PR #2545, opened 2021-09-26
[CALCITE-4805] Calcite should convert a small IN-list as if the user had written OR, even if the IN-list contains NULL

NobiGo:


core/src/test/resources/sql/sub-query.iq
```diff
@@ -3083,4 +3083,181 @@ EnumerableCalc(expr#0..4=[{inputs}], expr#5=[IS NOT NULL($t3)], proj#0..2=[{expr
             EnumerableTableScan(table=[[scott, EMP]])
 !plan
 
+# [CALCITE-4805] The IN sub-query can't be converted to OR when value list contains Null.
```
rubenada, line 3086: minor: this comment still contains the "old" title

NobiGo: done. Thanks for your review.

PR #2550, opened 2021-09-28
[CALCITE-4809] mixed merge join result isn't incomplete

mikasaakeman:


linq4j/src/main/java/org/apache/calcite/linq4j/EnumerableDefaults.java
```diff
@@ -4275,13 +4275,22 @@ private boolean advance() {
           // mergeJoin assumes inputs sorted in ascending order with nulls last,
           // if we reach a null key, we are done.
           if (leftKey == null || rightKey == null) {
-            if (joinType == JoinType.LEFT || (joinType == JoinType.ANTI && leftKey != null)) {
-              // all remaining items in left are results for left/anti join
-              remainingLeft = true;
-              return true;
+            if (rightKey == null || !(getLeftEnumerator() instanceof MergeJoinEnumerator
```
rubenada, line 4278: This ad-hoc code should not be needed, as long as MergeJoin inputs respect the MergeJoin pre-requisite: they must be sorted by join keys in ascending order nulls last.
I'd say the issue is not in this code, but on the module that creates the EnumerableMergeJoin, which perhaps does not introduce the appropriate Sort on its inputs.

mikasaakeman: In theory, there should be an orderby outside the inner left join, but there is no

rubenada: Then the problem is not here, it would be in the module that generates the plan.
Can you please try to produce a unit test that shows:
- The mismatched expected results vs actual result
- The generated plan, showing an EnumerableMergeJoin whose inputs are not sorted as they're supposed to

mikasaakeman: i see the plan is EnumerableMergeJoin(EnumerableMergeJoin(EnumerableSort(TableScan, EnumerableSort(TableScan)), EnumerableSort(TableScan))，but it shoud be EnumerableMergeJoin(EnumerableSort（EnumerableMergeJoin(EnumerableSort(TableScan, EnumerableSort(TableScan))), EnumerableSort(TableScan))

mikasaakeman: i see the plan is EnumerableMergeJoin(EnumerableMergeJoin(EnumerableSort(TableScan), EnumerableSort(TableScan)), EnumerableSort(TableScan)),but it should be EnumerableMergeJoin(EnumerableSort(EnumerableMergeJoin(EnumerableSort(TableScan), EnumerableSort(TableScan))), EnumerableSort(TableScan))

rubenada: Indeed, that looks like a problem.
We should try to find out the root cause of how / by which rule(s) that final plan was obtained. 
Perhaps there is an issue with the collations of EnumerableMergeJoin (or its inputs) ?

PR #2551, opened 2021-09-28
[CALCITE-4232] Support IN query in elasticsearch adapters

shlok7296:


elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/PredicateAnalyzer.java
```diff
@@ -148,6 +148,7 @@ private static boolean supportedRexCall(RexCall call) {
         case OR:
         case LIKE:
         case EQUALS:
+        case IN:
```
zabetak, line 151: If I remember well the `IN` operator has been superseded by `SEARCH` so this code may be unreachable at the moment.

zabetak (changes requested):
This seems to be a new feature so tests are definitely needed in order for this to get merged to master.

PR #2552, opened 2021-09-28
[CALCITE-4737] Add RelOptPlanner visualizer for debugging

thomasrebele:
An improved and refactored version of pull request https://github.com/apache/calcite/pull/2486.

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/visualizer/InputExcludedRelWriter.java
+++ b/core/src/main/java/org/apache/calcite/plan/visualizer/InputExcludedRelWriter.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.plan.visualizer;
+
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.RelWriter;
+import org.apache.calcite.sql.SqlExplainLevel;
+import org.apache.calcite.util.Pair;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * An implement of RelWriter for explaining a single RelNode.
+ * The result only contains the properties of the RelNode,
+ * but does not explain the children.
+ *
+ * <pre>{@code
+ * InputExcludedRelWriter relWriter = new InputExcludedRelWriter();
+ * rel.explain(relWriter);
+ * String digest = relWriter.toString();
+ * }</pre>
+ *
+ */
+class InputExcludedRelWriter implements RelWriter {
+
+  private final Map<String, @Nullable Object> values = new LinkedHashMap<>();
+
+  InputExcludedRelWriter() {
+  }
+
+
+  @Override public void explain(RelNode rel, List<Pair<String, @Nullable Object>> valueList) {
+    valueList.forEach(pair -> {
+      assert pair.left != null;
+      this.values.put(pair.left, pair.right);
+    });
+  }
+
+  @Override public SqlExplainLevel getDetailLevel() {
+    return SqlExplainLevel.EXPPLAN_ATTRIBUTES;
+  }
+
+  @Override public RelWriter input(String term, RelNode input) {
+    // do nothing, ignore input
+    return this;
+  }
+
+  @Override public RelWriter item(String term, @Nullable Object value) {
+    this.values.put(term, value);
+    return this;
+  }
+
+  @Override public RelWriter itemIf(String term, @Nullable Object value, boolean condition) {
+    if (condition) {
+      this.values.put(term, value);
+    }
+    return this;
+  }
+
+  @Override public RelWriter done(RelNode node) {
+    return this;
+  }
+
+  @Override public boolean nest() {
+    return false;
+  }
+
+  @Override public String toString() {
+    return values.toString();
+  }
+}
--- a/core/src/main/java/org/apache/calcite/plan/visualizer/RuleMatchVisualizer.java
+++ b/core/src/main/java/org/apache/calcite/plan/visualizer/RuleMatchVisualizer.java
@@ -1,0 +1,487 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.plan.visualizer;
+
+import org.apache.calcite.plan.RelOptCost;
+import org.apache.calcite.plan.RelOptListener;
+import org.apache.calcite.plan.RelOptPlanner;
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.hep.HepRelVertex;
+import org.apache.calcite.plan.volcano.RelSubset;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.metadata.RelMetadataQuery;
+
+import org.apache.commons.io.IOUtils;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Charsets;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.text.DecimalFormat;
+import java.text.MessageFormat;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * This is a tool to visualize the rule match process of a RelOptPlanner.
+ *
+ * <pre>{@code
+ * // create the visualizer
+ * RuleMatchVisualizer viz = new RuleMatchVisualizer("/path/to/output/dir", "file-name-suffix");
+ * viz.attachTo(planner)
+ *
+ * planner.findBestExpr();
+ *
+ * // extra step for HepPlanner: write the output to files
+ * // a VolcanoPlanner will call it automatically
+ * viz.writeToFile();
+ * }</pre>
+ */
+public class RuleMatchVisualizer implements RelOptListener {
+
+  private static final String INITIAL = "INITIAL";
+  private static final String FINAL = "FINAL";
+  public static final String DEFAULT_SET = "default";
+
+  // default HTML template can be edited at
+  // core/src/main/resources/org/apache/calcite/plan/visualizer/viz-template.html
+  private final String templateDirectory = "org/apache/calcite/plan/visualizer";
+  private final @Nullable String outputDirectory;
+  private final @Nullable String outputSuffix;
+
+  private String latestRuleID = "";
+  private int latestRuleTransformCount = 1;
+  private boolean initialized = false;
+
+  private @Nullable RelOptPlanner planner = null;
+
+  private boolean includeTransitiveEdges = false;
+  private boolean includeIntermediateCosts = false;
+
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/apache/calcite/plan/volcano/VolcanoRuleMatchVisualizer.java
```diff
@@ -0,0 +1,312 @@
...
+ *
+ * // writes the output to files
+ * visualizer.writeToFile(outputDirectory, "");
+ * </pre>
+ */
+public class VolcanoRuleMatchVisualizer {
+
+  public static VolcanoRuleMatchVisualizer createAndAttach(VolcanoPlanner volcanoPlanner) {
+    VolcanoRuleMatchVisualizer vis = new VolcanoRuleMatchVisualizer(volcanoPlanner);
+    // add listener outside of the constructor to make checker-framework happy
```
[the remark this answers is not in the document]
zabetak: I guess that if we incorporate by default in Calcite the method `createAndAttach` can be removed.

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoRuleMatchVisualizer.java
```diff
@@ -0,0 +1,312 @@
...
+ * // create the visualizer. This attaches a listener to VolcanoPlanner
+ * VolcanoRuleMatchVisualizer visualizer = VolcanoRuleMatchVisualizer.createAndAttach(planner);
+ *
+ * volcanoPlanner.findBestExpr();
+ *
+ * // writes the output to files
+ * visualizer.writeToFile(outputDirectory, "");
+ * </pre>
+ */
+public class VolcanoRuleMatchVisualizer {
```
zabetak, line 74: I am wondering if we can somehow unify this class with the `VolcanoRuleMatchVisualizerListener`? Actually, the main thing that I would like to avoid if possible is to have state dependent on the `VolcanoPlanner`. I see that we need the `VolcanoPlanner` to obtain some information about sets, subsets, etc., so I am wondering if we could obtain this information from `RelOptListener#relEquivalenceFound` API instead. The goal is to make this a bit more generic and dependent on the `RelOptListener` API instead of `VolcanoPlanner`.

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoRuleMatchVisualizer.java
```diff
@@ -0,0 +1,312 @@
...
+  List<String> ruleMatchSequence = new ArrayList<>();
+  // map of ruleMatch ID and the info, including the state snapshot at the time of ruleMatch
+  Map<String, VisualizerRuleMatchInfo> ruleInfoMap = new TreeMap<>();
+  // map of nodeID to the ruleID it's first added
+  Map<String, String> nodeAddedInRule = new TreeMap<>();
+
+  // a map of relNode ID to the actual RelNode object
+  // contains all the relNodes appear during the optimization
+  // all RelNode are immutable in Calcite, therefore only new nodes will be added
+  Map<String, RelNode> allNodes = new TreeMap<>();
```
zabetak, line 95: Reduce visibility?

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoRuleMatchVisualizer.java
```diff
@@ -0,0 +1,312 @@
...
+   *
+   * @param outputDirectory directory of the output files
+   * @param suffix          file name suffix, can be null
+   */
+  public void writeToFile(String outputDirectory, String suffix) {
+    // default HTML template is under "resources"
+    writeToFile("volcano-viz", outputDirectory, suffix);
+  }
+
+  public void writeToFile(String templateDirectory, String outputDirectory, String suffix) {
```
zabetak, line 273: Instead of forcing the client to keep a reference to the visualizer and call these methods explicitly after optimization can't we do it automatically at the end of the optimization phase (e.g., by relying on `RelOptListener#relChosen`)? 
If this gets integrated to the planner we should find a trick to avoid replacing the visualizations of previous queries; in other words avoid writing to the same file/directory.

core/src/main/java/org/apache/calcite/tools/visualizer/InputExcludedRelWriter.java
```diff
@@ -0,0 +1,90 @@
...
+ * <p>
+ *
+ * <pre>
+ * InputExcludedRelWriter relWriter = new InputExcludedRelWriter();
+ * rel.explain(relWriter);
+ * String digest = relWriter.toString();
+ * </pre>
+ *
+ */
+public class InputExcludedRelWriter implements RelWriter {
```
zabetak, line 43: Reduce visibility?

core/src/main/java/org/apache/calcite/tools/visualizer/VisualizerNodeInfo.java
```diff
@@ -0,0 +1,102 @@
...
+
+import java.text.DecimalFormat;
+import java.text.MessageFormat;
+import java.util.Locale;
+
+/**
+ * VisualizerNodeInfo is a helper data class for Volcano Visualizer.
+ * This class will be serialized to JSON and read by the Javascript file.
+ */
+public class VisualizerNodeInfo {
```
zabetak, line 31: Reduce visibility (if possible)?

core/src/main/java/org/apache/calcite/tools/visualizer/VolcanoRuleMatchVisualizerListener.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Override public void relDiscarded(RelDiscardedEvent event) {
+  }
+
+  @Override public void relEquivalenceFound(RelEquivalenceEvent event) {
+  }
+
+  public VolcanoRuleMatchVisualizer getVisualizer() {
+    return visualizer;
+  }
```
zabetak, line 95: Never called. Remove?

core/src/main/resources/volcano-viz/volcano-viz-data.js
```diff
@@ -0,0 +1,135 @@
...
+			explanation: "table=t2",
+			finalCost: "20cpu, 20io"
+		},
+		"node4": {
+			label: "IndexTableScan-4",
+			isSubset: false,
+			explanation: "table=t1, condition=c",
+			finalCost: "10cpu, 10io"
+		},
+	},
```
zabetak, line 46: A more realistic example from an actual run would be more easy to understand. The query doesn't have to be complicated just reflect actual classes/rules in Calcite.

zabetak (commented):
This is great work @zuozhiw @thomasrebele ! I would really like to see this merged. I think it is quite close. Can someone go over the comments and address them if possible?

core/src/main/resources/volcano-viz/volcano-viz-data.js
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// sample generated visualizer data used by the visualizer JavaScript code
+var data = {
```
vlsi, line 19: Should this be in `src/test/resources` ?

.gitignore
```diff
@@ -52,6 +52,8 @@ settings.xml
 nb-configuration.xml
 */nb-configuration.xml
 
+core/volcano-viz/*
+
```
vlsi, line 56: Is this needed?

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoRuleMatchVisualizer.java
```diff
@@ -0,0 +1,312 @@
...
+    volcanoPlanner.addListener(new VolcanoRuleMatchVisualizerListener(vis));
+    return vis;
+  }
+
+  VolcanoPlanner volcanoPlanner;
+
+  // a sequence of ruleMatch ID to represent the order of rule match
+  List<String> ruleMatchSequence = new ArrayList<>();
+  // map of ruleMatch ID and the info, including the state snapshot at the time of ruleMatch
+  Map<String, VisualizerRuleMatchInfo> ruleInfoMap = new TreeMap<>();
```
vlsi, line 88: Is there a need for `TreeMap`?
`HashMap` is faster and it consumes less memory.
If you want consistent iteration order, then prefer `LinkedHashMap`.

thomasrebele:
Thanks for the feedback. I'll try to make the visualizer more general by making it independent of the VolcanoPlanner. This might take a while.

zabetak:
> Thanks for the feedback. I'll try to make the visualizer more general by making it independent of the VolcanoPlanner. This might take a while.

@thomasrebele Only if you think it makes sense.

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoRuleMatchVisualizer.java
thomasrebele: I've completely refactored the Java code of the visualizer, removing the dependency on the `VolcanoPlanner`. With some slight adaptions the visualizer even works for the `HepPlanner`.

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoRuleMatchVisualizer.java
thomasrebele: I've changed it so that the method will be called automatically when using a `VolcanoPlaner`. It has to be called manually for a `HepPlanner`. The directory and the suffix will have to be chosen when creating the visualizer.

.gitignore
thomasrebele: I've removed it.

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoRuleMatchVisualizer.java
thomasrebele: Now using hash maps everywhere.

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoRuleMatchVisualizer.java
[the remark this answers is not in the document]
thomasrebele: Indeed, adding a property would make it easier to use the visualizer. To be discussed in the [Jira ticket](https://issues.apache.org/jira/browse/CALCITE-4737?focusedCommentId=17459849#comment-17459849)

core/src/main/resources/volcano-viz/volcano-viz-data.js
thomasrebele: A unit test has been added to check the json output of the visualizer.

core/src/main/resources/volcano-viz/volcano-viz-data.js
thomasrebele: I've removed the file completely. Examples of the output format can be seen in core/src/test/resources/org/apache/calcite/test/RuleMatchVisualizerTest.xml.

julianhyde (commented):
i've suggested a bunch of small changes, but this change looks very close.

core/src/main/java/org/apache/calcite/plan/visualizer/RuleMatchVisualizer.java
```diff
@@ -0,0 +1,479 @@
...
+  // core/src/main/resources/volcano-viz/viz-template.html
+  private final String templateDirectory = "volcano-viz";
+  private final String outputDirectory;
+  private final String outputSuffix;
+
+  private String latestRuleID = "";
+  private int latestRuleTransformCount = 1;
+  private boolean initialized = false;
+
+  private @Nullable RelOptPlanner planner = null;
```
zabetak, line 90: Do you need the planner only to obtain the root? If yes then maybe we can consider enriching the `RelOptListener` interface to provide this information (root, or initial plan, or whatever is necessary) so that we can drop this dependency with the listener? WDYT @thomasrebele ?

core/src/main/java/org/apache/calcite/plan/visualizer/InputExcludedRelWriter.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This RelWriter is indented to be used for getting a digest of a relNode,
+ *  excluding the field of the relNode's inputs.
+ * The result digest of the RelNode only contains its own properties.
+ * <p>
+ *
+ * <pre>
```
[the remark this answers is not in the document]
vlsi: I wonder if we can factor out this into an autostyle regexp. Frankly speaking, I am always in trouble when it comes to javadoc formatting.

core/src/main/java/org/apache/calcite/plan/visualizer/RuleMatchVisualizer.java
thomasrebele: It is also needed for getting the cost, so I would leave it as it is.

core/src/test/resources/org/apache/calcite/test/RuleMatchVisualizerTest.xml
```diff
@@ -0,0 +1,264 @@
...
+      LogicalProject(NAME=[$1])
+        LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
+    LogicalSort(sort0=[$0], dir0=[ASC], fetch=[10])
+      LogicalProject(NAME=[$1])
+        LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
+]]>
+    </Resource>
+    <Resource name="visualizer">
+      <![CDATA[{
+  "steps" : [ {
```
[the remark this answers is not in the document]
thomasrebele: Spaces removed to save some more bytes in the output files. The standard allows any number of spaces before and after the colon (https://www.json.org/json-en.html).

thomasrebele:
All remarks have been addressed. Additionally I've replaced the remaining references to the "Volcano" visualizer with a more general wording. All commits have been squashed into a single commit. The stalling Druid Test seems to be a more general issue (see https://github.com/apache/calcite/actions/runs/1580165192 on master).

Please let me know if you have any further comment. If nobody objects, could a committer merge the PR into master to have this feature in the upcoming release 1.29?

core/src/main/resources/org/apache/calcite/plan/visualizer/viz-template.html
```diff
@@ -0,0 +1,421 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+{% endcomment %}
+-->
+<meta charset="utf-8">
+<title>Calcite Rule Match Visualization</title>
+
+<script src="https://d3js.org/d3.v7.min.js" charset="utf-8"></script>
+<script src="https://dagrejs.github.io/project/dagre-d3/latest/dagre-d3.min.js"></script>
```
vlsi, line 25: Is there a way to use a fixed-version link?

thomasrebele: I don't know about any. Open to suggestions.

rubenada:
Thanks @thomasrebele for pushing this forward, and everyone else involved in the review.
I think the PR is in a good shape. If nobody objects, I will merge it at the end of the day, so that it will be part of the upcoming release 1.29

PR #2555, opened 2021-09-29
[CALCITE-4777]Casting from DECIMAL to BOOLEAN throws an exception

xuyangzhong:
According to the discussion in [calcite-4777], casting from decimal to boolean is invalid, and an exception like "Cast function cannot convert value of type DECIMAL(2,1) to type BOOLEAN" will be thrown in the period of validating. 
This pr changes:

- remove the coerce rule from decimal to boolean, which will lead to throw a conversion exception and let users know this is an invalid conversion.
- add some test cases in SqlOperatorBaseTest.
- change the reference.md to say this invalid conversion

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/type/SqlTypeCoercionRule.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/SqlTypeCoercionRule.java
@@ -202,7 +202,7 @@ public class SqlTypeCoercionRule implements SqlTypeMappingRule {
         coerceRules.copyValues(SqlTypeName.BOOLEAN)
             .add(SqlTypeName.CHAR)
             .add(SqlTypeName.VARCHAR)
-            .addAll(SqlTypeName.NUMERIC_TYPES)
+            .addAll(SqlTypeName.INT_TYPES)
             .build());
 
     // DATE, TIME, and TIMESTAMP are castable from
--- a/core/src/test/java/org/apache/calcite/sql/test/SqlOperatorBaseTest.java
+++ b/core/src/test/java/org/apache/calcite/sql/test/SqlOperatorBaseTest.java
@@ -202,6 +202,9 @@ public abstract class SqlOperatorBaseTest {
   public static final String INVALID_ARGUMENTS_NUMBER =
       "Invalid number of arguments to function .* Was expecting .* arguments";
 
+  public static final String INVALID_CAST_CONVERSION_ERROR =
+      "Cast function cannot convert value of type .* to type .*";
+
   public static final boolean TODO = false;
 
   /**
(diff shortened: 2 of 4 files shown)


core/src/test/java/org/apache/calcite/sql/test/SqlOperatorBaseTest.java
```diff
@@ -1595,6 +1598,17 @@ protected static Calendar getCalendarNotTooNear(int timeUnit) {
     tester.checkFails(
         "cast(cast('blah' as varchar(10)) as boolean)", INVALID_CHAR_MESSAGE,
         true);
+
+    tester.checkBoolean("cast(0 as boolean)", false);
```
rubenada, line 1602: For consistency, it should be `Boolean.FALSE` instead of `false`

core/src/main/java/org/apache/calcite/sql/type/SqlTypeCoercionRule.java
```diff
@@ -202,7 +202,8 @@ private SqlTypeCoercionRule(Map<SqlTypeName, ImmutableSet<SqlTypeName>> map) {
         coerceRules.copyValues(SqlTypeName.BOOLEAN)
             .add(SqlTypeName.CHAR)
             .add(SqlTypeName.VARCHAR)
-            .addAll(SqlTypeName.NUMERIC_TYPES)
+            .addAll(SqlTypeName.INT_TYPES)
+            .addAll(SqlTypeName.APPROX_TYPES)
```
rubenada, line 206: `SqlTypeName.APPROX_TYPES` include `FLOAT`, `REAL` and `DOUBLE`. Are we sure casting these types into boolean always work?
It's strange, it seems that `SELECT CAST(CAST(0 AS float) AS BOOLEAN)` works fine (returns `false`). But if we try a more "complex" query: `select cast(cast(p0 as float) as boolean) from (values (0)) as t(p0)` ; it will fail at runtime with `CompileException: Line 21, Column 31: Cannot cast "double" to "boolean"`.
This type of query will be generated if we add these tests into `SqlOperatorBaseTest#testCastToBoolean`:
```
tester.checkBoolean("cast(cast(0 as float) as boolean)", Boolean.FALSE);
tester.checkBoolean("cast(cast(0 as real) as boolean)", Boolean.FALSE);
tester.checkBoolean("cast(cast(0 as double) as boolean)", Boolean.FALSE);
```
which, if I am not mistaken, will fail if we add them to `SqlOperatorBaseTest#testCastToBoolean`. Therefore, perhaps we should consider `FLOAT`, `REAL` and `DOUBLE` (together with `DECIMAL`) as non-castable into boolean?

core/src/test/java/org/apache/calcite/sql/test/SqlOperatorBaseTest.java
```diff
@@ -1595,6 +1598,17 @@ protected static Calendar getCalendarNotTooNear(int timeUnit) {
...
+
```
rubenada, line 1603: Perhaps we could add also the test for the "other int types"?
```
tester.checkBoolean("cast(cast(0 as tinyint) as boolean)", Boolean.FALSE);
tester.checkBoolean("cast(cast(0 as smallint) as boolean)", Boolean.FALSE);
tester.checkBoolean("cast(cast(0 as bigint) as boolean)", Boolean.FALSE);
```

core/src/test/java/org/apache/calcite/sql/test/SqlOperatorBaseTest.java
xuyangzhong: got it

core/src/main/java/org/apache/calcite/sql/type/SqlTypeCoercionRule.java
xuyangzhong: About this bug I create a JIRA yesterday(in https://issues.apache.org/jira/browse/CALCITE-4810) before this pr.
I tested the sql "tester.checkBoolean("cast(cast(0.0 as double) as boolean)", Boolean.FALSE);" in master where i didn't apply my change and it threw the same exception just like you says.
I tried to find the reason and add the message into [CALCITE-4810]. It seems the bug that these two behavoirs codegen and constant simplifying is not unified.
I think your suggestion that avoids casting from float, real, double and decimal to boolean is reasonable and this also can solve this problem of inconsistent behavior. And more importantly, it follows the SQL standard.

core/src/test/java/org/apache/calcite/sql/test/SqlOperatorBaseTest.java
xuyangzhong: got it

core/src/main/java/org/apache/calcite/sql/type/SqlTypeCoercionRule.java
rubenada: IMO it makes sense, if we say that `DECIMAL` cannot be cast into `BOOLEAN`; consider the same for `FLOAT`, `REAL` and `DOUBLE`, I think it is a consistent approach.
However, if we confirm to go in this direction:
- It seems some test will need adjustments
- The cast matrix in `reference.md` will need to be modified for those rows too.

rubenada:
LGTM
@xuyangzhong the PR looks in a good shape, could you please squash commits into a single one so that we can merge it shortly?

xuyangzhong:
@rubenada  done

PR #2558, opened 2021-10-02
[CALCITE-4821] Move utility test classes into calcite-testkit and unpublish -test.jar

vlsi:


vlsi:
Just in case, I've logged :server:test failure as https://issues.apache.org/jira/browse/CALCITE-4824

PR #2560, opened 2021-10-04
CALCITE-4818 AggregateExpandDistinctAggregatesRule must infer correct data type for top aggregate calls

tledkov-gridgain:


wojustme:
+1

PR #2561, opened 2021-10-04
[CALCITE-4826] Set destination directory for core annotation processing

jacques-n:
Set destination directory for core annotation processing task

This task, used in the IDE, needs the property to be set even though the -proc:only property means no code will be compiled.


jacques-n:
Included this change in the original patch set via force push. Abandoning.

PR #2563, opened 2021-10-04
[CALCITE-4827] Move annotation processor configuration to base module

jacques-n:
This will allow use in multiple modules (not just core) with minimal repetition.


jacques-n:
@vlsi, would love your help on this. I'd like to avoid having to duplicate the afterSync stuff in each subproject but don't know enough gradle/kotlin to figure this out.

vlsi:
Thank you for looking into this, however, I would like to refrain from making the change now.
Currently, the task definition and configuration are co-located, so they are easier to discover and understand. If you move a half of the configuration to a different file, then it becomes harder to reason about.

I'm going to merge Gradle 7.2 migration soon, and then we'll be able to use https://docs.gradle.org/current/userguide/structuring_software_products.html approach, and virtually all code from the root file would be split across feature plugins.

PR #2565, opened 2021-10-06
Hang on build when using kapt

jacques-n:


jacques-n:
@vlsi , this appears to cause a gradle hang. You can see the jstack here: 
https://gist.github.com/jacques-n/417bcae348fafa79413d1b1aff9eabbb

Any help on why? Switching to kapt is necessary at a minimum in the core module in order to work with the kotlin test annotations combined with the annotation processing required in that test module.

vlsi:
It is going to fixed in the next kotlin-gradle-plugin version: https://youtrack.jetbrains.com/issue/KT-47853

PR #2566, opened 2021-10-06
[CALCITE-4832] Refactor Gradle to build-logic convention plugins

vlsi:
See https://docs.gradle.org/7.3/userguide/structuring_software_products.html
See https://github.com/jjohannes/idiomatic-gradle

This is WIP at the moment, however, the idea is that all the build logic moves to `build-logic/*`


build-logic/code-quality/src/main/kotlin/calcite.testing.gradle.kts
```diff
@@ -0,0 +1,69 @@
...
+
+plugins {
+    `java-library`
+}
+
+dependencies {
+    // TODO: add junit5 by default
+}
+
+tasks.withType<Test>().configureEach {
```
zinking, line 28: does it mean this occurs before in every subproject and now centralize here?

vlsi: The file is named `calcite.testing.gradle.kts`.
This code will be activated only for those `.gradle.kts` files that include

```kotlin
plugins {
    id("calcite.testing") // <-- gradle would apply code from calcite.testing.gradle.kts
}
```

It allows composition of the build logic, so the individual submodules would mention the bits they need in `plugins` section.

zinking:
always good to learn these new insights from you. thanks @vlsi

build-logic/code-quality/src/main/kotlin/calcite.testing.gradle.kts
vlsi: In other words, build files would never "apply automatically", and build logic would always be pulled by `plugins { .. }` section

vlsi:
I'm not interested in moving this forward

PR #2571, opened 2021-10-07
[CALCITE-4833] Complex nested correlated subquery failed.

zstan:


jamesstarr (commented):
I think are more generalized approach should be taken.  Very few queries will be written in way that this is handled.

PR #2572, opened 2021-10-08
[CALCITE-4841] Support numeric column type in CSV and File adapter.

xwkuang5:
- Introduce a new `NUMERIC` type in CsvFieldType and FileFieldType for
  more accurate representation of decimal data.
- Map `NUMERIC` type to Java BigDecimal type.


xwkuang5:
Done. Added the abilities to specify `decimal(precision/scale)` in the CSV header. PTAL, thank!

file/src/test/resources/sales-csv/DECIMAL.csv
```diff
@@ -0,0 +1,4 @@
+DEPTNO:int,BUDGET:decimal(18/2)
```
jacques-n, line 1: The standard pattern in most sql systems is a comma between scale and precision. It's really weird that you're using a slash. Can you make this a comma?

file/src/test/java/org/apache/calcite/adapter/file/FileAdapterTest.java
```diff
@@ -619,6 +621,72 @@ private String range(int first, int count) {
...
+            assertEquals(18, metaData.getPrecision(1));
+            assertEquals(2, metaData.getScale(1));
+          } catch (SQLException e) {
+            throw TestUtil.rethrow(e);
+          }
+        })
+        .ok();
+  }
+
+  @Test void testDecimalTypeArithmeticOperations() {
```
jacques-n, line 639: Can you also test cases for negative cases and out of space numbers? For example a CSV has 0.0000000000001 or 10000 but the it is declared as DECIMAL(4,1).

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
```diff
@@ -130,30 +129,108 @@ public static RelDataType deduceRowType(JavaTypeFactory typeFactory,
...
+            int precision = Integer.parseInt(decimalMatcher.group(1));
+            int scale = Integer.parseInt(decimalMatcher.group(2));
+            fieldType = typeFactory.createTypeWithNullability(typeFactory
+                .createSqlType(SqlTypeName.DECIMAL,
+                    precision, scale), true);
+          } else {
+            switch (typeString) {
+            case "string":
+              fieldType = typeFactory.createTypeWithNullability(typeFactory
+                      .createSqlType(typeFactory.createJavaType(String.class).getSqlTypeName()),
```
jacques-n, line 148: Please just use the SqlTypeName enum directly as opposed to the intermediate step of createJavaType()

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
```diff
@@ -130,30 +129,108 @@ public static RelDataType deduceRowType(JavaTypeFactory typeFactory,
...
+                  true);
+              break;
+            case "timestamp":
+              fieldType = typeFactory.createTypeWithNullability(
+                  typeFactory.createSqlType(
+                      typeFactory.createJavaType(java.sql.Timestamp.class).getSqlTypeName()),
+                  true);
+              break;
+            default:
+              System.out.println("WARNING: Found unknown type: "
```
jacques-n, line 218: I think this should be converted into a logging statement as opposed to a system.out (I know it was here before this change but we should clean this kind of thing up when working on the code.)

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
```diff
@@ -130,30 +129,108 @@ public static RelDataType deduceRowType(JavaTypeFactory typeFactory,
...
+                    precision, scale), true);
+          } else {
+            switch (typeString) {
+            case "string":
+              fieldType = typeFactory.createTypeWithNullability(typeFactory
+                      .createSqlType(typeFactory.createJavaType(String.class).getSqlTypeName()),
+                  true);
+              break;
+            case "boolean":
+              fieldType = typeFactory.createTypeWithNullability(
```
jacques-n, line 152: This pattern seems very redundant. It seems like 99% of the time, the pattern is string => sqlTypeName (everything except decimal). Can your refactor to avoid all the repetition?

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
```diff
@@ -284,6 +361,17 @@ static CSVReader openCsv(Source source) throws IOException {
...
         }
         return Double.parseDouble(string);
+      case DECIMAL:
+        if (string.length() == 0) {
+          return null;
+        }
+        BigDecimal result = new BigDecimal(string);
+        if (result.precision() > fieldType.getPrecision() || result.scale() != fieldType
+            .getScale()) {
+          throw new IllegalArgumentException(
```
jacques-n, line 371: Does throwing make sense here? Or should we coerce?

jacques-n: Also, best to use string format for this kind of thing as opposed to string concatenation.

jacques-n (changes requested):

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
xwkuang5: Done. Based coercion on PostgreSQL's implementation https://www.postgresql.org/docs/13/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeFactoryImpl.java
```diff
@@ -84,6 +84,10 @@ public SqlTypeFactoryImpl(RelDataTypeSystem typeSystem) {
         || (precision == RelDataType.PRECISION_NOT_SPECIFIED);
     final int maxPrecision = typeSystem.getMaxPrecision(typeName);
     if (maxPrecision >= 0 && precision > maxPrecision) {
+      System.out.println("WARNING: specified precision (" + precision
```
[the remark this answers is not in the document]
xwkuang5: Removed for now.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
xwkuang5: Done.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
xwkuang5: Refactored to a helper method now.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
xwkuang5: Done.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
xwkuang5: Done. Based coercion on PostgreSQL's implementation https://www.postgresql.org/docs/13/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL.

file/src/test/java/org/apache/calcite/adapter/file/FileAdapterTest.java
xwkuang5: Done. This has been added in the new CsvEnumeratorTest file as part of testing the parseDecimal method.

file/src/test/resources/sales-csv/DECIMAL.csv
xwkuang5: Done. Added double quote to disambiguate between column separator and the precision and scale separator.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
```diff
@@ -130,30 +137,70 @@ public static RelDataType deduceRowType(JavaTypeFactory typeFactory,
...
+            case "timestamp":
+              fieldType = toNullableRelDataType(typeFactory, SqlTypeName.TIMESTAMP);
+              break;
+            case "time":
+              fieldType = toNullableRelDataType(typeFactory, SqlTypeName.TIME);
+              break;
+            default:
+              LOGGER.warn(
+                  String.format(Locale.ROOT,
+                  "Found unknown type: %s in file: %s for column: %s. Will assume the type of "
```
jacques-n, line 191: when using a logger, you should use the loggers capability for value replacement as opposed to string format.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
```diff
@@ -130,30 +137,70 @@ public static RelDataType deduceRowType(JavaTypeFactory typeFactory,
...
+                      + "column is string.",
+                  typeString, source.path(), name));
+              fieldType = toNullableRelDataType(typeFactory, SqlTypeName.VARCHAR);
```
jacques-n, line 194: Shouldn't you set a max length on this varchar? (same for the one below)

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
```diff
@@ -314,22 +371,55 @@ static CSVReader openCsv(Source source) throws IOException {
...
+      case VARCHAR:
       default:
         return string;
       }
     }
   }
 
+  private static RelDataType parseDecimalSqlType(JavaTypeFactory typeFactory, int precision,
+      int scale) {
+    if (scale < 0) {
```
jacques-n, line 383: It seems like you should do some other checks here as well. For example, can precision be negative? What are the requirements for a relationship between precision and scale?

jacques-n (commented):
Looking pretty good. Thanks for the all updates. Just a few more things

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
xwkuang5: Done.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
xwkuang5: The current code does not set a max on the varchar either. They used the default in the type system https://github.com/apache/calcite/blob/6d51d277b158ff7073f29ada4a96a7a74c0b46fc/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystemImpl.java#L62.

This is a problem for file adapter since there is no validation for any of the types' precision during enumeration (except for the decimal type being added here). I think this should be addressed in a different PR.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
xwkuang5: Done. The following should be enough:
1. precision > 0: 0 has precision 1 and scale 0
2. scale >= 0
3. precision >= scale

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
jacques-n: Usually loggers use {}. Isn't that what should be used (instead of %s?)

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
jacques-n: makes sense.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
jacques-n: Do you want to update these to use Preconditions? I think it has convenience methods to use format strings (although I think it simplifies to %s for everything).

xwkuang5: Done.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
xwkuang5: Done.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
```diff
@@ -314,22 +372,56 @@ static CSVReader openCsv(Source source) throws IOException {
...
+        typeFactory.createSqlType(SqlTypeName.DECIMAL, precision, scale), true);
+  }
+
+  @VisibleForTesting
+  protected static BigDecimal parseDecimal(int precision, int scale, String string) {
+    BigDecimal result = new BigDecimal(string);
+    // If the parsed value has more fractional digits than the specified scale, round ties away
+    // from 0.
+    if (result.scale() > scale) {
+      LOGGER.warn(
```
jacques-n, line 399: This one is still using wrong pattern %s versus {}

jacques-n (approved):
LGTM, will merge after clean test runs. Nice work @xwkuang5 !

xwkuang5:
> LGTM, will merge after clean test runs. Nice work @xwkuang5 !

Thank you for the careful review and guidance!

jacques-n:
Merged in 065e323da89ca73a46bec195efc7bd4019dc4f2f

PR #2573, opened 2021-10-09
[CALCITE-4842] Support nested row type

snuyanzin:
The PR support nested row type for instance
```sql
select row(row(1, 2), row(3, 4), row(5, 6, 7));
```


snuyanzin:
closing since it was fixed within a different commit before 1.32.0 released

PR #2577, opened 2021-10-11
To_number support for column

vinayujeedm:
TO_number support for columns.


snuyanzin:
It looks there are lots of commits from other fork...
It would be great if you rebase your changes and leave here only those which are related to the current issue

PR #2581, opened 2021-10-13
[CALCITE-4848] Adding a HAVING condition to a query with a dynamic parameter makes the result always empty

thomasrebele:


chunweilei:
LGTM

PR #2587, opened 2021-10-18
[CALCITE-4858] Use Log4j2 instead of unsupported Log4j (1.x) in tests

zabetak:
1. Replace slf4j-log4j12 dependency (using Log4j 1.x)  with
log4j-slf4j-impl (using Log4j 2.x) when possible and change
log4j.properties to log4j2-test.xml. Cannot remove Log4j 1.x from
Pig/Piglet modules due to compilation dependency of the latter directly
to log4j.
2. Exclude slf4j-log4j12 when it comes transitively from other
dependencies. It gets in conflict with log4j-slf4j-impl and it may also
appear transitively in projects using Calcite.
3. Add log4j-slf4j-impl in modules (kafka, mongodb, plus, example) to
silence error messages due to missing logger implementation.
4. Always use testRuntimeOnly annotation for dependencies on SLF4J
logger implementation. Calcite shouldn't enforce a specific logger
implementation so they should never appear as compile or runtime
dependencies. A concrete implementation is only needed when we want to
log messages for testing purposes.
5. Bump Cassandra version to 3.11.3 to use Log4j2 in tests. Previous
versions did not allow the use of any custom logger implementation
(only logback). This was resolved CASSANDRA-13396 so now we can use
Log4j2 as the rest of the tests.
6. Unify display pattern in elasticsearch log4j2.xml with the rest of
tests.
6. Add/Modify Log4j2 configuration (log4j2-test.xml) to silence errors
and not display logging messages in tests. Instead of globally turning
off logs, specific logger entries were set in the configuration files:
(i) to avoid hiding by accident important problems;
(ii) to faciliate switching log levels during debugging in the future.

### DIFF
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -147,6 +147,7 @@ dependencies {
         apiv("org.postgresql:postgresql")
         apiv("org.scala-lang:scala-library")
         apiv("org.slf4j:slf4j-api", "slf4j")
+        // TODO: https://issues.apache.org/jira/browse/CALCITE-4862
         // Eventually we should get rid of slf4j-log4j12 dependency but currently it is not possible
         // since certain modules (Pig, Piglet) have dependencies using directly Log4j 1.x APIs
         runtimev("org.slf4j:slf4j-log4j12", "slf4j")


zinking:
looks good overall, thanks for the good work.

PR #2594, opened 2021-10-29
[CALCITE-4844] IN-list that references columns is wrongly converted to Values, and gives incorrect results

NobiGo:
…o Values, and gives incorrect results

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
@@ -48,6 +48,7 @@ import java.io.Reader;
 import java.io.Writer;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.math.BigDecimal;
 import java.net.URL;
 import java.sql.Connection;
 import java.sql.DriverManager;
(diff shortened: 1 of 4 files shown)


testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
```diff
@@ -150,6 +151,10 @@ protected void checkRun(String path) throws Exception {
                   && (Boolean) value;
               closer.add(Prepare.THREAD_EXPAND.push(b));
             }
+            if (propertyName.equals("insubquerythreshold")) {
```
zinking, line 154: is this temporary ? surprised code gen part follows naturely and requires no change.

NobiGo: No. This can be used in the .iq file to set the insubquerythreshold property.

NobiGo: `!set insubquerythreshold 0`

chunweilei (commented):
LGTM

testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
NobiGo: @zinking Please review again, Then we can merge.

PR #2598, opened 2021-11-04
[CALCITE-4875] Preserve Operand Nullability in NVL rewrite

jaystarshot:


core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -308,12 +308,20 @@ private static RexNode convertNvl(SqlRexContext cx, SqlCall call) {
...
+            rexBuilder.makeCast(
+                cx.getTypeFactory()
+                    .createTypeWithNullability(type, operand0.getType().isNullable()),
+                operand0),
+            rexBuilder.makeCast(
+                cx.getTypeFactory()
+                    .createTypeWithNullability(type, operand1.getType().isNullable()),
+                operand1)
+        ));
   }
```
chunweilei, line 325: Could you add a test case, @jaystarshot ?

jaystarshot: @chunweilei  There is no method to test the Operands of an Operator in the test framework, I think I need to create a new one and/or test it as an integration test.

core/src/test/java/org/apache/calcite/sql/test/SqlOperatorBaseTest.java
```diff
@@ -7202,6 +7218,41 @@ void assertSubFunReturns(boolean binary, String s, int start,
...
+        .build();
+    Planner planner = Frameworks.getPlanner(config);
+    SqlNode parse = planner.parse(sql);
+    SqlNode validate = planner.validate(parse);
+    RelRoot planRoot = planner.rel(validate);
+    return planRoot.rel;
+
+  }
+
+  @Test public void testNvlOperands() throws Exception {
```
NobiGo, line 7240: The testNvlFunc has existed in SqlOperatorTest, If it doesn't have enough tests for you, please perfect it.

If this can't suit your need. You can add a SQL type test in .iq file.

NobiGo (changes requested):

core/src/test/java/org/apache/calcite/sql/test/SqlOperatorBaseTest.java
jaystarshot: The testNvlFunc tests the function output and not the operands, same with .iq tests. Am I missing something?

NobiGo: You try to !plan command in .iq file. Same as https://github.com/apache/calcite/blob/master/core/src/test/resources/sql/agg.iq#L1707 .

NobiGo: Maybe in SqlToRelConverterTest. I think this test case in here is not standard.

jaystarshot: Thanks @NobiGo. I am new here so when I run the tests in a .iq file I get 
`No match found for function signature NVL(<CHARACTER>, <CHARACTER>)` . I guess its because NVl is supported only for Oracle Operator Tables. Is there a way to load it in the .iq tests?

jaystarshot: I tried to change org/apache/calcite/prepare/CalcitePrepareImpl.java:721 and added  oracle operators there, 
```
final SqlOperatorTable opTab0 =
    context.config().fun(SqlOperatorTable.class,
        SqlLibraryOperatorTableFactory.INSTANCE
            .getOperatorTable(SqlLibrary.STANDARD, SqlLibrary.ORACLE));
```
But i think we just want it configured for the Quidem tests and not in general.

NobiGo: I noticed the changes in SQL to RelNode So maybe SqlToRelConverterTest(The test result in SqlToRelConverterTest.xml) is more appropriate. Of course, If you think .iq is enough for this, You can try` !use oraclefunc ` expamle as https://github.com/apache/calcite/blob/master/core/src/test/resources/sql/functions.iq#L83 .

jaystarshot: Thanks a lot! Yes I think .iq test is enough as we are able to verify the type of the operand using the plan

core/src/test/resources/sql/functions.iq
```diff
@@ -179,5 +179,15 @@ SELECT EXISTSNODE(
...
 
+# [CALCITE-4875] Preserve Operand Nullability in NVL rewrite
+# Asserting that NVL does not change a Nullable operand to NOT Nullable
+
+!use oraclefunc
+select nvl("name", 'undefined') FROM "hr"."emps";
+
+EnumerableCalc(expr#0..4=[{inputs}], expr#5=[IS NOT NULL($t2)], expr#6=[CAST($t2):VARCHAR], expr#7=['undefined':VARCHAR], expr#8=[CASE($t5, $t6, $t7)], EXPR$0=[$t8])
+  EnumerableTableScan(table=[[hr, emps]])
+!plan
```
NobiGo, line 190: Can you add extra ` !ok` command to test the output?  Others is good for me.

NobiGo:
@jaystarshot  Can you add describe this plan in the Jira. (Before this PR and After) This can help other contributer understand what you resolve quickly.

jaystarshot:
@NobiGo Done

NobiGo (approved):

NobiGo (approved):

jaystarshot:
@NobiGo @chunweilei  Can we merge this PR?

NobiGo:
> @NobiGo @chunweilei  Can we merge this PR?

After @chunweilei review, then we can merge it.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -308,12 +308,20 @@ private static RexNode convertNvl(SqlRexContext cx, SqlCall call) {
...
             rexBuilder.makeCall(SqlStdOperatorTable.IS_NOT_NULL,
                 operand0),
-            rexBuilder.makeCast(type, operand0),
-            rexBuilder.makeCast(type, operand1)));
+            rexBuilder.makeCast(
+                cx.getTypeFactory()
+                    .createTypeWithNullability(type, operand0.getType().isNullable()),
+                operand0),
+            rexBuilder.makeCast(
+                cx.getTypeFactory()
```
chunweilei, line 321: I agree that operand1 should keep its nullability. But operand0 would be not nullable after rewrite.

jaystarshot: @chunweilei 
1) Consider If that operand depends on an input reference(which is nullable)
2) If in future that operand pushed down like in [ProjectJoinTransposeRule](https://github.com/apache/calcite/blob/bebe473fab2e242736614659ed6e5d04eeeb8bf5/core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java), there will be an issue as the operand and its reference have different nullability.
3) The output of the NVL operation is Not Null but there is no need to change the nullability of the operands as in calcite rules we might push the operands down in some rules causing errors ( operand referenced input type is nullable whereas we forced the operand to be not Null)
Do you think hence it makes sense to let the operand nullability what it was as the output is anyway not nullable?

jamesstarr: @chunweilei NVL(null, null) is valid sql.

jamesstarr: The return type should also be nullable if both inputs are nullable.

jaystarshot: `
select nvl(“name”, “name”) FROM “hr”.“emps”;
`
For this query the output type was indeed nullable (both names are nullable) so doesn't seem like an issue.

chunweilei: Got it! Thank you for your explanation.

chunweilei (approved):
+1

NobiGo:
@jaystarshot Please resolve the conflict files and squash the commits.

jaystarshot:
@NobiGo  Done

NobiGo:
@jaystarshot  Sorry I noticed the commit info has some invalid commit info.
For example:
"
Adding Integration tests
Fix style
Using Quidem Tests
Removing SqlOperator Test
Adding Ok to test
"
Can you just delete it ?

jaystarshot:
Thanks @NobiGo  I removed that info

NobiGo (approved):

jaystarshot:
@NobiGo  This was passing before squash and rebase, is that test flaky?

PR #2599, opened 2021-11-04
[CALCITE-4868] Elasticsearch adapter fails if GROUP BY is followed by ORDER BY

ILuffZhe:
The PR supports sorting aggregation result in Elasticsearch Adapter.
Like: **select val2, max(val3) as MAX_VAL3 from view group by val2 order by MAX_VAL3 desc**

Quoted from ES: "Buckets can only be sorted on a sub-aggregator path that is built out of zero or more single-bucket aggregations within the path and a final single-bucket or a metrics aggregation at the path end."

So, SQL like **select val1, val2, max(val3) as MAX_VAL3 from view group by val1, val2 order by MAX_VAL3 desc, val1, val2** is not supported, that's why I add a test case annotated with _Disable_ tag.


NobiGo:
1) Please reformat the commit info like[CALCITE-XXXX] XXXXXXXXX. 
2)The commit info needs to stay the same as The Jira's summary.

ILuffZhe:
Hi, @NobiGo . I've add some description for this fix, please take a look when you are free.

ILuffZhe:
I'll reopen this PR after https://github.com/apache/calcite/pull/2657 since those two are highly related.

PR #2600, opened 2021-11-06
[CALCITE-4877] Make the exception information more explicit for instantiate plugin.

sunjincheng121:
see here: https://github.com/apache/calcite-avatica/pull/161

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,18 +74,18 @@ jobs:
 
   linux-avatica:
     if: github.event.action != 'labeled'
-    name: 'Linux (JDK 11), Avatica-1.19.0'
+    name: 'Linux (JDK 11), Avatica master'
     runs-on: ubuntu-latest
     steps:
     - name: 'Set up JDK 11'
       uses: actions/setup-java@v1
       with:
         java-version: 11
-    - name: 'Clone Avatica-1.19.0 to Maven Local repository'
+    - name: 'Clone Avatica to Maven Local repository'
       run: |
-        git clone --branch rel/avatica-1.19.0 --depth 100 https://github.com/apache/calcite-avatica.git ../calcite-avatica
+        git clone --branch master --depth 100 https://github.com/apache/calcite-avatica.git ../calcite-avatica
     - uses: burrunan/gradle-cache-action@v1
-      name: Build Avatica-1.19.0
+      name: Build Avatica
       env:
         S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.S3_BUILD_CACHE_ACCESS_KEY_ID }}
         S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.S3_BUILD_CACHE_SECRET_KEY }}
@@ -95,13 +95,13 @@ jobs:
         build-root-directory: ../calcite-avatica
         arguments: publishToMavenLocal
         properties: |
-          calcite.avatica.version=1.19.0
+          calcite.avatica.version=1.0.0-dev-master
           skipJavadoc=
     - uses: actions/checkout@v2
       with:
         fetch-depth: 50
     - uses: burrunan/gradle-cache-action@v1
-      name: Test Avatica-1.19.0
+      name: Test
       env:
         S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.S3_BUILD_CACHE_ACCESS_KEY_ID }}
         S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.S3_BUILD_CACHE_SECRET_KEY }}
@@ -111,7 +111,7 @@ jobs:
         execution-only-caches: true
         arguments: --scan --no-parallel --no-daemon build javadoc
         properties: |
-          calcite.avatica.version=1.19.0-SNAPSHOT
+          calcite.avatica.version=1.0.0-dev-master-SNAPSHOT
           enableMavenLocal=
 
   linux-openj9:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -83,7 +83,7 @@ jobs:
         java-version: 11
     - name: 'Clone Avatica to Maven Local repository'
       run: |
-        git clone --branch master --depth 100 https://github.com/apache/calcite-avatica.git ../calcite-avatica
+        git clone --branch rel/avatica-1.19.0 --depth 100 https://github.com/apache/calcite-avatica.git ../calcite-avatica
     - uses: burrunan/gradle-cache-action@v1
       name: Build Avatica
       env:
@@ -95,7 +95,7 @@ jobs:
         build-root-directory: ../calcite-avatica
         arguments: publishToMavenLocal
         properties: |
-          calcite.avatica.version=1.0.0-dev-master
+          calcite.avatica.version=1.19.0
           skipJavadoc=
     - uses: actions/checkout@v2
       with:
@@ -111,7 +111,7 @@ jobs:
         execution-only-caches: true
         arguments: --scan --no-parallel --no-daemon build javadoc
         properties: |
-          calcite.avatica.version=1.0.0-dev-master-SNAPSHOT
+          calcite.avatica.version=1.19.0-SNAPSHOT
           enableMavenLocal=
 
   linux-openj9:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,18 +74,18 @@ jobs:
 
   linux-avatica:
     if: github.event.action != 'labeled'
-    name: 'Linux (JDK 11), Avatica master'
+    name: 'Linux (JDK 11), Avatica-1.19.0'
     runs-on: ubuntu-latest
     steps:
     - name: 'Set up JDK 11'
       uses: actions/setup-java@v1
       with:
         java-version: 11
-    - name: 'Clone Avatica to Maven Local repository'
+    - name: 'Clone Avatica-1.19.0 to Maven Local repository'
       run: |
         git clone --branch rel/avatica-1.19.0 --depth 100 https://github.com/apache/calcite-avatica.git ../calcite-avatica
     - uses: burrunan/gradle-cache-action@v1
-      name: Build Avatica
+      name: Build Avatica-1.19.0
       env:
         S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.S3_BUILD_CACHE_ACCESS_KEY_ID }}
         S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.S3_BUILD_CACHE_SECRET_KEY }}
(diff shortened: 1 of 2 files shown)


sunjincheng121:
The `FAILURE   0.6sec, org.apache.calcite.test.ServerQuidemTest > test(String)[3], [3] sql/schema.iq` will be fixed here: https://github.com/apache/calcite-avatica/pull/161

vlsi:
@sunjincheng121 would you please update the tests?

vlsi:
That is interesting. Tests with "avatica master branch" would fail since now it throws newer exceptions. On the other hand, if we merge this PR, then tests with the released avatica would fail.

It looks like we should postpone this PR till Avatica is released, and in the mean time we should use a fixed git commit (the one before the updated exceptions) for testing "avatica + calcite" integration.

NobiGo:
> That is interesting. Tests with "avatica master branch" would fail since now it throws newer exceptions. On the other hand, if we merge this PR, then tests with the released avatica would fail.
> 
> It looks like we should postpone this PR till Avatica is released, and in the mean time we should use a fixed git commit (the one before the updated exceptions) for testing "avatica + calcite" integration.

Yes. I have noticed this too. It is unreasonable that test Calcite's master branch with Avatica's master branch directly. I support use a fixed git commit (the one before the updated exceptions) for testing "avatica + calcite" integration.

sunjincheng121:
Thanks for the comments @vlsi and @NobiGo , If I understand correctly, we can make the 'calcite-avatica' version used in the test consistent with the property of 'calcite.avatica.version' configured in 'gradle.properties'. What do you think?

sunjincheng121:
Will improve the job name after the test check success.

vlsi:
There's no point in having a separate job if "released" avatica version is used.

The test with "latest avatica" enables to uncover bugs if new commits to calcite accidentally call APIs that are removed in the recent Avatica commts.

.github/workflows/main.yml
```diff
@@ -74,18 +74,18 @@ jobs:
 
   linux-avatica:
     if: github.event.action != 'labeled'
-    name: 'Linux (JDK 11), Avatica master'
+    name: 'Linux (JDK 11), Avatica-1.19.0'
```
vlsi, line 77: Do you suggest each upgrade of Avatica would trigger manual replacement of 1.19.0 in several places?
I am afraid it creates a really sad maintainer experience.

vlsi:
This is the first time we have had an intended failure like that, and what do you think of a temporary disabling `Avatica master+current Calcite` job? As Avatica is released, we could re-enable the check.

Another option is to ask @julianhyde or someone who knows Quidem the way to handle the case in `.iq` test somehow (e.g. use different assertions for different avatica versions). I do not know what is the way to make `.iq` conditional on Avatica version.

vlsi:
I'm inclined we should solve this at `.iq` level or rewrite `.iq` test to regular JUnit, so can support multiple Avatica versions (e.g. via "contains message1 or message2").

Otherwise, it would be the case we support one specific Avatica version only, which questions the split of Calcite and Avatica into different repositories. So we either support at least two Avatica versions, or we should just merge the repositories and stop pretending they are independent libraries.

julianhyde:
Simplest thing is to just support one version of Avatica in Calcite. Which is what we do already. 

Temporarily turn off the Avatica CI job that runs against Calcite latest. That job is a useful “early warning” but it’s fine if it doesn’t always work. 

Julian

vlsi:
> Simplest thing is to just support one version of Avatica in Calcite. Which is what we do already.

Should we merge the repositories then? Do we really gain much by splitting Avatica to a different repository?
Frankly speaking, it is harder to maintain when the repositories are separated: e.g. bump versions everywhere, bump Gradle, we have to vote twice, and so on.

julianhyde:
> Should we merge the repositories then? 

In my opinion, no. And in any case, it's orthogonal to this case, so if feel strongly, start a DISCUSS email thread.

vlsi:
@julianhyde , Here's the discussion: https://lists.apache.org/thread/zyrftnms4p31lsvm93x1cynhnjnccbq7

So far it looks like I broke both calcite-avatica and calcite builds by merging https://github.com/apache/calcite-avatica/pull/161. That is why I try to keep the thing going and try to figure out to heal both builds without reverting PR#161.

As an alternative option, I could revert PR#161 and let somebody else deal with the situation.

julianhyde:
@vlsi I'm trying to fix this by adding a conditional into server.iq.

But as usual, defeated by Gradle:
```
Execution failed for task ':linq4j:compileJava'.
> Could not resolve all files for configuration ':linq4j:compileClasspath'.
   > Could not find org.apache.calcite.avatica:avatica-core:1.20.0-SNAPSHOT.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/org/apache/calcite/avatica/avatica-core/1.20.0-SNAPSHOT/maven-metadata.xml
       - https://repo.maven.apache.org/maven2/org/apache/calcite/avatica/avatica-core/1.20.0-SNAPSHOT/avatica-core-1.20.0-SNAPSHOT.pom
     Required by:
         project :linq4j
   > Could not find org.apache.calcite.avatica:avatica-core:1.20.0-SNAPSHOT.
```
Avatica 1.20.0-SNAPSHOT exists. It is in .m2/repository. Gradle can't find it.

sunjincheng121:
Sorry for the late reply.  @vlsi @julianhyde 

> Temporarily turn off the Avatica CI job that runs against Calcite latest. That job is a useful “early warning” but it’s fine if it doesn’t always work.

+1 for turn off the related tests, or temporarily delete problematic test for solve the  calcite-avatica and calcite builds.

Of course, we can fix the issue by adding a conditional. but I feel a little strange that the tests of the two projects depend on each other's master.  Can we permanently eliminate the testing dependence on master？

julianhyde:
I have disabled Avatica running against Calcite master in https://github.com/apache/calcite-avatica/commit/4cf769f8ee32bb3520604e4f3284e6f233f90276. We can re-enable when the problem is fixed in Calcite master.

sunjincheng121:
> @vlsi I'm trying to fix this by adding a conditional into server.iq.

Are you mean that we should change the `schema.iq` ?

julianhyde:
I mean that someone should change `server.iq`.

sunjincheng121:
> I have disabled Avatica running against Calcite master in [apache/calcite-avatica@4cf769f](https://github.com/apache/calcite-avatica/commit/4cf769f8ee32bb3520604e4f3284e6f233f90276). We can re-enable when the problem is fixed in Calcite master.

Thank you @julianhyde ! and I have create the re-enable issue for mark the followup work. CALCITE-4880

julianhyde:
-1

Changing git workflows seems to be just masking problems.

julianhyde:
I don't see why we have Calcite running against latest Avatica AND ALSO Avatica running against latest Calcite. Too brittle. Let's do one or the other.

sunjincheng121:
> I don't see why we have Calcite running against latest Avatica AND ALSO Avatica running against latest Calcite. Too brittle. Let's do one or the other.

This is what puzzles me，as I mentioned before :)
> but I feel a little strange that the tests of the two projects depend on each other's master. Can we permanently eliminate the testing dependence on master？

sunjincheng121:
> @julianhyde , Here's the discussion: https://lists.apache.org/thread/zyrftnms4p31lsvm93x1cynhnjnccbq7
> 
> So far it looks like I broke both calcite-avatica and calcite builds by merging [apache/calcite-avatica#161](https://github.com/apache/calcite-avatica/pull/161). That is why I try to keep the thing going and try to figure out to heal both builds without reverting PR#161.
> 
> As an alternative option, I could revert PR#161 and let somebody else deal with the situation.

Thanks for helping review and merged the PR161. and broke the builds is my fault. and de-coupling the project's master dependency of the two projects is what we want to see.  Will Pay attention to the [discussion](https://lists.apache.org/thread/zyrftnms4p31lsvm93x1cynhnjnccbq7) you initiated before making any changes.

sunjincheng121:
> -1 to this pull request (modifying .github/workflows/main.yml).
> 
> Changing git workflows seems to be just masking problems.

I think calcite only be built against released calcite-avatica artifacts. I repaired this file because it relies on calcite-avatica's master, which is not the best choice.

vlsi:
> Avatica 1.20.0-SNAPSHOT exists. It is in .m2/repository. Gradle can't find it.

@julianhyde , this is intentional, and by default Gradle does *not* resolve dependencies from Maven Local repository.
This ensures every developer has a consistent resolution, and they do not get accidental artifacts lurking from no one knows which times.

For the rare cases when you really need to have `mavenLocal`, you can add `-PenableMavenLocal` command line argument which would add Maven Local to the list of repositories.

>Changing git workflows seems to be just masking problems

+1. I incline that the CI jobs uncovered true problem, so we should not disable them. Of course, temporary disabling is ok as a workaround, however, they served their purpose and they identified a bug.

julianhyde:
You may have intended this behavior, but it sucks for the rest of us.

I searched Google for how to get Gradle to read from my Maven repository. It said to add `mavenLocal()` to my `repositories` section. I did that. Nothing happened.

I have said it many times, and I'll say it again. Gradle sucks. If you are using a build script that someone else has written, Maven is far superior. If you have a problem, you can Google for the answer. I don't care if Gradle makes my build 5 seconds faster. Because a few times a year I waste a couple of hours finding the answer to a question.

vlsi:
@julianhyde , please stop blaming me and/or the tool. Please be friendly

vlsi:
The exact recipe is in GitHub actions: https://github.com/apache/calcite/blob/f1db79fb876ac9ba3c405283e99bb0438e4e97be/.github/workflows/main.yml#L115

> It said to add mavenLocal() to my repositories section. I did that. Nothing happened.

There are multiple `repositories` sections. If you add `mavenLocal` to *all* of them, it would eventually help.
If you know better, then you know that the one is enough, which is `allprojects { repositories { ... } }`

---

> If you are using a build script that someone else has written, Maven is far superior

Calcite was built with Maven for ages, and a perfectly-tuned Manven script was not really usable.

julianhyde:
@vlsi, Yeah, sorry, my comments were rather too personal. I know you've put a lot of work into Calcite's Gradle build scripts.

sunjincheng121:
Turn off the related tests、temporarily delete problematic test、or using the release version of `calcite-avatica` modify the `.github/workflows/main.yml`  can solve issue of CI builds  as a workaround.  So, we can keep the current states that disable avatica's travis job that uses calcite master or merge this PR and re-open the avatica's travis job in  CALCITE-4880. 

Anyway, Feel free to close this PR as we can tracing this issue in CALCITE-4880 according to the result of the [discussion](https://lists.apache.org/thread/zyrftnms4p31lsvm93x1cynhnjnccbq7).

Thanks all for pay attention to this PR. @vlsi @NobiGo @julianhyde 

Best,
Jincheng Sun

sunjincheng121:
Will close this PR as there is no further feedback coming.

PR #2603, opened 2021-11-10
[CALCITE-4882] Introduce new Lambda-based Metadata framework

jacques-n:
- Introduce a Lambda-based RelMetadataQuery creator for AOT environments
- Add new microbenchmark for metadata retrieval on a five-way-join
- Fix incorrect results for disabled test JdbcTest.testJoinFiveWay()
- Update jmh-gradle-plugin version.


jacques-n:
Working to update RelMdTest to be parameterized...

jacques-n:
Added reuse of RelMetdataTest. Will work on null checker next...

core/src/main/java/org/apache/calcite/rel/metadata/lambda/ReflectionToLambdaProvider.java
```diff
@@ -0,0 +1,359 @@
...
+ * This does reflection to lambda conversion based on a set of singleton objects
+ * with appropriate signatures, similar to how ReflectiveRelMetadataProvider works.
+ * Any class that can be discovered using ReflectiveRelMetadataProvider should also be consumable
+ * using this class. Ultimately, the goal may be to move to direct lambda registration as opposed to
+ * the old system of partial reflection discovery.
+ */
+@ThreadSafe
+public class ReflectionToLambdaProvider implements LambdaProvider {
+
+  public static final ImmutableList<Source> DEFAULT_SOURCES = ImmutableList.<Source>builder()
```
jamesstarr, line 83: If all handlers were refactored to only have 1 methods, couldn't we drive this off RelMetadataProvider, if a visiter pattern was exposed?

jacques-n: This is effectively a reimagining of `MetadataHandler` as a functional interface with a more useful hierarchy/generic definition. Could we convert MetdataHandler to provide this functionality: yes, we probably could. 

I haven't spent time trying to untangle the way we would get there. I would expect that it would take many releases to try minimize disruption while slowly converting all the interfaces/impls/etc.

jacques-n: I'll add, in general the whole pattern of arbitrary classes carrying arbitrary methods that name match and arguments kind-of match seems like something that wouldn't translate well to a lambda pattern (which is why the concrete RelNode argument is part of the generic interface definition in this lambda pattern).

jamesstarr: The Lambda API is leaking the dispatch implementation into the API for binding in custom implementations.  These APIs should be separate.  It is actually fairly easy to adept this approach to use handlers and not be distributive to down stream projects.

jacques-n: Can you expound on your comments. They are a bit brief. 

>Lambda API is leaking the dispatch implementation into the API for binding in custom implementations
Please be more specific about what exactly you think is leaking where. I think we may have different perspectives of what functionality is trying to be provided and how. The design here is that the declaration of a specific MetadataLambda implementation gives all the information about what RelNode types it applies to and what metadata it provides. I haven't yet posted the simple provider that simply returns hardcoded lambdas, but that was part of the goal for this design.

> It is actually fairly easy to adept this approach to use handlers
I assume you mean adapt. I can see how one could make MetadataLambdas pretend to be handlers with a bunch of boilerplate but I don't see why that is a good idea. What exactly would this buy? Using janino on top of lambdas?

> and not be distributive to down stream projects.
I assume you mean disruptive? I don't see how any of these changes are disruptive. This is just an alternative way to create RelMetadataQuery instances and specify metadata (via lambdas).

FYI, I will be updating this patch that to also include my vision for how to handle custom metadata types without any reflection and without excessive subclassing.

jamesstarr: If this is not the purposed end api, then we should not waste too much time discussing this implementation.  But just as the API for metadata retrieval should not expose internals, the API for registering metadata implementations.  This has not been the case for the previous api.

If your purposed API for this is to register every method the RelNode type instead of discover theming reflectively, then my questions is why do we need some convoluted framework for handling the dispatch instead of just harding coding in the dispatch.  Their is pretty much a one to one mapping at that point of the amount of configuration needed compare to explicitly maintaining the implementation.

vlsi:
Just wondering, does it have something in common with https://github.com/forax/exotic#structuralcall---javadoc ?

jacques-n:
> Just wondering, does it have something in common with https://github.com/forax/exotic#structuralcall---javadoc ?

Similar, but not the same. Here we're using LambdaMetafactory and actually convert the function call to a generic version. Core of that work is [here](https://github.com/apache/calcite/blob/b69ffa2c0eeb65b1bb5c46d745730cdace4d5e4c/core/src/main/java/org/apache/calcite/rel/metadata/lambda/ReflectionToLambdaProvider.java#L260).

linghengqian:
I'm curious, what is the reason why this PR is blocked?

jacques-n:
> I'm curious, what is the reason why this PR is blocked?

I was waiting for CALCITE-4942 to be resolved since it could potentially substantially reduce the amount of boilerplate and reimplementation in this patch. Note that I did get [ProxyingMetadataHandlerProvider|https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/rel/metadata/ProxyingMetadataHandlerProvider.java] which allows one to use metadata without runtime compilation. It isn't fast but at least it is functional in things like GraalVM.

jacques-n:
> I thought there was not agreement on how to go forward with [CALCITE-4942](https://issues.apache.org/jira/browse/CALCITE-4942)

I posted a comment on the JIRA for CALCITE-4942 in late December but never saw a response (afaik).

jacques-n:
Abandoning this work as there was too much contention around it.

PR #2605, opened 2021-11-10
[CALCITE-4783] RelFieldTrimmer incorrectly drops filter condition

shyzeshzye:
fix RelFieldTrimmer incorrectly drops calc condition when there is no field used by the upper 
https://issues.apache.org/jira/browse/CALCITE-4783

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
@@ -371,6 +371,15 @@ public class RelFieldTrimmer implements ReflectiveVisitor {
     final List<RexNode> projs = Util.transform(rexProgram.getProjectList(),
         rexProgram::expandLocalRef);
 
+    RexNode conditionExpr = null;
+    if (rexProgram.getCondition() != null) {
+      final List<RexNode> filter = Util.transform(
+          ImmutableList.of(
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -371,6 +371,15 @@ public TrimResult trimFields(
     final List<RexNode> projs = Util.transform(rexProgram.getProjectList(),
         rexProgram::expandLocalRef);
 
+    RexNode conditionExpr = null;
+    if (rexProgram.getCondition() != null) {
+      final List<RexNode> filter = Util.transform(
+          ImmutableList.of(
+              rexProgram.getCondition()), rexProgram::expandLocalRef);
+      assert filter.size() == 1;
```
xy2953396112, line 379: Why `filter.size() == 1`?

xy2953396112:
Please not use `fix` for PR title.

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
shyzeshzye: RexProgram has only one condition, so  it means rexProgram.getCondition() != null.

chunweilei (approved):
LGTM

PR #2607, opened 2021-11-12
[CALCITE-4846] IN-list that includes NULL converted to Values throws exception

NobiGo:


core/src/test/resources/sql/sub-query.iq
```diff
@@ -3306,4 +3306,28 @@ EnumerableCalc(expr#0..7=[{inputs}], expr#8=[7782], expr#9=[CAST($t0):INTEGER NO
...
+  EnumerableMergeJoin(condition=[=($10, $11)], joinType=[left])
+    EnumerableSort(sort0=[$10], dir0=[ASC])
+      EnumerableCalc(expr#0..9=[{inputs}], proj#0..9=[{exprs}], EMPNO0=[$t0])
+        EnumerableNestedLoopJoin(condition=[true], joinType=[inner])
+          EnumerableTableScan(table=[[scott, EMP]])
+          EnumerableAggregate(group=[{}], agg#0=[COUNT()], agg#1=[COUNT($0)])
+            EnumerableValues(tuples=[[{ null }, { 7782 }]])
+    EnumerableSort(sort0=[$0], dir0=[ASC])
+      EnumerableCalc(expr#0=[{inputs}], expr#1=[true], proj#0..1=[{exprs}])
+        EnumerableValues(tuples=[[{ null }, { 7782 }]])
```
vlsi, line 3330: The plan looks strange. Why are two joins are present here?

If that join duplication is not related to the fix, then I would suggest adding a comment and filing a JIRA ticket.
Otherwise, it looks like a bug.

core/src/test/resources/sql/sub-query.iq
```diff
@@ -3306,4 +3306,28 @@ EnumerableCalc(expr#0..7=[{inputs}], expr#8=[7782], expr#9=[CAST($t0):INTEGER NO
   EnumerableTableScan(table=[[scott, EMP]])
 !plan
 
+# [CALCITE-4846] IN-list that includes NULL converted to Values throws exception
+
+select * from "scott".emp where empno not in (null, 7782);
```
vlsi, line 3311: just wondering if `(empno, depnto) in ((1,2), (3, null))` work or not. Do you think it is worth adding a test like that?

vlsi (changes requested):

core/src/test/resources/sql/sub-query.iq
hsyuan: If I remembered it correctly, Calcite unnest NOT IN subquery to 2 joins, that is how it deals with NOT IN currently. :(

core/src/test/resources/sql/sub-query.iq
NobiGo: Yes. I have added it.

core/src/test/resources/sql/sub-query.iq
NobiGo: The Join duplication is not related to the fix. When I try to execute:
`select * from "scott".emp where (empno, deptno) not in ((7369, 20), (7499, 30));`
This SQL also has two joins. And return the right answer. So I think this is the right logic. (Unless we find a wrong unit
 test).

vlsi: When values are non-nullable, then the double join is not really needed. A single anti-join should be enough.
I have filed https://issues.apache.org/jira/browse/CALCITE-4889

vlsi (approved):

PR #2612, opened 2021-11-18
[CALCITE-4888] Unify type inferring logical for Sarg RexLiteral

beyond1920:

If `shouldConvertRaggedUnionTypesToVarying` is true, updates `RexBuilder#makeIn` to use less restrictive type based on all ranges types instead of first range type.

## Brief change log

  - Update `RexBuilder#makeIn`

## Verifying this change

  - Add tests in `RexBuilderTest`, `RelBuilderTest`

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexBuilder.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexBuilder.java
@@ -76,6 +76,8 @@ import java.util.function.IntPredicate;
 
 import static org.apache.calcite.linq4j.Nullness.castNonNull;
 
+import static java.util.Objects.requireNonNull;
+
 /**
  * Factory for row expressions.
  *
@@ -1332,10 +1334,14 @@ public class RexBuilder {
     if (areAssignable(arg, ranges)) {
       final Sarg sarg = toSarg(Comparable.class, ranges, RexUnknownAs.UNKNOWN);
       if (sarg != null) {
-        final RexNode range0 = ranges.get(0);
+        final List<RelDataType> types = Util.transform(ranges, RexNode::getType);
+        final List<RelDataType> distinctTypes = Util.distinctList(types);
+        final RelDataType sargType = requireNonNull(
+            typeFactory.leastRestrictive(distinctTypes),
+            () -> "Can't find leastRestrictive type among " + distinctTypes);
         return makeCall(SqlStdOperatorTable.SEARCH,
             arg,
-            makeSearchArgumentLiteral(sarg, range0.getType()));
+            makeSearchArgumentLiteral(sarg, sargType));
       }
     }
     return RexUtil.composeDisjunction(this, ranges.stream()
--- a/core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
@@ -4042,18 +4042,37 @@ public class RelBuilderTest {
   }
 
   /**
-   * Tests {@link RelBuilder#in} would throw exception if types of the ranges are not compatible.
+   * Tests {@link RelBuilder#in} would throw an exception when simplify filter predicates if types
+   * of the ranges are not compatible.
    */
-  @Test void testFilterInWithNotCompatibleArguments() {
+  @Test void testFilterInWithNotCompatibleArgumentsSimplification() {
     try {
       final RelBuilder b = RelBuilder.create(config().build());
       final RelNode root = b.scan("EMP")
           .filter(b.in(b.field(2), b.literal(1), b.literal(true)))
           .build();
       fail("expected error, got " + root);
-    } catch (ClassCastException e) {
-      // good
+    } catch (Exception e) {
+      assertThat(
+          e.getMessage(), is(
+          "Sarg arguments are not compatible with each other!\n"
+              + "Existed ranges are [[1..1]], new range is [true..true]."));
     }
+  }
+
+  /**
+   * Tests {@link RelBuilder#in} would generate a disjunction if types of the ranges are not
+   * compatible.
+   */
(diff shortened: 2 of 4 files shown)


core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
```diff
@@ -4031,6 +4033,22 @@ private static RelNode groupIdRel(RelBuilder builder, boolean extra) {
...
+        "LogicalFilter(condition=[SEARCH($2, Sarg['A':CHAR(5), 'CLERK']:CHAR(5))])\n"
+            + "  LogicalTableScan(table=[[scott, EMP]])\n";
+    final String expectedWithLeastRestrictive =
+        "LogicalFilter(condition=[SEARCH($2, Sarg['A':VARCHAR(5), 'CLERK':VARCHAR(5)]:VARCHAR(5))])\n"
+        + "  LogicalTableScan(table=[[scott, EMP]])\n";
+    assertThat(f.apply(createBuilder()), hasTree(expected));
+    assertThat(f.apply(RelBuilder.create(config().typeSystem(VaryingTypeSystem.INSTANCE).build())),
+        hasTree(expectedWithLeastRestrictive));
+  }
+
```
hsyuan, line 4051: Can you add a case for `(a,b) in (('1','1'), ('22', '333'))`?

beyond1920: @hsyuan It seems that the `RexNode` which represents a filter condition like `(a,b) in (('1','1'), ('22', '333'))` could not created by `RexBuilder.makeIn` because the condition would be translated into a composed condition like 
`OR(AND(=($0, _UTF-16LE'1'), =($1, _UTF-16LE'11')), AND(=($0, _UTF-16LE'2'), =($1, _UTF-16LE'333')))`. right?

NobiGo: @beyond1920 Before the code review, I think we need to reach an agreement about the method that how to resolve this issue.

beyond1920: @NobiGo Thanks for comment. I left some message in JIRA, welcome to further discussion.

Tartarus0zm (approved):

Tartarus0zm:
@beyond1920  LGTM, +1

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1332,10 +1334,14 @@ public RexNode makeIn(RexNode arg, List<? extends RexNode> ranges) {
...
       final Sarg sarg = toSarg(Comparable.class, ranges, RexUnknownAs.UNKNOWN);
       if (sarg != null) {
-        final RexNode range0 = ranges.get(0);
+        final List<RelDataType> types = Util.transform(ranges, RexNode::getType);
+        final List<RelDataType> distinctTypes = Util.distinctList(types);
+        final RelDataType sargType = requireNonNull(
+            typeFactory.leastRestrictive(distinctTypes),
+            () -> "Can't find leastRestrictive type among " + distinctTypes);
         return makeCall(SqlStdOperatorTable.SEARCH,
             arg,
```
danny0405, line 1343: `leastRestrictive` => `least restrictive`

core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
```diff
@@ -4031,6 +4033,41 @@ private static RelNode groupIdRel(RelBuilder builder, boolean extra) {
...
+   */
+  @Test void testFilterInWithNotCompatibleArguments() {
+    try {
+      final RelBuilder b = RelBuilder.create(config().build());
+      final RelNode root = b.scan("EMP")
+          .filter(b.in(b.field(2), b.literal(1), b.literal(true)))
+          .build();
+      fail("expected error, got " + root);
+    } catch (ClassCastException e) {
+      // good
```
danny0405, line 4047: Can we assert the exception message here ?

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
danny0405: Is this the right fix ? It seems that the code does some logic for implicit type coercion here, and there is a breaking change:
Before the change, the in-compatible values in `IN` clause does not throw while with this patch it does, we need to confirm the behavior.

If we want to add implicit type coercion here, we better unify the logic with what `TypeCoercion` does.

beyond1920: `leastRestrictive => least restrictive`
The exception message is keep consistent with `RexSargBuilder#RexSargBuilder` in `RexSimplify`.
I think it's a better user experience for user.

beyond1920: > It seems that the code does some logic for implicit type coercion here
The pr does not aim to fix implicit type coercion, it only want to solve the behavior of `RexBulder#makeIn` if the arguments are compatibule, just like discussion in the JIRA:
image.
I could create another JIRA to solve implicit type coercion.
Let's solve the problem one by one.
> Before the change, the in-compatible values in IN clause does not throw while with this patch it does, we need to confirm the behavior.
The old behavior is not correct which we need to fix.

core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
beyond1920: Good question, I once assert the exception message here. But I found assert only success for mac, failed in Window/Linux because ClassCastException message is minor different when run in the different environment. 
So I only  check the exception  class  here.

libenchao:
I'm closing this PR because https://github.com/apache/calcite/pull/2853 has been merged and fix this issue.

PR #2613, opened 2021-11-18
[CALCITE-4894] MV rewriting fails for conjunctive top expressions in SELECT clause

asolimando:
### Overview
MV rewrite fails when at least one expression in the project of either the view or the query references, directly or indirectly, to more than one field.

### Changes
- Added unit tests covering such "complex" expressions in both views and queries (highlighting the limitation)
- Improved the implementation to handle such cases as well
- Improved the error message for (possibly remaining) unhandled cases where a single expression can have multiple lineages, turned assertions into exceptions.

Regarding the tests, the generated plans look fine, I was wondering if we could add tests with some actual data to double check that the rewriting is working as expected. So far I could not find the right place to add that, I will keep digging, not sure it's strictly needed, it does not seem to be done for any other case, no matter how complex.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdExpressionLineage.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdExpressionLineage.java
@@ -493,7 +493,7 @@ public class RelMdExpressionLineage
       Map<RexInputRef, RexNode> singleMapping, Set<RexNode> result) {
     if (mapping.isEmpty()) {
       final RexReplacer replacer = new RexReplacer(singleMapping);
-      final List<RexNode> updatedPreds = new ArrayList<>();
+      final List<RexNode> updatedPreds = new ArrayList<>(1);
       updatedPreds.add(rexBuilder.copy(expr));
       replacer.mutate(updatedPreds);
       result.addAll(updatedPreds);
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdExpressionLineage.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdExpressionLineage.java
@@ -493,9 +493,8 @@ public class RelMdExpressionLineage
       Map<RexInputRef, RexNode> singleMapping, Set<RexNode> result) {
     if (mapping.isEmpty()) {
       final RexReplacer replacer = new RexReplacer(singleMapping);
-      final List<RexNode> updatedPreds = new ArrayList<>(
-          RelOptUtil.conjunctions(
-              rexBuilder.copy(expr)));
+      final List<RexNode> updatedPreds = new ArrayList<>();
+      updatedPreds.add(rexBuilder.copy(expr));
       replacer.mutate(updatedPreds);
       result.addAll(updatedPreds);
     } else {
--- a/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewJoinRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewJoinRule.java
@@ -279,8 +279,10 @@ public abstract class MaterializedViewJoinRule<C extends MaterializedViewRule.Co
           exprsLineage.add(RexUtil.composeDisjunction(rexBuilder, rewrittenExprs));
         }
       } else {
-        assert lineages.size() == 1 : "We only support project - filter - join, "
-            + "thus expression lineage should map to a single expression, got: " + lineages;
+        if (lineages.size() != 1) {
+          throw new IllegalStateException("We only support project - filter - join, "
+              + "thus expression lineage should map to a single expression, got: " + lineages);
+        }
         // Rewrite expr. Take first element from the corresponding equivalence class
         // (no need to swap the table references following the table mapping)
         exprsLineage.add(
--- a/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewJoinRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewJoinRule.java
@@ -29,6 +29,7 @@ import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexSimplify;
 import org.apache.calcite.rex.RexTableInputRef.RelTableRef;
 import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.tools.RelBuilder;
 import org.apache.calcite.util.Pair;
 
@@ -43,6 +44,7 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /** Materialized view rewriting for join.
  *
@@ -260,17 +262,28 @@ public abstract class MaterializedViewJoinRule<C extends MaterializedViewRule.Co
         : topProject.getProjects();
     List<RexNode> exprsLineage = new ArrayList<>(exprs.size());
     for (RexNode expr : exprs) {
-      Set<RexNode> s = mq.getExpressionLineage(node, expr);
-      if (s == null) {
+      Set<RexNode> lineages = mq.getExpressionLineage(node, expr);
+      if (lineages == null) {
         // Bail out
         return null;
       }
-      assert s.size() == 1;
-      // Rewrite expr. Take first element from the corresponding equivalence class
-      // (no need to swap the table references following the table mapping)
-      exprsLineage.add(
-          RexUtil.swapColumnReferences(rexBuilder,
-              s.iterator().next(), queryEC.getEquivalenceClassesMap()));
+      if (SqlKind.AND == expr.getKind() || SqlKind.OR == expr.getKind()) {
+        List<RexNode> rewrittenExprs = lineages.stream()
+            .map(
+                l -> RexUtil.swapColumnReferences(
+                rexBuilder, l, queryEC.getEquivalenceClassesMap()))
+            .collect(Collectors.toList());
+        if (SqlKind.AND == expr.getKind()) {
+          exprsLineage.add(RexUtil.composeConjunction(rexBuilder, rewrittenExprs));
+        } else {
+          exprsLineage.add(RexUtil.composeDisjunction(rexBuilder, rewrittenExprs));
+        }
+      } else {
+        assert lineages.size() == 1;
+        exprsLineage.add(
+            RexUtil.swapColumnReferences(rexBuilder,
+                lineages.iterator().next(), queryEC.getEquivalenceClassesMap()));
+      }
     }
     List<RexNode> viewExprs = topViewProject == null
         ? extractReferences(rexBuilder, viewNode)
--- a/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewJoinRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewJoinRule.java
@@ -267,7 +267,8 @@ public abstract class MaterializedViewJoinRule<C extends MaterializedViewRule.Co
       }
       if (lineages.size() != 1) {
         throw new IllegalStateException("We only support project - filter - join, "
-            + "thus expression lineage should map to a single expression, got: " + lineages);
+            + "thus expression lineage should map to a single expression, got: '"
+            + lineages + "' for expr '" + expr + "' in node '" + node + "'");
       }
       // Rewrite expr. Take first element from the corresponding equivalence class
       // (no need to swap the table references following the table mapping)
--- a/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewJoinRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewJoinRule.java
@@ -279,7 +279,10 @@ public abstract class MaterializedViewJoinRule<C extends MaterializedViewRule.Co
           exprsLineage.add(RexUtil.composeDisjunction(rexBuilder, rewrittenExprs));
         }
       } else {
-        assert lineages.size() == 1;
+        assert lineages.size() == 1 : "We only support project - filter - join, "
+            + "thus expression lineage should map to a single expression, got: " + lineages;
+        // Rewrite expr. Take first element from the corresponding equivalence class
+        // (no need to swap the table references following the table mapping)
         exprsLineage.add(
             RexUtil.swapColumnReferences(rexBuilder,
                 lineages.iterator().next(), queryEC.getEquivalenceClassesMap()));
--- a/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewJoinRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewJoinRule.java
@@ -29,7 +29,6 @@ import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexSimplify;
 import org.apache.calcite.rex.RexTableInputRef.RelTableRef;
 import org.apache.calcite.rex.RexUtil;
-import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.tools.RelBuilder;
 import org.apache.calcite.util.Pair;
 
@@ -44,7 +43,6 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 /** Materialized view rewriting for join.
  *
@@ -267,28 +265,15 @@ public abstract class MaterializedViewJoinRule<C extends MaterializedViewRule.Co
         // Bail out
         return null;
       }
-      if (SqlKind.AND == expr.getKind() || SqlKind.OR == expr.getKind()) {
-        List<RexNode> rewrittenExprs = lineages.stream()
-            .map(
-                l -> RexUtil.swapColumnReferences(
-                rexBuilder, l, queryEC.getEquivalenceClassesMap()))
-            .collect(Collectors.toList());
-        if (SqlKind.AND == expr.getKind()) {
-          exprsLineage.add(RexUtil.composeConjunction(rexBuilder, rewrittenExprs));
-        } else {
-          exprsLineage.add(RexUtil.composeDisjunction(rexBuilder, rewrittenExprs));
-        }
-      } else {
-        if (lineages.size() != 1) {
-          throw new IllegalStateException("We only support project - filter - join, "
-              + "thus expression lineage should map to a single expression, got: " + lineages);
-        }
-        // Rewrite expr. Take first element from the corresponding equivalence class
-        // (no need to swap the table references following the table mapping)
-        exprsLineage.add(
-            RexUtil.swapColumnReferences(rexBuilder,
-                lineages.iterator().next(), queryEC.getEquivalenceClassesMap()));
+      if (lineages.size() != 1) {
+        throw new IllegalStateException("We only support project - filter - join, "
+            + "thus expression lineage should map to a single expression, got: " + lineages);
       }
+      // Rewrite expr. Take first element from the corresponding equivalence class
+      // (no need to swap the table references following the table mapping)
+      exprsLineage.add(
+          RexUtil.swapColumnReferences(rexBuilder,
+              lineages.iterator().next(), queryEC.getEquivalenceClassesMap()));
     }
     List<RexNode> viewExprs = topViewProject == null
         ? extractReferences(rexBuilder, viewNode)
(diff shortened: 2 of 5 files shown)


core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1036,21 +1037,36 @@ protected NodeLineage generateSwapTableColumnReferencesLineage(
...
-      assert s.size() == 1;
-      // Rewrite expr. First we swap the table references following the table
-      // mapping, then we take first element from the corresponding equivalence class
-      final RexNode e = RexUtil.swapTableColumnReferences(rexBuilder,
-          s.iterator().next(), tableMapping, ec.getEquivalenceClassesMap());
-      exprsLineage.put(e, i);
-      if (RexUtil.isLosslessCast(e)) {
-        exprsLineageLosslessCasts.put(((RexCall) e).getOperands().get(0), i);
+      final RexNode expr = nodeExprs.get(i);
+      if (SqlKind.AND == expr.getKind() || SqlKind.OR == expr.getKind()) {
```
zabetak, line 1046: Why do we need to restrict ourselves to AND/OR expressions? I suppose the same problem can re-appear easily if we have other kind of expressions e.g., `s.time_id + s.time_id`. Can't we apply a generic solution here?

Moreover, I am not sure that `mq.expressionsLineage` works as expected. I checked the javadoc of `RelMdExpressionLineage#createAllPossibleCombinations` and it is supposed to return an equivalent expression with the one given as input. If the input is `A > 1 AND A <3 ` why in the output  we have `A > 1`, `A<3` decomposed. Is this normal?

asolimando: You are right, we can go for a more general solution and support all complex expressions like the one you mentioned. I will also add one or more tests for that.

Regarding `mq.expressionsLineage` I tend to agree with you, it speaks of returning the `set of resulting expressions equivalent to the input expression`, but then it splits it into multiple expressions.

If `mq.expressionsLineage` is broken then maybe we don't have an issue here at all in the first place. I can't see any unit tests for this part of the code, it's hard to tell the original intention behind the method.

asolimando: Finally my diagnosis is that `mq.expressionsLineage` was indeed broken, it was decomposing top `AND` expressions for no apparent good reason.

Luckily there were unit tests, just not where I expected them to be, I have managed to reproduce the issue in one of them, and double checked that the fix did handle it, and at the same time that nothing else broke.

Please have another look @zabetak if you have some spare cycles, thanks a lot for the tip, I was heading down a slippery road :)

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -2482,6 +2483,115 @@ private void checkPredicates(RelOptCluster cluster, RelOptTable empTable,
...
+    assertThat(literal2.getValueAs(String.class), is("abc"));
+
+    // check 'deptno > 1'
+    final RexCall gt = (RexCall) and.getOperands().get(1);
+    assertThat(gt.getKind(), is(SqlKind.GREATER_THAN));
+    final RexTableInputRef inputRef3 = (RexTableInputRef) gt.getOperands().get(0);
+    assertThat(inputRef3.getQualifiedName(), is(EMP_QNAME));
+    assertThat(inputRef3.getIndex(), is(7));
+    final RexLiteral literal3 = (RexLiteral) gt.getOperands().get(1);
+    assertThat(literal3.getValueAs(Integer.class), is(1));
```
zabetak, line 2533: I know that you are following the style of previous tests in this class and that is good but I find a bit verbose for what we really need. Instead of having 17 `assertThat` can't we just have only one based on the string representation of the expression?

Something like the following:
```
String actualExp = r.iterator().next().toString();
assertEquals("AND([CATALOG, SALES, EMP].#0.$0...", actualExp);
```
What do you think?

core/src/main/java/org/apache/calcite/rel/metadata/RelMdExpressionLineage.java
```diff
@@ -493,9 +493,8 @@ private static void createExpressions(RexBuilder rexBuilder,
       Map<RexInputRef, RexNode> singleMapping, Set<RexNode> result) {
     if (mapping.isEmpty()) {
       final RexReplacer replacer = new RexReplacer(singleMapping);
-      final List<RexNode> updatedPreds = new ArrayList<>(
-          RelOptUtil.conjunctions(
-              rexBuilder.copy(expr)));
+      final List<RexNode> updatedPreds = new ArrayList<>();
+      updatedPreds.add(rexBuilder.copy(expr));
```
zabetak, line 497: Micro optim, probably insignificant but since the list is always gonna have one element so let's keep it like that:

```suggestion
      final List<RexNode> updatedPreds = new ArrayList<>(1);
      updatedPreds.add(rexBuilder.copy(expr));
```

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
asolimando: As you say, I have just adapted to the style of the tests, but if breaking uniformity is an option I am more than happy to switch to string-based tests. The other format is hard to write, read and maintain, for no real added value IMO.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -2482,6 +2482,41 @@ private void checkPredicates(RelOptCluster cluster, RelOptTable empTable,
...
+  @Test void testExpressionLineageConjuntiveExpression() {
+    // empno is column 0 in catalog.sales.emp
+    // ename is column 1 in catalog.sales.emp
+    // deptno is column 7 in catalog.sales.emp
+    final RelNode rel = convertSql("select (empno = 1 or ename = 'abc') and deptno > 1 from emp");
+    final RelMetadataQuery mq = rel.getCluster().getMetadataQuery();
+
+    final RexNode ref = RexInputRef.of(0, rel.getRowType().getFieldList());
+    final Set<RexNode> r = mq.getExpressionLineage(rel, ref);
+    assertThat(r.size(), is(1));
```
vlsi, line 2494: It is sad when an assertion like this fails. It says like "expected 1 got 2", and you have absolutely no clue why 1 was expected, what caused 2 and so on.

`assert(size)` before asserting collection contents is almost always a bad idea. There are corner cases like `assertAll`, soft asserts, etc, however, you are not using them here.

In this case, I would suggest removing `assert(size` and use `set.toString()` with the relevant assertion message.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -2482,6 +2482,41 @@ private void checkPredicates(RelOptCluster cluster, RelOptTable empTable,
     assertThat(inputRef1.getIdentifier(), is(inputRef2.getIdentifier()));
   }
 
+  @Test void testExpressionLineageConjuntiveExpression() {
+    // empno is column 0 in catalog.sales.emp
+    // ename is column 1 in catalog.sales.emp
+    // deptno is column 7 in catalog.sales.emp
+    final RelNode rel = convertSql("select (empno = 1 or ename = 'abc') and deptno > 1 from emp");
```
vlsi, line 2489: It would be great if you could move those comments and the SQL in question to the last `assertEquals` in the test.

Then the assertion call would be paired with the intention, and the failure message would include the intended behavior as well.

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewJoinRule.java
```diff
@@ -260,17 +260,20 @@
...
-      if (s == null) {
+      Set<RexNode> lineages = mq.getExpressionLineage(node, expr);
+      if (lineages == null) {
         // Bail out
         return null;
       }
-      assert s.size() == 1;
+      if (lineages.size() != 1) {
+        throw new IllegalStateException("We only support project - filter - join, "
+            + "thus expression lineage should map to a single expression, got: " + lineages);
```
vlsi, line 270: Does it make sense to include the input `expr` and or `node` into the message as well?

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -2482,6 +2482,41 @@ private void checkPredicates(RelOptCluster cluster, RelOptTable empTable,
...
+    final RelMetadataQuery mq = rel.getCluster().getMetadataQuery();
+
+    final RexNode ref = RexInputRef.of(0, rel.getRowType().getFieldList());
+    final Set<RexNode> r = mq.getExpressionLineage(rel, ref);
+    assertThat(r.size(), is(1));
+
+    // 'dept.deptno + empno between 1 and 2' is translated into
+    // 'dept.deptno + empno >= 1 and dept.deptno + empno <= 2'
+    final String actualExp = r.iterator().next().toString();
+    assertEquals("AND(>=(+([CATALOG, SALES, DEPT].#0.$0, [CATALOG, SALES, EMP].#0.$0), 1),"
```
vlsi, line 2516: The comments are cool, however, they won't be visible in assertion failure, and it would be better to put them into the assertion message. Then the person that faces the failure would understand the reasons better.

asolimando: I have added an assertion message, failures now look like this:
```
For query 'select dept.deptno + empno between 1 and 2 from emp join dept on emp.deptno = dept.deptno':
'empno' is column 0 in 'catalog.sales.emp', 'deptno' is column 0 in 'catalog.sales.dept', and 'dept.deptno + empno between 1 and 2' is translated into 'dept.deptno + empno >= 1 and dept.deptno + empno <= 2' ==> expected: <AND(>=(+([CATALOG, SALES, DEPT].#0.$0, [CATALOG, SALES, EMP].#0.$0), 1), <=(+([CATALOG, SALES, DEPT].#0.$0, [CATALOG, SALES, EMP].#0.$0), 2)> but was: <AND(>=(+([CATALOG, SALES, DEPT].#0.$0, [CATALOG, SALES, EMP].#0.$0), 1), <=(+([CATALOG, SALES, DEPT].#0.$0, [CATALOG, SALES, EMP].#0.$0), 2))>
```

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
asolimando: I have added an assertion message, failure now look like this:
```
For query 'select (empno = 1 or ename = 'abc') and deptno > 1 from emp':
'empno' is column 0 in 'catalog.sales.emp', 'ename' is column 1 in 'catalog.sales.emp', and 'deptno' is column 7 in 'catalog.sales.emp' ==> expected: <AND(OR(=([CATALOG, SALES, EMP].#0.$0, 1), =([CATALOG, SALES, EMP].#0.$1, 'abc')), >([CATALOG, SALES, EMP].#0.$7, 1)> but was: <AND(OR(=([CATALOG, SALES, EMP].#0.$0, 1), =([CATALOG, SALES, EMP].#0.$1, 'abc')), >([CATALOG, SALES, EMP].#0.$7, 1))>
Expected :AND(OR(=([CATALOG, SALES, EMP].#0.$0, 1), =([CATALOG, SALES, EMP].#0.$1, 'abc')), >([CATALOG, SALES, EMP].#0.$7, 1)
Actual   :AND(OR(=([CATALOG, SALES, EMP].#0.$0, 1), =([CATALOG, SALES, EMP].#0.$1, 'abc')), >([CATALOG, SALES, EMP].#0.$7, 1))
```

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
asolimando: I have replaced the assertion with the use of `Iterables.getOnlyElement`, that is: `String actualExp = Iterables.getOnlyElement(r).toString();`

`Iterables.getOnlyElement` states explicitly that we were expecting a single element and we got a different number, I think it conveys the intention in case of failures and it keeps the test compact.

vlsi: `Iterables.getOnlyElement` would be slightly better, however, I believe it would not convey the reason for "only one element expected" (==input SQL would be missing, etc, etc)

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -2483,38 +2484,43 @@ private void checkPredicates(RelOptCluster cluster, RelOptTable empTable,
...
-    // deptno is column 7 in catalog.sales.emp
-    final RelNode rel = convertSql("select (empno = 1 or ename = 'abc') and deptno > 1 from emp");
+    final String sql = "select (empno = 1 or ename = 'abc') and deptno > 1 from emp";
+    final RelNode rel = convertSql(sql);
     final RelMetadataQuery mq = rel.getCluster().getMetadataQuery();
 
     final RexNode ref = RexInputRef.of(0, rel.getRowType().getFieldList());
-    final Set<RexNode> r = mq.getExpressionLineage(rel, ref);
-    assertThat(r.size(), is(1));
+    final Set<RexNode> r = Objects.requireNonNull(mq.getExpressionLineage(rel, ref));
```
vlsi: is `requireNonNull` needed here?

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
asolimando: I have merged the two assertions since they would be overlapping and provided a more comprehensive error message:

```
Lineage for expr '$0' in node 'rel#2:LogicalProject.(input=LogicalTableScan#1,exprs=[AND(OR(=($0, 1), =($1, 'abc')), >($7, 1))])' for query 'select (empno = 1 or ename = 'abc') and deptno > 1 from emp':
'empno' is column 0 in 'catalog.sales.emp', 'ename' is column 1 in 'catalog.sales.emp', and 'deptno' is column 7 in 'catalog.sales.emp' ==> expected: <AND(OR(=([CATALOG, SALES, EMP].#0.$0, 1), =([CATALOG, SALES, EMP].#0.$1, 'abc')), >([CATALOG, SALES, EMP].#0.$7, 1))]> but was: <[AND(OR(=([CATALOG, SALES, EMP].#0.$0, 1), =([CATALOG, SALES, EMP].#0.$1, 'abc')), >([CATALOG, SALES, EMP].#0.$7, 1))]>
Expected :AND(OR(=([CATALOG, SALES, EMP].#0.$0, 1), =([CATALOG, SALES, EMP].#0.$1, 'abc')), >([CATALOG, SALES, EMP].#0.$7, 1))]
Actual   :[AND(OR(=([CATALOG, SALES, EMP].#0.$0, 1), =([CATALOG, SALES, EMP].#0.$1, 'abc')), >([CATALOG, SALES, EMP].#0.$7, 1))]
```

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
vlsi: The motivation is exactly the same: test failure should include meaningful information.
If you add `requireNonNull`, `getOnlyElement`, etc, etc, then you make it harder to understand the nature of the failure.

This case is trivial: the test expects a set with a single element. The assertion message is more or less OK.
Why adding more checks before the final `assertEquals`?
I believe, the final `assertEquals` should be more than enough to tell that "expected ... got null".

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -2482,6 +2483,46 @@ private void checkPredicates(RelOptCluster cluster, RelOptTable empTable,
...
+  }
+
+  @Test void testExpressionLineageBetweenExpressionWithJoin() {
+    final String sql = "select dept.deptno + empno between 1 and 2"
+        + " from emp join dept on emp.deptno = dept.deptno";
+    final RelNode rel = convertSql(sql);
+    final RelMetadataQuery mq = rel.getCluster().getMetadataQuery();
+
+    final RexNode ref = RexInputRef.of(0, rel.getRowType().getFieldList());
+    final Set<RexNode> r = Objects.requireNonNull(mq.getExpressionLineage(rel, ref));
```
vlsi, line 2512: This seems to be duplicated across methods, so it would be nice to extract it into `assertExpressionLineage(sql, columnIndex, Supplier<String> message)`

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -2482,6 +2497,37 @@ private void checkPredicates(RelOptCluster cluster, RelOptTable empTable,
...
+  }
+
+  @Test void testExpressionLineageBetweenExpressionWithJoin() {
+    String sql = "select dept.deptno + empno between 1 and 2"
+        + " from emp join dept on emp.deptno = dept.deptno";
+    String expected = "[AND(>=(+([CATALOG, SALES, DEPT].#0.$0, [CATALOG, SALES, EMP].#0.$0), 1),"
+        + " <=(+([CATALOG, SALES, DEPT].#0.$0, [CATALOG, SALES, EMP].#0.$0), 2))]";
+
+    BiFunction<RexNode, RelNode, String> supplier = (ref, rel) -> "Lineage for expr '"
+        + ref + "' in node '" + rel + "'" + " for query '" + sql + "':\n"
```
vlsi, line 2522: Ah, it looks like `BiFunction` and `Supplier` is not realy needed here. Just `String comment` argument is enough for `assertExpressionLineage`, and add the prefix in `assertExpressionLineage`

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -2399,6 +2400,20 @@ private void checkPredicates(RelOptCluster cluster, RelOptTable empTable,
...
+  // ----------------------------------------------------------------------
+
+  private void assertExpressionLineage(String sql, int columnIndex, String expected,
+      BiFunction<RexNode, RelNode, String> message) {
+    RelNode rel = convertSql(sql);
+    RelMetadataQuery mq = rel.getCluster().getMetadataQuery();
+    RexNode ref = RexInputRef.of(columnIndex, rel.getRowType().getFieldList());
+    Set<RexNode> r = mq.getExpressionLineage(rel, ref);
+
+    assertEquals(expected, r.toString(), message.apply(ref, rel));
```
vlsi, line 2414: `r.toString()` -> `String.valueOf(r)`

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -2399,6 +2399,21 @@ private void checkPredicates(RelOptCluster cluster, RelOptTable empTable,
...
+
+  private void assertExpressionLineage(
+      String sql, int columnIndex, String expected, String comment) {
+    RelNode rel = convertSql(sql);
+    RelMetadataQuery mq = rel.getCluster().getMetadataQuery();
+    RexNode ref = RexInputRef.of(columnIndex, rel.getRowType().getFieldList());
+    Set<RexNode> r = mq.getExpressionLineage(rel, ref);
+
+    assertEquals(expected, String.valueOf(r), "Lineage for expr '" + ref + "' in node '"
+        + rel + "'" + " for query '" + sql + "':n" + comment);
```
vlsi, line 2414: `:n`?

asolimando: Sorry about that, I lost a character in the refactoring and I did not notice.

vlsi (approved):

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
asolimando: Actually it was not my doing, it was `./gradlew autostyleApply` doing that:

```
     assertEquals(expected, String.valueOf(r), "Lineage for expr '" + ref + "' in node '"
-        + rel + "'" + " for query '" + sql + "':\n" + comment);
+        + rel + "'" + " for query '" + sql + "':n"
+  + comment);
   }
```

Looks like a bug to me.

vlsi: https://github.com/apache/calcite/blob/fd3c1d0bdf907d192d42c18b3aff5306eec650e7/build.gradle.kts#L542 might be relevant

asolimando: I have removed the `\n` to avoid issues and filed https://github.com/autostyle/autostyle/issues/30

vlsi: `"\\n\"\n  +"` should probably be `"\\\\n\"\n  +"`

vlsi: @asolimando , I believe the issue has little to do with autostyle, and it seems to be the issue with Calcite's customization.

Please check Calcite-specific regex replacements (see build.gradle.kts#L542)

zabetak (approved):

PR #2614, opened 2021-11-19
[CALCITE-4895] MAP type in UDF cannot be externalized from json correctly

chunweilei:


core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -296,6 +296,13 @@ private RelDataType getRelDataType(RelDataTypeFactory typeFactory, Map<String, O
       componentType = toType(typeFactory, component);
       return typeFactory.createArrayType(componentType, -1);
 
+    case MAP:
+      Object key = requireNonNull(map.get("key"), "key");
+      Object value = requireNonNull(map.get("value"), "key");
```
vlsi, line 301:
```suggestion
      Object key = get(map, "key");
      Object value = get(map, "key");
```

chunweilei: Thanks for your view. All comments are addressed.

core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -387,6 +394,12 @@ private Object toJson(RelDataType node) {
       if (node.getComponentType() != null) {
         map.put("component", toJson(node.getComponentType()));
       }
+      if (node.getKeyType() != null) {
+        map.put("key", toJson(requireNonNull(node.getKeyType(), "key type")));
+      }
+      if (node.getValueType() != null) {
+        map.put("value", toJson(requireNonNull(node.getValueType(), "value type")));
+      }
```
vlsi, line 402: `requireNonNull` would not be required if you save `getKeyType()` result to a variable

chunweilei: Thank you for your kindly remind. It helps a lot.

PR #2320, opened 2021-01-10
[CALCITE-4458] The "rowTypeCoercion" for "SET" operations should be "OR" relationship (Jiatao Tao)

Aaaaaaron:
More details in JIRA.

PR #2615, opened 2021-11-20
[CALCITE-4897] Implicit type conversion is not complete for set operation in DML

JiajunBernoulli:
# insert union case

 Here is the sql:
```
insert into t3 
select 'a', 1.0, 1 union 
select 'b', 2, 2 union 
select 'c', 3.0, CAST(3 AS SMALLINT) union 
select 'd', 4.0, 4 union 
select 'e', 5.0, 5 
```
but the validated sql is :
```
INSERT INTO `CATALOG`.`SALES`.`T3`
SELECT 'a', CAST(1.0 AS INTEGER), CAST(1 AS SMALLINT) UNION 
SELECT 'b', 2, CAST(2 AS SMALLINT) UNION -- Encountered an integer and ended early
SELECT 'c', 3.0, CAST(3 AS SMALLINT) UNION -- Encountered an samlint and ended early
SELECT 'd', 4.0, 4 UNION -- should be cast  
SELECT 'e', 5.0, 5 -- should be cast 
```
This is why CALCITE-4458 changed '&&' to '||', the right validated sql is:
```
INSERT INTO `CATALOG`.`SALES`.`T3`
SELECT 'a', CAST(1.0 AS INTEGER), CAST(1 AS SMALLINT)
UNION
SELECT 'b', 2, CAST(2 AS SMALLINT)
UNION
SELECT 'c', CAST(3.0 AS INTEGER), CAST(3 AS SMALLINT)
UNION
SELECT 'd', CAST(4.0 AS INTEGER), CAST(4 AS SMALLINT)
UNION
SELECT 'e', CAST(5.0 AS INTEGER), CAST(5 AS SMALLINT)
 ```

# insert values case
Here is the sql
```
insert into t3 values ('a', 1.0, 1), ('b', 2, 2), ('c', 3.0, CAST(3 AS SMALLINT)), ('d', 4.0, 4), ('e', 5.0, 5)
```
 the validated sql is :
```
INSERT INTO `CATALOG`.`SALES`.`T3`
VALUES ROW('a', CAST(1.0 AS INTEGER), CAST(1 AS SMALLINT)),
ROW('b', 2, CAST(2 AS SMALLINT)),-- Encountered an integer and ended early
ROW('c', 3.0, CAST(3 AS SMALLINT)),Encountered an samllint and ended early
ROW('d', 4.0, 4),
ROW('e', 5.0, 5) 
```
the right validated sql is:
```
INSERT INTO `CATALOG`.`SALES`.`T3`
VALUES ROW('a', CAST(1.0 AS INTEGER), CAST(1 AS SMALLINT)),
ROW('b', 2, CAST(2 AS SMALLINT)),
ROW('c', CAST(3.0 AS INTEGER), CAST(3 AS SMALLINT)),
ROW('d', CAST(4.0 AS INTEGER), CAST(4 AS SMALLINT)),
ROW('e', CAST(5.0 AS INTEGER), CAST(5 AS SMALLINT))
```

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
@@ -121,8 +121,13 @@ public class TypeCoercionImpl extends AbstractTypeCoercion {
       // Set operations are binary for now.
       final SqlCall operand0 = ((SqlCall) query).operand(0);
       final SqlCall operand1 = ((SqlCall) query).operand(1);
-      final boolean coerced = rowTypeCoercion(scope, operand0, columnIndex, targetType)
-          && rowTypeCoercion(scope, operand1, columnIndex, targetType);
+      // Operand1 should be coerced even if operand0 not need to be coerced.
+      // For example, we have one table named t:
+      // INSERT INTO t -- only one column is c(int).
+      // SELECT 1 UNION   -- operand0 not need to be coerced.
+      // SELECT 1.0  -- operand1 should be coerced.
+      boolean coerced = rowTypeCoercion(scope, operand0, columnIndex, targetType);
+      coerced = rowTypeCoercion(scope, operand1, columnIndex, targetType) || coerced;
       // Update the nested SET operator node type.
       if (coerced) {
         updateInferredColumnType(
--- a/core/src/test/java/org/apache/calcite/test/TypeCoercionConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/TypeCoercionConverterTest.java
@@ -126,6 +126,24 @@ class TypeCoercionConverterTest extends SqlToRelTestBase {
     sql(sql).ok();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-4897">[CALCITE-4897]
+   * Set operation in DML, implicit type conversion is not complete</a>. */
+  @Test void testInsertUnionQuerySourceCoercion() {
+    final String sql = "insert into t1 "
+        + "select 'a', 1, 1.0,"
+        + " 0, 0, 0, 0, TIMESTAMP '2021-11-28 00:00:00', date '2021-11-28', x'0A', false union "
+        + "select 'b', 2, 2,"
+        + " 0, 0, 0, 0, TIMESTAMP '2021-11-28 00:00:00', date '2021-11-28', x'0A', false union "
+        + "select 'c', CAST(3 AS SMALLINT), 3.0,"
+        + " 0, 0, 0, 0, TIMESTAMP '2021-11-28 00:00:00', date '2021-11-28', x'0A', false union "
+        + "select 'd', 4, 4.0,"
+        + " 0, 0, 0, 0, TIMESTAMP '2021-11-28 00:00:00', date '2021-11-28', x'0A', false union "
+        + "select 'e', 5, 5.0,"
+        + " 0, 0, 0, 0, TIMESTAMP '2021-11-28 00:00:00', date '2021-11-28', x'0A', false";
+    sql(sql).ok();
+  }
+
   @Test void testUpdateQuerySourceCoercion() {
     final String sql = "update t1 set t1_varchar20=123, "
         + "t1_date=TIMESTAMP '2020-01-03 10:14:34', t1_int=12.3";
--- a/core/src/test/resources/org/apache/calcite/test/TypeCoercionConverterTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/TypeCoercionConverterTest.xml
@@ -117,6 +117,23 @@ LogicalTableModify(table=[[CATALOG, SALES, T1]], operation=[INSERT], flattened=[
 ]]>
     </Resource>
   </TestCase>
+  <TestCase name="testInsertUnionQuerySourceCoercion">
+    <Resource name="sql">insert into t1 select 'a', 1, 1.0, 0, 0, 0, 0, TIMESTAMP '2021-11-28 00:00:00', date '2021-11-28', x'0A', false union select 'b', 2, 2, 0, 0, 0, 0, TIMESTAMP '2021-11-28 00:00:00', date '2021-11-28', x'0A', false union select 'c', CAST(3 AS SMALLINT), 3.0, 0, 0, 0, 0, TIMESTAMP '2021-11-28 00:00:00', date '2021-11-28', x'0A', false union select 'd', 4, 4.0, 0, 0, 0, 0, TIMESTAMP '2021-11-28 00:00:00', date '2021-11-28', x'0A', false union select 'e', 5, 5.0, 0, 0, 0, 0, TIMESTAMP '2021-11-28 00:00:00', date '2021-11-28', x'0A', false</Resource>
+    <Resource name="plan">
+      <![CDATA[
+LogicalTableModify(table=[[CATALOG, SALES, T1]], operation=[INSERT], flattened=[false])
+  LogicalUnion(all=[false])
+    LogicalUnion(all=[false])
+      LogicalUnion(all=[false])
+        LogicalUnion(all=[false])
+          LogicalValues(tuples=[[{ 'a', 1, 1, 0, 0, 0, 0, 2021-11-28 00:00:00, 2021-11-28, X'0a', false }]])
+          LogicalValues(tuples=[[{ 'b', 2, 2, 0, 0, 0, 0, 2021-11-28 00:00:00, 2021-11-28, X'0a', false }]])
+        LogicalValues(tuples=[[{ 'c', 3, 3, 0, 0, 0, 0, 2021-11-28 00:00:00, 2021-11-28, X'0a', false }]])
+      LogicalValues(tuples=[[{ 'd', 4, 4, 0, 0, 0, 0, 2021-11-28 00:00:00, 2021-11-28, X'0a', false }]])
+    LogicalValues(tuples=[[{ 'e', 5, 5, 0, 0, 0, 0, 2021-11-28 00:00:00, 2021-11-28, X'0a', false }]])
+]]>
+    </Resource>
+  </TestCase>
   <TestCase name="testNotInOperation">
     <Resource name="sql">
       <![CDATA[select

### CODE REVIEW

**Comment by beyond1920** [comment]
@JiajunBernoulli LGTM. I only left a minor comment.

File: core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java [hunk 1]
```diff
@@ -104,14 +104,17 @@ public TypeCoercionImpl(RelDataTypeFactory typeFactory, SqlValidator validator)
...
-          return false;
+          coercedValues = false;
         }
       }
-      updateInferredColumnType(
-          requireNonNull(scope, "scope"), query, columnIndex, targetType);
-      return true;
+      if (coercedValues) {
+        updateInferredColumnType(
+            requireNonNull(scope, "scope"), query, columnIndex, targetType);
```
danny0405, line 115: What's the difference here ? The original code does the same behavior: `#updateInferredColumnType` only if all the operands are coerced.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -121,8 +124,8 @@ public TypeCoercionImpl(RelDataTypeFactory typeFactory, SqlValidator validator)
       // Set operations are binary for now.
       final SqlCall operand0 = ((SqlCall) query).operand(0);
       final SqlCall operand1 = ((SqlCall) query).operand(1);
-      final boolean coerced = rowTypeCoercion(scope, operand0, columnIndex, targetType)
-          && rowTypeCoercion(scope, operand1, columnIndex, targetType);
+      boolean coerced = rowTypeCoercion(scope, operand0, columnIndex, targetType);
```
danny0405, line 127: Change `&&` directly to `||` ?

core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
```diff
@@ -894,6 +894,13 @@ public MockCatalogReader init() {
       t2.addColumn("t2_binary", binaryType);
       t2.addColumn("t2_boolean", booleanType);
       registerTable(t2);
+
+      final MockTable t3 =
+          MockTable.create(this, tSchema, "T3", false, 7.0, null);
+      t3.addColumn("t3_varchar20", varchar20Type);
+      t3.addColumn("t3_int", intType);
```
danny0405, line 901: Why add a new table here ? 't1' and 't2' have all the data types you needed.

danny0405 (changes requested):

core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
JiajunBernoulli: I'm worried that too **many columns** will affect everyone's reading, so I only use **a few columns.** If you think it is more appropriate to use the existing table, I can change it later.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
JiajunBernoulli: I thought so at first. But many test will be failed, for example:
```
 sql("select 1 from (values(true)) union values (select '1' from (values (true)) as tt)")
        .type("RecordType(VARCHAR EXPR$0) NOT NULL");
```
```
org.opentest4j.AssertionFailedError: expected: <RecordType(VARCHAR EXPR$0) NOT NULL> 
but was: <RecordType(VARCHAR NOT NULL EXPR$0) NOT NULL>
```
I think it is only executed when coerced values is true, like the following code
```      
// Update the nested SET operator node type.
if (coerced) {
  updateInferredColumnType(
      requireNonNull(scope, "scope"), query, columnIndex, targetType);
}
```

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
JiajunBernoulli: This will not fix the problem. Because when `rowTypeCoercion(scope, operand0, columnIndex, targetType)=false`,  `rowTypeCoercion(scope, operand1, columnIndex, targetType) `will not be run.
I will use `|`, do you agree?
```
final boolean coerced = rowTypeCoercion(scope, operand0, columnIndex, targetType)
    | rowTypeCoercion(scope, operand1, columnIndex, targetType);
```

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
danny0405: > org.opentest4j.AssertionFailedError: expected: <RecordType(VARCHAR EXPR$0) NOT NULL> 
but was: <RecordType(VARCHAR NOT NULL EXPR$0) NOT NULL>

Shouldn't this be the correct behavior ? Two constant literals are type nullable false, and implicit type coercion does not solve nullability in-consistency.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
danny0405: Yeah, you code is correct, we should not add any short-cut logic here.

core/src/test/java/org/apache/calcite/test/TypeCoercionConverterTest.java
```diff
@@ -133,4 +162,24 @@
...
+    } catch (Exception e) {
+      throw TestUtil.rethrow(e);
+    }
+    final RelDataTypeFactory typeFactory = tester.getValidator().getTypeFactory();
+    final Prepare.CatalogReader catalogReader =
+        tester.createCatalogReader(typeFactory);
+    final SqlValidator validator =
+        tester.createValidator(catalogReader, typeFactory);
+
+    final String actual = validator.validate(sqlQuery).toString();
```
danny0405, line 182: Somehow we need some refactoring for this testing infrustructure, for e.g, a `Sql` util, we can do that in another PR.

core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
danny0405: Yeah, i'm sliding to reuse the original tables.

JiajunBernoulli: I deleted `t3` and used `t1`.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
JiajunBernoulli: done

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
JiajunBernoulli: I modified the expected values of failed unit tests. They should all be not null. Are there any other questions?

core/src/test/java/org/apache/calcite/test/TypeCoercionConverterTest.java
JiajunBernoulli: I'm interested in it. `checkValidateSqlEquals` function can be abstracted. 
Do you think it's more appropriate to put it in the parent class `SqlToRelTestBase` or create an interface `SqlToValidateTestBase`?

core/src/test/java/org/apache/calcite/test/TypeCoercionConverterTest.java
```diff
@@ -133,4 +172,24 @@
...
+
+  private void checkValidateSqlEquals(String sql) {
+    Objects.requireNonNull(sql, "sql");
+    final SqlNode sqlQuery;
+    try {
+      sqlQuery = tester.parseQuery(sql);
+    } catch (RuntimeException | Error e) {
+      throw e;
+    } catch (Exception e) {
+      throw TestUtil.rethrow(e);
```
vlsi, line 184: Consider using `catch(Throwable e) { TestUtil.rethrow(e, "sql=" + sql) }`
That would be shorter, it would catch all throwables, and it would add the SQL in question

core/src/test/java/org/apache/calcite/test/TypeCoercionConverterTest.java
vlsi: @JiajunBernoulli , have you seen https://lists.apache.org/thread/3l9y4fv31kthmpn9pbzqdmfp80yrjphf by @julianhyde ?

core/src/test/java/org/apache/calcite/test/TypeCoercionConverterTest.java
JiajunBernoulli: I have modified it, thank you very much for your advice.

core/src/test/java/org/apache/calcite/test/TypeCoercionConverterTest.java
JiajunBernoulli: I didn't know before. I will continue to pay attention to it

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -121,8 +122,9 @@ public TypeCoercionImpl(RelDataTypeFactory typeFactory, SqlValidator validator)
       // Set operations are binary for now.
       final SqlCall operand0 = ((SqlCall) query).operand(0);
       final SqlCall operand1 = ((SqlCall) query).operand(1);
-      final boolean coerced = rowTypeCoercion(scope, operand0, columnIndex, targetType)
-          && rowTypeCoercion(scope, operand1, columnIndex, targetType);
+      @SuppressWarnings("ShortCircuitBoolean")
+      boolean coerced = rowTypeCoercion(scope, operand0, columnIndex, targetType)
+          | rowTypeCoercion(scope, operand1, columnIndex, targetType);
       // Update the nested SET operator node type.
```
danny0405, line 128: We should not use `|` in the logical operator.

JiajunBernoulli: I didn't find a more suitable way, just modify the original code.

danny0405 (approved):
Overall looks good now, would refactor the test code when i merge it ~

JiajunBernoulli:
> Overall looks good now, would refactor the test code when i merge it ~
Hello, the PR has been approved for five months. Do you have any plans to refactor the unit test recently? I want to create a new JIRA ticket to refactor the unit test this month. After refactoring, merge the commit.

JiajunBernoulli:
@danny0405  Hello, I refactor the tests and rebase main branch. Ci failed for other reasons, so the PR is ready for Merge. Would you please merge it?  
BTW, I delete insert values case because many conflicts need to discuss, I will create new jiar ticket to solve it. The PR only solve problem for set operation.

JiajunBernoulli:
@danny0405  Would you please review the pr again?

JiajunBernoulli:
@rubenada, You left a [comment](https://github.com/apache/calcite/pull/2320#issuecomment-757456022) on CALCITE-4458(The solution is not correct and lacks unit tests).  
I believe you should be familiar with implicit type cast. The PR has been approved by dannny, but I can't contact him now. Would you please review this PR? Thanks.

rubenada:
LGTM

danny0405 (approved):
+1， thanks for the fix, looks good from my side.

JiajunBernoulli:
@danny0405, @rubenada . Thanks for your review.

PR #2616, opened 2021-11-23
[CALCITE-4871] CAST a literal to DECIMAL type return wrong result

NobiGo:


core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -677,6 +684,74 @@ Expression translateCast(
...
+      convert = Expressions.call(operand, BuiltInMethod.BIG_DECIMAL_SET_SCALE.method,
+          Expressions.constant(targetType.getScale()), Expressions.constant(RoundingMode.HALF_UP));
+      break;
+    default:
+      break;
+    }
+    return convert;
+  }
+
+  private @NotNull Expression translateCharToDecimal(Expression operand, RelDataType targetType,
```
vlsi, line 731: We assume notnull by default, so there's no need for explicit notnull

NobiGo: Ok. Thanks you. When the PR is ready I will repair this. 
Actually I need more advire about this PR.
Because When I try to make the cast operator to return the right result(I can confirm the test added in misc.iq have right behavior), then many unit tests throw exception or result have a different scale. So I want to know  how to handle this?

zinking: > 

guess I don't have a direct reply, but usually I will create another fork of the repo. and compare the behavior simultaneously with the patch and without the patch and adjust correpsondingly.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
```diff
@@ -1744,7 +1744,7 @@ public static boolean isValidDecimalValue(@Nullable BigDecimal value, RelDataTyp
     case DECIMAL:
       final int intDigits = value.precision() - value.scale();
       final int maxIntDigits = toType.getPrecision() - toType.getScale();
-      return intDigits <= maxIntDigits;
+      return intDigits <= maxIntDigits && value.scale() == toType.getScale();
```
zinking, line 1747: this might be the culpit of the change, I would see this as an obivious change of the original intention

NobiGo: @zinking Thanks for your review. 
The method is used to judge can remove the cast operator from literal or not. 
If we just judge the precision. The example:
`cast(15.345 as decimal(5,1)) should be return 15.4 but return 15.345. `
If I have any misunderstanding please let me know.

PR #2617, opened 2021-11-25
fix parameter order of  rebuilding PLUS/TIMES

shpodg:
The SqlDatetimePlusOperator parameter order can not be changed


core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -376,7 +376,7 @@ private static RexNode canonizeNode(RexBuilder rexBuilder, RexNode condition) {
         return rexBuilder.makeCall(call.getOperator(), left, right);
       }
 
-      RexNode newCall = rexBuilder.makeCall(call.getOperator(), right, left);
+      RexNode newCall = rexBuilder.makeCall(call.getOperator(), left, right);
```
vlsi, line 379: I believe it is not an error.
The reordering is intentional: note the comparison in the previous line

shpodg: Thanks for review ,I have a testcase that can prove it and I will open an another PR .

vlsi: Please create a branch for the PR next time. PRs from master branches do not really work well.

PR #2620, opened 2021-11-28
[CALCITE-4907] JdbcJoin does not convert Join with always true/false …

francesco-gini88:
…RexNode

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcRules.java
+++ b/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcRules.java
@@ -332,6 +332,9 @@ public class JdbcRules {
      * @return Whether condition is supported
      */
     private static boolean canJoinOnCondition(RexNode node) {
+      if (node.isAlwaysTrue() || node.isAlwaysFalse()) {
+        return true;
+      }
       final List<RexNode> operands;
       switch (node.getKind()) {
       case AND:
--- a/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcRules.java
+++ b/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcRules.java
@@ -332,11 +332,11 @@ public class JdbcRules {
      * @return Whether condition is supported
      */
     private static boolean canJoinOnCondition(RexNode node) {
-      if (node.isAlwaysTrue() || node.isAlwaysFalse()) {
-        return true;
-      }
       final List<RexNode> operands;
       switch (node.getKind()) {
+      case LITERAL:
+        // literal on a join condition would be TRUE or FALSE
+        return true;
       case AND:
       case OR:
         operands = ((RexCall) node).getOperands();
--- a/core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
@@ -50,23 +50,24 @@ class JdbcAdapterTest {
    * same time. */
   private static final ReentrantLock LOCK = new ReentrantLock();
 
-  /** VALUES is not pushed down, currently. */
   @Test void testValuesPlan() {
     final String sql = "select * from \"days\", (values 1, 2) as t(c)";
-    final String explain = "PLAN="
-        + "EnumerableNestedLoopJoin(condition=[true], joinType=[inner])\n"
-        + "  JdbcToEnumerableConverter\n"
+    final String explain = "PLAN=JdbcToEnumerableConverter\n"
+        + "  JdbcJoin(condition=[true], joinType=[inner])\n"
         + "    JdbcTableScan(table=[[foodmart, days]])\n"
-        + "  EnumerableValues(tuples=[[{ 1 }, { 2 }]])";
+        + "    JdbcValues(tuples=[[{ 1 }, { 2 }]])";
     final String jdbcSql = "SELECT *\n"
-        + "FROM \"foodmart\".\"days\"";
+        + "FROM \"foodmart\".\"days\",\n"
+        + "(VALUES (1),\n"
+        + "(2)) AS \"t\" (\"C\")";
     CalciteAssert.model(FoodmartSchema.FOODMART_MODEL)
         .query(sql)
         .explainContains(explain)
         .runs()
         .enable(CalciteAssert.DB == CalciteAssert.DatabaseInstance.HSQLDB
             || CalciteAssert.DB == DatabaseInstance.POSTGRESQL)
-        .planHasSql(jdbcSql);
+        .planHasSql(jdbcSql)
+        .returnsCount(14);
   }
 
   @Test void testUnionPlan() {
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcRules.java
```diff
@@ -332,6 +332,9 @@ protected JdbcJoinRule(Config config) {
      * @return Whether condition is supported
      */
     private static boolean canJoinOnCondition(RexNode node) {
+      if (node.isAlwaysTrue() || node.isAlwaysFalse()) {
+        return true;
+      }
```
NobiGo, line 337: Add another case condition:
```
// ***************
case LITERAL:
        return true;
```
is more suitable ?

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -402,6 +400,29 @@
             + "FROM \"SCOTT\".\"DEPT\") AS \"t1\" ON \"t0\".\"DEPTNO\" = \"t1\".\"DEPTNO\"");
   }
 
+  @Test void testJoinConditionAlwaysTruePushDown() {
+    CalciteAssert.model(JdbcTest.SCOTT_MODEL)
+        .query(
+            "select empno, ename, d.deptno, dname\n"
+                + "from scott.emp e,scott.dept d\n"
+                + "where true")
+        .explainContains(
```
julianhyde, line 409: please match the formatting of strings in other tests. double quotes right after open parentheses.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -50,23 +50,24 @@
    * same time. */
   private static final ReentrantLock LOCK = new ReentrantLock();
 
-  /** VALUES is not pushed down, currently. */
   @Test void testValuesPlan() {
```
julianhyde, line 53: update the comment; don't remove it.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcRules.java
julianhyde: i agree. with a comment that this would be TRUE or FALSE.

francesco-gini88: shouldn't I check that then the literal is in fact a boolean?
```
case LITERAL:
        return ((RexLiteral) node).getTypeName() == SqlTypeName.BOOLEAN;
```

or perhaps there is some guarantee that something that is not a boolean can't show up there?

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -50,23 +50,25 @@
...
     final String sql = "select * from \"days\", (values 1, 2) as t(c)";
-    final String explain = "PLAN="
-        + "EnumerableNestedLoopJoin(condition=[true], joinType=[inner])\n"
-        + "  JdbcToEnumerableConverter\n"
+    final String explain = "PLAN=JdbcToEnumerableConverter\n"
+        + "  JdbcJoin(condition=[true], joinType=[inner])\n"
         + "    JdbcTableScan(table=[[foodmart, days]])\n"
-        + "  EnumerableValues(tuples=[[{ 1 }, { 2 }]])";
+        + "    JdbcValues(tuples=[[{ 1 }, { 2 }]])";
```
vlsi, line 59: In practice, this becomes a less efficient plan, and it would be more efficient to "duplicate in-memory" rows rather than "duplicate rows at the database side, and spend extra network and CPU resources for processing the duplicates".

So in a proper cost model, this should not be pushed to Jdbc.

julianhyde: The JDBC rules' job is to apply all valid transformations. RelToSql's job is to convert a plan to SQL. Somewhere else there is a cost model that determines whether executing the whole query in the backend database is the optimal plan. So this test is correct.

vlsi: > So this test is correct

`VALUES is pushed down` can't be correct as it is unclear **why** it should be pushed.

The test could be created for one of the following:
a) "SQL with VALUES doesn't result in a runtime error"
b) "this specific SQL pushes VALUES to JDBC because..."
c) "this specific SQL does NOT push VALUES to JDBC because..."

I think the test case is more like "doesn't result in a runtime error", so `VALUES is pushed down` comment above is misleading.

If someone adjusts the costing model, the planner would realize pushing the values to JDBC is less efficient (e.g. it produces more network traffic, it consumes more CPU), so the "proper" behavior should be "values is NOT pushed to JDBC".

----

If you want to test that "VALUES can be pushed at all", then the test case should better include some join condition, so nobody questions the efficiency of the pushing VALUES.

---

I don't really like how this discussion consumes time, however, you suggested "update the comment" instead of removing it, and I clarify that the comments (both old and new) are misleading. They add no value, and even the test method name could convey the same (or better) message.

francesco-gini88: FWIW I think the test and comment `VALUES is not pushed down, currently.` was giving the false impression that  `Values` can't be pushed down into the db. A query like `select * from "days", (values 1, 2) as t(c) where t.c = "days"."day"` would have had a plan with VALUES pushed down
```
"PLAN=JdbcToEnumerableConverter\n" 
        + "  JdbcJoin(condition=[=($2, $0)], joinType=[inner])\n" 
        + "    JdbcTableScan(table=[[foodmart, days]])\n" 
        + "    JdbcValues(tuples=[[{ 1 }, { 2 }]])"
```
The reason why `VALUES` was not pushed down is because the cross join couldn't be pushed down in the first place. 

On the point that the plan is less efficient and that it would be more efficient to "duplicate in-memory" rows. I can see where your observation is coming from. However, this test case is quite simple. In reality I would expect other aspects to matter as well. For instance in the in memory implementation of the cross join I _think_ that the inner iterator is instanciated multiple times. If the inner iterator happens to be `JdbcToEnumerableConverter` that would equate to running a query multiple times, because the inner iterator is created multiple times. This would mean fetching rows multiple times plus adding latency multiple times.

In any case, I don't think this is the pr to address VALUES optimisations

amaliujia: is the Values pushed because 
```
      case LITERAL:
        // literal on a join condition would be TRUE or FALSE
        return true;
```
so that Values were treated as Literal? 

If we have a strong argument here maybe only returns true when it is a boolean (true/false) and leave a TODO/Jira for whether pushing Values?

NobiGo: I think This PR has addressed what is described in the Summary. And @vlsi comments are reasonable. So I support creating another Jira to add more unit tests about the VALUES is pushed down or not.

francesco-gini88: @amaliujia my understanding is that Values is pushed down because in that way the join is also pushed down. What's changed is that now the join can be pushed down, because it is a cross join and 
```
    case LITERAL:
        // literal on a join condition would be TRUE or FALSE
        return true;
```
the condition is always the literal true.
Therefore I figure the planner is picking the plan where the join is pushed down (generally speaking might be better to push down the join, although maybe not this particular case) and consequently the Values are also pushed down. For this reason I don't think a TODO belongs to that part of the code. I think a Jira describing exactly in which cases/queries Values should not be pushed down would be a better alternative.

I'm happy to add a join condition if that makes the test clearer.
```
select * from \"days\", (values 1, 2) as t(c) where \"day\" = t.c
```
A query like this one wouldn't be affected by my change (the join condition is not a literal) and the Values would still be pushed down

francesco-gini88:
hi there. This PR hasn't made the cut for one of the previous release. I'd like to complete the work, but I am unclear on what is left that I should be doing. There is an open thread around  the efficiency of a plan generated by a join with `values`, can someone advise if any change is needed for the PR to go in or whether we don't want the change because of concern of plan efficiency, in which case I'd like to close the pr

zabetak (approved):
I think the PR is in good shape to be merged. I went over the existing discussions and I don't see something really blocking. There are quite a few people involved in the reviews but if nobody objects I will merge it in a few days.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcRules.java
zabetak: `Join#isValid` ensures that the condition, in this case the literal, is always a Boolean so there is no need for additional checks thus marking this conversation as resolved.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
zabetak: I think that the discussion around the test naming, comments, and purpose is not particularly crucial for this PR. I am not sure what was the purpose of this test before the changes introduced in this PR, and I am not sure either what is the purpose from now on going forward. Thus, I plan to accept the latest changes which merely update the comment.

The **main point** I keep from this conversation is that the change in this PR can potentially lead to important performance regression when a Cartesian product is pushed into the JDBC backend. Nevertheless, as @julianhyde mentioned, the decision to push a cartesian product or not should be part of the cost model rather than the rule which defines valid transformations thus I think it makes sense to go forward with the current PR as it is.

PR #2621, opened 2021-11-30
[CALCITE-4908] Support classpath resource name as model resource path too

woonsan:


ILuffZhe:
Hi, @woonsanko , something goes wrong during CI which requires code modification, and how about adjusting the headline of PR to "[CALCITE-4908] Support classpath resource URI as model URI".

vlsi:
@woonsan , Do you think you can create an end-to-end test case with a filename that includes whitespace?

julianhyde:
@vlsi, @woonsan, Such a test case would be useful. However, if you add a file to git that contains whitespace it will probably break tooling. We have no files that contain whitespace currently. Maybe you can accomplish the test case some other way.

woonsan:
@vlsi , I guess we will probably need to take care of the potential cases where people may use file paths with spaces on different machines, but I personally prefer keeping this ticket as small as possible only focusing on `classpath:` model resource url support for now.
Can we discuss in the dev@ list for the potential improvements regarding file paths - not classpath resource paths - with space(s) separately? It would be also great if someone can provide a real use (test) cases, too and what the real problems look like.

vlsi:
@woonsan , see https://github.com/apache/calcite/commit/2f33a0c57b7b7e77b8193d0fff1e3531119aee0a#diff-7350bd9b77a86a4eef5e2b2b000cd9c9ae8bf30f1264222c1fed43a688b967e4

You add the following documentation note:

> resource URI like `classpath:/com/example/calcite/model.yaml`

However, the URI standard does have a way to encode special symbols.
What is missing is that you mention URI, and you miss to clarify how special symbols (e.g. `+`, `:`, `/`, `space`, emojis, whatever) should be represented in that URI.
Shoud it be `classpath:/com/hello%20world.yml` or `classpath:/com/hello world.yml`?

----

Side note: when you add `.getResource...`, you need a class loader. Sometimes Calcite's classloader is fine, and sometimes `threadcontext classloader` would be slightly better. So I would suggest you allow both options (e.g. via different prefixes or something).

---

All in all, I do not block the PR in any way (even though I think it should be improved before the merge), and I feel I won't have time to review it later.

woonsan:
Thanks @vlsi !

I've rephrased the docs to use classpath resource name or path consistently, not 'URI' any more. Looking at `Class.getResource(String name)` API doc, the `name` is just a Java-specific 'path', not anything like 'URI'.
Regarding support for the thread's current context classloader's resource loading, I'm not sure if it is really necessary, practically speaking.

If someone wants to put calcite-core.jar in a shared classloader while trying to load the model classpath resource in an sub-classloader in some container env, it could be a problem to them. Even in that case, I'd recommend them to deploy calcite-core.jar in their specific app (in sub-classloader), for safety. There could be a good use case in a new emerging env in the future too, but I'd rather wait for hearing about a practical use case first.

Thanks,

Woonsan

woonsan:
Closing this as it's too simple to take time any more.

PR #2622, opened 2021-11-30
[CALCITE-4898] Upgrading Elasticsearch version from 7.0.1 to 7.15.2

ILuffZhe:
This PR upgrades embedded Elasticsearch version from 7.0.1 to 7.15.2.

**Description**:

- New dependencies: _org.codelibs.elasticsearch.module:scripting-painless-spi_, as module "org.elasticsearch.painless.spi" is removed after ES 7.15.0 in lang-painless
- Third maven repo: _org.codelibs.elasticsearch.module:lang-painless_ is no longer maintained after ES 7.10.2, which is migrated to https://maven.codelibs.org/
- RestClient Upgrading: the low level rest client in ES has good compatibilities(just http request) among 7.X, which is also upgraded to 7.15.2

**Self-verification**: I've run some tests locally to make sure new feature can be applied(not added in unit test).

- Supported: new features like RareTerms、minimun_interval in auto_date_histogram can be successfully applied, which are not supported in ES 7.0.1
- Not supported: top_metrics、multi_terms、rate and other features in x-pack are not supported currently. Those features can be registered when AnalyticsPlugin is loaded to build ES Node(test environment), however, dependency _org.elasticsearch.plugin:x-pack-analytics_ cannot be reached through maven central at present


elasticsearch/build.gradle.kts
```diff
@@ -20,6 +20,12 @@ plugins {
     id("com.github.vlsi.gradle-extensions")
 }
 
+repositories {
+    maven {
+        setUrl("https://maven.codelibs.org/")
```
jacques-n, line 25: Why do we need to add a repository? It seems like we already source some stuff from org.codelibs, is this new things now available in the existing repositories?

elasticsearch/build.gradle.kts
```diff
@@ -40,7 +46,7 @@ dependencies {
     // https://github.com/elastic/elasticsearch/issues/49218
```
jacques-n, line 46: Is this issue not fixed in elastic 7.15.2? It looks like it was fixed a couple years ago...

elasticsearch/build.gradle.kts
ILuffZhe: After ES 7.10.2, "org.codelibs.elasticsearch.module:lang-painless" is no longer maintained in central maven repo, which migrated to https://maven.codelibs.org/, so does "org.codelibs.elasticsearch.module:scripting-painless-spi".
That's why I added a new repo.(see https://github.com/codelibs/elasticsearch-module#version)

jacques-n (commented):
This looks good, can you squash and rebase?

Thanks

elasticsearch/build.gradle.kts
ILuffZhe: Yes, this issue is already solved, and I've removed the constraints. Thanks!

ILuffZhe:
Done. Thanks for your review. @jacques-n

jacques-n:
@ILuffZhe , I just realized that your commit message isn't the standard message format for Calcite commits. It should be `[CALCITE-XXXX] Short description.`

ILuffZhe:
@jacques-n Sorry for that, I've corrected it.

jacques-n:
Thanks @ILuffZhe

julianhyde:
@jacques-n This change adds new a package repository. Are we OK with that?

vlsi:
@ILuffZhe , @jacques-n , if we have to use a new repository, can you please configure content filtering so we don't attempt resolving other artifacts in the codelibs repository?
For instance, https://docs.gradle.org/current/userguide/declaring_repositories.html#declaring_content_exclusively_found_in_one_repository

julianhyde:
I really hope we don't use a new repository.

vlsi:
+100500
@julianhyde , thanks for noticing this.

laurentgo:
What about the licensing change which happens post 7.10.x? As covered in https://www.elastic.co/pricing/faq/licensing,tThe main elasticsearch jars are now under the Elastic licence or the SSPL v1, but not under the ASL 2.0. The SSPL license is in category X (https://www.apache.org/legal/resolved.html#category-x), which means it is actually forbidden to include in the project.

As for the other codelibs module, the POM file says they are licensed under ASL 2.0, but it seems they are rebundling elasticsearch artifacts from the distribution, and as far as I can tell they are also redistributed under the Elastic license or the SSPL v1.

ILuffZhe:
@laurentgo Thanks for the clarification and gradle tips by @vlsi .
@julianhyde I can understand your concern, if a new repo is unacceptable, we may consider upgrading ES to 7.10.2.

jacques-n:
> This change adds new a package repository. Are we OK with that?

I did identify the addition and was okay with it given the following:
- it was constrained to test dependencies
- it was constrained to a single module
- it seemed benign

I didn't know about the ability to do content filtering or I would have asked that it was included.  However, if many others have a strong aversion, we can revert. What do we think we should suggest people do in this kind of situation? Especially in this situation where new functionality isn't being added here which sources a new repo, instead existing dependency jars have moved to a new repo.

> Category X license

I didn't realize that the ES license change was in a minor version (sigh). That being said, I don't see a problem with this patch. We already depend on category X licenses for build/test (JMH is an example, I believe) and this is consistent with ASF policies. Category X are prohibited in source releases and convenience binaries and I don't believe this change introduces any violation of that. As I understand it, the rest client we depend on for non-test code remains ASL2.0 licensed (only the "highlevel" library moved to the category X licenses: [ll license](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/_license.html), [ll-hl distinction](https://stackoverflow.com/questions/62340904/elasticsearch-rest-high-level-client-vs-elasticsearch-rest-client)).

#### Summary
- This seemed like a relatively innocuous change which is why I merged it. If there is a strong discomfort to the inclusion of a new repository (even at this constrained level), we can revert this and ask the contributor to find an alternative path.
- If we are uncomfortable using category X licenses in the ES test module, we should probably remove ES as a supported adapter and either deprecate entirely or move to the AWS opensearch fork.

vlsi:
> We already depend on category X licenses for build/test (JMH is an example, I believe) 

JMH is GPL with classpath exception, and that is explicitly allowed. There's no X in JMH.

> Category X are prohibited in source releases and convenience binaries and I don't believe this change introduces any violation of that

The spirit of ASF is that we produce code that **can** be modified by the consumers.
Unfortunately, category X does limit the consumers, so it is much better to avoid X licenses for all the cases.

Here's a recent case when ASF convinced **third-party** to remove a transitive LGPL dependency: https://issues.apache.org/jira/browse/LEGAL-572

> As I understand it, the rest client we depend on for non-test code

Code without tests is not really helpful 🤷 
Do you consider "ES adapter" to be a test code for Calcite?
I just thought "ES adapter" was a standalone helpful library.

>X licenses in the ES test module, we should probably remove ES as a supported adapter

I am not sure what are the options regarding forks, however, removing ES adapter or staying with the older version are reasonable options.

jacques-n:
> JMH is GPL with classpath exception, and that is explicitly allowed

Honest question: where is this explicitly allowed? It's been a while since I've done ASF license validation. It seems like it is still category X according to my quick skim of the [3rd party licensing page](https://www.apache.org/legal/resolved.html) as well as various legal discussions such as [this one](https://issues.apache.org/jira/browse/LEGAL-396?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=16551865#comment-16551865) and [this one](https://markmail.org/thread/4tvidrumrjnbfqhj).

> The spirit of ASF is that we produce code...

My intention wasn't to talk about spirit. I was trying to talk about rules. I agree that the spirit is to avoid category X.  

> Code without tests is not really helpful 🤷 

Most Calcite users use Calcite + Adapters as libraries without using the Calcite tests. I think they find it very helpful.

vlsi:
> Honest question: where is this explicitly allowed?

I do not want to spend hours on licensing stuff, so let me try with an example.
`java.sql.Connection` is licensed under GPL + Classpath_Exception.
Apparently Calcite and Avatica implement those interfaces, so there's a really tight link between Calcite and `GPL + Classpath_Exception`.

If `GPL + Classpath_Exception` were forbidden, then Calcite was not possible in the first place.

The same applies to JMH and JOL. They have exactly the same license as `java.sql.Connection` and `java.lang.Object`. Apparenlty we are allowed to depend on `java.lang.Object`, so there's no way `JMH` is forbidden.

> Most Calcite users use Calcite + Adapters as libraries without using the Calcite tests. I think they find it very helpful.

If we keep "tests for Calcite" private, and we release "production code only", then clients won't be able to produce derived works. They won't know if their customizations break invariants.

jacques-n:
> java.sql.Connection is licensed under GPL + Classpath_Exception

You're mixing api with implementation. There are implementations of java.sql.Connection that are ASL2.0 such as [here](https://github.com/apache/harmony/blob/trunk/classlib/modules/sql/src/main/java/java/sql/Connection.java).

> there's no way JMH is forbidden

I didn't say JMH is forbidden, I said it is Category X according to Apache legal (GPL+CPE). Category X is allowed as optional and for build.

> I have no idea if the problematic classes are used for testing ES adapter only or if they are used for implementing it.

I did do the legwork on review. For reference from gradle build file (with my added license notes):

```kotlin
// Category A
api("org.elasticsearch.client:elasticsearch-rest-client")

// Category A (according to github license, Laurent comment not withstanding)
testImplementation("org.codelibs.elasticsearch.module:lang-painless")
testImplementation("org.codelibs.elasticsearch.module:scripting-painless-spi")

// Category X
testImplementation("org.elasticsearch:elasticsearch")
```

So unless I misunderstand the gradle syntax, only the tests are relying on Category X (same as JMH).

julianhyde:
Regarding category X.

I accept that we can use category X licensed components in testing.

However, they are radioactive, and I think they should be labeled as category X in the source code. Without the 'Category X' comment, it would be easy for someone to change `testImplementation` to `implementation`.

It's especially worth calling out because not everyone knows that Elasticsearch has changed license.

Regarding repositories.

I don't think it's worth the pain and fragility of adding a new repository, even for one release, and even with gradle guard rails that limit the impact to just this component.

Anyone who is writing serious open source Java code needs to get it into Maven Central.

A hundred IT people are going to have to add a new repository to their corporate Maven mirror, cursing Calcite for being high maintenance.

How much do we need to back out or downgrade in order to do without the repository?

jacques-n:
> I don't think it's worth the pain and fragility of adding a new repository, even for one release, and even with gradle guard rails that limit the impact to just this component.

Sounds good to me. I've opened [CALCITE-4948](https://issues.apache.org/jira/browse/CALCITE-4948) to address.

PR #2623, opened 2021-12-01
CALCITE-4913 Correlated variables in a select list are not deduplicated

korlov42:


core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4346,7 +4336,21 @@ private void convertSelectList(
...
-    bb.setRoot(relBuilder.build(), false);
+
+    RelNode project = relBuilder.build();
+
+    final RelNode r;
+    final CorrelationUse p = getCorrelationUse(bb, project);
+    if (p != null) {
+      assert p.r instanceof Project;
+      Project proj = (Project) p.r;
+      r = LogicalProject.create(proj.getInput(), proj.getHints(),
```
zabetak, line 4347: I've seen that this pattern appears in other places in `SqlToRelConverter` but it is a bit weird that from one hand we are asserting that the project is an instance of a superclass (`Project`) and on the other hand we are creating a `LogicalProject`. Can we use the `relbuilder` to create the new project? Would it be better to rely on `project#copy`?

core/src/main/java/org/apache/calcite/rel/logical/LogicalProject.java
```diff
@@ -35,52 +37,63 @@
...
 import java.util.List;
+import java.util.Objects;
+import java.util.Set;
 
 /**
  * Sub-class of {@link org.apache.calcite.rel.core.Project} not
  * targeted at any particular engine or calling convention.
  */
 public final class LogicalProject extends Project {
+  private final ImmutableSet<CorrelationId> variablesSet;
```
zabetak, line 53: I think it makes more sense to put `variablesSet` in the superclass (`Project`) rather than here.

vlsi: Does `variablesSet` belong to `Project` at all? I thought it was `Correlate` that sets variables.

zabetak: @vlsi  I thought so too but apparently `variablesSet` is also on `LogicalFilter`, `Join`, and `Correlate`.

vlsi: Oh, it looks like https://github.com/apache/calcite/commit/505a9064b96a6c8399735fc2fa4d0ac9d5f3ed87 and https://issues.apache.org/jira/browse/CALCITE-1045 are relevant.

core/src/main/java/org/apache/calcite/rel/logical/LogicalProject.java
```diff
@@ -137,14 +170,20 @@ public static LogicalProject create(final RelNode input, List<RelHint> hints,
 
   @Override public RelNode withHints(List<RelHint> hintList) {
     return new LogicalProject(getCluster(), traitSet, hintList,
-        input, getProjects(), getRowType());
+        input, getProjects(), getRowType(), variablesSet);
+  }
+
+  @Override public RelWriter explainTerms(RelWriter pw) {
+    return super.explainTerms(pw)
+        .itemIf("variablesSet", variablesSet, !variablesSet.isEmpty());
```
vlsi, line 178: Please print `variablesSet` **before** the regular project fields as it would be easier to relate project and its variables.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -11183,7 +11183,7 @@ LogicalProject(EXPR$0=[> SOME($0, {
 LogicalProject(DEPTNO=[$0])
   LogicalFilter(condition=[=($cor0.JOB, $1)])
     LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
-})])
+})], variablesSet=[[$cor0]])
```
vlsi, line 11186: This looks really confusing. We should print variables side-by-side with Project, and variables should go before fields names.

core/src/main/java/org/apache/calcite/rel/logical/LogicalProject.java
korlov42: It does make sense for project with correlated scalar queries. For example, the query like `SELECT t0.v, (SELECT t0.v + t1.v FROM test_tbl t1) AS j FROM test_tbl t0` has a logical plan:
```
LogicalProject(V=[$2], J=[$SCALAR_QUERY({
LogicalProject(T0.V + T1.V=[+($cor0.V, $2)])
  IgniteLogicalTableScan(table=[[PUBLIC, TEST_TBL]])
})])
  IgniteLogicalTableScan(table=[[PUBLIC, TEST_TBL]])
```
In order to evaluate this plan you need either implement filter that set correlation variable, or convert it to LogicalCorrelate (what is actually done by `SubQueryRemoveRule` ).

With that said, it seems to me LogicalFilter has to expose the variables they set, but not every physical implementation would do so.

vlsi: @korlov42 , thanks for the clarifications. I see. I thought we did not have that in Calcite (I thought Correlate was the only way to set correlate vars).

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
korlov42: I think this assertion primary used to avoid "Unchecked cast" warning throwing by a static code analyzers. Here we are converting sqlNode to a relation tree, so LogicalProject is a reasonable result. Parent class `Project` give us sufficent information to create a new `LogicalProject` with found set of correlated variables.

We could use the `relBuilder`, but it doesn't change a lot.

core/src/main/java/org/apache/calcite/rel/logical/LogicalProject.java
korlov42: So, I moved `variableSet` to a `Project` class in order to explain this term before the regular project fields

core/src/main/java/org/apache/calcite/rel/logical/LogicalProject.java
korlov42: fixed

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
korlov42: fixed, I suppose

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -673,7 +707,7 @@ LogicalProject(DEPTNO=[$7], EXPR$1=[$9])
     </Resource>
     <Resource name="planNotExpanded">
       <![CDATA[
-LogicalProject(DEPTNO=[$7], EXPR$1=[$SCALAR_QUERY({
+LogicalProject(variablesSet=[[$cor0]], DEPTNO=[$7], EXPR$1=[$SCALAR_QUERY({
```
vlsi, line 710: `variablesSet` looks really odd here. What do you think of `correlation=[[$cor0]]`?

korlov42: A `correlation` sounds more natural to me too, but `variablesSet` is consistent with `LogicalFilter`. Do we need to change this too?

core/src/main/java/org/apache/calcite/rel/core/Project.java
```diff
@@ -89,25 +94,33 @@ protected Project(
...
-      RelDataType rowType) {
+      RelDataType rowType,
+      Set<CorrelationId> variablesSet) {
     super(cluster, traits, input);
     assert rowType != null;
     this.exps = ImmutableList.copyOf(projects);
     this.hints = ImmutableList.copyOf(hints);
     this.rowType = rowType;
+    this.variablesSet = Objects.requireNonNull(
+        ImmutableSet.copyOf(variablesSet), "variablesSet");
```
vlsi, line 105: Can `copyOf` ever return `null`?

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
korlov42: ok, renamed this term for project only for now

core/src/main/java/org/apache/calcite/rel/core/Project.java
korlov42: No, it's an artifact of moving this check to parent class :(. `ImmutableSet.copyOf` will throw an NPE if someone will pass a null parameter. So we need to wrap an argument `variablesSet` with `requiredNonNull`, and pass the result to `copyOf`.

Thanks for noticing! Fixed.

core/src/main/java/org/apache/calcite/rel/core/Project.java
```diff
@@ -70,17 +71,21 @@
...
-   * @param hints    Hints of this relation expression
-   * @param input    Input relational expression
-   * @param projects List of expressions for the input columns
-   * @param rowType  Output row type
+   * @param cluster      Cluster that this relational expression belongs to
+   * @param traits       Traits of this relational expression
+   * @param hints        Hints of this relation expression
+   * @param input        Input relational expression
+   * @param projects     List of expressions for the input columns
+   * @param rowType      Output row type
```
vlsi, line 86: Frankly speaking, I think we should avoid reflowing the comments. There's no point in changing all the previous lines when you add a new one.
See https://google.github.io/styleguide/javaguide.html#s4.6.3-horizontal-alignment

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
zabetak: The class allows somebody to pass their own customized `RelBuilder` which can create a `HiveProject` instead of `LogicalProject`. In various places the `SqlToRelConverter` creates explicitly `Logical` expressions instead passing via the `relBuilder` which currently makes the `RelBuilder` customization kind of useless but this may change in the near future.

zabetak:
@korlov42  In order to facilitate the review, can you put in the description the SQL query along with the respective (problematic) plan? If it causes wrong results then please include expected and actual results. It would be good also to add a test in `sub-query.iq` or another similar place to ensure that the query actually runs and produces the expected results.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
korlov42: Oh, now I see. Then it definitely would be better to rewrite with relBuilder.

core/src/main/java/org/apache/calcite/rel/core/Project.java
korlov42: reverted this change

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
korlov42: Hi, @zabetak! I've reworked PR in order to use `relBuilder` to create a correlated project after deduplication. Could you take a look again please?

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableRelFactories.java
```diff
@@ -68,7 +70,9 @@
       implements org.apache.calcite.rel.core.RelFactories.ProjectFactory {
     @Override public RelNode createProject(RelNode input, List<RelHint> hints,
         List<? extends RexNode> childExprs,
-        @Nullable List<? extends @Nullable String> fieldNames) {
+        @Nullable List<? extends @Nullable String> fieldNames,
+        Set<CorrelationId> variablesSet) {
+      Preconditions.checkArgument(variablesSet.isEmpty());
```
vlsi, line 75: Would you please add a clarification message?

core/src/main/java/org/apache/calcite/rel/core/RelFactories.java
```diff
@@ -162,10 +162,20 @@ private RelFactories() {
...
     RelNode createProject(RelNode input, List<RelHint> hints,
-        List<? extends RexNode> childExprs, @Nullable List<? extends @Nullable String> fieldNames);
+        List<? extends RexNode> childExprs, @Nullable List<? extends @Nullable String> fieldNames,
+        Set<CorrelationId> variablesSet);
+
+    @Deprecated // to be removed before 2.0
+    default RelNode createProject(RelNode input, List<RelHint> hints,
+        List<? extends RexNode> childExprs, @Nullable List<? extends @Nullable String> fieldNames) {
+      return createProject(input, hints, childExprs, fieldNames, ImmutableSet.of());
+    }
```
vlsi, line 178: I wonder if there's a better way to evolve interfaces like this.

1) I am not sure we should deprecate `variablesSet=empty` method. I believe, it would be the most commonly used case, and it would be sad to pass `ImmutableSet.of()` all over the place.

2) I am not sure how often Calcite consumers implement `ProjectFactory`, however, adding a non-default method breaks consumers.

3) Does this change mean that every implementation has to implement `variablesSet` in `Project`? In other words, what if `mongo` adapter does not support correlation columns as a part of `Project`? What should it do? Should the rules be updated to convert the project into something else?

vlsi:
@korlov42 , What do you suggest for cases like https://github.com/apache/calcite/blob/f3c09365c2eecbb5198d9acbef638f76053a49a9/mongodb/src/main/java/org/apache/calcite/adapter/mongodb/MongoRules.java#L302-L318 ?

It looks like you did not refactor `mongodb` adapter, and I guess it would result in loss of semantics. It looks like `variablesSet` would be discarded.
The sad thing is Calcite users would get similar issues: they would get wrong results/wrong plans, and nobody would know the reason.

WDYT?

core/src/main/java/org/apache/calcite/rel/core/RelFactories.java
korlov42: I used [CALCITE-3183](https://issues.apache.org/jira/browse/CALCITE-3183) as reference. According to Julian Hydes review comments, `Factories should be minimal. The old method should be default and deprecated. See SortFactory.`

korlov42:
> The sad thing is Calcite users would get similar issues: they would get wrong results/wrong plans, and nobody would know the reason.

Before this patch, they could get a correlated projection without being notified this particular project should set a correlation variable (`project#getVariablesSet()` returns an empty set). The only change is that now a `project` returns a valid set of variables it set.

core/src/main/java/org/apache/calcite/rel/logical/LogicalProject.java
```diff
@@ -123,12 +143,27 @@ public static LogicalProject create(final RelNode input, List<RelHint> hints,
         cluster.traitSet().replace(Convention.NONE)
             .replaceIfs(RelCollationTraitDef.INSTANCE,
                 () -> RelMdCollation.project(mq, input, projects));
-    return new LogicalProject(cluster, traitSet, hints, input, projects, rowType);
+    return new LogicalProject(cluster, traitSet, hints, input, projects, rowType,
+        ImmutableSet.of());
+  }
```
vlsi, line 148: Could you delegate to the new method instead of duplicating the logic?

korlov42: fixed

core/src/main/java/org/apache/calcite/rel/core/Project.java
```diff
@@ -89,25 +94,33 @@ protected Project(
...
     this.rowType = rowType;
+    this.variablesSet = ImmutableSet.copyOf(
+        Objects.requireNonNull(variablesSet, "variablesSet"));
     assert isValid(Litmus.THROW, null);
   }
 
+  protected Project(RelOptCluster cluster, RelTraitSet traits, List<RelHint> hints,
+      RelNode input, List<? extends RexNode> projects, RelDataType rowType) {
+    this(cluster, traits, hints, input, projects, rowType, ImmutableSet.of());
+  }
```
vlsi, line 112: In practice, this change breaks backward compatibility: you add a "convenience" constructor, and now the clients do not know they **must** somehow handle the new `variablesSet` parameter (e.g. by throwing on non-empty set)

I would suggest the method to be removed or at least marked with `@Deprecated` so the clients know which places to fix.

korlov42: mark this constructor as `@Deprecated`

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableRelFactories.java
korlov42: done

korlov42:
@vlsi, I've added a `@Deprecated` on the Project's constructor. This highlighted a bunch of missed places (primarily, adapters). So all of them are fixed now, I suppose.

With this deprecation users should be aware of changes. But, IMO, despite now the project rel returns correlation variables, the semantic is still the same. Before this patch it was possible to set correlated scalar subquery as a projection expression. The only difference now is there is an easy way to extract all variables  required by such a correlated scalar subqueries.

vlsi:
@korlov42 , I've lost interest in Calcite, and I have no time for reviewing the PR.

zabetak:
Hi @korlov42 , I am very busy in the next 10 days or so I don't think I will have time to review the latest changes immediately. After that however, if it is still un-merged please ping me again and I will try to get this in; it seems to be very close.

korlov42:
Hi, @zabetak! Seems like 10 days have passed. Now it's my turn to be unavailable for next 10 days, but I really hope you'll find time to review this.

korlov42:
Hi, @zabetak! Sorry for pushing, but is there any chance you'll find the time to help me get this PR done?

cassandra/src/main/java/org/apache/calcite/adapter/cassandra/CassandraRules.java
```diff
@@ -258,7 +258,8 @@ private static boolean isEqualityOnKey(RexNode node, List<String> fieldNames,
   public static class CassandraProjectRule extends CassandraConverterRule {
     /** Default configuration. */
     private static final Config DEFAULT_CONFIG = Config.INSTANCE
-        .withConversion(LogicalProject.class, Convention.NONE,
+        .withConversion(LogicalProject.class, p -> p.getCorrelVariable() == null
+                || p.getVariablesSet().isEmpty(), Convention.NONE,
```
zabetak, line 262: Do we need to check both `p.getCorrelVariable()` and `p.getVariablesSet()`? From the API and existing implementations I get the impression that `p.getVariablesSet()` is a superset of `p.getCorrelVariable()`.

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -664,6 +664,46 @@ LogicalProject(EMPNO=[$0], JOB=[$2])
...
+      LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+})])
+  LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+]]>
+    </Resource>
+    <Resource name="planExpanded">
+      <![CDATA[
+LogicalProject(DEPTNO=[$7], EXPR$1=[$9], EXPR$2=[$10])
+  LogicalCorrelate(correlation=[$cor1], joinType=[left], requiredColumns=[{7}])
+    LogicalCorrelate(correlation=[$cor0], joinType=[left], requiredColumns=[{7}])
```
zabetak, line 688: Are the changes in `SubqueryRemoveRule` done so that you get a different correlate variable here?

If yes can you explain why it is a problem to have the same correlated variable. I am asking this cause we have other plans in the repo where multiple correlates have the same correlation id. Check plans for:

* RelOptRulesTest#testDecorrelateTwoExists
* RelOptRulesTest#testDecorrelateTwoExists#testDecorrelateTwoIns

core/src/test/resources/sql/sub-query.iq
```diff
@@ -3257,6 +3257,34 @@ EnumerableCalc(expr#0..7=[{inputs}], expr#8=[null:BOOLEAN], proj#0..8=[{exprs}])
...
+|     30 | 14 |  0 |
+|     30 | 14 |  0 |
+|     30 | 14 |  0 |
+|     30 | 14 |  0 |
+|     30 | 14 |  0 |
+|     30 | 14 |  0 |
++--------+----+----+
+(14 rows)
+
+!ok
```
zabetak, line 3286: Apart from the result please also include the `!plan`

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -7074,17 +7114,17 @@ from dept]]>
     </Resource>
     <Resource name="plan">
```
zabetak, line 7115: If we don't have a similar query in `sub-query.iq` please include this as well so that test that it actually runs.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcRules.java
```diff
@@ -544,7 +547,9 @@ public JdbcProject(RelOptCluster cluster, RelTraitSet traitSet,
     }
 
     @Override public JdbcProject copy(RelTraitSet traitSet, RelNode input,
-        List<RexNode> projects, RelDataType rowType) {
+        List<RexNode> projects, RelDataType rowType, Set<CorrelationId> variablesSet) {
+      Preconditions.checkArgument(variablesSet.isEmpty(),
+          "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

core/src/main/java/org/apache/calcite/interpreter/Bindables.java
```diff
@@ -403,12 +405,14 @@ protected BindableProjectRule(Config config) {
...
-      super(cluster, traitSet, ImmutableList.of(), input, projects, rowType);
+      super(cluster, traitSet, ImmutableList.of(), input, projects, rowType, ImmutableSet.of());
       assert getConvention() instanceof BindableConvention;
     }
 
     @Override public BindableProject copy(RelTraitSet traitSet, RelNode input,
-        List<RexNode> projects, RelDataType rowType) {
+        List<RexNode> projects, RelDataType rowType, Set<CorrelationId> variablesSet) {
+      Preconditions.checkArgument(variablesSet.isEmpty(),
+          "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

cassandra/src/main/java/org/apache/calcite/adapter/cassandra/CassandraProject.java
```diff
@@ -42,13 +46,15 @@
...
+    super(cluster, traitSet, ImmutableList.of(), input, projects, rowType, ImmutableSet.of());
     assert getConvention() == CassandraRel.CONVENTION;
     assert getConvention() == input.getConvention();
   }
 
   @Override public Project copy(RelTraitSet traitSet, RelNode input,
-      List<RexNode> projects, RelDataType rowType) {
+      List<RexNode> projects, RelDataType rowType, Set<CorrelationId> variablesSet) {
+    Preconditions.checkArgument(variablesSet.isEmpty(),
+        "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableProject.java
```diff
@@ -80,7 +84,9 @@ public static EnumerableProject create(final RelNode input,
   }
 
   @Override public EnumerableProject copy(RelTraitSet traitSet, RelNode input,
-      List<RexNode> projects, RelDataType rowType) {
+      List<RexNode> projects, RelDataType rowType, Set<CorrelationId> variablesSet) {
+    Preconditions.checkArgument(variablesSet.isEmpty(),
+        "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableRelFactories.java
```diff
@@ -68,7 +70,10 @@
       implements org.apache.calcite.rel.core.RelFactories.ProjectFactory {
     @Override public RelNode createProject(RelNode input, List<RelHint> hints,
         List<? extends RexNode> childExprs,
-        @Nullable List<? extends @Nullable String> fieldNames) {
+        @Nullable List<? extends @Nullable String> fieldNames,
+        Set<CorrelationId> variablesSet) {
+      Preconditions.checkArgument(variablesSet.isEmpty(),
+          "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

core/src/test/java/org/apache/calcite/plan/volcano/TraitPropagationTest.java
```diff
@@ -386,7 +390,9 @@ public static PhysProj create(final RelNode input,
     }
 
     public PhysProj copy(RelTraitSet traitSet, RelNode input,
-        List<RexNode> exps, RelDataType rowType) {
+        List<RexNode> exps, RelDataType rowType, Set<CorrelationId> variablesSet) {
+      Preconditions.checkArgument(variablesSet.isEmpty(),
+          "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6565,11 +6568,13 @@ protected MyFilterRule(Config config) {
...
         RelDataType rowType) {
-      super(cluster, traitSet, ImmutableList.of(), input, projects, rowType);
+      super(cluster, traitSet, ImmutableList.of(), input, projects, rowType, ImmutableSet.of());
     }
 
     public MyProject copy(RelTraitSet traitSet, RelNode input,
-        List<RexNode> projects, RelDataType rowType) {
+        List<RexNode> projects, RelDataType rowType, Set<CorrelationId> variablesSet) {
+      Preconditions.checkArgument(variablesSet.isEmpty(),
+          "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

pig/src/main/java/org/apache/calcite/adapter/pig/PigProject.java
```diff
@@ -35,12 +39,14 @@
...
-    super(cluster, traitSet, ImmutableList.of(), input, projects, rowType);
+    super(cluster, traitSet, ImmutableList.of(), input, projects, rowType, ImmutableSet.of());
     assert getConvention() == PigRel.CONVENTION;
   }
 
   @Override public Project copy(RelTraitSet traitSet, RelNode input, List<RexNode> projects,
-      RelDataType rowType) {
+      RelDataType rowType, Set<CorrelationId> variablesSet) {
+    Preconditions.checkArgument(variablesSet.isEmpty(),
+        "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

mongodb/src/main/java/org/apache/calcite/adapter/mongodb/MongoProject.java
```diff
@@ -56,7 +60,9 @@ public MongoProject(RelOptCluster cluster, RelTraitSet traitSet,
   }
 
   @Override public Project copy(RelTraitSet traitSet, RelNode input,
-      List<RexNode> projects, RelDataType rowType) {
+      List<RexNode> projects, RelDataType rowType, Set<CorrelationId> variablesSet) {
+    Preconditions.checkArgument(variablesSet.isEmpty(),
+        "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

innodb/src/main/java/org/apache/calcite/adapter/innodb/InnodbProject.java
```diff
@@ -42,13 +46,15 @@
...
+    super(cluster, traitSet, ImmutableList.of(), input, projects, rowType, ImmutableSet.of());
     assert getConvention() == InnodbRel.CONVENTION;
     assert getConvention() == input.getConvention();
   }
 
   @Override public Project copy(RelTraitSet traitSet, RelNode input,
-      List<RexNode> projects, RelDataType rowType) {
+      List<RexNode> projects, RelDataType rowType, Set<CorrelationId> variablesSet) {
+    Preconditions.checkArgument(variablesSet.isEmpty(),
+        "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

geode/src/main/java/org/apache/calcite/adapter/geode/rel/GeodeProject.java
```diff
@@ -45,13 +49,15 @@
...
+    super(cluster, traitSet, ImmutableList.of(), input, projects, rowType, ImmutableSet.of());
     assert getConvention() == GeodeRel.CONVENTION;
     assert getConvention() == input.getConvention();
   }
 
   @Override public Project copy(RelTraitSet traitSet, RelNode input,
-      List<RexNode> projects, RelDataType rowType) {
+      List<RexNode> projects, RelDataType rowType, Set<CorrelationId> variablesSet) {
+    Preconditions.checkArgument(variablesSet.isEmpty(),
+        "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchProject.java
```diff
@@ -43,13 +47,15 @@
...
+    super(cluster, traitSet, ImmutableList.of(), input, projects, rowType, ImmutableSet.of());
     assert getConvention() == ElasticsearchRel.CONVENTION;
     assert getConvention() == input.getConvention();
   }
 
   @Override public Project copy(RelTraitSet relTraitSet, RelNode input, List<RexNode> projects,
-      RelDataType relDataType) {
+      RelDataType relDataType, Set<CorrelationId> variablesSet) {
+    Preconditions.checkArgument(variablesSet.isEmpty(),
+        "Correlated scalar subqueries is not supported");
```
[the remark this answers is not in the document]
korlov42: done

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -905,11 +905,11 @@ public static RelNode createCastRel(
     if (rename) {
       // Use names and types from castRowType.
       return projectFactory.createProject(input, hints, castExps,
-          castRowType.getFieldNames());
+          castRowType.getFieldNames(), ImmutableSet.of());
```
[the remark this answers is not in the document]
korlov42: Good catch! Fixed

core/src/main/java/org/apache/calcite/rel/core/Project.java
```diff
@@ -120,14 +134,15 @@ protected Project(RelInput input) {
         ImmutableList.of(),
         input.getInput(),
         requireNonNull(input.getExpressionList("exprs"), "exprs"),
-        input.getRowType("exprs", "fields"));
+        input.getRowType("exprs", "fields"),
+        ImmutableSet.of());
```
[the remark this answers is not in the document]
korlov42: fixed

core/src/main/java/org/apache/calcite/rel/logical/LogicalProject.java
```diff
@@ -58,29 +61,34 @@
...
       List<RelHint> hints,
       RelNode input,
       List<? extends RexNode> projects,
-      RelDataType rowType) {
-    super(cluster, traitSet, hints, input, projects, rowType);
+      RelDataType rowType,
+      Set<CorrelationId> variablesSet) {
+    super(cluster, traitSet, hints, input, projects, rowType, variablesSet);
     assert traitSet.containsIfApplicable(Convention.NONE);
   }
```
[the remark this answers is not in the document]
korlov42: yes, it's better to preserve compatibility. So, I've returned and deprecated the old constructor

cassandra/src/main/java/org/apache/calcite/adapter/cassandra/CassandraRules.java
korlov42: This looks like attempt to check if collection is null or empty, but something went wrong... Fixed this to verify both `p.getCorrelVariable()` and `p.getVariablesSet()`

core/src/main/java/org/apache/calcite/rel/rules/ProjectWindowTransposeRule.java
```diff
@@ -99,7 +100,7 @@ public ProjectWindowTransposeRule(RelBuilderFactory relBuilderFactory) {
 
     final LogicalProject projectBelowWindow =
         new LogicalProject(cluster, window.getTraitSet(), ImmutableList.of(),
-            window.getInput(), exps, builder.build());
+            window.getInput(), exps, builder.build(), ImmutableSet.of());
```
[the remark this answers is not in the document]
korlov42: If my understanding correct, we only transpose an identity project, otherwise the identity project will be derived and passed below the window and the topmost project will be adjusted. In that case it looks safe.

> Also it seems that here we are going to lose the variables set in the original projection

It's a bug, thank you for noticing

core/src/main/java/org/apache/calcite/rel/rules/FilterProjectTransposeRule.java
```diff
@@ -187,7 +187,7 @@ protected FilterProjectTransposeRule(
     RelNode newProject =
         config.isCopyProject()
             ? project.copy(project.getTraitSet(), newFilterRel,
```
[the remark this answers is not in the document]
korlov42: Do we really need to investigate _all the rules_ within this PR? This requires a lot of effort, whereas the benefit is small. I mean, for years downstream projects were dealing with correlated subqueries in the project, and no one cares about empty set returned from `getVariablesSet()`. Does it make sence to fix those rules step by step whenever the problem arises?

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -1794,7 +1794,74 @@ public RelBuilder project(Iterable<? extends RexNode> nodes,
...
+   * @param nodes Expressions
+   * @param fieldNames Suggested field names
+   * @param force create project even if it is identity
+   */
+  public RelBuilder projectCorrelated(
+      Iterable<CorrelationId> variablesSet,
+      Iterable<? extends RexNode> nodes,
+      Iterable<? extends @Nullable String> fieldNames,
+      boolean force) {
+    return project_(variablesSet, nodes, fieldNames, ImmutableList.of(), force);
```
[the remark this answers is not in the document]
korlov42: RelBuilder is used by downstream projects. Since now we could pass correlation variables to the project, I believe it's good idea to give an ability to do it with builder

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -2654,16 +2654,6 @@ protected void convertCollectionTable(
...
-    final SqlValidatorScope selectScope =
-        ((DelegatingScope) bb.scope()).getParent();
-    final Blackboard seekBb = createBlackboard(selectScope, null, false);
-
-    final CorrelationUse p = getCorrelationUse(seekBb, callRel);
-    if (p != null) {
-      assert p.r instanceof LogicalTableFunctionScan;
-      callRel = (LogicalTableFunctionScan) p.r;
-    }
-
```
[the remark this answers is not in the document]
korlov42: This snippet were introduced by CALCITE-4673 and addresses only particular case with table functions. My patch is targeted to fix the problem in general

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -129,16 +132,33 @@ protected RexNode apply(RexSubQuery e, Set<CorrelationId> variablesSet,
...
+    for (CorrelationId corrId : variablesSet) {
+      CorrelationId newId = cluster.createCorrel();
+      newVariablesSet.add(newId);
+      r = DeduplicateCorrelateVariables.go(cluster.getRexBuilder(), newId,
+          ImmutableSet.of(corrId), r);
+    }
+    builder.push(r);
+    final RelMetadataQuery mq = cluster.getMetadataQuery();
     final Boolean unique = mq.areColumnsUnique(builder.peek(),
         ImmutableBitSet.of());
```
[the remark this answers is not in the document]
korlov42: I've reverted this change. See [this](https://github.com/apache/calcite/pull/2623#discussion_r887873550) comment for details

core/src/main/java/org/apache/calcite/sql2rel/RelStructuredTypeFlattener.java
```diff
@@ -534,10 +535,21 @@ public void rewriteRel(LogicalProject rel) {
     RelNode newInput = getNewForOldRel(rel.getInput());
     List<RexNode> newProjects = Pair.left(flattenedExpList);
     List<String> newNames = Pair.right(flattenedExpList);
-    final RelNode newRel = relBuilder.push(newInput)
+    RelNode renamedProject =  relBuilder.push(newInput)
         .projectNamed(newProjects, newNames, true)
```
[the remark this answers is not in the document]
korlov42: I would prefer not to touch `RelBuilder#projectNamed` within this PR, otherwise it might require revising of all the places where `projectNamed` is called

core/src/test/resources/sql/sub-query.iq
korlov42: fixed

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
korlov42: fixed

PR #2624, opened 2021-12-02
[CALCITE-4914] Introduce HandlerBasedRelMetdataQuery as an extension point

jacques-n:
This is a precursor to implementing CALCITE-4879. It tries to better encapsulate/separate 
the consumption side of RelMetaDataQuery from the production side. This includes:
- Mark the public fields in RelMetadataQueryBase as INTERNAL fields
- Deprecate the static initialHandler method and replace it with a class level createInitialHandler
- Deprecate the public constructor of RelMetadataQuery to avoid direct extension
- Introduce a new HandlerBasedRelMetdataQuery that will house the handler specific
  concepts in a future release.


jacques-n:
Adding @julianhyde as a reviewer as he expressed some concerns around this work on Jira.

jacques-n:
Due to discussionsing on jira, abandoning.

PR #2625, opened 2021-12-02
[wip] Introduce a new generic visitor stateless visitor pattern that RelShuttle can extend.

jacques-n:


core/src/main/java/org/apache/calcite/rel/GenericRelVisitor.java
```diff
@@ -0,0 +1,76 @@
...
+  T visit(LogicalMatch match) throws E;
+
+  T visit(LogicalSort sort) throws E;
+
+  T visit(LogicalExchange exchange) throws E;
+
+  T visit(LogicalTableModify modify) throws E;
+
+  T visit(T other) throws E;
+
```
jaystarshot, line 75: Maybe we can add
`T visit(RelNode other) throws E; too`

jaystarshot (commented):
nit

core/src/main/java/org/apache/calcite/rel/GenericRelVisitor.java
jacques-n: Fixed. Darn search and replace...

core/src/main/java/org/apache/calcite/rel/RelNode.java
```diff
@@ -410,7 +410,20 @@ default boolean isEnforcer() {
...
+  /**
+   * Accepts a visit from a generic visitor.
+   *
+   * @param visitor
+   * @param <T> The return type of the visitor.
+   * @param <E> An exception that maybe thrown by the acceptance of the visitor.
+   * @return
+   * @throws E A configurable exception type based on the visitor.
+   */
+  <T, E extends Throwable> T accept(GenericRelVisitor<T, E> visitor) throws E;
```
vlsi, line 426: It seems to break backward compatibility as all rel nodes would have to implement the new method :-/

Suppose there's `DrillRelNode extends RelNode` or even `MockRel extends RelNode`.
How should they upgrade?

Would it be better to have a default implementation that throws?

core/src/main/java/org/apache/calcite/rel/GenericRelVisitor.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.calcite.rel.logical.LogicalProject;
+import org.apache.calcite.rel.logical.LogicalSort;
+import org.apache.calcite.rel.logical.LogicalTableModify;
+import org.apache.calcite.rel.logical.LogicalUnion;
+import org.apache.calcite.rel.logical.LogicalValues;
+
+/**
+ * Visitor that has methods for the common logical relational expressions.
+ */
+public interface GenericRelVisitor<T, E extends Throwable> {
```
vlsi, line 40: Naming is hard, however, I would like to refrain from `Generic`. We have a lot of generics in code, and it would be sad to have `Generic` prefix everywhere.

jacques-n:
Abandoning as no clear solution was found.

PR #2627, opened 2021-12-02
Fixed incorrect asserts

rendel:
It is not possible to use a range distribution with the current assert, it will always fail.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/RelDistributions.java
+++ b/core/src/main/java/org/apache/calcite/rel/RelDistributions.java
@@ -103,7 +103,7 @@ public class RelDistributions {
           || Ordering.natural().isOrdered(keys)
           : "key columns of hash distribution must be in order";
       assert type == Type.HASH_DISTRIBUTED
-          || type == Type.RANDOM_DISTRIBUTED
+          || type == Type.RANGE_DISTRIBUTED
           || keys.isEmpty();
     }
 


core/src/main/java/org/apache/calcite/rel/RelDistributions.java
```diff
@@ -103,7 +103,7 @@ private RelDistributionImpl(Type type, ImmutableIntList keys) {
           || Ordering.natural().isOrdered(keys)
           : "key columns of hash distribution must be in order";
       assert type == Type.HASH_DISTRIBUTED
-          || type == Type.RANDOM_DISTRIBUTED
+          || type == Type.RANGE_DISTRIBUTED
           || keys.isEmpty();
     }
```
chunweilei, line 108: @rendel Thank you for reporting this. I think it is a bug. Could you file a JIRA for this?

chunweilei:
BTW, you can find more details about contribution in the doc[1].

[1] https://calcite.apache.org/develop/#contributing

core/src/main/java/org/apache/calcite/rel/RelDistributions.java
vlsi: I would say a test case is more important than JIRA.
If the assertion is user-facing, then it should be a regular if + illegalargumentexception

chunweilei:
Let us move to https://github.com/apache/calcite/pull/2822.

PR #2628, opened 2021-12-02
[CALCITE-4917] Add test for 'IS NOT NULL(a) AND a=b' simplification

asolimando:
Simplification (from `RexSimplify` class) is mostly covered in `RexProgramTest`, but tests for expressions of the form `IS NOT NULL(a) AND a=b` (with nullable and not nullable types) seem to be missing.

Since I had to write tests to make sure the simplification was as expected, I assume others might end up doing the same, and that the tests will both act as documentation and it will also protect against regressions.

### DIFF
--- a/core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
+++ b/core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
@@ -1710,6 +1710,18 @@ class RexProgramTest extends RexProgramTestBase {
         .expandedSearch(expanded);
   }
 
+  @Test void testSimplifyAndIsNotNullWithEquality() {
+    // "AND(IS NOT NULL(x), =(x, y)) => AND(IS NOT NULL(x), =(x, y)) (unknownAsFalse=false),
+    // "=(x, y)" (unknownAsFalse=true)
+    checkSimplify2(and(isNotNull(vInt(0)), eq(vInt(0), vInt(1))),
+        "AND(IS NOT NULL(?0.int0), =(?0.int0, ?0.int1))",
+        "=(?0.int0, ?0.int1)");
+
+    // "AND(IS NOT NULL(x), =(x, y)) => "=(x, y)"
+    checkSimplify(and(isNotNull(vIntNotNull(0)), eq(vIntNotNull(0), vInt(1))),
+        "=(?0.notNullInt0, ?0.int1)");
+  }
+
   @Test void testSimplifyAndIsNull() {
     final RexNode aRef = input(tInt(true), 0);
     final RexNode bRef = input(tInt(true), 1);


core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -1710,6 +1710,17 @@ private void checkExponentialCnf(int n) {
...
   }
 
+  @Test void testSimplifyAndIsNotNullEquality() {
+    RelDataType intNullableType =
+        typeFactory.createTypeWithNullability(
+            typeFactory.createSqlType(SqlTypeName.INTEGER), true);
+    final RexInputRef i0 = rexBuilder.makeInputRef(intNullableType, 0);
+    final RexInputRef i1 = rexBuilder.makeInputRef(intNullableType, 1);
+
+    checkSimplify2(and(isNotNull(i0), eq(i0, i1)),
```
vlsi, line 1720:
```suggestion
    checkSimplify2(and(isNotNull(vIntNotNull(0)), eq(vIntNotNull(0), vIntNotNull(1))),
```

vlsi (approved):
Thanks, it looks good

vlsi:
Will merge after https://github.com/apache/calcite/pull/2629

asolimando:
@vlsi please hold on to merging this, I'd like to address @julianhyde's comments first, thanks!

chunweilei (approved):
LGTM

asolimando:
@vlsi, @chunweilei: do you want me to squash before merging or will you do it? In case, please use the updated Jira ticket title "Add test for 'a IS NOT NULL AND a = b' simplification" as commit message, the first commit message is stale. Thanks!

julianhyde:
@asolimando I acknowledge this is a particular challenge with the current workflow. (@vlsi please don't hit me over the head with it.)
* We want really good commit messages
* We want people to not squash/amend commits, to make them easier to review the deltas. Therefore the contributor can't change the commit message
* Some committers forget to look at JIRA comments before merging (and squashing and rebasing) the PR

asolimando:
> * We want people to not squash/amend commits, to make them easier to review the deltas. Therefore the contributor can't change the commit message

I agree on that when the review is on-going, but I have always done my squashing once all the active reviewers mark the PR "approved", just to spare some pain to the reviewers/committers (but also to have a final chance to write an extended commit message if the PR is a bit complex, definitely this is not the case). Does it pose problems too or you were referring to pending reviews?

> * Some committers forget to look at JIRA comments before merging (and squashing and rebasing) the PR

Probably I should have updated the PR title to reflect the new JIRA title.

vlsi:
I think we have already discussed "when it is time to squash", and I believe the outcome is like "as soon as one thinks there's a final state, it is a good time to squash".

AFAIK GitHub does show "force push" changes, so "force push" alone does not create problems for the review.

However, there's a notable case: "rebase on top of current master and squash at the same time". That results in "combined diff" where both "changes from master branch" and "changes from the squash" appear in the same diff.

In that case, I prefer to rebase first, force push, then squash and force-push again. It shows as two "force push" events in GitHub UI.

vlsi:
> In case, please use the updated Jira ticket title

I think PR author and contributor is in a much better position to update PR/JIRA titles than the reviewer/committer.
All in all, I would like for the contributors to update the PR/JIRA/whatever title so it reflects the actual intention rather than "the first idea which might even be abandoned long ago".

asolimando:
I have squashed the commits, updated the commit message to reflect the latest title of the JIRA ticket.
The second force push is a rebase on current master.

julianhyde:
@vlsi I agree with that workflow. I have to tell contributors quite often not to amend/rebase while review is ongoing, so we should make an effort to document for contributors the workflow that we would like them to follow.

asolimando:
If there are no objections I will merge in the next 24 hours

julianhyde:
@asolimando You should announce your intentions on the '[Calcite Release] Preparing for 1.29.0 Release' thread. As RM, Rui is about to lock down master branch.

PR #2631, opened 2021-12-03
[CALCITE-4915] Query with unqualified common column and NATURAL JOIN fails

alex-plekhanov:
Alternative version with a couple of methods refactoring


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -466,14 +466,14 @@ private boolean expandSelectItem(
...
+  private SqlNode expandExprFromJoin(SqlJoin join, SqlIdentifier identifier,
       @Nullable SelectScope scope) {
-    if (join.getConditionType() != JoinConditionType.USING) {
+    if (join.getConditionType() != JoinConditionType.USING && !join.isNatural()) {
       return identifier;
     }
 
     for (String name
-        : SqlIdentifier.simpleNames((SqlNodeList) getCondition(join))) {
+        : requireNonNull(usingNames(join), () -> "usingNames for " + join)) {
```
zinking, line 476: er, can this be extracted into a variable? it would be more debug friendly

alex-plekhanov: Fixed

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -286,6 +286,57 @@ public static void checkActualAndReferenceFiles() {
...
+        + "grouping(deptno, job), count(*)\n"
+        + "from emp\n"
+        + "natural join dept\n"
+        + "group by grouping sets ((deptno), (deptno, job))";
+    sql(sql).ok();
+  }
+
+  /** Similar to {@link #testJoinNaturalWithUnqualifiedCommonColumn()},
+   * but with multiple join. */
+  @Test void testJoinNaturalWithMultipleJoin() {
```
ViggoC, line 431: This test case failed after rebasing main branch.

mihaibudiu: how does it fail?

alex-plekhanov: @mihaibudiu plan have changed a little bit, additional cast added (from CHAR to VARCHAR). Related change https://issues.apache.org/jira/browse/CALCITE-6413
I've fixed the test to match the current plan, but perhaps it's an issue and we shouldn't cast CHAR types to VARCHAR

mihaibudiu:
CI fails, please ask for a review once it works again.

alex-plekhanov:
@mihaibudiu, done. CI passed now.

alex-plekhanov:
@mihaibudiu, looks like issue is fixed by #4410 and production code fix is not required anymore.
Do we still need test cases implemented by this ticket?

mihaibudiu:
Are you asking whether to add more test cases?
If you think that the test cases increase code coverage they are good.
If you are unsure, more test cases are usually good, especially the quidem test.
If you think they are redundant, then you can just close this PR.

alex-plekhanov:
@mihaibudiu, I mean #4410 fixes another problem and mainly has test cases related to that problem, but also fixes NATURAL JOIN and has a simple test for that. 
This pull-request (#2631) also has extended tests for NATURAL JOIN problem (aggregation with "group by" by natural join column, for example) and quidem test. I think these extedned tests still useful, because "group by" with common natural join column also can be broken somehow. For example, we still have problem with "order by" by natural join column: https://issues.apache.org/jira/browse/CALCITE-4922

core/src/test/resources/sql/join.iq
```diff
@@ -1004,4 +1004,30 @@ ON t1.a IS NOT DISTINCT FROM t2.a;
 
 !ok
 
+### [CALCITE-4915] Query with unqualified common column and NATURAL JOIN fails
```
mihaibudiu, line 1007: I assume you have validated the correctness of this output using some other means.

mihaibudiu (approved):
So this will enable us to close [CALCITE-4915]?

core/src/test/resources/sql/join.iq
mihaibudiu: @alex-plekhanov if you can answer these two questions maybe we can resolve this issue?

alex-plekhanov: Sorry for delay with answer. I've tested this query with PG, result is the same, except ordering. I've added ORDER BY now, to get more predictable results.
Yes, we can close CALCITE-4915

PR #2633, opened 2021-12-07
[CALCITE-4851] Build gives lots of 'Execution optimizations have been disabled' warnings

NobiGo:
…en disabled when build CALCITE

### DIFF
--- a/babel/build.gradle.kts
+++ b/babel/build.gradle.kts
@@ -14,6 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import com.github.autostyle.gradle.AutostyleTask
+
 plugins {
     id("com.github.vlsi.ide")
     calcite.fmpp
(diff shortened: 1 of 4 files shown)


vlsi:
We do not verify the code style for the generated code, so the dependencies should be `mustRunAfter` rather than `dependsOn`.

babel/build.gradle.kts
```diff
@@ -52,6 +52,14 @@ val javaCCMain by tasks.registering(org.apache.calcite.buildtools.javacc.JavaCCT
     packageName.set("org.apache.calcite.sql.parser.babel")
 }
 
+tasks.autostyleJavaCheck {
+    dependsOn(javaCCMain)
+}
+
+tasks.checkstyleMain {
+    dependsOn(javaCCMain)
+}
```
vlsi, line 61: This would break in case `-PskipCheckstyle` or `-PskipAutostyle`.
Something like https://github.com/apache/jmeter/blob/0938f50e15cc62e2de645b823def881364e64b78/src/core/build.gradle.kts#L145-L160 might do.

NobiGo: @vlsi  Please help review again. Thanks very much.

NobiGo:
> We do not verify the code style for the generated code, so the dependencies should be `mustRunAfter` rather than `dependsOn`.

done.

babel/build.gradle.kts
NobiGo: > This would break in case `-PskipCheckstyle` or `-PskipAutostyle`. Something like https://github.com/apache/jmeter/blob/0938f50e15cc62e2de645b823def881364e64b78/src/core/build.gradle.kts#L145-L160 might do.

done.

vlsi (approved):
LGTM

PR #2639, opened 2021-12-09
[CALCITE-4929] Implement default getDef() methods in MetadataHandler interfaces

jacques-n:
- Add default getDef() methods.
- Mark all function metadata handler interfaces as FunctionalInterface where possible.
- Update MetdataDef and RelMetdataHandlerGeneratorUtil to ignore getDef() methods.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
@@ -61,8 +61,13 @@ public abstract class BuiltInMetadata {
     @Nullable Double getSelectivity(@Nullable RexNode predicate);
 
     /** Handler API. */
+    @FunctionalInterface
     interface Handler extends MetadataHandler<Selectivity> {
       @Nullable Double getSelectivity(RelNode r, RelMetadataQuery mq, @Nullable RexNode predicate);
+
+      @Override default MetadataDef<Selectivity> getDef() {
+        return DEF;
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


jamesstarr:
Have you consider adding a default implementation directly to handler interface and deprecating the method?

jacques-n:
> Have you consider adding a default implementation directly to handler interface and deprecating the method?

It seems like your saying to add a default implementation to `MetadataHandler.getDef()`. I considered this but it requires the use of reflection to retrieve an expected static field on sub-interfaces. To me, this perpetuates a kind of brittle runtime binding that I think we should avoid/reduce in Calcite.

I didn't consider deprecating the method as I believe it is used in ReflectiveRelMetadataProvider and it seems like it would need to be refactored to support removal. My main goal was that people implementing handlers didn't have to build a bunch of boilerplate.

jamesstarr:
> 

@jacques-n, it is used in ReflectiveRelMetadataProvider to support deprecated behavior, so no refactoring or minimal refactoring is required.  You could just have it throw an exception since the value is never actually needed.

jacques-n:
> You could just have it throw an exception since the value is never actually needed.

That would be a breaking change (downstream projects may be using `getDef()`). We should deprecate for at least one release before removing and/or throwing. This means for at least one release (possibly many), the PR alleviates downstream boilerplate implementations and allows most handlers to be functional. It also allows implementers to be less impacted once the method deprecates and then is removed.

While getDef() deprecation may make sense ultimately, I'd like to see any patch that proposes that to also modify `ReflectiveRelMetadataProvider` to avoid it's use in the non-deprecated path of `public static <M extends Metadata> RelMetadataProvider reflectiveSource(
      MetadataHandler<? extends M>, Class<? extends MetadataHandler<M>>)`. This seems beyond the scope of this change. Happy to review that patch separately if you'd like to propose it.

jamesstarr:
All of calcites instances would still return value, and previously all of
their instances would have had to have an implementation.  You can not
break down stream code by adding a default implementation, after wall it
was one of the design goals of java 8.

jacques-n:
@jamesstarr, you're right that technically it wouldn't be a breaking change. However, it doesn't seem like a good approach since it is a misleading change. Having getDef() throw and marking the subinterfaces as functional means you have a suggested interface that doesn't actually work in the reflective provider. The change here avoids causing that unexpected/misleading behavior. As I said, if you want to get rid of the method moving forward, I'm fully supportive of doing that.

PR #2641, opened 2021-12-11
[CALCITE-4934] Deploy snapshots releases on master merge

jacques-n:


zabetak (approved):
LGTM

build.gradle.kts
```diff
@@ -806,6 +806,18 @@ allprojects {
...
+            repositories {
+                maven {
+                    name = "ApacheSnapshot" //  optional target repository name
+                    url = uri("https://repository.apache.org/content/repositories/snapshots")
+                    credentials {
+                        username = (project.properties["nexusSnapshotUser"] ?: "").toString()
+                        password = (project.properties["nexusSnapshotPassword"] ?: "").toString()
+                    }
+                }
+            }
```
vlsi, line 819: The repository is already configured, so there's no need to add one more

vlsi (changes requested):
I think https://ci-builds.apache.org/job/Calcite/job/Calcite-snapshots/ is a better way to publish snapshots

jacques-n:
> I think https://ci-builds.apache.org/job/Calcite/job/Calcite-snapshots/ is a better way to publish snapshots

Why? We use Github Workflows for everything else, why wouldn't we use it for this too? I for one don't regularly look at a random Jenkins instance. 

I'd also note that the job linked above was created after this PR. It'd be better to discuss and come to agreement on a route before building a "competing" solution and pushing it to production. I could have easily just pushed this PR to production as well.

build.gradle.kts
vlsi: https://calcite.apache.org/docs/howto.html#making-a-snapshot

jacques-n: Can you please answer my questions: 

Where is this defined in the build files? What are the username and password property names?

jacques-n:
I've modified this to try to use the existing -Pasf definition (wherever it is defined)

jacques-n:
I've confirmed that the updated version of this is working successfully:

https://github.com/apache/calcite/runs/4509981678?check_suite_focus=true

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,35 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+name: "Publish Snapshot to Maven"
```
vlsi, line 20:
```suggestion
name: "Publish Snapshot to Nesus"
```

build.gradle.kts
vlsi: @jacques-n , I understand it might be frustrating, however, could you please try finding the answer at howto.html?

Calcite releases are pushed to ASF Nexus for a long time, so the repository must be declared.

Of course, I can answer your questions, however, in that case, I would like to remove howto.html since it seems to add no value.

jacques-n: Closing conversation due to disrespectful behavior

jacques-n:
Abandoning due to another version being done by another committer.

PR #2643, opened 2021-12-13
Remove duplicate entries from RelOptRules.CALC_RULES

zabetak:
It makes no difference for the planner since duplicate rules are ignored.


NobiGo (approved):
LGTM

zabetak:
Thanks for the review @NobiGo !

PR #2644, opened 2021-12-13
Improve debug message in IterativeRuleDriver

zabetak:
Remove PLANNER logging from the message and better formulate at which point during the rule matching the cost is printed.

The original message is misleading cause it is not the planner that is printed but the queue. Moreover, `IterativeRuleDriver#toString` just prints the class name and hashcode of the instance which is useless. 
Class name might be useful but can be obtained by changing the logger configuration (pattern layout).

### Before
```
PLANNER = org.apache.calcite.plan.volcano.IterativeRuleDriver@3f3065ea; COST = {inf}
...
PLANNER = org.apache.calcite.plan.volcano.IterativeRuleDriver@3f3065ea; COST = {11.0 rows, 42.0 cpu, 0.0 io}
```

### After
```
Best cost before rule match: {inf}
...
Best cost before rule match: {11.0 rows, 42.0 cpu, 0.0 io}
```

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/volcano/IterativeRuleDriver.java
+++ b/core/src/main/java/org/apache/calcite/plan/volcano/IterativeRuleDriver.java
@@ -21,8 +21,6 @@ import org.apache.calcite.util.trace.CalciteTrace;
 
 import org.slf4j.Logger;
 
-import static java.util.Objects.requireNonNull;
-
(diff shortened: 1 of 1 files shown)


NobiGo (approved):
LGTM

zabetak:
Thanks for the review @NobiGo !

PR #2645, opened 2021-12-14
[CALCITE-4939] Upgrade log4j2 version to 2.16.0

zabetak:
The 2.16.0 release has two additional fixes related to CVE-2021-4422 for hardening security. Calcite uses Log4j2 only for testing purposes so the risk is minimal but it would be good to have them.


guiyanakuang:
LGTM

asolimando (approved):
LGTM

rubenada:
LGTM

PR #2646, opened 2021-12-14
[CALCITE-4936] Generalize FilterCalcMergeRule and ProjectCalcMergeRule

mgramin:
[CALCITE-4936] Generalize FilterCalcMergeRule and ProjectCalcMergeRule. Like CalcMergeRule, use only Project, Filter and Calc interfaces, which will expand the range of their applying.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/ProjectCalcMergeRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/ProjectCalcMergeRule.java
@@ -70,9 +70,7 @@ public class ProjectCalcMergeRule
 
     // Don't merge a project which contains windowed aggregates onto a
     // calc. That would effectively be pushing a windowed aggregate down
-    // through a filter. Transform the project into an identical calc,
-    // which we'll have chance to merge later, after the over is
-    // expanded.
+    // through a filter.
     final RelOptCluster cluster = project.getCluster();
     RexProgram program =
         RexProgram.create(
--- a/core/src/main/java/org/apache/calcite/rel/rules/ProjectCalcMergeRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/ProjectCalcMergeRule.java
@@ -35,13 +35,13 @@ import org.immutables.value.Value;
 
 /**
  * Planner rule that merges a
- * {@link org.apache.calcite.rel.logical.LogicalProject} and a
- * {@link org.apache.calcite.rel.logical.LogicalCalc}.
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/rel/rules/ProjectCalcMergeRule.java
```diff
@@ -82,8 +82,6 @@ public ProjectCalcMergeRule(RelBuilderFactory relBuilderFactory) {
             project.getRowType(),
             cluster.getRexBuilder());
     if (RexOver.containsOver(program)) {
-      LogicalCalc projectAsCalc = LogicalCalc.create(calc, program);
```
rubenada, line 85: Isn't this a breaking change? Currently the Project is removed, with this proposal the Project remains.

mgramin: Unfortunately, yes. But i assume that special rule `ProjectToCalcRule` can perform simple conversion from `LogicalProject` to `LogicalCalc`. Whereas `ProjectCalcMergeRule` will be able to stay more clear and more generalize.

zabetak: This is a small breaking change. I doubt people will notice but let's put a small note in `history.md` to be on the safe side. Moreover, please remove the now obsolete comments in lines 73-75 above.

zabetak (approved):
LGTM, only minor comments left.

General comment independent of this change: Do we really need `FilterCalcMergeRule` and `ProjectCalcMergeRule`? They look a bit redundant in the presence of `FilterToCalcRule` `ProjectToCalcRule` and `CalcMergeRule`.

mgramin:
> General comment independent of this change: Do we really need `FilterCalcMergeRule` and `ProjectCalcMergeRule`? They look a bit redundant in the presence of `FilterToCalcRule` `ProjectToCalcRule` and `CalcMergeRule`.

`FilterCalcMergeRule` and `ProjectCalcMergeRule` produce only `LogicalCalc` also, it can make further optimizations  more difficult.

hsyuan (approved):
LGTM. Shall we merge it?

rubenada:
@hsyuan ok for me, but please be aware that we are currently in code freeze since we are in the middle of building a new release. We could merge this patch once master re-opens.

zabetak:
I will merge this in a few minutes.

PR #2647, opened 2021-12-14
[CALCITE-4941] SemiJoinRule loses hints

rubenada:
[[CALCITE-4941](https://issues.apache.org/jira/browse/CALCITE-4941)] SemiJoinRule loses hints
SemiJoinRule does not propagate the (potential) hints from the input Join to the output SemiJoin

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -58,6 +58,9 @@ import org.apache.calcite.rel.core.JoinRelType;
 import org.apache.calcite.rel.core.Minus;
 import org.apache.calcite.rel.core.Project;
 import org.apache.calcite.rel.core.Union;
+import org.apache.calcite.rel.hint.HintPredicates;
+import org.apache.calcite.rel.hint.HintStrategyTable;
+import org.apache.calcite.rel.hint.RelHint;
 import org.apache.calcite.rel.logical.LogicalAggregate;
 import org.apache.calcite.rel.logical.LogicalCorrelate;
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -1077,6 +1081,48 @@ private void checkSemiOrAntiJoinProjectTranspose(JoinRelType type) {
...
+
+    // verify hint
+    final RelBuilder relBuilder = RelBuilder.create(RelBuilderTest.config().build());
+    final RelNode input = relFn.apply(relBuilder);
+    final HepProgram program = new HepProgramBuilder()
+        .addRuleInstance(CoreRules.PROJECT_TO_SEMI_JOIN)
+        .build();
+    final HepPlanner hepPlanner = new HepPlanner(program);
+    hepPlanner.setRoot(input);
+    final RelNode output = hepPlanner.findBestExp();
```
amaliujia, line 1121: I didn't touch this before so out of curiosity: 

Is HepPlanner deterministic  that we can guarantee this test will not be flaky?

amaliujia:
LGTM

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
rubenada: Yes, it is deterministic. HepPlanner applies its rule(s) systematically until no rule can be fired.
In this particular test it will simply apply {{CoreRules.PROJECT_TO_SEMI_JOIN}}.

NobiGo (approved):
LGTM

PR #2648, opened 2021-12-17
[CALCITE-4947] Checkstyle fails on classes generated by Intellij when using option "build and run [tests] using Intellij IDEA"

thomasrebele:


adiroy05:
assign to me

vlsi:
@thomasrebele , did you check the files are not duplicated across `generated` and `generated_test` folders?
In other words, is it right that each `.java` class appears in a single folder only?

thomasrebele:
@vlsi, I could not find any duplicates. What would happen if there were duplicates?

vlsi:
The question was not really related to the change. I think the change is mergeable.

PR #2649, opened 2021-12-17
[CALCITE-4948] Revert Elasticsearch to 7.10.2

jacques-n:
Remove extra maven repository.

### DIFF
--- a/elasticsearch/build.gradle.kts
+++ b/elasticsearch/build.gradle.kts
@@ -18,12 +18,6 @@ plugins {
     id("com.github.vlsi.gradle-extensions")
 }
 
-repositories {
-    maven {
-        setUrl("https://maven.codelibs.org/")
-    }
-}
-
 dependencies {
     api(project(":core"))
     api(project(":linq4j"))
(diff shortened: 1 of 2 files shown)


elasticsearch/build.gradle.kts
```diff
@@ -44,7 +38,6 @@ dependencies {
     testImplementation("org.apache.logging.log4j:log4j-api")
     testImplementation("org.apache.logging.log4j:log4j-core")
     testImplementation("org.codelibs.elasticsearch.module:lang-painless")
-    testImplementation("org.codelibs.elasticsearch.module:scripting-painless-spi")
```
vlsi, line 47: Is `org.codelibs` an authentic source of the dependencies? C
It looks like they publish non AL2 code as if it was AL2.

Should we use `org.elasticsearch.plugin:elasticsearch-scripting-painless-spi` instead?

jacques-n: I've opened a [separate ticket|https://issues.apache.org/jira/browse/CALCITE-4948] to evaluate since we've had the `org.codelibs.elasticsearch.module:lang-painless dependency` for 2+years. For reference, it looks like [this comment](https://github.com/apache/calcite/blob/94e6272d57478a5913cb51e7df5c25e67511229a/elasticsearch/pom.xml#L113) was lost when we migrated from maven to gradle.

PR #2650, opened 2021-12-18
[CALCITE-4949] Evaluate if using org.codelibs is appropriate for an Elasticsearch painless dependency

ILuffZhe:
This PR adds some old comments and removes "scripting-painless-spi" dependency, which is included in "lang-painless".

### DIFF
--- a/elasticsearch/build.gradle.kts
+++ b/elasticsearch/build.gradle.kts
@@ -37,7 +37,11 @@ dependencies {
 
     testImplementation("org.apache.logging.log4j:log4j-api")
     testImplementation("org.apache.logging.log4j:log4j-core")
-    testImplementation("org.codelibs.elasticsearch.module:lang-painless")
+    // See https://github.com/elastic/elasticsearch/issues/18131.
+    testImplementation("org.codelibs.elasticsearch.module:lang-painless") {
+        because("Elasticsearch doesn't export painless script artifact to maven central, " +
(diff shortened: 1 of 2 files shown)


elasticsearch/build.gradle.kts
```diff
@@ -37,6 +37,9 @@ dependencies {
 
     testImplementation("org.apache.logging.log4j:log4j-api")
     testImplementation("org.apache.logging.log4j:log4j-core")
+    // Elasticsearch doesn't export painless script artifact to maven central.
+    // Using 3rd party version (codelibs) only for test.
```
vlsi, line 41: Please use `because("...")` instead of comments

ILuffZhe: @vlsi  I've made some changes if I understand you correctly.

NobiGo: @ILuffZhe  You have some misunderstanding about the @vlsi comment. You can find an example in [https://github.com/gradle/gradle-profiler/blob/master/build.gradle.kts#L33](https://github.com/gradle/gradle-profiler/blob/master/build.gradle.kts#L33)

ILuffZhe: Thanks for your reminder, @NobiGo. I've corrected it.

NobiGo (approved):
LGTM

PR #2652, opened 2021-12-18
Set timeout for running Druid tests in GitHub CI

zabetak:
When tests run successfully they don't take more than a few minutes.
Sometimes though they get stuck and run forever (till the max GitHub
timeout of 6h) so set a timeout to 10minutes to avoid wasting resources
and fail-fast.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -287,6 +287,7 @@ jobs:
         path: calcite
     - uses: burrunan/gradle-cache-action@v1
       name: 'Run Druid tests'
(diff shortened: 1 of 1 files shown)


asolimando (approved):
10 minutes sounds reasonable and having a timeout is a good workaround while waiting to fix the root the cause.

rubenada:
Sounds reasonable. LGTM

NobiGo (approved):
LGTM

zabetak:
Thanks for the reviews @asolimando @rubenada @NobiGo !

PR #2653, opened 2021-12-19
[CALCITE-4952] (Re)introduce a simplistic RelMetadataQuery option

jacques-n:
- Introduce new ProxyingMetadataHandlerProvider that avoids code compilation
- Update RelMetadataTest to factor out metadata handling from tests
- Introduce tests for both Janino and the proxying/simplistic path.
- A simple microbenchmark for metadata
- Minor version bump to JMH
- Enable clearing of static cache within JaninoRelMetadataProvider


core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -3686,4 +3742,43 @@ Sql assertRowsUnique(boolean[] ignoreNulls, Matcher<Boolean> matcher,
     }
 
   }
+
+  /**
+   * Test class for janino metadata path.
+   */
+  @Nested
```
zabetak, line 3749: I am not sure if this is an appropriate use of the `@Nested` annotation. Citing the [JUnit5 user guide](https://junit.org/junit5/docs/current/user-guide/#writing-tests-nested):
`Only non-static nested classes (i.e. inner classes) can serve as @Nested test classes`.

I think the test run in this case cause you are using also inheritance but again I am not sure if those were meant to be combined. Instead of relying on test inheritance or `@Nested`, [parameterized tests](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests) would better fit here I think.

jacques-n: Yeah, I was torn on making it parameterized given the sheer amount of minor changes. 

I just went through and did so. It quadrupled the size of the patch but I agree it is more idiomatic.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -3686,4 +4157,71 @@ Sql assertRowsUnique(boolean[] ignoreNulls, Matcher<Boolean> matcher,
...
+
+  /**
+   * A configuration that describes how metadata should be configured.
+   */
+  public static class MetadataConfig {
+
+    private final String name;
+    private final Function<RelMetadataProvider, MetadataHandlerProvider> converter;
+    private final Supplier<RelMetadataProvider> defaultProviderSupplier;
+    private final boolean isCaching;
```
jamesstarr: Caching is required volcano planner.

jacques-n: Can you point me to where that is? That seems like it should be a bug in volcano, not a requirement.

jacques-n: I'm wondering if you're referring of the requirement to ensure we're breaking cycles (as opposed to actually caching). Do you know if there are any tests to verify this behavior (cycle-break)?

jamesstarr: Yes, I am refering to cycle detection.  Sorry, about not being clearer.  I am not sure if their are any tests in calcite that trigger it.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
zabetak: Oh, sorry about that. It wasn't a "must do" comment from my side but thanks for taking care of it.

core/src/main/java/org/apache/calcite/rel/metadata/ProxyingMetadataHandlerProvider.java
```diff
@@ -0,0 +1,114 @@
...
+          () -> "Not supported: " + method);
+      RelNode rel = Objects.requireNonNull((RelNode) args[0], "rel must be non null");
+      RelMetadataQuery mq = Objects.requireNonNull((RelMetadataQuery) args[1],
+          "mq must be non null");
+
+      // using deprecated RelMetadataProvider method here as the non-deprecated methods completely
+      // sidestep the purpose of RelMetadataProvider reflection-based functionality.
+      UnboundMetadata metadata = provider.apply(
+          (Class<? extends RelNode>) rel.getClass(),
+          (Class<? extends Metadata>) metadataType);
```
zabetak, line 88: I don't think it is a good idea to use deprecated APIs. If we really need this method one way or another should we revert the deprecation?

ubenchmark/src/jmh/java/org/apache/calcite/benchmarks/MetadataBenchmark.java
```diff
@@ -0,0 +1,140 @@
...
+            + "          EnumerableTableScan(table=[[foodmart2, time_by_day]])\n"
+            + "    EnumerableHashJoin(condition=[=($0, $2)], joinType=[inner])\n"
+            + "      EnumerableCalc(expr#0..14=[{inputs}], proj#0..1=[{exprs}])\n"
+            + "        EnumerableTableScan(table=[[foodmart2, product]])\n"
+            + "      EnumerableCalc(expr#0..4=[{inputs}], product_class_id=[$t0], "
+            + "product_family=[$t4])\n"
+            + "        EnumerableTableScan(table=[[foodmart2, product_class]])")
+        .returns("c0=USA; c1=1997; c2=Non-Consumable; m0=16414\n"
+            + "c0=USA; c1=1997; c2=Drink; m0=7978\n"
+            + "c0=USA; c1=1997; c2=Food; m0=62445\n");
```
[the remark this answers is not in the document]
jacques-n: I agree in theory. In reality, I'm looking for baby steps here. 

Based on my reviewing of metadata tickets back to the original refactor Julian did several years ago, it appears that multiple developers of the metadata module would manually use this query to test the performance. I wanted to at least get it in a place where it could be run in a reliable/standard/consistent way. Realistically, we should move to something that is actually a microbenchmark. Having the entirety of query planning rule firing/transformations/etc being part of a test to understand metadata performance leads to a lot of noise in the measurement (beyond just the plan validation). 

For now, I thought it was better to keep this in since it also serves to confirm that the providers are behaving consistently.

core/src/main/java/org/apache/calcite/rel/metadata/ProxyingMetadataHandlerProvider.java
jacques-n: This isn't my favorite either but I think it's the right choice for now. The apis are so much in flux at the moment that reverting deprecation isn't ideal since I'm not convinced that the deprecated interface is best either. (In a perfect world, we would have included the second impl before deciding to deprecate this method.) A risk associated with the ongoing refactor in this module is too much focus on a single impl (the janino one). It's important to have this second impl in place so we can make sure the refactoring makes sense across impls.

zabetak: I didn't follow closely the changes in metadata APIs but personally I would refrain from deprecating/removing APIs when in doubt. Taking out a release with a new deprecation has an impact on end-users and we cannot change our mind very often. Having said that, this is not a discussion for this PR so feel free to move it forward as you see fit.

ubenchmark/src/jmh/java/org/apache/calcite/benchmarks/MetadataBenchmark.java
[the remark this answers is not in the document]
zabetak: I am not saying to turn this into a real microbenchmark since we are quite far from there anyways. However, if we need to ensure correctness then wouldn't be better to add unit tests for this (if there are aren't already). Anyways minor comment feel free to ignore.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
jacques-n: I've just added a test to verify no cycles in both impls.

PR #2655, opened 2021-12-22
[CALCITE-4942] Remove metadata boilerplate

jamesstarr:
MetadataHandler's method getDef() and the generic referencing a Metadata
subtype provide no value in current implementation of Metadata.  Removing
them reduces the amount of code to define a Metadata type as well as
clearify the api.  This will require downstream projects that have custom
metadata type to remove the generic from their Handler class.

* Deprecating MetadataDef, BuiltInMetadata.{extends Metadata},
  BuiltInMetadata.{extends Metadata}.Handler, internal methods of
  ChainedRelMetadataProvider, BuiltInMethod referencing Metadata
  methods.
* Reworking BuiltInMetadata to have the Handlers at the top level.
  Extending the new Handlers with the old hanlders to reduces breaking
  downstream projects.
* Adding default implementation to MetadataHandler#getDef
* Removing generic from RelMetadataHandler
* Reworking creation validation in ReflectiveRelMetadataProvider

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/hep/HepRelMetadataProvider.java
+++ b/core/src/main/java/org/apache/calcite/plan/hep/HepRelMetadataProvider.java
@@ -18,7 +18,6 @@ package org.apache.calcite.plan.hep;
 
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.metadata.Metadata;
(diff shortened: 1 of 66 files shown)


vlsi:
I was not involved in Metadata development a lot, nor did I use it a lot, however, I think `MetadataDef` was not that useful, so +1 for removing the code and making the code easier to follow.

PR #2657, opened 2021-12-22
[CALCITE-4860] In Elasticsearch adapter, support NULLS FIRST and NULLS LAST query

ILuffZhe:
Support NULLS LAST/FIRST SQL semantics by modifying Elasticsearch scripts.

### DIFF
--- a/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchRel.java
+++ b/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchRel.java
@@ -57,6 +57,13 @@ public interface ElasticsearchRel extends RelNode {
     final List<Map.Entry<String, RelFieldCollation.Direction>> sort = new ArrayList<>();
 
     /**
+     * Sorting missing values.
+     *
+     * @see <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/sort-search-results.html#_missing_values">Missing Values</a>
+     */
+    final List<Map.Entry<String, RelFieldCollation.NullDirection>> nullsSort = new ArrayList<>();
+
+    /**
      * Elastic aggregation ({@code MIN / MAX / COUNT} etc.) statements (functions).
      *
      * @see <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html">aggregations</a>
@@ -108,6 +115,11 @@ public interface ElasticsearchRel extends RelNode {
     void addSort(String field, RelFieldCollation.Direction direction) {
       requireNonNull(field, "field");
       sort.add(Pair.of(field, direction));
+    }
+
+    void addNullsSort(String field, RelFieldCollation.NullDirection nullDirection) {
+      requireNonNull(field, "field");
+      nullsSort.add(new Pair<>(field, nullDirection));
     }
 
     void addAggregation(String field, String expression) {
--- a/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSort.java
+++ b/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSort.java
@@ -68,6 +68,8 @@ public class ElasticsearchSort extends Sort implements ElasticsearchRel {
     for (RelFieldCollation fieldCollation : collation.getFieldCollations()) {
       final String name = fields.get(fieldCollation.getFieldIndex()).getName();
       final String rawName = implementor.expressionItemMap.getOrDefault(name, name);
+      // if nulls order is not specified, default NULLS LAST/FIRST for ASC/DESC
+      implementor.addNullsSort(rawName, fieldCollation.nullDirection);
       implementor.addSort(rawName, fieldCollation.getDirection());
     }
 
--- a/elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/AggregationAndSortTest.java
+++ b/elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/AggregationAndSortTest.java
@@ -48,7 +48,7 @@ import java.util.Map;
  * Testing Elasticsearch aggregation transformations.
  */
 @ResourceLock(value = "elasticsearch-scrolls", mode = ResourceAccessMode.READ)
-class AggregationTest {
+class AggregationAndSortTest {
 
   public static final EmbeddedElasticsearchPolicy NODE = EmbeddedElasticsearchPolicy.create();
 
@@ -127,24 +127,24 @@ class AggregationTest {
   @Test void searchInRange() {
     Assumptions.assumeTrue(Bug.CALCITE_4645_FIXED, "CALCITE-4645");
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select count(*) from view where val1 >= 10 and val1 <=20")
         .returns("EXPR$0=1\n");
 
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select count(*) from view where val1 <= 10 or val1 >=20")
         .returns("EXPR$0=2\n");
 
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select count(*) from view where val1 <= 10 or (val1 > 15 and val1 <= 20)")
         .returns("EXPR$0=2\n");
   }
 
   @Test void countStar() {
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select count(*) from view")
         .queryContains(
             ElasticsearchChecker.elasticsearchChecker(
@@ -152,34 +152,34 @@ class AggregationTest {
         .returns("EXPR$0=3\n");
 
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select count(*) from view where cat1 = 'a'")
         .returns("EXPR$0=1\n");
 
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select count(*) from view where cat1 in ('a', 'b')")
         .returns("EXPR$0=2\n");
 
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select count(*) from view where val1 in (10, 20)")
         .returns("EXPR$0=0\n");
 
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select count(*) from view where cat4 in ('2018-01-01', '2019-12-12')")
         .returns("EXPR$0=2\n");
 
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select count(*) from view where cat4 not in ('2018-01-01', '2019-12-12')")
         .returns("EXPR$0=1\n");
   }
 
   @Test void all() {
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select count(*), sum(val1), sum(val2) from view")
         .queryContains(
             ElasticsearchChecker.elasticsearchChecker(
@@ -190,7 +190,7 @@ class AggregationTest {
         .returns("EXPR$0=3; EXPR$1=8.0; EXPR$2=47.0\n");
 
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select min(val1), max(val2), count(*) from view")
         .queryContains(
             ElasticsearchChecker.elasticsearchChecker(
@@ -203,14 +203,14 @@ class AggregationTest {
 
   @Test void cat1() {
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select cat1, sum(val1), sum(val2) from view group by cat1")
         .returnsUnordered("cat1=null; EXPR$1=0.0; EXPR$2=5.0",
                         "cat1=a; EXPR$1=1.0; EXPR$2=0.0",
                         "cat1=b; EXPR$1=7.0; EXPR$2=42.0");
 
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select cat1, count(*) from view group by cat1")
         .returnsUnordered("cat1=null; EXPR$1=1",
             "cat1=a; EXPR$1=1",
@@ -218,14 +218,14 @@ class AggregationTest {
 
     // different order for agg functions
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select count(*), cat1 from view group by cat1")
         .returnsUnordered("EXPR$0=1; cat1=a",
             "EXPR$0=1; cat1=b",
             "EXPR$0=1; cat1=null");
 
     CalciteAssert.that()
-        .with(AggregationTest::createConnection)
+        .with(AggregationAndSortTest::createConnection)
         .query("select cat1, count(*), sum(val1), sum(val2) from view group by cat1")
         .returnsUnordered("cat1=a; EXPR$1=1; EXPR$2=1.0; EXPR$3=0.0",
                 "cat1=b; EXPR$1=1; EXPR$2=7.0; EXPR$3=42.0",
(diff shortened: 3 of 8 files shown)


elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSort.java
```diff
@@ -62,6 +62,8 @@
     for (RelFieldCollation fieldCollation : collation.getFieldCollations()) {
       final String name = fields.get(fieldCollation.getFieldIndex()).getName();
       final String rawName = implementor.expressionItemMap.getOrDefault(name, name);
+      // if nulls order is not specified, default NULLS LAST/FIRST for ASC/DESC
+      implementor.addNullsSort(rawName, fieldCollation.nullDirection);
```
NobiGo, line 66: The default behavior is stay same as ES ? Assording [_missing_values](https://www.elastic.co/guide/en/elasticsearch/reference/current/sort-search-results.html#_missing_values) The default behavior is always _LAST.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/AggregationTest.java
```diff
@@ -400,6 +400,21 @@ public static void setupInstance() throws Exception {
             "aggregations:{'v1.max.field': 'val1'",
             "'v2.min.field': 'val2'}"))
         .returnsUnordered("v1=7; v2=5");
+  }
+
+  @Test void testNullsSort() {
```
NobiGo, line 405: Can we remove this tests out the AggregationTest ? This unit test isn't belong to the type.

NobiGo (changes requested):

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSort.java
ILuffZhe: No, it's not the same.
I think it would be better to follow SQL behavior of NULLS order(NULLS LAST/FIRST for ASC/DESC), since Elasticsearch Adapter converts SQL semantics to ES scripts.

NobiGo: Ok. I get it.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/AggregationTest.java
ILuffZhe: I've thought about adding NULLS ORDER tests in ElasticSearchAdapterTest, which is more reasonable.
However, many tests are disabled after CALCITE-4101(see https://github.com/apache/calcite/commit/e7aca6914c65bd812a9b177c438f687b77e090af), as a result of CI failure.
I think it's better to recover those test cases. What do you suggest? @NobiGo

zabetak: @ILuffZhe Re-enabling the tests that pass sounds a reasonable thing to do.

ILuffZhe: @zabetak OK, I am gonna to file a JIRA case to document it. Test case for this PR shall be adjusted after that.

NobiGo: +1 Enabling the tests.

ILuffZhe: Hi, @NobiGo , @zabetak , I need a little advice on this patch.
I'm gonna to add some unit tests for NULLS SORT, however, ElasticSearchAdapterTest use zips-mini.json as its data, which doesn't contain any null values for each column, so adding or deleting data would cause a lot code changes. I've thought about two ways:

- add test cases in ElasticSearchAdapterTest without changing zips-mini.json, only check the generated Elasticsearch Scripts 
- add a new class for NULLS SORT, kinda of redundant since there are several test classes in this module(if any new features cannot be testified based on the existing tests, always adding a class might be unacceptable)

What do you think?

zabetak: @ILuffZhe I don't have a strong preference.

There are already "feature" specific classes such as "ProjectionTest", "MatchTest", etc., so I wouldn't be surprised if there is a new class emphasizing on sorting and nulls.

On the other hand you could definitely add another dataset in `ElasticSearchAdapterTest` and add test cases in that class. You may draw some inspiration from `JdbcAdapterTest` where each test uses a specific `model`.

ILuffZhe: @zabetak Thanks for the helpful hints. I'll take a look afterwards.

NobiGo: I think we should add another like zips-mini-nullable.json file include NULL value in doc, then after when we need add another test for NULL value inlucde aggregate... can base on this doc. About Where the unit test should be, I don't have a strong preference.

ILuffZhe: Thanks for the advice @NobiGo .
I've added a new dataset(zips-mini-nullable.json) for NULLS SORT, and made some adjustments on managing different datasets in ElasticSearchAdapterTest. Please take a look when you are free.

elasticsearch/src/test/resources/zips-mini-nullable.json
```diff
@@ -0,0 +1,6 @@
+{"age" :  21, "name" : "testName1", "score" : 91, "weight" : 65.0}
```
NobiGo, line 1: This different between zips-mini-nullable and zips-min file should be the NULL value, Others please stay same.

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchRel.java
```diff
@@ -54,6 +55,12 @@
      */
     final List<Map.Entry<String, RelFieldCollation.Direction>> sort = new ArrayList<>();
 
+    /**
+     * Sorting missing values.
+     * @see <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/sort-search-results.html#_missing_values">Missing Values</a>
+     */
+    final Map<String, RelFieldCollation.NullDirection> nullsSort = new HashMap<>();
```
NobiGo, line 62: I suspect this use Map have some hidden problems when handle have duplicate columns. I suggest we use  `List<Map.Entry<String, RelFieldCollation.Direction>>` to handle this.

NobiGo (changes requested):
Have some problems need to confirm.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/AggregationTest.java
asereda-gs: Hi @ILuffZhe,

I agree with Stamatis that you should validate actual query similarly to how it is done in `ElasticSearchAdapterTest`.

Perhaps, you can create separate class `SortTest` (or `OrderTest`) to validate output as well as actual queries ? For examples look at `BooleanLogicTest` or `AggregationTest`.

ILuffZhe: Hi @asereda-gs ,
I've though about creating a new class for testing NULLS SORT before, since NULLS SORT is one small sorting feature, I decided to add a dataset to verify its function at last, like @zabetak mentioned above.
What's more, for those contributors who want to add new features or fixing bugs in ES module, I think creating different indexes in one class(like ElasticSearchAdapterTest) for unit tests is also a good choice.

ILuffZhe: Hi @asereda-gs ,
I've though about creating a new class for testing NULLS SORT before, since NULLS SORT is one small sorting feature, I decided to add a dataset to verify its function at last, like @zabetak mentioned above.
What's more, for those contributors who want to add new features or fixing bugs in ES module, I think creating different indexes in one class(like ElasticSearchAdapterTest) for unit tests is also a good choice.

elasticsearch/src/test/resources/zips-mini-nullable.json
ILuffZhe: OK, I've updated the dataset.

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchRel.java
ILuffZhe: While sorting duplicated columns, the latter ones will be treated as redundant.
Since sort and nulls sort are corresponding for one field, I've changed the Collection to store NULLS SORT DIRECTION. Thanks @NobiGo .

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/AggregationTest.java
asereda-gs: I'm not sure you need a very big dataset, just a smart one.

Personally I find it easier to add test documents (for ES) directly in java code see [AggregationTest](https://github.com/apache/calcite/blob/cdf9d700183d926780f4e5b57d6211a85a548506/elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/AggregationTest.java#L72-L74) and then run different queries against them.

ILuffZhe: For this case, one small dataset is enough. And personally speaking, adding a new test class is much more convenient. So, your advice is reasonable.
As for why I add a new dataset in ElasticSearchAdapterTest, I think creating an index seems a little better than creating a test class when the existing data is not satisfied for fixing bugs, which require not many tests.
We might need a third opinion here. What do you think? @zabetak @NobiGo

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/ElasticSearchAdapterTest.java
```diff
@@ -746,4 +759,40 @@ public static void setupInstance() throws Exception {
...
+                + "id=null; pop=45442\n"
+                + "id=null; pop=null\n")
+        .queryContains(
+            ElasticsearchChecker.elasticsearchChecker(
+                "_source:[ 'id', 'pop' ]",
+                "sort:[ {pop:{missing:'_last', order:'desc'}} ]",
+                "size:5196"));
+
+    calciteAssertZipsWithNull()
+        .query("select pop, count(1) as CNT from zipswithnull group by pop "
```
NobiGo, line 778: Please use line breaking 
```
\n
```
 to reformat the query.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/ElasticSearchAdapterTest.java
```diff
@@ -746,4 +759,40 @@ public static void setupInstance() throws Exception {
...
+            + "order by pop asc nulls last")
+        .returns("pop=45442; CNT=1\n"
+            + "pop=47687; CNT=1\n"
+            + "pop=57871; CNT=1\n"
+            + "pop=59498; CNT=1\n"
+            + "pop=65046; CNT=1\n"
+            + "pop=null; CNT=1\n");
```
NobiGo, line 785: Please add elasticsearch script check in this unit test.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/AggregationTest.java
```diff
@@ -401,6 +401,34 @@ public static void setupInstance() throws Exception {
             "aggregations:{'v1.max.field': 'val1'",
             "'v2.min.field': 'val2'}"))
         .returnsUnordered("v1=7; v2=5");
+  }
+
+  @Test void testMultiNullsSortInGroup() {
+    CalciteAssert.that()
+        .with(newConnectionFactory())
```
NobiGo, line 408: According the query  and result, Actually I can't judge the result is wrong or right. So please make the test data more clearly for your test.

NobiGo (changes requested):
In addition, Please  use line break to reformat the SQL and Elasticsearch script.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/AggregationTest.java
NobiGo: As above, I don't care this test cases should be in ElasticSearchAdapterTest or another new Class, It's up to you. But what I care is the PR is really address this issue and have a more comprehensive test cases.

asereda-gs: I think there are too many unnecessary changes in `ElasticSearchAdapterTest`. You can reuse `AggregationTest` or create your own class (eg. `SortTest`). I assume you would want to test both aggregation and sorting in the same query.

ILuffZhe: Yeah, I want to confirm order by in aggregation terms still works.
Moreover, I'm going to create a SortTest to address this issue based on the overall discussions here. Thanks for your advice @asereda-gs .

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchMapping.java
```diff
@@ -102,19 +105,28 @@ String index() {
...
-    private static final Set<JsonNode> MISSING_VALUES =
+    private static final Set<JsonNode> MIN_MISSING_VALUES =
         Stream.of("string", // for ES2
             "text", "keyword",
             "date", "long", "integer", "double", "float")
-            .map(Datatype::missingValueForType)
+            .map(Datatype::minMissingValueForType)
+            .collect(Collectors.toSet());
+
+    private static final Set<JsonNode> MAX_MISSING_VALUES =
```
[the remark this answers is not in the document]
ILuffZhe: You are right, the former code only deals with the MIN value for missing field, which is not enough for NULLS SORT in agg terms. Such test will be added in SortTest too.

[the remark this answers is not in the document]
asereda-gs: What would be min/max value for a string type ?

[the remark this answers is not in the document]
ILuffZhe: Currently, "\_\_MISSING\_\_" is used for text/keyword type in agg terms. Just like [CALCITE-4896](https://issues.apache.org/jira/browse/CALCITE-4896) mentioned, some certain context cannot be dealt well, and I still don't find a better way. Do you have any good ideas on this?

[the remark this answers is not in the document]
asereda-gs: But how min/max missing string value will be sorted ? I assume it also needs two values for nulls last/first ?

Should it be empty string `""` (for min) and `zzzzzz`  (for max) ?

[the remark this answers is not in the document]
ILuffZhe: You are right. I'll add those two value for text/keyword type. By the way, it seems hard to define a specific value for MAX.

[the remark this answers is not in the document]
ILuffZhe: Hi @asereda-gs . I've updated the PR according to the discussions. For text/keyword type, I use "~"(for max) to replace missing value, which has greater ASCII value than "z".
If you have any other opinions, please let me know.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/AggregationTest.java
ILuffZhe: @NobiGo SortTest has been added and its results are more intuitive.

ILuffZhe:
Hi, @asereda-gs . Sorry to bother you again, as for String type, MAX_VALUE for missing doc seems impossible to determine(char[Integer.MAX_VALUE] is also unacceptable). Maybe we should add some java doc to illustrate this. What do you think?

ILuffZhe:
Almost forgot this PR.
Min/Max value for a string type seems unlikely to deal with, so can we just partially support Numeric and Date nulls sort in ES adapter? What you guys think? @NobiGo @asereda-gs @zabetak

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/AggregationAndSortTest.java
```diff
@@ -412,16 +412,62 @@ private static Connection createConnection() throws SQLException {
...
             "cat1=b; cat6=null",
             "cat1=null; cat6=text1");
   }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-4860">[CALCITE-4860]
+   * In Elasticsearch adapter, support NULLS FIRST and NULLS LAST query</a>.
+   */
+  @Test void testNullsSort() {
+    CalciteAssert.that()
```
ILuffZhe, line 433: I don't put this test to ElasticSearchAdapterTest is because we need some null values, which may cause many unnecessary changes since we need to modify the zips-mini.json.

NobiGo (approved):
LGTM

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchMapping.java
```diff
@@ -150,7 +162,37 @@ private Datatype(final String name) {
...
+      // this is unknown type
+      return null;
+    }
+
+    private static @Nullable JsonNode maxMissingValueForType(String name) {
+      switch (name) {
+      case "string": // for ES2
+      case "text":
+      case "keyword":
+        return FACTORY.textNode("~~~~~~~~");
```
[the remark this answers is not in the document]
ILuffZhe: > How to evaluate the length of the text field. If the value length is less than the "~~~~~~~" It maybe can get the right answer. But if the value length is longer than this?

I think this is an outdated comment.

PR #2658, opened 2021-12-23
[CALCITE-4963] Make behavior in SqlDialectFactoryImpl reusable

marcobjorge:
Make behaviour in `SqlDialectFactoryImpl` reusable by refactoring the creation of a `SqlDialect.Context` from a `DatabaseMetaData` into a `SqlContexts` class.


marcobjorge:
My understanding is that this was some kind of timeout error but I don't have permissions to retry - can someone please give it a push?

PR #2420, opened 2021-05-27
[CALCITE-4606]Translate SEARCH rexnode to termsQuery in ElasticSearch Adapter

yinchuanwang:
Currently the elastic search adapter does not support the 'SEARCH' call which usually comes from the 'in (a, b)'. This is the reason of the failure of test case "AggregationTest" in es adapter.  In this PR, I added the translation from 'SEARCH' call to termsQuery of es. Some test cases were also added in the "AggregationTest" to test the 'SEARCH' with different SQL types (char, varchar, number...)
Please help review.

PR #2659, opened 2021-12-24
[CALCITE-4960] Enable unit tests in Elasticsearch Adapter

ILuffZhe:
This PR aims at **re-enabling unit tests** in Elasticsearch Adapter. The changes are:

- Increasing SocketTimeout and ConnectionTimeout for RestClient to avoid CI failure

- Remove one case in BooleanLogicTest.java, which is documented at https://issues.apache.org/jira/browse/CALCITE-4965

- Disable testFilterSortDesc() in ElasticSearchAdapterTest.java, which is not supported before CALCITE-4645 merged.

I think re-enabling those tests has higher priorities than fixing bugs in ES Adapter. When other contributors make changes in this module, their codes cannot be guaranteed well without unit tests. Last but not least, the disabled cases mentioned above will be reopened after bug fixing.

### DIFF
--- a/elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/BooleanLogicTest.java
+++ b/elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/BooleanLogicTest.java
@@ -21,6 +21,7 @@ import org.apache.calcite.schema.SchemaPlus;
 import org.apache.calcite.schema.impl.ViewTable;
 import org.apache.calcite.schema.impl.ViewTableMacro;
 import org.apache.calcite.test.CalciteAssert;
+import org.apache.calcite.util.Bug;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.collect.ImmutableMap;
@@ -118,6 +119,10 @@ class BooleanLogicTest {
     assertEmpty("select * from view where num > 42 and num < 42 and num = 42");
     assertEmpty("select * from view where num > 42 or num < 42 and num = 42");
     assertSingle("select * from view where num > 42 and num < 42 or num = 42");
+    if (Bug.CALCITE_4965_FIXED) {
+      assertSingle("select * from view where num > 42 or num < 42 or num = 42");
+      assertEmpty("select * from view where num is null");
+    }
     assertSingle("select * from view where num >= 42 and num <= 42 and num = 42");
     assertEmpty("select * from view where num >= 42 and num <= 42 and num <> 42");
     assertEmpty("select * from view where num < 42");
--- a/elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/BooleanLogicTest.java
+++ b/elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/BooleanLogicTest.java
@@ -26,7 +26,6 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.collect.ImmutableMap;
 
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.parallel.ResourceAccessMode;
 import org.junit.jupiter.api.parallel.ResourceLock;
@@ -42,7 +41,6 @@ import java.util.Map;
 /**
  * Test of different boolean expressions (some more complex than others).
  */
-@Disabled("RestClient often timeout in PR CI")
 @ResourceLock(value = "elasticsearch-scrolls", mode = ResourceAccessMode.READ)
 class BooleanLogicTest {
 
@@ -120,7 +118,6 @@ class BooleanLogicTest {
     assertEmpty("select * from view where num > 42 and num < 42 and num = 42");
     assertEmpty("select * from view where num > 42 or num < 42 and num = 42");
     assertSingle("select * from view where num > 42 and num < 42 or num = 42");
-    assertSingle("select * from view where num > 42 or num < 42 or num = 42");
     assertSingle("select * from view where num >= 42 and num <= 42 and num = 42");
     assertEmpty("select * from view where num >= 42 and num <= 42 and num <> 42");
     assertEmpty("select * from view where num < 42");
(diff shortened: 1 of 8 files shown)


elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/EmbeddedElasticsearchPolicy.java
```diff
@@ -197,7 +197,11 @@ RestClient restClient() {
       return client;
     }
 
-    final RestClient client = RestClient.builder(httpHost()).build();
+    final RestClient client = RestClient.builder(httpHost())
+        .setRequestConfigCallback(requestConfigBuilder -> requestConfigBuilder
+            .setConnectTimeout(60 * 1000)  // default 1000
+            .setSocketTimeout(3 * 60 * 1000))  // default 30000
+        .build();
```
zabetak, line 204: How did you choose the new values? Did you do any tests?

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/ElasticSearchAdapterTest.java
```diff
@@ -476,8 +484,8 @@ public static void setupInstance() throws Exception {
 
   @Test void testInPlan() {
     final String[] searches = {
-        "query: {'constant_score':{filter:{bool:{should:"
```
zabetak, line 479: Why did `bool:should` disappeared?

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/ElasticSearchAdapterTest.java
```diff
@@ -462,6 +462,14 @@ public static void setupInstance() throws Exception {
...
   }
 
+  /**
+   * Range query is not supported currently.
+   * @see <a href="https://issues.apache.org/jira/browse/CALCITE-4645">[CALCITE-4645]
+   * In Elasticsearch adapter, a range predicate should be translated to a range query</a>
+   * <p>Description about translating search call to range query
+   * {@link org.apache.calcite.adapter.elasticsearch.PredicateAnalyzer.Visitor#canBeTranslatedToTermsQuery(RexCall)}
+   */
+  @Disabled
```
zabetak, line 472: I think it is better to create an entry in https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/util/Bug.java class and use that value to skip/disable the test.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/BooleanLogicTest.java
```diff
@@ -120,7 +118,6 @@ public static void setupInstance() throws Exception {
     assertEmpty("select * from view where num > 42 and num < 42 and num = 42");
     assertEmpty("select * from view where num > 42 or num < 42 and num = 42");
     assertSingle("select * from view where num > 42 and num < 42 or num = 42");
-    assertSingle("select * from view where num > 42 or num < 42 or num = 42");
```
zabetak, line 123: Instead of removing completely the test can you please skip the line using a (new) reference from https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/util/Bug.java.

zabetak (approved):
LGTM thanks for pushing this forward @ILuffZhe! Left some minor comments and will merge it afterwards.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/EmbeddedElasticsearchPolicy.java
ILuffZhe: (1) I’ve checked some CI tests, where I find tests running time for Elasticsearch:test **ranges from 18s to 60s**, it’s hard to determine a perfect timeout value for RestClient due to the network bandwidth and resources fluctuations. In our own project, we set those two timeout for 60s and 180s based on experience and actual use.
(2) As for timeout tests, I’ve debug this several times by setting socketTimeout to 1, and “1 milliseconds timeout on connection http-outgoing-0”(exception msg changed when you set different timeout) occurred when timeout reached. So, it's working.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/ElasticSearchAdapterTest.java
ILuffZhe: After CALCITE-4606, Search RexCall will be translated to Terms Query, you can see the full explanation here: https://github.com/apache/calcite/pull/2420/files#r647115126.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/ElasticSearchAdapterTest.java
ILuffZhe: Hi, @zabetak .I didn't know such an entry for Bug label before. 
I've added those two bugs there, thanks for your reminder!

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/BooleanLogicTest.java
ILuffZhe: Done.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/ElasticSearchAdapterTest.java
zabetak: OK, got it!

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/EmbeddedElasticsearchPolicy.java
zabetak: Thanks for the clarification.

ILuffZhe:
Thanks for your review, @zabetak !

PR #2660, opened 2021-12-25
[CALCITE-4966] Make CsvEnumerator.RowConverter public

ieugen:
CsvEnumerator.RowConverter is exposed in consturctor and public static method so it should be public.


NobiGo:
Please reformat the commit info like 
`[CALCITE-345] AssertionError in RexToLixTranslator comparing to date literal`
Then the PR can auto-link to JIRA

ieugen:
@NobiGo : As long as the jira ID is in the commit log, it will work.
It does work now.
I will change the commit id as you mentioned and force push the changes.

ieugen:
@NobiGo : Done.

PR #2661, opened 2021-12-26
[CALCITE-4957] Add HASH_RANDOM in RelDistribution.Type to solve data skew caused by Hash Distribution type

beyond1920:

The pr aims to introduce a new RelDistribution.Type (HASH_RANDOM) to solve data skew caused by Hash type.

## Brief change log

  - Add a value `HASH_RANDOM_DISTRIBUTED` in enum `RelDistribution#Type`
  - Add a method `getHashRandomNumber` in interface `RelDistribution`, which default return null.
  - Update `RelDistributions` to apply for the new enum value in `RelDistribution#Type`
  - Update `ExchangeRemoveConstantKeysRule` to take new enum value in `RelDistribution#Type` into consideration
  - Update `RelJson` to take the hashRandomNumber into consideration when convert to/From Json

## Verifying this change

  - Add tests in `RelWriterTest`, `RelDistributionTest`, `RelBuilderTest`, `RelOptRulesTest`

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -6308,6 +6308,30 @@ class RelOptRulesTest extends RelOptTestBase {
         .check();
   }
 
+  @Test void testHashRandomExchangeRemoveConstantKeysRule() {
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .scan("EMP")
+        .filter(
+            b.call(
+                SqlStdOperatorTable.EQUALS,
+                b.field("EMPNO"),
+                b.literal(10)))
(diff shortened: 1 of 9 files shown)


core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6297,7 +6297,7 @@ private Sql spatial(String sql) {
             b.field(0),
             b.field(1))
         .sortExchange(
-            RelDistributions.hash(ImmutableList.of(0, 1)),
+            RelDistributions.hashRandom(ImmutableList.of(0, 1), 8),
```
chunweilei, line 6300: Should we add a new test?

beyond1920: Make sense. I've updated the pr. Thanks for review.

PR #2664, opened 2021-12-29
[CALCITE-4967] Support SQL hints for temporal table join

beyond1920:

The pr aims to support SQL hints for temporal table join.

## Brief change log

  - Extends `Correlate` to be a `Hintable`
  - Add a hint predicate in `HintPredicates`

## Verifying this change

  - Add tests in `SqlHintsConverterTest` to validate the hint could be propagated to `Correlate`
  - Add tests in `RelBuilderTest`

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/core/RelFactories.java
+++ b/core/src/main/java/org/apache/calcite/rel/core/RelFactories.java
@@ -406,14 +406,6 @@ public class RelFactories {
     RelNode createCorrelate(RelNode left, RelNode right, List<RelHint> hints,
         CorrelationId correlationId, ImmutableBitSet requiredColumns,
         JoinRelType joinType);
-
-    @Deprecated // to be removed before 1.23
-    default RelNode createCorrelate(RelNode left, RelNode right,
-        CorrelationId correlationId, ImmutableBitSet requiredColumns,
-        JoinRelType joinType) {
-      return createCorrelate(left, right, ImmutableList.of(), correlationId,
-          requiredColumns, joinType);
-    }
   }
 
   /**
--- a/core/src/main/java/org/apache/calcite/rel/core/RelFactories.java
+++ b/core/src/main/java/org/apache/calcite/rel/core/RelFactories.java
@@ -392,18 +392,28 @@ public class RelFactories {
    * <p>The result is typically a {@link Correlate}.
    */
   public interface CorrelateFactory {
+
     /**
      * Creates a correlate.
      *
      * @param left             Left input
      * @param right            Right input
+     * @param hints            Hints
      * @param correlationId    Variable name for the row of left input
      * @param requiredColumns  Required columns
      * @param joinType         Join type
      */
-    RelNode createCorrelate(RelNode left, RelNode right,
+    RelNode createCorrelate(RelNode left, RelNode right, List<RelHint> hints,
         CorrelationId correlationId, ImmutableBitSet requiredColumns,
         JoinRelType joinType);
+
+    @Deprecated // to be removed before 1.23
+    default RelNode createCorrelate(RelNode left, RelNode right,
+        CorrelationId correlationId, ImmutableBitSet requiredColumns,
+        JoinRelType joinType) {
+      return createCorrelate(left, right, ImmutableList.of(), correlationId,
+          requiredColumns, joinType);
+    }
   }
 
   /**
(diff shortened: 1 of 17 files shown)


core/src/main/java/org/apache/calcite/rel/core/RelFactories.java
```diff
@@ -392,29 +392,39 @@ RelNode createJoin(RelNode left, RelNode right, List<RelHint> hints,
...
      * @param joinType         Join type
      */
-    RelNode createCorrelate(RelNode left, RelNode right,
+    RelNode createCorrelate(RelNode left, RelNode right, List<RelHint> hints,
         CorrelationId correlationId, ImmutableBitSet requiredColumns,
         JoinRelType joinType);
+
+    @Deprecated // to be removed before 1.23
+    default RelNode createCorrelate(RelNode left, RelNode right,
+        CorrelationId correlationId, ImmutableBitSet requiredColumns,
```
danny0405, line 412: No need to add this default impl, `createCorrelate` is kind of internal API.

core/src/main/java/org/apache/calcite/rel/logical/LogicalCorrelate.java
```diff
@@ -86,26 +107,38 @@ public LogicalCorrelate(RelInput input) {
...
       JoinRelType joinType) {
     final RelOptCluster cluster = left.getCluster();
     final RelTraitSet traitSet = cluster.traitSetOf(Convention.NONE);
-    return new LogicalCorrelate(cluster, traitSet, left, right, correlationId,
+    return new LogicalCorrelate(cluster, traitSet, hints, left, right, correlationId,
         requiredColumns, joinType);
   }
 
+  @Deprecated // to be removed before 1.23
+  public static LogicalCorrelate create(RelNode left, RelNode right,
```
danny0405, line 120: Do you mean 'to be removed before 2.0' ?

danny0405 (commented):
Thanks for the contribution，overall looks good, only a few minor comments.

beyond1920:
@danny0405 Thanks for review. I've updated the pr based on you comments.

core/src/test/java/org/apache/calcite/test/SqlHintsConverterTest.java
```diff
@@ -771,30 +786,49 @@ static HintStrategyTable createHintStrategies(HintStrategyTable.Builder builder)
...
+
+      List<String> run(RelNode input) {
+        go(input);
+        return names;
+      }
+
+      @Override public void visit(RelNode node, int ordinal, @Nullable RelNode parent) {
+        if (node instanceof TableScan) {
+          RelOptTable table = node.getTable();
+          String tableName = Util.last(table.getQualifiedName());
```
danny0405, line 825: Should we use a visitor here ? The original logic force equals only when the join inputs are both direct table reference, but now it is not, which is not for our purpose ?

beyond1920: Hi danny, I updated the logical to extract table names. Please have a look at the pr again, thanks a lot.

core/src/test/java/org/apache/calcite/test/SqlHintsConverterTest.java
```diff
@@ -771,14 +788,42 @@ static HintStrategyTable createHintStrategies(HintStrategyTable.Builder builder)
...
+            && ((Snapshot) ((Filter) rightInput).getInput()).getInput() instanceof TableScan;
+        if (!rightIsFilterOnSnapshot) {
+          return false;
+        }
+        String leftTableName = getTableNameOfScan(leftInput);
+        String rightTableName = getTableNameOfScan(
+            ((Snapshot) ((Filter) rightInput).getInput()).getInput());
+        final List<String> tableNames = hint.listOptions;
+        return tableNames.contains(leftTableName) && tableNames.contains(rightTableName);
+      };
```
danny0405, line 824: Can we use `equalsStringList` here ?

beyond1920: > And there must have a filter ?
Yes for temporal join. Because there must a filter which contain join condition.

> Can we use equalsStringList here ?
I'm not sure if there is necessary here to build a list which contain leftTableName and rightTableName, then call equalsStringList. Because we have already get  leftTableName and rightTableName separately.

danny0405 (approved):
+1, looks good from my side.

PR #2666, opened 2021-12-30
[CALCITE-3627] New Row Null policy

talatuyarer:
As describe under CALCITE-3627 ticket. Current row policy is wrong. When ROW has a null column. Calcite returns null for ROW value. I am working on Apache Beam SQL Row support. Because of ANY policy RexToLixTranslator generate wrong code. This is my first pr for calcite project. Thanks for your comments in advance.

I recreated my pr because old pr did not puck up updated code. 

@DonnyZone could you re-review ?  I updated my pr after your RexNode-to-Expression CodeGen Implementation.

Thanks


talatuyarer:
I only changed 3 files but somehow it shows more than 7 files.

talatuyarer:
So sorry I have to close this pr too :(

PR #2667, opened 2021-12-30
[CALCITE-3627] New Row Null Policy

talatuyarer:
As describe under CALCITE-3627 ticket. Current row policy is wrong. When ROW has a null column. Calcite returns null for ROW value. I am working on Apache Beam SQL Row support. Because of ANY policy RexToLixTranslator generate wrong code. This is my first pr for calcite project. Thanks for your comments in advance. 

I recreated my pr because old pr did not puck up updated code.

@DonnyZone could you re-review ? I updated my pr after your RexNode-to-Expression CodeGen Implementation.

Thanks


talatuyarer:
Finally I created clean pr. @DonnyZone I look forward your comments. So sorry to create multiple PRs

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3107,7 +3107,13 @@ Expression getCondition(final List<Expression> argIsNullList) {
       if (nullPolicy == NullPolicy.ARG0) {
         return argIsNullList.get(0);
       }
```
DonnyZone, line 3109: It can be simplified as
```
if (nullPolicy == NullPolicy.ALL) {
    // Condition for NullPolicy.ALL: v0 == null && v1 == null
    return Expressions.foldAnd(argIsNullList);
}
// Condition for regular cases: v0 == null || v1 == null
return Expressions.foldOr(argIsNullList);
```

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3107,7 +3107,13 @@ Expression getCondition(final List<Expression> argIsNullList) {
...
-      return Expressions.foldOr(argIsNullList);
+      // Condition for regular cases: v0 == null || v1 == null
+      Expression condition = Expressions.foldOr(argIsNullList);
```
DonnyZone, line 3111: It can be simplified as
```
if (nullPolicy == NullPolicy.ALL) {
        // Condition for NullPolicy.ALL: v0 == null && v1 == null
        return Expressions.foldAnd(argIsNullList);
      }
// Condition for regular cases: v0 == null || v1 == null
 return Expressions.foldOr(argIsNullList);
```

DonnyZone:
Some tests in `PigRelOpTest` fail.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
talatuyarer: I changed this code block as you suggested. @DonnyZone

talatuyarer:
> Some tests in `PigRelOpTest` fail.

@DonnyZone previous Calcite does not support ROW nullability. so for those failing tests row is null but it returns full row rather than null value for row. I changed tests' expected output with right one. Please let me know if you see any issue

DonnyZone:
LGTM!
@talatuyarer Could you squash the commits and update the commit message as the format described in [guideline](https://calcite.apache.org/develop/)?

talatuyarer:
@DonnyZone I believe I squashed my commits. Please let me know if you need anything from me.

DonnyZone:
> @DonnyZone I believe I squashed my commits. Please let me know if you need anything from me.

Sorry for the inconvenience, could you squash your 5 commits into one, i.e., "`[CALCITE-3224] Incorrect null semantic for ROW function`"? So that I can merge this PR quickly through Github web UI.

talatuyarer:
Ok it is done. Now I see only one commit @DonnyZone

DonnyZone:
Hi @talatuyarer, sorry for trouble you again. The ticket number in commit message should be `[CALCITE-3627]`.

talatuyarer:
@DonnyZone it is done I did not realized too 😂

talatuyarer:
Thank you @DonnyZone for reviewing and merging 🙏

PR #2669, opened 2021-12-31
[CALCITE-4968] Use TOP N for MsSQL instead of FETCH without OFFSET

vvysotskyi:
MS SQL doesn't allow using `FETCH` without the `ORDER BY` statement. With this change, the `TOP N` statement will be generated instead of `FETCH` to prevent such errors.

Also, added `START AT` for Sybase, since initially OFFSET was ignored there.


cgivre:
Thanks @vvysotskyi for this!

jbalint:
looks good. thanks @vvysotskyi

PR #2672, opened 2022-01-03
CALCITE-4975 Removing dependency to commons-lang3

gunnarmorling:
Creating this as a draft PR for now. Provided there's interest in this, I'll log an issue in Jira and attribute the change to it. This is based on the recent mailing list discussion around reducing the number of dependencies of Calcite. I chose commons-lang3 as a rather easy target for getting my feet wet with the code base. Any feedback welcome 🙏.

### DIFF
--- a/file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
+++ b/file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
@@ -17,7 +17,6 @@
 package org.apache.calcite.adapter.file;
 
 import org.apache.calcite.adapter.java.JavaTypeFactory;
-import org.apache.calcite.avatica.util.DateTimeUtils;
 import org.apache.calcite.linq4j.Enumerator;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.type.SqlTypeName;
@@ -26,8 +25,6 @@ import org.apache.calcite.util.ImmutableNullableList;
 import org.apache.calcite.util.Pair;
 import org.apache.calcite.util.Source;
 import org.apache.calcite.util.trace.CalciteLogger;
-
-import org.apache.commons.lang3.time.FastDateFormat;
 
 import au.com.bytecode.opencsv.CSVReader;
 
@@ -39,13 +36,16 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
-import java.text.ParseException;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
-import java.util.TimeZone;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -54,6 +54,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import static org.apache.calcite.linq4j.Nullness.castNonNull;
 
+import static java.time.temporal.ChronoField.MILLI_OF_DAY;
 /** Enumerator that reads from a CSV file.
  *
  * @param <E> Row type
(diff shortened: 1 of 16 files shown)


file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
```diff
@@ -68,18 +69,16 @@
   private final RowConverter<E> rowConverter;
   private @Nullable E current;
 
-  private static final FastDateFormat TIME_FORMAT_DATE;
-  private static final FastDateFormat TIME_FORMAT_TIME;
-  private static final FastDateFormat TIME_FORMAT_TIMESTAMP;
+  private static final DateTimeFormatter TIME_FORMAT_DATE;
+  private static final DateTimeFormatter TIME_FORMAT_TIME;
+  private static final DateTimeFormatter TIME_FORMAT_TIMESTAMP;
```
gunnarmorling, line 74: So `DateTimeFormatter` was added in Java 8. Can we use this? I.e. what's the minimum Java baseline for Calcite?

gunnarmorling:
Changes fall in four categories:

* Replace `FastDateFormat` with Java 1.8's `DateTimeFormatter`
* Use built-in `String::isEmpty`
* Replace `Triple` class with dedicated POJO
* Use regexp for counting occurrences of a substring

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
```diff
@@ -68,18 +69,16 @@
...
 
   static {
-    final TimeZone gmt = TimeZone.getTimeZone("GMT");
-    TIME_FORMAT_DATE = FastDateFormat.getInstance("yyyy-MM-dd", gmt);
-    TIME_FORMAT_TIME = FastDateFormat.getInstance("HH:mm:ss", gmt);
-    TIME_FORMAT_TIMESTAMP =
-        FastDateFormat.getInstance("yyyy-MM-dd HH:mm:ss", gmt);
+    TIME_FORMAT_DATE = DateTimeFormatter.ofPattern("yyyy-MM-dd", Locale.ENGLISH);
+    TIME_FORMAT_TIME = DateTimeFormatter.ofPattern("HH:mm:ss", Locale.ENGLISH);
+    TIME_FORMAT_TIMESTAMP = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss", Locale.ENGLISH);
```
gunnarmorling, line 81:
```suggestion
    TIME_FORMAT_DATE = DateTimeFormatter.ofPattern("yyyy-MM-dd", Locale.ROOT);
    TIME_FORMAT_TIME = DateTimeFormatter.ofPattern("HH:mm:ss", Locale.ROOT);
    TIME_FORMAT_TIMESTAMP = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss", Locale.ROOT);
```

NobiGo:
@gunnarmorling Please reformat the commit info like this:
```
[CALCITE-345] AssertionError in RexToLixTranslator comparing to date literal
```

NobiGo:
I noticed this commons-lang3 also existed in other modules that doesn't remove. So It is missed or intentional?

gunnarmorling:
> I noticed this commons-lang3 also existed in other modules that doesn't remove. So It is missed or intentional?

Could you share a link to these references? Thx!

NobiGo:
> > I noticed this commons-lang3 also existed in other modules that doesn't remove. So It is missed or intentional?
> 
> Could you share a link to these references? Thx!

You can try search "org.apache.commons:commons-lang3" in project. For example:
https://github.com/apache/calcite/blob/master/geode/build.gradle.kts#L33

NobiGo:
@gunnarmorling  Sorry, It's my mistake. It is complete.

file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
NobiGo: We claimed support 8+ jdk in https://github.com/apache/calcite/blob/master/site/_docs/howto.md . So I think we can use it.

zabetak: It is a good idea to use `DateTimeFormatter` and we should do it. We should be a bit careful though cause every time there is a change in Date/Time/Timestamp parsing there are problems (many times backward compatibility) that go unnoticed.

Would it be possible to add some tests for dates/timestamps in the range `[0000-01-01 ,9999-12-29]` (a snippet generating random timestamps can be found [here](https://github.com/apache/hive/blob/3c9ea5d8d6ecc0d3749e22b28bfb7690dd3f1be1/ql/src/test/org/apache/hadoop/hive/ql/io/parquet/serde/TestParquetTimestampsHive2Compatibility.java#L118)) and check/document the differences with the old formatter. It's a bit unfortunate that we don't have such tests already.

zabetak (commented):
Thanks for taking care of this @gunnarmorling . Definitely useful change! I went quickly over the PR and left some suggestions.

PR #2673, opened 2022-01-04
[CALCITE-4910] Enhance simplify to reduce ((A OR D) and (A OR C) AND A AND B) to (A AND B) (Ziwei Liu)

xiayutianwei1994:
Enhance simplify to reduce ((A OR D) AND (A OR C) AND A AND B) to (A AND B) (Ziwei Liu)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexSimplify.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexSimplify.java
@@ -49,6 +49,7 @@ import com.google.common.collect.Sets;
 import com.google.common.collect.TreeRangeSet;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.codehaus.commons.nullanalysis.NotNull;
 
 import java.math.BigDecimal;
 import java.util.ArrayList;
@@ -629,7 +630,7 @@ public class RexSimplify {
     } else {
       e2 = rexBuilder.makeCall(e.op, operands);
     }
-    return simplifyUsingPredicates(e2, clazz);
+    return simplifyUsingPredicates(e2, clazz, unknownAs);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


NobiGo:
@xiayutianwei1994 Until we have a final state, please don't force push the commit before that.

xiayutianwei1994:
> @xiayutianwei1994 Until we have a final state, please don't force push the commit before that.

Ok, Thanks

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -1506,8 +1506,46 @@ RexNode simplifyAnd(RexCall e, RexUnknownAs unknownAs) {
...
+      if (satisfy.test(term)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
   // package-protected only to support a deprecated method; treat as private
   RexNode simplifyAnd2(List<RexNode> terms, List<RexNode> notTerms) {
+    Set<RexNode> termsSet = new HashSet<>(terms);
```
kgyrtkirk, line 1541: I think we already have similar logic here - based on the predicates - and most likely we only miss the handling of an existing predicate somewhere around in the `simplifyIsNull/simpliftIsNotNull` methods

I think this approach is a little bit too much by the way as at first glance I would be afraid that it would make "unsafe" simplifications as well

let me know if you think otherwise - I'll try to take a closer look tomorrow

xiayutianwei1994: Thanks for your comment.
If use unknown as unknown, some predicate will not be simplified to IS NOT NULL or IS NULL like the third test case I added.
And sometimes, predicate like (C0+C1) IS NOT NULL will not be pulled up.
There also has another question: If (A OR B) AND A, predicate A is nondeterministic, whether it can be simplified to A.
Now we only pull up predicate which is deterministic.

xiayutianwei1994: I have updated the code, please review it again.
Thanks very much.

xiayutianwei1994:
I have updated the code, please review it again.
Thanks very much.

PR #2678, opened 2022-01-06
Updated to clarify DB value in system property setting

mans2singh:
Update to clarify system property value **DB** which can be `h2`, `hsqldb` etc (updated <pre>-Dcalcite.test.db=DB (where <b>db</b> is...</pre> to <pre>-Dcalcite.test.db=DB (where <b>DB</b> is...</pre>).

### DIFF
--- a/site/_docs/howto.md
+++ b/site/_docs/howto.md
@@ -109,7 +109,7 @@ You can use `./gradlew assemble` to build the artifacts and skip all tests and v
 There are other options that control which tests are run, and in what
 environment, as follows.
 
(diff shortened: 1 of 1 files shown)


ILuffZhe:
@mans2singh That's a sharp catch. LGTM.

NobiGo:
Please change the Commit info as "Site: Fix typo in howto.md" Then we can merge it.

mans2singh:
@NobiGo  - I've updated the commit message based on your recommendation.  Please let me know if there is anything else required.  Thanks @ILuffZhe @NobiGo for your advice.

PR #2679, opened 2022-01-08
[CALCITE-4323] View query should be top node avoid losing sorting

JiajunBernoulli:
I think in the **expandView** method, the view should be considered as the top node. 
Here are some reasons:
1. **convertQuery** is for SqlNode, the SqlNode of view is an independent new SqlNode.
2. The SqlNode of **expandView** is not directly in the subquery.
3. When creating a view, the same query will also be used as the top node.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -683,12 +683,16 @@ public class SqlToRelConverter {
     convertFrom(
         bb,
         select.getFrom());
-    if (RelOptUtil.isOrder(castNonNull(bb.root))
-        && config.isRemoveSortInSubQuery()
-        && (!bb.top || validator().isAggregate(select) || select.isDistinct()
-        || select.hasOrderBy() || select.hasWhere())
-    ) {
-      bb.setRoot(bb.root.getInput(0), true);
+    // The existence of other nodes will cause the view to be in the subquery
+    boolean hasOtherNodes = select.hasWhere()
+        || validator().isAggregate(select) || select.isDistinct()
+        || select.hasOrderBy() || select.getFetch() != null;
+    // When the query has other nodes, the sort node should be removed
+    // even if the view is at the top level
+    boolean isRemoveSort = config.isRemoveSortInSubQuery() && (!bb.top || hasOtherNodes);
+    //  ORDER BY ... LIMIT is about semantics, so only remove pure order
+    if (RelOptUtil.isPureOrder(castNonNull(bb.root)) && isRemoveSort) {
+      bb.setRoot(castNonNull(bb.root).getInput(0), true);
     }
     convertWhere(
         bb,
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -683,6 +683,13 @@ protected void convertSelectImpl(
     convertFrom(
         bb,
         select.getFrom());
+    if (RelOptUtil.isOrder(bb.root)
```
julianhyde: this complex 'if' needs a comment explaining what it is doing.

Remove the space before ')'.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -5898,17 +5965,32 @@ private void checkMapSchemaDirectConnection(String s) throws SQLException {
             + "order by \"empid\" limit 2", null);
     with
         .query("select \"name\" from \"adhoc\".V order by \"name\"")
+        .withHook(Hook.SQL2REL_CONVERTER_CONFIG_BUILDER,
+            (Consumer<Holder<Config>>) configHolder ->
+                configHolder.set(configHolder.get().withRemoveSortInSubQuery(false)))
```
julianhyde, line 5970: I don't understand why this hook is still needed. The view is not top-level, so its results should not be sorted, but the LIMIT on the sorted data should still apply. Does the test succeed if you remove the hook?

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -5898,17 +5965,32 @@ private void checkMapSchemaDirectConnection(String s) throws SQLException {
...
     // ORDER BY and LIMIT in sub-query were not standard SQL until SQL:2008.)
+    with
+        .query("select \"name\" from (\n"
+            + "select * from \"adhoc\".\"EMPLOYEES\" where \"deptno\" = 10\n"
+            + "order by \"empid\" limit 2)\n"
+            + "order by \"name\"")
+        .withHook(Hook.SQL2REL_CONVERTER_CONFIG_BUILDER,
+            (Consumer<Holder<SqlToRelConverter.Config>>) configHolder ->
+                configHolder.set(configHolder.get().withRemoveSortInSubQuery(false)))
+        .returns("name=Bill\n"
```
julianhyde, line 5984: Same comments as previously. I would want the test to succeed even without the hook.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -5898,17 +5965,32 @@ private void checkMapSchemaDirectConnection(String s) throws SQLException {
...
+            + "name=Theodore\n");
+
     with
         .query("select \"name\" from (\n"
             + "select * from \"adhoc\".\"EMPLOYEES\" where \"deptno\" = 10\n"
             + "order by \"empid\" limit 2)\n"
             + "order by \"name\"")
         .returns("name=Bill\n"
+            + "name=Sebastian\n"
             + "name=Theodore\n");
```
julianhyde, line 5994: This is wrong. The LIMIT should not be removed. The sub-query should continue to return two rows.

JiajunBernoulli: You are right, I modified the code to preserve the previous behavior.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
JiajunBernoulli: The new code doesn't need a hook.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
JiajunBernoulli: I rewritten and added some comments to explain the code.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
JiajunBernoulli: Also revised, Would you please review this request again?

PR #2684, opened 2022-01-10
Corrected json stream property name

mans2singh:
Corrected stream property name (`steram` to `stream`) in JsonView constructor.

### DIFF
--- a/core/src/main/java/org/apache/calcite/model/JsonView.java
+++ b/core/src/main/java/org/apache/calcite/model/JsonView.java
@@ -91,7 +91,7 @@ public class JsonView extends JsonTable {
   @JsonCreator
   public JsonView(
       @JsonProperty(value = "name", required = true) String name,
-      @JsonProperty("steram") JsonStream stream,
+      @JsonProperty("stream") JsonStream stream,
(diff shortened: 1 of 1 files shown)


asolimando:
The fix is of course good, I am not familiar with this class though, but I guess it's used to de/serialize views into json files.

Therefore, I guess this qualifies as a breaking change, because existing json files with the field written as per the typo, won't be loaded anymore by Calcite after the fix. 

I'd like to hear a second opinion before merging (is it enough to mark the PR with the label `breaking change` to have it marked in the next release notes, or something more is needed?).

mans2singh:
Hi @asolimando - 

Thanks for your review. 

I did a quick search and did not find any reference to `steram` in the code.  Please let me know if you have any advice for me.  

Thanks

asolimando:
It's not necessary to find code references in order to have a breaking change (consider json files created with Calcite <= 1.29.0, if we get this in for 1.30.0, they won't be loaded anymore since the property name changed). 

Anyway it's a minor thing, just a run of `sed` would be able to fix by massive replacement, so I will merge this shortly.

asolimando (approved):
LGTM

PR #2686, opened 2022-01-11
[CALCITE-4982] NonNull field shouldn't be pushed down into leaf of outer-join in 'ProjectJoinTransposeRule'

wojustme:
This pr is prepared for [ISSUE-4982](https://issues.apache.org/jira/browse/CALCITE-4982)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
@@ -169,8 +169,8 @@ public class ProjectJoinTransposeRule
             final RexCall castCall = (RexCall) expr;
             final RelDataType operand0Type = castCall.getOperands().get(0).getType();
             if (relType.getSqlTypeName() == operand0Type.getSqlTypeName()
-                && relType.isNullable() != operand0Type.isNullable()) {
-              // Non push down cast's expression with the same type by default
+                && operand0Type.isNullable() && !relType.isNullable()) {
+              // Do not push down not nullable cast's expression with the same type by default
               // eg: CAST($1):VARCHAR(10) NOT NULL, and type of $1 is nullable VARCHAR(10)
               return false;
             }
--- a/core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
@@ -31,15 +31,12 @@ import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexOver;
 import org.apache.calcite.rex.RexShuttle;
 import org.apache.calcite.sql.SqlKind;
-import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.tools.RelBuilderFactory;
 
 import org.immutables.value.Value;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.function.Function;
-import java.util.function.Predicate;
 
 import static java.util.Objects.requireNonNull;
 
--- a/core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
@@ -24,17 +24,22 @@ import org.apache.calcite.rel.core.Join;
 import org.apache.calcite.rel.core.Project;
 import org.apache.calcite.rel.logical.LogicalJoin;
 import org.apache.calcite.rel.logical.LogicalProject;
+import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexOver;
 import org.apache.calcite.rex.RexShuttle;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.tools.RelBuilderFactory;
 
 import org.immutables.value.Value;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Function;
+import java.util.function.Predicate;
 
 import static java.util.Objects.requireNonNull;
 
@@ -157,7 +162,24 @@ public class ProjectJoinTransposeRule
   @Value.Immutable(singleton = false)
   public interface Config extends RelRule.Config {
     Config DEFAULT = ImmutableProjectJoinTransposeRule.Config.builder()
-        .withPreserveExprCondition(expr -> !(expr instanceof RexOver))
+        .withPreserveExprCondition(expr -> {
+          // Non push down over's expression by default
+          if (expr instanceof RexOver) {
+            return false;
+          }
+          final RelDataType relType = expr.getType();
+          if (SqlKind.CAST == expr.getKind()) {
+            final RexCall castCall = (RexCall) expr;
+            final RelDataType operand0Type = castCall.getOperands().get(0).getType();
+            if (relType.getSqlTypeName() == operand0Type.getSqlTypeName()
+                && relType.isNullable() != operand0Type.isNullable()) {
+              // Non push down cast's expression with the same type by default
+              // eg: CAST($1):VARCHAR(10) NOT NULL, and type of $1 is nullable VARCHAR(10)
+              return false;
+            }
+          }
+          return true;
+        })
         .build()
         .withOperandFor(LogicalProject.class, LogicalJoin.class);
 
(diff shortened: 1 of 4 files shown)


core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -9004,6 +9005,38 @@ LogicalProject(EXPR$0=[$1], EXPR$1=[$0])
...
+      LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+      LogicalTableScan(table=[[CATALOG, SALES, BONUS]])
+    LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
+]]>
+    </Resource>
+    <Resource name="planAfter">
+      <![CDATA[
+LogicalProject(EXPR$0=[CASE(IS NOT NULL($4), CAST($4):VARCHAR(10) NOT NULL, $1, $2, '':VARCHAR(10))])
+  LogicalJoin(condition=[=($0, $3)], joinType=[left])
+    LogicalProject(DEPTNO=[$1], EXPR$0=[IS NOT NULL($3)], EXPR$1=[CAST($3):VARCHAR(10) NOT NULL])
```
NobiGo, line 9032: @wojustme If the $3 is NULL, then what will happen in `EXPR$1=[CAST($3):VARCHAR(10) NOT NULL])`? Because the EXPR1 will execute any time. right? 
But  the plan bafore, when we confirm $3 is not NULL, we will to calcute `[CAST($3):VARCHAR(10) NOT NULL])`,
So please try when job include NULL value not just claim this filed can NULL, Plase check the data, I don't know, Just want to confirm.

NobiGo (changes requested):

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
wojustme: @NobiGo 
It's good question for me. 
In my project, we use calcite as an sql-proxy, and send dialectal sql to various query engine.
`EXPR$1=[CAST($3):VARCHAR(10) NOT NULL])` will be translated to sql fragment of `CAST(job AS VARCHAR)`
So, I didn't pay attention to this problem. I try to find some codes or docs to explain this question.

In calcite's linq4j, `CAST($3):VARCHAR(10) NOT NULL` will be translated to `input_isNull0 ? (String) null : input_value0`

NobiGo: Hi, Please try validate this SQL result in out.iq and submit the test in this PR? I don't know.

wojustme:
Hi @NobiGo 
Take your time to review this conversation.
Thanks.

wojustme:
Hi @NobiGo 
Are you going on reviewing this pr?

libenchao:
What's the status of this PR?

wojustme:
@libenchao 
Thank you for reminding me, and I almost forgot this issue.
I will update this pr as soon as possible.

wojustme:
@libenchao 
This issue/pr was so long that it took me a little time to review it.
Two solutions:
1. Avoid pushing down with the operand of coalesce's operator
2. Keep pushing down, and fix `org.apache.calcite.plan.Strong#policy(org.apache.calcite.rex.RexNode)`

Let's discuss in the [JIRA](https://issues.apache.org/jira/browse/CALCITE-4982)

wojustme:
Add default config: PreserveExprCondition
Shouldn't push non-null field, which is wrapped by CAST's operator.
Hi @libenchao 
Please review this pr again, thanks a lot.

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
```diff
@@ -157,7 +159,24 @@ public ProjectJoinTransposeRule(
   @Value.Immutable(singleton = false)
   public interface Config extends RelRule.Config {
     Config DEFAULT = ImmutableProjectJoinTransposeRule.Config.builder()
-        .withPreserveExprCondition(expr -> !(expr instanceof RexOver))
+        .withPreserveExprCondition(expr -> {
+          // Non push down over's expression by default
```
libenchao, line 163:
```suggestion
          // Do not push down over's expression by default
```

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
```diff
@@ -157,7 +159,24 @@ public ProjectJoinTransposeRule(
...
+          if (expr instanceof RexOver) {
+            return false;
+          }
+          final RelDataType relType = expr.getType();
+          if (SqlKind.CAST == expr.getKind()) {
+            final RexCall castCall = (RexCall) expr;
+            final RelDataType operand0Type = castCall.getOperands().get(0).getType();
+            if (relType.getSqlTypeName() == operand0Type.getSqlTypeName()
+                && relType.isNullable() != operand0Type.isNullable()) {
```
libenchao, line 172: casting `not null` to `nullable` could be pushed down?

libenchao (commented):
@wojustme Thanks for updating. 
In the JIRA discussion, I was suggesting not pushing `CASE`'s operands partially through `Join`, which is a more general way. However, it seems your approach could also fix this issue.  

I'm fine with current approach since it indeed works for this issue, and leave others to later issues, maybe in CALCITE-5315.

wojustme:
In my opinion, this case only works on in CALCITE's optimizer or validator, which some expresion only change nullable attribute.
we couldn't write an sql `cast(c1 as not integer)`, and we only write an sql `cast(c1 as integer)`.

This pr will fix the issue with a little modification. Maybe we get a good solution from the discussion of [CALCITE-5315](https://issues.apache.org/jira/browse/CALCITE-5315).

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
libenchao: kindly ping.

wojustme:
Hi @libenchao 
PR is ready, and could you take your time to review it again, thanks.

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
libenchao: @wojustme Here is one missing comment which hasn't been addressed.

wojustme: @libenchao Thanks for your reminder, but I check this pr, and I found that I have update this code, screenshot as follows:
image

libenchao: It's in different line, I'm saying line 163

wojustme: I'm sorry for it.
Please review it, thanks a lot.
@libenchao

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
```diff
@@ -157,7 +159,24 @@ public ProjectJoinTransposeRule(
   @Value.Immutable(singleton = false)
   public interface Config extends RelRule.Config {
     Config DEFAULT = ImmutableProjectJoinTransposeRule.Config.builder()
-        .withPreserveExprCondition(expr -> !(expr instanceof RexOver))
+        .withPreserveExprCondition(expr -> {
+          // Do not push down over's expression by default
+          if (expr instanceof RexOver) {
+            return false;
+          }
+          final RelDataType relType = expr.getType();
```
libenchao, line 167: move this into to `if` branch

libenchao (commented):
@wojustme LGTM with one minor comment, after fixing that, I'll merge this.

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinTransposeRule.java
wojustme: Hi @libenchao 
I have moved this code into `else` branch, I guess this's your suggestion.

libenchao: Sorry that I was not clear enough on this. I mean:
```java
          if (expr instanceof RexOver) {
            return false;
          }
          if (SqlKind.CAST == expr.getKind()) {
            final RelDataType relType = expr.getType();
            final RexCall castCall = (RexCall) expr;
            final RelDataType operand0Type = castCall.getOperands().get(0).getType();
            if (relType.getSqlTypeName() == operand0Type.getSqlTypeName()
                && operand0Type.isNullable() && !relType.isNullable()) {
              // Do not push down not nullable cast's expression with the same type by default
              // eg: CAST($1):VARCHAR(10) NOT NULL, and type of $1 is nullable VARCHAR(10)
              return false;
            }
          }
```

wojustme: Got it, and pr has been updated.
Please review it.
@libenchao

libenchao (approved):
+1, merging.

libenchao:
fixed via https://github.com/apache/calcite/commit/43db14cf8e08230c309eeb425a568e77db964a46

PR #2688, opened 2022-01-12
[MINOR] remove unused methods

xuzifu666:


xuzifu666:
@jcamachor  hi, the pr remove some unused methods, and have a review please, thanks

michaelmior:
These are public methods that are being removed. I'm not convinced that there are no uses of these methods by users of Calcite.

xuzifu666:
> These are public methods that are being removed. I'm not convinced that there are no uses of these methods by users of Calcite.

user can use the method by they self in other way， these methods are useless in the project.

NobiGo (commented):
I think we can't remove this method directly. Maybe we can add java deprecated annotation. like 
```
  @Deprecated // to be removed before 2.0
```

and add some java doc why need to remove.

xuzifu666:
> I think we can't remove this method directly. Maybe we can add java deprecated annotation. like
> 
> ```
>   @Deprecated // to be removed before 2.0
> ```
> 
> and add some java doc why need to remove.

i see, i would add @Deprecated on these methods @NobiGo

rubenada:
> user can use the method by they self in other way， these methods are useless in the project. so i advise remove these methods to improve calcite code for other developers
@xuzifu666 could you please elaborate why these methods are useless? I agree with @michaelmior , we need to be careful when removing public features.

xuzifu666:
> elaborate
i found no place use these method in whole project or tests，so i though these methods are useless. if these methods are used at a place in Calcite or tests, the ci/cd would not be successful, but we can see ci/cd success. i added   @Deprecated on these methods now. whether remove can decide in next version. now i though mark it deprecated is nesscessary. @rubenada  do you aggree with me?

asolimando:
CI/CD won't necessarily break if you remove them, they might be unused _within_ Calcite, but they can be very well used by Calcite users, removing them might break their code.

What others suggested, going through a `@Deprecated` annotation is what we generally do.

However, it would be good to at least comment on why some methods are deprecated, and what to use instead in order to achieve the same goal, if possible (e.g., multiple versions of a constructor for a given class).

rubenada:
Agree with @asolimando .
@xuzifu666 I agree it is not ideal to have untested features inside Calcite code. However, having unused public methods inside Calcite does not mean that some downstream projects using Calcite do not use them. Those projects will stop working if we remove them. In this particular case, at first glance, these methods don't seem out-of-date / useless in my opinion.

xuzifu666:
> CI/CD won't necessarily break if you remove them, they might be unused _within_ Calcite, but they can be very well used by Calcite users, removing them might break their code.
> 
> What others suggested, going through a `@Deprecated` annotation is what we generally do.
> 
> However, it would be good to at least comment on why some methods are deprecated, and what to use instead in order to achieve the same goal, if possible (e.g., multiple versions of a constructor for a given class).
ok, these methods can be used by normal non agrs constructor to build, need provided with a specified method，according  to design patterns the way of providing a method which is not used in project internally is not perfect. @asolimando @rubenada

xuzifu666:
hi, i change add @SuppressWarnings to these methods. do you aggree with me? @asolimando

PR #2692, opened 2022-01-14
[CALCITE-4980] Babel parser support MySQL NULL-safe equal operator '<=>' (xurenhe&&DuanXiong)

wojustme:


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -706,4 +707,17 @@ private SqlLibraryOperators() {
   @LibraryOperator(libraries = { POSTGRESQL })
   public static final SqlOperator INFIX_CAST =
       new SqlCastOperator();
+
+  /** Null-safe "&lt;=&gt;" equal operator used by MySQL, for example
+   * {@code 1<=>NULL}. */
+  @LibraryOperator(libraries = { MYSQL })
+  public static final SqlOperator NULL_SAFE_EQUAL =
```
NobiGo, line 714: I think The NULL_SAFE_EQUAL **is** IS_NOT_DISTINCT_FROM. Not belong to. So

```
public static final SqlOperator NULL_SAFE_EQUAL = SqlStdOperatorTable.IS_NOT_DISTINCT_FROM;
```
right ?

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -261,6 +261,15 @@ private void checkParseInfixCast(String sqlType) {
     sql(sql).ok(expected);
   }
 
+  /** Tests parsing MySQL-style "<=>" equal operator. */
+  @Test void testParseNullSafeEqual()  {
+    String sql = "SELECT x <=> 3 FROM (VALUES (1, 2)) as tbl(x,y)";
+    String expected = ""
```
NobiGo, line 267: Please reformat the SQL.

NobiGo (changes requested):

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
wojustme: Yes, you are right.
I update it.
Thanks.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
wojustme: Got it. I have updated.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
NobiGo: Why do we need to change the operator's name? If we use the SqlStdOperatorTable.IS_NOT_DISTINCT_FROM directly, Then can support parse and execute. Now only can support parse. If I miss something please let me know. 
and Add another parse test cases for `(a, b) <=> (x, y)`

wojustme: @NobiGo 
When using SqlStdOperatorTable#IS_NOT_DISTINCT_FROM, '`X` <=> 3' will return '`X` IS NOT DISTINCT FROM 3'.
I think it destroys the user's original input.

NobiGo: According to [MySQL '<=>'](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#operator_equal-to) . We should cover it.

NobiGo: Please don't force fush commit, until we have a final state.

wojustme: @NobiGo 
Ok, got it.
In my project, I didn't meet these complex scenes.
I will add some related test cases.
Thanks a lot.

NobiGo:
@wojustme Because refactoring tests in CALCITE-4885, So please fix the conflicts and the typo above.

NobiGo (changes requested):
Some small typos need fix.

site/_docs/reference.md
```diff
@@ -1203,13 +1203,13 @@ The operator precedence and associativity, highest to lowest.
 | * / % &#124;&#124;                                | left
 | + -                                               | left
 | BETWEEN, IN, LIKE, SIMILAR, OVERLAPS, CONTAINS etc. | -
-| < > = <= >= <> !=                                 | left
+| < > = <= >= <> != <=>                                | left
```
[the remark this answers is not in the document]
wojustme: ok, I format it.

NobiGo (approved):
LGTM

PR #2693, opened 2022-01-17
[CALCITE-4987] JDBC adapter generates incorrect query when ORDER BY alias collides with field in SELECT ITEM

hannerwang:
After this pr, JDBC adapter generates ORDER BY clause with the following preference:
1. if dialect conformance support sort by ordinal, prefer ORDER BY ordinal.
2. if dialect conformance support sort by alias, prefer ORDER BY alias.
3. resolve to ORDER BY SELECT ITEM without as.

Exceptions:
1. if the ORDER BY clause is in WINDOW or WITHIN GROUP, always resolve to ORDER BY SELECT ITEM.
2. if the ORDER BY ordinal is greater than size of SELECT LIST, always resolve to ORDER BY SELECT ITEM.


FDoKE:
If you mind me, i'll drop related error here.
query (some_aggregate something like string_agg)
`select some_aggregate(id, ',') WITHIN GROUP(ORDER BY id) as agg, somecol as id GROUP BY somecol`
when one column have exact same alias as aggregate `order by` column name this causes incorrect **ordinal** usage in ORDER BY (it produces ordinal=4 which has no use inside aggregate function)

PR #2694, opened 2022-01-17
[CALCITE-4965] IS NOT NULL failed in Elasticsearch Adapter

ILuffZhe:
NamedFieldExpression is packed in CastExpression while translating bool condition.


asereda-gs:
Thank you!

PR #2695, opened 2022-01-18
[CALCITE-4954] Group TEXT field failed in Elasticsearch Adapter

ILuffZhe:
Set "**fielddata**" to "**true**" in Elasticsearch mapping, since aggregations and sorting are disabled by default for text type.

It's only set in the test module, which provides instruction for SQL users when they want to group or sort text field.


NobiGo (approved):
LGTM

mihaibudiu:
This seems to have been approved more than 1 year ago, so if @ILuffZhe can squash the commits we can merge it?

ILuffZhe:
@mihaibudiu Sorry for the delay, I've rebased this PR, please take a look when you are free.

mihaibudiu (approved):

PR #2696, opened 2022-01-21
[CALCITE-4899] Support hint for lateral table syntax

Tartarus0zm:
Support hint for lateral table syntax

## Verifying this change
* Add tests in `SqlHintsConverterTest` to validate the hint could be propagated to Correlate
* Add tests in `RelToSqlConverterTest`


Tartarus0zm:
@danny0405  please take a look, if you have time, thanks

core/src/main/codegen/templates/Parser.jj
```diff
@@ -2173,10 +2176,23 @@ SqlNode TableRef2(boolean lateral) :
...
+            <COMMENT_END>
+        ]
         {
             if (lateral) {
-                tableRef = SqlStdOperatorTable.LATERAL.createCall(
-                    s.end(this), tableRef);
+                if (hints.size() > 0) {
+                    hintList = new SqlNodeList(hints, s.addAll(hints).end(this));
+                    SqlNode tmpTableRef = SqlStdOperatorTable.LATERAL.createCall(
+                                        s.end(this), tableRef);
```
danny0405, line 2189: `tmpTableRef ` => `lateralCall`

core/src/main/java/org/apache/calcite/sql/SqlCorrelateTableRef.java
```diff
@@ -0,0 +1,73 @@
...
+
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+import com.google.common.collect.ImmutableList;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.List;
+
+import static java.util.Objects.requireNonNull;
```
danny0405, line 27: Can we just reuse `SqlTableRef` code ? We can switch the `tableName` to `tableRef` which is a `SqlNode`.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -1225,7 +1226,7 @@ private SqlValidatorScope getScopeOrThrow(SqlNode node) {
     return getNamespace(id);
   }
 
-  @Override public @Nullable SqlValidatorNamespace getNamespace(SqlNode node) {
+  @Override public @Nullable SqlValidatorNamespace  getNamespace(SqlNode node) {
     switch (node.getKind()) {
```
danny0405, line 1230: unnecessary change.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -2441,9 +2447,8 @@ private SqlNode registerFrom(
               null,
               forceRightNullable,
               lateral);
-      if (newRight != right) {
-        join.setRight(newRight);
-      }
+      // for check style.
+      swapJoinRightNode(join, right, newRight);
```
danny0405, line 2451: Don't think this is a right change, you should make the `newRight` correct first.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -3001,12 +3009,20 @@ private void convertJoin(Blackboard bb, SqlJoin join) {
         throw Util.unexpected(conditionType);
       }
     }
+
+    final List<RelHint> hints;
+    if (right instanceof SqlCorrelateTableRef) {
+      hints = SqlUtil.getRelHint(hintStrategies, ((SqlCorrelateTableRef) right).operand(1));
+    } else {
```
danny0405, line 3016: A way to avoid these `instanceOf` is add a new `SqlKind`.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
Tartarus0zm: good catch

core/src/main/java/org/apache/calcite/sql/SqlCorrelateTableRef.java
Tartarus0zm: good catch

Tartarus0zm:
@danny0405  I have updated the commit, please take a look

core/src/main/codegen/templates/Parser.jj
danny0405: I feel it is more reasonable to include the `tableRef` into the `LATERAL` call not the other way around, say we should change the lateral call to support `tableRef` as the table reference, WDYT ?

Tartarus0zm: I've already changed it

Tartarus0zm:
@danny0405  I have updated the commit, please take a look, thanks

danny0405:
I can not find any vendor that supports hints for functions, let's be conservative here, does the query hints like

```java
/*+ tvf_join("hash", 'table name') */
```

works here ? You can add custom validations for the hint anyway.

PR #2698, opened 2022-01-21
[CALCITE-4992] Close the ElasticSearch RestClient in ElasticsearchTransport

jnturton:

The change attempts to address a resource leak bug believed to result from no call to RestClient#close() being made.  [Docs describing the need for this call are online here](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/java-rest-low-usage-initialization.html).

~~The call to RestClient#close() is added to a new `finalize()` method in ElasticsearchTransport because no other suitable lifecycle method was evident to me.  Any suggestion of a better place to make this call is welcome.~~

Instead of creating a new RestClient instance every time a new Schema is created, shared RestClient instances are cached and reused. If the cache grows beyond a fixed number, currently 100, of distinct RestClients then evictions begin along with the logging of warnings.  Upon the eviction of a RestClient, a call to its `close()` method is made.

## Verifying this change

File descriptor resource leaks such as we have observed in the context connecting Apache Drill to ElasticSearch (which goes via Calcite) are not easy to unit test.  I'm open to ideas here...

### DIFF
--- a/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
+++ b/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
@@ -67,7 +67,7 @@ public class ElasticsearchSchemaFactory implements SchemaFactory {
       .maximumSize(1000)
       .expireAfterAccess(1, TimeUnit.HOURS)
       .removalListener(new RemovalListener<Integer, RestClient>() {
-        public void onRemoval(RemovalNotification<Integer, RestClient> notice) {
+        @Override public void onRemoval(RemovalNotification<Integer, RestClient> notice) {
           try {
             // Free resources allocated by this RestClient
             notice.getValue().close();
--- a/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
+++ b/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
@@ -47,7 +47,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 /**
@@ -60,19 +59,34 @@ public class ElasticsearchSchemaFactory implements SchemaFactory {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ElasticsearchSchemaFactory.class);
 
+  private static final int REST_CLIENT_CACHE_SIZE = 100;
+
   // RestClient objects allocate system resources and are thread safe. Here, we
   // cache them using a key derived from the hashCode()s of the parameters that
-  // define a RestClient.
+  // define a RestClient. The primary reason to do this is to limit the resource leak
+  // that results from Calcite's current inability to close clients that it creates.
+  // Amongst the OS resources leaked are file descriptors which are limited to
+  // 1024 per process by default on Linux at the time of writing.
   private static Cache<Integer, RestClient> restClients = CacheBuilder.newBuilder()
-      .maximumSize(1000)
-      .expireAfterAccess(1, TimeUnit.HOURS)
+      .maximumSize(REST_CLIENT_CACHE_SIZE)
       .removalListener(new RemovalListener<Integer, RestClient>() {
         @Override public void onRemoval(RemovalNotification<Integer, RestClient> notice) {
+          LOGGER.warn(
+            "Will close an ES REST client to keep the number of open clients under {}",
+            REST_CLIENT_CACHE_SIZE
+          );
+          LOGGER.warn(
+            "Any schema objects created with this client are now broken!\n"
+  +
+            "Do not try to access more than {} distinct ES REST APIs through this adapter.",
+            REST_CLIENT_CACHE_SIZE
+          );
+
           try {
             // Free resources allocated by this RestClient
             notice.getValue().close();
           } catch (IOException ex) {
-            LOGGER.warn("Could not close RestClient %s", notice.getValue());
+            LOGGER.warn("Could not close RestClient {}", notice.getValue());
           }
         }
       })
--- a/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
+++ b/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
@@ -31,6 +31,10 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.RemovalListener;
+import com.google.common.cache.RemovalNotification;
 
 import org.elasticsearch.client.RestClient;
 import org.elasticsearch.client.RestClientBuilder;
@@ -41,6 +45,9 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 /**
@@ -52,6 +59,24 @@ import java.util.stream.Collectors;
 public class ElasticsearchSchemaFactory implements SchemaFactory {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ElasticsearchSchemaFactory.class);
+
+  // RestClient objects allocate system resources and are thread safe. Here, we
+  // cache them using a key derived from the hashCode()s of the parameters that
+  // define a RestClient.
+  private static Cache<Integer, RestClient> REST_CLIENTS = CacheBuilder.newBuilder()
+      .maximumSize(1000)
+      .expireAfterAccess(1, TimeUnit.HOURS)
+      .removalListener(new RemovalListener<Integer, RestClient>() {
+        public void onRemoval(RemovalNotification<Integer, RestClient> notice) {
+          try {
+            // Free resources allocated by this RestClient
+            notice.getValue().close();
+          } catch (IOException ex) {
+            LOGGER.warn("Could not close RestClient %s", notice.getValue());
+          }
+        }
+      })
+      .build();
 
   public ElasticsearchSchemaFactory() {
   }
@@ -110,7 +135,7 @@ public class ElasticsearchSchemaFactory implements SchemaFactory {
    * @param hosts list of ES HTTP Hosts to connect to
    * @param username the username of ES
    * @param password the password of ES
-   * @return newly initialized low-level rest http client for ES
+   * @return new or cached low-level rest http client for ES
    */
   private static RestClient connect(List<HttpHost> hosts, String pathPrefix,
                                     String username, String password) {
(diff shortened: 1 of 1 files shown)


ILuffZhe:
- One small reminder: the commit info should be like "[CALCITE-4992] RestClient resource leak in ElasticSearch adapter", which can be linked to JIRA automatically

-  Could you please remove the unnecessary blank lines?

-  Instead of closing the RestClient by JVM GC, how about initiatively closing the client after each query?

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchTransport.java
```diff
@@ -224,6 +224,16 @@ void closeScroll(Iterable<String> scrollIds) {
       return rawHttp(ElasticsearchJson.Result.class).apply(post);
     };
   }
+  
+  
+
+  @Override
+  protected void finalize() throws Throwable {
```
zabetak, line 231: Finalizers are a java concept that is strongly discouraged for quite some time now. Moreover, they are going to [disappear soon](https://www.youtube.com/watch?v=eDgBnjOid-g) so let's find another way to handle this.

zabetak (changes requested):

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchTransport.java
NobiGo: [JEP 421: Deprecate Finalization for Removal](https://openjdk.java.net/jeps/421?continueFlag=abef66f10493295d8376160baf41892b) This address have some way to replace the finalize method and Why we should avoid using it. I hope it can help.

jnturton:
@ILuffZhe @zabetak, a colleague of mine has raised a concern that recreating the RestClient for every request, as I've done in this iteration, might negatively affect effeciency.  It would be nice to get your thoughts on this.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/ElasticSearchAdapterTest.java
```diff
@@ -105,7 +105,7 @@ private static Connection createConnection() throws SQLException {
         connection.unwrap(CalciteConnection.class).getRootSchema();
 
     root.add("elastic",
-        new ElasticsearchSchema(NODE.restClient(), NODE.mapper(), ZIPS));
+             new ElasticsearchSchema(NODE.clientBuilder(), NODE.mapper(), ZIPS));
```
ILuffZhe, line 108: Please check the indent here, so does somewhere else.

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/EmbeddedElasticsearchPolicy.java
```diff
@@ -192,19 +201,18 @@ ObjectMapper mapper() {
    * Low-level http rest client connected to current embedded elastic search instance.
    * @return http client connected to ES cluster
    */
-  RestClient restClient() {
-    if (client != null) {
-      return client;
+  synchronized RestClientBuilder clientBuilder() {
```
ILuffZhe, line 204: Do we need a synchronized here?

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchTransport.java
```diff
@@ -279,22 +276,25 @@ private Response applyInternal(final HttpRequest request)
           request.getRequestLine().getMethod(),
           request.getRequestLine().getUri());
       r.setEntity(entity);
-      final Response response = restClient.performRequest(r);
 
-      final String payload = entity != null && entity.isRepeatable()
-          ? EntityUtils.toString(entity) : "<empty>";
+      try (RestClient restClient = clientBuilder.build()) {
```
ILuffZhe, line 280: If I understand this correctly, for each unit test, we will create a new RestClient to request, which is what you mentioned.

- In my team, we manage the RestClients by Cache(for lifecycle and something else). Since it is actual business related, we might don't make it here
- For unit tests, I'm wandering if there is a suitable way to manage the RestClient in BeforeClass and AfterClass. Just my personal opinion.

jnturton:
@ILuffZhe @zabetak, sorry about the long delay.  Please see this alternative approach which aims to both address the leak and make efficient use of RestClient.

ILuffZhe (commented):
Hi, @jnturton . Sorry for the delay. I've been a little busy lately. Overall look good to me.

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
```diff
@@ -53,6 +60,24 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ElasticsearchSchemaFactory.class);
 
+  // RestClient objects allocate system resources and are thread safe. Here, we
+  // cache them using a key derived from the hashCode()s of the parameters that
+  // define a RestClient.
+  private static Cache<Integer, RestClient> restClients = CacheBuilder.newBuilder()
+      .maximumSize(1000)
```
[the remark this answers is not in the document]
jnturton: @ILuffZhe the choices were just made conservatively (no one will connect to more than 1000 ElasticSearch servers and replacing each RestClient at most hourly will not amount to a great load.  Testing a custom build of Drill based on a custom build Calcite would be quite lot of effort for us so can I propose that a first PR just aims to stop the resource leaking?  We can then monitor it for an extended period in one of our Drill environments and report back to you with our data if it looks like these parameters need fine tuning.

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
```diff
@@ -53,6 +60,24 @@
...
+      .expireAfterAccess(1, TimeUnit.HOURS)
+      .removalListener(new RemovalListener<Integer, RestClient>() {
+        public void onRemoval(RemovalNotification<Integer, RestClient> notice) {
```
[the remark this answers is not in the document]
jnturton: Thanks, fixing now...

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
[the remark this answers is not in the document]
ILuffZhe: Thanks for the clarification. I think this PR is better than the former one(as far as I can remember).
It's better not to force push before the final decision, so other reviewers can take a look.

jnturton:
Hi @ILuffZhe, is there anything more I should do here?

asereda-gs:
Hi @jnturton ,

Thanks for this PR. I'm not sure that having cache inside `ElasticsearchSchemaFactory.java` is the best approach. 

I understand that you want to properly manage resource lifecycle. Is there a reason you can't share existing `ElasticsearchSchema` ?

Is there a way in your environment to manually instantiate `ElasticsearchSchema` with existing `RestClient` instance (see [ElasticsearchSchema ctor](https://github.com/apache/calcite/blob/master/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchema.java#L65)? This way you can control when `RestClient` is closed.

asereda-gs:
Hi @jnturton ,

Thanks for this PR. I'm not sure that having cache inside `ElasticsearchSchemaFactory.java` is the best approach. 

I understand that you want to properly manage resource lifecycle. Is there a reason you can't share existing `ElasticsearchSchema` ?

Is there a way in your environment to manually instantiate `ElasticsearchSchema` with existing `RestClient` instance (see [ElasticsearchSchema ctor](https://github.com/apache/calcite/blob/master/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchema.java#L65)? This way you can control when `RestClient` is closed.

jnturton:
> I understand that you want to properly manage resource lifecycle. Is there a reason you can't share existing `ElasticsearchSchema` ?
> 
> Is there a way in your environment to manually instantiate `ElasticsearchSchema` with existing `RestClient` instance (see [ElasticsearchSchema ctor](https://github.com/apache/calcite/blob/master/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchema.java#L65)? This way you can control when `RestClient` is closed.

@asereda-gs I can see if we can lift the rest client management out to Drill but do you really want to retain the situation where this library creates internal objects that allocate OS resources and never calls `close()` on them, either of its own accord or in response to some message from the application?

Surely that's just bad library code and now we're arguing "yes it leaks but applications that are in position to to do a lot of reuse can make that a slow leak, alternatively they can not use the API in the obvious way and do their own resource management.  So we don't need to fix the library"?  You have to think that there are probably many applications out there using Calcite with slow resource leaks today as a result.

asereda-gs:
The problem is that `ElasticsearchSchemaFactory` (or other `SchemaFactory` implementations for that matter) creates new resource but it doesn't know exactly when to close it, only application does.

Freeing resources on `finalize()` or cache timeout looks more like a hack.

I believe there should be an explicit `close()` call from library user for `ElasticsearchSchemaFactory` to free its resources. 

Maybe `Schema` or `SchemaFactory` should be implementing `AutoCloseable`. But closing event should be initiated explicitly by the application (not GC or cache expiration thread).

So either application provides a `RestClient` and then it knows how to manage it or ask `ElasticsearchSchema` to close it. Unfortunately I don't see Calcite having `Schema.close` API.

For example `JdbcSchema`  relies on external `javax.sql.DataSource`.

Perhaps you can raise this question on the dev lists since I see similar issues with Mongo/Geode adapters.

jnturton:
@asereda-gs I agree that the right design is for the application to request the freeing of resources.  An hour or so ago I started work on making Schema implement AutoCloseable with a default implementation of `close()` that is a no-op to preserve backwards compat, but then I encountered CassandraSchemaFactory which keeps a cache of shared CqlSessions, very much like the current version of this PR except that the CqlSession cache entries never expire and are no doubt also a resource leak although a bounded one, unlike Elasticsearch without this PR.  So if simply set the cache expiry in this PR to "never" then I've ended up with exactly what Calcite does for Cassandra and I assume that would be mergeable?

While that would stop the worst bleeding, and is perhaps worth doing for stop-gap, it would still be pretty horrible.  I think Calcite might have to make a choice now.  Either 

- it abstains from connection management entirely in adapters, which breaks its public APIs since then applications must start to pass connections in or 
- it starts to use connections in single-use way, freeing immediately and taking a performance hit or
- it continues with its connection caching and to starts to run a housekeeping thread to clean up its caches.

asereda-gs:
@jnturton I believe this should be a question to the dev list.

The correct way is to address resource management issue for all Calcite adapters.

Should `Schema` extend `AutoCloseable`? Maybe `SchemaPlus` could expose something like [Closer](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/io/Closer.html) for child schemas to register for close event.

Perhaps this can be addressed in more elegant way than existing implementations.

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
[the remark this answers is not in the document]
zabetak: What happens if after one hour the client is removed and closed but there is still an active reference inside the `ElasticsearchSchema`?

zabetak:
@jnturton @asereda-gs there are many points where I agree with you.

It would be a good idea to bring up the general problem of closing resources to the dev list but I don't consider this a must for moving this forward. If we find a satisfactory solution without revising all adapter APIs then I could leave with it till we tackle the general problem.

I wouldn't mind going forward with the Guava cache but as Andrei mentioned the cache doesn't know when to close the client. What happens if the `RestClient` that we are evicting from the cache is still in use in a schema? If this is possible then I guess we are going to create a new bug while trying to fix this one.

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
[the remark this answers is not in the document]
jnturton: @zabetak thank you, that is indeed a mistake.  The ElasticsearchSchema must either be made to either only obtain a rest client from the cache (remove its own reference) or the cache expiry should be set to "never" which is the approach in the Cassandra adapter.  Which is preferred?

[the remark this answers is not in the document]
jnturton: Resolved, somewhat.  Clients can still be removed and closed but only if their total number has reached an excessive number, not because of any configured expiry time.  A warning that trouble has begun is logged.  Note that if the clients were not removed and closed and no warning was logged then trouble would nevertheless have begun because the application was heading the road to getting killed for leaking resources.

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchemaFactory.java
```diff
@@ -53,6 +60,24 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ElasticsearchSchemaFactory.class);
 
+  // RestClient objects allocate system resources and are thread safe. Here, we
+  // cache them using a key derived from the hashCode()s of the parameters that
+  // define a RestClient.
+  private static Cache<Integer, RestClient> REST_CLIENTS = CacheBuilder.newBuilder()
```
julianhyde: I’m surprised this isn’t final.

jnturton: @julianhyde in the latest push the member is final.  Also, these ES clients are dedicated to a single ES server so the consequences of a hash collision would indeed have been bad.  So the cache is no longer keyed on a hash.

PR #2699, opened 2022-01-22
[CALCITE-4988] ((A IS NOT NULL OR B) AND A IS NOT NULL) can't be simp…

NobiGo:
…lify to (A IS NOT NULL) When A is deterministic

### DIFF
--- a/core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
+++ b/core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
@@ -2157,6 +2157,82 @@ class RexProgramTest extends RexProgramTestBase {
     checkSimplify(isNotNull(cast(i2, intType)), "true");
   }
 
+  /**
+   * Unit test for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-4988">[CALCITE-4988]
+   * ((A IS NOT NULL OR B) AND A IS NOT NULL) can't be simplify to (A IS NOT NULL)
+   * When A is deterministic</a>. */
+  @Test void testSimplifyIsNotNullWithDeterministic() {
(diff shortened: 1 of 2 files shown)


ILuffZhe:
Does this case cover the situation like "(A is not null and B is null) or A is not null"?

NobiGo:
> Does this case cover the situation like "(A is not null and B is null) or A is not null"?

@ILuffZhe This can't cover this situation. The OR condition has a different way to pull up the predicate.

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -2157,6 +2157,82 @@ private void checkExponentialCnf(int n) {
...
+    // "(A IS NOT NULL OR B) AND A IS NOT NULL" when A is not deterministic
+    // ==>
+    // "(A IS NOT NULL OR B) AND A IS NOT NULL"
+    SqlOperator ndc = getNoDeterministicOperator();
+    checkSimplifyUnchanged(
+        and(or(isNotNull(rexBuilder.makeCall(ndc)), gt(vInt(2), literal(2))),
+            isNotNull(rexBuilder.makeCall(ndc))));
+  }
+
+  @Test void testSimplifyIsNotNullWithNoDeterministic2() {
```
asolimando, line 2198: I can't spot any difference w.r.t. `testSimplifyIsNotNullWithNoDeterministic`, is this a copy-paste error or am I missing something?

asolimando: Never mind, just saw the OR vs AND, my bad!

asolimando (approved):
LGTM!

PR #2701, opened 2022-01-25
[CALCITE-4994] SQL-to-RelNode conversion is slow if table contains hundreds of fields

jaystarshot:
PS-    I am new in calcite so wanted your suggestions if any benchmarking, testing is required .

 fix issue:[ CALCITE-4994](https://issues.apache.org/jira/browse/CALCITE-4994)
This was a bottleneck in In my database, I had observed that for a large table involving 1200 columns and a huge select having multiple expressions and operators, this part was a bottleneck.

I wrote some jmh tests but i could not load this table as I could not find the API to load create tables into catalog for testing

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -4106,8 +4106,9 @@ public class SqlToRelConverter {
     } else {
       qualified = SqlQualified.create(null, 1, null, identifier);
     }
-    final Pair<RexNode, @Nullable BiFunction<RexNode, String, RexNode>> e0 =
-        bb.lookupExp(qualified);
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4817,13 +4819,17 @@ void setRoot(List<RelNode> inputs) {
         if (node == null) {
```
julianhyde, line 4819: is node ever null? if not, remove this `if`.

julianhyde: also, does `lookupExp` ever return null due to a user error? I don't think so. There's only one caller, and it throws if it receives null. So let's tighten this down, and make the function throw rather than ever returning null.

julianhyde (commented):
a couple of observations how we can simplify program flow. not your fault, but if we want to improve performance we must simplify.

PR #2704, opened 2022-01-28
Added calcite-clj references to adpater page and talks

ieugen:


zabetak:
Thanks for the PR @ieugen . I left out the changes in the adapter page since we are not talking about an adapter there. Let's put that in the new community section ;)

PR #2706, opened 2022-01-31
[CALCITE-4995] AssertionError caused by RelFieldTrimmer on SEMI/ANTI join

rubenada:
More info in [CALCITE-4995](https://issues.apache.org/jira/browse/CALCITE-4995)


thomasrebele:
LGTM.

PR #2709, opened 2022-02-02
[CALCITE-4996] in RelJson, add a new public toRex method which can ov…

mkou:
Creates a new public static method that transforms a RexNode tree defined in a map (from a JSON) into a RexNode, with a way to apply a special method to handle the inputs
```
/**
   * Transforms a RexNode tree defined in a map (from a JSON) into a RexNode
   * Applying a special method to inputs instead of transforming them into inputRef
   * @param cluster The optimization environment
   * @param apply is a InputTranslator lambda that transforms the map representing input
   *               references into a RexNode
   * @param o the map derived from a RexNode transformed into a JSON
   * @return the transformed RexNode
   */
   ```

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
+++ b/core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
@@ -19,6 +19,8 @@ package org.apache.calcite.rel.externalize;
 import org.apache.calcite.avatica.AvaticaUtils;
 import org.apache.calcite.avatica.util.TimeUnit;
 import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.plan.RelOptTable;
+import org.apache.calcite.plan.RelTraitSet;
 import org.apache.calcite.rel.RelCollation;
 import org.apache.calcite.rel.RelCollationImpl;
 import org.apache.calcite.rel.RelCollations;
@@ -76,7 +78,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.function.BiFunction;
 
 import static org.apache.calcite.rel.RelDistributions.EMPTY;
 import static org.apache.calcite.util.Static.RESOURCE;
@@ -90,6 +91,7 @@ import static java.util.Objects.requireNonNull;
 public class RelJson {
   private final Map<String, Constructor> constructorMap = new HashMap<>();
   private final @Nullable JsonBuilder jsonBuilder;
+  private final InputTranslator inputTranslator;
 
   public static final List<String> PACKAGES =
       ImmutableList.of(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by asolimando** [comment]
@mkou for general discussions and tips I suggest to send an email to the ML, you have better chances to get a reply than someone reading the PR description

**Comment by julianhyde** [comment]
@asolimando, I see that @mkou added similar description in the Jira case https://issues.apache.org/jira/browse/CALCITE-4996. I think that the Jira case is the right place for discussion.

IMHO each Jira case is its own conversation thread. Cases appear on the dev list when they are logged. If you are interested in seeing how the discussion develops, press the 'watch' button.

File: core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java [hunk 1]
```diff
@@ -776,4 +905,22 @@ private void addRexFieldCollationList(
     map.put("syntax", operator.getSyntax().toString());
     return map;
   }
+
+  /**
```
julianhyde, line 909: I wouldn't mention the 'functional interface'. Document its purpose, which is to translate.

Rename `apply` to something like `translate`. `apply` only made sense when this was a `BiFunction`.

In the method javadoc change `Lambda that defines how to transform` to `Transforms`. Less is more.

Rename `stringObjectMap` to `map`.

core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
```diff
@@ -429,6 +435,11 @@
     return Stream.of(SqlExplainFormat.TEXT, SqlExplainFormat.DOT);
   }
 
+  private static RexNode apply(Map<String, Object> map, RexBuilder rexBuilder, List<RelNode> inputs) {
```
julianhyde, line 438: need better name than `apply`. Also a bit of javadoc describing intend. In the javadoc I'd mention that it is intended to implement the translator interface.

core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
```diff
@@ -706,6 +717,51 @@
...
+        + "        {\n"
+        + "          \"literal\": 1,\n"
+        + "          \"type\": { \"type\": \"INTEGER\", \"nullable\": false }\n"
+        + "        }\n"
+        + "      ]\n"
+        + "    }";
+    assertReadRex(jsonString2, "+(1, 1)");
+  }
+
+  private void assertReadRex(String jsonString1, String expected)
```
julianhyde, line 752: better name would be `assertThatReadExpressionResult`

change parameter to from `String` to `Matcher<String>`. then there will be an `is` at the call site, and the code will read somewhat like English (per junit best practices, e.g. `assertThat(1 + 2, is(3));`

core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -99,7 +102,143 @@
...
+   * Transforms a RexNode tree defined in a map (from a JSON) into a RexNode,
+   * applying a special method to inputs instead of transforming them into inputRef.
+   * @param cluster The optimization environment
+   * @param apply is a InputTranslator lambda that transforms the map representing input
+   *               references into a RexNode
+   * @param o the map derived from a RexNode transformed into a JSON
+   * @return the transformed RexNode
+   */
+  public static RexNode readExpression(
```
julianhyde, line 143: Let's move this method a bit further down the file. It's not the most important method in this class.

The description 'Transforms a RexNode tree' is a bit misleading, given that the input is JSON not RexNode. Maybe something like 'Converts a JSON expression...'?

Convert the anonymous inner class that implements RelInput into a private static inner class. That gives us an opportunity to document it. Also it will make it easier to refactor in future.

core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -99,7 +102,143 @@
...
+  public RelJson(@Nullable JsonBuilder jsonBuilder, InputTranslator inputTranslator) {
     this.jsonBuilder = jsonBuilder;
+    this.inputTranslator = inputTranslator;
+  }
+
+  /**
+   * Transforms a RexNode tree defined in a map (from a JSON) into a RexNode,
+   * applying a special method to inputs instead of transforming them into inputRef.
+   * @param cluster The optimization environment
+   * @param apply is a InputTranslator lambda that transforms the map representing input
```
julianhyde, line 138: rename `apply` parameter to `translator`

julianhyde (commented):
suggest a few small edits

I am still thinking about whether the `Map<String, Object` should be an `int ordinal` but that's a bigger design question.

PR #2710, opened 2022-02-02
[CALCITE-4997] Keep APPROX_COUNT_DISTINCT in some SqlDialects

JiajunBernoulli:
`SELECT APPROX_COUNT_DISTINCT(product_id)
FROM foodmart.product`
will be
`SELECT COUNT(DISTINCT product_id)
FROM foodmart.product`
This can cause many tasks to run too slowly, we should keep approx_count_distinct function.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -1151,18 +1151,18 @@ public abstract class SqlImplementor {
     public SqlNode toSql(AggregateCall aggCall) {
       return toSql(aggCall.getAggregation(), aggCall.isDistinct(),
           Util.transform(aggCall.getArgList(), this::field),
-          aggCall.filterArg, aggCall.collation);
+          aggCall.filterArg, aggCall.collation, aggCall.isApproximate());
     }
 
     /** Converts a call to an aggregate function, with a given list of operands,
      * to an expression. */
     private SqlCall toSql(SqlOperator op, boolean distinct,
-        List<SqlNode> operandList, int filterArg, RelCollation collation) {
+        List<SqlNode> operandList, int filterArg, RelCollation collation, boolean approximate) {
       final SqlLiteral qualifier =
           distinct ? SqlSelectKeyword.DISTINCT.symbol(POS) : null;
       if (op instanceof SqlSumEmptyIsZeroAggFunction) {
         final SqlNode node = toSql(SqlStdOperatorTable.SUM, distinct,
-            operandList, filterArg, collation);
+            operandList, filterArg, collation, approximate);
         return SqlStdOperatorTable.COALESCE.createCall(POS, node, ZERO);
       }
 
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1179,7 +1179,9 @@ private SqlCall toSql(SqlOperator op, boolean distinct,
 
       // Handle filter by generating FILTER (WHERE ...)
       final SqlCall call2;
-      if (filterArg < 0) {
```
jaystarshot, line 1182: Why cannot we just check if `op instanceof SqlOperatorTable.APPROX_COUNT_DISTINCT`

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
```diff
@@ -725,6 +725,11 @@ public boolean supportsAggregateFunction(SqlKind kind) {
     return false;
   }
 
+  /** Returns whether this dialect supports APPROX_COUNT_DISTINCT functions)}. */
```
NobiGo, line 728: Please remove the redundant brackets.

core/src/main/java/org/apache/calcite/sql/SqlDialectFactoryImpl.java
```diff
@@ -221,6 +221,8 @@
       return VerticaSqlDialect.DEFAULT;
     case SPARK:
       return SparkSqlDialect.DEFAULT;
+    case SNOWFLAKE:
```
NobiGo, line 224: Alphabetic order? Looks the SPARK out of position too?

NobiGo (changes requested):

core/src/main/java/org/apache/calcite/sql/SqlDialectFactoryImpl.java
JiajunBernoulli: Good advice, I've changed it.

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
JiajunBernoulli: done

NobiGo (approved):
LGTM.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -6043,6 +6043,25 @@ private void checkLiteral2(String expression, String expected) {
     sql.ok(expected);
   }
 
+  @Test void testSelectApproxCountDistinct() {
+    final String query = "select approx_count_distinct(\"product_id\") from \"product\"";
+    final String expectedAccurate = "SELECT COUNT(DISTINCT \"product_id\")\n"
```
asolimando, line 6048: Nitpick: not a native speaker here, but I think that the opposite of "approximate" is "exact" here

asolimando (approved):
Left a very minor comment which is totally optional, for the rest LGTM!

NobiGo:
@JiajunBernoulli Looks good. Please rebase the commits and I will merge.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
JiajunBernoulli: I changed the word by the way when I rebase.

JiajunBernoulli:
> @JiajunBernoulli Looks good. Please rebase the commits and I will merge.

I merged into one commit and add summary.

NobiGo (approved):
LGTM

PR #2712, opened 2022-02-04
[CALCITE-4999] ARRAY, MULTISET functions should return a collection of scalars if a sub-query returns 1 column

dssysolyatin:


rubenada:
Could you please provide a unit test with a SQL query? Probably the best option would be adding the problematic query into `big-query.iq`

rubenada:
Kindly reminder: normally the commit message should be the same as the Jira title (in case of a bug, it should normally describe the bug) https://calcite.apache.org/develop/#contributing

julianhyde:
This proposed change is not an "improvement".

core/src/main/java/org/apache/calcite/rel/core/Collect.java
```diff
@@ -132,6 +133,47 @@ public static Collect create(RelNode input,
...
   }
 
+  /**
+   * Creates a Collect.
+   *
+   * @param input          Input relational expression
+   * @param sqlKind        SqlKind
+   * @param fieldName      Name of the sole output field
+   */
+  public static Collect create(RelNode input,
```
rubenada, line 143: Should we deprecate any of the previous `create` methods to encourage people to use this new one? Or are they all still valid?

dssysolyatin: I think yes, `create(RelNode input,SqlTypeName collectionType, String fieldName)` should be deprecated.

rubenada: Ok, in that case we should not be calling it anymore.
If I am not mistaken, the PR has removed all the previously existing calls (in favor of the new `create` method), but said new method calls the deprecated one (Collect.java:174): it should call the constructor directly in that line.

dssysolyatin: Fixed

PR #2717, opened 2022-02-09
[CALCITE-5007] Upgrade H2 database version to 2.1.210

zabetak:
The 1.4.197 version suffers from https://nvd.nist.gov/vuln/detail/CVE-2018-10054.

In the project, we use H2 only for testing purposes thus the H2 binaries are not present in the runtime classpath thus CVE-2018-10054 does not pose a problem for Calcite or its users.


NobiGo (approved):
LGTM

PR #2718, opened 2022-02-10
[CALCITE-5008] Ignore synthetic and static methods in MetadataDef

rpuch:
According to https://www.jacoco.org/jacoco/trunk/doc/faq.html , Jacoco
adds a synthetic static method called $jacocoInit() during
instrumentation.

MetadataDef constructor breaks if it encounters such method on
the class given to it because it does not expect it. But it looks like
it simply should not consider synthetic methods (as well as static
methods) at all.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/MetadataDef.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/MetadataDef.java
@@ -40,8 +40,7 @@ public class MetadataDef<M extends Metadata> {
     this.metadataClass = metadataClass;
     this.handlerClass = handlerClass;
     this.methods = ImmutableList.copyOf(methods);
-    final Method[] handlerMethods = Arrays.stream(handlerClass.getDeclaredMethods())
-        .filter(m -> !m.getName().equals("getDef")).toArray(i -> new Method[i]);
+    final Method[] handlerMethods = MetadataHandler.handlerMethods(handlerClass);
 
     // Handler must have the same methods as Metadata, each method having
     // additional "subclass-of-RelNode, RelMetadataQuery" parameters.
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -114,6 +114,7 @@ dependencies {
         apiv("org.apiguardian:apiguardian-api")
         apiv("org.bouncycastle:bcpkix-jdk15on", "bouncycastle")
         apiv("org.bouncycastle:bcprov-jdk15on", "bouncycastle")
+        apiv("net.bytebuddy:byte-buddy")
         apiv("org.cassandraunit:cassandra-unit")
         apiv("org.codehaus.janino:commons-compiler", "janino")
         apiv("org.codehaus.janino:janino")
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -85,6 +85,7 @@ dependencies {
 
     testImplementation(project(":testkit"))
     testImplementation("commons-lang:commons-lang")
+    testImplementation("net.bytebuddy:byte-buddy")
     testImplementation("net.hydromatic:foodmart-queries")
     testImplementation("net.hydromatic:quidem")
     testImplementation("org.apache.calcite.avatica:avatica-server")
--- a/core/src/main/java/org/apache/calcite/rel/metadata/janino/RelMetadataHandlerGeneratorUtil.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/janino/RelMetadataHandlerGeneratorUtil.java
@@ -62,8 +62,7 @@ public class RelMetadataHandlerGeneratorUtil {
         .getName();
     final String name =
         "GeneratedMetadata_" + simpleNameForHandler(handlerClass);
-    final Method[] declaredMethods = Arrays.stream(handlerClass.getDeclaredMethods())
-        .filter(m -> !m.getName().equals("getDef")).toArray(i -> new Method[i]);
+    final Method[] declaredMethods = MetadataHandler.handlerMethods(handlerClass);
     Arrays.sort(declaredMethods, Comparator.comparing(Method::getName));
 
     final Map<MetadataHandler<?>, String> handlerToName = new LinkedHashMap<>();
--- a/core/src/test/java/org/apache/calcite/rel/metadata/MetadataDefTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/metadata/MetadataDefTest.java
@@ -16,30 +16,24 @@
  */
 package org.apache.calcite.rel.metadata;
 
-import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
-import java.util.Arrays;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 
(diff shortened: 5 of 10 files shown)


rpuch:
A test has failed.

```
FAILURE   0.1sec, org.apache.calcite.test.DruidAdapter2IT > testAggOverStringToLong()
    java.lang.IllegalStateException: Unable to create connection: connectionFactory = org.apache.calcite.test.ConnectionFactories$MapConnectionFactory@f20d5180
        at org.apache.calcite.test.CalciteAssert$AssertQuery.createConnection(CalciteAssert.java:1375)
        at org.apache.calcite.test.CalciteAssert$AssertQuery.withConnection(CalciteAssert.java:1383)
        at org.apache.calcite.test.CalciteAssert$AssertQuery.queryContains(CalciteAssert.java:1699)
        at org.apache.calcite.test.DruidAdapter2IT.testAggOverStringToLong(DruidAdapter2IT.java:3463)
        Caused by: java.sql.SQLException: No suitable driver found for jdbc:calcite:
            at java.sql.DriverManager.getConnection(DriverManager.java:689)
            at java.sql.DriverManager.getConnection(DriverManager.java:208)
            at org.apache.calcite.test.ConnectionFactories$MapConnectionFactory.createConnection(ConnectionFactories.java:115)
            at org.apache.calcite.test.CalciteAssert$AssertQuery.createConnection(CalciteAssert.java:1373)
            ... 3 more
            Next exception 1: [CIRCULAR REFERENCE SQLException]
```

It does not seem to be related to my changes. Is it possible to re-run just the failed sub-task? I did not find one, maybe I'm not looking at the correct place, or maybe I just don't have enough privileges.

core/src/main/java/org/apache/calcite/rel/metadata/MetadataDef.java
```diff
@@ -41,7 +42,10 @@ private MetadataDef(Class<M> metadataClass,
     this.handlerClass = handlerClass;
     this.methods = ImmutableList.copyOf(methods);
     final Method[] handlerMethods = Arrays.stream(handlerClass.getDeclaredMethods())
-        .filter(m -> !m.getName().equals("getDef")).toArray(i -> new Method[i]);
+        .filter(m -> !m.getName().equals("getDef"))
```
rubenada, line 45: @jacques-n it seems this filter was originally introduced via CALCITE-4929 (see [4ff5fa8](https://github.com/apache/calcite/commit/4ff5fa89c180ebc30d8fb324c2a50b0e9797b9ca) ), what do you think about the current patch?
Also, I noticed that in  CALCITE-4929 a similar filter was introduced in [RelMetadataHandlerGeneratorUtil.java](https://github.com/apache/calcite/commit/4ff5fa89c180ebc30d8fb324c2a50b0e9797b9ca#diff-b5bac3d2a905799f710aef8e5fd02dd659dc9e6e2c8e7b05ed0c34022025b8e1) , if we apply this patch, should we apply the change in both files too?

jacques-n: Yes, we should apply the same logic in both places. I was thinking that we should probably extract this to a utility method as opposed to continue duplication. As part of this, let's also add specific units tests which verify the utility method. You up for that @rpuch?

rpuch: @jacques-n sure, I'll do it. Thank you guys for your suggestions.

rpuch:
@rubenada @jacques-n I extracted the code for finding handler methods and also added tests concerning synthetic methods. Could you please take a look?

core/src/main/java/org/apache/calcite/rel/metadata/MetadataDef.java
rpuch: Done

core/src/test/java/org/apache/calcite/rel/metadata/BlankMetadataHandler.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.metadata;
+
+/**
+ * A blank {@link MetadataHandler} that is used as a base for adding a synthetic method.
+ */
+interface BlankMetadataHandler extends MetadataHandler<TestMetadata> {
```
jacques-n, line 22: Can you move the test interfaces into the test class files?

core/src/main/java/org/apache/calcite/rel/metadata/MetadataHandlerMethods.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public class MetadataHandlerMethods {
+  /**
+   * Finds handler methods defined by a {@link MetadataHandler}. Static and synthetic methods
+   * are ignored.
+   *
+   * @param handlerClass the handler class to inspect
+   * @return handler methods
+   */
+  public static Method[] handlerMethods(Class<? extends MetadataHandler<?>> handlerClass) {
```
jacques-n, line 34: How about we put this as a static method on MetadataHandler instead of a separate class?

jacques-n (commented):
Generally looks good. Would be great if you could collapse a bunch of the small test files to static inner classes/interfaces (within the test classes).

core/src/main/java/org/apache/calcite/rel/metadata/MetadataHandlerMethods.java
rpuch: Moved the method to `MetadataHandler` class

core/src/test/java/org/apache/calcite/rel/metadata/BlankMetadataHandler.java
rpuch: Moved `BlankMetadataHandler` into a test class.

`TestMetadata` and `MetadataHandlerWithStaticMethod` are used by two distinct test classes, so they cannot be 'internalized' without fusing the test classes together. If your suggestion is to still fuse them, please elaborate what test structure you would prefer to see.

rpuch:
@rubenada @jacques-n guys is anything still missing?

build.gradle.kts
```diff
@@ -893,3 +893,7 @@ allprojects {
         }
     }
 }
+java {
+    sourceCompatibility = JavaVersion.VERSION_17
```
jacques-n, line 897: We shouldn't be changing this as part of this patch. (I didn't notice it before.)

The rest looks good.

jacques-n (changes requested):

build.gradle.kts
rpuch: Oops, good catch. For some reason, it's already second time when this change was added. First time I noticed and reverted it, second time I didn't. I wonder whether it's added by an IDE (I use IntelliJ IDEA) or by some build script.

Anyway, I removed the change. Could we please proceed?

jacques-n (approved):
LGTM.

rpuch:
Thank you for your review!

jacques-n:
Thanks for the contribution @rpuch!

rpuch:
Thank you guys!

PR #2720, opened 2022-02-15
[CALCITE-5011] CassandraAdapterDataTypesTest fails with initializatio…

asolimando:
…n error

Cache Cassandra sessions based on "hostname, port, keyspace, username, password" information.


asolimando:
I plan to merge in around 48 hours if nobody objects

PR #2721, opened 2022-02-16
[CALCITE-4912] Confusing javadoc of RexSimplify.simplify

NobiGo:


core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -204,8 +204,8 @@ public RexNode simplifyPreservingType(RexNode e, RexUnknownAs unknownAs,
    *
    * <p>In particular:</p>
    * <ul>
-   * <li>{@code simplify(x = 1 AND y = 2 AND NOT x = 1)}
```
wojustme, line 207: Origin doc maybe means reduce condition of x.
Why you change here?

NobiGo: It is a wrong simplify.

wojustme: @NobiGo Yes, you are right.
My description may not be very accurate.
I mean that this expression should return FALSE
Should we just change return's doc?

NobiGo: No. What the expression should be returned is decided by how we treat the UNKNOW. For example:
```
x=1 AND NOT x=1

If UNKNOW as FALSE then return FALSE;
If UNKNOW as TRUE  then return FALSE;
If UNKNOW as UNKNOW then return UNKNOW.(Then (x = 1 AND y = 2 AND NOT x = 1) result will not be a definite value)
```

wojustme: Yes, It's my negligence
LTGM~

PR #2722, opened 2022-02-16
Make sure the test kit can be used externally

mkou:


julianhyde:
@mkou,

Can you please log a JIRA case and use the case number in the commit message.

Move `FixtureTest` from `org.apache.calcite.test` to some other package (keeping within `testkit` module). This will keep us honest in future.

This PR looks like 30 commits because you used a merge. Can you rebase, so it looks like one small commit? Then force-push and ask for review again.

julianhyde:
> This PR looks like 30 commits because you used a merge

Oops, not true. Someone force-pushed to master. Not your fault. Please ignore.

PR #2724, opened 2022-02-19
[CACLITE-5013] Retain parentheses if a setop has an LIMIT clause

JiajunBernoulli:
In standard SQL, the operand of union should not have limit  or order by.So parse will fail in SqlParserTetest#testLimitUnion and SqlParserTetest#OrderUnion.
However, when users use parentheses, most engines allow it. We should retain the parentheses written by users.

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -3374,6 +3374,84 @@ class RelToSqlConverterTest {
     sql(query).ok(expected);
   }
 
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5013">[CALCITE-5013]
+   * Unparse SqlSetOperator should be retained parentheses
+   * when the operand has limit or offset</a>. */
+  @Test void testSetOpRetainParentheses() {
+    // Parentheses will be discarded, because semantics not be affected.
+    final String discardedParenthesesQuery = "SELECT \"product_id\" FROM \"product\""
+        + "UNION ALL\n"
+        + "(SELECT \"product_id\" FROM \"product\" WHERE \"product_id\" > 10)\n"
+        + "INTERSECT ALL\n"
+        + "(SELECT \"product_id\" FROM \"product\" )";
+    final String discardedParenthesesRes = "SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\"\n"
+        + "UNION ALL\n"
+        + "SELECT *\n"
+        + "FROM (SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\"\n"
+        + "WHERE \"product_id\" > 10\n"
+        + "INTERSECT ALL\n"
+        + "SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\")";
+    sql(discardedParenthesesQuery).ok(discardedParenthesesRes);
+
+    // Parentheses will be retained because sub-query has LIMIT or OFFSET.
+    // If parentheses are discarded the semantics of parsing will be affected.
+    final String allSetOpQuery = "SELECT \"product_id\" FROM \"product\""
+        + "UNION ALL\n"
+        + "(SELECT \"product_id\" FROM \"product\" LIMIT 10)\n"
+        + "INTERSECT ALL\n"
+        + "(SELECT \"product_id\" FROM \"product\" OFFSET 10)\n"
+        + "EXCEPT ALL\n"
+        + "(SELECT \"product_id\" FROM \"product\" LIMIT 5 OFFSET 5)";
+    final String allSetOpRes = "SELECT *\n"
+        + "FROM (SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\"\n"
+        + "UNION ALL\n"
+        + "SELECT *\n"
+        + "FROM ((SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\"\n"
+        + "FETCH NEXT 10 ROWS ONLY)\n"
+        + "INTERSECT ALL\n"
+        + "(SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\"\n"
+        + "OFFSET 10 ROWS)))\n"
+        + "EXCEPT ALL\n"
+        + "(SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\"\n"
+        + "OFFSET 5 ROWS\n"
+        + "FETCH NEXT 5 ROWS ONLY)";
+    sql(allSetOpQuery).ok(allSetOpRes);
+
+    // After the config is enabled, order by will be retained, so parentheses are required.
+    final String retainOrderQuery = "SELECT \"product_id\" FROM \"product\""
+        + "UNION ALL\n"
+        + "(SELECT \"product_id\" FROM \"product\" ORDER BY \"product_id\")";
+    final String retainOrderResult = "SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\"\n"
+        + "UNION ALL\n"
+        + "(SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\"\n"
+        + "ORDER BY \"product_id\")";
+    sql(retainOrderQuery).withConfig(c -> c.withRemoveSortInSubQuery(false)).ok(retainOrderResult);
+
+    // Parentheses are required to keep ORDER and LIMIT on the sub-query.
+    final String retainLimitQuery = "SELECT \"product_id\" FROM \"product\""
+        + "UNION ALL\n"
+        + "(SELECT \"product_id\" FROM \"product\" ORDER BY \"product_id\" LIMIT 2)";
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/SqlUtil.java
```diff
@@ -417,24 +418,52 @@ public static void unparseSqlIdentifierSyntax(
     writer.endList(frame);
   }
 
+  private static void unparseSetOpSyntax(
```
julianhyde, line 421: It's not obvious what this method does. Can you add brief javadoc.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3301,6 +3301,85 @@ private void checkHavingAliasSameAsColumn(boolean upperAlias) {
     sql(query).ok(expected);
   }
 
+  @Test void testSetOpLimitShouldRetainParentheses() {
```
julianhyde, line 3304: capitalize 'limit' and other sql keywords

julianhyde, line 3304: add javadoc referencing jira

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3301,6 +3301,85 @@ private void checkHavingAliasSameAsColumn(boolean upperAlias) {
...
+        + "(SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\"\n"
+        + "ORDER BY \"product_id\"\n"
+        + "OFFSET 5 ROWS\n"
+        + "FETCH NEXT 5 ROWS ONLY)) AS \"t6\"";
+    sql(unionOffsetQuery).ok(unionOffsetRes);
+  }
+
+  @Test void testSetOpOrderMaybeRetainParentheses() {
+    // Order by without limit/offset will be removed, so parentheses are not required.
```
julianhyde, line 3351: this test is testing RelToSql, not SqlToRel. So I would remove this case (the first half of this method).

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
julianhyde, line 3304: There are only tests for UNION currently. Add a test for INTERSECT or EXCEPT, or perhaps a combination of set-ops.

core/src/main/java/org/apache/calcite/sql/SqlUtil.java
```diff
@@ -417,24 +418,52 @@ public static void unparseSqlIdentifierSyntax(
...
+      unparseSetOpSyntax(operand0, writer, leftPrec, operator.getLeftPrec());
+    } else {
+      operand0.unparse(writer, leftPrec, operator.getLeftPrec());
+    }
     final boolean needsSpace = operator.needsSpace();
     writer.setNeedWhitespace(needsSpace);
     writer.sep(operator.getName());
     writer.setNeedWhitespace(needsSpace);
-    call.operand(1).unparse(writer, operator.getRightPrec(), rightPrec);
+    SqlNode operand1 = call.operand(1);
```
julianhyde, line 461: Is the logic of the `if` statement below repeated inside the `unparseSetOpSyntax` function?

core/src/main/java/org/apache/calcite/sql/SqlUtil.java
julianhyde, line 421: why is the function named `unparseSetOpSyntax`? The `Syntax` seems redundant, in which case it would be `unparseSetOp`. But then I see that it's unparsing a `SqlSelect`, not a set op. So I'm thoroughly confused.

JiajunBernoulli: Ok, I've added a doc to explain it.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
JiajunBernoulli: I changed it, but I'm a little confused about how to choose case. 
In `testUnionWrappedInASelect` and `testAliasOnStarNoExcessQuotes`, sql keywords is lowercase.

JiajunBernoulli: Done, and I renamed this function to `testSetOpRetainParentheses`.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
JiajunBernoulli: Ok, I've deleted this test case.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
JiajunBernoulli: There has been a case of UNION and INTERSECT before, I've added a combination of set-ops.

core/src/main/java/org/apache/calcite/sql/SqlUtil.java
JiajunBernoulli: I moved it to `unparseBinaryOperand`.

core/src/main/java/org/apache/calcite/sql/SqlUtil.java
JiajunBernoulli: Sorry, the previous naming was really inappropriate. 
I renamed it to `unparseBinaryOperand`(Write parentheses for special operands in it)

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3301,6 +3301,71 @@ private void checkHavingAliasSameAsColumn(boolean upperAlias) {
...
+        + "FROM (SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\"\n"
+        + "WHERE \"product_id\" > 10\n"
+        + "INTERSECT ALL\n"
+        + "SELECT \"product_id\"\n"
+        + "FROM \"foodmart\".\"product\")";
+    sql(discardedParenthesesQuery).ok(discardedParenthesesRes);
+
+    // Parentheses will be retained because subquery has limit or offset.
+    // If parentheses  are discarded the semantics of parsing will be affected.
```
NobiGo, line 3329: redundant blank space after the parentheses

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3301,6 +3301,71 @@ private void checkHavingAliasSameAsColumn(boolean upperAlias) {
     sql(query).ok(expected);
   }
 
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5013">[CALCITE-5013]
+   * Unparse SqlSetOperator should be retained parentheses
+   * when the operand has limit or offset</a>. */
+  @Test void testSetOpRetainParentheses() {
```
NobiGo, line 3309: I'm suspect whether we should remove the parentheses?
If the SQL is :
```
"SELECT \"product_id\" FROM \"product\""
        + "UNION ALL\n"
        + "(SELECT \"product_id\" FROM \"product\")\n"
        + "INTERSECT ALL\n"
        + "(SELECT \"product_id\" FROM \"product\" WHERE \"product_id\" > 10)"
```
```
SELECT \"product_id\"
FROM \"foodmart\".\"product\"
UNION ALL
SELECT *
FROM (SELECT \"product_id\"
FROM \"foodmart\".\"product\"
INTERSECT ALL
SELECT \"product_id\"
FROM \"foodmart\".\"product\"
WHERE \"product_id\" > 10)
```

I think it has become unclear? So maybe we should stay the parentheses?

NobiGo (changes requested):

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
JiajunBernoulli: Do you mean to we should retain it when there is a `where` condition? Or we always retain parentheses?

julianhyde: I disagree that we should add parentheses for aesthetics. See my comment in https://issues.apache.org/jira/browse/CALCITE-5013 and let's discuss further there.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
JiajunBernoulli: done

NobiGo (approved):

jbalint:
@JiajunBernoulli looks good. can you please fix the merge conflict?

JiajunBernoulli:
> @JiajunBernoulli looks good. can you please fix the merge conflict?

Ok. Thanks for your review, would you please merge it?

PR #2727, opened 2022-02-21
[CALCITE-1794] Expressions with numeric comparisons are not simplified when CAST is present

NobiGo:


asolimando:
Hi @NobiGo, in the associated Jira ticket you suggested a re-ordering of predicates, from some test changes it seems that this is indeed happening now, but I see no code change related to that. Is the addition of the case for `CAST` causing that somehow?

NobiGo:
> Hi @NobiGo, in the associated Jira ticket you suggested a re-ordering of predicates, from some test changes it seems that this is indeed happening now, but I see no code change related to that. Is the addition of the case for `CAST` causing that somehow?

@asolimando Yes. It looks need to reorder it, but as @julianhyde is concerned (It Will introduce plan diffs and make this more expensive). So When I try to find another way to resolve it, I found after the RexNode is optimized, The RexNode is converted to a Search argument, Then this case can be simplified to a point.

julianhyde:
This PR looks good. @NobiGo, Can you add one or two more comments that the problem is the CAST. The .iq tests should point out the the simplified plan has only 'deptno = 25' and has dropped the 'deptno <> 20' condition.

When I am squashing these commits I'll probably use 'Expressions with numeric comparisons are not simplified when CAST is present' as the commit message.

PR #2729, opened 2022-02-23
[CALCITE-5019] Avoid multiple scans when table is ProjectableFilterableTable

IceMimosa:
```sql
select a from ProjectableFilterableTable where b = 'xx'
```

The ProjectableFilterableTable will scan two times when projections and filters are different columns.

### DIFF
--- a/core/src/main/java/org/apache/calcite/interpreter/TableScanNode.java
+++ b/core/src/main/java/org/apache/calcite/interpreter/TableScanNode.java
@@ -200,8 +200,6 @@ public class TableScanNode implements Node {
       } else {
         projectInts = projects.toIntArray();
       }
-      final Enumerable<@Nullable Object[]> enumerable1 =
-          pfTable.scan(root, mutableFilters, projectInts);
       for (RexNode filter : mutableFilters) {
         if (!filters.contains(filter)) {
           throw RESOURCE.filterableTableInventedFilter(filter.toString())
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/interpreter/TableScanNode.java
```diff
@@ -227,6 +225,8 @@ private static TableScanNode createProjectableFilterable(Compiler compiler,
           continue;
```
IceMimosa, line 225: If `changeCount > 0`, the scan will trigger again.

core/src/test/java/org/apache/calcite/test/ScannableTableTest.java
```diff
@@ -155,7 +155,7 @@
             "j=Paul");
     // Only 2 rows came out of the table. If the value is 4, it means that the
     // planner did not pass the filter down.
-    assertThat(buf.toString(), is("returnCount=2, filter=<0, 4>, projects=[1]"));
+    assertThat(buf.toString(), is("returnCount=2, filter=<0, 4>, projects=[1, 0]"));
```
rubenada, line 158: Is it expected to have this side effect (and the one in line 191)?

IceMimosa: @rubenada Yes, it's expected. The test ignored this case.

JiajunBernoulli: @IceMimosa Do you know what projetcs means?

IceMimosa: @JiajunBernoulli This test wants to pushdown filter to table scan, but it's not suitable here, it should use RelRule to do this in optimize phase. `TableScan's projects` and `ProjectableFilterableTable's projects` are not the same, The test here is `select j where i`, in physical execution, the ProjectableFilterableTable's projects must contain `j` and `i` if you don't change the AST.

rubenada:
@IceMimosa the change LGTM, could you please squash commits into a single one?
May I suggest as commit message something along the lines "[CALCITE-5019] Avoid multiple scans when table is ProjectableFilterableTable and projections and filters act on different columns"

PR #2730, opened 2022-02-24
Modify parser to allow subqueries in JSON "VALUE"

GavinRay97:
Allows to execute expressions like the below:

```sql
SELECT JSON_OBJECT(
  KEY 'foo'
  VALUE (SELECT 'bar')
)
```

This feature is useful for returning graph-like responses (get a user, their friends, and their friends pets for example) since it allows arbitrary depth nesting in queries, but with a key unlike multisets/arrays.

### DIFF
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -6287,7 +6287,7 @@ List<SqlNode> JsonNameAndValue() :
             }
         }
     )
-    e = Expression(ExprContext.ACCEPT_NON_QUERY) {
+    e = Expression(ExprContext.ACCEPT_ALL) {
         list.add(e);
     }
     {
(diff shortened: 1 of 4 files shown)


GavinRay97:
[Second commit fixes 4989](https://issues.apache.org/jira/browse/CALCITE-4989)

GavinRay97:
This is trickier than I thought actually, since each JSON call invokes `jsonize()`, if you have nested JSON functions they will be parsed `n` number of times.

Each parse will add another set of escapes. The only proper way to handle this is to parse only once, at the top-level object, but this isn't currently possibly as far as I know =/

GavinRay97:
I've reverted the changes in formatting, based on the discussion in the mailing list here:
- https://lists.apache.org/thread/5t5nmj05ocz3z40gyfvj2gw98r1b6m54

It turns out that the trick is to use `FORMAT JSON` on every JSON expression, it prevents the issue with escaping.

PR is ready for review now

PR #2731, opened 2022-02-25
[CALCITE-5023] JOIN whose inputs projecting same field names shouldn'…

hannerwang:
…t generate SELECT STAR when converting rel to sql

create table users(id int);
create table depts(id int);
select a.id, b.id as id0 from users1 a, depts1 b
after converting the sql to rel then back rel to sql, we get 
SELECT *
FROM `default`.`users1`
CROSS JOIN `default`.`depts1`

but it can't be used in CTAS in some dialects such as mysql, spark sql.

This change will affect other tests and I have modified their SELECT \* to SELECT fields.


xiedeyantu:
Same as #4405

PR #2734, opened 2022-02-27
[CALCITE-5025] Update commons-io version

sdreynolds:
In commons-io:commons-io < 2.7, it is possible to traverse into the
parent directory:
`CVE-2021-29425`

This changes the version 2.11.0 which was released on
2021-07-10. There have been a number of commits to this project in
recent months and seems like a good time to update to a modern relase
2.4 was released on 2012-06-12


zabetak (approved):
LGTM will merge in a few minutes.

PR #2735, opened 2022-03-01
[CALCITE-5021] Double JOIN is created for NOT IN when IN-list that the values all non-nullable is converted to Values

NobiGo:


core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
```diff
@@ -294,6 +294,33 @@ public static boolean containsNullable(RelDataType type) {
     return false;
   }
 
+  /**
+   * Recreates a RelDataType with the type keep same as the param sourceRelDataType and
+   * nullability keep same as the param targetRelDataType.
```
asolimando, line 299: What about rephrasing the comment as follows:
```suggestion
   * Creates a RelDataType having the same type of the sourceRelDataType,
   * and the same nullability as the targetRelDataType.
```

core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
```diff
@@ -294,6 +294,33 @@ public static boolean containsNullable(RelDataType type) {
...
+   */
+  public static RelDataType keepNullableSame(RelDataType sourceRelDataType,
```
asolimando, line 301: Picking a good name for this method seems hard, what about `keepSourceTypeAndTargetNullability`?

NobiGo: Looks good, I will fix it.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
NobiGo: Looks good, I will fix it.

asolimando (approved):
LGTM, left few minor comments on javadoc and method naming

asolimando:
@NobiGo please make sure to have a green CI run before merging, the travis failure does not seem related but I'd rather see it running.

PR #2745, opened 2022-03-15
[CALCITE-4512] GROUP BY expression with argument name same with SELEC…

hannerwang:
…T field and alias causes validation error

This PR restricts the alias/ordinal expanding GROUP BY expr to identifier or CUBE/ROLLUP/GROUPING SETS top level expression.
For example:
select replace(name, 'a', 'b') as name from users group by replace(name, 'a', 'b')
The name in GROUP BY clause will not be replaced by replace(name, 'a', 'b'), it will be resolved as users.name.

SELECT empno + deptno AS x, empno - deptno AS y FROM emp GROUP BY ROLLUP(x, y, empno) 
The x, y in GROUP BY clause will be replaced by empno + deptno, empno - deptno.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
@@ -6601,15 +6601,17 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
       this.select = select;
       this.root = root;
       this.havingExpr = havingExpr;
-      addExpandableExpressions();
+      if (!havingExpr) {
+        addExpandableExpressions();
+      }
     }
 
     @Override public @Nullable SqlNode visit(SqlIdentifier id) {
       if (id.isSimple()
-          && aliasOrdinalExpandSet.contains(id)
           && (havingExpr
               ? validator.config().conformance().isHavingAlias()
-              : validator.config().conformance().isGroupByAlias())) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -6590,20 +6593,25 @@ static class ExtendedExpander extends Expander {
...
+        addExpandableExpressions();
+      }
     }
 
     @Override public @Nullable SqlNode visit(SqlIdentifier id) {
       if (id.isSimple()
           && (havingExpr
               ? validator.config().conformance().isHavingAlias()
-              : validator.config().conformance().isGroupByAlias())) {
+              : (validator.config().conformance().isGroupByAlias()
```
jaystarshot, line 6613: Why not create a new Expander which extends Expander and use that in validateGroupBy if flag is set?

jbalint: @hannerwang are you able to respond and get this PR merged?

hannerwang: There may be a lot of duplicate codes and changes so that review will be difficult.

hannerwang:
This pr is out of date, will be solved at [pr-3929](https://github.com/apache/calcite/pull/3929)

PR #2747, opened 2022-03-18
[CALCITE-5051] Test for reproducing UNION query plan issue

zgramana:
Per email group, this captures a change in behavior from 1.24 to 1.29.

In 1.24, `UNION` and `UNION ALL` query plans were identical, with the
exception of the `EnumerableUnion.all` value.

In 1.29, the reproduced plan eliminates some `LogicalProject` nodes,
resulting in the inability to pushdown projections of very
wide tables, causing performance to regress sometimes rather
a lot

### DIFF
--- a/core/src/test/java/org/apache/calcite/tools/PlannerTest.java
+++ b/core/src/test/java/org/apache/calcite/tools/PlannerTest.java
@@ -18,16 +18,22 @@ package org.apache.calcite.tools;
 
 import org.apache.calcite.adapter.enumerable.EnumerableConvention;
 import org.apache.calcite.adapter.enumerable.EnumerableProject;
+import org.apache.calcite.adapter.enumerable.EnumerableRel;
+import org.apache.calcite.adapter.enumerable.EnumerableRelImplementor;
 import org.apache.calcite.adapter.enumerable.EnumerableRules;
 import org.apache.calcite.adapter.enumerable.EnumerableTableScan;
+import org.apache.calcite.adapter.enumerable.EnumerableUnion;
 import org.apache.calcite.adapter.java.ReflectiveSchema;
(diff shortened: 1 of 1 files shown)


rubenada:
@zgramana  I have verified the unit test in this PR, and it runs fine with latest Calcite.
Therefore, as I suspected, it seems the bug has been fixed as a side effect of [CALCITE-5061](https://issues.apache.org/jira/browse/CALCITE-5061), so I'll close the Jira [CALCITE-5051](https://issues.apache.org/jira/browse/CALCITE-5051)

zgramana:
> @zgramana I have verified the unit test in this PR, and it runs fine with latest Calcite.
> Therefore, as I suspected, it seems the bug has been fixed as a side effect of [CALCITE-5061](https://issues.apache.org/jira/browse/CALCITE-5061), so I'll close the Jira [CALCITE-5051](https://issues.apache.org/jira/browse/CALCITE-5051).
> Shall we close this PR?

That's great to hear, thanks. This PR might be worth merging to catch any potential regressions.

rubenada:
Thanks for the feedback @zgramana 

> This PR might be worth merging to catch any potential regressions.

IMO there's no need for that, the PR that fixed [CALCITE-5061](https://issues.apache.org/jira/browse/CALCITE-5061) (and as a side effect also [CALCITE-5051](https://issues.apache.org/jira/browse/CALCITE-5051) ) included a test that is very similar to the one proposed in this PR: https://github.com/apache/calcite/blob/7de37aefaaa72959a0145053319fae2956fbb9f1/core/src/test/java/org/apache/calcite/sql2rel/RelFieldTrimmerTest.java#L520

zgramana:
Per @rubenada's [comment above](https://github.com/apache/calcite/pull/2747#issuecomment-1216902172), closing this as it is has been superseded.

PR #2750, opened 2022-03-22
[CALCITE-5052] Allow resources created with a JAR protocol (Bazel)

mkou:


core/src/test/java/org/apache/calcite/util/SourceTest.java
```diff
@@ -121,6 +131,54 @@ void testAbsoluteFileToUrl(String path, String expectedUrl) throws URISyntaxExce
         () -> "Sources.of(Sources.of(file(" + path + ").absolutePath).url()).file().getPath()");
   }
 
+  /**
+   * Resources created with the JAR protocol should also work.
```
tanclary, line 135: nit could reference JIRA here and in below test.

core/src/test/java/org/apache/calcite/util/SourceTest.java
```diff
@@ -121,6 +131,54 @@ void testAbsoluteFileToUrl(String path, String expectedUrl) throws URISyntaxExce
...
+    assertThat("Sources.of(file " + jarPath + ").url()).file().getPath()",
+        slashify(source.file().getPath()),
+        is("sources!/abcdef.txt"));
+
+  }
+
+  /** Tests {@link Sources#of(URL)} with code similar to
+   * {@code DiffRepository.Key#toRepo()}. */
+  @Test void testJarFileUrlWrite() {
+    final URL refFile = FoodmartQuerySet.class.getResource("/queries.json");
```
tanclary, line 154: To be honest I'm a bit confused at what all is going on here. Would you mind either:
- Extracting some of the setup into a helper (maybe easier testing is possible since this PR was opened).
- Breaking this into multiple tests
- Adding more comments

If it's all common sense and I'm OOTL then disregard.

testkit/src/main/java/org/apache/calcite/test/DiffRepository.java
```diff
@@ -895,7 +898,17 @@ private static class Key {
     DiffRepository toRepo() {
       final URL refFile = findFile(clazz, ".xml");
       final String refFilePath = Sources.of(refFile).file().getAbsolutePath();
-      final String logFilePath = refFilePath.replace(".xml", "_actual.xml");
+      final String logFilePath;
+      if (StringUtils.containsIgnoreCase(refFilePath, ".jar!")) {
```
tanclary, line 902: I notice this block is identical to the one above, can we put it somewhere helpful and refactor

tanclary (commented):
Probably worth checking how much the test fixtures have changed since this was last update (maybe some or maybe none at all)

core/src/test/java/org/apache/calcite/util/SourceTest.java
mkou: Crap it is so long ago I don't remember why I did that to test the diff repo

core/src/test/java/org/apache/calcite/util/SourceTest.java
```diff
@@ -40,7 +42,7 @@
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.hasToString;
+import static org.hamcrest.Matchers.*;
```
tanclary, line 45: I think Calcite uses explicit imports rather than .*; this is configurable in intellij settings 

https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports

core/src/test/java/org/apache/calcite/util/SourceTest.java
```diff
@@ -118,6 +120,22 @@ void testAbsoluteFileToUrl(String path, String expectedUrl) throws URISyntaxExce
         () -> "Sources.of(Sources.of(file(" + path + ").absolutePath).url()).file().getPath()");
   }
 
+  /**
+   * [CALCITE-5052] Resources created with the JAR protocol should also work.
```
tanclary, line 124: if youre going to link a jira case you can format it like this example: https://github.com/apache/calcite/blob/main/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java#L277

testkit/src/main/java/org/apache/calcite/test/DiffRepository.java
```diff
@@ -877,7 +880,7 @@ public static DiffRepository lookup(Class<?> clazz,
     return REPOSITORY_CACHE.getUnchecked(key);
   }
 
-  /**
+    /**
```
tanclary, line 883: mistake?

testkit/src/main/java/org/apache/calcite/test/DiffRepository.java
```diff
@@ -900,18 +903,21 @@ String filter(
...
   /** Cache key. */
-  private static class Key {
+  protected static class Key {
     private final Class<?> clazz;
     private final DiffRepository baseRepository;
     private final Filter filter;
     private final int indent;
-
+    private URL refFile;
+    String refFilePath;
```
tanclary, line 912: To be honest I like the newline after all the variables and before the `Key() {}`

testkit/src/test/java/org/apache/calcite/test/DiffRepositoryTest.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+}
+public class DiffRepositoryTest {
+
+  /**
+   * Test that a DiffRepository from a file in a JAR
+   * would create a log file outside of the JAR.
+   **/
+  @Test void testLogFilePathInJar() {
+    final DiffRepository diffRepo = JarStubDiffRepository.lookup(FixtureTest.class);
```
tanclary, line 74: Looks so much better now

testkit/src/test/java/org/apache/calcite/test/DiffRepositoryTest.java
```diff
@@ -0,0 +1,77 @@
...
+   * @param filter         Filter or null
+   * @param indent         Indentation of XML file
+   */
+  private JarStubDiffRepository(URL refFile, File logFile,
+      @Nullable DiffRepository baseRepository, Filter filter, int indent) {
+    super(refFile, logFile, baseRepository, filter, indent);
+  }
+
+  public static DiffRepository lookup(Class<?> clazz) {
+    final JarKey key = new JarKey(clazz, null, null, 2);
```
tanclary, line 54: should these have comments // be spaced out ? I think the actual logic looks fine !

tanclary (changes requested):

PR #2755, opened 2022-03-26
[CALCITE-4989] Nested JSON_OBJECT creation does not produce proper json

libenchao:
This pr makes nested json creation functions produce correct results. Before this pr, these json creation functions (json_object/json_array/json_objectagg/json_arrayagg) takes the output of inner json functions' result as raw string.


XuQianJin-Stars (approved):
LGTM

GavinRay97:
Thank you so much for this @libenchao!

PR #2756, opened 2022-03-27
[CALCITE-5032] RelOptUtil#splitJoinCondition returns wrong when there is no equal condition

libenchao:
Fix the bug for `RelOptUtil#splitJoinCondition`:
```Java
  public static RexNode splitJoinCondition(
      List<RelDataTypeField> sysFieldList,
      List<RelNode> inputs,
      RexNode condition,
      List<List<RexNode>> joinKeys,
      @Nullable List<Integer> filterNulls,
      @Nullable List<SqlOperator> rangeOp) {
```
 for following case:
```SQL
select
 *
from emp left join dept
on emp.empno > 0
```
The expected result is to make `(empno > 0) = true` as the equal join key.

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
@@ -1387,41 +1387,6 @@ public abstract class RelOptUtil {
         }
       }
 
-      if ((rangeOp == null)
-          && ((leftKey == null) || (rightKey == null))) {
-        // no equality join keys found yet:
-        // try transforming the condition to
-        // equality "join" conditions, e.g.
-        //     f(LHS) > 0 ===> ( f(LHS) > 0 ) = TRUE,
-        // and make the RHS produce TRUE, but only if we're strictly
-        // looking for equi-joins
-        final ImmutableBitSet projRefs = InputFinder.bits(condition);
-        leftKey = null;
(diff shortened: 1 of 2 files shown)


hsyuan (commented):
It doesn't make sense to treat `empno > 0` as equal condition and generate equal join keys.

libenchao:
@hsyuan Thanks for your review.
I also think this is a rare case, that's why I added the param `strictEqualJoin` to control it. I'm also fine to just remove this logic from `splitJoinCondition` if you agree with this.

hsyuan (approved):
Hi @libenchao, thanks for your quick response. The change looks good to me.
+1

.ratignore
```diff
@@ -39,6 +39,7 @@ site/_layouts/news.html
 site/_layouts/news_item.html
 site/_layouts/page.html
 site/_sass/**
+site/Gemfile.lock
```
hsyuan, line 42: I believe this change has been in master branch?

libenchao: Sure, I've removed it.

libenchao:
@hsyuan Is there anything I need to do before this get merged, do I need to squash my commits?

hsyuan:
Squash the commits, and we are good to go.

libenchao:
"Druid Tests" failed due to timeout, it is not caused by this change.

PR #2757, opened 2022-03-29
[CALCITE-4401] SqlJoin toString throws RuntimeException

abhishek-das-gupta:
Add a "SELECT *" wrap over SqlJoin in SqlJoin#toSqlString

Resolves the RuntimeException seen in SqlJoin#toString

Override SqlNode#toSqlString in SqlJoin

Reason: SqlJoin#unparse method calls SqlPrettyWriter#sep with frame being NULL

This change is less fragile than to mutate internal frame state


core/src/test/java/org/apache/calcite/sql/test/SqlPrettyWriterTest.java
```diff
@@ -438,6 +441,30 @@ private SqlPrettyWriterFixture tableDotStar() {
...
+    final String expectedJoinString = "SELECT *\r\n"
+        + "FROM (SELECT `C_CUSTKEY`, `REGION_NAME`\r\n"
+        + "FROM `TPCH`.`OUT_TPCH_VW__CUSTOMER`) AS `T`\r\n"
+        + "INNER JOIN (SELECT `O_CUSTKEY`, `O_TOTALPRICE`\r\n"
+        + "FROM `TPCH`.`OUT_TPCH_VW__ORDERS`) AS `T0` ON `T`.`C_CUSTKEY` = `T0`.`O_CUSTKEY`";
+
+    assert join != null;
+    Assertions.assertEquals(join.toString(), expectedJoinString);
+  }
+
```
julianhyde, line 467: Please use the fixture, to be consistent with other methods in this test.

julianhyde: Also, move your test next to other related tests.

(Contributors seem to think it is 'polite' to add code to the end of a file. It is not. If everyone does that, it creates merge conflicts. Better to put the code in the most logical place, next to related code.)

abhishek-das-gupta: I've moved the test case to most logical place (among join tests). 

I first tried to use fixture only but the issue I faced was how to check toString() on JOIN ? The SQLs containing JOIN when parsed give back a SqlSelect. Hence  if I use `sql(sql).check()`  it would assert on the SqlSelect#toString instead of SqlJoin#toString. That is already working on current master since SqlSelect#unparse will start the frame for SqlJoin to conusme in the fromList condition.    

It is only when explicit toString on Join is called then frame is null and we get RuntimeException.

I want to mimick the steps present in the jira description to  in the testcase. 

I also think this testcase does look not consistant with other test present. 

Is there an api present in fixture that I'm missing ? If so, can you please guide me ? If not,  should I introduce an api in fixture? Otherwise will `sql(sql).check()` suffice ?

Appreciate the review!

julianhyde: I think you will need to extend the fixture.

abhishek-das-gupta: Done. Pls check

PR #2763, opened 2022-04-05
[CALCITE-5048] Optimizer support dynamic parameter in LIMIT/OFFSET

JiajunBernoulli:
I searched the code of `RexLiteral#intValue` in core module. Found that four class failed when LIMIT/OFFSET parameter is Dynamic. 
- RelMdMaxRowCount
- RelMdMinRowCount
- RelMdUtil
- RelDecorrelator

I fixed them, and added corresponding unit tests.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
@@ -477,13 +477,10 @@ public class RelDecorrelator implements ReflectiveVisitor {
     RelCollation oldCollation = rel.getCollation();
     RelCollation newCollation = RexUtil.apply(mapping, oldCollation);
 
-    final int offset = rel.offset == null ? -1 : RexLiteral.intValue(rel.offset);
-    final int fetch = rel.fetch == null ? -1 : RexLiteral.intValue(rel.fetch);
-
     final RelNode newSort = relBuilder
-            .push(newInput)
-            .sortLimit(offset, fetch, relBuilder.fields(newCollation))
-            .build();
+        .push(newInput)
+        .sortLimit(rel.offset, rel.fetch, relBuilder.fields(newCollation))
+        .build();
 
     // Sort does not change input ordering
     return register(rel, newSort, frame.oldToNewOutputs, frame.corDefOutputs);
--- a/core/src/main/java/org/apache/calcite/tools/RelBuilder.java
+++ b/core/src/main/java/org/apache/calcite/tools/RelBuilder.java
@@ -70,6 +70,7 @@ import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexCallBinding;
 import org.apache.calcite.rex.RexCorrelVariable;
+import org.apache.calcite.rex.RexDynamicParam;
 import org.apache.calcite.rex.RexExecutor;
 import org.apache.calcite.rex.RexFieldCollation;
 import org.apache.calcite.rex.RexInputRef;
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -3196,12 +3197,23 @@ public RelBuilder sort(RelCollation collation) {
    */
   public RelBuilder sortLimit(int offset, int fetch,
       Iterable<? extends RexNode> nodes) {
+    return sortLimit(offset, fetch, nodes, null, null);
+  }
+
+  public RelBuilder sortLimit(int offset, int fetch,
```
rubenada, line 3203: Personally, I find this API confusing, because it could lead to situations where e.g. a valid offset and a valid dynamicOffset are passed at the same time.
What about keeping the "old" sortLimit method and adding a new one with RexNode parameters (with the corresponding check to verify that they are literals or parameters or null), i.e. something like:
```
  /** Creates a {@link Sort} ... */
  public RelBuilder sortLimit(int offset, int fetch,
      Iterable<? extends RexNode> nodes) {
    final @Nullable RexNode offsetNode = offset <= 0 ? null : literal(offset);
    final @Nullable RexNode fetchNode = fetch < 0 ? null : literal(fetch);
    if (offsetNode == null && fetch == 0 && config.simplifyLimit()) {
      return empty();
    }
    return sortLimit(offsetNode, fetchNode, nodes);
  }

  /**
   * TODO javadoc
   */
  public RelBuilder sortLimit(@Nullable RexNode offsetNode, @Nullable RexNode fetchNode,
      Iterable<? extends RexNode> nodes) {
    if (offsetNode != null) {
      if (!(offsetNode instanceof RexLiteral || offsetNode instanceof RexDynamicParam)) {
        throw new IllegalArgumentException("...");
      }
    }
    if (fetchNode != null) {
      if (!(fetchNode instanceof RexLiteral || fetchNode instanceof RexDynamicParam)) {
        throw new IllegalArgumentException("...");
      }
    }

    final Registrar registrar = new Registrar(fields(), ImmutableList.of());
    ...
```
I think this approach would also simplify the new code in `RelDecorrelator`, where this new method could be used like: `sortLimit(rel.offset, rel.fetch, relBuilder.fields(newCollation))`
WDYT?

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -476,14 +477,29 @@ protected RexNode removeCorrelationExpr(
 
     RelCollation oldCollation = rel.getCollation();
     RelCollation newCollation = RexUtil.apply(mapping, oldCollation);
-
-    final int offset = rel.offset == null ? -1 : RexLiteral.intValue(rel.offset);
-    final int fetch = rel.fetch == null ? -1 : RexLiteral.intValue(rel.fetch);
+    int offset = -1;
```
rubenada, line 480: See my comment in RelBuilder, which could help simplify this code.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
JiajunBernoulli: Thank you for your advice, I have finished the refactor.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
JiajunBernoulli: Good advice.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -3196,12 +3197,40 @@ public RelBuilder sort(RelCollation collation) {
    */
   public RelBuilder sortLimit(int offset, int fetch,
       Iterable<? extends RexNode> nodes) {
+    final @Nullable RexNode offsetNode = offset <= 0 ? null : literal(offset);
+    final @Nullable RexNode fetchNode = fetch < 0 ? null : literal(fetch);
+    if (offsetNode == null && fetch == 0 && config.simplifyLimit()) {
```
rubenada, line 3202: Sorry, I know I was the one that proposed this code, but I think we should not include this `if (...) return empty()` simplification, because it will be a duplication (this same simplification will be part any way of the overloaded method with RexNode parameters, so I think we should remove it from this one).

JiajunBernoulli: Never mind, I deleted the duplicate code. Thank you again~

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -644,6 +644,15 @@ final RelMetadataFixture sql(String sql) {
     fixture.assertThatRowCount(is(9D), is(0D), is(10d));
   }
 
+  @Test void testRowCountSortLimitOffsetDynamic() {
+    sql("select * from emp order by ename limit ? offset ?")
+        .assertThatRowCount(is(14D), is(0D), is(Double.POSITIVE_INFINITY));
```
rubenada, line 649: minor: in these tests, instead of `is(14D)`, I think it would be a bit better to use `is(EMP_SIZE)` (as some other tests in the same class do).
Similarly, instead of `is(13D)` we could use `is(EMP_SIZE - 1)`.

rubenada:
@JiajunBernoulli I left you a minor comment in a test. Apart from that, I think the PR is in a good shape and could be merged soon (after squashing commits into a single one, I'd suggest to keep the Jira title as message for the single commit after squash).

JiajunBernoulli:
> @JiajunBernoulli I left you a minor comment in a test. Apart from that, I think the PR is in a good shape and could be merged soon (after squashing commits into a single one, I'd suggest to keep the Jira title as message for the single commit after squash).
Thanks for your review, I used EMP_SIZE to replace 14D in single commit.

PR #2765, opened 2022-04-06
[CALCITE-4861] Chained cast calls can lead to unexpected behaviour

sbroeder:
- For cast to integer types check the max and min values of
             the return types and throw an exception if the literal
             value is too large for the return type.  Handles single
             casts as well as chained casts.
           - Casts of string values allow NumberFormatException to
             surface.


asolimando:
@sbroeder FYI you can run checker framework locally as follows to ease fixing the current issues detected by CI: `./gradlew -PenableCheckerframework`

matriv:
If I'm not mistaken, I don't think this solves the issue, it only addresses literals (?)
Chained calls where the arg is a table column, will still get simplified, and overflow can still occur.

sbroeder:
@matriv Yes, that's true.  So far I have only addressed literals cast as INT_TYPES.

PR #2767, opened 2022-04-07
[CALCITE-5027] Incorrect format for TIMESTAMP literals in SqlDialect#quoteTimestampLiteral

NobiGo:
`Date and Time Patterns` can find in SimpleDateFormat Java Doc.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlDialect.java
@@ -682,7 +682,7 @@ public class SqlDialect {
    * @return SQL timestamp literal
    */
   public String quoteTimestampLiteral(Timestamp timestamp) {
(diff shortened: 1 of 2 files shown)


asolimando:
@NobiGo, did you try implementing a test via SQL and extract function as Julian suggested in the Jira ticket? 

Such a test would be more self-describing and easier to check/understand w.r.t. one based on ms from epoch like in your test. 

IMO the PR is already valuable as-is, but it would be even better with such a test in place.

NobiGo:
@asolimando This method is only used in SqlBuilder for the purposes of creating SQL queries and expressions with String. So  I don't add a SQL test.

rubenada:
LGTM
(IMO perhaps the Jira title and the commit message could be a bit more specific, e.g. "Incorrect format for timestamp literals in SqlDialect#quoteTimestampLiteral")

PR #2768, opened 2022-04-07
[CALCITE-5086] Core parser should allow "OFFSET" to occur before "LIMIT"

hannerwang:
support trino style query like: SELECT * FROM users OFFSET 10 LIMIT 10

### DIFF
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -644,7 +644,11 @@ SqlNode OrderedQueryOrExpr(ExprContext exprContext) :
     [
         OffsetClause(startCount)
         [
-            LimitClause(startCount)
+            LimitClause(startCount) {
+                if (!this.conformance.isOffsetStartLimitCountAllowed()) {
+                    throw SqlUtil.newContextException(getPos(), RESOURCE.offsetStartLimitCountNotAllowed());
+                }
+            }
         ]
         |
         LimitClause(startCount)
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -606,6 +606,13 @@ JAVACODE boolean matchesPrefix(int[] seq, int[][] prefixes)
  *    [ OFFSET start ]</pre>
  * </blockquote>
  *
+ * <p>Trino syntax for limit:
+ *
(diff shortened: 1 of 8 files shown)


core/src/main/codegen/templates/Parser.jj
```diff
@@ -623,8 +630,7 @@ SqlNode OrderedQueryOrExpr(ExprContext exprContext) :
 {
     SqlNode e;
     SqlNodeList orderBy = null;
-    SqlNode start = null;
-    SqlNode count = null;
+    SqlNode[] startCount = {null, null};
 }
 {
```
chunweilei, line 635: Is it necessary to use Array to indicate the first value is the offset and the second value is the limit? It maybe not a good practice.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -3369,6 +3369,48 @@ void checkPeriodPredicate(Checker checker) {
...
+        .withConformance(SqlConformanceEnum.DEFAULT)
+        .fails(error);
+
+    // "limit 2" overrides the earlier "offset 4"
+    final String expected3 = "SELECT `A`\n"
+        + "FROM `FOO`\n"
+        + "OFFSET 2 ROWS\n"
+        + "FETCH NEXT 3 ROWS ONLY";
+    sql("select a from foo offset 4 limit 2,3")
+        .withConformance(SqlConformanceEnum.LENIENT)
```
chunweilei, line 3396: Do other systems support 'offset xx limit xx,xx'? It looks a little wired.

core/src/main/codegen/templates/Parser.jj
hannerwang: Hi @chunweilei , thanks for your review, I'm glad to hear your suggestions to improve it.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
hannerwang: Calcite supports limit xx,xx offset xx now, and the offset will override previous limit, the syntax looks strange also, however, I think it will be ok in lenient mode.

PR #2773, opened 2022-04-19
[CALCITE-5095] Support Java 18 and Guava 31.1-jre

libenchao:


libenchao:
"CI / CheckerFramework (JDK 11) (pull_request)" failed due to the upgrade of guava's version, it may be related to v31.0's "Annotated Guava much more thoroughly for nullness"(https://github.com/google/guava/releases/tag/v31.0)
I'll try to fix it ASAP.

core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
```diff
@@ -429,7 +430,8 @@ private static Properties loadProperties() {
         Thread.currentThread().getContextClassLoader(),
         CalciteSystemProperty.class.getClassLoader());
     // Read properties from the file "saffron.properties", if it exists in classpath
-    try (InputStream stream = classLoader.getResourceAsStream("saffron.properties")) {
```
NobiGo, line 432: Why do we need to add these restrictions？ I noticed a lot of changes in this.

NobiGo (commented):
The PR about the version looks good. Only left something needs to confirm.

core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
libenchao: Thanks for the review @NobiGo, these changes is because we upgraded guava version, which checks nullability more strictly, many apis added nullability annotation.

NobiGo (approved):
LGTM

core/src/main/java/org/apache/calcite/rel/externalize/RelEnumTypes.java
```diff
@@ -105,6 +107,6 @@ public static String fromEnum(Enum enumValue) {
   /** Converts a string to an enum value.
    * The converse of {@link #fromEnum(Enum)}. */
   static <E extends Enum<E>> E toEnum(String name) {
-    return (E) ENUM_BY_NAME.get(name);
+    return (E) Objects.requireNonNull(ENUM_BY_NAME.get(name));
```
julianhyde, line 110: This seems to be a change in the method contract. Previously the method would return null, now it throws. At least change the javadoc to make the contract explicit.

core/src/main/java/org/apache/calcite/rex/RexFieldCollation.java
```diff
@@ -44,7 +45,7 @@ public class RexFieldCollation extends Pair<RexNode, ImmutableSet<SqlKind>> {
           .build();
 
   public RexFieldCollation(RexNode left, Set<SqlKind> right) {
-    super(left, KINDS.get(right));
+    super(left, Objects.requireNonNull(KINDS.get(right)));
```
julianhyde, line 48: I tend to use a static import for Objects.requireNinNull to keep the code short.

libenchao:
@julianhyde Thanks for your review. I've updated the pr.

PR #2774, opened 2022-04-21
[CALCITE-5091] RelMdRowCount can return more accurate rowCount

JiajunBernoulli:
Here is the sql that fetch is deterministic and offset is dynamic 
`select * from emp order by ename limit 1 offset ?`
**RelMdRowCount#getRowCount** will return 14d, because it always return rowCount when offset is dynamic.

```
  if (rel.offset instanceof RexDynamicParam) {
    return rowCount;
  }
```
Obviously, rowCount will not exceed 1 because LIMIT 1. We can improve it

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdRowCount.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdRowCount.java
@@ -33,7 +33,6 @@ import org.apache.calcite.rel.core.TableModify;
 import org.apache.calcite.rel.core.TableScan;
 import org.apache.calcite.rel.core.Union;
 import org.apache.calcite.rel.core.Values;
-import org.apache.calcite.rex.RexDynamicParam;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.util.Bug;
 import org.apache.calcite.util.ImmutableBitSet;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/rel/metadata/RelMdRowCount.java
```diff
@@ -147,45 +146,27 @@ public Double getRowCount(Calc rel, RelMetadataQuery mq) {
...
     if (rowCount == null) {
       return null;
     }
-    if (rel.offset instanceof RexDynamicParam) {
-      return rowCount;
-    }
-    final int offset = rel.offset == null ? 0 : RexLiteral.intValue(rel.offset);
+
+    final int offset = rel.offset instanceof RexLiteral ? RexLiteral.intValue(rel.offset) : 0;
     rowCount = Math.max(rowCount - offset, 0D);
```
NobiGo, line 165: If we don't care about the offset value when it is a dynamic param, the row count value will be wrong? I think the origin logic is good.

NobiGo (changes requested):

core/src/main/java/org/apache/calcite/rel/metadata/RelMdRowCount.java
JiajunBernoulli: Yes, I think the row count value will be wrong. 
Do you think **14d** is right for the sql?
`select * from emp order by ename limit 1 offset ?`

chunweilei (commented):
LGTM

core/src/main/java/org/apache/calcite/rel/metadata/RelMdRowCount.java
NobiGo: The limit number is the max row count that the SQL can return.

NobiGo (approved):

PR #2775, opened 2022-04-21
[CALCITE-5107] Support SQL hint for Filter, SetOp, Sort, Window, Values

godfreyhe:
Currently, only part of RelNodes support SQL hint, such as: Scan, Project, Join. Filter, SetOp, Sort, Window, Values do not support hint yet, which cause hint propagation can not be supported. This pr aims to support those nodes


core/src/main/java/org/apache/calcite/rel/core/Sort.java
```diff
@@ -86,17 +126,7 @@ protected Sort(
       RelCollation collation,
       @Nullable RexNode offset,
       @Nullable RexNode fetch) {
-    super(cluster, traits, child);
```
danny0405, line 89: Thanks for the contribution, overall looks good, there are some compile failure, can you fix that ?

godfreyhe:
cc @danny0405

danny0405 (approved):
+1, LGTM.

PR #2776, opened 2022-04-24
[CALCITE-5081] Group keys of Aggregate are wrongly changed during decorrelation

libenchao:


core/src/test/java/org/apache/calcite/sql2rel/RelDecorrelatorTest.java
```diff
@@ -0,0 +1,89 @@
...
+        .scan("DEPT")
+        .filter(
+            builder.equals(builder.field(0),
+                builder.call(
+                    SqlStdOperatorTable.PLUS,
+                    builder.literal(10),
+                    builder.field(v.get(), "DEPTNO"))))
+        .correlate(JoinRelType.LEFT, v.get().id, builder.field(2, 0, "DEPTNO"))
+        .aggregate(builder.groupKey("ENAME"), builder.max(builder.field("EMPNO")))
+        .build();
```
chunweilei, line 67: Could you add a test that shows the plan before the change is correct and remain correct after the change?

libenchao: @chunweilei Thanks for the review. However I don't quite understand this comment, could you kindly elaborate it a little more?

chunweilei: The error happens in some special cases while it won't in other cases. What I propose is to add a test that shows the change works well in those cases.

libenchao: @chunweilei Currently, all tests for decorrelation is in `SqlToRelConverterTest`, enabled by `SqlToRelFixture#withDecorrelate`. 
The reason why I added the `RelDecorrelatorTest` is that we cannot reproduce this bug using SQL because in `SqlToRelConverter`, it has already [put the group columns in the front](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java#L5465-L5472).
Hence, all existing tests about decorrelation in `SqlToRelConverterTest` proves that this change does not break anything. Does this sounds reasonable to you?

chunweilei (commented):
LGTM

chunweilei (commented):
LGTM

chunweilei:
@libenchao Could you please resolve the conflicts?

libenchao:
@chunweilei Thanks for the review, I've resolved the conflicts. 
Druid tests failed due to timeout which is not related to this change.

PR #2781, opened 2022-04-27
[CALCITE-5117]  Optimize the EXISTS sub-query by Metadata RowCount

NobiGo:


core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -450,8 +450,18 @@ private static RexNode rewriteSome(RexSubQuery e, Set<CorrelationId> variablesSe
...
+    // If the sub-query is guaranteed to produce at least one row, just return
+    // TRUE.
+    final RelMetadataQuery mq = e.rel.getCluster().getMetadataQuery();
+    final Double minRowCount = mq.getMinRowCount(e.rel);
+    if (minRowCount != null && minRowCount >= 1D) {
+      return builder.literal(true);
+    }
+    final Double maxRowCount = mq.getMaxRowCount(e.rel);
+    if (maxRowCount != null && maxRowCount < 1D) {
+      return builder.literal(false);
```
chunweilei, line 462: Could you add a test in RelOptRuleTest so that we can see the optimized plan?

NobiGo: Done.

chunweilei (approved):
LGTM

PR #2783, opened 2022-04-28
[CALCITE-5109] RelMdAllPredicates and RelMdExpressionLineage support to analyze left/right join

wojustme:
JIRA: https://issues.apache.org/jira/browse/CALCITE-5109

Summary:
- Expand Model of RexTableInputRef
- RelMdExpressionLineage support the analysis of left/right join
- RelMdAllPredicates support the analysis of left/right join

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexTableInputRef.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexTableInputRef.java
@@ -18,6 +18,7 @@ package org.apache.calcite.rex;
 
 import org.apache.calcite.plan.RelOptTable;
 import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.sql.SqlKind;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/calcite/rex/RexTableInputRef.java
```diff
@@ -46,10 +47,23 @@ public class RexTableInputRef extends RexInputRef {
...
+  private final boolean wrapJoinNullable;
+
+  private RexTableInputRef(RelTableRef tableRef, int index, RelDataType type,
+      boolean wrapJoinNullable) {
     super(index, type);
+    if (wrapJoinNullable) {
+      // Nullable of type must be true, if this input ref is wrapped by left/right join.
+      assert type.isNullable();
+    }
     this.tableRef = tableRef;
```
chunweilei, line 64: Use ```assert wrapJoinNullable && type.isNulable()```.

core/src/main/java/org/apache/calcite/rex/RexTableInputRef.java
```diff
@@ -46,10 +47,23 @@ public class RexTableInputRef extends RexInputRef {
...
 
-  private RexTableInputRef(RelTableRef tableRef, int index, RelDataType type) {
+  /**
+   * It's a tag, which ref is wrapped by out join. And it always emits nullable = true.
+   * Input ref will be true, if it exists in right's input of left-join
+   * or left's input of right-join are true
+   */
+  private final boolean wrapJoinNullable;
+
+  private RexTableInputRef(RelTableRef tableRef, int index, RelDataType type,
```
chunweilei, line 57: Is it better to use ```forceNullable``` instead of ```wrapJoinNullable```?

core/src/main/java/org/apache/calcite/rex/RexTableInputRef.java
wojustme: Yes, It looks more concise
But, Code should be `assert !forceNullable || type.isNullable();`

core/src/main/java/org/apache/calcite/rex/RexTableInputRef.java
wojustme: It seems good, Thanks

wojustme:
@chunweilei 
I'm sorry to reply you so late, because I've been a little busy lately.
Code is updated, please take a time to review it. Thanks.

core/src/main/java/org/apache/calcite/rex/RexTableInputRef.java
asolimando: If possible, try also to add an error message to newly added assertions

wojustme: @asolimando 
Yes, you are right. It's a good habit.

chunweilei:
Hi, @wojustme . I will look through the PR in the coming days.

chunweilei:
To be honest, I am not familiar with this part. So I need more time to review it. Besides, It would be great if more people can help review it.

wojustme:
> To be honest, I am not familiar with this part. So I need more time to review it. Besides, It would be great if more people can help review it.

I believe that the process of discussion is to let us grow together

PR #2784, opened 2022-04-29
Remove useless explicit type cast when using SqlValidator

chunweilei:


chunweilei:
file a new PR: use https://github.com/apache/calcite/pull/2785

PR #2785, opened 2022-04-29
Remove useless explicit type cast when using SqlValidator

chunweilei:


rubenada:
+1.
I think the same change can be applied to `SqlInOperator:148` and `SqlJsonObjectAggAggFunction:77`

chunweilei:
> +1. I think the same change can be applied to `SqlInOperator:148` and `SqlJsonObjectAggAggFunction:77`

Nice catch. I have updated the PR. Thank you for your review.

PR #2786, opened 2022-04-29
[CALCITE-5003] MergeUnion on types with different collators produces wrong result

rubenada:
[[CALCITE-5003](https://issues.apache.org/jira/browse/CALCITE-5003)] MergeUnion on types with different collators produces wrong result.
Make sure that all inputs are sorted in the same way, adding a projection in case the input's type collation is not the same as the union's.

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableMergeUnionRule.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableMergeUnionRule.java
@@ -19,18 +19,24 @@ package org.apache.calcite.adapter.enumerable;
 import org.apache.calcite.plan.RelOptRuleCall;
 import org.apache.calcite.plan.RelRule;
 import org.apache.calcite.rel.RelCollation;
+import org.apache.calcite.rel.RelFieldCollation;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.Sort;
 import org.apache.calcite.rel.core.Union;
 import org.apache.calcite.rel.logical.LogicalSort;
 import org.apache.calcite.rel.logical.LogicalUnion;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeField;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableMergeUnionRule.java
```diff
@@ -90,9 +96,42 @@ public EnumerableMergeUnionRule(Config config) {
       }
     }
 
+    final RelBuilder builder = call.builder();
     final List<RelNode> inputs = new ArrayList<>(unionInputsSize);
     for (RelNode input : union.getInputs()) {
-      final RelNode newInput = sort.copy(sort.getTraitSet(), input, collation, null, inputFetch);
+      // Check if type collations match, otherwise store it in this map to generate a cast later
+      // to guarantee that all inputs will be sorted in the same way
+      final Map<Integer, RelDataType> castMap = new HashMap<>();
```
thomasrebele, line 104: How about using a bit set instead of a map? The value is only used once and can easily be looked up again.

thomasrebele (commented):
A small nitpick, otherwise looks good to me.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableMergeUnionRule.java
rubenada: Makes sense. Change applied.

thomasrebele: The union.getRowType().getFieldList() could be factored out of the for-loop. I guess that it does not impact the performance much, so up to you to decide.

rubenada: ok, done.

rubenada:
Thanks for the review @thomasrebele !
I plan to merge the patch in the coming days, if no objection appears.

PR #2789, opened 2022-05-03
[CALCITE-5126] Implicit column alias for single-column UNNEST should work with any single-column UNNEST’s input

dssysolyatin:


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -2335,11 +2335,7 @@ private SqlNode registerFrom(
       expr = call.operand(0);
       final boolean needAlias = call.operandCount() > 2
           || expr.getKind() == SqlKind.VALUES
-          || expr.getKind() == SqlKind.UNNEST
-          && (((SqlCall) expr).operand(0).getKind()
-                  == SqlKind.ARRAY_VALUE_CONSTRUCTOR
-              || ((SqlCall) expr).operand(0).getKind()
-                  == SqlKind.MULTISET_VALUE_CONSTRUCTOR);
+          || expr.getKind() == SqlKind.UNNEST;
```
rubenada, line 2338: @julianhyde sorry to ping you directly.
I think this PR looks in a good shape, I was just wondering if you have any problem or concern about this particular change, which alters a condition that you explicitly added in [c2527cc](https://github.com/apache/calcite/commit/c2527ccf440f7750bfbabd2063c402440c5b32a0).
If you don't have any objection, I think we should move forward and merge the patch, wdyt?

core/src/test/resources/sql/unnest.iq
```diff
@@ -105,6 +105,29 @@ from UNNEST(array ['apple', 'banana']) as "fruit";
...
++--------+
+| FRUIT  |
++--------+
+| banana |
++--------+
+(1 row)
+
+!ok
+
+#TODO: Uncomment when CALCITE-5127 will be fixed
```
julianhyde, line 119: @dssysolyatin Please remove the comments and use `!if (false) {` instead. (Results in fewer diffs.) Also change 'CALCITE-5127' to '[CALCITE-5127]' for easier search.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
julianhyde: Thanks for checking with me. I think it's OK to loosen that condition; that's kind of the point of this bug.

@dssysolyatin Can you add a validator test that `select * from depts as d, unnest (d.emps)`, which was previously valid because there was not an array or multiset constructor inside the unnest, now requires an alias. (Or some SQL to that effect.)

julianhyde: +1 to merge this PR when we have that extra validator test and the change I requested in `unnest.iq`.

dssysolyatin: @julianhyde Can you explain why `select * from depts as d, unnest (d.emps)` should require alias. It works with PR changes as well. For example:
```
select * from (SELECT ARRAY['banana'] as fruit_arr) as q, UNNEST(q.fruit_arr);
```

returns (if use current branch)
```
+-----------+--------+
| FRUIT_ARR | EXPR$0 |
+-----------+--------+
| [banana]  | banana |
+-----------+--------+
```

returns (if use PG)
```
 fruit_arr | unnest 
-----------+--------
 {banana}  | banana
(1 row)
```

julianhyde: I think it should require an alias because the variable is called `needAlias`. (I'm being deliberately stupid because I am reviewing the code, not writing the PR.) If the variable's name is misleading please change it.

dssysolyatin: @julianhyde I would rename `needAlias` to `needAliasNamespace`. Also it should be taken into account that this code will be executed only if there is `AS` operator. If there is no `AS` operator like in your case, it will be work as usual
```
switch (kind) {
    case AS:
      call = (SqlCall) node;
      if (alias == null) {
        alias = String.valueOf(call.operand(1));
      }
      expr = call.operand(0);
      final boolean needAliasNamespace = call.operandCount() > 2
          || expr.getKind() == SqlKind.VALUES
          || expr.getKind() == SqlKind.UNNEST;
```

dssysolyatin: @julianhyde @rubenada Can we merge this PR ?

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -8363,6 +8363,11 @@ public void _testGroupExpressionEquivalenceParams() {
     // UNNEST MULTISET gets the same treatment as UNNEST ARRAY.
     sql("select * from unnest(multiset [1, 2, 1]) as f")
         .type("RecordType(INTEGER NOT NULL F) NOT NULL");
+
+    // The magic doesn't happen if the UNNEST is used without AS operator.
+    sql("select * from (SELECT ARRAY['banana'] as fruits) as t, UNNEST(t.fruits)")
+            .type("RecordType(CHAR(6) NOT NULL ARRAY NOT NULL FRUITS, "
+                    + "CHAR(6) NOT NULL EXPR$0) NOT NULL").ok();
```
julianhyde, line 8370: please fix indentation

dssysolyatin: Fixed.

PR #2790, opened 2022-05-03
[CALCITE-3557] ClassCastException for using nested multiset or array …

JiajunBernoulli:
…inside multiset

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/JdbcTest.java
+++ b/core/src/test/java/org/apache/calcite/test/JdbcTest.java
@@ -2118,6 +2118,40 @@ public class JdbcTest {
         .returnsUnordered("A=[{10}, {20}, {10}, {10}]");
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-3557">[CALCITE-3557]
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -2118,6 +2118,40 @@ void checkMultisetQueryWithSingleColumn() {
         .returnsUnordered("A=[{10}, {20}, {10}, {10}]");
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-3557">[CALCITE-3557]
+   * ClassCastException for using nested multiset or array inside multiset</a>.
```
chunweilei, line 2124: Does this change miss something? It seems CALCITE_3557_FIXED is used only in the test.

JiajunBernoulli: Sorry, I forget to say the fix is in calcite-avatica: https://github.com/apache/calcite-avatica/pull/179/files
After updating calcite-avatica version, we can disable CALCITE_3557_FIXED and enable test. So the PR only add some tests in calcite.

PR #2791, opened 2022-05-05
[CALCITE-5134] Queries with subquery inside select list does not work if subquery uses table from left join

libenchao:


dssysolyatin (approved):

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4883,7 +4883,14 @@ void setRoot(List<RelNode> inputs) {
           int i = 0;
           int offset = 0;
           for (SqlValidatorNamespace c : ancestorScope1.getChildren()) {
-            builder.addAll(c.getRowType().getFieldList());
+            final boolean isNullable = ancestorScope1.children.get(i).isNullable();
```
rubenada, line 4886: Thanks for the PR @libenchao .
I'm not the most expert on the impacted classes, but I guess some people might frown upon making `ScopeChild` public just for this fix. What about...
- Leaving `ScopeChild` unchanged
- Add a new method in `ListScope`:
```
  public boolean isChildNullable(int i) {
    return children.get(i).nullable;
  }
```
- Changing the `if` condition on the `SqlToRelConverter` fix code to use this method:
```
for (SqlValidatorNamespace c : ancestorScope1.getChildren()) {
  if (ancestorScope1.isChildNullable(i)) {
    for (final RelDataTypeField f : c.getRowType().getFieldList()) {
      builder.add(f.getName(), typeFactory.createTypeWithNullability(f.getType(), true));
    }
  } else {
    builder.addAll(c.getRowType().getFieldList());
  }
  ...
```              
In my opinion, it would be a slightly cleaner solution than the current proposal, WDYT?

libenchao: Yes, it's more cleaner. I've revised it. Thanks for your review.

rubenada (approved):

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
libenchao: done.

PR #2795, opened 2022-05-08
[CALCITE-5135] Implement SQL DAY function

JiajunBernoulli:


core/src/main/java/org/apache/calcite/sql/SqlJdbcFunctionCall.java
```diff
@@ -732,6 +732,7 @@ private JdbcToInternalLookupTable() {
       map.put("MONTH", simple(SqlStdOperatorTable.MONTH));
       map.put("WEEK", simple(SqlStdOperatorTable.WEEK));
       map.put("DAYOFYEAR", simple(SqlStdOperatorTable.DAYOFYEAR));
+      map.put("DAY", simple(SqlStdOperatorTable.DAYOFMONTH));
```
xy2953396112, line 735: How is it different from `DAYOFMONTH`?

JiajunBernoulli: Thanks for your review. It is same as `DAYOFMONTH `, but it is not a valid JDBC function. I've deleted it.

PR #2800, opened 2022-05-10
[CALCITE-3890] Derive IS NOT NULL filter for the inputs of inner join

chunweilei:


asolimando (approved):
LGTM, I left few minor comments which are optional

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
```diff
@@ -0,0 +1,108 @@
...
+    if (newLeftNode != join.getLeft() || newRightNode != join.getRight()) {
+      List<RelNode> inputs = new ArrayList<>();
+      inputs.add(newLeftNode);
+      inputs.add(newRightNode);
+      RelNode newJoin = join.copy(join.getTraitSet(), inputs);
+      call.transformTo(newJoin);
+    }
+  }
+
+  private RelNode createIsNotNullFilter(RelNode input, ImmutableIntList keys,
```
julianhyde, line 73: This method would be simpler if it didn't return a RelNode but just operated on the RelBuilder. You wouldn't need to call simplify, or check whether the condition is true. It's quite possible it would be just one line.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -411,6 +411,37 @@ private static boolean skipItem(RexNode expr) {
...
+   * should create is not null filter in the right input of join. */
+  @Test void testJoinDeriveIsNotNullFilterRule2() {
+    final String sql = "select t1.deptno from (select * from emp where mgr is not null) t1 "
+        + "join emp t2 on t1.mgr = t2.mgr";
+    sql(sql).withRule(CoreRules.JOIN_DERIVE_IS_NOT_NULL_FILTER_RULE).check();
+  }
+
+  /** As {@link #testJoinDeriveIsNotNullFilterRule1()};
+   * should not create is not null filter in both sides since it is full join. */
+  @Test void testJoinDeriveIsNotNullFilterRule3() {
```
julianhyde, line 432: Capitalize SQL keywords. "create is not null filter" is ungrammatical, but "create IS NOT NULL filter" makes sense.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
chunweilei: Simplify is used to avoid creating redundant IS NOT NULL filter. Otherwise, it would create a new IS NOT NULL filter every time the rule applies.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
chunweilei: That's true.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
```diff
@@ -0,0 +1,108 @@
...
+    final RexSimplify simplify = new RexSimplify(rexBuilder, relOptPredicateList, executor);
+    List<RexNode> nodes = new ArrayList<>();
+    keys.forEach(i -> {
+      final RexNode expr = relBuilder
+          .call(SqlStdOperatorTable.IS_NOT_NULL, rexBuilder.makeInputRef(input, i));
+      RexNode simplified = simplify.simplify(expr);
+      if (!simplified.isAlwaysTrue()) {
+        nodes.add(expr);
+      }
+    });
```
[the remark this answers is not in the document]
chunweilei: I prefer the current implementation~~.

wojustme:
Hi @chunweilei 
I think we need consider more possibilities about join-condition.
An example:
`select t1.deptno from emp t1 inner join emp t2 on t1.mgr IS NOT DISTINCT FROM t2.mgr`
I think we needn't create new filter of `is_not_null`, because the join condition will return true(`t1.mgr is null and t2.mgr is null`)

chunweilei:
> Hi @chunweilei I think we need consider more possibilities about join-condition. An example: `select t1.deptno from emp t1 inner join emp t2 on t1.mgr IS NOT DISTINCT FROM t2.mgr` I think we needn't create new filter of `is_not_null`, because the join condition will return true(`t1.mgr is null and t2.mgr is null`)

That's a good point.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
julianhyde: If you call {{RelBuilder.filter}} it will simplify, and will also ignore {{true}} filters. Same effect, fewer lines of code.

chunweilei: If I understand correctly, what you propose can only handle the following case:
```
LogicalProject(DEPTNO=[$7])
  LogicalJoin(condition=[=($3, $12)], joinType=[inner])
      LogicalFilter(condition=[IS NOT NULL($3)])
```
. But it can not handle:
```
LogicalProject(DEPTNO=[$1])
  LogicalJoin(condition=[=($0, $3)], joinType=[inner])
    LogicalProject(ENAME=[$1], DEPTNO=[+($7, 1)])
      LogicalFilter(condition=[IS NOT NULL($1)])
        LogicalTableScan(table=[[CATALOG, SALES, EMPNULLABLES]])
    LogicalTableScan(table=[[CATALOG, SALES, EMPNULLABLES]])
```
.

chunweilei:
The PR is updated.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
```diff
@@ -0,0 +1,130 @@
...
+
+    final RelNode newLeftNode = createIsNotNullFilter(join.getLeft(), leftKeys,
+        relBuilder, mq);
+    final RelNode newRightNode = createIsNotNullFilter(join.getRight(), rightKeys,
+        relBuilder, mq);
+
+    if (newLeftNode != join.getLeft() || newRightNode != join.getRight()) {
+      final List<RelNode> inputs = new ArrayList<>();
```
liyafan82, line 87: minor: we can specify the initial capacity here?

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
```diff
@@ -0,0 +1,130 @@
...
+    final List<RexNode> nodes = RelOptUtil.conjunctions(condition);
+    final ImmutableBitSet.Builder notNullableKeys = ImmutableBitSet.builder();
+    nodes.forEach(node -> {
+      if (Strong.isStrong(node)) {
+        notNullableKeys.addAll(RelOptUtil.InputFinder.bits(node));
+      }
+    });
+
+    final List<Integer> leftKeys = new ArrayList<>();
+    final List<Integer> rightKeys = new ArrayList<>();
```
liyafan82, line 70: why not declare them as 'ImmutableBitSet'?

liyafan82:
LGTM, with some minor comments.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
chunweilei: Done.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
chunweilei: Done.

chunweilei:
Thank you for your review. Comments are addressed @liyafan82 .

wojustme:
I looked through the code, and this pr is good for me.
I try an another test case `select t1.deptno from emp t1 join emp t2 on coalesce(t1.mgr = t2.mgr, false)`, and the result of rule isn't what I want.
In my opinion, this rule is OK now, because it has no side effects, although this rule doesn't work. I think `org.apache.calcite.plan.Strong` could be enhanced.

LGTM for this pr.

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
```diff
@@ -570,6 +570,9 @@ private CoreRules() {}
       JOIN_PUSH_TRANSITIVE_PREDICATES =
       JoinPushTransitivePredicatesRule.Config.DEFAULT.toRule();
 
+  public static final JoinDeriveIsNotNullFilterRule JOIN_DERIVE_IS_NOT_NULL_FILTER_RULE =
+      JoinDeriveIsNotNullFilterRule.Config.DEFAULT.toRule();
+
```
julianhyde, line 575: no javadoc comment? every other field has a comment

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.calcite.util.Util;
+
+import org.immutables.value.Value;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Planner rule that derives IS NOT NULL filter from inner join.
+ */
```
julianhyde, line 46: as i've said elsewhere 'derives a filter' isn't very clear. you don't say that it is under the join. you don't say that it is a relnode (it could be a constraint)

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
chunweilei: Done.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
chunweilei: Thank you for your suggestion. I updated the java doc comment.

julianhyde:
Please take a look at https://github.com/julianhyde/calcite/tree/3890-join-is-not-null. I prototyped how this change would look if we let `RelBuilder.filter` do the work. (It needed a change to `RelBuilder` to simplify using the current `RelNode`'s pulled up predicates.)

julianhyde:
It might be worth noting, somewhere, the similarity between this rule and CoreRules#FILTER_INTO_JOIN.

julianhyde:
I saw you used `Strong`, which is great. At some point we could improve this - add a method say `ImmutableBitSet Strong.strongKeys(RexNode e)` that returns, for a given expression `e`, the columns that must not be null in order for `e` to be not null.

The rule currently does
```
    final ImmutableBitSet.Builder notNullableKeys = ImmutableBitSet.builder();
    RelOptUtil.conjunctions(join.getCondition()).forEach(node -> {
      if (Strong.isStrong(node)) {
        notNullableKeys.addAll(RelOptUtil.InputFinder.bits(node));
      }
    });
```
but that `strongKeys` would not need to call `conjunctions` because it would work on `AND` just like any other operator.

liyafan82 (approved):
LGTM, thanks.

chunweilei:
> Please take a look at https://github.com/julianhyde/calcite/tree/3890-join-is-not-null. I prototyped how this change would look if we let `RelBuilder.filter` do the work. (It needed a change to `RelBuilder` to simplify using the current `RelNode`'s pulled up predicates.)

Awesome!

chunweilei:
> It might be worth noting, somewhere, the similarity between this rule and CoreRules#FILTER_INTO_JOIN.

Done. I wrote some notes about the similarity in `JoinDeriveIsNotNullFilterRule`.

PR #2801, opened 2022-05-10
[CALCITE-5138] Join on condition generates wrong plan when the condit…

NobiGo:
…ion is sub-query


NobiGo:
This issue will address the normal sub-query, not the Correlate sub-query.

asolimando (approved):
LGTM, I am particularly happy about the comprehensive set of tests you have added, good work @NobiGo!

PR #2803, opened 2022-05-11
[CALCITE-5130] AssertionError: "Conversion to relational algebra failed to preserve datatypes" when union VARCHAR literal and CAST(null AS INTEGER)

yingyuwang:
In AbstractTypeCoercion#promoteToVarChar() return resultType with nullability if any of the types being promoted are nullable.

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
+++ b/core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
@@ -216,7 +216,7 @@ class TypeCoercionTest {
     sql("select '1' from (values(true)) union values 2")
         .type("RecordType(VARCHAR NOT NULL EXPR$0) NOT NULL");
     sql("select (select 1+2 from (values true)) tt from (values(true)) union values '2'")
-        .type("RecordType(VARCHAR NOT NULL TT) NOT NULL");
+        .type("RecordType(VARCHAR TT) NOT NULL");
     // union with star
     sql("select * from (values(1, '3')) union select * from (values('2', 4))")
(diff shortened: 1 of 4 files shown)


libenchao (approved):
@yingyuwang LGTM, Thanks for pr.

core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
```diff
@@ -216,7 +216,7 @@ private static ImmutableList<RelDataType> combine(
     sql("select '1' from (values(true)) union values 2")
         .type("RecordType(VARCHAR NOT NULL EXPR$0) NOT NULL");
     sql("select (select 1+2 from (values true)) tt from (values(true)) union values '2'")
-        .type("RecordType(VARCHAR NOT NULL TT) NOT NULL");
+        .type("RecordType(VARCHAR TT) NOT NULL");
     // union with star
```
chunweilei, line 220: Is this change expected?

yingyuwang: What I found is Calcite seems to add `isNullable=true` when there is a sub query, for example we have:
```
    sql("select 1 from (values 1)")
        .type("RecordType(INTEGER NOT NULL EXPR$0) NOT NULL"); // isNullable=false for EXPR$0 when not query from subquery
    sql("select (select 1 from (values 1))")    
        .type("RecordType(INTEGER EXPR$0) NOT NULL"); // isNullable=true for EXPR$0 when query from subquery
```
So before the change, `isNullable=true` is lost at `promoteToVarChar()` for this test.

PR #2804, opened 2022-05-12
[CALCITE-5150] Parser should parse subquery with order by inside array constructor

dssysolyatin:
…y constructor


chunweilei (commented):
LGTM

PR #2806, opened 2022-05-13
[CALCITE-5061] Improve recursive application of the field trimming

rubenada:
[CALCITE-5061](https://issues.apache.org/jira/browse/CALCITE-5061) - Improve recursive application of the field trimming: if a RelNode itself cannot be trimmed, try to trim inside its subtree as much as possible


hsyuan (approved):
LGTM.

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -352,11 +352,24 @@ public TrimResult trimFields(
...
+    }
+
+    // We don't know how to trim this RelNode, but we can try to trim inside its inputs
+    List<RelNode> newInputs = new ArrayList<>(rel.getInputs().size());
+    for (RelNode input : rel.getInputs()) {
+      ImmutableBitSet inputFieldsUsed = ImmutableBitSet.range(input.getRowType().getFieldCount());
+      TrimResult trimResult = dispatchTrimFields(input, inputFieldsUsed, extraFields);
+      if (!trimResult.right.isIdentity()) {
+        throw new IllegalArgumentException();
+      }
```
chunweilei, line 368: > throw new IllegalArgumentException();

Should we add some clear exception messages instead?

chunweilei (approved):
LGTM. Left a minor comment.

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
rubenada: Exception message improved.

rubenada:
Thanks for the review @hsyuan @chunweilei .
I have squashed commits into a single one. I will merge this PR in the coming days if no objection appears.

PR #2133, opened 2020-09-02
[CALCITE-2040] Arrow adapter

KarshitShah:

PR #2810, opened 2022-05-19
[CALCITE-2040] Arrow adapter (continued)

jonathanswenson:
Builds on the work done by #2133 and additional commits by @julianhyde on his branch at https://github.com/julianhyde/calcite/tree/2040-arrow

Added contributions: 
+ upgrade to arrow 8.0.0 (latest) allowing removal of the checked in jar - 7.0.0 likely will fit the same requirements, but I have not tried it.
+ Integrate the ArrowData class directly into the tests rather than depending on a gradle task to generate the arrow test file.
+ Fix a bug in early termination of the ArrowEnumerable if a filter does not match a given record batch. 
+ Begin implementing the tests that @julianhyde has specified. 

There are a number of additional tests that have been called out. But many of them are currently not supported by the current approach. This appears to match the high level approach implemented by the cassandra adapter (hence the lack of support for several features).


jonathanswenson:
Closing in favor of #3666 (includes all of this code)

PR #2811, opened 2022-05-20
[CALCITE-5156] Support implicit type cast for IN Sub-query

NobiGo:


core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
```diff
@@ -257,15 +257,6 @@ protected boolean needToCast(SqlValidatorScope scope, SqlNode node, RelDataType
       return false;
     }
 
-    // No need to cast if the source type precedence list
```
JiajunBernoulli, line 260: I found that you only changed the conversion rules between tinyint, int and bigint. Do you want to support more types for IN Sub-query?  For example:
```
  @Test void testInSubQuery() {
    sql("select * from t1 where t1_bigint in (select t1_decimal from t1)")
        .withExpand(false)
        .ok();
  }
```
> Learn from you.

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -1704,6 +1704,23 @@ void checkCorrelatedMapSubQuery(boolean expand) {
     sql(sql).withExpand(false).ok();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5156">[CALCITE-5156]
+   * Support implicit type cast for IN Sub-query</a>. */
+  @Test void testInSubQueryWithTypeCast() {
```
JiajunBernoulli, line 1710: Maybe the UnitTest can be writtren in TypeCoercionConverterTest
> Just minor comment

chunweilei: > Maybe the UnitTest can be written in TypeCoercionConverterTest

+1.

core/src/test/resources/sql/sub-query.iq
```diff
@@ -1174,7 +1173,7 @@ EnumerableCalc(expr#0..3=[{inputs}], expr#4=[IS NULL($t3)], expr#5=[null:BOOLEAN
     EnumerableLimit(fetch=[1])
       EnumerableSort(sort0=[$0], dir0=[DESC])
         EnumerableAggregate(group=[{0}], c=[COUNT()])
-          EnumerableCalc(expr#0..2=[{inputs}], expr#3=[true], cs=[$t3])
+          EnumerableCalc(expr#0..2=[{inputs}], expr#3=[CAST($t0):INTEGER], expr#4=[IS NOT NULL($t3)], cs=[$t4])
             EnumerableTableScan(table=[[scott, DEPT]])
 !plan
```
chunweilei, line 1178: Are these plan changes desirable?

core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
```diff
@@ -91,11 +91,21 @@ public abstract class AbstractTypeCoercion implements TypeCoercion {
...
       RelDataType targetType) {
+    return coerceOperandType(scope, call, index, targetType, true);
+  }
+
+  protected boolean coerceOperandType(
+      @Nullable SqlValidatorScope scope,
+      SqlCall call,
+      int index,
+      RelDataType targetType,
+      boolean isUpdateValidate) {
```
mihaibudiu, line 108: I would name this parameter "updateValidator"

mihaibudiu (approved):
This looks very useful, can you please resolve the conflicts?

NobiGo:
@mihaibudiu Hi, Sorry to bother you. If you have free time, please review this again.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypePrecedenceList.java
```diff
@@ -31,7 +31,7 @@ public interface RelDataTypePrecedenceList {
    * Determines whether a type appears in this precedence list.
    *
    * @param type type to check
-   * @return true iff this list contains type
+   * @return true if this list contains type
```
mihaibudiu, line 34: iff means "if and only if"
was the original comment wrong?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -1958,7 +1959,33 @@ private RelOptUtil.Exists convertExists(
       if (!values.getTuples().isEmpty()) {
         unionInputs.add(values);
       }
-      resultRel = LogicalUnion.create(unionInputs, true);
+      RelDataType finalRowType = rowType;
```
mihaibudiu, line 1962: This is a nice optimization, but I don't understand how it relates to the issue that is being solved.
This should probably be a separate PR.
Moreover, this looks like this should be part of UNION_TO_VALUES, and not always part of SqlToRelConverter.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypePrecedenceList.java
NobiGo: I misunderstood. I will rollback this.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
NobiGo: ﻿@mihaibudiu Yes, as you can see, it is the same as UNION_TO_VALUES but works in different phases. One is the convert SQL to RelNode phase, and the other is the optimize phase.
﻿
Before this PR. When we convert the SQL to RelNode(convert IN list to Join):
```
SELECT empno FROM emp AS e WHERE cast(e.empno as bigint) in (130, 131, 132, 133, 134)
```
The generated RelNode contains:
﻿
```
LogicalAggregate(group=[{0}])
  LogicalValues(tuples=[[{ 130 }, { 131 }, { 132 }, { 133 }, { 134 }]])
```
But when we add CAST in the IN list value:
The generated RelNode will be:
```
LogicalUnion(all=[true])
  LogicalValues(tuples=[[{ 130 }]])
  LogicalValues(tuples=[[{ 131 }]])
  LogicalValues(tuples=[[{ 132 }]])
  LogicalValues(tuples=[[{ 133 }]])
  LogicalValues(tuples=[[{ 134 }]])
```
I want to keep this unchanged, so add the same processing method here.
The unit tests in `SqlToRelConverterTest#testInToSemiJoin`.

mihaibudiu: I think you are fixing a symptom of the bug, and not the actual bug.
This fix should work no matter which subquery is present. It would work even if the subquery is a union, or a LogicalValues, or any other query.
The fact that you have to handle union of values specially suggests that you are not fixing the bug in the issue, but another bug. Maybe the issue is wrong, and this is a bug with Union.

mihaibudiu: The way to find out is to write another test query, where the query in the IN right-hand-side returns a result with a type that will also require a cast.

NobiGo: Are you referring to this SQL?
```
select * from dept where cast(deptno as bigint) in (select deptno + 20 from dept)
```
I add a test case in SqlToRelConverterTest#testInSubQueryWithTypeCast.

mihaibudiu: whatever fixes this query should also handle the union case.

NobiGo: The commit info describes a usage phenomenon caused by this defect, not an actual bug description.

mihaibudiu: Where is the actual bug that is being fixed?

NobiGo: > whatever fixes this query should also handle the union case.

Can you give a unit test case or an example SQL? I will cover it.

NobiGo: In Calcite, we don't add cast if the source type precedence list contains a target type if type is integer types (smallint, tinyint, integer, and bigint). But if we don't add cast, the SQL output will be wrong. 
So In this PR, I will add cast even if the source type precedence list contains the target type(only works in integer types).  The corresponding code in AbstractTypeCoercion.java.

mihaibudiu: Great, so you deleted code which was omitting the casts.
The question I have: is this change handling union in the SqlToRelConverter required to make the test pass? If it's still required, it means that there is somewhere another bug. 
If it's not required, perhaps you don't need to add this change about union in this PR.

NobiGo: SqlToRelConverter is still required. Because it occurred during the conversion of IN list to Values.

Before this PR. 

The SQL:
```
SELECT empno FROM emp AS e WHERE cast(e.empno as bigint) in (130, 131, 132, 133, 134)
```
The Calcite will convert it to :
```
LogicalAggregate(group=[{0}])
  LogicalValues(tuples=[[{ 130 }, { 131 }, { 132 }, { 133 }, { 134 }]])
```
But when we add CAST like:
```
SELECT empno FROM emp AS e WHERE cast(e.empno as bigint) in (cast(130 as bigint), cast(131 as bigint), cast(132 as bigint), cast(133 as bigint), cast(134 as bigint))
````
The Calcite will convert it to:
```
LogicalUnion(all=[true])
  LogicalValues(tuples=[[{ 130 }]])
  LogicalValues(tuples=[[{ 131 }]])
  LogicalValues(tuples=[[{ 132 }]])
  LogicalValues(tuples=[[{ 133 }]])
  LogicalValues(tuples=[[{ 134 }]])
```
This test case was not covered before. The change in this PR makes this bug apparent. Do we need a separate PR to resolve it first? I hope this can answer your question.

Comment (mihaibudiu) [reply core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java:1962]: Why is the form with union not correct? Has the cast information been lost?

Comment (NobiGo) [reply core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java:1962]: The Union is correct, but VALUE is simpler. Not lost anything in UNION.
Before this PR, the IN-LIST SQL without cast can be converted to values, I hope it can still be done here after we add the CAST.

Comment (mihaibudiu) [reply core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java:1962]: If it works, I would prefer a solution in which the UNION rewrite is done using the UNION_TO_VALUES rule. There is no point in duplicating the logic of that rule in the SqlToRelConverter.

Comment (NobiGo) [reply core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java:1962]: Yes, I get it. UNION_TO_VALUES works in the optimization phase. But here is to convert the SQL to the RelNode phase. 
Before this PR. SQL like `cast(number as bigint) in (12,13,14) ` can convert to 'VALUES' when converting the SQL to the RelNode. I hope to keep the semantics here unchanged, so I added it here. If we don't do this, the RelNode converted here will be different from the previous version. According to the review comment in Jira https://issues.apache.org/jira/browse/CALCITE-5156 it is unacceptable.

Comment (mihaibudiu) [reply core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java:1962]: But if you leave out the optimization of union from SqlToRelConverter, won't the plans in RelOptRulesTest look more like they used to?

I think that the insertion of the cast is unavoidable, that's after all what this change is about.

Comment (NobiGo) [reply core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java:1962]: I understand what you mean, but I need to clarify that this does not include all UNION. It is only when converting IN-LIST to values. If you think we should remove the changes in SqlToRelConverter then I will do it. I think this is not a bug, as long as we reach an agreement, it's fine.

Comment (mihaibudiu) [reply core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java:1962]: I think SqlToRelConverter should do the minimal amount of rewrites necessary to implement the program with the correct semantics, and no optimizations. The code is already complex, it's best to separate the concerns. But, clearly, the Rel representation produced should contain enough information to represent the programs, including some casts that cannot be inferred otherwise.

Comment (NobiGo) [reply core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java]: @mihaibudiu I have rollback the SqlToRelConverter. Now this PR includes CALCITE-6592. If you have free time, please review it.

File: core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml [hunk 7]
```diff
@@ -5248,10 +5287,15 @@ from emp]]>
...
-LogicalProject(EMPNO=[$0], EXPR$1=[CAST(IS NOT TRUE($11)):BOOLEAN])
-  LogicalJoin(condition=[=($9, $10)], joinType=[left])
-    LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], DEPTNO0=[$7])
-      LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+LogicalProject(EMPNO=[$0], EXPR$1=[OR(=($9, 0), AND(<($10, $9), null, IS NULL($13)), AND(IS NULL($13), >=($10, $9)))])
+  LogicalJoin(condition=[=($11, $12)], joinType=[left])
+    LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], $f0=[$9], $f1=[$10], DEPTNO0=[$7])
+      LogicalJoin(condition=[true], joinType=[inner])
+        LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+        LogicalAggregate(group=[{}], agg#0=[COUNT()], agg#1=[COUNT($0)])
```
NobiGo, line 5295: LogicalAggregate should be unnecessary, I will check this.

mihaibudiu: it looks like this query was decorrelated

mihaibudiu (approved):

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
NobiGo: @mihaibudiu After fixing the CALCITE-6599, I have rollbacked this change.

core/src/test/resources/sql/sub-query.iq
NobiGo: Has been fixed in https://issues.apache.org/jira/browse/CALCITE-5769. I have rolled back this change.

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
NobiGo: Done!

core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
NobiGo: This PR focuses on integer types. We can add it in another PR.

NobiGo:
If there are no other comments, I will merge it in the next few days.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
mihaibudiu: I have already approved these changes.

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -290,7 +290,7 @@ public static void checkActualAndReferenceFiles() {
    */
   @Test void testAsOfCast() {
     final String sql = "SELECT * "
-        + "FROM (SELECT deptno % 10 as m, CAST(deptno AS BIGINT) as deptno FROM dept) D\n"
+        + "FROM (SELECT CAST(deptno % 10 AS BIGINT) as m, CAST(deptno AS BIGINT) as deptno FROM dept) D\n"
```
NobiGo, line 293: @mihaibudiu If you have time please review this unit test again.
If I don't change it, `D.m = E.empno` will become `(cast(D.m as BIGINT) = E.empno)` so it throws an exception.

mihaibudiu: Maybe this is cause by the recent changes I made to the TypeCoercion which got merged into main: https://github.com/apache/calcite/pull/3998
In this case this change is fine.
The ASOF join seems a bit brittle. I have followed the spec from Snowflake for its definition, but I wonder whether we can improve that.

NobiGo: The exception is `ASOF JOIN condition must be a conjunction of equality comparisons`. The ON condition will converted to `CAST(D.M AS BIGINT) = E.EMPNO` if we don't change the D.M type. 
Before this PR, Integer and BigInt don't need to cast. But now it needs.
Before #3998, we don't need to make this change.

mihaibudiu: Adding the cast is fine for now. The spec for ASOF requires it, unfortunately.
Today the validator checks the shape of the ASOF join before optimizations.
But optimizations may change it. Ideally it should only be checked after optimizations.

NobiGo: If you don't think we should solve it in this PR, I am preparing to merge it.

mihaibudiu: yes, it is completely unrelated to this PR.

PR #2819, opened 2022-05-27
[CALCITE-5159] Postgres dialect should support implicit cast from string literal to array literal

dssysolyatin:


core/src/main/codegen/templates/Parser.jj
```diff
@@ -238,6 +238,39 @@ public class ${parser.class} extends SqlAbstractParserImpl
         return SqlStmtList();
     }
 
+    // Q: Is it the best way ???
```
julianhyde, line 241: I'd add a method `parseArray` to `SqlAbstractParserImpl`, as a top-level entry point to the parser. 

Also add a `parseArrayLiteral` method in `SqlParserUtil` similar to `parseTimestampLiteral`. The implementation will create a fresh parser then call `parseArray`.

If the validator needs to convert a char literal into an array literal, it can call that method.

julianhyde: I would not preemptively convert every char literal into an array literal. Which means that `class SqlArrayCharStringLiteral` is probably not needed.

dssysolyatin: Then I will need to pass ParserFactory to SqlValidatorImpl. If it is OK that validator will know about Parser then let's do like this.

julianhyde: To do it properly, I agree, we should pass parser factory and config in the validator config. But for now it's probably OK to use the default parser.

babel/src/test/resources/sql/postgresql.iq
```diff
@@ -0,0 +1,59 @@
+# postgresql.iq - Babel test for Postgresql dialect of SQL
```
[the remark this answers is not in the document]
dssysolyatin: Fixed.

core/src/main/codegen/templates/Parser.jj
dssysolyatin: Done.

core/src/main/resources/org/apache/calcite/runtime/CalciteResource.properties
```diff
@@ -319,4 +319,5 @@ InvalidInputForExtractXml=Invalid input for EXTRACT xpath: ''{0}'', namespace: '
 InvalidInputForExistsNode=Invalid input for EXISTSNODE xpath: ''{0}'', namespace: ''{1}''
 DifferentLengthForBitwiseOperands=Different length for bitwise operands: the first: {0,number,#}, the second: {1,number,#}
 NoOperator=No operator for ''{0}'' with kind: ''{1}'', syntax: ''{2}'' during JSON deserialization
+IllegalArrayExpression=Illegal array expression {0}
```
julianhyde, line 322: please put the line in the correct place in the file; putting it at the end tends to create merge conflicts

core/src/main/java/org/apache/calcite/sql/parser/SqlAbstractParserImpl.java
```diff
@@ -547,6 +547,11 @@ protected SqlCall createCall(
    */
   public abstract void setConformance(SqlConformance conformance);
 
+  /**
+   * Parse string to array literal.
```
julianhyde, line 551: should be 'Parses a string to an array literal'

core/src/main/java/org/apache/calcite/sql/parser/SqlParserUtil.java
```diff
@@ -205,6 +207,12 @@ public static SqlIntervalLiteral parseIntervalLiteral(SqlParserPos pos,
     return SqlLiteral.createInterval(sign, intervalStr, intervalQualifier, pos);
   }
 
+  public static SqlNode parseArrayLiteral(String s) throws Exception {
```
julianhyde, line 210: Needs a javadoc comment. Especially the fact that you use the default parser.

core/src/main/resources/org/apache/calcite/runtime/CalciteResource.properties
julianhyde: I don't see any tests for 'Illegal array expression'. Please add at least one test.

core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
```diff
@@ -721,4 +738,31 @@ boolean canImplicitTypeCast(List<RelDataType> types, List<SqlTypeFamily> familie
...
+        && SqlTypeUtil.isArray(targetType)
+        && operand instanceof SqlCharStringLiteral
+    ) {
+      try {
+        SqlNode arrayValue = SqlParserUtil.parseArrayLiteral(
+            ((SqlCharStringLiteral) operand).getValueAs(String.class));
+        call.setOperand(index, arrayValue);
+        updateInferredType(arrayValue, targetType);
+      } catch (Throwable e) {
+        return false;
```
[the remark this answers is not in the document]
julianhyde: Yes, `catch Error` is just as bad as `catch Throwable`.

Is it possible for `parseArrayLiteral` to declare a short list of exceptions that it may throw? Some of them will be user errors (i.e. the string is not valid format); internal errors (e.g. `NullPointerException`, `ArrayIndexOutOfBoundsException`) should be rethrown, not become `return false`.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -243,6 +243,17 @@ public class ${parser.class} extends SqlAbstractParserImpl
         return SqlStmtList();
     }
 
+    public SqlNode parseArray() throws SqlParseException {
+        switchTo(LexicalState.BQID);
+        try {
+          return ArrayLiteral();
+        } catch (ParseException ex) {
```
dssysolyatin, line 250: I used catch twice because javaCC can not parse `catch (ParseException | TokenMsgError ex)` for some reason.

core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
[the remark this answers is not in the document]
dssysolyatin: I managed to replace throwable by SqlParseException

dssysolyatin:
@julianhyde if you have time, please review this PR again. This PR is actively used in one of our project, I have to fix conflicts quite often. Thanks !

PR #2820, opened 2022-05-29
[CALCITE-35] Support parenthesized sub-clause in JOIN

julianhyde:


michaelmior:
LGTM although I only did a cursory check. Thanks!

PR #2822, opened 2022-05-29
[CALCITE-5170] Assertion error on range distribution creation

rkondakov:


libenchao (approved):
Good catch, LGTM.

chunweilei (approved):
+1. 
P.S. This PR is the same as https://github.com/apache/calcite/pull/2627.

PR #2823, opened 2022-05-31
[CALCITE-5147] Parser configured with BigQuery dialect cannot parse timestamp literal

wenruimeng:
Update the paser.jj file to support BigQuery time, timestamp, date, and interval literal with BigQuery compatible string literal.

Intend to fix the issue
https://issues.apache.org/jira/browse/CALCITE-5147


core/src/main/codegen/templates/Parser.jj
```diff
@@ -4528,17 +4528,50 @@ SqlLiteral DateTimeLiteral() :
         return SqlParserUtil.parseTimestampLiteral(p, s.end(this));
     }
 |
-    <DATE> { s = span(); } <QUOTED_STRING> {
-        return SqlParserUtil.parseDateLiteral(token.image, s.end(this));
-    }
+    <DATE> { s = span(); } ((<QUOTED_STRING> | <BIG_QUERY_QUOTED_STRING>)
```
julianhyde, line 4531: Can you use the same indentation for '{' and '|' as other code.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -4528,17 +4528,50 @@ SqlLiteral DateTimeLiteral() :
...
+      {
+          p = SqlParserUtil.stripQuotes(token.image, "'", "'", "\\'",
+              Casing.UNCHANGED);
+          return SqlParserUtil.parseDateLiteralWithoutQuote(p, s.end(this));
+      }
+     | <BIG_QUERY_DOUBLE_QUOTED_STRING>
```
julianhyde, line 4537: Why do we need this BIG_QUERY_DOUBLE_QUOTED_STRING case?

julianhyde: Oh I see. You opened a '(' several lines back and didn't indent. Confusing.

I suggest you add a rule (function) that matches <QUOTED_STRING>, <BIG_QUERY_QUOTED_STRING> and <BIG_QUERY_DOUBLE_QUOTED_STRING> and use it in all of these literals.

core/src/main/java/org/apache/calcite/sql/parser/SqlParserUtil.java
```diff
@@ -139,38 +139,45 @@ public static java.sql.Timestamp parseTimestamp(String s) {
...
     return SqlLiteral.createDate(d, pos);
   }
 
-  public static SqlTimeLiteral parseTimeLiteral(String s, SqlParserPos pos) {
+  public static SqlDateLiteral parseDateLiteral(String s, SqlParserPos pos) {
     final String dateStr = parseString(s);
+    return parseDateLiteralWithoutQuote(dateStr, pos);
+  }
+
+  public static SqlTimeLiteral parseTimeLiteralWithoutQuote(String dateStr, SqlParserPos pos) {
```
julianhyde, line 160: Why have a parameter `dateStr` for a function that parses a time literal? That seems confusing. Should it be `timeStr`? Or just stick to `s`.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -5420,8 +5441,8 @@ public void subTestIntervalYearPositive() {
    */
   public void subTestIntervalYearToMonthPositive() {
     // default precision
-    expr("interval '1-2' year to month")
-        .ok("INTERVAL '1-2' YEAR TO MONTH");
+    expr("interval '1:2' year to month")
+        .ok("INTERVAL '1:2' YEAR TO MONTH");
```
julianhyde, line 5445: Did you intend to change the syntax from hyphen to colon? I believe hyphen was correct already.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -191,10 +191,10 @@ class BabelParserTest extends SqlParserTest {
         .ok("SELECT TIMESTAMP '1969-07-20 00:00:00'");
     // PostgreSQL allows the following. We should too.
     sql("select ^timestamp '1969-07-20 1:2'^")
-        .fails("Illegal TIMESTAMP literal '1969-07-20 1:2': not in format "
+        .fails("Illegal TIMESTAMP literal 1969-07-20 1:2: not in format "
```
julianhyde, line 194: Removing the single-quotes from the error message doesn't seem to be an improvement.

core/src/main/java/org/apache/calcite/sql/parser/SqlParserUtil.java
wenruimeng: Good point. Let me change it. I keep it due to it's named as that in previous code.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
wenruimeng: Oh. This is a typo. Thanks for catching that.

core/src/main/codegen/templates/Parser.jj
wenruimeng: Will do

wenruimeng: @julianhyde Updated based on your feedback. Let me know whether it's what you expected.

core/src/main/codegen/templates/Parser.jj
wenruimeng: Addressed based on the rule added.

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
julianhyde: Yes, it's worth keeping the error message the same. You probably just need to edit `CalciteResource.java` and `CalciteResource.properties`.

wenruimeng: Updated based on your suggestion.

PR #2448, opened 2021-06-30
[CALCITE-4673] Correlate variables deduplication is not triggered when custom table function presented.

zstan:

PR #2825, opened 2022-06-03
CALCITE-4913 Correlated variables in a select list are not deduplicated

korlov42:


core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -2696,16 +2696,6 @@ protected void convertCollectionTable(
             validator().getValidatedNodeType(call),
             columnMappings);
 
-    final SqlValidatorScope selectScope =
```
korlov42, line 2699: this snippet were introduced by a [1] and addresses the very particular case with LogicalTableFunctionScan relation.

Current patch should solve the problem in general, thus this snippet is not required anymore

[1] https://github.com/apache/calcite/pull/2448

korlov42:
A few travis task failed with "ERROR: JAVA_HOME is set to an invalid directory". I assume it's not related to the changes

zabetak (approved):
LGTM, will merge soon!

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -664,6 +664,46 @@ LogicalProject(EMPNO=[$0], JOB=[$2])
...
+      LogicalProject($f0=[0])
+        LogicalFilter(condition=[AND(=($7, $cor1.DEPTNO), =($1, 'SMITH'), >($cor1.DEPTNO, 0))])
+          LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+]]>
+    </Resource>
+    <Resource name="sql">
+      <![CDATA[select e.deptno,
+  (select count(*) from emp where e.deptno > 0),
+  (select count(*) from emp where e.deptno > 0 and e.deptno < 10)
+from emp e]]>
```
zabetak, line 704: @korlov42 This is not the same query that is used in the test. To avoid this kind of problems in the future it is better to rely on the `./core/build/resources/test/org/apache/calcite/test/SqlToRelConverterTest_actual.xml` that is generated automatically by the framework. You can copy-paste the latter instead of manually editing the resources file yourself.

korlov42: got it... Thanks!

PR #2831, opened 2022-06-13
[CALCITE-5177] Query loses hint after decorrelation (minor follow-up)

rubenada:
CALCITE-5177 Query loses hint after decorrelation minor follow-up (following [this discussion](https://github.com/apache/calcite/pull/2829/files#r895412046))

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
@@ -475,6 +475,10 @@ public abstract class RelOptUtil {
    * or {@code newRel} directly if one of them are not {@link Hintable}
    */
   public static RelNode copyRelHints(RelNode originalRel, RelNode newRel, boolean filterHints) {
+    if (originalRel == newRel && !filterHints) {
+      return originalRel;
+    }
+
     if (originalRel instanceof Hintable
         && newRel instanceof Hintable
         && ((Hintable) originalRel).getHints().size() > 0) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -475,6 +475,10 @@ public static RelNode copyRelHints(RelNode originalRel, RelNode newRel) {
    * or {@code newRel} directly if one of them are not {@link Hintable}
    */
   public static RelNode copyRelHints(RelNode originalRel, RelNode newRel, boolean filterHints) {
+    if (originalRel == newRel && !filterHints) {
+      return originalRel;
```
danny0405, line 479: Do we need to make decision for `filterHints` ?

rubenada: If we have `filterHints` we can end up with a "new rel result" right? The result rel will be the originalRel (or newRel) with potentially fewer hints (if at least one of the original hints is filtered out). Maybe it is a far-fetched scenario that does not happen in practice, but I guess it is theoretically possible, so I'd prefer to be have this check.

danny0405: We better check first which code is calling the method with `filterHint` as true first.

rubenada: But this is a public method in RelOptUtil, hence a "public API" of Calcite. Theoretically, any "Calcite client code" (downstream project) could call this method with `filterHint` true.

rubenada: @danny0405 given the imminent code freeze and RC process for the next release, we need to hurry if we want to include this patch in the next Calcite version. wdyt?

rubenada: @danny0405 gentle reminder about this "to-be-merged PR" :)

rubenada: @danny0405 sorry to insist, but I think we should move forward to squeeze this (minor) change in the next release. I plan to merge the PR as it is in the next 24h (unless you see any issue that should "block" it). Thanks.

PR #2832, opened 2022-06-14
Remove unreachable condition in CsvTableScan

henneberger:
This removes an unreachable condition in CsvTableScan when checking if a table is a JsonTable. The table object can only take the form of a RelOptTable and not a Table.


chunweilei (approved):
Good catch. +1

chunweilei:
Thank you for your PR, @henneberger!

PR #2834, opened 2022-06-15
[CALCITE-5192] CodeGenerationBenchmark throws IllegalStateException

libenchao:


chunweilei (commented):
Should we add a test?

libenchao:
@chunweilei Thanks for your review.
> Should we add a test?

I'm not sure about this, as existing benchmarks all do not have tests, I think maybe we do not need it?

chunweilei:
> @chunweilei Thanks for your review.
> 
> > Should we add a test?
> 
> I'm not sure about this, as existing benchmarks all do not have tests, I think maybe we do not need it?

How do you find the bug if there is no test? We'd better add a test case to reproduce the error.

libenchao:
> How do you find the bug if there is no test? 
I spotted it because I was running the benchmark manually to verify some performance issues, the bug do not leads to failoure/exception, the benchmark will catch it, and print error logs.

From what I can see, the benchmarks are always run manually, and the result are used by hand, not by code (such as UT).

ubenchmark/src/jmh/java/org/apache/calcite/adapter/enumerable/CodeGenerationBenchmark.java
```diff
@@ -183,23 +184,25 @@ public void setup() {
...
-        info.cbe = cbe;
+        info.cbeSupplier = () -> {
+          ICompilerFactory compilerFactory;
+          try {
+            compilerFactory = CompilerFactoryFactory.getDefaultCompilerFactory(
+                CodeGenerationBenchmark.class.getClassLoader());
+          } catch (Exception e) {
+            throw new IllegalStateException(
+                "Unable to instantiate java compiler", e);
+          }
```
zabetak, line 195: Do we also need to include the `compilerFactory` in the `Supplier`? Whatever we include in the supplier becomes part of the benchmark (and not part of the setup) so we should reduce it to the minimal.

ubenchmark/src/jmh/java/org/apache/calcite/adapter/enumerable/CodeGenerationBenchmark.java
```diff
@@ -183,23 +184,25 @@ public void setup() {
...
+          IClassBodyEvaluator cbe = compilerFactory.newClassBodyEvaluator();
+          cbe.setClassName(info.classExpr.name);
+          cbe.setExtendedClass(Utilities.class);
+          cbe.setImplementedInterfaces(
+              plan.getRowType().getFieldCount() == 1
+                  ? new Class[]{Bindable.class, Typed.class}
+                  : new Class[]{ArrayBindable.class});
+          cbe.setParentClassLoader(EnumerableInterpretable.class.getClassLoader());
```
zabetak, line 203: Since the setup of the CBE is now becoming part of the benchmark maybe it would make more sense to extract it to a method and call it directly in the benchmark methods but I don't have strong feelings about this. Keeping it in a supplier should be fine as well.

zabetak (approved):
LGTM, left only some minor comments.

libenchao:
@zabetak Thanks for your review comments, they are helpful, I've addressed both of them.

libenchao:
merging

PR #2838, opened 2022-06-22
[CALCITE-5156] Alias within GroupingSets throws type mis-match exception

hannerwang:


rubenada:
@hannerwang Thanks for the PR. It seems there are some `autostyleJavaCheck FAILED` issues

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -6661,6 +6661,11 @@ static class ExtendedExpander extends Expander {
           SqlIdentifier sid = (SqlIdentifier) expr;
           final SqlIdentifier fqId = getScope().fullyQualify(sid).identifier;
           expr = expandDynamicStar(sid, fqId);
+        } else {
+          requireNonNull(expr, "expr");
```
rubenada, line 6665: Are we sure `expr` will always be non-null here?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -6661,6 +6661,11 @@ static class ExtendedExpander extends Expander {
...
+          SqlNode exprCopy = expr.clone(expr.getParserPosition());
```
rubenada, line 6666: Could you please explain a bit the logic behind the fix? (maybe it is worth to add a comment in the code).
As far as I can see, a clone of `expr` is created, using `expr`'s own position, so an exact copy is generated; and somehow using this copy instead of the original fixes the problem? (even though they are supposed to be equal, or I am missing something?)

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
hannerwang: Thanks ruben, I dove into SqlNodeList which is type of selectList and found its item could be nullable, but I couldn't found any possible nullable selectItem in parser.jj, so I think selectItem must be non-nullable, do you think so?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
hannerwang: Look the new test case, before the commit,  the **X** alias in CUBE will be expanded to 'empno / 2' expression without copy, so the 'empno / 2' in SELECT LIST and CUBE is same,  being type of **Integer not null**. 
When calcite derive the select item 'empno / 2 AS x', the type of 'empno / 2' will be **reused**, so the 'empno / 2 AS x' will be also **Integer not null**, this is not consistent with CUBE operator, GROUPING SETS () will make 'empno / 2 AS x' nullable. 
I make a copy in GROUP expression to make sure the 'empno / 2' type couldn't be **reused** and  the type of 'empno / 2' in SELECT LIST could be derive in correct AggregatingSelectScope.

hannerwang: In SqlValidatorImpl, node type will be cached into nodeToTypeMap(IdentityHashMap), so if calcite doesn't create a copy for 'empno / 2' in CUBE, the 'empno / 2' in SELECT LIST will **reuse** the type in CUBE which is **Integer not null** but it should be **nullable**.

rubenada: I understand. Thanks for the explanation and for your patience.
LGTM. As I said, I'm not the most expert in SqlValidator (but there are not may reviewers available at the moment, so I'll take responsibility for this PR).
This patch fixes the bug without any regression, so I'll merge it in the next 24h / 48h (if no objection appears).
@hannerwang could you please squash commits into a single one? Thanks.

rubenada (approved):

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
hannerwang: Hi @rubenada , sorry for bothering you, I have found the [CALCITE-5145](https://issues.apache.org/jira/browse/CALCITE-5145) is an another scenario for wrong type mis-match, I looked the [pr-2818](https://github.com/apache/calcite/pull/2818) @yingyuwang submitted and found a general solution for the two scenarios, please review when you are available.

yingyuwang: Hi @hannerwang, many thanks for taking CALCITE-5145 my pr-2818 into consideration. Your change LGTM and it looks much cleaner than I experimented in pr-2818!

rubenada: Thanks for your work @hannerwang ; I agree with @yingyuwang the new solution seems cleaner.
So, if I understand correctly, the current PR will fix [CALCITE-5045](https://issues.apache.org/jira/browse/CALCITE-5045) and also [CALCITE-5145](https://issues.apache.org/jira/browse/CALCITE-5145), right? In which case, could you please amend the commit message to reflect both fixes:
```
[CALCITE-5045] Alias within GroupingSets throws type mis-match exception
[CALCITE-5145] CASE statement within GROUPING SETS throws type mis-match exception
```

hannerwang: Thanks ruben, I changed the commit message without newline between the two jira titiles, It seems the commit message doesn't allow newline, is it good?

rubenada: Thanks, it's good.

PR #2843, opened 2022-06-26
[CALCITE-5178] Single column with ROW type generates wrong plan

libenchao:


core/src/main/java/org/apache/calcite/adapter/enumerable/JavaRowFormat.java
```diff
@@ -82,6 +83,11 @@ public enum JavaRowFormat {
         JavaTypeFactory typeFactory,
         RelDataType type) {
       assert type.getFieldCount() == 1;
+      RelDataType field0Type = type.getFieldList().get(0).getType();
```
JiajunBernoulli, line 86: Does the code can be simplified?
```      
      RelDataType field0Type = type.getFieldList().get(0).getType();
      // nested ROW type is always represented as array.
      return field0Type.getSqlTypeName() == SqlTypeName.ROW
          ? Object[].class
          : typeFactory.getJavaClass(field0Type);
    ```

libenchao: @JiajunBernoulli Thanks for your review. I don't have a preference for either way actually.

chunweilei (approved):
LGTM

libenchao:
merging now.

PR #2606, opened 2021-11-11
[CALCITE-4865] Extends Table function to support Polymorphic Table Function (PTF)

beyond1920:
## What is the purpose of the change

This pr aims to extend table function to support Polymorphic Table Function. 

## Brief change log

  - Update `Parser.jj` to support partition by clause and order by clause for input table with set semantics of PTF
  - Introduce `TableCharacteristics` which contains three characteristics of input table of table function
  - Update `SqlTableFunction` to add a method `tableCharacteristics`,  the method returns the table characteristics for the ordinal-th argument to this table function. Default return value is Optional.empty which means the ordinal-th argument is not table.
  - Introduce `SqlSetSemanticsTable` which represents input table with set semantics of Table Function, its `SqlKind` is `SET_SEMANTICS_TABLE`
  - Updates `SqlValidatorImpl` to validate only set semantic table of Table Function could have partition by and order by clause
  - Update `SqlToRelConverter#substituteSubQuery` to parse subQuery which represents set semantics table.

## Verifying this change

  - Add tests in `SqlToRelConveterTest`, `SqlValidatorTest`

PR #2844, opened 2022-06-27
[CALCITE-4865] Allow table functions to be polymorphic

beyond1920:

This pr aims to extend table function to support Polymorphic Table Function. 

## Brief change log

  - Update `Parser.jj` to support partition by clause and order by clause for input table with set semantics of PTF
  - Introduce `TableCharacteristics` which contains three characteristics of input table of table function
  - Update `SqlTableFunction` to add a method `tableCharacteristics`,  the method returns the table characteristics for the ordinal-th argument to this table function. Default return value is Optional.empty which means the ordinal-th argument is not table.
  - Introduce `SqlSetSemanticsTable` which represents input table with set semantics of Table Function, its `SqlKind` is `SET_SEMANTICS_TABLE`
  - Updates `SqlValidatorImpl` to validate only set semantic table of Table Function could have partition by and order by clause
  - Update `SqlToRelConverter#substituteSubQuery` to parse subQuery which represents set semantics table.

## Verifying this change

  - Add tests in `SqlToRelConveterTest`, `SqlValidatorTest`

### DIFF
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -1580,20 +1580,22 @@ SqlNode NamedRoutineCall(
 SqlNode TableParam() :
 {
     final Span s;
+    final SqlNodeList partitionList;
+    final SqlNodeList orderList;
     SqlNode tableRef;
-    SqlNodeList partitionList = SqlNodeList.EMPTY;
-    SqlNodeList orderList = SqlNodeList.EMPTY;
 }
 {
     { s = span(); }
     tableRef = ExplicitTable(getPos())
-    [
+    (
         <PARTITION> <BY>
         partitionList = SimpleIdentifierOrList()
-    ]
-    [
+    |   { partitionList = SqlNodeList.EMPTY; }
+    )
+    (
         orderList = OrderByInTableParam()
-    ]
+     |  { orderList = SqlNodeList.EMPTY; }
+    )
     {
         if (partitionList.isEmpty() && orderList.isEmpty()) {
           return tableRef;
@@ -1615,21 +1617,23 @@ SqlNode PartitionedQueryOrQueryOrExpr(ExprContext exprContext) :
     { return e; }
 }
 
-SqlNode PartitionedByAndOrderBy(SqlNode e):
+SqlNode PartitionedByAndOrderBy(SqlNode e) :
 {
     final Span s;
-    SqlNodeList partitionList = SqlNodeList.EMPTY;
-    SqlNodeList orderList = SqlNodeList.EMPTY;
+    final SqlNodeList partitionList;
+    final SqlNodeList orderList;
 }
 {
     { s = span(); }
-    [
+    (
         <PARTITION> <BY>
         partitionList = SimpleIdentifierOrList()
-    ]
-    [
+    |   { partitionList = SqlNodeList.EMPTY; }
+    )
+    (
         orderList = OrderByInTableParam()
-    ]
+     |  { orderList = SqlNodeList.EMPTY; }
+    )
     {
         if (partitionList.isEmpty() && orderList.isEmpty()) {
           return e;
@@ -1642,8 +1646,7 @@ SqlNode PartitionedByAndOrderBy(SqlNode e):
 
 SqlNodeList OrderByInTableParam() :
 {
-    List<SqlNode> list;
-    SqlNode e;
+    final List<SqlNode> list = new ArrayList<SqlNode>();
     final Span s;
 }
 {
@@ -1651,25 +1654,18 @@ SqlNodeList OrderByInTableParam() :
   <BY>
   (
       LOOKAHEAD(2)
-      <LPAREN> e = OrderItem()
-      {
-          list = startList(e);
-      }
+      <LPAREN> AddOrderItem(list)
       (
           LOOKAHEAD(2)
           <COMMA>
-          e = OrderItem()
-          {
-              list.add(e);
-          }
+          AddOrderItem(list)
       )*
       <RPAREN> {
           return new SqlNodeList(list, s.addAll(list).pos());
       }
   |
-      e = OrderItem()
+      AddOrderItem(list)
       {
-          list = startList(e);
           return new SqlNodeList(list, s.addAll(list).pos());
       }
   )
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -1014,7 +1014,9 @@ void AddArg0(List<SqlNode> list, ExprContext exprContext) :
     (
         e = Default()
     |
-        e = OrderedQueryOrExpr(firstExprContext)
+        e = TableArg()
+    |
+        e = PartitionedQueryOrQueryOrExpr(firstExprContext)
     )
     {
         if (name != null) {
@@ -1039,6 +1041,8 @@ void AddArg(List<SqlNode> list, ExprContext exprContext) :
         e = Default()
     |
         e = Expression(exprContext)
+    |
+        e = TableArg()
     )
     {
         if (name != null) {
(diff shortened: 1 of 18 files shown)


beyond1920:
This pr is a follow up the [pr#2606](https://github.com/apache/calcite/pull/2606).

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1653,6 +1657,108 @@ SqlNode NamedRoutineCall(
     }
 }
 
+/**
+ * Table parameter of a table function.
+ * The input table with set semantics may be partitioned/ordered on one or more columns.
+ */
+SqlNode TableArg() :
+{
```
danny0405, line 1665: TableArg -> TableParam

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1653,6 +1657,108 @@ SqlNode NamedRoutineCall(
...
+
+    { return e; }
+}
+
+SqlNode PartitionedByAndOrderBy(SqlNode e):
+{
+    final Span s;
+    SqlNodeList partitionList = SqlNodeList.EMPTY;
+    SqlNodeList orderList = SqlNodeList.EMPTY;
+}
```
danny0405, line 1707: The over window also has partition by order by clause, can that parse block be reused ?

beyond1920: The code could not be reused here because partition by clause and order by clause are different  from the partition by clause and order by clause in some other contexts. 
Test cases could cover these cases.
Based on the SQL standard, Order by clause of Polymorphic table function is different from other order by clause.
The list of order keys must be parenthesized if more than one column is listed.
Similarly, PARTITION BY clause with a list of zero or more columns. The list can always be enclosed in parentheses. If there is only one partitioning column, then the parentheses are optional.

core/src/main/java/org/apache/calcite/sql/SqlSetSemanticsTableOperator.java
```diff
@@ -0,0 +1,95 @@
...
+          orderList);
+    }
+  }
+
+  @Override public RelDataType deriveType(
+      SqlValidator validator,
+      SqlValidatorScope scope, SqlCall call) {
+    final List<SqlNode> operands = call.getOperandList();
+    assert operands.size() == 3;
+    RelDataType tableType = validator.deriveType(scope, operands.get(0));
```
danny0405, line 87: We already assert the operands number on construction.

core/src/main/java/org/apache/calcite/sql/SqlSetSemanticsTableOperator.java
```diff
@@ -0,0 +1,95 @@
...
+    assert tableType != null;
+    return tableType;
+  }
+
+  @Override public boolean argumentMustBeScalar(int ordinal) {
+    return ordinal != 0;
+  }
```
danny0405, line 94: Is this right ? Not sure whether the partition by /order by belongs to scalar argument ?

core/src/main/java/org/apache/calcite/sql/SqlTableFunction.java
```diff
@@ -30,4 +32,17 @@ public interface SqlTableFunction {
...
+   * Returns the table parameter characteristics for <code>ordinal</code>th
+   * parameter to this table function.
+   *
+   * <p>Returns <code>null</code> if the <code>ordinal</code>th argument is
+   * not table parameter or the <code>ordinal</code> is smaller than 0 or
+   * the <code>ordinal</code> is greater than or equals to the number of
+   * parameters.
+   */
+  default @Nullable TableCharacteristic tableCharacteristic(int ordinal) {
+    return null;
```
danny0405, line 46: Seems the functionality is only supported by window function ? Then moving this interface to `SqlWindowTableFunction` is more proper way.

core/src/main/java/org/apache/calcite/sql/TableCharacteristic.java
```diff
@@ -0,0 +1,333 @@
...
+      boolean pruneIfEmpty,
+      boolean passColumnsThrough) {
+    this.semantics = semantics;
+    this.pruneIfEmpty = pruneIfEmpty;
+    this.passColumnsThrough = passColumnsThrough;
+  }
+
+  public static TableCharacteristic withRowSemantic(
+      boolean columnsPassThrough) {
+    // Tables with row semantics are always effectively prune when empty.
```
danny0405, line 274: Is there any possibility we can add a `Builder` for this clazz ?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3380,6 +3384,66 @@ protected void validateFrom(
...
+        // parameter with row semantics, then it should not be with PARTITION BY OR ORDER BY.
+        SqlNode currentNode = call.operand(idx);
+        if (currentNode instanceof SqlCall) {
+          final SqlOperator op = ((SqlCall) currentNode).getOperator();
+          if (op == SqlStdOperatorTable.SET_SEMANTICS_TABLE) {
+            throwInvalidRowSemanticsTable(call, idx, (SqlCall) currentNode);
+          } else if (op == SqlStdOperatorTable.ARGUMENT_ASSIGNMENT) {
+            // Dig out the underlying operand
+            SqlNode realNode = ((SqlBasicCall) currentNode).operand(0);
+            if (realNode instanceof SqlCall) {
```
danny0405, line 3419: Can move the digging in the first so that the validation logic only need on.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3380,6 +3384,66 @@ protected void validateFrom(
     getNamespaceOrThrow(node, scope).validate(targetRowType);
   }
 
+  protected void validateTableFunction(SqlCall node, SqlValidatorScope scope,
+      RelDataType targetRowType) {
+    // Dig out real call; TABLE() wrapper is just syntactic.
+    SqlCall call = node.operand(0);
+    if (call.getOperator() instanceof SqlTableFunction) {
+      SqlTableFunction tableFunction = (SqlTableFunction) call.getOperator();
```
danny0405, line 3392: What kind of function it is if the SqlKind is `COLLECTION_TABLE` but not a `SqlTableFunction` ?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -1371,13 +1373,153 @@ private void substituteSubQuery(Blackboard bb, SubQuery subQuery) {
...
+    SqlNode query;
+    call = (SqlBasicCall) subQuery.node;
+    query = call.operand(0);
+    final SqlValidatorScope innerTableScope =
+        (query instanceof SqlSelect)
+            ? validator().getSelectScope((SqlSelect) query)
+            : null;
+    final Blackboard setSemanticsTableBb = createBlackboard(
+        innerTableScope, null, false);
+    final RelNode inputOfSetSemanticsTable = convertQueryRecursive(
```
danny0405, line 1401: What is the semantics to create black board with null scope ?

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -958,7 +960,17 @@ private static List<Integer> elseArgs(int count) {
 
   private static List<RexNode> convertOperands(SqlRexContext cx,
       SqlCall call, SqlOperandTypeChecker.Consistency consistency) {
-    return convertOperands(cx, call, call.getOperandList(), consistency);
+    List<SqlNode> operandList;
+    if (call.getOperator() instanceof SqlTableFunction) {
+      // skip set semantic table node of table function
```
danny0405, line 965: I see too many instance of for `SqlTableFunction`, can it has its own SqlKind ? Or can we use the COLLECTION_TABLE as its sql kind ?

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -4164,6 +4164,77 @@ void checkPeriodPredicate(Checker checker) {
...
+  }
+
+  @Test void testTableFunctionWithMultiplePartitionKeys() {
+    // test multiple partition keys for input table
+    final String sql =
+        "select * from table(topn(table orders partition by (orderId, productid), 3))";
+    final String expected = "SELECT *\n"
+        + "FROM TABLE(`TOPN`(((TABLE `ORDERS`) PARTITION BY `ORDERID`, `PRODUCTID`), 3))";
+    sql(sql).ok(expected);
+  }
```
danny0405, line 4189: We need some tests for multiple partition/ordering keys.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -1564,6 +1564,50 @@ public void _testLikeAndSimilarFails() {
...
+  @Test void testInvalidTableFunction() {
+    // A table function at most have one input table with row semantics
+    sql("select * from table(^invalid(table orders, table emp)^)")
+        .fails("A table function at most has one input table with row semantics."
+            + " Table function 'INVALID' has multiple input tables with row semantics");
+    // Only tables with set semantics may be partitioned
+    sql("select * from table(^score(table orders partition by productid)^)")
+        .fails("Only tables with set semantics may be partitioned."
+            + " Invalid PARTITION BY clause in the 0-th operand of table function 'SCORE'");
+    // Only tables with set semantics may be ordered
```
danny0405, line 1576: We need some tests for multiple partition/ordering keys.
We need some tests for multiple table params.

danny0405 (commented):
Thanks for the contribution @beyond1920 , i have left some comments.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
beyond1920: Currently, there is no valid case like this.
It requires argument of COLLECTION_TABLE must be a table function which is already checked by ProcedureNamespace#validateImpl.
SqlValidatorTest#testCollectionTable cover these cases.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
beyond1920: Scope is not null here, it's second argument (nameToNodeMap) is null.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
beyond1920: > We need some tests for multiple partition/ordering keys.

There are already exist cases to cover these in SqlValidatorTest#testTableFunctionWithTableParam.

> We need some tests for multiple table params.

Based on SQL standard, a table function at most has one input table with row semantics, there are already exist cases to cover these based on MockSqlOperatorTable#InvalidTableFunction.

I add some cases to cover a table function with multiple input tables with set semantics based on MockSqlOperatorTable#SimilarlityTableFunction.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
beyond1920: > We need some tests for multiple partition/ordering keys.

There are already exist cases to cover these, for example, SqlParserTest#testTableFunctionWithMultiplePartitionKeys, SqlParserTest#testTableFunctionWithMultipleOrderKeys


> We need some tests for multiple table params.

Based on SQL standard, a table function at most has one input table with row semantics, there are already exist cases to cover these based on MockSqlOperatorTable#InvalidTableFunction.

I add some cases to cover a table function with multiple input tables with set semantics based on MockSqlOperatorTable#SimilarlityTableFunction.

beyond1920:
@danny0405 Thanks for your review.
I've updated pull request to address your comments.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
danny0405: So the `instance of ` is not necessary here ?

beyond1920: I prefer to keep `instance of` here.
because `ProcedureNamespace#validateImpl` do a validation after this step.
If I remove `instance of` here, I need to add the `else` branch here to handle illegal cases.
WDYT?

beyond1920:
Hi, @danny0405 @julianhyde

beyond1920:
Merging

PR #2845, opened 2022-06-29
[CALCITE-4223] Metadata handlers for TableScan should see whether the RelOptTable implements the handler

chunweilei:


asolimando (approved):
LGTM

PR #2848, opened 2022-07-03
[CALCITE-5201] Improve SemiJoinRule to match Join's right input which is unique for Join keys

libenchao:


core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
```diff
@@ -232,6 +237,79 @@ default JoinToSemiJoinRuleConfig withOperandFor(Class<Join> joinClass,
...
+     */
+    @Value.Immutable
+    public interface JoinOnUniqueToSemiJoinRuleConfig extends SemiJoinRule.Config {
+      JoinOnUniqueToSemiJoinRuleConfig DEFAULT = ImmutableJoinOnUniqueToSemiJoinRuleConfig.of()
+          .withDescription("SemiJoinRule:unique")
+          .withOperandSupplier(b ->
+              b.operand(Project.class).oneInput(
+                  b2 -> b2.operand(Join.class).predicate(SemiJoinRule::isJoinTypeSupported).inputs(
+                      b3 -> b3.operand(RelNode.class).anyInputs(),
+                      b4 -> b4.operand(RelNode.class).anyInputs()
```
chunweilei, line 303: Is it necessary to capture the join's input?

core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
```diff
@@ -232,6 +237,79 @@ default JoinToSemiJoinRuleConfig withOperandFor(Class<Join> joinClass,
     }
   }
 
+  /**
+   * SemiJoinRule that matches a Join with a RelNode which is unique
+   * for Join's right keys.
+   *
+   * @see CoreRules#JOIN_ON_UNIQUE_TO_SEMI_JOIN */
+  public static class JoinOnUniqueToSemiJoinRule extends SemiJoinRule {
```
chunweilei, line 245: The comment should be 'SemiJoinRule that matches a Project on top of a Join xxxx'.

core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
```diff
@@ -42,14 +43,18 @@
...
 public abstract class SemiJoinRule
     extends RelRule<SemiJoinRule.Config>
     implements TransformationRule {
   private static boolean isJoinTypeSupported(Join join) {
     final JoinRelType type = join.getJoinType();
-    return type == JoinRelType.INNER || type == JoinRelType.LEFT;
+    return type == JoinRelType.INNER
+        || type == JoinRelType.LEFT
+        || type == JoinRelType.SEMI;
   }
```
chunweilei, line 58: If it is already a semi join, do we still need to convert it?

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableSort.java
```diff
@@ -72,11 +72,7 @@ public static EnumerableSort create(RelNode child, RelCollation collation,
     final BlockBuilder builder = new BlockBuilder();
     final EnumerableRel child = (EnumerableRel) getInput();
     final Result result = implementor.visitChild(this, 0, child, pref);
-    final PhysType physType =
-        PhysTypeImpl.of(
-            implementor.getTypeFactory(),
-            getRowType(),
-            result.format);
+    final PhysType physType = result.physType;
```
rubenada, line 75: This change does not seem related to CALCITE-5201

core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
```diff
@@ -232,6 +237,79 @@ default JoinToSemiJoinRuleConfig withOperandFor(Class<Join> joinClass,
...
+      final RelNode left = call.rel(2);
+      final RelNode right = call.rel(3);
+
+      // return if right's fields are used.
+      final ImmutableBitSet bits =
+          RelOptUtil.InputFinder.bits(project.getProjects(), null);
+      final ImmutableBitSet rightBits =
+          ImmutableBitSet.range(left.getRowType().getFieldCount(),
+              join.getRowType().getFieldCount());
+      if (bits.intersects(rightBits)) {
```
rubenada, line 264: Shall we move the applicability conditions (`!bits.intersects(rightBits)` and `unique != null && unique`) to the `matches` method instead of `onMatch` ?

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableLimit.java
```diff
@@ -96,11 +96,7 @@ public static EnumerableLimit create(final RelNode input, @Nullable RexNode offs
     final BlockBuilder builder = new BlockBuilder();
     final EnumerableRel child = (EnumerableRel) getInput();
     final Result result = implementor.visitChild(this, 0, child, pref);
-    final PhysType physType =
-        PhysTypeImpl.of(
-            implementor.getTypeFactory(),
-            getRowType(),
-            result.format);
+    final PhysType physType = result.physType;
```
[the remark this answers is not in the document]
libenchao: Yes, you are right. I also struggled with this change for some time.
Without this change, `JdbcTest#testDynamicParameterInLimitOffset` will fail. The reason is that in this PR, we strengthened the ability to transform to semi join, and this test case is the affected one. The physical implementation for semi join and inner join is different in `EnumerableHashJoin`, and this leads to the test failure.

I also thought about to leave it as a follow-up issue, but I found it's very hard to reproduce this error without this PR. Hence I put the fix in this PR.
What do you think about this?

core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
rubenada: Not sure about this... can we have a SEMI with an aggregate inside in a real scenario?

The code that you mention seems to be "legacy code" that we forgot to clean-up when CALCITE-4623 was applied. In practice, the existing `JoinToSemiJoinRule` should not match SEMI because it is not included on its `isJoinTypeSupported` method (which was changed via CALCITE-4623 to consider just INNER and LEFT):
```
  private static boolean isJoinTypeSupported(Join join) {
    final JoinRelType type = join.getJoinType();
    return type == JoinRelType.INNER || type == JoinRelType.LEFT;
  }
```

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableLimit.java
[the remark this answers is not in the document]
libenchao: The plan before this PR:
```xml
EnumerableLimit(offset=[?1], fetch=[?0])
  EnumerableSort(sort0=[$3], dir0=[ASC])
    EnumerableCalc(expr#0..6=[{inputs}], empid=[$t2], deptno=[$t3], name=[$t4], salary=[$t5], commission=[$t6])
      EnumerableHashJoin(condition=[AND(=($0, $2), =($1, $3))], joinType=[inner])
        EnumerableAggregate(group=[{0}], EXPR$0=[MAX($1)])
          EnumerableTableScan(table=[[hr, emps]])
        EnumerableTableScan(table=[[hr, emps]])
```
after this PR:
```xml
EnumerableSort(offset=[?1], fetch=[?0])
  EnumerableSort(sort0=[$3], dir0=[ASC])
    EnumerableHashJoin(condition=[AND(=($0, $5), =($1, $6))], joinType=[semi])
      EnumerableTableScan(table=[[hr, emps]])
      EnumerableAggregate(group=[{0}], EXPR$0=[MAX($1)])
        EnumerableTableScan(table=[[hr, emps]])
```

`EnumerableHashJoin(joinType=[semi])`'s physical type is `Employee`. However, `EnumerableSort` and `EnumerableSort`'s physical type is `Record5_0`. That's why after this PR, the test will fail.

> This change seems a bit ad-hoc. Why is it applied only on {{EnumerableLimit}} and {{EnumerableSort}}, but we keep the "old code" in other operators like {{EnumerableLimitSort}} or {{EnumerableAggregate}}?

You are right, we should check all places, or figure out another way to fix this.

This is indeed orthogonal with current PR, maybe we can leave it as a follow-up issue. I've considered this way before, there may be two ways to to this:
* Remove `JoinOnUniqueToSemiJoinRule` from `RelOptRules#ABSTRACT_RELATIONAL_RULES` and `Prorams#RULE_SET`
* Disable `JdbcTest#testDynamicParameterInLimitOffset` and link the follow-up issue id to it  

What do you think about these two ways?

core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
libenchao: It's good to know there is already a discussion about this before in CALCITE-4623.

Consider this case, we have a plan below:
```xml
  LogicalAggregate(group=[{0}], agg#0=[MIN($1)])
    LogicalProject(DEPTNO=[$0], $f0=[true])
```
After `JoinOnUniqueToSemiJoinRule`, it would be transformed to:
```xml
LogicalJoin(condition=[=($7, $9)], joinType=[semi])
  LogicalAggregate(group=[{0}], agg#0=[MIN($1)])
    LogicalProject(DEPTNO=[$0], $f0=[true])
```
And we can still use `JoinToSemiJoinRule` to remove the `Aggregate`:
```xml
LogicalJoin(condition=[=($7, $9)], joinType=[semi])
  LogicalProject(DEPTNO=[$0], $f0=[true])
```

In many cases, we would not expect the rules firing orders, especially in `HepPlanner`, hence the above transforming order is valid and should be valid IMHO.

chunweilei: It seems a bit weird to do this optimization in a rule named JoinToSemiJoinRule. Maybe it's better to move this optimization to another place.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableLimit.java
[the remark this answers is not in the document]
libenchao: @rubenada I removed these changes in this PR, and logged [CALCITE-5229](https://issues.apache.org/jira/browse/CALCITE-5229) as a follow-up work.

core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
libenchao: @chunweilei I figured out another way to do this. In `JoinOnUniqueToSemiJoinRule`, if we found that right is `Aggregate`, we can fallback to `ProjectToSemiJoinRule`.

libenchao:
@chunweilei @rubenada Thanks for the review, I've addressed the comments, do you have other concerns?

rubenada:
Thanks @libenchao LGTM.
Have you run (at least locally) slowTests? (just in case there's a slow test that needs to be adapted with this patch)

core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
```diff
@@ -232,6 +234,87 @@ default JoinToSemiJoinRuleConfig withOperandFor(Class<Join> joinClass,
...
+    }
+
+    @Override public void onMatch(RelOptRuleCall call) {
+      final Project project = call.rel(0);
+      final Join join = call.rel(1);
+      final RelNode left = call.rel(2);
+      final RelNode right = call.rel(3);
+
+      // Same with ProjectToSemiJoinRule
+      if (right instanceof Aggregate) {
```
rubenada, line 269: Should we rather add in the new rule's config a predicate to not match Aggregates on the join RHS:
```
...
b2 -> b2.operand(Join.class).predicate(SemiJoinRule::isJoinTypeSupported).inputs(
  b3 -> b3.operand(RelNode.class).anyInputs(),
  b4 -> b4.operand(RelNode.class).predicate(n -> !(n instanceof Aggregate)).anyInputs()
)))
```
Rather than matching it with two rules (the old and the new) and creating two times the same result?

libenchao: Good idea!

libenchao:
@rubenada I've addressed your last comment, and slowTests passed locally.

core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
rubenada: Thanks for the reactivity @libenchao !
Nitpick: perhaps adding a comment in the predicate, to clarify that the new rule does not match aggregates on the join RHS because this case is already covered by ProjectToSemiJoinRule ?

libenchao: sounds good, added.

rubenada: Thanks

rubenada:
Thanks for the patch @libenchao , no more comments from my side. LGTM

chunweilei (approved):
LGTM

libenchao:
@rubenada @chunweilei Thanks for the review, merging.

core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
```diff
@@ -232,6 +234,84 @@ default JoinToSemiJoinRuleConfig withOperandFor(Class<Join> joinClass,
...
+          builder.push(right);
+          builder.join(JoinRelType.SEMI, join.getCondition());
+          break;
+        case LEFT:
+          builder.push(left);
+          break;
+        default:
+          throw new AssertionError(join.getJoinType());
+        }
+        builder.project(project.getProjects());
```
sdreynolds, line 286: :doh: this doesn't carry through the field aliases like it does in the other [rule](https://github.com/libenchao/calcite/blob/65e07f71c3e98115f114ee05986daa6baab28b96/core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java#L127)
```java
builder.project(project.getProjects(), project.getRowType().getFieldNames());
```

libenchao: @sdreynolds thanks for catching this, I noticed that you have opened CALCITE-5391 to track this, let's go to there for further discussions.

core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
```diff
@@ -232,6 +234,84 @@ default JoinToSemiJoinRuleConfig withOperandFor(Class<Join> joinClass,
...
+    @Override public void onMatch(RelOptRuleCall call) {
+      final Project project = call.rel(0);
+      final Join join = call.rel(1);
+      final RelNode left = call.rel(2);
+      final RelNode right = call.rel(3);
+
+      final JoinInfo joinInfo = join.analyzeCondition();
+      final RelOptCluster cluster = join.getCluster();
+      final RelMetadataQuery mq = cluster.getMetadataQuery();
+      final Boolean unique = mq.areColumnsUnique(right, joinInfo.rightSet());
```
yashmayya, line 271: This doesn't take into account any `nonEquiConditions` on the join? We also aren't verifying beforehand whether the join is an equi-join as is being done for the other two semi-join rules `ProjectToSemiJoinRule` and `JoinToSemiJoinRule` here - https://github.com/apache/calcite/blob/4def07066b6cb370e37739985b19f6e6b408addb/core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java#L96-L98

@libenchao do you think this is a potential bug in the `JoinOnUniqueToSemiJoinRule`?

PR #2853, opened 2022-07-08
[CALCITE-4632] SARG datatype should be less restictive than any of the input types

abhishek-das-gupta:
Instead of first row expression's datatype, SARG datatype should consider datatype
of every row expression present in the `IN` and `BETWEEN` clause and should return the most general datatype from a
set of datatypes.


core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1355,10 +1356,14 @@ public RexNode makeIn(RexNode arg, List<? extends RexNode> ranges) {
     if (areAssignable(arg, ranges)) {
       final Sarg sarg = toSarg(Comparable.class, ranges, RexUnknownAs.UNKNOWN);
       if (sarg != null) {
-        final RexNode range0 = ranges.get(0);
+        final List<RelDataType> types = ranges.stream()
+            .map(RexNode::getType)
+            .collect(Collectors.toList());
+        RelDataType searchType = typeFactory.leastRestrictive(types);
+        searchType = searchType == null ? ranges.get(0).getType() : searchType;
```
libenchao, line 1363: If`searchType == null` is `true`, do we really need to proceed with `ranges.get(0).getType()`?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1252,6 +1253,26 @@ private static String toSql(RelNode root, SqlDialect dialect,
     relFn(relFn).optimize(rules, null).ok(expected);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-4632">[CALCITE-4632]
+   * Find the least restrictive datatype for SARG</a>. */
+  @Test void testLeastRestrictiveTypeForSarg() {
```
libenchao, line 1259: As discussed in the JIRA, the root cause is not from `RelToSqlConverter`, so I'm not sure this is the right place to add tests. How about we add more tests in `RexBuilderTest` and assert the `RexNode`'s type returned from `RexBuilder#makeIn`?

libenchao (commented):
@abhishek-das-gupta Thanks for your PR, generally it looks great, I've left some minor comments.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
abhishek-das-gupta: ACK

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
abhishek-das-gupta: All the usages of `RelDataTypeFactory.leastRestrictive` handles the edge case where the return type is null. So I thought about defaulting to how it get handles now in current master.  Is it really not necessary to handle this case? Can you provide your thoughts on this?

libenchao: The question is: "is it correct to use `ranges.get(0).getType()`?". If it's not correct, then we'd better throw exception. If we decide to keep it, then we should know in what cases will `searchType == null` is true, and we still want to use `ranges.get(0).getType()` as the final type.

abhishek-das-gupta: Thanks @libenchao .  I think if we don't find a  least restrictive `searchType`  among the `types`, then selecting `ranges.get(0).getType()` is wrong.  Since it might eventually fail at Rel to Sql conversion.  Hence throwing an exception

libenchao:
I found that there is a similar work before: [CALCITE-4888](https://issues.apache.org/jira/browse/CALCITE-4888).   
@beyond1920 What do you think about this? I think we can add you as a co-author of this issue, and resolve both CALCITE-4888 and CALCITE-4632 together.

abhishek-das-gupta:
@libenchao I agree with your comment. Please let when can resolve this issue, If nothing more needs to be done.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1258,6 +1259,26 @@ private static String toSql(RelNode root, SqlDialect dialect,
...
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .scan("EMP")
+        .filter(
+            b.or(b.isNull(b.field("COMM")),
+                  b.in(
+                  b.field("COMM"),
+                  b.literal(new BigDecimal("1.0")), b.literal(new BigDecimal("20000.0")))
+            )
+        )
```
julianhyde, line 1274: Please remove newlines before close-parentheses. This is not Calcite's coding style.

core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
```diff
@@ -624,6 +624,29 @@ private void checkDate(RexLiteral literal) {
...
+    RexNode left = rexBuilder.makeInputRef(decimalType, 0);
+    final RexNode literal1 = rexBuilder.makeExactLiteral(new BigDecimal("1.0"));
+    final RexNode literal2 = rexBuilder.makeExactLiteral(new BigDecimal("20000.0"));
+
+    RexNode inCall = rexBuilder.makeIn(left, ImmutableList.of(literal1, literal2));
+    assertThat(inCall.getKind(), is(SqlKind.SEARCH));
+
+    final RexNode sarg = ((RexCall) inCall).operands.get(1);
+    final RelDataType leastRestrictiveType =
+        typeFactory.leastRestrictive(ImmutableList.of(literal1.getType(), literal2.getType()));
```
julianhyde, line 643: Can you assert what the least restrictive type is? It will make the test more concrete.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
abhishek-das-gupta: Sorry about this. Addressed this.

core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
abhishek-das-gupta: ACK

libenchao: I'm not sure your new commit addressed Julian's comment, but per my understanding, I prefer the below way (and this is how I understand Julian's comment), what do you think?

```java
    RelDataType expected = typeFactory.createSqlType(SqlTypeName.DECIMAL, 6, 1);
    assertEquals(sarg.getType(), expected);
```

abhishek-das-gupta: Yes, i think I mistinterpreted it. Apologies.

libenchao (approved):
LGTM, will merge if there is no more objections.

libenchao:
@abhishek-das-gupta I found that `RexBuilder#makeBetween` also may have the same problem, could you help to confirm this, and fix it if it exists? (In this issue, or a follow-up issue, it's up to you, I'm ok with both way)

PR #2860, opened 2022-07-23
[CALCITE-5217] Add support for INTERVAL qualifier for Firebolt

aymeric-dispa:
Adds support for the INTERVAL qualifier for [Firebolt](https://www.firebolt.io/). 

Ticket: https://issues.apache.org/jira/browse/CALCITE-5217

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/FireboltSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/FireboltSqlDialect.java
@@ -16,14 +16,17 @@
  */
 package org.apache.calcite.sql.dialect;
 
+import org.apache.calcite.avatica.util.TimeUnit;
 import org.apache.calcite.avatica.util.TimeUnitRange;
 import org.apache.calcite.config.NullCollation;
 import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeSystem;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/FireboltSqlDialect.java
```diff
@@ -196,4 +197,18 @@ public FireboltSqlDialect(Context context) {
...
+      SqlIntervalLiteral literal, int leftPrec, int rightPrec) {
+    SqlIntervalLiteral.IntervalValue interval =
+        literal.getValueAs(SqlIntervalLiteral.IntervalValue.class);
+    writer.keyword("INTERVAL");
+    writer.print("'");
+    if (interval.getSign() == -1) {
+      writer.print("-");
+    }
+    writer.literal(interval.getIntervalLiteral());
+    writer.print(interval.getIntervalQualifier().toString());
```
JiajunBernoulli, line 211: `toString` is AnsiSqlDialect.DEFAULT.  

I prefer to use `interval.getIntervalQualifier().unparse(writer, leftPrec, rightPrec);`

aymeric-dispa: I just realised that the implementation was a bit too simplistic and was not handling ranges (I just pushed a change)

jbalint (approved):

PR #2864, opened 2022-08-01
Ravtdbq 60 table name and column name are same

akash-pawar1:


akash-pawar1:
closing as target branch is incorrect

PR #2868, opened 2022-08-09
[CALCITE-5230] Fix PERCENTILE_DISC return type derivation

itiels:
Changed the return type of PERCENTILE_DISC to the type of the `order_by_expression`.

Ticket: [CALCITE-5230](https://issues.apache.org/jira/browse/CALCITE-5230)

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlWithinGroupOperator.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlWithinGroupOperator.java
@@ -19,9 +19,13 @@ package org.apache.calcite.sql;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.type.OperandTypes;
 import org.apache.calcite.sql.type.ReturnTypes;
+import org.apache.calcite.sql.type.SqlTypeUtil;
 import org.apache.calcite.sql.validate.SqlValidator;
+import org.apache.calcite.sql.validate.SqlValidatorNamespace;
 import org.apache.calcite.sql.validate.SqlValidatorScope;
 import org.apache.calcite.sql.validate.SqlValidatorUtil;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.util.Objects;
 
@@ -62,12 +66,13 @@ public class SqlWithinGroupOperator extends SqlBinaryOperator {
       SqlValidatorScope operandScope) {
     assert call.getOperator() == this;
     assert call.operandCount() == 2;
+
     final SqlValidatorUtil.FlatAggregate flat = SqlValidatorUtil.flatten(call);
-    if (!flat.aggregateCall.getOperator().isAggregator()) {
-      throw validator.newValidationError(call,
-          RESOURCE.withinGroupNotAllowed(
-              flat.aggregateCall.getOperator().getName()));
+    final SqlOperator operator = flat.aggregateCall.getOperator();
+    if (!operator.isAggregator()) {
+      throw validator.newValidationError(call, RESOURCE.withinGroupNotAllowed(operator.getName()));
     }
+
     for (SqlNode order : Objects.requireNonNull(flat.orderList)) {
       Objects.requireNonNull(validator.deriveType(scope, order));
     }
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -988,4 +990,15 @@ public static SqlCall stripSeparator(SqlCall call) {
       return relDataType;
     }
   };
+
+  public static final SqlReturnTypeInference PERCENTILE_DISC = opBinding -> {
+    if (opBinding instanceof SqlWithinGroupOperator.PercentileDiscCallBinding) {
```
julianhyde, line 995: This 'if .. instanceof ... else if ... instanceof' pattern isn't good. How about adding a method `getCollationExpression()` to `SqlCallBinding`, with a default implementation that throws. Then you can remove both `getCollationType` methods.

core/src/main/java/org/apache/calcite/sql/SqlWithinGroupOperator.java
```diff
@@ -79,7 +81,56 @@ public SqlWithinGroupOperator() {
...
+      SqlCall call,
+      SqlValidator validator,
+      SqlOperator operator
+  ) {
+    if (!operator.isAggregator()) {
+      throw validator.newValidationError(call,
+          RESOURCE.withinGroupNotAllowed(
+              operator.getName()));
+    }
+  }
```
julianhyde, line 115: I would remove the `throwIfNotAggregate` function. Methods that always throw are confusing to the reader, and leave regions of code that the compiler doesn't know are unreachable.

core/src/main/java/org/apache/calcite/sql/SqlWithinGroupOperator.java
```diff
@@ -79,7 +81,56 @@ public SqlWithinGroupOperator() {
...
+      SqlCallBinding opBinding =
+          new PercentileDiscCallBinding(validator, scope, inner, getCollationColumn(call));
+      inner.getOperator().checkOperandTypes(opBinding, true);
+      RelDataType ret = inner.getOperator().inferReturnType(opBinding);
+      validator.setValidatedNodeType(inner, ret);
+      return ret;
+    } else {
+      return validateOperands(validator, scope, call);
+    }
+  }
```
julianhyde, line 99: There's a lot of duplicated code here. Maybe if you replace `new SqlCallBinding(validator, scope, call)` in the base method with a protected method, you can save all that duplication?

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
itiels: I can't implement `getCollationExpression` in `PercentileDiscAggCallBinding`, as I don't have the collation expression there. 
(maybe i'm missing something). I can however implement `getCollationType` which will remove the need to use the `instance of` pattern here.

core/src/main/java/org/apache/calcite/sql/SqlWithinGroupOperator.java
itiels: I don't think we can, because the validation/derivation is needed to be done in terms of the `inner` operator. even if we supply the correct `SqlCallBinding` to `validateOperands`, it will still be done in the context of the outer operator(`SqlWithinGroupOperator`)

libenchao (commented):
@itiels Thanks for your PR, generally it looks good now, I only left one minor comment.

Besides, `PERCENTILE_COUT` also should be fixed, we can do it in this issue, or a follow-up issue.

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -2303,7 +2303,7 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
    */
   public static final SqlAggFunction PERCENTILE_DISC =
       SqlBasicAggFunction
-          .create(SqlKind.PERCENTILE_DISC, ReturnTypes.DOUBLE,
+          .create(SqlKind.PERCENTILE_DISC, ReturnTypes.PERCENTILE_DISC,
```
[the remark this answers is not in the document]
itiels: Thanks, fixed it

itiels:
Thanks @libenchao for the review. 

I've fixed the comment.
As for `PERCENTILE_CONT`, based on conversation with Julian in the Jira ticket, there is nothing to be done.

Is there anything else that I need/should to do for this PR?

libenchao:
> As for PERCENTILE_CONT, based on conversation with Julian in the Jira ticket, there is nothing to be done.

I'm not sure about this conclusion, I've checked the SQL standard (2011), 10.9:

> If PERCENTILE_CONT is specified, then:
> 1) Let ROW0 be the greatest exact numeric value with scale 0 (zero) that is less than or equal
> to NVE*(N–1). Let ROWLIT0 be a <literal> representing ROW0.
> 2) Let ROW1 be the least exact numeric value with scale 0 (zero) that is greater than or equal to NVE*(N–1). Let ROWLIT1 be a <literal> representing ROW1.
> 3) Let FACTOR be an <approximate numeric literal> representing NVE*(N–1)–ROW0.
> 4) The result is the result of the <scalar subquery>
>      ( WITH TEMPTABLE(X, Y) AS
>            ( SELECT ROW_NUMBER()
> OVER (ORDER BY WSP) - 1, TXCOLNAME
> FROM TXANAME )
> SELECT CAST ( T0.Y + FACTOR * (T1.Y - T0.Y) AS DT ) FROM TEMPTABLE T0, TEMPTABLE T1
> WHERE T0.ROWNUMBER = ROWLIT0
> AND T1.ROWNUMBER = ROWLIT1 )

We can see that the standard is using `CAST ( T0.Y + FACTOR * (T1.Y - T0.Y) AS DT )` as the result for `PERCENTILE_COUT`, whose type is the same as the sort key.

And also I've checked some databases, and found: [oracle](https://docs.oracle.com/cd/B19306_01/server.102/b14200/functions110.htm) and [postgrelsql](https://www.postgresql.org/docs/9.4/functions-aggregate.html) are consistent with the standard.

There are some databases, such as [server sql](https://learn.microsoft.com/en-us/sql/t-sql/functions/percentile-cont-transact-sql?view=sql-server-ver16), are declaring the return type which is not consistent with the standard.

Hence, IMO, we should make it consistent with the standard by default. Of course if could be extensible for some dialects, if we want to achieve that, maybe the proper place to add the extension is in `RelDataTypeSystem`

libenchao:
@itiels Sorry to ping you.
We plan to release 1.33 recently, it would be great if we can make it in this release. Do you have any comments about my last comment?

itiels:
HI @libenchao, 
Sorry for the late reply.

I agree with your conclusion, however I prefer to do that in a separate PR if it's ok with you.

libenchao:
I'm ok with finishing it to another separate issue.
This PR now looks good to me. The CI failure seems unrelated, I'll merge it.

libenchao (approved):

julianhyde:
@libenchao when you merge, please change the commit message to something that doesn’t start with “fix”

libenchao:
@julianhyde Very thanks for the reminder 👍 , I did plan to do that while merging.

PR #2876, opened 2022-08-17
[CALCITE-5240] Enhance MaterializedViewRule so that it applies to rol…

twdsilva:
…lup view for queries that contain a predicate on the rollup column

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
@@ -248,7 +248,7 @@ public abstract class MaterializedViewAggregateRule<C extends MaterializedViewAg
       @Nullable Project topProject,
       RelNode node,
       BiMap<RelTableRef, RelTableRef> queryToViewTableMapping,
-      EquivalenceClasses viewEC, EquivalenceClasses queryEC, boolean pushQueryFilter) {
+      EquivalenceClasses viewEC, EquivalenceClasses queryEC, boolean viewRewritten) {
     Aggregate aggregate = (Aggregate) node;
 
     // Our target node is the node below the root, which should have the maximum
@@ -274,7 +274,7 @@ public abstract class MaterializedViewAggregateRule<C extends MaterializedViewAg
     if (!compensationColumnsEquiPred.isAlwaysTrue()) {
       RexNode newCompensationColumnsEquiPred = rewriteExpression(rexBuilder, mq,
           target, target, queryExprs, queryToViewTableMapping, queryEC, false,
-          compensationColumnsEquiPred);
+          compensationColumnsEquiPred, viewRewritten);
       if (newCompensationColumnsEquiPred == null) {
         // Skip it
         return null;
@@ -285,7 +285,7 @@ public abstract class MaterializedViewAggregateRule<C extends MaterializedViewAg
     if (!otherCompensationPred.isAlwaysTrue()) {
       RexNode newOtherCompensationPred = rewriteExpression(rexBuilder, mq,
           target, target, queryExprs, queryToViewTableMapping, viewEC, true,
-          otherCompensationPred);
+          otherCompensationPred, viewRewritten);
       if (newOtherCompensationPred == null) {
         // Skip it
         return null;
@@ -298,30 +298,11 @@ public abstract class MaterializedViewAggregateRule<C extends MaterializedViewAg
                 otherCompensationPred)));
 
     // Generate query rewriting.
-    RelNode rewrittenPlan = null;
-    if (pushQueryFilter) {
+    RelNode rewrittenPlan;
+    if (viewRewritten) {
       rewrittenPlan = target.accept(
-          // Push the queryCompensationPred down into any filter that is present in the plan
-          new RelShuttleImpl() {
-            @Override public RelNode visit(LogicalFilter filter) {
-              RexNode condition =
-                  RexUtil.flatten(rexBuilder,
-                      rexBuilder.makeCall(
-                          SqlStdOperatorTable.AND,
-                          filter.getCondition(),
-                          queryCompensationPred));
-              return filter.copy(filter.getTraitSet(), filter.getInput(), condition);
-            }
-
-            @Override public RelNode visit(RelNode other) {
-              if (other instanceof RelSubset) {
-                RelSubset relSubset = (RelSubset) other;
-                return relSubset.getBestOrOriginal().accept(this);
-              } else {
-                return visitChildren(other);
-              }
-            }
-          });
+          new QueryCompensationPredicateHandler(
+          rexBuilder, queryCompensationPred));
     } else {
       rewrittenPlan = relBuilder
           .push(target)
@@ -335,6 +316,38 @@ public abstract class MaterializedViewAggregateRule<C extends MaterializedViewAg
                   ImmutableList.of(rewrittenPlan))));
     }
     return aggregate.copy(aggregate.getTraitSet(), ImmutableList.of(rewrittenPlan));
+  }
+
+  /**
+   * Push the queryCompensationPred down into any filter that is present in the plan.
+   */
+  private static class QueryCompensationPredicateHandler extends RelShuttleImpl {
+    private final RexBuilder rexBuilder;
+    private final RexNode queryCompensationPred;
+
+    QueryCompensationPredicateHandler(RexBuilder rexBuilder, RexNode queryCompensationPred) {
+      this.rexBuilder = rexBuilder;
+      this.queryCompensationPred = queryCompensationPred;
+    }
+
+    @Override public RelNode visit(LogicalFilter filter) {
+      RexNode condition =
+          RexUtil.flatten(rexBuilder,
+              rexBuilder.makeCall(
+                  SqlStdOperatorTable.AND,
+                  filter.getCondition(),
+                  queryCompensationPred));
+      return filter.copy(filter.getTraitSet(), filter.getInput(), condition);
+    }
+
+    @Override public RelNode visit(RelNode other) {
+      if (other instanceof RelSubset) {
+        RelSubset relSubset = (RelSubset) other;
+        return relSubset.getBestOrOriginal().accept(this);
+      } else {
+        return visitChildren(other);
+      }
+    }
   }
 
   @Override protected @Nullable RelNode createUnion(RelBuilder relBuilder, RexBuilder rexBuilder,
@@ -942,18 +955,20 @@ public abstract class MaterializedViewAggregateRule<C extends MaterializedViewAg
     return Pair.of(resultTopViewProject, requireNonNull(resultViewNode, "resultViewNode"));
   }
 
+
+
   /**
-   * If the view contains a FLOOR(col) and the query contains a range predicate on the col then
+   * If the view contains FLOOR(col) and the query contains a range predicate on col then
    * rewrite the view to include a predicate based on the query predicate so that the view
    * can be used.
    * @return pair of view and added view predicate, or null if the rewrite can't be done
    */
-  @Override public @Nullable  Pair<RelNode, RexNode> rewriteInputView(RelOptRuleCall call,
+  @Override public @Nullable Pair<RelNode, RexNode> rewriteInputView(RelOptRuleCall call,
       RelNode view, RelNode viewNode, RexBuilder rexBuilder, Pair<RexNode, RexNode> queryPreds,
       RexNode viewPred) {
-    if (!queryPreds.right.isAlwaysTrue() && viewPred.isAlwaysTrue()) {
+    if (queryPreds.right != null && !queryPreds.right.isAlwaysTrue() && viewPred.isAlwaysTrue()) {
       // If there is no view predicate add one based on the query predicate if the
-      // view is a rollup ( contains an aggregation that groups by FLOOR )
+      // view is a rollup (contains an aggregation that groups by FLOOR)
       if (viewNode instanceof Aggregate) {
         Aggregate aggregate = (Aggregate) viewNode;
         RelNode input = aggregate.getInput();
@@ -964,18 +979,10 @@ public abstract class MaterializedViewAggregateRule<C extends MaterializedViewAg
             if (rexNode instanceof RexCall) {
               RexCall rexCall = (RexCall) rexNode;
               if (rexCall.getKind() == SqlKind.FLOOR) {
-                RexInputRef rexInputRef = RexInputRef.of(viewColumnIndex, view.getRowType());
-                ImplicitViewPredicateShuttle viewPredicateShuttle =
-                    new ImplicitViewPredicateShuttle(rexBuilder, rexCall, rexInputRef, false);
-                ImplicitViewPredicateShuttle viewPredicateShuttle2 =
-                    new ImplicitViewPredicateShuttle(rexBuilder, rexCall, rexInputRef, true);
-                RexNode viewPredicate = queryPreds.right.accept(viewPredicateShuttle);
-                if (viewPredicateShuttle.isRangeMatched()) {
-                  RexNode viewFilter = queryPreds.right.accept(viewPredicateShuttle2);
-                  RelBuilder builder =
-                      call.builder().transform(c -> c.withPruneInputOfAggregate(false));
-                  RelNode rewrittenView = builder.push(view).filter(viewFilter).build();
-                  return new Pair<>(rewrittenView, viewPredicate);
+                Pair<RelNode, RexNode> viewAndViewPredicate = getViewAndViewPredicate(call, view,
+                    rexBuilder, queryPreds, rexCall, viewColumnIndex);
+                if (viewAndViewPredicate != null) {
+                  return viewAndViewPredicate;
                 }
               }
             }
@@ -987,6 +994,29 @@ public abstract class MaterializedViewAggregateRule<C extends MaterializedViewAg
     }
     // return original view and viewPred
     return Pair.of(view, viewPred);
+  }
+
+  /**
+   * Uses ImplicitViewPredicateShuttle to generate a view predicate that will be added to the view.
+   * @return pair of view and added view predicate, or null if the rewrite can't be done
+   */
+  private @Nullable Pair<RelNode, RexNode> getViewAndViewPredicate(RelOptRuleCall call,
+      RelNode view, RexBuilder rexBuilder, Pair<RexNode, RexNode> queryPreds, RexCall rexCall,
+      int viewColumnIndex) {
+    ImplicitViewPredicateShuttle viewPredicateShuttleRangeMatcher =
+        new ImplicitViewPredicateShuttle(rexBuilder, rexCall, null);
+    RexNode viewPredicate = queryPreds.right.accept(viewPredicateShuttleRangeMatcher);
+    if (viewPredicateShuttleRangeMatcher.isRangeMatched()) {
+      RexInputRef viewInputRef = RexInputRef.of(viewColumnIndex, view.getRowType());
+      ImplicitViewPredicateShuttle viewPredicateShuttleFilterGenerator =
+          new ImplicitViewPredicateShuttle(rexBuilder, rexCall, viewInputRef);
+      RexNode viewFilter = queryPreds.right.accept(viewPredicateShuttleFilterGenerator);
+      RelBuilder builder =
+          call.builder().transform(c -> c.withPruneInputOfAggregate(false));
+      RelNode rewrittenView = builder.push(view).filter(viewFilter).build();
+      return new Pair<>(rewrittenView, viewPredicate);
+    }
+    return null;
   }
 
   /**
--- a/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
@@ -23,13 +23,16 @@ import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.plan.hep.HepPlanner;
 import org.apache.calcite.plan.hep.HepProgram;
 import org.apache.calcite.plan.hep.HepProgramBuilder;
+import org.apache.calcite.plan.volcano.RelSubset;
 import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.RelShuttleImpl;
 import org.apache.calcite.rel.core.Aggregate;
 import org.apache.calcite.rel.core.AggregateCall;
 import org.apache.calcite.rel.core.Filter;
 import org.apache.calcite.rel.core.JoinRelType;
 import org.apache.calcite.rel.core.Project;
 import org.apache.calcite.rel.core.TableScan;
+import org.apache.calcite.rel.logical.LogicalFilter;
 import org.apache.calcite.rel.metadata.RelMetadataQuery;
 import org.apache.calcite.rel.rules.AggregateProjectPullUpConstantsRule;
 import org.apache.calcite.rel.rules.CoreRules;
@@ -242,7 +245,7 @@ public abstract class MaterializedViewAggregateRule<C extends MaterializedViewAg
       @Nullable Project topProject,
       RelNode node,
       BiMap<RelTableRef, RelTableRef> queryToViewTableMapping,
-      EquivalenceClasses viewEC, EquivalenceClasses queryEC) {
+      EquivalenceClasses viewEC, EquivalenceClasses queryEC, boolean pushQueryFilter) {
     Aggregate aggregate = (Aggregate) node;
 
     // Our target node is the node below the root, which should have the maximum
@@ -292,10 +295,36 @@ public abstract class MaterializedViewAggregateRule<C extends MaterializedViewAg
                 otherCompensationPred)));
 
     // Generate query rewriting.
-    RelNode rewrittenPlan = relBuilder
-        .push(target)
-        .filter(simplify.simplifyUnknownAsFalse(queryCompensationPred))
-        .build();
+    RelNode rewrittenPlan = null;
+    if (pushQueryFilter) {
+      rewrittenPlan = target.accept(
+          // Push the queryCompensationPred down into any filter that is present in the plan
+          new RelShuttleImpl() {
+            @Override public RelNode visit(LogicalFilter filter) {
+              RexNode condition =
+                  RexUtil.flatten(rexBuilder,
+                      rexBuilder.makeCall(
+                          SqlStdOperatorTable.AND,
+                          filter.getCondition(),
+                          queryCompensationPred));
+              return filter.copy(filter.getTraitSet(), filter.getInput(), condition);
+            }
+
+            @Override public RelNode visit(RelNode other) {
+              if (other instanceof RelSubset) {
+                RelSubset relSubset = (RelSubset) other;
+                return relSubset.getBestOrOriginal().accept(this);
+              } else {
+                return visitChildren(other);
+              }
+            }
+          });
+    } else {
+      rewrittenPlan = relBuilder
+          .push(target)
+          .filter(simplify.simplifyUnknownAsFalse(queryCompensationPred))
+          .build();
+    }
     if (config.unionRewritingPullProgram() != null) {
       return aggregate.copy(aggregate.getTraitSet(),
           ImmutableList.of(
--- a/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
@@ -19,6 +19,7 @@ package org.apache.calcite.rel.rules.materialize;
 import org.apache.calcite.avatica.util.TimeUnitRange;
 import org.apache.calcite.linq4j.Ord;
 import org.apache.calcite.plan.RelOptRule;
+import org.apache.calcite.plan.RelOptRuleCall;
 import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.plan.hep.HepPlanner;
 import org.apache.calcite.plan.hep.HepProgram;
@@ -41,6 +42,7 @@ import org.apache.calcite.rel.rules.FilterProjectTransposeRule;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexPermuteInputsShuttle;
@@ -50,6 +52,7 @@ import org.apache.calcite.rex.RexTableInputRef.RelTableRef;
 import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlFunction;
+import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.fun.SqlMinMaxAggFunction;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.type.SqlTypeName;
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -485,6 +516,54 @@ protected void perform(RelOptRuleCall call, @Nullable Project topProject, RelNod
...
+    if (viewNode instanceof Aggregate) {
+      Aggregate aggregate = (Aggregate) viewNode;
+      RelNode input = aggregate.getInput();
+      if (input instanceof Project) {
+        Project project = (Project) input;
+        int viewColumnIndex = 0;
+        for (RexNode rexNode : project.getProjects()) {
+          if (rexNode instanceof RexCall) {
+            RexCall rexCall = (RexCall) rexNode;
+            if (rexCall.getOperator() instanceof SqlFloorFunction) {
```
sdreynolds, line 547: Should this compare based on `SqlKind` ? https://github.com/apache/calcite/blob/b9c2099ea92a575084b55a206efc5dd341c0df62/core/src/main/java/org/apache/calcite/sql/SqlKind.java

twdsilva: Update the PR with this change, thanks for the review @sdreynolds.

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
```diff
@@ -292,10 +298,36 @@
                 otherCompensationPred)));
 
     // Generate query rewriting.
-    RelNode rewrittenPlan = relBuilder
-        .push(target)
-        .filter(simplify.simplifyUnknownAsFalse(queryCompensationPred))
-        .build();
+    RelNode rewrittenPlan = null;
```
asolimando, line 301: Nit: `null` here is redundant and can be omitted

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
```diff
@@ -292,10 +298,36 @@
...
+                RelSubset relSubset = (RelSubset) other;
+                return relSubset.getBestOrOriginal().accept(this);
+              } else {
+                return visitChildren(other);
+              }
+            }
+          });
+    } else {
+      rewrittenPlan = relBuilder
+          .push(target)
```
asolimando, line 327: I think it's worth extracting the `RelShuttleImpl` to a private class to improve readability

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
```diff
@@ -910,6 +942,53 @@ protected SqlFunction getFloorSqlFunction(TimeUnitRange flag) {
     return Pair.of(resultTopViewProject, requireNonNull(resultViewNode, "resultViewNode"));
   }
 
+  /**
+   * If the view contains a FLOOR(col) and the query contains a range predicate on the col then
```
asolimando, line 946: Nit:
```suggestion
   * If the view contains FLOOR(col) and the query contains a range predicate on col then
```

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
```diff
@@ -910,6 +942,53 @@ protected SqlFunction getFloorSqlFunction(TimeUnitRange flag) {
...
+   * rewrite the view to include a predicate based on the query predicate so that the view
+   * can be used.
+   * @return pair of view and added view predicate, or null if the rewrite can't be done
+   */
+  @Override public @Nullable  Pair<RelNode, RexNode> rewriteInputView(RelOptRuleCall call,
```
asolimando, line 951: Remove extra whitespace
```suggestion
  @Override public @Nullable Pair<RelNode, RexNode> rewriteInputView(RelOptRuleCall call,
```

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
```diff
@@ -910,6 +942,53 @@ protected SqlFunction getFloorSqlFunction(TimeUnitRange flag) {
...
+      RelNode view, RelNode viewNode, RexBuilder rexBuilder, Pair<RexNode, RexNode> queryPreds,
+      RexNode viewPred) {
+    if (!queryPreds.right.isAlwaysTrue() && viewPred.isAlwaysTrue()) {
```
asolimando, line 954: `queryPreds.right` is nullable here, we need to check for `queryPreds.right != null` before invoking a method to avoid NPE

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
```diff
@@ -910,6 +942,53 @@ protected SqlFunction getFloorSqlFunction(TimeUnitRange flag) {
...
+      // If there is no view predicate add one based on the query predicate if the
+      // view is a rollup ( contains an aggregation that groups by FLOOR )
```
asolimando, line 956: Nit:
```suggestion
      // view is a rollup (contains an aggregation that groups by FLOOR)
```

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
```diff
@@ -910,6 +942,53 @@ protected SqlFunction getFloorSqlFunction(TimeUnitRange flag) {
...
+              RexCall rexCall = (RexCall) rexNode;
+              if (rexCall.getKind() == SqlKind.FLOOR) {
+                RexInputRef rexInputRef = RexInputRef.of(viewColumnIndex, view.getRowType());
+                ImplicitViewPredicateShuttle viewPredicateShuttle =
+                    new ImplicitViewPredicateShuttle(rexBuilder, rexCall, rexInputRef, false);
+                ImplicitViewPredicateShuttle viewPredicateShuttle2 =
+                    new ImplicitViewPredicateShuttle(rexBuilder, rexCall, rexInputRef, true);
+                RexNode viewPredicate = queryPreds.right.accept(viewPredicateShuttle);
+                if (viewPredicateShuttle.isRangeMatched()) {
+                  RexNode viewFilter = queryPreds.right.accept(viewPredicateShuttle2);
```
asolimando, line 974: We can move the init of the second shuttle inside the `if`, the only place where it's used.
```suggestion
                RexNode viewPredicate = queryPreds.right.accept(viewPredicateShuttle);
                if (viewPredicateShuttle.isRangeMatched()) {
                  ImplicitViewPredicateShuttle viewPredicateShuttle2 = 
                      new ImplicitViewPredicateShuttle(rexBuilder, rexCall, rexInputRef, true);
                  RexNode viewFilter = queryPreds.right.accept(viewPredicateShuttle2);
```

Additionally, let's find a more descriptive name than `viewPredicateShuttle`, something like`viewPredicateShuttleFilterGenerator`?

Possibly let's improve also `viewPredicateShuttle`, with something like `viewPredicateShuttleRangeMatcher`?

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
```diff
@@ -910,6 +942,53 @@ protected SqlFunction getFloorSqlFunction(TimeUnitRange flag) {
...
+      // view is a rollup ( contains an aggregation that groups by FLOOR )
+      if (viewNode instanceof Aggregate) {
+        Aggregate aggregate = (Aggregate) viewNode;
+        RelNode input = aggregate.getInput();
+        if (input instanceof Project) {
+          Project project = (Project) input;
+          int viewColumnIndex = 0;
+          for (RexNode rexNode : project.getProjects()) {
+            if (rexNode instanceof RexCall) {
+              RexCall rexCall = (RexCall) rexNode;
```
asolimando, line 965: I feel this is the right spot where an auxiliary method could be extracted, `rewriteInputView` is too long and does a lot of things, what do you think?

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1252,6 +1285,193 @@ protected RexNode replaceWithOriginalReferences(final RexBuilder rexBuilder,
...
+    private final RexBuilder rexBuilder;
+    private final RexCall floorCall;
+    private final RexInputRef rexInputRef;
+    private final boolean generateViewFilter;
+    private long lowerBound;
+    private long upperBound;
+
+    ImplicitViewPredicateShuttle(RexBuilder rexBuilder, RexCall floorCall,
+        RexInputRef rexInputRef,
+        boolean generateViewFilter) {
```
asolimando, line 1303: Nit: we generally don't stack up arguments in Calcite
```suggestion
        RexInputRef rexInputRef, boolean generateViewFilter) {
```

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1252,6 +1285,193 @@ protected RexNode replaceWithOriginalReferences(final RexBuilder rexBuilder,
...
+          } else {
+            upperBound = truncatedVal;
+          }
+          if (generateViewFilter) {
+            tableInputRefOperand = rexInputRef;
+          }
+          RexNode modifiedCall = rexBuilder.makeCall(call.getType(), transformedCallOperator,
+              reverseOperands ? ImmutableList.of(tableInputRefOperand, truncatedLiteral)
+                  : ImmutableList.of(truncatedLiteral, tableInputRefOperand));
+          return modifiedCall;
```
asolimando, line 1348: I guess you use the variable for easing debug, for modern debuggers can step past the return expression, I think we lose clarity and we don't gain much, I suggest to drop the variable and return immediately.

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1252,6 +1285,193 @@ protected RexNode replaceWithOriginalReferences(final RexBuilder rexBuilder,
...
+    ImplicitViewPredicateShuttle(RexBuilder rexBuilder, RexCall floorCall,
+        RexInputRef rexInputRef,
+        boolean generateViewFilter) {
+      this.floorCall = floorCall;
+      this.rexBuilder = rexBuilder;
+      this.rexInputRef = rexInputRef;
+      this.generateViewFilter = generateViewFilter;
+    }
+
+    private RexNode transformCall(RexCall call, boolean isLowerBound) {
```
asolimando, line 1310: We need to split this method using auxiliary ones, can you rework it to reduce the overall complexity?

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1252,6 +1285,193 @@ protected RexNode replaceWithOriginalReferences(final RexBuilder rexBuilder,
...
+      RexNode literalOperand = call.operands.get(0);
+      RexNode tableInputRefOperand = call.operands.get(1);
+      final int floorIndex = ((RexInputRef) floorCall.getOperands().get(0)).getIndex();
+      boolean reverseOperands = false;
+      if ((literalOperand.getKind() == SqlKind.LITERAL
+          && tableInputRefOperand.getKind() == SqlKind.TABLE_INPUT_REF)
+          || (literalOperand.getKind() == SqlKind.TABLE_INPUT_REF
+          && tableInputRefOperand.getKind() == SqlKind.LITERAL)) {
+        if (literalOperand.getKind() == SqlKind.TABLE_INPUT_REF
+            && tableInputRefOperand.getKind() == SqlKind.LITERAL) {
```
asolimando, line 1323: What about simplifying the if condition by introducing some boolean variables like `isLeftLiteral`, `isRightLiteral` etc?

```suggestion
      final boolean isLeftLiteral = literalOperand.getKind() == SqlKind.LITERAL;
      final boolean isRightLiteral = tableInputRefOperand.getKind() == SqlKind.LITERAL;
      final boolean isLeftTableInputRef = literalOperand.getKind() == SqlKind.TABLE_INPUT_REF;
      final boolean isRightTableInputRef = tableInputRefOperand.getKind() == SqlKind.TABLE_INPUT_REF;
      boolean reverseOperands = false;
      if (isLeftLiteral && isRightTableInputRef || isLeftTableInputRef && isRightLiteral) {
        if (literalOperand.getKind() == SqlKind.TABLE_INPUT_REF
            && tableInputRefOperand.getKind() == SqlKind.LITERAL) {
```

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1252,6 +1285,193 @@ protected RexNode replaceWithOriginalReferences(final RexBuilder rexBuilder,
...
+          literalOperand = call.operands.get(1);
+          tableInputRefOperand = call.operands.get(0);
+          reverseOperands = true;
+        }
+        int predicateIndex = ((RexTableInputRef) tableInputRefOperand).getIndex();
+        if (floorIndex == predicateIndex) {
```
asolimando, line 1329: Here we could probably replace the if/else with a call to an aux method

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1252,6 +1285,193 @@ protected RexNode replaceWithOriginalReferences(final RexBuilder rexBuilder,
...
+      RexNode truncatedLiteral = rexBuilder.makeCall(floorCall.getType(),
+          floorOperator, ImmutableList.of(literalOperand, floorCall.getOperands().get(1)));
+      Comparable v0 = RexInterpreter.evaluate(truncatedLiteral, Collections.emptyMap());
+      if (v0 == null) {
+        throw new AssertionError("interpreter returned null for " + v0);
+      }
+      Comparable v1 = RexInterpreter.evaluate(literalOperand, Collections.emptyMap());
+      if (v1 == null) {
+        throw new AssertionError("interpreter returned null for " + v1);
+      }
```
asolimando, line 1370: `v0` and `v1` are always null when you build the assertion message, replace the value with something that makes sense

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1252,6 +1285,193 @@ protected RexNode replaceWithOriginalReferences(final RexBuilder rexBuilder,
...
+      long modifiedVal = (long) v0;
+      final long originalVal = (long) v1;
+      // Since the view contains a FLOOR() if the query does a > or <= and the operand is already
+      // a FLOORED value we have to shift the value to the next higher or lower floored value
+      // If the query contains a >= or < we don't need to shift the modified value
```
asolimando, line 1375: Can you add an example here?

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1252,6 +1285,193 @@ protected RexNode replaceWithOriginalReferences(final RexBuilder rexBuilder,
...
+        return transformCall(call, false);
+      case EQUALS:
+        return rexBuilder.makeLiteral(true);
+      default:
+        return super.visitCall(call);
+      }
+    }
+
+    public boolean isRangeMatched() {
+      return lowerBound != 0L || upperBound != 0L && lowerBound != upperBound;
```
asolimando, line 1470: Don't you mean the following?

```suggestion
      return (lowerBound != 0L || upperBound != 0L) && lowerBound != upperBound;
```

core/src/test/java/org/apache/calcite/test/MaterializedViewRelOptRulesTest.java
```diff
@@ -338,6 +341,104 @@ protected final MaterializedViewFixture sql(String materialize,
         .ok();
   }
 
+  @Test void testAggregateMaterializationAggregateFuncsRange1() {
```
asolimando, line 344: Please find better names for tests, if you have multiple tests, they are covering something specific each, so why don't put that in the test name? This is true for all tests1...n in the PR

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1252,6 +1285,193 @@ protected RexNode replaceWithOriginalReferences(final RexBuilder rexBuilder,
...
+        if (floorIndex != predicateIndex) {
+          // ignore predicates on different columns
+          return rexBuilder.makeLiteral(true);
+        }
+        if (!(r.lowerEndpoint() instanceof TimestampString)
+            || !(r.upperEndpoint() instanceof TimestampString)) {
+          // ignore predicates where the type isn't a timestamp
+          return rexBuilder.makeLiteral(true);
+        }
+        RexNode lb = rexBuilder.makeTimestampLiteral((TimestampString) r.lowerEndpoint(), 0);
```
asolimando, line 1419: Please avoid cryptic names like `lb`, `ub`, there is already a lot of complex logic in these methods, let's try to help the code reader with more explicit names like `lowerBound` and `upperBound`. 

They are surely longer, but nowadays we have IDEs for that :)

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
asolimando, line 1310: I rename `transformCall` into `adjustComparisonBoundary`, `transformComparisonBoundary` or something closer to what this method is actually doing

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1252,6 +1285,193 @@ protected RexNode replaceWithOriginalReferences(final RexBuilder rexBuilder,
...
+
+        if (generateViewFilter) {
+          tableInputRefOperand = rexInputRef;
+        }
+
+        RelDataType type = call.getOperands().get(1).getType();
+        Sarg modSarg = Sarg.of(RexUnknownAs.UNKNOWN,
+            ImmutableRangeSet.of(Range.closedOpen(lbTimestampString, ubTimestampString)));
+        RexNode searchArgumentLiteral = rexBuilder.makeSearchArgumentLiteral(modSarg, type);
+        RexNode res = rexBuilder.makeCall(SqlStdOperatorTable.SEARCH, tableInputRefOperand,
```
asolimando, line 1441: What about dropping `res` and returning the expressions?

core/src/main/java/org/apache/calcite/rex/RexInterpreter.java
```diff
@@ -342,22 +346,15 @@ private static Comparable ceil(RexCall call, List<Comparable> values) {
...
-    switch (unit) {
-    case QUARTER:
-      return TimeUnitRange.MONTH;
+  private static long ceilSubDay(SqlKind kind, Long v, TimeUnitRange unit) {
+    switch (kind) {
+    case FLOOR:
+      return SqlFunctions.floor(v, unit.startUnit.multiplier.longValue());
     default:
-      return TimeUnitRange.DAY;
+      return SqlFunctions.ceil(v, unit.startUnit.multiplier.longValue());
```
asolimando, line 357: If the `default` case is just for `ceil`, I'd rather cover it explicitly

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -3027,6 +3028,43 @@ private void checkSarg(String message, Sarg sarg,
         is(false));
   }
 
+  private void helpTestFloorCeil(String timestampString, String expectedFloorString,
```
asolimando, line 3031: `assertFloorCeil`?

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -3027,6 +3028,43 @@ private void checkSarg(String message, Sarg sarg,
...
+      String expectedCeilString, TimeUnitRange timeUnitRange) {
+    final RexLiteral timestampLiteral = rexBuilder.makeTimestampLiteral(
+        new TimestampString(timestampString), 3);
```
asolimando, line 3034: Is precision `3` always appropriate?

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -3027,6 +3028,43 @@ private void checkSarg(String message, Sarg sarg,
...
+    final RexLiteral flagLiteral = rexBuilder.makeFlag(timeUnitRange);
+
+    RexNode flooredLiteral = rexBuilder.makeCall(SqlStdOperatorTable.FLOOR,
+        ImmutableList.of(timestampLiteral, flagLiteral));
+    long expectedFloorValue = new TimestampString(expectedFloorString).getMillisSinceEpoch();
+    assertThat(eval(flooredLiteral), is(expectedFloorValue));
+
+    RexNode ceiledLiteral = rexBuilder.makeCall(SqlStdOperatorTable.CEIL,
```
asolimando, line 3042: Please be consistent with the `final` modifier, since you have added it on some other vars, if you omit it here, the reader thinks it's going to mutate, while this is not the case

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -3027,6 +3028,43 @@ private void checkSarg(String message, Sarg sarg,
...
+    String timestampString = "2011-07-20 12:34:18.078";
+    helpTestFloorCeil(timestampString, "2011-07-20 12:34:18", "2011-07-20 12:34:19",
+        TimeUnitRange.SECOND);
+    helpTestFloorCeil(timestampString, "2011-07-20 12:34:00", "2011-07-20 12:35:00",
+        TimeUnitRange.MINUTE);
+    helpTestFloorCeil(timestampString, "2011-07-20 12:00:00", "2011-07-20 13:00:00",
+        TimeUnitRange.HOUR);
+    helpTestFloorCeil(timestampString, "2011-07-20 00:00:00", "2011-07-21 00:00:00",
+        TimeUnitRange.DAY);
+    helpTestFloorCeil("2011-07-20 12:34:59.078", "2011-07-17 00:00:00", "2011-07-24 00:00:00",
```
asolimando, line 3058: The start of week here is on Sunday, which is locale dependant, is there a way to make this locale dependant? If not, why would it be OK as is?

core/src/test/java/org/apache/calcite/test/MaterializedViewRelOptRulesTest.java
```diff
@@ -338,6 +341,104 @@ protected final MaterializedViewFixture sql(String materialize,
...
+            ChainedRelMetadataProvider.of(
+                ImmutableList.of(TestMetadataHandlers.TestRelMdRowCount.SOURCE,
+                    TestMetadataHandlers.TestRelMdSelectivity.SOURCE,
+                    DefaultRelMetadataProvider.INSTANCE))
+        )
+        .ok();
+  }
+
+  @Test void testAggregateMaterializationAggregateFuncsRange31() {
+    // test using multiple views
```
asolimando, line 401: can you improve the comment to match the explanation of the tests above which is of great help? Notably, please cover why `mv0` is picked over `mv1` and sketch what happens when multiple views are available

core/src/test/java/org/apache/calcite/test/MaterializedViewRelOptRulesTest.java
```diff
@@ -338,6 +341,104 @@ protected final MaterializedViewFixture sql(String materialize,
...
+        .withMetadataProvider(
+            ChainedRelMetadataProvider.of(
+                ImmutableList.of(TestMetadataHandlers.TestRelMdRowCount.SOURCE,
+                    TestMetadataHandlers.TestRelMdSelectivity.SOURCE,
+                    DefaultRelMetadataProvider.INSTANCE))
+        )
+        .ok();
+  }
+
+
```
asolimando, line 441: Nit: remove the extra newline here

core/src/test/java/org/apache/calcite/test/TestMetadataHandlers.java
```diff
@@ -0,0 +1,108 @@
...
+          if (!r.hasLowerBound() || !r.hasUpperBound()
+              || !(r.lowerEndpoint() instanceof TimestampString)
+              || !(r.lowerEndpoint() instanceof TimestampString)) {
+            // ignore predicates where the type isn't a timestamp
+            return RelMdUtil.guessSelectivity(predicate);
+          }
+          long lowerBound = ((TimestampString) r.lowerEndpoint()).getMillisSinceEpoch();
+          long upperBound = ((TimestampString) r.upperEndpoint()).getMillisSinceEpoch();
+          // only used for a range less than one day
+          selectivity += (upperBound - lowerBound) / (Double.valueOf(DateTimeUtils.MILLIS_PER_DAY));
```
asolimando, line 101: Unnecessary boxing, it can be simplified as:
```suggestion
          selectivity += (upperBound - lowerBound) / ((double) DateTimeUtils.MILLIS_PER_DAY);
```

core/src/test/java/org/apache/calcite/test/MaterializedViewRelOptRulesTest.java
```diff
@@ -338,6 +341,104 @@ protected final MaterializedViewFixture sql(String materialize,
         .ok();
   }
```
asolimando, line 343: Could you add one or more tests explicitly covering [your claim](https://issues.apache.org/jira/browse/CALCITE-5240?focusedCommentId=17608900&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17608900
) that the behaviour is unchanged when we have more complex cases (a predicate already exists, we have a predicate in a join, etc.)?

asolimando (changes requested):
@twdsilva thanks for the PR, I think it's a very useful feature, I still share the doubts that @julianhyde raised in a [jira comment](https://issues.apache.org/jira/browse/CALCITE-5240?focusedCommentId=17598705&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17598705) (copying a filter into the view seems hacky and does not always seem safe) and that @zabetak raised [in the ML](https://lists.apache.org/thread/6c1x985qrzjvd1tc38kmcj56hs9q43bw) (isn't there a way to modify the query to match the view, instead of changing the view and making view matching more complex?).

In the same ML discussion, you have raised a question on this point, I don't have an answer to that myself, but I feel that the we all need to agree that we are going into the right direction before being able to merge the PR.

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewAggregateRule.java
twdsilva: Added a method called `getViewAndViewPredicate`.

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
twdsilva: removed the `generateViewFilter` param as its no longer required

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
twdsilva: added a method `transformCallHelper`.

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
twdsilva: I added the new method `transformCallHelper` to handle the `if` case here.

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
twdsilva: fixed

core/src/test/java/org/apache/calcite/test/MaterializedViewRelOptRulesTest.java
twdsilva: changed test names to be more descriptive

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
twdsilva: done

core/src/main/java/org/apache/calcite/rex/RexInterpreter.java
twdsilva: done

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
twdsilva: done

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
twdsilva: added final modifier here

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
twdsilva: From the comment on CALCITE-3412 I think `FLOOR(timestamp TO WEEK)` will always round down to the previous Sunday. 
https://issues.apache.org/jira/browse/CALCITE-3412?focusedCommentId=16953504&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16953504

core/src/test/java/org/apache/calcite/test/MaterializedViewRelOptRulesTest.java
twdsilva: Added a more descriptive comment.

core/src/test/java/org/apache/calcite/test/MaterializedViewRelOptRulesTest.java
twdsilva: done

core/src/test/java/org/apache/calcite/test/TestMetadataHandlers.java
twdsilva: done

core/src/test/java/org/apache/calcite/test/MaterializedViewRelOptRulesTest.java
twdsilva: I have added `testAggregateViewWithPredicate` to test when a predicate already exists in the view and `testJoinViewWithoutPredicate`, `testJoinViewWithPredicateSameAsQueryPredicate`, `testJoinViewWithPredicateDifferentThanQueryPredicate` to test views that have a join with predicate.

twdsilva:
> @twdsilva thanks for the PR, I think it's a very useful feature, I still share the doubts that @julianhyde raised in a [jira comment](https://issues.apache.org/jira/browse/CALCITE-5240?focusedCommentId=17598705&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17598705) (copying a filter into the view seems hacky and does not always seem safe) and that @zabetak raised [in the ML](https://lists.apache.org/thread/6c1x985qrzjvd1tc38kmcj56hs9q43bw) (isn't there a way to modify the query to match the view, instead of changing the view and making view matching more complex?).
> 
> In the same ML discussion, you have raised a question on this point, I don't have an answer to that myself, but I feel that the we all need to agree that we are going into the right direction before being able to merge the PR.

Thank you @asolimando for the review. 
We have a use case where we maintain materialized preaggregated cubes that store data rolled up to the UTC day or hour. The queries in our system always have a time range predicate that doesn't neccesarily align on the UTC day or boundary. We would like to use the materialized cube table whenever possible. Here is an [example](https://github.com/twilio/calcite-kudu/blob/30c0d4e7b1d15288f7c30cdfda885c0c3a4666bf/adapter/src/test/java/com/twilio/kudu/sql/ScenarioIT.java#L278).
I could not figure how to get `MaterializedViewAggregateRule`  to handle the case when a view contains a rollup (has a group by with a `FLOOR(col)`) other than rewriting the view to include a predicate based on the query predicate. The existing code that uses the union operator works if we include this generated predicate. The other main change is to map `FLOOR(col)` to col which is valid because the predicate added to the view is aligned on the floored value of the query predicate. If there is a cleaner way to make this work I would appreciate any pointers.

PR #2878, opened 2022-08-21
[CALCITE-5241] Implement CHAR function

JiajunBernoulli:
Notes: `char(n)` is different from `chr(n)` when n large than 256. `chr(n)` be added in https://issues.apache.org/jira/browse/CALCITE-2510


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -462,6 +462,16 @@ public static ByteString right(ByteString s, int n) {
     return s.substring(len - n);
   }
 
+  /**
+   * SQL CHAR(int) function.
+   */
```
julianhyde, line 467: explain the difference between this and `chr`. It isn't int vs long.

julianhyde: also make the javadoc consistent with other javadoc

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -650,7 +650,16 @@ private SqlLibraryOperators() {
...
+  @LibraryOperator(libraries = {MYSQL, SPARK})
+  public static final SqlFunction CHAR =
+      new SqlFunction("CHAR",
+          SqlKind.OTHER_FUNCTION,
+          ReturnTypes.CHAR_FORCE_NULLABLE,
+          null,
+          OperandTypes.INTEGER,
+          SqlFunctionCategory.STRING);
+
+  @LibraryOperator(libraries = {ORACLE, POSTGRESQL})
```
julianhyde, line 662: CHR and CHAR need javadoc

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -347,6 +347,13 @@ public static SqlCall stripSeparator(SqlCall call) {
   public static final SqlReturnTypeInference CHAR =
           explicit(SqlTypeName.CHAR);
 
+  /**
+   * Type-inference strategy whereby the result type of a call is a nullable
+   * Char.
+   */
+  public static final SqlReturnTypeInference CHAR_FORCE_NULLABLE =
+      CHAR.andThen(SqlTypeTransforms.FORCE_NULLABLE);
+
```
julianhyde, line 356: not 'nullable Char', but 'nullable CHAR(1)'

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1492,9 +1492,6 @@ protected static Calendar getCalendarNotTooNear(int timeUnit) {
     f.checkScalar("{fn ASCII('ABC')}", "65", "INTEGER NOT NULL");
     f.checkNull("{fn ASCII(cast(null as varchar(1)))}");
 
-    if (false) {
-      f.checkScalar("{fn CHAR(code)}", null, "");
-    }
```
julianhyde, line 1497: Why remove this test? The goal of the change is to implement JDBC '{fn CHAR}'

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
```diff
@@ -339,6 +339,7 @@ public enum BuiltInMethod {
   UPPER(SqlFunctions.class, "upper", String.class),
   LOWER(SqlFunctions.class, "lower", String.class),
   ASCII(SqlFunctions.class, "ascii", String.class),
+  CHAR(SqlFunctions.class, "charN", int.class),
   REPEAT(SqlFunctions.class, "repeat", String.class, int.class),
```
julianhyde, line 343: should this be `CHAR(SqlFunctions.class, "charN", long.class)`?

and don't you also need `CHR(SqlFunctions.class, "chr", long.class)`?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
JiajunBernoulli: Yes, it should be bigint.

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
JiajunBernoulli: Thank you for your careful review, I changed to them.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
JiajunBernoulli: Added javadoc.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
JiajunBernoulli: Changed to it.

k-jamroz:
Interesting changes in 1.32.0:
* https://issues.apache.org/jira/browse/CALCITE-5241 - `CHAR` is now a reserved function name. I am not sure if this may impact parsing. Maybe we want to implement this function?
* https://issues.apache.org/jira/browse/CALCITE-5232 (Upgrade protobuf-java from 3.17.1 to 3.21.5) - we currently have 3.21.9, so should be fine
* https://issues.apache.org/jira/browse/CALCITE-5178 - we currently do not support scalar queries, should not impact us

PR #2883, opened 2022-08-25
[CALCITE-5245] Array element accessing in BigQuery should invoke `ORDINAL` operator

goldmedal:
Invoke `ORDINAL` operator when accessing the array elements.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
@@ -188,9 +188,9 @@ public class BigQuerySqlDialect extends SqlDialect {
       unparseTrim(writer, call, leftPrec, rightPrec);
       break;
     case ITEM:
-      SqlBasicCall firstOperand = call.operand(0);
-      if (firstOperand.getOperator().equals(ARRAY_VALUE_CONSTRUCTOR)
-          || firstOperand.getOperator().equals(ARRAY_QUERY)) {
+      if (call.operand(0) instanceof SqlBasicCall
+          && (((SqlBasicCall) call.operand(0)).getOperator().equals(ARRAY_VALUE_CONSTRUCTOR)
+          || ((SqlBasicCall) call.operand(0)).getOperator().equals(ARRAY_QUERY))) {
         call.operand(0).unparse(writer, leftPrec, rightPrec);
         final SqlWriter.Frame indexFrame = writer.startList("[", "]");
         // Because Calcite use 1-based array index, use `ORDINAL` to access the target array.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -183,6 +183,15 @@ public BigQuerySqlDialect(SqlDialect.Context context) {
     case TRIM:
       unparseTrim(writer, call, leftPrec, rightPrec);
       break;
+    case ITEM:
```
libenchao, line 186: `ITEM` is not only used for `array`, but also for `map`, `struct`. So it's better to check whether the `operand(0)` is an `array`.
Besides, does `ORDINAL` only work in the array accessing in BigQuery? Are there any other places that also need to be handled like this?

goldmedal: `ORIDANL` is an [Array subscript operator](https://cloud.google.com/bigquery/docs/reference/standard-sql/operators#array_subscript_operator) in BigQuery. I didn't see any other places needs this operator.

BTW, because `arrray subscript operator` has `ORDIANL`, `SAFE_ORDINAL`, `OFFST` and `SAFE_OFFSET`, should we offer a config for the user to switch which one is preferred?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -6466,6 +6466,32 @@ private void checkLiteral2(String expression, String expected) {
...
+  @Test void testBigQueryArrayElementAccessing() {
+    final String arrayCtorSql = ""
+        + "SELECT ARRAY[1, 2, 3][1]";
+    final String expectedArrayCtorSql = ""
+        + "SELECT ARRAY[1, 2, 3][ORDINAL(1)]";
+    sql(arrayCtorSql)
+        .withBigQuery().ok(expectedArrayCtorSql);
+
+    // TODO: Got error from
+    //  `SqlImplementor`: Need to implement org.apache.calcite.rel.core.Collect
```
goldmedal, line 6478: I tried to add test for `ARRAY_QUERY` but got this error. Should we create another PR for it?

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -183,6 +187,21 @@ public BigQuerySqlDialect(SqlDialect.Context context) {
...
+      if (call.operand(0) instanceof SqlBasicCall
```
goldmedal, line 191: I tried to handle the SQL which access the element of a table column like
```sql
WITH POST.DEPT_GBY_EMP_NAME AS {
      select * from (values
              (10, array['Jane', 'Bob']),
              (20, array['Eric']),
              (30, array['Susan', 'Alice']),
              (40, array[null])) as t(deptno, employees)
}
SELECT D.EMPLOYEES[1] FROM POST.DEPT_GBY_EMP_NAME D
```
I found it's hard to recognize which type the column is here. It could be a `ARRAY`, `MAP` or `STRUCT`.  Is there any better way to do this here?

libenchao: Sorry that I didn't realized that in `SqlDialect` we do not have validated types. It's indeed not easy to know the operand's type here.
I would suggest that we go back to the JIRA to discuss the design first, I'll post my concerns there.

PR #2885, opened 2022-08-26
[CALCITE-5249] RelToSqlConverter can't access the field of an struct array element

goldmedal:


core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -658,16 +659,18 @@ public SqlNode toSql(@Nullable RexProgram program, RexNode rex) {
           final SqlNode expr = toSql(program, referencedExpr);
           sqlIdentifier = new SqlIdentifier(expr.toString(), POS);
           break;
+        case ITEM:
```
libenchao, line 662: The switch branch for `ROW` is not correct too. Maybe we should unify them and make this the default behavior?

PR #2886, opened 2022-08-27
[CALCITE-5243] SELECT NULL AS C causes NoSuchMethodException: java.sql.ResultSet.getVoid(int)

l4wei:
Support NULL case in JdbcToEnumerableConverter.generateGet


libenchao (approved):
LGTM, will merge if no other objections in 24 hours.

PR #2887, opened 2022-08-27
[CALCITE-5247]  FilterJoinRule cannot simplify left join to inner join for `WHERE RHS.C1 IS NOT NULL OR RHS.C2 IS NOT NULL`

alimans3:


core/src/main/java/org/apache/calcite/plan/StrongPredicateAnalyzer.java
```diff
@@ -0,0 +1,68 @@
...
+
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexNode;
+
+import java.util.List;
+
+/**
+ * Utilities for checking if a predicate is definitely false based on a strong policy.
+ */
+public final class StrongPredicateAnalyzer {
```
libenchao, line 27: Do you really need to add a new util class? How about just adding these methods to `Strong` since we already have many similar methods.
Besides, adding methods to `Strong` makes Calcite more extensible, e.g., some downstream projects could extends `Strong` and override some methods to add their own logics.

libenchao (commented):
@alimans3 Thanks for your contribution. The PR generally looks good to me, I only left one minor comment.

core/src/main/java/org/apache/calcite/plan/StrongPredicateAnalyzer.java
alimans3: Thanks for the remark, done!

core/src/main/java/org/apache/calcite/plan/Strong.java
```diff
@@ -149,14 +149,45 @@ public static boolean allStrong(List<RexNode> operands) {
...
 
-  /** Returns whether an expression is definitely not true. */
+  /** Returns whether an expression is definitely not true by either it being analyzed to be not
+   * true or null. */
   public boolean isNotTrue(RexNode node) {
+    return isPredicateNotTrue(node) || isNull(node);
+  }
+
+  /** Returns whether an expression is definitely not true. */
+  public boolean isPredicateNotTrue(RexNode node) {
```
libenchao, line 159: Why do you add this public method? Can it be merged into `isNotTrue`?

alimans3: We need this method since the methods below (allNotTrue, anyNotTrue) cannot rely on the isNotTrue due to the || isNull call.
I will make this method private

core/src/main/java/org/apache/calcite/plan/Strong.java
```diff
@@ -149,14 +149,45 @@ public static boolean allStrong(List<RexNode> operands) {
     return operands.stream().allMatch(Strong::isStrong);
   }
 
-  /** Returns whether an expression is definitely not true. */
+  /** Returns whether an expression is definitely not true by either it being analyzed to be not
```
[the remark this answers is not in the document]
alimans3: done

core/src/main/java/org/apache/calcite/plan/Strong.java
alimans3: made private.

core/src/main/java/org/apache/calcite/plan/Strong.java
```diff
@@ -151,12 +151,41 @@ public static boolean allStrong(List<RexNode> operands) {
 
   /** Returns whether an expression is definitely not true. */
   public boolean isNotTrue(RexNode node) {
+    return isPredicateNotTrue(node) || isNull(node);
```
libenchao, line 154: I'm still not sure why we cannot merge `isNotTrue` and `isPredicateNotTrue` like below:

```
  public boolean isNotTrue(RexNode node) {
    switch (node.getKind()) {
    //TODO Enrich with more possible cases?
    case IS_NOT_NULL:
      return isNull(((RexCall) node).getOperands().get(0));
    case OR:
      return allNotTrue(((RexCall) node).getOperands());
    case AND:
      return anyNotTrue(((RexCall) node).getOperands());
    default:
      return isNull(node);
    }
  }

  /** Returns whether all expressions in a list are definitely not true. */
  private boolean allNotTrue(List<RexNode> operands) {
    for (RexNode operand : operands) {
      if (!isNotTrue(operand)) {
        return false;
      }
    }
    return true;
  }

  /** Returns whether any expressions in a list are definitely not true. */
  private boolean anyNotTrue(List<RexNode> operands) {
    for (RexNode operand : operands) {
      if (isNotTrue(operand)) {
        return true;
      }
    }
    return false;
  }
```

alimans3: it is doable, it covers more cases this way hence the new change in RelOptRules.xml
The case is AND/OR(DEFINITELY FALSE, DEFINITELY NULL) -> DEFINITELY FALSE
done!

libenchao (commented):
@alimans3 Thanks for your patience, the PR lgtm now. However, the CI still fails, could you fix them?

alimans3:
> @alimans3 Thanks for your patience, the PR lgtm now. However, the CI still fails, could you fix them?

should be fixed now!

libenchao (approved):
LGTM, will merge~ 
Thank you for the contribution @alimans3

suibianwanwank:
This issue should have been resolved in [CALCITE-5247], maybe we can close this PR.

PR #2888, opened 2022-08-29
[CALCITE-5251] Support SQL hint for Snapshot

lincoln-lil:
Current, `snapshot` does not support hint yet and this blocks hint propagation to the underlying temporal table.
This pr aims to support Snapshot node.


danny0405 (approved):
+1

PR #2890, opened 2022-09-01
[CALCITE-5259] Add getParameterRowType method to Planner interface

dssysolyatin:


core/src/main/java/org/apache/calcite/prepare/PlannerImpl.java
```diff
@@ -237,6 +237,15 @@ private void ready() {
     return Pair.of(validatedNode, type);
   }
 
+  @Override public RelDataType getParameterRowType() {
+    if (state.ordinal() < State.STATE_4_VALIDATED.ordinal()) {
+      throw new RuntimeException("Need to call #validate() first");
+    }
```
jbalint, line 243:
```suggestion
ensure(State.STATE_4_VALIDATED);```

jbalint (changes requested):

core/src/main/java/org/apache/calcite/prepare/PlannerImpl.java
dssysolyatin: @jbalint 
`ensure` works differently. `ensure` checks if planner can move to next state. if ensure is used and user calls getParameterRowType from STATE_5_CONVERTED state, the `getParameterRowType` will throw exception

jbalint: yes - you're right. the javadoc is wrong... :(

jbalint (approved):

PR #2891, opened 2022-09-02
[CALCITE-5260] Pop the last RelNode added into the bindings list when failing to apply a rule

wangyeweikuer:
When fail to match a subtree, HepPlanner should pop the last RelNode added into the bindings list, where the nodes left are 'real' matched when applying a rule. 

Otherwise, using some temporary and unnecessary nodes to construct a HepRuleCall will cause an assertion error.

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/hep/HepPlanner.java
+++ b/core/src/main/java/org/apache/calcite/plan/hep/HepPlanner.java
@@ -67,6 +67,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
+import java.util.Stack;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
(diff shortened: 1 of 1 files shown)


libenchao:
@wangyeweikuer Thanks for your contribution, could you add some tests which will fail without your fix? This will make us easier to understand the PR.

core/src/main/java/org/apache/calcite/plan/hep/HepPlanner.java
```diff
@@ -614,7 +615,7 @@ private List<HepRelVertex> getVertexParents(HepRelVertex vertex) {
   private static boolean matchOperands(
       RelOptRuleOperand operand,
       RelNode rel,
-      List<RelNode> bindings,
+      Stack<RelNode> bindings,
       Map<RelNode, List<RelNode>> nodeChildren) {
     if (!operand.matches(rel)) {
       return false;
```
chunweilei, line 621: Could you add a test case?

PR #2892, opened 2022-09-03
[CALCITE-5263] Improve XmlFunctions by using an XML DocumentBuilder

rubenada:
[CALCITE-5263](https://issues.apache.org/jira/browse/CALCITE-5263) Improve XmlFunctions by using an XML DocumentBuilder

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/XmlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/XmlFunctions.java
@@ -24,7 +24,9 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 import org.xml.sax.InputSource;
+import org.xml.sax.SAXException;
 
+import java.io.IOException;
 import java.io.StringReader;
 import java.io.StringWriter;
 import java.util.ArrayList;
@@ -33,6 +35,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import javax.xml.XMLConstants;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.transform.ErrorListener;
 import javax.xml.transform.OutputKeys;
 import javax.xml.transform.Source;
@@ -48,6 +54,7 @@ import javax.xml.xpath.XPathConstants;
 import javax.xml.xpath.XPathExpression;
 import javax.xml.xpath.XPathExpressionException;
 import javax.xml.xpath.XPathFactory;
+import javax.xml.xpath.XPathFactoryConfigurationException;
 
 import static org.apache.calcite.linq4j.Nullness.castNonNull;
 import static org.apache.calcite.util.Static.RESOURCE;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/runtime/XmlFunctions.java
```diff
@@ -215,6 +238,27 @@ private static String convertNodeToString(Node node) throws TransformerException
     return writer.toString();
   }
 
+  private static Node getDocumentNode(final String xml) {
+    final DocumentBuilderFactory documentBuilderFactory = DocumentBuilderFactory.newInstance();
```
zabetak, line 242: Do we need to create a factory on every call to this method? Since the configuration does not change it may be safe to reuse the instance (per application or per thread).

zabetak (approved):
LGTM. I left some minor comments.

core/src/test/java/org/apache/calcite/test/SqlXmlFunctionsTest.java
```diff
@@ -36,6 +36,18 @@
  */
 class SqlXmlFunctionsTest {
 
+  private static final String XML = "<document>string</document>";
+  private static final String FILE = System.getProperty("os.name").startsWith("Windows")
+      ? "file:///C:/Windows/System32/drivers/etc/hosts"
+      : "file:///etc/hosts";
```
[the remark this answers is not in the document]
rubenada: Switched to temp file

core/src/main/java/org/apache/calcite/runtime/XmlFunctions.java
```diff
@@ -215,6 +238,27 @@ private static String convertNodeToString(Node node) throws TransformerException
...
+    documentBuilderFactory.setXIncludeAware(false);
+    documentBuilderFactory.setExpandEntityReferences(false);
+    documentBuilderFactory.setNamespaceAware(true);
+    try {
+      documentBuilderFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+      documentBuilderFactory
+          .setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+    } catch (final ParserConfigurationException e) {
+      throw new IllegalArgumentException("Document Builder configuration failed", e);
```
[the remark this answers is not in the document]
rubenada: Changed into IllegalStateException

core/src/main/java/org/apache/calcite/runtime/XmlFunctions.java
rubenada: Seems reasonable: created ThreadLocal DocumentBuilderFactory

core/src/main/java/org/apache/calcite/runtime/XmlFunctions.java
```diff
@@ -60,11 +67,24 @@
 public class XmlFunctions {
 
   private static final ThreadLocal<@Nullable XPathFactory> XPATH_FACTORY =
-      ThreadLocal.withInitial(XPathFactory::newInstance);
+      ThreadLocal.withInitial(() -> {
+        final XPathFactory xPathFactory = XPathFactory.newInstance();
+        try {
+          xPathFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        } catch (XPathFactoryConfigurationException e) {
+          throw new IllegalStateException("XPath Factory configuration failed", e);
```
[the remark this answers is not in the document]
rubenada: So, if I understand it correctly, we are breaking applications that cannot create a secure XPathFactory and which require such factory because they are calling an xml operator. If an application cannot create a secure XPathFactory but does not make use of xml operators, it will be fine.
If this is the case, I think we should leave the code as it is: if an xml operation is invoked, it needs to be done via a secure factory, that seems like the safest approach.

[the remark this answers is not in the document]
zabetak: Sounds good let's leave it like that and re-evaluate in the future if necessary.

exceptionfactory (approved):

core/src/test/java/org/apache/calcite/test/SqlXmlFunctionsTest.java
```diff
@@ -36,6 +39,23 @@
...
 
+  private static final String XML = "<document>string</document>";
+  private static final String XSLT =
+      "<xsl:stylesheet xmlns:xsl=\"http://www.w3.org/1999/XSL/Transform\"></xsl:stylesheet>";
+  private static final String DOCUMENT_PATH = "/document";
+  private static @Nullable String xmlExternalEntity = null;
+  private static @Nullable String xsltExternalEntity = null;
+
+  @BeforeAll public static void setup() throws Exception {
+    final File testFile = File.createTempFile("foo", "temp");
```
exceptionfactory, line 50: Although this is a temporary test file, the documentation for [java.io.File.createTempFile()](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#createTempFile(java.lang.String,%20java.lang.String)) notes that the [java.nio.Files.createTempFile()](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createTempFile-java.nio.file.Path-java.lang.String-java.lang.String-java.nio.file.attribute.FileAttribute...-) method creates the file with more restrictive permissions.

core/src/main/java/org/apache/calcite/runtime/XmlFunctions.java
```diff
@@ -60,13 +67,41 @@
...
+        final TransformerFactory transformerFactory = TransformerFactory.newInstance();
         transformerFactory.setErrorListener(new InternalErrorListener());
+        try {
+          transformerFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        } catch (TransformerConfigurationException e) {
+          throw new IllegalStateException("Transformer Factory configuration failed", e);
+        }
         return transformerFactory;
       });
+  private static final ThreadLocal<@Nullable DocumentBuilderFactory> DOCUMENT_BUILDER_FACTORY =
```
pjfanning, line 90: Would it be possible to apply similar secure features to the DocumentBuilderFactory used in https://github.com/apache/calcite/blob/b9c2099ea92a575084b55a206efc5dd341c0df62/testkit/src/main/java/org/apache/calcite/test/DiffRepository.java#L210 ?

rubenada: Thanks for the comment @pjfanning . The scope of this PR is just XmlFunctions, but I will create a separate ticket + PR to carry out your suggestion.

core/src/test/java/org/apache/calcite/test/SqlXmlFunctionsTest.java
rubenada: Modification applied.
Let me know if everything is ok on your side @exceptionfactory . If so, I will proceed to finalize the PR (squash commits into a single one + merge).

exceptionfactory (approved):
Thanks for making the adjustment @rubenada, looks good!

core/src/main/java/org/apache/calcite/runtime/XmlFunctions.java
pjfanning: I raised https://issues.apache.org/jira/browse/CALCITE-5274 just in case - thanks

rubenada: Thanks @pjfanning , I will take care of it.

PR #2893, opened 2022-09-03
[CALCITE-5262] Add WKB, EWKB, GeoJSON, and GML support

bchapuis:


bchapuis:
In addition to the support for WKB, EWKB, GeoJSON, and GML, this PR adds support for a lot of the simplest ST_ functions. Most of the test cases have been adapted from examples found in the H2GIS and Postgis documentation.

PR #2895, opened 2022-09-06
[CALCITE-5267] Remove useless variable 'newCasts' in AggregateCaseToF…

zhuwenzhuang:
Remove useless variable 'newCasts' in AggregateCaseToFilter.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/AggregateCaseToFilterRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/AggregateCaseToFilterRule.java
@@ -105,30 +105,15 @@ public class AggregateCaseToFilterRule
     final List<AggregateCall> newCalls =
         new ArrayList<>(aggregate.getAggCallList().size());
     final List<RexNode> newProjects = new ArrayList<>(project.getProjects());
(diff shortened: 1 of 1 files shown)


chunweilei (commented):
I checked the history. The rule is introduced from Druid[1], in which newCasts is used. It may be better to correct it instead of removing it.

[1] https://github.com/apache/druid/pull/4360/files#diff-809e3a73d1485dad83629598ea5bcbab8f5ce7724e9572975b052176c44fbdf7

zhuwenzhuang:
> I checked the history. The rule is introduced from Druid[1], in which newCasts is used. It may be better to correct it instead of removing it.
> 
> [1] https://github.com/apache/druid/pull/4360/files#diff-809e3a73d1485dad83629598ea5bcbab8f5ce7724e9572975b052176c44fbdf7

I guess the code "relBuilder.aggregate(groupKey, newCalls).convert(aggregate.getRowType(), false)" is a better way to do CAST. So we don't need to save every project to keep types stable.

libenchao (approved):
LGTM, Thanks for your contribution @zhuwenzhuang, I'll merge this if there is no objections from @chunweilei

PR #2896, opened 2022-09-07
[CALCITE-5270] Make Firebolt dialect return false for supportsAggregateFunctionFilter

tjbanghart:
Firebolt does not support filter clauses in aggregate functions. e.g. COUNT(*) FILTER (WHERE a = 2).
The [dialect implementation](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/dialect/FireboltSqlDialect.java) should override the `supportsAggregateFunctionFilter` method it currently inherits from the  base [SqlDialect class](https://github.com/apache/calcite/blob/b9c2099ea92a575084b55a206efc5dd341c0df62/core/src/main/java/org/apache/calcite/sql/SqlDialect.java#L758)

Firebolt SQL reference: https://docs.firebolt.io/sql-reference/functions-reference/

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/FireboltSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/FireboltSqlDialect.java
@@ -158,6 +158,10 @@ public class FireboltSqlDialect extends SqlDialect {
         SqlParserPos.ZERO);
   }
 
+  @Override public boolean supportsAggregateFunctionFilter() {
+    return false;
+  }
+
   @Override public boolean supportsFunction(SqlOperator operator,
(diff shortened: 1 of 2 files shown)


JiajunBernoulli:
@tjbanghart , would you please add some unit tests?

zabetak:
@tjbanghart The best place to add tests about the change is `RelToSqlConverterTest`. Please verify that generated queries actually run in `Firebolt`.

tjbanghart:
Thanks for taking a look. I added a simple unit test in `RelToSqlConverterTest`. I think that should be sufficient but please let me know if it needs more.

PR #2899, opened 2022-09-08
[CALCITE-5277] Make EnumerableRelImplementor stashedParameters order deterministic to increase BINDABLE_CACHE hit rate

rubenada:
More info: [CALCITE-5277](https://issues.apache.org/jira/browse/CALCITE-5277)


thomasrebele:
Thanks for the PR! What's the advantage of using a `LinkedHashMap` and `Equivalence.Identity` compared to a simple `ArrayList` containing the `ParameterExpression`s in the order they were added? The latter might consume less memory.

thomasrebele:
Seems that we have different definitions of simple code ;) I have the impression that an `IdentityHashMap` and an `ArrayList` is a bit easier to understand than a `LinkedHashMap` with a `Equivalence.Wrapper<Object>`. A comment at `stashedParameters` could indicate to update the list as well. A check whether the map and the list have the same number of elements could be added.

However, this is only a minor detail. If you insist on using the Equivalence class, the PR is fine for me.

PR #2901, opened 2022-09-08
[CALCITE-5264] Apply hint exclusion check for all rels in the RelOptCall

Aitozi:
This PR is meant to apply hint exclusion check for all rels in the RelOptCall. This will help to cover the case that the hint exclusion is attatched not on the rel root.


Aitozi:
Hi @danny0405  @julianhyde please help review when you are free, thanks

core/src/main/java/org/apache/calcite/plan/RelOptRuleCall.java
```diff
@@ -201,13 +201,18 @@ public RelOptPlanner getPlanner() {
    * @return true iff rule should be excluded
    */
   public boolean isRuleExcluded() {
-    if (!(rels[0] instanceof Hintable)) {
-      return false;
+    boolean excluded = false;
```
asolimando, line 204: You don't need the `excluded` variable at all it seems, remove this line and replace the latest return statement with `return false;` (sorry but multi-line comment is broken in Github UI in this case so I can't use code suggestion).

core/src/test/java/org/apache/calcite/test/SqlHintsConverterTest.java
```diff
@@ -578,6 +579,21 @@ public final Fixture sql(String sql) {
...
+            + " where deptno = 20)";
+
+    final RelNode rel = ruleFixture().sql(sql).toRel();
+    HepProgram program = new HepProgramBuilder()
+            .addRuleInstance(CoreRules.FILTER_PROJECT_TRANSPOSE)
+            .build();
+    HepPlanner planner = new HepPlanner(program);
+    planner.setRoot(rel);
+    RelNode newRel = planner.findBestExp();
+    Assertions.assertTrue(rel.deepEquals(newRel));
```
asolimando, line 594: It's convenient to show expected vs computed when test fails, otherwise you need to add prints or open the debugger to see what's going on.

Consider something along this line:
```suggestion
    Assertions.assertTrue(rel.deepEquals(newRel),
        "Expected:\n" + RelOptUtil.toString(rel) + "Computed:\n" + RelOptUtil.toString(newRel));
```

Alternatively, you might compare the strings directly if possible, the diff would be even easier to inspect.

asolimando:
Apart from few minor comments, the PR looks good to me. I have checked the original PR and the discussion there, it makes sense to me too to check all rels, but since I am not very familiar with hints I'd prefer a second opinion around this before merging.

core/src/main/java/org/apache/calcite/plan/RelOptRuleCall.java
Aitozi: Good catch! Removed

Aitozi:
@asolimando Thanks for your comments, I have addressed them. And rebase on the latest master force pushed.

asolimando:
> @asolimando Thanks for your comments, I have addressed them. And rebase on the latest master force pushed.

Thanks @Aitozi for addressing the comments so fast!

However, rebasing on the target branch is not necessary and we avoid it when the review is still in progress as it makes reviewing the diff over the previous versions way harder.

We only rebase/squash at the very end when the active reviewers are done reviewing (they generally ask for it explicitly).

In this case it's not a big deal since the changes are rather small, but please bear it in mind for the future :)

Aitozi:
Hi @asolimando can you help me here:

when I develop on the calcite, I found it's not easy to conform its checkstyle, especially the indentation. I have tried 
- `./gradlew autostyleApply` to auto fix some problem, but the indentation still there
- Import the checktyle config as described [here](https://calcite.apache.org/docs/howto.html#setting-up-intellij-idea), but it still can not show the checkstyle problem, when writing the code.

The last way I tried is to run the `./gradlew autostyleCheck checkstyleAll` and fix the problem one by one. But it's a bit inefficient. Do you have some best practice for it ? Thanks.

asolimando:
> Hi @asolimando can you help me here:
> 
> when I develop on the calcite, I found it's not easy to conform its checkstyle, especially the indentation. I have tried
> 
> * `./gradlew autostyleApply` to auto fix some problem, but the indentation still there
> * Import the checktyle config as described [here](https://calcite.apache.org/docs/howto.html#setting-up-intellij-idea), but it still can not show the checkstyle problem, when writing the code.
> 
> The last way I tried is to run the `./gradlew autostyleCheck checkstyleAll` and fix the problem one by one. But it's a bit inefficient. Do you have some best practice for it ? Thanks.

I use `./gradlew autostyleApply` or `./gradlew check` (fix + check), but that's not addressing all possible issues as you have experienced, I guess this is the best we got at the moment. 

Others might know better, in general it's preferable to send an email to the mailing list for such general questions to increase your chances of getting an answer.

Aitozi:
Thanks for your guidance :)

asolimando:
The PR LGTM, I will merge by the end of this week if there will be no objections

asolimando:
@Aitozi, can you squash the commits into a single one having as commit message `[CALCITE-5264] HintStrategy rule exclusion does not match innermost rels`? As soon as this is done I am ready to merge the PR, thanks!

Aitozi:
> @Aitozi, can you squash the commits into a single one having as commit message `[CALCITE-5264] HintStrategy rule exclusion does not match innermost rels`? As soon as this is done I am ready to merge the PR, thanks!

Sure, done!

PR #2906, opened 2022-09-14
[CALCITE-5285] Preventing CalcReduceExpressionsRule to transform to self

krzysztofslusarski:
Fixing https://issues.apache.org/jira/browse/CALCITE-5285


viliam-durina:
Shouldn't we also add a check that will fail in case `call.transformTo` is supposed to transform to a deep-equal relnode, if this is not allowed in general?

PR #2910, opened 2022-09-17
[CALCITE-5265] Select operator' parentheses should be same with Union operator

l4wei:
remove necessary parentheses on select clause.

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -6354,7 +6354,7 @@ class RelToSqlConverterTest {
 
   /** Test case for
    * <a href="https://issues.apache.org/jira/browse/CALCITE-5265">[CALCITE-5265]
-   * Select operator' parentheses should be same with Union operator</a>. */
+   * Unparsing should not add parentheses for SELECT in INSERT</a>. */
   @Test void testInsertValueWithDynamicParams() {
     final String sql = "insert into \"DEPT\" values (?,?,?)";
     final String expected = ""
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -6354,13 +6354,37 @@ class RelToSqlConverterTest {
 
   /** Test case for
    * <a href="https://issues.apache.org/jira/browse/CALCITE-5265">[CALCITE-5265]
-   * Unparsing should not add parentheses for SELECT in INSERT</a>. */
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -6352,6 +6352,20 @@ private void checkLiteral2(String expression, String expected) {
         .withCalcite().ok(expectedCalciteX);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5265">[CALCITE-5265]
+   * Select operator' parentheses should be same with Union operator</a>. */
+  @Test void testInsertValueWithDynamicParams() {
```
libenchao, line 6358: Are there any tests which cover what Julian has mentioned in the JIRA comment?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -6352,6 +6352,20 @@ private void checkLiteral2(String expression, String expected) {
         .withCalcite().ok(expectedCalciteX);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5265">[CALCITE-5265]
+   * Select operator' parentheses should be same with Union operator</a>. */
```
libenchao, line 6357: The comment should also be adapted according to the changes in JIRA title.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
l4wei: There are many unit tests cover "insert into aTable select ? from bTable" in SqlParserTest.java. And RelToSqlConverterTest::testInsertValuesWithDynamicParams will cover "insert into aTable select ? from bTable union all select ? from cTable" mentioned by Julian in JIRA comment.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
l4wei: OK, done.

libenchao (approved):
@l4wei Thanks the updating, the PR LGTM now, merging.

libenchao:
@l4wei I didn't notice Julian has another comment when I commented with "LGTM", could you add another test which addresses Julian's comment?

l4wei:
> 

OK

l4wei:
> 

I have added test that mentioned by Julian' last comment in third commit in this PR. FYI

libenchao (approved):
LGTM, will merge in 24 hours if there is no more objections appear.

PR #2911, opened 2022-09-18
[CALCITE-5127] Error when executing query with subquery in select lis…

libenchao:
…t that uses outer column of array type

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
@@ -481,15 +481,8 @@ public class RelFieldTrimmer implements ReflectiveVisitor {
     final int fieldCount = rowType.getFieldCount();
     final RelNode input = project.getInput();
 
-    boolean containsSubQuery = false;
-    for (RexNode node : project.getProjects()) {
-      if (RexUtil.containsSubQuery(node)) {
-        containsSubQuery = true;
-        break;
-      }
-    }
     // Do not trim Project's fields before SubQueryRemoveRule applies.
-    if (containsSubQuery) {
+    if (RexUtil.containsSubQuery(project.getProjects())) {
       return result(project, Mappings.createIdentity(fieldCount));
     }
 
--- a/core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
@@ -481,6 +481,18 @@ public class RelFieldTrimmer implements ReflectiveVisitor {
     final int fieldCount = rowType.getFieldCount();
     final RelNode input = project.getInput();
 
+    boolean containsSubQuery = false;
+    for (RexNode node : project.getProjects()) {
+      if (RexUtil.containsSubQuery(node)) {
+        containsSubQuery = true;
+        break;
+      }
+    }
+    // Do not trim Project's fields before SubQueryRemoveRule applies.
+    if (containsSubQuery) {
+      return result(project, Mappings.createIdentity(fieldCount));
+    }
+
     // Which fields are required from the input?
     final Set<RelDataTypeField> inputExtraFields =
         new LinkedHashSet<>(extraFields);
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -1908,9 +1908,19 @@ private RelBuilder project_(
...
 
+    // Do not merge projections when top project contains RexSubQuery
+    boolean containsSubQuery = false;
+    for (RexNode node : nodes) {
+      if (RexUtil.containsSubQuery(node)) {
+        containsSubQuery = true;
+        break;
+      }
+    }
+
```
zabetak, line 1919: I would move the check inside the if block (maybe after line 1934/1944) so that we don't spend time looking for subqueries if we don't need to.

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -481,6 +481,18 @@ public TrimResult trimFields(
...
+      if (RexUtil.containsSubQuery(node)) {
+        containsSubQuery = true;
+        break;
+      }
+    }
+    // Do not trim Project's fields before SubQueryRemoveRule applies.
+    if (containsSubQuery) {
+      return result(project, Mappings.createIdentity(fieldCount));
+    }
+
```
zabetak, line 495: Does this code has test coverage? I don't see any plan changes due to this. Also the effects of this change are much broader than the changes in `RelBuilder` thus I would suggest to treat this separately in another JIRA.

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -3383,6 +3383,16 @@ void checkCorrelatedMapSubQuery(boolean expand) {
     sql(sql).withDecorrelate(true).ok();
   }
 
+  /**
+   * Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-5127">[CALCITE-5127]
+   * Error when executing query with subquery in select list that uses outer column of array
+   * type</a>.
+   */
+  @Test void testCorrelationWithProjection() {
```
zabetak, line 3391: Possibly more descriptive names for the test case:
* `testNotMergeProjectionsWhenCorrelationPresent`
* `testProjectionWithCorrelationNoMerge`

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
zabetak: Also this is a breaking change since it changes the old behavior. It fixes a bug but at the same time prevents some optimizations to take effect. I think it is worth adding a few lines explaining the change in behavior in the respective section in `history.md`.

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
libenchao: The test added in `sub-query.iq` will fail without this change. It surprised me too that this does not affect any existing plan tests. I guess that's because we don't have enough plan tests for sub query. If there is a test case which is affected by this change, then it was already wrong.

Due to the integrity of this issue I think this is ok to keep it in the same PR, WDYT?

zabetak: I didn't notice that the new test in `sub-query.iq` was passing from here, great! Since there are tests then I guess it is OK to keep the changes as part of this PR.

> If there is a test case which is affected by this change, then it was already wrong.
It may be wrong but not always, right? This is a rough check as we discussed not precise.

Moreover if there is a sub-query in the top project then field trimmer will stop without doing anything. I think we can do better than that. Is it valid to trim its input? I mean something like the following:

```java
if (containsSubQuery) {
  return trimFields((RelNode) project, fieldsUsed, extraFields);
}
```

zabetak: Maybe it is worth adding the new test also in `RelFieldTrimmerTest` since it is explicitly targeting this code path.

libenchao: tests added.

libenchao:
closed due to we changed to https://github.com/apache/calcite/pull/2813 again, see more discussions in the JIRA

PR #2913, opened 2022-09-19
[CALCITE-5180] Implement BigQuery TIME_TRUNC and TIMESTAMP_TRUNC functions

tjbanghart:


core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -91,7 +91,10 @@ public class CalciteCatalogReader implements Prepare.CatalogReader {
 
   public CalciteCatalogReader(CalciteSchema rootSchema,
       List<String> defaultSchema, RelDataTypeFactory typeFactory, CalciteConnectionConfig config) {
-    this(rootSchema, SqlNameMatchers.withCaseSensitive(config != null && config.caseSensitive()),
+    this(rootSchema,
+        // TODO: BigQuery has case-insensitive functions
```
tjbanghart, line 95: https://github.com/apache/calcite/pull/2914 should fix this.

babel/src/test/resources/sql/big-query.iq
```diff
@@ -16,7 +16,809 @@
 # limitations under the License.
 #
 !use scott-big-query
-!set outputformat csv
+!set outputformat mysql
```
julianhyde, line 19: when I merge this I'll factor out the big-query.iq change as a separate commit

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -520,7 +520,9 @@ public class SqlParserTest {
       "TEMPORARY",                     "92", "99",
       "THEN",                          "92", "99", "2003", "2011", "2014", "c",
       "TIME",                          "92", "99", "2003", "2011", "2014", "c",
+      "TIME_TRUNC", // BigQuery
```
julianhyde, line 523: I'm surprised there are new keywords. We needed a new keyword for `ILIKE` because it's infix, but these can just be function names.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1604,6 +1604,10 @@ protected static Calendar getCalendarNotTooNear(int timeUnit) {
     f.checkScalar("{fn TIMESTAMPDIFF(MONTH,"
         + " TIMESTAMP '2019-09-01 00:00:00',"
         + " TIMESTAMP '2020-03-01 00:00:00')}", "6", "INTEGER NOT NULL");
+    f.checkScalar("{fn TIMESTAMP_TRUNC(TIMESTAMP '2019-09-20 15:30:00', MONTH) }",
```
julianhyde, line 1607: I don't think JDBC syntax is a requirement. You should remove these lines, and revert the changes to `SqlJdbcFunctionCall.java`.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -4083,6 +4083,55 @@ void subTestIntervalSecondNegative() {
...
+              + "TIME_TRUNC\\(<TIME>,.*\\)");
+      expr(String.format(Locale.ROOT, test, unit))
+          .withOperatorTable(opTable)
+          .columnType("TIME(0) NOT NULL");
+    }
+    // should fail on incompatible time unit
+    expr("TIME_TRUNC(TIME '15:30:00.00', ^DAY^)")
+        .fails("(?s).*Was expecting one of.*");
+  }
+
```
julianhyde, line 4134: Are these tests superfluous? between `SqlParserTest` and `SqlOperatorTest` I think you have it covered. I don't think there's any non-trivial validation behavior.

site/_docs/reference.md
```diff
@@ -2625,6 +2625,8 @@ semantics.
 | b m o p | SUBSTR(string, position [, substringLength ]) | Returns a portion of *string*, beginning at character *position*, *substringLength* characters long. SUBSTR calculates lengths using characters as defined by the input character set
 | m | STRCMP(string, string)                         | Returns 0 if both of the strings are same and returns -1 when the first argument is smaller than the second and 1 when the second one is smaller than the first one
 | o | TANH(numeric)                                  | Returns the hyperbolic tangent of *numeric*
+| b | TIME_TRUNC(time, timeUnit)                     | Truncates a *time* value to the granularity of *timeUnit*. The *time* value is always rounded to the beginning of timeUnit, which can be one of the following: MILLISECOND, SECOND, MINUTE, HOUR.
+| b | TIMESTAMP_TRUNC(timestamp, timeUnit)           | Truncates a *timestamp* value to the granularity of *timeUnit*. The *timestamp* value is always rounded to the beginning of the *timeUnit*.
```
julianhyde, line 2629: move down a few lines, to retain alphabetical order

julianhyde (changes requested):
almost there. one or two changes and I'll approve.

you don't need to do anything with big-query.iq - I'll handle that during merge.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
tjbanghart: These can probably be simplified to run a single time unit but they validate for an error message if the BigQuery operator table is not used. I can add similar check in `SqlOperatorTest` so this file isn't touched.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
tjbanghart: This is me following an example commit too closely.

PR #2915, opened 2022-09-20
 [CALCITE-5293] Support general set operators in PruneEmptyRules

kasakrisz:


core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -94,7 +94,7 @@ public interface Config extends RelRule.Config {
   public static final RelOptRule UNION_INSTANCE =
       ImmutableUnionEmptyPruneRuleConfig.of()
           .withOperandSupplier(b0 ->
-              b0.operand(LogicalUnion.class).unorderedInputs(b1 ->
+              b0.operand(Union.class).unorderedInputs(b1 ->
```
zabetak, line 97: Better leave the default configurations unchanged to avoid accidentally breaking other projects. Projects who wish to enlarge the matching scope can modify the configuration.

zabetak (commented):
LGTM, just one small comment about backwards compatibility.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
kasakrisz: Only set operator rules configured for Logical operators so it makes a bit inconsistent the default config.
Anyway I got your point and reverted.

zabetak: Hmm, I didn't notice that the other rules were configured for the general operators. Your initial proposal makes sense. Please revert https://github.com/apache/calcite/pull/2915/commits/ad4cfeaf49100f0ce7f5668c0c55fc2fbd8a0a8a and add a note in `history.md` about the slight change in behavior in the default configuration regarding these rules. 

I will merge the PR afterwards; sorry for the back n forth.

asolimando (approved):
LGTM modulo pending tests

site/_docs/history.md
```diff
@@ -44,6 +44,10 @@ z.
 #### Breaking Changes
 {: #breaking-1-33-0}
 
+* [<a href="https://issues.apache.org/jira/browse/CALCITE-5293">CALCITE-5293</a>]
+  Support general set operators in PruneEmptyRules. The default configuration of PruneEmptyRules for Set operators is changed: the rules matching scope has increased.
```
asolimando, line 48: Nitpick:
```suggestion
  Support general set operators in PruneEmptyRules. The default configuration of PruneEmptyRules for Set operators has changed: the rules matching scope has increased.
```

PR #2916, opened 2022-09-21
[CALCITE-5294] Improve PruneEmptyRules join instances to remove join if generates null and one branch is empty

kasakrisz:


core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -480,9 +486,24 @@ public interface JoinLeftEmptyRuleConfig extends PruneEmptyRule.Config {
       return new PruneEmptyRule(this) {
         @Override public void onMatch(RelOptRuleCall call) {
           Join join = call.rel(0);
+          Values empty = call.rel(1);
+          RelNode right = call.rel(2);
+          RexBuilder rexBuilder = call.builder().getRexBuilder();
           if (join.getJoinType().generatesNullsOnLeft()) {
             // "select * from emp right join dept" is not necessarily empty if
```
rubenada, line 493: Probably this comment (and the analogous one in `JoinRightEmptyRuleConfig`) needs to be adjusted because of this change.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -491,6 +512,30 @@ public interface JoinLeftEmptyRuleConfig extends PruneEmptyRule.Config {
...
+      RexNode nullLiteral = rexBuilder.makeNullLiteral(relDataTypeField.getType());
+      projectFields.add(nullLiteral);
+      newColumnNames.add(empty.getRowType().getFieldList().get(i).getName());
+    }
+  }
+
+  public static void copyProjects(RexBuilder rexBuilder, RelDataType inRowType,
+      RelDataType castRowType, int castRowTypeOffset,
+      List<RexNode> outProjects, List<String> outProjectNames) {
+
```
rubenada, line 528: Nitpick: could you please remove this empty line to keep it consistent with the rest of the code?

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
kasakrisz: Updated the comment.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
kasakrisz: removed

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -491,6 +513,29 @@ public interface JoinLeftEmptyRuleConfig extends PruneEmptyRule.Config {
...
+    for (int i = 0; i < inRowType.getFieldCount(); ++i) {
+      RelDataTypeField relDataTypeField = inRowType.getFieldList().get(i);
+      RexInputRef inputRef = rexBuilder.makeInputRef(relDataTypeField.getType(), i);
+      RexNode cast = rexBuilder.makeCast(
+          castRowType.getFieldList().get(castRowTypeOffset + i).getType(), inputRef);
+      outProjects.add(cast);
+      outProjectNames.add(relDataTypeField.getName());
+    }
+  }
+
```
asolimando, line 538: As highlighted in the corresponding Hive PR, since those two methods are used by both rules, it would be better to move them either before them, or after them, rather than having them in between the rules as of now.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -491,6 +513,29 @@ public interface JoinLeftEmptyRuleConfig extends PruneEmptyRule.Config {
...
+      List<RexNode> projectFields, List<String> newColumnNames) {
+    for (int i = 0; i < empty.getRowType().getFieldList().size(); ++i) {
+      RelDataTypeField relDataTypeField = empty.getRowType().getFieldList().get(i);
+      RexNode nullLiteral = rexBuilder.makeNullLiteral(relDataTypeField.getType());
+      projectFields.add(nullLiteral);
+      newColumnNames.add(empty.getRowType().getFieldList().get(i).getName());
+    }
+  }
+
+  public static void copyProjects(RexBuilder rexBuilder, RelDataType inRowType,
```
asolimando, line 526: Does it need to be `public`?

kasakrisz:
Should this patch also be mentioned in `history.md` in `#### Bug-fixes, API changes and minor enhancements` section ?

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -480,9 +486,25 @@ public interface JoinLeftEmptyRuleConfig extends PruneEmptyRule.Config {
...
+            List<String> columnNames = new ArrayList<>(
+                empty.getRowType().getFieldCount() + right.getRowType().getFieldCount());
+            // left
+            addNullLiterals(rexBuilder, empty, projects, columnNames);
+            // right
+            copyProjects(rexBuilder, right.getRowType(),
+                join.getRowType(), empty.getRowType().getFieldCount(), projects, columnNames);
+
+            RelNode project = call.builder().push(right).project(projects, columnNames).build();
+            call.transformTo(project);
```
julianhyde, line 507: Can this be written in a functional style? Rather than creating empty lists then filling them, use `map`. `RelBuilder` has plenty of methods that return lists to types, lists of names, etc.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
kasakrisz: Renamed, refactored the methods and also moved after them.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
kasakrisz: No it should be private so I changed. Thanks.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -551,17 +529,15 @@ public interface JoinRightEmptyRuleConfig extends PruneEmptyRule.Config {
...
-            RelNode project = call.builder().push(left).project(projects, columnNames).build();
+            List<Pair<? extends RexNode, String>> projects =
+                Stream.concat(
+                    getProjectStream(left.getRowType(), join.getRowType(), 0, rexBuilder),
+                    getNullLiteralStream(empty, rexBuilder)).collect(Collectors.toList());
+
+            RelNode project = call.builder()
+                .push(left)
+                .project(Pair.left(projects), Pair.right(projects))
+                .build();
```
julianhyde, line 540: I believe that the field names are the same before and after. If so, you don't need to create pairs of (RexNode, String), just RexNode. Which I think would simplify getProjectStream and getNullLiteralStream.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
kasakrisz: Refactored the utility methods to use the Java stream api. I defined stream of List<Pair<>> as a return type to ensure the order of expressions.
If this is not what you mean could you please give some example?

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
kasakrisz: Sorry, just read your comment, let me try it.

kasakrisz: The field names can be inherited from the left branch but it has to be set for the `NULL` literals otherwise names like `$f9` are generated.
explected
```
LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], DEPTNO0=[$9], NAME=[$10])
  LogicalProject(EMPNO=[CAST($0):INTEGER], ENAME=[CAST($1):VARCHAR(20)], JOB=[CAST($2):VARCHAR(10)], MGR=[$3], HIREDATE=[CAST($4):TIMESTAMP(0)], SAL=[CAST($5):INTEGER], COMM=[CAST($6):INTEGER], DEPTNO=[CAST($7):INTEGER], SLACKER=[CAST($8):BOOLEAN], DEPTNO0=[null:INTEGER], NAME=[null:VARCHAR(10)])
```
actual
```
LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], DEPTNO0=[$9], NAME=[$10])
  LogicalProject(EMPNO=[CAST($0):INTEGER], ENAME=[CAST($1):VARCHAR(20)], JOB=[CAST($2):VARCHAR(10)], MGR=[$3], HIREDATE=[CAST($4):TIMESTAMP(0)], SAL=[CAST($5):INTEGER], COMM=[CAST($6):INTEGER], DEPTNO=[CAST($7):INTEGER], SLACKER=[CAST($8):BOOLEAN], $f9=[null:INTEGER], $f10=[null:VARCHAR(10)])
```

kasakrisz: It works, thanks for the hint.

julianhyde: I have a feeling that it could be even shorter if you use Guava's `List.concat()` rather than `Stream.concat()`, and use `RelBuilder.fields()`. You'll have to assign the `RelBuilder` to a variable.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -551,15 +554,13 @@ public interface JoinRightEmptyRuleConfig extends PruneEmptyRule.Config {
...
-      RelDataType inRowType,
-      RelDataType castRowType,
-      int castRowTypeOffset,
-      RexBuilder rexBuilder) {
-    return inRowType.getFieldList().stream().map(
-        typeField -> rexBuilder.makeCast(
-                castRowType.getFieldList().get(castRowTypeOffset + typeField.getIndex()).getType(),
-                rexBuilder.makeInputRef(typeField.getType(), typeField.getIndex())));
+  private static Stream<RexNode> castToNullable(
+      RexBuilder rexBuilder, Stream<RexNode> leftExpressions) {
```
asolimando, line 558: We probably want to use `expressions` instead of `leftExpressions` here, since the method can be called on left or right, depending on the kind of outer join.

kasakrisz: fixed

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -479,13 +481,24 @@ public interface JoinLeftEmptyRuleConfig extends PruneEmptyRule.Config {
     @Override default PruneEmptyRule toRule() {
       return new PruneEmptyRule(this) {
         @Override public void onMatch(RelOptRuleCall call) {
-          Join join = call.rel(0);
+          final Join join = call.rel(0);
+          final Values empty = call.rel(1);
+          final RelNode right = call.rel(2);
+          final RelBuilder b = call.builder();
```
asolimando, line 487: Nitpick: in the codebase this is generally called `relBuilder`, the method is not big but you might wonder what `b` is in the rest of the code, while with `relBuilder` it will be pretty clear

julianhyde: That's reasonable. I'll apply the change to my branch, and will merge tomorrow when my batch of changes passes.

kasakrisz: Renamed to `relBuilder`

asolimando (approved):
LGTM, solid contribution @kasakrisz!

PR #2918, opened 2022-09-23
[CALCITE-5297] Casting dynamic variable twice throws exception

dssysolyatin:
It was originally intended that dynamic parameter are nullable (see SqlValidatorImpl#inferUnknownTypes), but DeriveTypeVisitor made them non nullable. In the end row type has not been preserved after converting query from SqlNode to RelNode.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
@@ -235,8 +235,11 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
 
   private int nextGeneratedId;
   protected final RelDataTypeFactory typeFactory;
-
-  /** The type of dynamic parameters until a type is imposed on them. */
+  /**
+   * The type of dynamic parameters until a type is imposed on them.
+   * It is assumed that dynamic parameter is nullable.
+   * See {@link #inferUnknownTypes(RelDataType, SqlValidatorScope, SqlNode)}
+   */
   protected final RelDataType unknownType;
   private final RelDataType booleanType;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -235,8 +235,11 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
 
   private int nextGeneratedId;
   protected final RelDataTypeFactory typeFactory;
-
-  /** The type of dynamic parameters until a type is imposed on them. */
+  /**
+   * The type of dynamic parameters until a type is imposed on them.
+   * It is assumed that dynamic parameter is nullable.
+   * See {@link #inferUnknownTypes(RelDataType, SqlValidatorScope, SqlNode)}
```
libenchao, line 241: The fix seems right. However, I'm confused about the comment for `unknownType`, it is not only used for 'dynamic parameters' actually.

dssysolyatin: I can remove the comment from here and add new comment near assignment operator like:
```
/**
* It is assumed that unknown type is nullable by default
*/
unknownType = typeFactory.createTypeWithNullability(typeFactory.createUnknownType(), true);
```

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -4678,4 +4678,14 @@ void checkCorrelatedMapSubQuery(boolean expand) {
...
         .ok();
   }
+
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5297">[CALCITE-5297]
+   * Casting dynamic variable twice throws exception</a>.
+   */
+  @Test void testDynamicParameterDoubleCast() {
+    String sql = "SELECT CAST(CAST(? AS INTEGER) AS INTEGER)";
```
clesaec, line 4688: [JIRA](https://issues.apache.org/jira/browse/CALCITE-5297) is about Cast Integer, then Char (_CAST(CAST(? AS INTEGER) AS CHAR)_), test is for double CAST to INTEGER. Is it intentional ?

dssysolyatin: but not to confuse anyone with Jira ticket. I think I will return back old test case

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
libenchao: The new comment looks good.

libenchao (approved):
@dssysolyatin LGTM, thanks for the PR, merging.

PR #2919, opened 2022-09-25
[CALCITE-5298] CalciteSystemProperty calcite.test.dataset path check fails under Java Security Manager

risdenk:
SOLR-16433 found that Calcite does not handle the Java security manager returning permission denied when checking that the calcite.test.dataset path exists. Solr runs with a security manager that doesn't allow arbitrary filesystem access. This failure causes Calcite to not load and therefore unusable.

The code in question is here: https://github.com/apache/calcite/blame/main/core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java#L189

A few other places in Calcite already check for SecurityException: https://github.com/apache/calcite/search?q=SecurityException

### DIFF
--- a/core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
+++ b/core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
@@ -196,8 +196,13 @@ public final class CalciteSystemProperty<T> {
             "../../calcite-test-dataset"
         };
         for (String s : dirs) {
-          if (new File(s).exists() && new File(s, "vm").exists()) {
-            return s;
+          try {
+            if (new File(s).exists() && new File(s, "vm").exists()) {
+              return s;
+            }
+          } catch (SecurityException ignore) {
+            // Ignore SecurityException on purpose because if
+            // we can't get to the file we fall through.
           }
         }
         return ".";
@@ -449,10 +454,9 @@ public final class CalciteSystemProperty<T> {
       }
     } catch (IOException e) {
       throw new RuntimeException("while reading from saffron.properties file", e);
-    } catch (RuntimeException e) {
-      if (!"java.security.AccessControlException".equals(e.getClass().getName())) {
-        throw e;
-      }
+    } catch (SecurityException ignore) {
+      // Ignore SecurityException on purpose because if
+      // we can't get to the file we fall through.
     }
 
     // Merge system and saffron properties, mapping deprecated saffron
--- a/core/src/main/java/org/apache/calcite/util/SaffronProperties.java
+++ b/core/src/main/java/org/apache/calcite/util/SaffronProperties.java
@@ -133,10 +133,9 @@ public interface SaffronProperties {
         }
       } catch (IOException e) {
         throw new RuntimeException("while reading from saffron.properties file", e);
-      } catch (RuntimeException e) {
-        if (!"java.security.AccessControlException".equals(e.getClass().getName())) {
-          throw e;
-        }
+      } catch (SecurityException ignore) {
+        // Ignore SecurityException on purpose because if
+        // we can't get to the file we fall through.
       }
 
       // copy in all system properties which start with "saffron."


core/src/main/java/org/apache/calcite/util/Util.java
```diff
@@ -1048,6 +1048,19 @@ public static String getStackTrace(Throwable t) {
...
   }
 
+  /**
+   * Checks a RuntimeException for AccessControlException without importing JDK
+   * classes that are deprecated with the anticipated removal of Java security manager.
+   * Ignores the RuntimeException if it is an AccessControlException otherwise throws.
+   * @param e RuntimeException to check if it is an AccessControlException
+   */
+  @API(since = "1.33", status = API.Status.EXPERIMENTAL)
+  public static void ignoreAccessControlException(RuntimeException e) {
```
risdenk, line 1058: So I thought about trying to put the whole `try/catch` in here but that got into issues with return values and handling other exceptions. There may be a way to do this with lambdas or functions passed in. I just punted on that and made this a simple method just for checking `AccessControlException`

julianhyde: Is this issue tied to any particular JDK version? E.g. SecurityManager became deprecated in JDK x.x and is going to be removed in JDK y.y? If so it would be helpful to mention in this comment.

julianhyde (commented):
Looks good. With a couple of javadoc changes this can go in to 1.33.

core/src/main/java/org/apache/calcite/util/Util.java
risdenk: Fixed in ede956c7aecf560ea88c45b48ef7a351f0c8e891

risdenk:
> SonarCloud Quality Gate failed.    [Quality Gate failed](https://sonarcloud.io/dashboard?id=apache_calcite&pullRequest=2919)
> 
> [Bug](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=2919&resolved=false&types=BUG) [C](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=2919&resolved=false&types=BUG) [1 Bug](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=2919&resolved=false&types=BUG) [Vulnerability](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=2919&resolved=false&types=VULNERABILITY) [A](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=2919&resolved=false&types=VULNERABILITY) [0 Vulnerabilities](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=2919&resolved=false&types=VULNERABILITY) [Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=apache_calcite&pullRequest=2919&resolved=false&types=SECURITY_HOTSPOT) [A](https://sonarcloud.io/project/security_hotspots?id=apache_calcite&pullRequest=2919&resolved=false&types=SECURITY_HOTSPOT) [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=apache_calcite&pullRequest=2919&resolved=false&types=SECURITY_HOTSPOT) [Code Smell](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=2919&resolved=false&types=CODE_SMELL) [A](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=2919&resolved=false&types=CODE_SMELL) [0 Code Smells](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=2919&resolved=false&types=CODE_SMELL)
> 
> [](https://sonarcloud.io/component_measures?id=apache_calcite&pullRequest=2919&metric=new_coverage&view=list) [10.0% Coverage](https://sonarcloud.io/component_measures?id=apache_calcite&pullRequest=2919&metric=new_coverage&view=list) [](https://sonarcloud.io/component_measures?id=apache_calcite&pullRequest=2919&metric=new_duplicated_lines_density&view=list) [0.0% Duplication](https://sonarcloud.io/component_measures?id=apache_calcite&pullRequest=2919&metric=new_duplicated_lines_density&view=list)

I left a comment on this about the one "bug" and how instanceof should not be used in this case - we need to avoid the AccessControlException import.

core/src/main/java/org/apache/calcite/util/Util.java
```diff
@@ -1048,6 +1048,22 @@ public static String getStackTrace(Throwable t) {
...
+   * with the anticipated removal of Java security manager.
+   *
+   * @param e RuntimeException to check if it is an AccessControlException
+   */
+  @API(since = "1.33", status = API.Status.EXPERIMENTAL)
+  public static void ignoreAccessControlException(RuntimeException e) {
+    if (!"java.security.AccessControlException".equals(e.getClass().getName())) {
+      throw e;
+    }
+  }
```
zabetak, line 1065: Do we really need this hacky check? `AccessControlException` is a subclass of `SecurityException`. The latter is not deprecated (https://download.java.net/java/early_access/jdk21/docs/api/java.base/java/lang/SecurityException.html) so I assume we can do regular `catch(SecurityException se)` where needed can't we?

zabetak (changes requested):

core/src/main/java/org/apache/calcite/util/Util.java
risdenk: Seems like a reasonable idea. From https://openjdk.org/jeps/411

```
...
We will not deprecate some classes in the java.security package that are related to the Security Manager, for various reasons:
...
SecurityException — A runtime exception thrown by Java APIs when a permission check fails. We may deprecate this API for removal at a later date, but for now the impact of doing so would be too high.
...
```

risdenk: Switched to this approach in 88c31c9f29f22754443e2c67b3e9e9e293972fcf

zabetak (approved):
LGTM, pending tests!

risdenk:
Squashed all the changes together to make it one commit.

PR #2924, opened 2022-09-29
Add ST_Explode function

bchapuis:


core/src/main/java/org/apache/calcite/runtime/SpatialTypeTableFunctions.java
```diff
@@ -0,0 +1,158 @@
...
+ * Helper methods to implement spatial type (ST) table functions.
+ */
+@Deterministic
+@Strict
+@Experimental
+public class SpatialTypeTableFunctions {
+
+  private SpatialTypeTableFunctions() {}
+
+  public static final Method EXPLODE =
```
julianhyde, line 59: I think you should move `EXPLODE` and `EXPLODE_CURSOR` into `enum BuiltInMethod`

core/src/main/java/org/apache/calcite/runtime/SpatialTypeTableFunctions.java
```diff
@@ -0,0 +1,158 @@
...
+      Types.lookupMethod(SpatialTypeTableFunctions.class,
+          "explode", Geometry.class);
+
+  public static final Method EXPLODE_CURSOR =
+      Types.lookupMethod(SpatialTypeTableFunctions.class,
+          "explodeCursor", Enumerable.class);
+
+  public static ScannableTable explode(Geometry geometry) {
```
julianhyde, line 67: The method needs javadoc. 

A better way to slice the functionality might be to move `explode` into `SpatialTypeFunctions` (from SQL's perspective, a table function is just a function) and make the anonymous subclass of `ScannableTable` into a static inner class of `SpatialTypeTableFunctions` that is package-private.

core/src/main/java/org/apache/calcite/runtime/SpatialTypeTableFunctions.java
```diff
@@ -0,0 +1,158 @@
...
+        final Enumerable<Geometry> enumerable = a.selectMany(a0 -> {
+          final Geometry geometry = (Geometry) a0[0];
+          return new GeometryElements(geometry);
+        });
+        return (Queryable) enumerable.asQueryable();
+      }
+    };
+  }
+
+  /**
```
julianhyde, line 118: How about using the following?
```
Linq4j.asEnumerable(
  Functions.generate(geometry.getNumGeometries(),
      i -> geometry.getGeometryN(i)));
```

It seems to capture the random-access-generating-list nature, and yours is rather more specific to Geometry than it needs to be. Also it has optimized implementations of `skip` , `take`, `elementAt`, `count`, etc.

core/src/test/java/org/apache/calcite/test/SpatialTypeTableFunctionsTest.java
```diff
@@ -0,0 +1,106 @@
...
+              + "geom=POINT (3 3)\n"
+              + "geom=POINT (4 4)\n"));
+    }
+  }
+
+  /**
+   * Tests the ST_EXPLODECURSOR table function.
+   */
+  @Disabled()
+  @Test void testExplodeCursorFunction() throws SQLException {
```
julianhyde, line 78: Is it a goal to implement `ST_ExploreCursor` in this change? If so this test should be enabled.

core/src/main/java/org/apache/calcite/runtime/SpatialTypeTableFunctions.java
bchapuis: Ok, I will do that.

Notice that the EXPLODE_CURSOR function is currently not working and the corresponding examples in `TableFunctionTest.testTableFunctionCursorInputs` neither (the test is disabled). Is there another way to achieve the same result?

core/src/test/java/org/apache/calcite/test/SpatialTypeTableFunctionsTest.java
bchapuis: The EXPLODE_CURSOR function is currently not working and the corresponding examples in `TableFunctionTest.testTableFunctionCursorInputs` neither (the original test is disabled as well). Is there another way to achieve the same result?

core/src/test/java/org/apache/calcite/test/SpatialTypeTableFunctionsTest.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Tests for spatial table functions.
+ */
+public class SpatialTypeTableFunctionsTest {
+
+  /**
+   * Tests the ST_Explode table function.
+   */
+  @Test void testExplodeFunction() throws SQLException {
+    try (Connection connection = DriverManager.getConnection("jdbc:calcite:")) {
```
julianhyde, line 50: Is it possible to run this test using CalciteAssert? This test has a lot of boilerplate to maintain, and I don't want to see this boilerplate copy-pasted 100 times.

core/src/test/java/org/apache/calcite/test/SpatialTypeTableFunctionsTest.java
julianhyde: I don't recall what is and isn't working with table functions. Best thing to do is to log a Jira case with what's not working, and make [CALCITE-5281] depend on it.

bchapuis: Thanks, i logged [CALCITE-5309](https://issues.apache.org/jira/browse/CALCITE-5309) and will continue the discussion in [CALCITE-5281](https://issues.apache.org/jira/browse/CALCITE-5281).

core/src/main/java/org/apache/calcite/runtime/SpatialTypeTableFunctions.java
bchapuis: Same comment as above.

core/src/main/java/org/apache/calcite/runtime/SpatialTypeTableFunctions.java
bchapuis: Thank you for the pointer on Linq4j, this has been fixed.

core/src/test/java/org/apache/calcite/test/SpatialTypeTableFunctionsTest.java
bchapuis: I removed the cursor functions for now and tried to clarify the proble in [CALCITE-5281](https://issues.apache.org/jira/browse/CALCITE-5281).

core/src/test/java/org/apache/calcite/test/SpatialTypeTableFunctionsTest.java
bchapuis: This class has been removed and the tests added to `spatial.iq`.

PR #2925, opened 2022-09-30
[CALCITE-4871] Fix CAST a literal to DECIMAL type return wrong result

wanda1416:


wanda1416:
@libenchao Can you review this? thanks.

libenchao:
@wanda1416 Thanks for your contribution, I can review this. However, before that, I would like to hear @NobiGo 's opinion since he is the original reporter and assignee of this issue.

PR #2928, opened 2022-10-04
[CALCITE-5306] Support LTS JDKs and latest in testing

asolimando:
Dropped JDK 15 and 16 by adopting the latest LTS version (JDK 17) or latest overall (JDK18). JDK19 is at the moment incompatible with `forbiddenAPIs`, so it's not an option at the moment.


zabetak (approved):
LGTM, thanks for stabilizing the CI.

.github/workflows/main.yml
```diff
@@ -123,7 +123,7 @@ jobs:
 
   linux-openj9:
```
michaelmior, line 124: It seems there is still some inconsistency here. I'm not sure if the original intent was to run on macOS or Linux, but I think the key here, the name, and the actual OS used should all match.

rubenada: Just to provide some info. IIRC correctly, this was added due to some strange Error that appeared only on WebSphere (ie with IBM's JVM), see details on email discussion: https://lists.apache.org/thread/qmo39634fpdxnwconoo80zbczo0wy1wq

asolimando: Thanks @rubenada for pointing to the original discussion. Did you experience the issue described there on a linux server I guess, would you recommend switching to a linux-based image?

If we are not covering the specific use-case on the original OS we might even consider removing that job altogether and focus on more recent JDK versions or reduce the load on CI.

rubenada: @asolimando  IIRC the problem appeared on an Ubutu + WebSphere environment.
I agree, IMO we could consider removing this job and, and you say, focus on recent JDKs.

PR #2929, opened 2022-10-06
[CALCITE-5310] JSON_OBJECT in scalar sub-query throws AssertionError

libenchao:


core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -1790,10 +1790,11 @@ private RexNode createCaseExpression(
 
     @Override public RexNode visitLiteral(RexLiteral literal) {
       // Use nullIndicator to decide whether to project null.
-      // Do nothing if the literal is null.
+      // Do nothing if the literal is null or symbol.
       if (!RexUtil.isNull(literal)
```
dssysolyatin, line 1794: Do I understand correctly that if I have query like:
```
SELECT (SELECT CARDINALITY(ARRAY[false, true, a.attidentity = 'a']) FROM UNNEST(ARRAY[1]) as v) as options
FROM UNNEST(ARRAY['a', 'b']) AS a(attidentity);
```
RelDecorrelator will try to wrap true and false into case expression ?

Another question. why do we need to wrap literals into case expression at all ?

dssysolyatin: Ok, I got why calcite wraps literals if literal is in SELECT list like `SELEC (SELECT 1 ....)`. But I still don't understand why `CARDINALITY` arguments should be wrapped in case expression ?
```
CARDINALITY(ARRAY[false, true, a.attidentity = 'a'])
```
In my understanding only CARDINALITY should be wrapped to caseExpression

libenchao: The whole story lies in `RelDecorrelator#RemoveCorrelationRexShuttle`. When `projectPulledAboveLeftCorrelator && (nullIndicator != null)` is `true`, currently it will add a case for all expressions.  

Maybe your feeling is right, I haven't thought it thorough.

dssysolyatin: Ok, I was just trying to understand if we can reduce number of caseExpression. Anyway it is not directly related to this ticket.

LGTM

dssysolyatin (approved):

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
libenchao: @dssysolyatin Thanks for your review.

It's good to have such questions and discussions.

PR #2931, opened 2022-10-06
[CALCITE-5248] Support extended column aliasing in select list and where, on and filter clauses

jaystarshot:
Based on the ticket https://issues.apache.org/jira/browse/CALCITE-5248

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -2823,14 +2823,17 @@ class RelToSqlConverterTest {
     sql(query).ok(expected);
   }
 
-  @Test public void testSelectQueryAliasInWhereClauseAndOrderBy() {
-    String query = "select  \"cases_per_pallet\" as c, upper(c) from \"product\" WHERE c > 100 ORDER BY c";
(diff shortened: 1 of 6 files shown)


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -2815,6 +2815,45 @@ private SqlDialect nonOrdinalDialect() {
...
+
+  @Test public void testSelectQueryAliasPreferenceForInnerSubQuery() {
+    String query = "select  \"employee_id\" as c, lower(c) from "
+            + "( SELECT \"employee_id\", \"gender\" AS c " +
+           "FROM \"employee\")";
+
+    final String expected = "SELECT \"employee_id\" AS \"C\", LOWER(\"gender\")\nFROM \"foodmart\".\"employee\"";
+    sql(query).ok(expected);
+  }
+
```
JiajunBernoulli, line 2854: Calcite support group by alias, would you please add some unit test that alias be used by aggregate and filter?

jaystarshot: Hi Jianjun, thanks for the review and I have added a simple unit test where alias is used both in filter and aggregate

Akanksha-kedia:
can u add test case and scenarios for btr understanding of the pr.

PR #2932, opened 2022-10-07
[CALCITE-5304] RelJson should support RexSubQuery

libenchao:


core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -122,10 +127,32 @@ public RelJson withInputTranslator(InputTranslator inputTranslator) {
...
+
   private JsonBuilder jsonBuilder() {
     return requireNonNull(jsonBuilder, "jsonBuilder");
   }
 
+  private RelJsonWriter relJsonWriter() {
+    return requireNonNull(relJsonWriter, "relJsonWriter");
+  }
+
+  private RelJsonReader relJsonReader() {
```
chunweilei, line 152: It's a little weird that you add RelJsonWriter and RelJsonReader to RelJson. Why RelJson has to use them?

libenchao: Because there is a `RelNode` in `RexSubQuery`, to serialize/deserialize a `RelNode`, we need to use `RelJsonWriter` and `RelJsonReader`.

PR #2935, opened 2022-10-10
[CALCITE-5314] Prune empty parts of a query by exploiting stats/metadata

HanumathRao:
The following changes are made to implement an optimization for pruning out the sub trees when a base-table is empty. The detection of a table being empty is done using MaxRow stat.

I have not used existing stats like rowcount etc for triggering the optimization as they are estimates and can be stale. I also thought about introducing a new stat like exactRowCount but it seems like an overkill (just for base table it is valid and it soon will be an estimate after base table for any other node in the tree). After due consideration, it seems like MaxRow stat fits well for this scenario.
1) It is not a default stat and it needs to be supplied for a table by overloading the MaxRowHandler.
2) Default value of the MaxRow is large value (Infinity) and hence the optimization doesn't trigger by default.

EmptyTableOptimizationConfig change adds a new rule to transform the base table to empty values node when maxRowCount is zero. All the existing PruneEmptyRules will do the necessary optimizations to prune out the complex sub-tree once the emptyValues node is created.

Please review the changes.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
@@ -470,6 +470,7 @@ public abstract class PruneEmptyRules {
             call.transformTo(emptyValues);
           }
         }
+
       };
     }
   }
@@ -542,28 +543,35 @@ public abstract class PruneEmptyRules {
   }
 
 
-  public static final RelOptRule EMPTY_TABLE =
-      ImmutableEmptyTableOptimizationConfig.of()
-          .withOperandSupplier(b0 -> {
-            return b0.operand(TableScan.class).noInputs();
-          })
-          .withDescription("ConvertEmptyTableToValues")
+  public static final RelOptRule EMPTY_TABLE_INSTANCE =
+      ImmutableZeroMaxRowsRuleConfig.of()
+          .withOperandSupplier(b0 -> b0.operand(TableScan.class).noInputs())
+          .withDescription("PruneZeroRowsTable")
           .toRule();
 
   /** Configuration for rule that transforms an empty table into an empty values node.
    * MaxRowCount is used as the stat to transform, hence Table implementer needs
    * to supply this metadata if this optimization needs to be applied.*/
   @Value.Immutable
-  public interface EmptyTableOptimizationConfig extends PruneEmptyRule.Config {
+  public interface ZeroMaxRowsRuleConfig extends PruneEmptyRule.Config {
 
     @Override default PruneEmptyRule toRule() {
       return new PruneEmptyRule(this) {
         @Override public void onMatch(RelOptRuleCall call) {
-          TableScan tableScan = call.rel(0);
-          Double maxRowCount = call.getMetadataQuery().getMaxRowCount(tableScan);
-          if (maxRowCount != null && maxRowCount == 0.0) {
-            call.transformTo(call.builder().push(tableScan).empty().build());
+          RelNode node = call.rel(0);
+          Double maxRowCount = call.getMetadataQuery().getMaxRowCount(node);
+          if (maxRowCount == null || maxRowCount > 0.0) {
+            return;
           }
+          RelNode emptyValues = call.builder().push(node).empty().build();
+          RelTraitSet traits = node.getTraitSet();
+          // propagate all traits (except convention) from the original tableScan
+          // into the empty values
+          if (emptyValues.getConvention() != null) {
+            traits = traits.replace(emptyValues.getConvention());
+          }
+          emptyValues = emptyValues.copy(traits, Collections.emptyList());
+          call.transformTo(emptyValues);
         }
       };
     }
--- a/core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
@@ -572,10 +572,6 @@ public abstract class PruneEmptyRules {
 
         @Override public void onMatch(RelOptRuleCall call) {
           RelNode node = call.rel(0);
-          Double maxRowCount = call.getMetadataQuery().getMaxRowCount(node);
-          if (maxRowCount == null || maxRowCount > 0.0) {
-            return;
-          }
           RelNode emptyValues = call.builder().push(node).empty().build();
           RelTraitSet traits = node.getTraitSet();
           // propagate all traits (except convention) from the original tableScan
--- a/core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
@@ -33,6 +33,7 @@ import org.apache.calcite.rel.core.JoinRelType;
 import org.apache.calcite.rel.core.Minus;
 import org.apache.calcite.rel.core.Project;
 import org.apache.calcite.rel.core.Sort;
+import org.apache.calcite.rel.core.TableScan;
 import org.apache.calcite.rel.core.Union;
 import org.apache.calcite.rel.core.Values;
 import org.apache.calcite.rel.logical.LogicalValues;
@@ -469,7 +470,6 @@ public abstract class PruneEmptyRules {
             call.transformTo(emptyValues);
           }
         }
-
       };
     }
   }
@@ -536,6 +536,34 @@ public abstract class PruneEmptyRules {
             return;
           }
           call.transformTo(relBuilder.push(join).empty().build());
+        }
+      };
+    }
+  }
+
+
+  public static final RelOptRule EMPTY_TABLE =
+      ImmutableEmptyTableOptimizationConfig.of()
+          .withOperandSupplier(b0 -> {
+            return b0.operand(TableScan.class).noInputs();
+          })
+          .withDescription("ConvertEmptyTableToValues")
+          .toRule();
+
+  /** Configuration for rule that transforms an empty table into an empty values node.
+   * MaxRowCount is used as the stat to transform, hence Table implementer needs
+   * to supply this metadata if this optimization needs to be applied.*/
+  @Value.Immutable
+  public interface EmptyTableOptimizationConfig extends PruneEmptyRule.Config {
+
+    @Override default PruneEmptyRule toRule() {
+      return new PruneEmptyRule(this) {
+        @Override public void onMatch(RelOptRuleCall call) {
+          TableScan tableScan = call.rel(0);
+          Double maxRowCount = call.getMetadataQuery().getMaxRowCount(tableScan);
+          if (maxRowCount != null && maxRowCount == 0.0) {
+            call.transformTo(call.builder().push(tableScan).empty().build());
+          }
         }
       };
     }
--- a/core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
@@ -167,6 +167,19 @@ public abstract class PruneEmptyRules {
   }
 
   /**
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -540,4 +540,32 @@ public interface JoinRightEmptyRuleConfig extends PruneEmptyRule.Config {
...
+   * to supply this metadata if this optimization needs to be applied.*/
+  @Value.Immutable
+  public interface EmptyTableOptimizationConfig extends PruneEmptyRule.Config {
+
+    @Override default PruneEmptyRule toRule() {
+      return new PruneEmptyRule(this) {
+        @Override public void onMatch(RelOptRuleCall call) {
+          TableScan tableScan = call.rel(0);
+          Double maxRowCount = call.getMetadataQuery().getMaxRowCount(tableScan);
+          if (maxRowCount != null && maxRowCount == 0.0) {
```
zabetak, line 564: It is better to put the row count check in the `matches` method. When the condition is not satisfied the rule can be eliminated much earlier.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -540,4 +540,32 @@ public interface JoinRightEmptyRuleConfig extends PruneEmptyRule.Config {
...
+  /** Configuration for rule that transforms an empty table into an empty values node.
+   * MaxRowCount is used as the stat to transform, hence Table implementer needs
+   * to supply this metadata if this optimization needs to be applied.*/
+  @Value.Immutable
+  public interface EmptyTableOptimizationConfig extends PruneEmptyRule.Config {
+
+    @Override default PruneEmptyRule toRule() {
+      return new PruneEmptyRule(this) {
+        @Override public void onMatch(RelOptRuleCall call) {
+          TableScan tableScan = call.rel(0);
```
zabetak, line 562: It is better to be more general here (use `RelNode` here instead of `TableScan`); the rule will still work correctly and if people want to reduce or enlarge the scope they can do it by changing the matching operands of the configuration.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -540,4 +540,32 @@ public interface JoinRightEmptyRuleConfig extends PruneEmptyRule.Config {
...
+          Double maxRowCount = call.getMetadataQuery().getMaxRowCount(tableScan);
+          if (maxRowCount != null && maxRowCount == 0.0) {
+            call.transformTo(call.builder().push(tableScan).empty().build());
```
zabetak, line 565: The effect of this rule is very similar to what `RemoveEmptySingleRule` is doing. The latter is also trying to conserve the traits so maybe it is a good idea to attemp to conserve the traits as well here.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -540,4 +540,32 @@ public interface JoinRightEmptyRuleConfig extends PruneEmptyRule.Config {
...
     }
   }
+
+
+  public static final RelOptRule EMPTY_TABLE =
+      ImmutableEmptyTableOptimizationConfig.of()
+          .withOperandSupplier(b0 -> {
+            return b0.operand(TableScan.class).noInputs();
+          })
+          .withDescription("ConvertEmptyTableToValues")
```
zabetak, line 550: nit: To keep things uniform maybe rename to `"PruneZeroRowsTable"`

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -540,4 +540,32 @@ public interface JoinRightEmptyRuleConfig extends PruneEmptyRule.Config {
       };
     }
   }
+
+
+  public static final RelOptRule EMPTY_TABLE =
+      ImmutableEmptyTableOptimizationConfig.of()
+          .withOperandSupplier(b0 -> {
+            return b0.operand(TableScan.class).noInputs();
+          })
```
zabetak, line 549: I think it can be simplified to `.withOperandSupplier(b0 -> b0.operand(TableScan.class).noInputs());`

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -540,4 +540,32 @@ public interface JoinRightEmptyRuleConfig extends PruneEmptyRule.Config {
       };
     }
   }
+
+
+  public static final RelOptRule EMPTY_TABLE =
```
zabetak, line 545: All other rules in this class are suffixed with `INSTANCE`. To keep things uniform I would suggest one of the following:
* `EMPTY_TABLE_INSTANCE`
* `TABLE_ZERO_ROWS_INSTANCE`
* `ZERO_ROWS_TABLE_INSTANCE`

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReader.java
```diff
@@ -618,6 +621,57 @@ public StructKind getKind() {
     }
   }
 
+  /**
+   * Mock implementation of
+   * {@link org.apache.calcite.prepare.Prepare.PreparingTable} which supplies MaxRow stat.
+   */
+  public static class MaxRowMockTable
+      extends MockTable implements BuiltInMetadata.MaxRowCount.Handler {
```
zabetak, line 629: Maybe you can avoid introducing a new sub-class and make `MockTable` implement the `MaxRowCount.Handler` assuming that rowCount == maxRowCount at this point. If we need to differentiate in the future we can add or modify the constructors/factories in `MockTable`.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -3413,6 +3413,54 @@ RelOptFixture checkDynamicFunctions(boolean treatDynamicCallsAsConstant) {
...
+  @Test void testEmptyTableJoinRight() {
+    // inner join is eliminated after the right table is transformed to empty values.
+    final String sql = "select * from products as e\n"
+        + ",EMPTY_PRODUCTS as d where e.PRODUCTID = d.PRODUCTID\n";
+    sql(sql)
+        .withRule(
+            PruneEmptyRules.EMPTY_TABLE,
+            PruneEmptyRules.JOIN_RIGHT_INSTANCE)
+        .check();
+  }
```
zabetak, line 3446: I don't think need a test case for every possible combination of rules of `EMPTY_TABLE` with the others. I think we can keep just two:
* `testEmptyTable` which just tests the new rule.
* `testEmptyTableInComplexQuery` which exploits multiple pruning rules including `EMPTY_TABLE` to simplify a plan completely leaving only an empty `LogicalValues` at the end.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
zabetak, line 545: I think it would be a good idea to add the new rule to `org.apache.calcite.plan.RelOptRules#ABSTRACT_RULES` list along with the other pruning rules.

HanumathRao:
Thanks @zabetak for the review. I have addressed all the comments except the following. 

_It is better to put the row count check in the matches method. When the condition is not satisfied the rule can be eliminated much earlier._

I am using the call context to get the RelMetadataQuery, I think this is not fully formed during the matching of the rule. I thought about using the MetadataQuery singleton INSTANCE, but it can get complex during the matching. Please let me know if you have an easier way in mind.

chunweilei (commented):
LGTM

zabetak:
> I am using the call context to get the RelMetadataQuery, I think this is not fully formed during the matching of the rule

@HanumathRao can you elaborate a bit what you mean when you say that the metadata query is not formed during `matches`? Both `matches` and `onMatch` take a `RelOptRuleCall` as parameter and in both cases `call.getMetadataQuery()` is available.

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java
```diff
@@ -261,6 +261,14 @@ protected MockCatalogReaderSimple(RelDataTypeFactory typeFactory,
     productsTable.addColumn("SUPPLIERID", fixture.intType);
     registerTable(productsTable);
 
+    // Register "EMPTY_PRODUCTS" table.
+    MockTable emptyProductsTable = MockTable.create(this, salesSchema, "EMPTY_PRODUCTS",
+        false, 2000D, 0.0);
```
zabetak, line 266: Is it intentional to have rows > masRows?

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -3058,6 +3058,50 @@ LogicalSort(sort0=[$7], dir0=[ASC], fetch=[0])
...
+      LogicalFilter(condition=[>($2, 10)])
+        LogicalJoin(condition=[=($0, $3)], joinType=[inner])
+          LogicalTableScan(table=[[CATALOG, SALES, PRODUCTS]])
+          LogicalTableScan(table=[[CATALOG, SALES, EMPTY_PRODUCTS]])
+]]>
+    </Resource>
+    <Resource name="planAfter">
+      <![CDATA[
+LogicalProject(PRODUCTID=[$0], NAME=[$1], SUPPLIERID=[$2], PRODUCTID0=[$3], NAME0=[$4], SUPPLIERID0=[$5], PRODUCTID00=[$6], NAME00=[$7], SUPPLIERID00=[$8])
+  LogicalJoin(condition=[=($0, $3)], joinType=[left])
```
zabetak, line 3102: I think that by adding the appropriate rules the left join can also be removed.

HanumathRao: I have tried to use most of the PruneEmptyRules and it didn't prune the Left join. This is because of the way the rules are applied one after the other. There is a dependency of the rules in this particular example PROJECT (EMPTY) needs to be applied before LEFT JOIN(TABLE, EMPTY). PROJECT(EMPTY) depends on the INNER JOIN (TABLE, EMPTY) to be applied. Please let me know if you have any particular rule in mind, then I can try it out. thanks

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java
HanumathRao: It was intentional but fixed it anyway in the latest commit.

HanumathRao:
> > I am using the call context to get the RelMetadataQuery, I think this is not fully formed during the matching of the rule
> 
> @HanumathRao can you elaborate a bit what you mean when you say that the metadata query is not formed during `matches`? Both `matches` and `onMatch` take a `RelOptRuleCall` as parameter and in both cases `call.getMetadataQuery()` is available.

Thanks @zabetak, I think I misunderstood your earlier comment. I thought that you are asking me to introduce a predicate in operand supplier itself to rule out the application of the rule. I fixed the code in the latest commit to return false for not applicable cases in matches method.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
zabetak: What I had in mind is something like https://github.com/apache/calcite/pull/2935/commits/f4d0777db3f2f8618c8d796024e9bd3db4f733f8.

zabetak:
I pushed a few small changes to the PR. @HanumathRao let me know if you are OK with those so that I can rebase and merge the PR.

HanumathRao:
Thank you @zabetak for the review and improvements. Please go ahead and merge the PR.

PR #2937, opened 2022-10-13
[CALCITE-5332] Configuring PruneEmptyRules is cumbersome

zabetak:


rubenada:
@zabetak IMO the change makes sense, lgtm.
Have you checked if there are other rules that could benefit from the same refactoring?

zabetak:
@rubenada I guess there are other rules which miss a `DEFAULT` config instance. The problem though may not be so serious in every case if rules are not declared as 'RelOptRule` but as a subclass of `RelRule`.

Maybe it makes sense to do this refactoring for every rule to keep things uniform. WDYT?

rubenada:
@zabetak as you already mentioned in Jira, it seems PruneEmptyRule is the only one that lacks a default config, so we should definitely add it and align with the rest (see e.g. `ReduceExpressionsRule` and subclasses, which seems similar case, and which does define default configs).

rubenada:
@zabetak shall we move on and merge this PR?

zabetak:
@rubenada Thanks for the reminder. I added an additional `DEFAULT` instance for the new `ZeroMaxRowsRuleConfig` and rebased against `main`. I will merge this once all tests are run.

PR #2947, opened 2022-10-22
[MINOR] Add default catalog link to test fixtures

JiajunBernoulli:
- When I write a unit test, I often see which tables and columns are available. At present, I have to search the catalog. If there is a link to jump directly, it is very convenient.
- For new contributor, it will be easier to find the default catalog.

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/RelOptTestBase.java
+++ b/testkit/src/main/java/org/apache/calcite/test/RelOptTestBase.java
@@ -34,7 +34,10 @@ abstract class RelOptTestBase {
     return RelOptFixture.DEFAULT;
   }
 
-  /** Creates a fixture and sets its SQL statement. */
+  /**
+   * Creates a test context with a SQL query.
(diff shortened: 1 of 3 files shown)


asolimando:
I have recently had the same issue, so +1 on the change.

I noticed that the comments for the `sql()` method are all different for these three classes, maybe we can try to provide a single description, so that it's clear that the three methods are doing the same thing?

JiajunBernoulli:
Thanks for your review.@asolimando I agree with your suggestion and I use a single description:
> Creates a test context with a SQL query.

 There are two reasons to use it:
- I think it is more appropriate to stress SQL query, because `expr()` function description is 
> Creates a test context with a SQL expression. 
- **fixture** is abstract concept, **test context** is a better understanding vocabulary.
> Creates a fixture and sets its SQL statement.

asolimando:
Thanks @JiajunBernoulli, I agree on your reasoning. Can I ask you to squash the commits? Once done, I will merge it in the next 24 hours.

PR #2952, opened 2022-10-25
[CALCITE-5252] JDBC adapter sometimes miss parentheses around SELECT in WITH_ITEM body

l4wei:
adjust precedence of `SqlWithItem#unparse` to fix

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlWithItem.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlWithItem.java
@@ -97,7 +97,7 @@ public class SqlWithItem extends SqlCall {
         withItem.columnList.unparse(writer, getLeftPrec(), getRightPrec());
       }
       writer.keyword("AS");
-      withItem.query.unparse(writer, 100, 100);
+      withItem.query.unparse(writer, MDX_PRECEDENCE, MDX_PRECEDENCE);
     }
 
     @SuppressWarnings("argument.type.incompatible")
--- a/core/src/main/java/org/apache/calcite/sql/SqlWithItem.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlWithItem.java
@@ -97,7 +97,7 @@ public class SqlWithItem extends SqlCall {
         withItem.columnList.unparse(writer, getLeftPrec(), getRightPrec());
       }
       writer.keyword("AS");
-      withItem.query.unparse(writer, 10, 10);
+      withItem.query.unparse(writer, 100, 100);
     }
 
     @SuppressWarnings("argument.type.incompatible")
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/SqlWithItem.java
```diff
@@ -97,7 +97,7 @@ private static class SqlWithItemOperator extends SqlSpecialOperator {
         withItem.columnList.unparse(writer, getLeftPrec(), getRightPrec());
       }
       writer.keyword("AS");
-      withItem.query.unparse(writer, 10, 10);
+      withItem.query.unparse(writer, 100, 100);
```
libenchao, line 100: To be honest, I don't like current approach (deferring adding parentheses in `SqlCall`). Since we should always add parentheses after `AS`, how about doing this like:
```java
final SqlWriter.Frame frame = writer.startList("(", ")");
withItem.query.unparse(writer, 0, 0);
writer.endList(frame);
```

l4wei: IMHO, in essence, parentheses should be provided by subquery like `SqlSelect` or `SqlCall` to indicate the order of execution. Return to the calcite repo, I noticed current implements have followed the principle I mentioned just now, so the approach in current PR has done it in the same style. On the other hand, redundant  parentheses around `SqlSelect` will be produced in many test cases if using the approach mentioned by you, e.g. testWith2, testWithOrderBy, testWith, testSingleQuotedAlias and so on.

libenchao: Then could you document why you choose `100`?

l4wei: I think "leftPrec: 100, rightPrec: 100" is a agreement or a habit in Calcite project to indicate that it's a high enough priority. you can find the same setting in `SqlWith#unparse` before https://github.com/apache/calcite/pull/2938 be mergerd, and `SqlFunction#SqlFunction`.

l4wei: It looks like MDX_PRECEDENCE will be better, How about I change 100 to MDX_PRECEDENCE?

libenchao: Yes, `MDX_PRECEDENCE` is better than hard coded `100`.

l4wei: OK, done

libenchao (approved):
+1, merging.

PR #2957, opened 2022-10-31
[CALCITE-5355] Use the Presto SQL dialect for AWS Athena

jnturton:
I used manual testing of the Drill query 
```
select * from athena.test_table limit 10 offset 3
```
where the storage plugin `athena` uses Drill's Calcite-based JDBC storage plugin. With Calcite 1.32.0 `FETCH NEXT 10 ROWS ONLY` is generated and rejected by Athena. With Calcite built from this branch the query succeeds and the expected limit and offset are applied.

A minor aside that I'd appreciate any pointers on is that Calcite prints the following to stdout for every query made against Athena.
```
Unknown table type: EXTERNAL_TABLE
```
I've seen the same thing when querying the H2 DBMS though IIRC the table type involved was different. When Drill runs in embedded mode then stdout is the same place as the user's sqlline session so these messages are quite noisy and will be disconcerting for users. Can I do something about this? Add new table types, change the warning to go to a log rather than stdout?

Thanks
James

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
+++ b/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
@@ -49,6 +49,8 @@ import com.google.common.collect.Multimap;
 import com.google.common.collect.Ordering;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


vvysotskyi (approved):
+1

jnturton:
Thank you for the review @vvysotskyi. Please be aware that I have just amended my commit with an unrelated and minor logging fix.

PR #2960, opened 2022-11-02
[CALCITE-5155] Custom time frame

julianhyde:
This is a draft PR. I will shortly rebase, re-order and squash commits. At that point I will force-push to the branch.

To understand the intent of this change, please refer to [CALCITE-5155](https://issues.apache.org/jira/browse/CALCITE-5155). If you have comments about the specification/design, make them in that case.


julianhyde:
I just force-pushed to the branch (squash and rebase). If you're curious, the un-squashed changes are still at https://github.com/julianhyde/calcite/tree/5155-time-units.1.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -1826,12 +1828,32 @@ private static class TimestampAddConvertlet implements SqlRexConvertlet {
...
+      final TimeFrame timeFrame = cx.getValidator().validateTimeFrame(qualifier);
+      final TimeUnit unit = first(timeFrame.unit(), TimeUnit.EPOCH);
+      final RelDataType type = cx.getValidator().getValidatedNodeType(call);
+      final RexNode op1 = cx.convertExpression(call.operand(1));
+      final RexNode op2 = cx.convertExpression(call.operand(2));
+      if (unit == TimeUnit.EPOCH && qualifier.timeFrameName != null) {
+        // Custom time frames have a different path. They are kept as names, and
+        // then handled by Java functions.
+        final RexLiteral timeFrameName =
+            rexBuilder.makeLiteral(qualifier.timeFrameName);
```
tjbanghart, line 1841: Where does this handling occur? SqlFunctions.java? A reference to where these are resolved would be nice.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -1826,12 +1828,32 @@ private static class TimestampAddConvertlet implements SqlRexConvertlet {
...
+        // If the TIMESTAMPADD call has type TIMESTAMP and op2 has type DATE
+        // (which can happen for sub-day time frames such as HOUR), cast op2 to
+        // TIMESTAMP.
```
tjbanghart, line 1844: Was this a new issue discovered when implementing custom timeframes or something introduced?

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -1858,8 +1880,22 @@ private static class TimestampDiffConvertlet implements SqlRexConvertlet {
...
+      final TimeFrame timeFrame = cx.getValidator().validateTimeFrame(qualifier);
+      final TimeUnit unit = first(timeFrame.unit(), TimeUnit.EPOCH);
+
+      final RexNode op1 = cx.convertExpression(call.operand(1));
+      final RexNode op2 = cx.convertExpression(call.operand(2));
+      if (unit == TimeUnit.EPOCH && qualifier.timeFrameName != null) {
+        // Custom time frames have a different path. They are kept as names, and
+        // then handled by Java functions.
+        final RexLiteral timeFrameName =
+            rexBuilder.makeLiteral(qualifier.timeFrameName);
```
tjbanghart, line 1893: Again I am not sure where this resolution happens. Is it possible to complete the mapping here?

core/src/main/java/org/apache/calcite/rel/type/TimeFrameSet.java
```diff
@@ -0,0 +1,762 @@
...
+    public Builder addDivision(String name, Number count, String baseName) {
+      return addSub(name, true, count, baseName, TimestampString.EPOCH);
+    }
+
+    /** Defines a rollup from one frame to another.
+     *
+     * <p>An explicit rollup is not necessary for frames where one is a multiple
+     * of another (such as MILLISECOND to HOUR). Only use this method for frames
+     * that are not multiples (such as DAY to MONTH). */
+    public Builder addRollup(String fromName, String toName) {
```
tjbanghart, line 441: Would all custom timeframes require explicit rollups?

core/src/main/java/org/apache/calcite/rel/type/TimeFrameSet.java
```diff
@@ -0,0 +1,762 @@
...
+
+    @Override protected BigFraction coreMultiplier() {
+      return coreMultiplier;
+    }
+  }
+
+  /** Frame that defines is based on a minor frame and resets whenever the major
+   * frame resets. For example, "DOY" (day of year) is based on DAY and resets
+   * every YEAR. */
+  static class QuotientFrame extends TimeFrameImpl {
```
tjbanghart, line 716: MONTH is a "Core" timeframe but could it also be considered a `QuotientFrame` for YEAR?

core/src/main/java/org/apache/calcite/rel/type/TimeFrameSet.java
```diff
@@ -0,0 +1,762 @@
...
+          || numerator
+          .divide(denominator)
+          .getNumerator()
+          .abs()
+          .equals(BigInteger.ONE);
+    }
+  }
+
+  /** Core time frame (such as SECOND, MONTH, ISOYEAR). */
+  static class CoreFrame extends TimeFrameImpl {
```
tjbanghart, line 599: Are there certain timeframes that couldn't be derived from those already in `TimeFrames.CORE`? 

Should all custom timeframes be a `SubFrame` of these known `CoreFrame`s? Perhaps I misread the builder but I wonder if people could end up making garbage `CoreFrame`s which could throw others sub and quotient ones off.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2553,6 +2555,70 @@ public static Locale locale(DataContext root) {
...
+    final TimeFrameSet timeFrameSet =
+        requireNonNull(DataContext.Variable.TIME_FRAME_SET.get(root));
+    final TimeFrame timeFrame = timeFrameSet.get(timeFrameName);
+    return timeFrameSet.addDate(date, interval, timeFrame);
+  }
+
+  public static long customTimestampAdd(DataContext root,
+      String timeFrameName, long interval, long timestamp) {
+    final TimeFrameSet timeFrameSet =
+        requireNonNull(DataContext.Variable.TIME_FRAME_SET.get(root));
```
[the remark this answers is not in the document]
julianhyde: I plan to add the following javadoc, plus improve the javadoc of `class TimeFrameSet`:
```
  /** SQL {@code DATEADD} function applied to a custom time frame.
   *
   * <p>Custom time frames are created as part of a {@link TimeFrameSet}.
   * This method retrieves the session's time frame set from the
   * {@link DataContext.Variable#TIME_FRAME_SET} variable, then looks up the
   * time frame by name. */
```

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
julianhyde: I'll add " such as SqlFunctions.customTimestampAdd".

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
julianhyde: The issue doesn't arise for built-in time frames because we implement those by expanding to an arithmetic expression (multiply, add, modulo literals such as 86,400) and when we are dealing with timestamps the values are implicitly widened from `int` to `long`.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
julianhyde: It would be more complicated to complete the mapping here, because we are (for enumerable convention, anyway) converting the plan into a Java program. We can't pass objects into that program, only simple values (string, int etc.) because we want to invoke the program as you would a `public static void main` method.

The only object is the `DataContext root` parameter. Functions such as `SqlFunctions.customDateAdd` look up the time frame from `root` using the time frame name.

core/src/main/java/org/apache/calcite/rel/type/TimeFrameSet.java
julianhyde: No, we can infer that you can rollup from, say, `minute15` to `hour7` because we know that one is 15 * 60 seconds and the other is 7 * 60 * 60 seconds.

A month is not a fixed multiple of days. An explicit rollup is basically saying that the start of a month is aligned with the start of a day. And therefore you can compute a month total by adding up the day totals for all days in that month.

The same is not true for weeks in (non-ISO) years. A year does not start on the first day of a week. You cannot roll up weeks to years.

Good question, though. I'll add an explanation in the javadoc.

core/src/main/java/org/apache/calcite/rel/type/TimeFrameSet.java
julianhyde: Yes, you could define a month as 1/12th of a year. (In the same way that we define a millisecond as a 1/1000th of a second.)

I just chose to make SECOND and MONTH the fundamental units, somewhat arbitrarily.

core/src/main/java/org/apache/calcite/rel/type/TimeFrameSet.java
julianhyde: Yes, in this feature we handle time frames that are multiples of the SECOND, MONTH, ISOWEEK/ISOYEAR core, possibly with an offset, or are aligned. 

Reasonable time frames that cannot be represented are Easter (years starting with Easter, and numbering days from then); and the Roman system (months divided into 8-9 day periods, kalends, ides etc.).

If those requirements arise, we can evolve the design.

I'll add a note to the Jira case.

julianhyde: I'm sure people can define garbage frames if they try. The majority of people using this feature will be writing SQL against time frames that someone else has defined. Validation of time frames could be improved but is lower priority.

PR #2962, opened 2022-11-03
[CALCITE-5362] Geometry measurement functions

bchapuis:
This PR adds support for the following spatial measurement functions:
- ST_Area
- ST_ClosestCoordinate
- ST_ClosestPoint
- ST_FurthestCoordinate
- ST_Length
- ST_LocateAlong
- ST_LongestLine
- ST_MaxDistance
- ST_Perimeter
- ST_ProjectPoint

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SpatialTypeFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SpatialTypeFunctions.java
@@ -38,6 +38,7 @@ import org.locationtech.jts.geom.GeometryCollection;
 import org.locationtech.jts.geom.GeometryComponentFilter;
 import org.locationtech.jts.geom.GeometryFactory;
 import org.locationtech.jts.geom.GeometryFilter;
+import org.locationtech.jts.geom.LineSegment;
 import org.locationtech.jts.geom.LineString;
 import org.locationtech.jts.geom.LinearRing;
 import org.locationtech.jts.geom.MultiLineString;
@@ -49,6 +50,8 @@ import org.locationtech.jts.geom.Polygon;
 import org.locationtech.jts.geom.PrecisionModel;
 import org.locationtech.jts.geom.util.AffineTransformation;
 import org.locationtech.jts.geom.util.GeometryFixer;
+import org.locationtech.jts.linearref.LengthIndexedLine;
+import org.locationtech.jts.operation.distance.DistanceOp;
 import org.locationtech.jts.operation.linemerge.LineMerger;
 import org.locationtech.jts.operation.overlay.snap.GeometrySnapper;
 import org.locationtech.jts.operation.polygonize.Polygonizer;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/runtime/SpatialTypeFunctions.java
```diff
@@ -1318,6 +1322,195 @@ public static Geometry ST_Translate(Geometry geom, BigDecimal x, BigDecimal y) {
     return transformation.transform(geom);
   }
 
+  // Geometry measurement functions
+
+  /**
+   * Returns the area of the {@code geom}.
+   */
+  public static @Nullable Double ST_Area(Geometry geom) {
+    return geom.getArea();
```
rubenada, line 1331: Just curious: why `ST_Area` and  `ST_Length` don't have a null-check and the other methods do?

core/src/main/java/org/apache/calcite/runtime/SpatialTypeFunctions.java
```diff
@@ -1318,6 +1322,195 @@ public static Geometry ST_Translate(Geometry geom, BigDecimal x, BigDecimal y) {
...
+    for (int i = 0; i < geom.getNumGeometries(); i++) {
+      Geometry geometry = geom.getGeometryN(i);
+      if (geometry instanceof Polygon) {
+        perimeter += geometry.getLength();
+      }
+    }
+    return perimeter;
+  }
+
+  public static @Nullable Geometry ST_ProjectPoint(Geometry point, Geometry line) {
```
rubenada, line 1501: Minor: javadoc seems to be missing here.

core/src/main/java/org/apache/calcite/runtime/SpatialTypeFunctions.java
bchapuis: This is a good point, I fixed these two methods. I started adding test cases for null values after realizing that they were supported by postgis. I will open an issue to ensure that we are consistent when it come to null values in the other methods.

core/src/main/java/org/apache/calcite/runtime/SpatialTypeFunctions.java
bchapuis: Thank you, I added the javadoc.

core/src/main/java/org/apache/calcite/runtime/SpatialTypeFunctions.java
bchapuis: After reading a the javadoc for [SqlFunctions](https://javadoc.io/doc/org.apache.calcite/calcite-core/1.26.0/org/apache/calcite/runtime/SqlFunctions.html), it looks like checks for nulls are performed by Calcite before calling the functions. From what I understand, the default behavior is as follow: the functions always return null if one or more of its arguments are null. As this behavior looks similar to what postgis does, I removed all my null checks. Can you confirm that this is the right approach?

bchapuis: Thanks a lot for the clarification and the pointers.

PR #2965, opened 2022-11-10
[CALCITE-5105] Add MEASURE type and AGGREGATE aggregate function

julianhyde:


core/src/main/java/org/apache/calcite/sql/type/ApplySqlType.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.common.collect.ImmutableList;
+
+/** Type that applies generic type to type parameters. */
+abstract class ApplySqlType extends AbstractSqlType {
+  protected final ImmutableList<? extends RelDataType> types;
+
+  ApplySqlType(SqlTypeName typeName, boolean isNullable,
+      Iterable<? extends RelDataType> types) {
+    super(typeName, isNullable, null);
+    this.types = ImmutableList.copyOf(types);
```
rubenada, line 31: Would it make sense to check that `types` is not empty?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorUtil.java
```diff
@@ -1287,6 +1287,20 @@ private static FlatAggregate flattenRecurse(@Nullable SqlCall filterCall,
...
+    return getMeasure(selectItem) != null;
+  }
+
+  /** Returns the measure expression if a select item is a measure, null
+   * otherwise.
+   *
+   * <p>For a measure, {@code selectItem} will have the form
+   * {@code AS(MEASURE(exp), alias)} and this method returns {@code exp}. */
+  public static @Nullable SqlNode getMeasure(SqlNode selectItem) {
+    return null;
```
rubenada, line 1301: This method seems unfinished, will it be part of future work? Maybe add a TODO or a comment in that case?

rubenada:
@julianhyde Thanks for this contribution. Looks good. I have taken a look, and left some minor comments (I'm not very familiar with SQL parser and validator parts though, so I'm probably not the most suited reviewer in there).

site/_docs/reference.md
```diff
@@ -49,6 +49,13 @@ here to appease testAllFunctionsAreDocumented:
 | SUCCEEDS       | Documented as a period operator
 | TABLE          | Documented as part of FROM syntax
 | VARIANCE()     | In SqlStdOperatorTable, but not fully implemented
+
+Dialect-specific:
+
+| C | Function       | Reason not documented
+|:--|:-------------- |:---------------------
+| c | AGGREGATE(m)   | TODO: document; also AS MEASURE
```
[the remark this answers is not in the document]
julianhyde: Good point. I've added the code 'c' now, but not much explanation of measures at this point.

core/src/main/java/org/apache/calcite/sql/type/ApplySqlType.java
julianhyde: It's possible that 0 type arguments makes sense. So I don't think we should add the check.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorUtil.java
julianhyde: Yes, it will be extended in [CALCITE-4496](https://issues.apache.org/jira/browse/CALCITE-4496). I added a note.

PR #2967, opened 2022-11-13
[CALCITE-5377] RelFieldTrimmer support Sort with dynamic param

JiajunBernoulli:


rubenada (approved):
Thanks for the patch @JiajunBernoulli ! LGTM.

PR #2968, opened 2022-11-13
[CALCITE-5354] JDBC with UNNEST not working

JiajunBernoulli:


core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -2442,18 +2472,27 @@ private void convertUnnest(Blackboard bb, SqlCall call, @Nullable List<String> f
     });
     RelNode child =
         (null != bb.root) ? bb.root : LogicalValues.createOneRow(cluster);
+
+    RelNode newChild;
+    final CorrelationUse correlationUse = getCorrelationUse(bb, child);
+    if (correlationUse != null && !(child instanceof Collect)) {
+      newChild = addCorrelation(child, correlationUse, relBuilder);
```
libenchao, line 2479: I'm not sure this is the right place to add the correlation variable. Actually the correlation variable should be in the join node. What do you think about this?

JiajunBernoulli: Yes, I made a mistake. Origin plan should be
```
EnumerableCalc(expr#0..25=[{inputs}], proj#0..23=[{exprs}], EXPR$0=[$t25]): rowcount = 100.0, cumulative cost = {780.0 rows, 7501.0 cpu, 0.0 io}, id = 161
  EnumerableCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{24}]): rowcount = 100.0, cumulative cost = {680.0 rows, 2401.0 cpu, 0.0 io}, id = 157
    JdbcToEnumerableConverter: rowcount = 100.0, cumulative cost = {190.0 rows, 2111.0 cpu, 0.0 io}, id = 148
      JdbcProject(store_id=[$0], store_type=[$1], region_id=[$2], store_name=[$3], store_number=[$4], store_street_address=[$5], store_city=[$6], store_state=[$7], store_postal_code=[$8], store_country=[$9], store_manager=[$10], store_phone=[$11], store_fax=[$12], first_opened_date=[$13], last_remodel_date=[$14], store_sqft=[$15], grocery_sqft=[$16], frozen_sqft=[$17], meat_sqft=[$18], coffee_bar=[$19], video_store=[$20], salad_bar=[$21], prepared_food=[$22], florist=[$23], $f24=[ARRAY($0)]): rowcount = 100.0, cumulative cost = {180.0 rows, 2101.0 cpu, 0.0 io}, id = 146
        JdbcTableScan(table=[[foodmart, store]]): rowcount = 100.0, cumulative cost = {100.0 rows, 101.0 cpu, 0.0 io}, id = 1
    EnumerableUncollect: rowcount = 1.0, cumulative cost = {2.9000000000000004 rows, 2.9000000000000004 cpu, 0.0 io}, id = 155
      JdbcToEnumerableConverter: rowcount = 1.0, cumulative cost = {1.9000000000000001 rows, 1.9000000000000001 cpu, 0.0 io}, id = 153
        JdbcProject(EXPR$0=[$cor0.$f24]): rowcount = 1.0, cumulative cost = {1.8 rows, 1.8 cpu, 0.0 io}, id = 151
          JdbcValues(tuples=[[{ 0 }]]): rowcount = 1.0, cumulative cost = {1.0 rows, 1.0 cpu, 0.0 io}, id = 116
```

`JdbcToEnumerableConverter` create a new `JdbcImplementor` that extends `RelToSqlConverter`. **Root cause is the new  `RelToSqlConverter` doesn't have  correlation variable.**

In my option, we should FORCE_DECORRELATE in JdbcAdapter so that JdbcToEnumerableConverter can generate SQL independently.

libenchao: I haven't got too much time on the detail of this issue yet, but the plan seems correct to me now.

> In my option, we should FORCE_DECORRELATE in JdbcAdapter so that JdbcToEnumerableConverter can generate SQL independently.

Does `RelToSqlConverter` handle this correctly now? By saying correctly, I mean it could translate the plan to a correct SQL.

PR #2972, opened 2022-11-15
[CALCITE-2989] Use ISO calendar system when converting from java.sql.Date/Time/Timestamp to a unix timestamp

freastro:
This changes the unix timestamp to Date/Time/Timestamp conversion in `SqlFunctions` to convert between the ISO calendar system and the standard Gregorian calendar. This ensures that unix timestamps created by `SqlFunctions` will have the same value as unix timestamps created by `DateTimeUtils`.

Some of the tests may fail without this corresponding change to Avatica:
https://github.com/apache/calcite-avatica/pull/197

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexUtil.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexUtil.java
@@ -210,6 +210,20 @@ public class RexUtil {
   }
 
   /**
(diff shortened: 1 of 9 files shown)


freastro:
Will reopen targeting the CALCITE-5414 JIRA issue instead.

PR #2973, opened 2022-11-17
[CALCITE-5180] Implement some of the overloads for BigQuery's DATE and TIMESTAMP

wnob:
There is an existing implementation for a BQ `DATE` function but it wrongly thinks this is simply a typecast. See documentation [here][1].

This PR implements 2 of the 4 possible overloads for `DATE` (considering the optional time zone in the second form) as well as 1 of the possible overloads for `TIMESTAMP`, and was able to get a quidem test to pass utilizing both functions. I intend to implement the rest of the overloads in a follow-up commit but was hoping for some feedback now and thought this was a decent-sized commit to review at once.

[1]: https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date

### DIFF
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -24,7 +24,7 @@ JoinType LeftSemiJoin() :
 
 SqlNode DateFunctionCall() :
 {
-    final SqlFunctionCategory funcType = SqlFunctionCategory.USER_DEFINED_FUNCTION;
+    final SqlFunctionCategory funcType = SqlFunctionCategory.TIMEDATE;
     final SqlIdentifier qualifiedName;
     final Span s;
     final SqlLiteral quantifier;
@@ -44,7 +44,7 @@ SqlNode DateFunctionCall() :
 
 SqlNode TimestampFunctionCall() :
 {
-    final SqlFunctionCategory funcType = SqlFunctionCategory.USER_DEFINED_FUNCTION;
+    final SqlFunctionCategory funcType = SqlFunctionCategory.TIMEDATE;
     final SqlIdentifier qualifiedName;
     final Span s;
     final SqlLiteral quantifier;
@@ -64,7 +64,7 @@ SqlNode TimestampFunctionCall() :
 
 SqlNode TimeFunctionCall() :
 {
-    final SqlFunctionCategory funcType = SqlFunctionCategory.USER_DEFINED_FUNCTION;
+    final SqlFunctionCategory funcType = SqlFunctionCategory.TIMEDATE;
     final SqlIdentifier qualifiedName;
     final Span s;
     final SqlLiteral quantifier;
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -62,6 +62,26 @@ SqlNode TimestampFunctionCall() :
     }
 }
 
+SqlNode TimeFunctionCall() :
+{
+    final SqlFunctionCategory funcType = SqlFunctionCategory.USER_DEFINED_FUNCTION;
+    final SqlIdentifier qualifiedName;
+    final Span s;
+    final SqlLiteral quantifier;
+    final List<? extends SqlNode> args;
+}
+{
+    <TIME> {
+        s = span();
+        qualifiedName = new SqlIdentifier(unquotedIdentifier(), getPos());
+    }
+    args = FunctionParameterList(ExprContext.ACCEPT_SUB_QUERY) {
+        quantifier = (SqlLiteral) args.get(0);
+        args.remove(0);
+        return createCall(qualifiedName, s.end(this), funcType, quantifier, args);
+    }
+}
+
 SqlNode DateaddFunctionCall() :
 {
     final Span s;
--- a/babel/src/main/codegen/includes/parserImpls.ftl
+++ b/babel/src/main/codegen/includes/parserImpls.ftl
@@ -42,6 +42,26 @@ SqlNode DateFunctionCall() :
     }
 }
 
+SqlNode TimestampFunctionCall() :
+{
+    final SqlFunctionCategory funcType = SqlFunctionCategory.USER_DEFINED_FUNCTION;
+    final SqlIdentifier qualifiedName;
+    final Span s;
+    final SqlLiteral quantifier;
+    final List<? extends SqlNode> args;
+}
+{
+    <TIMESTAMP> {
+        s = span();
+        qualifiedName = new SqlIdentifier(unquotedIdentifier(), getPos());
+    }
+    args = FunctionParameterList(ExprContext.ACCEPT_SUB_QUERY) {
+        quantifier = (SqlLiteral) args.get(0);
+        args.remove(0);
+        return createCall(qualifiedName, s.end(this), funcType, quantifier, args);
+    }
+}
+
 SqlNode DateaddFunctionCall() :
 {
     final Span s;
--- a/babel/src/test/resources/sql/big-query.iq
+++ b/babel/src/test/resources/sql/big-query.iq
@@ -1051,29 +1051,27 @@ select unix_date(timestamp '2008-12-25') as d;
 
 !ok
 
-# DATE
-# 'date(x) is shorthand for 'cast(x as date)'
-select date('1970-01-01') as d;
+# DATE(year, month, day)
+select date(2022, 11, 15) as d;
 +------------+
 | d          |
 +------------+
-| 1970-01-01 |
+| 2022-11-15 |
 +------------+
 (1 row)
 
 !ok
 
-!if (false) {
-select date(cast(null as varchar(10))) as d;
-+---+
-| D |
-+---+
-|   |
-+---+
+# DATE(timestamp)
+select date(timestamp("2008-12-25 15:30:00+00")) as d;
++------------+
+| d          |
++------------+
+| 2008-12-25 |
++------------+
 (1 row)
 
 !ok
-!}
 
 #####################################################################
 # DATE_ADD
--- a/babel/src/test/resources/sql/big-query.iq
+++ b/babel/src/test/resources/sql/big-query.iq
@@ -1062,13 +1062,84 @@ select date(2022, 11, 15) as d;
 
 !ok
 
+# TIMESTAMP(string)
+# All these timestamps should be equal.
+# This tests the BQ timestamp literal string formatter
+# (option 'T', optional leading zeros, optional offset with proper conversion).
+select timestamp("2008-01-01 01:03:05+00") as t_space,
+       timestamp("2008-01-01T01:03:05+00") as t_iso,
+       timestamp("2008-01-01 01:03:05") as t_no_offset,
+       timestamp("2008-1-1 3:5:7+02:02:02") as t_offset;
++---------------------+---------------------+---------------------+---------------------+
+| t_space             | t_iso               | t_no_offset         | t_offset            |
++---------------------+---------------------+---------------------+---------------------+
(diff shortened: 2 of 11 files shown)


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -42,6 +42,26 @@ SqlNode DateFunctionCall() :
     }
 }
 
+SqlNode TimestampFunctionCall() :
```
olivrlee, line 45: Can you explain a bit on the whole purpose of the `/babel` directory? What I'm understanding here is that its used to be able to parse new SQL statements coming in, the `date` function existed already above this block so it doesn't need to be added

wnob: Here's my understanding: Babel is the "universal" parser ¹ that does not understand the semantics of individual dialects (hence the name). Dialect-specific semantic validation is a separate step that occurs after dialect-agnostic parsing.

This is a [FreeMarker template file][1]. It does not contain any interpolation or tags, so running this through FreeMarker would simply remove the license comment at the top of the file. It's processed by [FMPP][2] in `config.fmpp` [here][3], which is in-turn referenced in the [JavaCC][4] grammar [here][5], as well as [here][6] via the `builtinFunctionCallMethods` list where the function name was added above in `config.fmpp`.

Together, these 2 additions to the grammar indicate that there is a *production* called `TimestampFunctionCall` which consists of the [`<TIMESTAMP>` token][7] followed by a [`FunctionParameterList`][8] production, which consists of left-parenthesis, comma-separated arguments, right-parenthesis. The parser does nothing to validate how many arguments there are or what types they are.

So yes, long story short, you're right that we didn't have to modify the `DateFunctionCall` production because it's just a function call either way, and I just added this new production with a different name token but which is otherwise identical to the `DATE` one.

¹ I've heard Julian make reference to other parsers being potentially usable with Calcite but I don't know of any examples.

[1]: https://freemarker.apache.org/docs/dgui_template_overallstructure.html
[2]: https://fmpp.sourceforge.net/index.html
[3]: https://github.com/apache/calcite/blob/a505b25eacc473c6ec0ef8abd40c1ccae86297b6/babel/src/main/codegen/config.fmpp#L566
[4]: https://javacc.github.io/javacc/
[5]: https://github.com/apache/calcite/blob/a0e119ea42def418957f214f539469f1aba76c18/core/src/main/codegen/templates/Parser.jj#L1163
[6]: https://github.com/apache/calcite/blob/a0e119ea42def418957f214f539469f1aba76c18/core/src/main/codegen/templates/Parser.jj#L5992
[7]: https://github.com/apache/calcite/blob/a0e119ea42def418957f214f539469f1aba76c18/core/src/main/codegen/templates/Parser.jj#L7922
[8]: https://github.com/apache/calcite/blob/a0e119ea42def418957f214f539469f1aba76c18/core/src/main/codegen/templates/Parser.jj#L948

wnob:
I've now implemented all the overloads that do not involve `DATETIME` objects. I'm not sure what is the best way to proceed with these because there doesn't seem to be a good fit in standard SQL to act as an alias. `TIMESTAMP_WITH_LOCAL_TIME_ZONE` is not a great fit; it has an implied time zone, whereas a `DATETIME` has no time zone at all.

It's not necessary for MVP since we don't yet support `DATETIME` at all, so I'm inclined to punt on this for now.

tjbanghart:
Wouldn't `DATETIME` map to Calcite's interpretation of a `TIMESTAMP` due to the lack of TZ info?

Also, should we move this into the core parser? I don't feel strongly either way (or if it matters all that much) but the BQ style literals and some functions live in core already.

Otherwise, LGTM and a great example for adding other functions.

babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -42,6 +42,46 @@ SqlNode DateFunctionCall() :
...
+{
+    final SqlFunctionCategory funcType = SqlFunctionCategory.USER_DEFINED_FUNCTION;
```
julianhyde, line 47: not sure it should be user-defined function. a library function, maybe

babel/src/test/resources/sql/big-query.iq
```diff
@@ -1051,29 +1051,98 @@ select unix_date(timestamp '2008-12-25') as d;
...
 !ok
 
-!if (false) {
-select date(cast(null as varchar(10))) as d;
-+---+
-| D |
-+---+
-|   |
-+---+
+# TIMESTAMP(string)
```
julianhyde, line 1065: big-query.iq was created by pasting in doc from google's site. please continue the tradition.

babel/src/main/codegen/includes/parserImpls.ftl
wnob: Ya this was just copied from the existing `DateFunctionCall`. I assumed this vaguely meant library function since there is no obvious "library function" value in the `SqlFunctionCategory` enum. Perhaps these should all be in the `TIMEDATE` category. I suppose since this is the babel parser, there's no real concept of a library function.

BTW, is my assessment of babel in my above comment with Oliver correct? Should we be putting these productions in the core parser instead (or in addition to here)?

babel/src/test/resources/sql/big-query.iq
wnob: Cool, I think the most recent commit has what you asked for. The BigQuery docs aren't completely self-consistent style-wise but I copied verbatim. Also, the 3 latter tests involve a combination of different functions, so just kept the original descriptive comments there.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -149,6 +159,40 @@ public class SqlFunctions {
...
+          .appendValue(ChronoField.SECOND_OF_MINUTE, 1, 2, SignStyle.NOT_NEGATIVE)
+          // ISO 8601 supports up to nanosecond precision, but BQ only up to microsecond.
+          .optionalStart()
+          .appendFraction(ChronoField.MICRO_OF_SECOND, 0, 6, true)
+          .optionalEnd()
+          .optionalStart()
+          .parseLenient()
+          .appendOffsetId()
+          .toFormatter(Locale.ROOT);
+
```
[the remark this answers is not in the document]
wnob: Yes, it would appear that JDK8 had a different implementation for `DateTimeFormatter` that cannot handle offsets (or at least not in the same way as later versions). I'm inclined to switch to regex-based conversion, which will require more logic but be more widely compatible. Thoughts?

site/_docs/reference.md
```diff
@@ -2596,7 +2596,16 @@ semantics.
...
-| b | DATE(string)                                   | Equivalent to `CAST(string AS DATE)`
+| b | DATE(integer, integer, integer)                | Returns a date object given year, month, and day
+| b | DATE(timestamp)                                | Extracts the UTC date from a timestamp
+| b | DATE(timestamp, string)                        | Extracts the date from a timestamp, in a given time zone
+| b | TIMESTAMP(string)                              | Equivalent to `CAST(string AS TIMESTAMP)`
+| b | TIMESTAMP(string, string)                      | Equivalent to `CAST(string AS TIMESTAMP)`, converted to a given time zone
+| b | TIMESTAMP(date)                                | Convert a UTC date to a timestamp (at midnight)
+| b | TIMESTAMP(date, string)                        | Convert a date to a timestamp (at midnight), in a given time zone
+| b | TIME(integer, integer, integer)                | Returns a time object given hour, minute, and second
+| b | TIME(timestamp)                                | Extracts the UTC time from a timestamp
```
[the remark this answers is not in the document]
wnob: Reworded for clarity.

wnob:
@tjbanghart

> Wouldn't `DATETIME` map to Calcite's interpretation of a `TIMESTAMP` due to the lack of TZ info?

Neither has tz info. I included a little table in the design doc that shows similarities between types in BQ and Calcite, and BQ `TIMESTAMP` maps pretty well to Calcite's `TIMESTAMP`. There isn't really a good existing type for `DATETIME` in Calcite. I think we'll have to implement one.

julianhyde:
BQ `DATETIME` = Calcite `TIMESTAMP`

wnob:
> BQ `DATETIME` = Calcite `TIMESTAMP`

I don't see the connection. In Calcite, a timestamp is represented internally as an offset in milliseconds since epoch. In BigQuery, a timestamp is represented internally as an offset in microseconds since epoch. Apart from the difference in resolution, they're essentially the same.

Of course, this can easily be re-interpreted as a `DATETIME` by assuming a default time zone, but I struggle to see how a Calcite `TIMESTAMP` is better mapped to a BQ `DATETIME` than to a BQ `TIMESTAMP`.

julianhyde:
No, BigQuery, a timestamp is represented internally as an offset in microseconds since epoch *UTC*.

The "UTC" is a crucial difference.

Sure, they're both a bunch of bits. But so is a 64 bit IEEE floating point number. The interpretation of those bits is crucial.

wnob:
Of course, we could simply re-interpret those bits as `DATETIME`, as I mentioned. We don't need a new internal representation; we could just make a new type that happens to have the same internal representation.

UTC is not exactly relevant for a BQ timestamp. Sure, the epoch is 1970-01-01 00:00:00 UTC, but you could also think of it as 1969-01-01 16:00:00 PST. It's just a point in time.

On the other hand, a time zone would be very relevant if we decided to represent BQ `DATETIME` as a Calcite `TIMESTAMP`, since we would have to interpret Calcite's offset-from-epoch in a particular time zone, UTC being the obvious choice.

Alternatively, we could pick a new internal representation for BQ `DATETIME` -- e.g. as a Java `LocalDateTime`, which is internally represented as clock/calendar parameters with no time zone -- and avoid the implementation detail of converting from offset to clock/calendar parameters in a particular time zone. It doesn't really matter from an engineering standpoint.

Basically, we have 2 equally possible options:
1. Map BQ `TIMESTAMP` to Calcite `TIMESTAMP`, and map BQ `DATETIME` to something else (probably a whole new type that's BQ-specific, call it `BIGQUERY_DATETIME`)
2. Map BQ `DATETIME` to Calcite `TIMESTAMP`, and map BQ `TIMESTAMP` to something else (also probably a whole new type that's BQ-specific, call it `BIGQUERY_TIMESTAMP`).

I just don't see any advantages to option 1 over 2. Option 1 would cause confusion, as well as requiring a slightly awkward internal representation for `DATETIME` because it requires a time zone to convert from internal representation to semantic representation.

wnob:
I would also note that both BQ types have microsecond precision, so either way the existing Calcite `TIMESTAMP` is not a perfect match.

julianhyde:
Let's move design discussions to the Jira case.

julianhyde:
The precision issue really is orthogonal, so at best it muddies the waters. See https://issues.apache.org/jira/browse/CALCITE-5266.

mkou:
> 

I see this was last month, I just want to make sure that there is better visibility on that and that indeed we agree that 
BQ TIMESTAMP => Calcite TIMESTAMP WITH LOCAL TIMEZONW
BQ DATETIME => Calcite DATETIME

julianhyde:
@mkou, In `SqlLibraryOperators`, the `TIMESTAMP` function should return a Calcite `TIMESTAMP`, `DATE` will return `DATE`, and there is no `DATETIME` function.

Elsewhere (or in non-open source code), there can be a function called `DATETIME` that returns a Calcite `TIMESTAMP` (what BQ calls a `DATETIME`), and a function called `TIMESTAMP` that returns a Calcite `TIMESTAMP WITH LOCAL TIME ZONE` (what BQ calls a `TIMESTAMP`). The `DATETIME` function would in fact be `SqlLibraryOperators.TIMESTAMP.withName("DATETIME")`. And so forth.

Sorry it's complicated. But we need to keep the function library orthogonal to any type aliasing.

mkou:
Hey Julian, 

Yes we agree this is what I meant 

>

PR #2974, opened 2022-11-17
[CALCITE-5209] Proper sub-query handling if it is used inside select list and group by

dssysolyatin:
The problem was that SqlRelConverter tried to convert IN sub-query inside the select list to left join, but in the current case IN sub-query in the select list refers to IN sub-query inside GROUP BY.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -71,6 +71,7 @@ import org.apache.calcite.rel.logical.LogicalUnion;
 import org.apache.calcite.rel.logical.LogicalValues;
 import org.apache.calcite.rel.metadata.RelColumnMapping;
 import org.apache.calcite.rel.metadata.RelMetadataQuery;
+import org.apache.calcite.rel.rel2sql.SqlImplementor;
 import org.apache.calcite.rel.stream.Delta;
 import org.apache.calcite.rel.stream.LogicalDelta;
 import org.apache.calcite.rel.type.RelDataType;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -6374,10 +6390,13 @@ private static class SubQuery {
     final SqlNode node;
     final RelOptUtil.Logic logic;
     @Nullable RexNode expr;
+    final SqlImplementor.@Nullable Clause parentClause;
```
libenchao, line 6393: How about just name it `clause` instead of `parentClause`? ('parent' makes think that it's something like 'parent scope')

libenchao (commented):
@dssysolyatin Thanks for the PR, it looks good, I only have one small comment.

core/src/test/resources/sql/agg.iq
```diff
@@ -3467,4 +3467,38 @@ order by ename, deptno;
...
++--------+
+|      0 |
++--------+
+(1 row)
+
+!ok
+
+!set insubquerythreshold 5
+select
+    case when deptno in (1, 2, 3, 4, 5) THEN 1 else 0 end
```
libenchao, line 3490: The exception has nothing to do with `case`, I would suggest to make it simpler (at least for the commit message and Jira title)

dssysolyatin: What about "Proper sub-query handling if it is used inside select list and group by" ?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
dssysolyatin: Fixed.

core/src/test/resources/sql/agg.iq
libenchao: This looks better!

dssysolyatin: @libenchao Done. I also rebased branch.

libenchao:
LGTM~

PR #2975, opened 2022-11-18
[CALCITE-5391] JoinOnUniqueToSemiJoinRule should preserve field names, if possible

sdreynolds:
## Why:
When rewriting the query for a semi join, the
`JoinOnUniqueToSemiJoinRule` pushes a new project via `RelBuilder`. This project is *almost* a good copy of the original `Project` -- it is just missing the field names.

## How:
This change uses `RelDataType` of the `Project` to get the field names for the new `Project` pushed on the `Relbuilder` stack. This is similar to what is done in `SemiJoinRule#perform`

## Related PRs
#2848 introduced this rule and the bug.


libenchao (approved):
+1, @sdreynolds  thanks for catching and fixing this!

PR #2977, opened 2022-11-21
[CALCITE-5388] No result when using ROW_NUMBER in two common table expressions

dssysolyatin:


core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableWindow.java
```diff
@@ -693,7 +696,7 @@ private static Pair<Expression, Expression> getPartitionIterator(
     }
     Expression multiMap_ =
         builder.append(
-            "multiMap", Expressions.new_(SortedMultiMap.class));
+            "multiMap", Expressions.new_(SortedMultiMap.class), false);
```
libenchao, line 699: Current test case would success without this change, it would be good to have a test to cover this change.

libenchao (commented):
@dssysolyatin Thanks for your PR, generally I'm ok with it, I only left one minor comment.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableWindow.java
dssysolyatin: multiMap works OK at the moment even with optimize = true. But it is very unsafe. I will try to explain why. The different between multiMap and tempList cases are following.

Code for tempList case looks like:
```
final java.util.List tempList = (java.util.List) org.apache.calcite.linq4j.Linq4j.asEnumerable(new Integer[] {1,2})
                                                             .into(new java.util.ArrayList()); <-- this expression will be reused.
tempList.clear()
<----> Second EnumerableWindow will use the same tempList and will not fill tempList again
```

Code for multiMap looks like:
```
final org.apache.calcite.runtime.SortedMultiMap multiMap = new org.apache.calcite.runtime.SortedMultiMap(); <--this expression will be reused
org.apache.calcite.linq4j.Linq4j.asEnumerable(new Integer[] {1,2}).foreach(<code for multiMap filling>)
multiMap.clear()
<----> Second EnumerableWindow will refill 
```
Code works OK for multiMap only because the creation and filling of the multiMap are in different expressions. But it is very unsafe. If for some reason creating and filling multiMap are in the same expression code will also stop working.

libenchao: Thanks for your detailed explanation. I knew it's unsafe, what I propose it to have a test case to reproduce it and cover it if we can.

rubenada: I agree with @libenchao , would it be possible to have a test that fails (i.e. produces wrong result) without this patch, and gets fixed with the patch?

dssysolyatin: @libenchao @rubenada As I wrote above "multiMap" case works without this fix. 
But it works only because expressions which create and fill multiMap are different. 
I set optimize flag = true for "multiMap" only for safety, because it is really easy to make a mistake here. 
I am not against optimizations, but in my opinion, this optimization does more harm than good. 

```
Pseudo code for "multiMap" case
<-- first EnumerableWindow -->
multiMap = new SortedMultiMap() <-- luckely optimizer caches only this expression
fillMultimap(multiMap)
multiMap.clear()

<-- second EnumerableWindow -->
fillMap(multiMap)
multiMap.clear()
```

rubenada: @dssysolyatin thanks for the clarification. Sorry if I insist, but in the Jira you talk about a scenario that returns empty result instead of the expected rows; wouldn't it be possible to reproduce that in a unit test?

dssysolyatin: @rubenada For case when tempList is used, the query returns empty result without this fix. 
The following test case covers it:
```
with
    CTE1(rownr1, val1) as ( select ROW_NUMBER() OVER(ORDER BY id ASC), id from (values (1), (2)) as Vals1(id) ),
    CTE2(rownr2, val2) as ( select ROW_NUMBER() OVER(ORDER BY id ASC), id from (values (1), (2)) as Vals2(id) )
select
    CTE1.rownr1, CTE1.val1, CTE2.rownr2, CTE2.val2
from
    CTE1,CTE2
where
    CTE1.val1 = CTE2.val2;

+--------+------+--------+------+
| ROWNR1 | VAL1 | ROWNR2 | VAL2 |
+--------+------+--------+------+
|      1 |    1 |      1 |    1 |
|      2 |    2 |      2 |    2 |
+--------+------+--------+------+
(2 rows)
```

libenchao: With the newly added test, I would approve this PR. (Although we are fixing a non-exist problem).  
How about we add a comment in the commit message about this, to avoid leaving a false impression that maintainers will assume that there is a problem currently and this change fix that？

dssysolyatin: Yes, the problem does not exists for "multiMap" at the moment. But this change avoids possible problem. Although I started to think, that I should revert changes for "multiMap" and create separate ticket for improving optimizer.

It is quite bad idea to reuse mutable collections at all:
https://github.com/apache/calcite/commit/a7bedd6595ebe69bd93e9fcc3f05850da5f673b2

dssysolyatin: @libenchao I reverted my changes for "multiMap" and did rebase

libenchao (approved):
+1, LGTM

PR #2980, opened 2022-11-22
[CALCITE-5389] Add STARTS_WITH() and ENDS_WITH() for BIG_QUERY

olivrlee:
Adding `STARTS_WITH()` and `ENDS_WITH()` for `BIG_QUERY`


babel/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -178,6 +199,7 @@ SqlCreate SqlCreateTable(Span s, boolean replace) :
 |   < DATEPART: "DATEPART" >
 |   < NEGATE: "!" >
 |   < TILDE: "~" >
+|   < STARTS_WITH: "STARTS_WITH" >
```
wnob, line 202: nit: I believe this should be alphabetically sorted

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -206,6 +206,14 @@ private SqlLibraryOperators() {
               .andThen(SqlTypeTransforms.TO_VARYING), null,
           OperandTypes.STRING, SqlFunctionCategory.STRING);
 
+  /** BigQuery's "SUBSTR(value1, value2)" function. */
```
wnob, line 209: typo?

babel/src/test/resources/sql/big-query.iq
```diff
@@ -2519,4 +2519,31 @@ SELECT PARSE_TIMESTAMP("%c", "Thu Dec 25 07:30:00 2008") AS parsed;
...
 
+#####################################################################
+#
+# STARTS_WITH(value1, value2)
+#
+# Takes two STRING or BYTES values. Returns TRUE if the second value is a prefix of the first.
+#
+# This function supports specifying collation.
+WITH items AS
+  (SELECT 'foo' as item
```
wnob, line 2530: These are all strings, but your function accepts `OperandTypes.STRING_STRING` which includes bytes objects as well. We should test those too (I think you need another overload in `SqlFunctions`).

I also think you may need to specify that both arguments must have the same type, and I'm not sure if `OperandTypes.STRING_STRING` accomplishes that. There should be a test to make sure you can't call `STARTS_WITH` passing a string and a bytes together. Double check that this is consistent with BigQuery by running a sample query in the BQ console, since their documentation isn't 100% clear.

olivrlee: Update: BQ Console will throw an error if attempting to use string+bytes, so I've updated this PR to use `STRING_SAME_SAME`

PR #2981, opened 2022-11-23
[CALCITE-5283] Add ARG_MIN, ARG_MAX aggregate functions

zoudan:
Add ARG_MIN, ARG_MAX (aka MIN_BY, MAX_BY) aggregate functions.
- **ARG_MAX(value, comp)**   Returns *value* for the maximum value of *comp* in the group
- **ARG_MIN(value, comp)**    Returns *value* for the minimum value of *comp* in the group

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -1974,6 +1974,17 @@ public class SqlFunctions {
     return b0 == null || b1 != null && b0.compareTo(b1) < 0 ? b1 : b0;
   }
 
+  /** less than */
+  public static <T extends Comparable<T>> boolean lessThan(T b0, T b1) {
+    return b1 == null || b0 != null && b0.compareTo(b1) < 0;
+  }
+
+  /** grater than */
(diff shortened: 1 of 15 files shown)


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1849,6 +1849,17 @@ public static <T extends Comparable<T>> T greatest(T b0, T b1) {
...
   }
 
+  /** Less than. */
+  public static <T extends Comparable<T>> boolean lessThan(T b0, T b1) {
+    return b1 == null || b0 != null && b0.compareTo(b1) < 0;
+  }
+
+  /** Grater than. */
+  public static <T extends Comparable<T>> boolean greaterThan(T b0, T b1) {
+    return b1 == null || b0 != null && b0.compareTo(b1) > 0;
```
chunweilei, line 1859: What's the difference between `greatest` and `greaterThan`?

zoudan: `greaterThan` return wether b0 is greater than b1 while `greatest` return the larger one between b0 and b1

PR #2983, opened 2022-11-26
[CALCITE-5395] RelToSql converter fails when SELECT * is under a semi…

JiajunBernoulli:
…-join node


rubenada:
@JiajunBernoulli the fix LGTM, but I am not very familiar with SqlToRelConverter's logic.
Let's wait to see if some other contributor that knows this part of the code better can take a look. Otherwise I'll take the responsibility and merge myself.

PR #2988, opened 2022-11-28
[CALCITE-5399] Remove Proj4J from the api dependencies

bchapuis:
Proj4J uses the EPSG dataset and has restricting terms of use. Making Proj4J a compileOnly dependency addresses this issue.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -48,7 +48,11 @@ dependencies {
 
     api("org.locationtech.jts:jts-core")
     api("org.locationtech.jts.io:jts-io-common")
-    api("org.locationtech.proj4j:proj4j")
+
+    // Due to restricting terms of use, we cannot include Proj4J as an API dependency.
+    compileOnly("org.locationtech.proj4j:proj4j")
+    testRuntimeOnly("org.locationtech.proj4j:proj4j")
+
     api("com.fasterxml.jackson.core:jackson-annotations")
     api("com.google.errorprone:error_prone_annotations")
     api("com.google.guava:guava")
--- a/core/src/main/java/org/apache/calcite/runtime/ProjectionTransformer.java
+++ b/core/src/main/java/org/apache/calcite/runtime/ProjectionTransformer.java
@@ -40,6 +40,14 @@ import java.util.stream.Stream;
 
 /**
  * Transforms the projection of a geometry.
+ *
+ * This class uses Proj4J to transform the projection of a geometry
+ * and should not be used beyond the scope of the Spatial Type Extensions.
+ * Proj4J is released under the Apache License 2.0, however, it also uses the EPSG dataset,
+ * which has restricting <a href="https://epsg.org/terms-of-use.html">terms of use</a>.
+ * As a result, Proj4J is not suitable for inclusion in Apache Calcite
+ * and this class will throw {@code ClassNotFoundException}s
+ * if Proj4J is not added to the classpath by the user.
  */
 public class ProjectionTransformer extends GeometryTransformer {
 
--- a/site/_docs/howto.md
+++ b/site/_docs/howto.md
@@ -841,6 +841,8 @@ your key to the keyservers used by Nexus, see above.
   that the `META-INF` directory contains `LICENSE`,
   `NOTICE`
 * Check PGP, per [this](https://httpd.apache.org/dev/verification.html)
+* Check that Proj4J is not among the api and implementation dependencies,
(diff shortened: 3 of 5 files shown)


core/src/main/java/org/apache/calcite/runtime/ProjectionTransformer.java
```diff
@@ -40,6 +40,14 @@
...
 /**
  * Transforms the projection of a geometry.
+ *
+ * This class uses Proj4J to transform the projection of a geometry
+ * and should not be used beyond the scope of the Spatial Type Extensions.
+ * Proj4J is released under the Apache License 2.0, however, it also uses the EPSG dataset,
+ * which has restricting <a href="https://epsg.org/terms-of-use.html">terms of use</a>.
+ * As a result, Proj4J is not suitable for inclusion in Apache Calcite
+ * and this class will throw {@code ClassNotFoundException}s
+ * if Proj4J is not added to the classpath by the user.
```
rubenada, line 50: Would we need to add a similar warning in the web documentation (`reference.md`) in the spatial section?

core/build.gradle.kts
```diff
@@ -57,6 +54,11 @@ dependencies {
     api("org.checkerframework:checker-qual")
     api("org.slf4j:slf4j-api")
 
+    api("org.locationtech.jts:jts-core")
+    api("org.locationtech.jts.io:jts-io-common")
+    compileOnly("org.locationtech.proj4j:proj4j")
+    testImplementation("org.locationtech.proj4j:proj4j")
```
zabetak, line 60: `testImplementation` means that we need it in the compile and runtime classpath for tests; is this true? If we can use `testRuntimeOnly` I think it would be better.

zabetak: Moreover, consider moving it closer to other dependencies in the same group.

site/_docs/howto.md
```diff
@@ -841,6 +841,8 @@ your key to the keyservers used by Nexus, see above.
   that the `META-INF` directory contains `LICENSE`,
   `NOTICE`
 * Check PGP, per [this](https://httpd.apache.org/dev/verification.html)
+* Check that Proj4J is not among the api and implementation dependencies,
```
zabetak, line 844: We could enforce this with a new CI job as well. Not necessary to do it now but maybe worth considering as a follow-up.

zabetak (commented):
Maybe you did this already but if not it would be nice to create a very simple maven project that depends on `calcite-core` and uses the Spatial extension to verify that indeed you get the desired `ClassNotFoundException`.

Also worth confirming the output of `mvn dependency:tree` on that sample project as well.

In case you are not aware, you can use `./gradlew publishToMavenLocal` to test things locally.

core/src/main/java/org/apache/calcite/runtime/ProjectionTransformer.java
bchapuis: This is a good idea, I added notices in `reference.md` and `spatial.md`.

site/_docs/howto.md
bchapuis: Yes, this is a good idea. Would adding something like `if ./gradlew dependencies | grep -q proj4j; then exit 1; else exit 0; fi` in a workflow be sufficient?

julianhyde: I think proj4j are going to fix the licensing issue on their end. I think a CI job would be overkill given that this fix will be short-lived.

I suggest adding a 'Add Proj4j as a dependency' Jira case, to be fixed in 1.34, and revisit then.

julianhyde: I've made some further remarks in the jira case, and +1 when those are fixed.

zabetak (commented):
LGTM, many thanks for pushing this forward @bchapuis !

zabetak (approved):

rubenada (approved):

bchapuis:
@zabetak Thanks a lot for your help. 

I just reread the [develop documentation](https://calcite.apache.org/develop/) in prevision of merging the PR. To keep the history clean, I rebased the branch, squashed the commits, and force pushed the result. As a new (and intimidated ;) committer I should now be able to click "rebase and merge" to close this PR. Is that it? Or am I missing something?

zabetak:
@bchapuis Thanks for pushing this forward! You are correct; either you click "Rebase and merge" button or you do the same thing via command line. Sharing also another link that you may find useful in the future: https://calcite.apache.org/docs/howto.html#merging-pull-requests

PR #2989, opened 2022-11-28
[CALCITE-5280] Add geometry aggregate functions (ST_Accum, ST_Collect, and ST_Union)

bchapuis:
Here, the main point of caution is the creation of the operator table in `SqlLibraryOperatorTableFactory`.

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SpatialTypeFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SpatialTypeFunctions.java
@@ -52,6 +52,7 @@ import org.locationtech.jts.geom.util.GeometryFixer;
 import org.locationtech.jts.operation.linemerge.LineMerger;
 import org.locationtech.jts.operation.overlay.snap.GeometrySnapper;
 import org.locationtech.jts.operation.polygonize.Polygonizer;
+import org.locationtech.jts.operation.union.UnaryUnionOp;
 import org.locationtech.jts.precision.GeometryPrecisionReducer;
 import org.locationtech.jts.simplify.DouglasPeuckerSimplifier;
 import org.locationtech.jts.simplify.TopologyPreservingSimplifier;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperatorTableFactory.java
```diff
@@ -16,6 +16,15 @@
...
 package org.apache.calcite.sql.fun;
 
+import org.apache.calcite.config.CalciteConnectionConfigImpl;
+import org.apache.calcite.jdbc.CalciteSchema;
+import org.apache.calcite.jdbc.JavaTypeFactoryImpl;
+import org.apache.calcite.model.ModelHandler;
+import org.apache.calcite.prepare.CalciteCatalogReader;
+import org.apache.calcite.runtime.SpatialTypeFunctions;
+import org.apache.calcite.schema.Function;
+import org.apache.calcite.schema.SchemaPlus;
```
julianhyde, line 26: The large number of new imports is an indication that the code to define the aggregate functions should be somewhere else.

bchapuis: I aggree, this code should have been placed in `SqlOPeratorTables.createSpatial`. However, as it does not address the number of imports, I looked at the `SqlStdOperatorTable` and wondered if we should create a `SqlSpatialOperatorTable`. I will probably dig into this a bit, but please stop me if it is a bad idea.

bchapuis: @julianhyde I moved this part of the code to a new class called [SqlSpatialTypeOperatorTable](https://github.com/apache/calcite/pull/2989/files#diff-0a0f94412167ef22f588b72b5452b6e8c693cf7b22239690b71b2433189fabe4), which implements the `SqlOperatorTable` interface. I also experimented with the `ReflectiveSqlOperatorTable`, but I believe that using this approach would require declaring properties of type `SqlUserDefinedAggFunction` for each `AggregateFunction`. What do you think about continuing down this path?

PR #2993, opened 2022-12-01
[CALCITE-5410] Assertion error on PERCENT_REMAINDER operator with DECIMAL type

alex-plekhanov:
PERCENT_REMAINDER operator should have the same return type inference as MOD operator.


libenchao (approved):
+1, good catch, @alex-plekhanov  thanks for your contribution!

PR #2994, opened 2022-12-01
[CALCITE-5411] Updated Spark dialect to reflect support of ROLLUP and CUBE standard syntax

LeonidChistov:
https://issues.apache.org/jira/browse/CALCITE-5411

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -5972,7 +5972,13 @@ class RelToSqlConverterTest {
     final String expected = "SELECT COUNT(*)\n"
         + "FROM \"foodmart\".\"product\"\n"
         + "GROUP BY CUBE(\"product_id\", \"product_class_id\")";
-    sql(query).ok(expected);
+    final String expectedSpark = "SELECT COUNT(*)\n"
+        + "FROM foodmart.product\n"
+        + "GROUP BY CUBE(product_id, product_class_id)";
+    sql(query)
+        .ok(expected)
+        .withPresto().ok(expected)
+        .withSpark().ok(expectedSpark);
   }
 
   @Test void testRollupWithGroupBy() {
@@ -5982,7 +5988,13 @@ class RelToSqlConverterTest {
     final String expected = "SELECT COUNT(*)\n"
         + "FROM \"foodmart\".\"product\"\n"
         + "GROUP BY ROLLUP(\"product_id\", \"product_class_id\")";
-    sql(query).ok(expected);
+    final String expectedSpark = "SELECT COUNT(*)\n"
+        + "FROM foodmart.product\n"
+        + "GROUP BY ROLLUP(product_id, product_class_id)";
+    sql(query)
+        .ok(expected)
+        .withPresto().ok(expected)
+        .withSpark().ok(expectedSpark);
   }
 
   @Test void testJsonType() {
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -5972,16 +5972,7 @@ class RelToSqlConverterTest {
     final String expected = "SELECT COUNT(*)\n"
         + "FROM \"foodmart\".\"product\"\n"
         + "GROUP BY CUBE(\"product_id\", \"product_class_id\")";
-    final String expectedInSpark = "SELECT COUNT(*)\n"
-        + "FROM foodmart.product\n"
-        + "GROUP BY product_id, product_class_id WITH CUBE";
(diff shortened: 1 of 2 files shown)


devozerov (approved):
LGTM

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -5972,16 +5972,7 @@ private void checkLiteral2(String expression, String expected) {
...
         + "GROUP BY CUBE(\"product_id\", \"product_class_id\")";
-    final String expectedInSpark = "SELECT COUNT(*)\n"
-        + "FROM foodmart.product\n"
-        + "GROUP BY product_id, product_class_id WITH CUBE";
-    final String expectedPresto = "SELECT COUNT(*)\n"
-        + "FROM \"foodmart\".\"product\"\n"
-        + "GROUP BY CUBE(\"product_id\", \"product_class_id\")";
-    sql(query)
-        .ok(expected)
-        .withPresto().ok(expectedPresto)
```
libenchao, line 5983: Why need you remove the test for presto?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -5972,16 +5972,7 @@ private void checkLiteral2(String expression, String expected) {
     final String expected = "SELECT COUNT(*)\n"
         + "FROM \"foodmart\".\"product\"\n"
         + "GROUP BY CUBE(\"product_id\", \"product_class_id\")";
-    final String expectedInSpark = "SELECT COUNT(*)\n"
-        + "FROM foodmart.product\n"
-        + "GROUP BY product_id, product_class_id WITH CUBE";
```
libenchao, line 5977: We have already support `WITH CUBE` and `WITH ROLLUP` in CALCITE-3077, will the test still pass after your change?

libenchao (commented):
Spark seems supporting both `GROUP BY CUBE (x, y)` and `GROUP BY x, y WITH CUBE`[1] syntax, and we have already support the latter in CALCITE-3077, is it necessary to change it to the first one?

[1] https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-groupby.html

LeonidChistov:
@libenchao 

First reason is that syntax "WITH ROLLUP" is non standard, while "GROUP BY ROLLUP(...)" syntax is standard.

Second reason is following:

- We have a [hasTrickyRollup](https://github.com/apache/calcite/blob/01002b4a16ddc2bdb7aa59a5f5f22649111d2ea9/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java#L984) check that has a meaning of something like "dialect supports WITH ROLLUP, but not GROUP BY ROLLUP".
-  In our code for `Sort` implementation we assume that we are not allowed to have "WITH ROLLUP" clause followed by "ORDER" by clause. That is actually not true for the Spark dialect, but true for MySQL5. And that constraint can make generated SQL code for Spark more complicated that is actually required (two SQL SELECT clause will be needed).
- Now we have a solution: either to fix dialect attributes of Spark dialect so that we always generate "GROUP BY ROLLUP(..)" or we need to implement more fine-grained distinction between "WITH ROLLUP supporting ORDER BY" and "WITH ROLLUP not supporing ORDER BY"
- IMO, first solution looks less intrusive for current code base

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
LeonidChistov: I think I misinterpreted how the test checks works, will revert this change, thanks.

libenchao:
> @libenchao
> 
> First reason is that syntax "WITH ROLLUP" is non standard, while "GROUP BY ROLLUP(...)" syntax is standard.
> 
> Second reason is following:
> 
> * We have a [hasTrickyRollup](https://github.com/apache/calcite/blob/01002b4a16ddc2bdb7aa59a5f5f22649111d2ea9/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java#L984) check that has a meaning of something like "dialect supports WITH ROLLUP, but not GROUP BY ROLLUP".
> * In our code for `Sort` implementation we assume that we are not allowed to have "WITH ROLLUP" clause followed by "ORDER" by clause. That is actually not true for the Spark dialect, but true for MySQL5. And that constraint can make generated SQL code for Spark more complicated that is actually required (two SQL SELECT clause will be needed).
> * Now we have a solution: either to fix dialect attributes of Spark dialect so that we always generate "GROUP BY ROLLUP(..)" or we need to implement more fine-grained distinction between "WITH ROLLUP supporting ORDER BY" and "WITH ROLLUP not supporing ORDER BY"
> * IMO, first solution looks less intrusive for current code base

Sounds good to me, standard syntax is better since Spark supports it.

libenchao (approved):
+1, merging!

PR #2998, opened 2022-12-05
[CALCITE-5360] Implement BigQuery TIMESTAMP_ADD

tanclary:
Implement BigQuery's TIMESTAMP_ADD. This implementation is similar to that of the standard timestampadd but accounts for the difference in operand number and type. JIRA issue https://issues.apache.org/jira/browse/CALCITE-5360 includes TIMESTAMP_DIFF as well but this PR is being opened primarily for feedback to avoid potential duplicate mistakes as TIMESTAMP_DIFF should follow a similar implementation.

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -7913,7 +7913,50 @@ public class SqlOperatorTest {
         "2015-01-01 00:00:00", "TIMESTAMP(0) NOT NULL");
   }
 
-  @Test void testDenseRankFunc() {
+  @Test void testBigQueryTimestampAdd() {
+    final SqlOperatorFixture nonBigQuery = fixture()
+        .setFor(SqlLibraryOperators.TIMESTAMP_ADD_BIG_QUERY);
+    nonBigQuery.checkFails("^timestamp_add(timestamp '2008-12-25 15:30:00', "
+            + "interval 5 minute)^",
+        "No match found for function signature "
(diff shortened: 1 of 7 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7913,7 +7913,50 @@ private static void checkArrayConcatAggFuncFails(SqlOperatorFixture t) {
...
+        .setFor(SqlLibraryOperators.TIMESTAMP_ADD_BIG_QUERY);
+    nonBigQuery.checkFails("^timestamp_add(timestamp '2008-12-25 15:30:00', "
+            + "interval 5 minute)^",
+        "No match found for function signature "
+            + "TIMESTAMP_ADD\\(<TIMESTAMP>, <INTERVAL_DAY_TIME>\\)", false);
+
+    final SqlOperatorFixture f = fixture()
+        .withLibrary(SqlLibrary.BIG_QUERY)
+        .setFor(SqlLibraryOperators.TIMESTAMP_ADD_BIG_QUERY);
+    f.checkScalar("timestamp_add(timestamp '2008-12-25 15:30:00', "
```
julianhyde, line 7927: I don't think Calcite supports NANOSECOND intervals right now (maybe not MICROSECOND either). You should disable this statement with an `if` and log a Calcite bug to track.

julianhyde (commented):
A good start. I suggested some improvements in comments.

I edited the code to check whether my suggestions made sense, so feel free to build on my changes in https://github.com/julianhyde/calcite/tree/5360-timestamp-add.

PR #2999, opened 2022-12-09
[CALCITE-5428] Drop minimum Guava version to 16.0.1.

gianm:
See https://issues.apache.org/jira/browse/CALCITE-5428.


gianm:
Hmm. The "Checks" tab says "This job was skipped" for "Linux (JDK 8), Oldest Guava, America/New_York Timezone". I think we need that to run to verify that this change works properly. Does anyone know why that doesn't run, or how to get it to run?

PR #3002, opened 2022-12-12
[CALCITE-5417] Upgrade Proj4J from 1.1.5 to 1.2.2

bchapuis:
This PR is intended at testing Proj4J version 1.2.2.


site/_docs/reference.md
```diff
@@ -2438,6 +2438,9 @@ Not implemented:
 
 #### Geometry projection functions
 
+The EPSG dataset is released separately from Proj4J due to its restrictive [terms of use](https://epsg.org/terms-of-use.html).
+In order to use the projection functions in Apache Calcite, users must include the EPSG dataset in their dependencies.
```
julianhyde, line 2442: Can you shorten the lines? In my opinion there's little reason not to wrap text at 60 or 70 characters. Long lines are difficult to view in diff output.

bom/build.gradle.kts
```diff
@@ -58,6 +58,7 @@ dependencies {
         apiv("org.locationtech.jts:jts-core")
         apiv("org.locationtech.jts.io:jts-io-common")
         apiv("org.locationtech.proj4j:proj4j")
+        apiv("org.locationtech.proj4j:proj4j-epsg")
```
julianhyde, line 61: If proj4j and proj4j-epsg are likely to have the same version number there's no need for this line, or the line in gradle.properties.

core/build.gradle.kts
```diff
@@ -98,6 +98,7 @@ dependencies {
     testImplementation(kotlin("stdlib-jdk8"))
     testImplementation(kotlin("test"))
     testImplementation(kotlin("test-junit5"))
+    testRuntimeOnly("org.locationtech.proj4j:proj4j-epsg")
```
julianhyde, line 101: add a comment that proj4j-epsg must not be converted 'implementation' due to its license

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
```diff
@@ -66,6 +66,9 @@ public interface CalciteResource {
   @BaseMessage("Geo-spatial extensions and the GEOMETRY data type are not enabled")
   ExInst<SqlValidatorException> geometryDisabled();
 
+  @BaseMessage("Proj4J EPSG is missing from the classpath")
+  ExInst<CalciteException> proj4jEpsgIsMissing();
```
julianhyde, line 70: I would add '; to resolve this problem, download the EPSG data set and agree to its terms of use'

PR #3003, opened 2022-12-13
5430/Add IFNULL() function for BIG_QUERY dialect

olivrlee:


core/src/main/java/org/apache/calcite/sql/fun/SqlIfNullFunction.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.calcite.sql.type.OperandTypes;
+import org.apache.calcite.sql.type.ReturnTypes;
+import org.apache.calcite.sql.type.SqlTypeTransforms;
+import org.apache.calcite.sql.validate.SqlValidator;
+import org.apache.calcite.util.Util;
+
+import java.util.List;
+
+/**
+ * The <code>IFULL</code> function.
```
wnob, line 35: nit: maybe mention that this is for BigQuery.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6315,6 +6315,30 @@ void assertSubFunReturns(boolean binary, String s, int start,
...
   }
 
+
+  @Test void testIfNullCoalesceFunc() {
+    final SqlOperatorFixture f = fixture()
+        .withLibrary(SqlLibrary.BIG_QUERY)
+        .setFor(SqlLibraryOperators.IFNULL, VM_EXPAND);
+
+    f.checkString("ifnull('a','b')", "a", "CHAR(1) NOT NULL");
+    f.checkString("ifnull(null,'b')", "b", "CHAR(1)");
```
wnob, line 6325: nit: I'm a bit surprised that this isn't `NOT NULL`, since the javadoc for `LEAST_NULLABLE` says:

"the result is NOT NULL if any of the arguments is NOT NULL; otherwise the type is unchanged"

and it would seem the type of `'b'` should be `NOT NULL`. Any ideas as to why? Not sure if this is really a problem or not.

core/src/main/java/org/apache/calcite/sql/fun/SqlIfNullFunction.java
```diff
@@ -0,0 +1,73 @@
...
+        ReturnTypes.LEAST_RESTRICTIVE.andThen(SqlTypeTransforms.LEAST_NULLABLE),
+        null,
+        OperandTypes.SAME_SAME,
+        SqlFunctionCategory.SYSTEM);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  // override SqlOperator
+  @Override public SqlNode rewriteCall(SqlValidator validator, SqlCall call) {
```
wnob, line 52: Oh cool, I didn't know about this. Super clean.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
olivrlee: Updated to follow NVL pattern and the result is CHAR(1) NOT NULL, so fixed!

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -190,7 +190,10 @@ private SqlLibraryOperators() {
 
   /** The "IFNULL(value, value)" function. */
   @LibraryOperator(libraries = {BIG_QUERY})
-  public static final SqlFunction IFNULL = new SqlIfNullFunction();
+  public static final SqlFunction IFNULL = SqlBasicFunction.create("IFNULL",
```
julianhyde, line 193: better, `NVL.withName("IFNULL")` (you'll need to change the type of `NVL`)

core/src/main/java/org/apache/calcite/sql/fun/SqlIfNullFunction.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.calcite.sql.type.SqlTypeTransforms;
+import org.apache.calcite.sql.validate.SqlValidator;
+import org.apache.calcite.util.Util;
+
+import java.util.List;
+
+/**
+ * The <code>IFULL</code> function.
+ */
+public class SqlIfNullFunction extends SqlFunction {
```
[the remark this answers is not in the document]
olivrlee: I deleted this file in my last commit, no we don't need it anymore

julianhyde:
Merged as e426bd26. (The PR link in that commit is incorrect; it should point to this PR, 3003.)

PR #3005, opened 2022-12-15
[CALCITE-5435] Insert multi row VALUES failed when column has values of different types

alex-plekhanov:


core/src/test/java/org/apache/calcite/test/TypeCoercionConverterTest.java
```diff
@@ -173,6 +173,20 @@ public static void checkActualAndReferenceFiles() {
     sql(sql).ok();
   }
 
+  @Test void testInsertValuesDifferentColumnTypes() {
```
zoudan, line 176: `testInsertValuesDifferentColumnTypes` may not be a perfect name here, how about `testInsertValuesWithConvertibleTypes` ?

alex-plekhanov: Thanks, fixed

PR #3006, opened 2022-12-17
[CALCITE-5439] Validation of Pivot fails after creating a deep copy of SqlNode

vvysotskyi:


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -6653,13 +6654,44 @@ private void checkLiteral2(String expression, String expected) {
         .withBigQuery().ok(expected);
   }
 
+  @Test void testPivotWithSqlShuttle() {
```
libenchao, line 6657: I'm wondering if this is the best place for this. Another place would be `SqlParserTest`, there is a similar issue CALCITE-4406 which you can take a look.

libenchao (commented):
@vvysotskyi Thanks for your PR, I've left one comment.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
vvysotskyi: Agree, thanks, I moved the test there.

core/src/main/java/org/apache/calcite/sql/SqlPivot.java
```diff
@@ -195,5 +197,17 @@ static class Operator extends SqlSpecialOperator {
     Operator(SqlKind kind) {
       super(kind.name(), kind);
     }
+
+    @Override public SqlCall createCall(
+        @Nullable SqlLiteral functionQualifier,
+        SqlParserPos pos,
+        @Nullable SqlNode... operands) {
+      assert operands.length == 4;
+      pos = pos.plusAll(operands);
```
libenchao, line 206: Why do you add `operands` position to `pos`? I don't see other `SqlOperator#createCall`s are doing this.

libenchao (commented):
@vvysotskyi Thanks for the updating, generally looks good to me now, I only have one minor comment.

core/src/main/java/org/apache/calcite/sql/SqlPivot.java
vvysotskyi: `SqlOperator#createCall` does that. It seemed useful to pass operands for the parser position.

libenchao: I debugged your test case, and found an interesting result. The `pos` for original `SqlPivot` is from 'line 3 column 1, to line 3 column 54', which represents the whole `PIVOT (...)`. However, after copying using `createCall`, it changes to 'line 2 column 7, to line 3 column 54'. The reason for this is the first operand is the `query` which does not count into the original pos.
Hence, for this case, I think we should not add this line, since it gives the wrong result. WDYT?

vvysotskyi: Thanks for investigating it. In this case, it makes sense to remove this line, I made this change and squashed it with the existing commit.

libenchao (approved):
LGTM~

PR #3008, opened 2022-12-20
[CALCITE-5436] Implement DATE_SUB, TIME_SUB, TIMESTAMP_SUB (compatible w/ BigQuery)

tanclary:
Add support for BigQuery's DATE_SUB, TIME_SUB, and TIMESTAMP_SUB functions. Create MINUS_DATE2 operator to handle subtracting an interval from a timestamp, time, or date expression. This differs from the standard MINUS_DATE operator which takes 3 arguments which is designed to subtract one time expression from another. Add week and quarter as valid time units for intervals in the parser.


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -660,6 +669,24 @@ private SqlLibraryOperators() {
           OperandTypes.TIMESTAMP_INTERVAL)
           .withFunctionType(SqlFunctionCategory.TIMEDATE);
 
+  /** The "TIME_SUB(time_expression, INTERVAL int64_expression date_part)"
+   * function (BigQuery) subtracts int64_expression units of date_part from
```
julianhyde, line 673: s/INTERVAL int64_expression date_part/interval_expression/

In Calcite the INTERVAL keyword isn't specific to the TIME_SUB function, it's part of the interval expression.

Also Calcite doesn't have an int64 type. (Yeah I know BQ does but we're using Calcite terminology here.)

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -1893,6 +1899,39 @@ private static class TimestampAddConvertlet implements SqlRexConvertlet {
...
+      // TIMESTAMP_SUB(timestamp, interval)
+      //  => timestamp - count * INTERVAL '1' UNIT
+      final RexBuilder rexBuilder = cx.getRexBuilder();
+      final SqlBasicCall operandCall = call.operand(1);
+      SqlIntervalQualifier qualifier = operandCall.operand(1);
+      final RexNode op1 = cx.convertExpression(operandCall.operand(0));
+      final RexNode op2 = cx.convertExpression(call.operand(0));
+      final TimeFrame timeFrame = cx.getValidator().validateTimeFrame(qualifier);
+      final TimeUnit unit = timeFrame.unit();
+      RexNode interval2Sub;
```
julianhyde, line 1914: `interval2Sub` should be `final`.

Fix the checkerframework errors.

Add comment why special handling for NANO, MICRO

site/_docs/reference.md
```diff
@@ -2615,6 +2619,7 @@ semantics.
 | p q | DATEADD(timeUnit, integer, datetime)         | Equivalent to `TIMESTAMPADD(timeUnit, integer, datetime)`
 | p q | DATEDIFF(timeUnit, datetime, datetime2)      | Equivalent to `TIMESTAMPDIFF(timeUnit, datetime, datetime2)`
 | q | DATEPART(timeUnit, datetime)                   | Equivalent to `EXTRACT(timeUnit FROM  datetime)`
+| b | DATE_SUB(date, interval int64 date_part)       | Subtracts int64_expression units of date_part from the date, independent of any time zone.
```
julianhyde, line 2622: not `int64_expression units of date_part`; it's just an interval expression

remove '.' from end of line

DATE_SUB, TIME_SUB, TIMESTAMP_SUB should have very similar descriptions to TIMESTAMP_ADD. I dropped the ball when reviewing TIMESTAMP_ADD. It should have been
```
TIMESTAMP_ADD(timestamp, interval) | Returns the TIMESTAMP value that occurs *interval* after *timestamp*
```

`TIMESTAMP_SUB` should be "the TIMESTAMP value that occurs *interval* before *timestamp*", and similarly DATE_SUB, TIME_SUB

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -700,6 +727,14 @@ private SqlLibraryOperators() {
           ReturnTypes.TIMESTAMP_NULLABLE, OperandTypes.INTEGER,
           SqlFunctionCategory.TIMEDATE);
 
+  /** 2-argument form of the special minus-date operator
```
julianhyde, line 730: I'm not sure how MINUS_DATE2 gets found by the validator. The syntax is 'special' so it isn't found by name. Users can't reference it in their queries. If the operator is only used in intermediate expressions move it to `class SqlInternalOperators`.

I also don't see why the return-type inference needs to be different.

tanclary: I parameterized the return-type inference because the standard MINUS_DATE operator has three arguments, and the return type is specified as ARG2. This is incompatible with BigQuery because there are only two arguments (the time/date/timestamp expression, and the interval being subtracted).

PR #3009, opened 2022-12-20
CALCITE-5447/ DATE_TRUNC() function for BIG_QUERY

olivrlee:


babel/src/test/resources/sql/redshift.iq
```diff
@@ -1238,7 +1238,7 @@ select date_part(minute, time '02:08:01');
 select date_part(minute, date '2009-01-01');
 !ok
 !}
-
+!if (false) {
```
olivrlee, line 1241: DATE_TRUNC() in redshift has the operands reversed (DATE_TRUNC(date_part, date_expression) and the babel test has it marked as LENIENT for functions that aren't implemented, so I'm commenting this out.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -5050,6 +5050,86 @@ SqlIntervalQualifier TimeUnitOrName() : {
...
+    LOOKAHEAD(1)
+    timeUnit = TimeUnitForDateTrunc() {
+        return new SqlIntervalQualifier(timeUnit, null, getPos());
+    }
+|
+    weekdayIntervalQualifier = WeekWeekdaySqlIntervalQualifier() {
+        return weekdayIntervalQualifier;
+    }
+|
+    unitName = DateTruncIdentifiers() {
```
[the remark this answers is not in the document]
olivrlee: I was following this description for the permitted operands: https://g3doc.corp.google.com/company/teams/googlesql/reference/functions-and-operators.md?cl=head#date-trunc
I was under the assumption that only these operands would be accepted and handled since I saw tests in `SqlOperatorTest` that fails when called with things like `microsecond`, `nanosecond`

core/src/main/codegen/templates/Parser.jj
```diff
@@ -5050,6 +5050,86 @@ SqlIntervalQualifier TimeUnitOrName() : {
...
+
+SqlIntervalQualifier WeekWeekdaySqlIntervalQualifier() : {
+    final String weekdayName;
+}
+{
+    weekdayName =  WeekWeekday() {
+        return new SqlIntervalQualifier(weekdayName, getPos());
+    }
+}
+String WeekWeekday() :
```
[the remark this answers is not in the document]
olivrlee: Yep I have this followup https://issues.apache.org/jira/browse/CALCITE-5450 to add to all the other functions, but I'd like to get this reviewed before I go and do that

babel/src/test/resources/sql/redshift.iq
julianhyde: This test should be re-enabled before we accept the PR. If the test is incorrect, fix it. If you need a different DATE_TRUNC function for BQ than for Redshift, that can be accomplished.

core/src/main/codegen/templates/Parser.jj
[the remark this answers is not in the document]
julianhyde: @olivrlee FYI that link is not available to people outside your VPN

babel/src/test/resources/sql/big-query.iq
```diff
@@ -1691,16 +1691,38 @@ SELECT TIMESTAMP_DIFF("2001-02-01 01:00:00", "2001-02-01 00:00:01", HOUR) AS neg
 #     corresponding Gregorian calendar year.
 #
 # Returns DATE
-
-!if (false) {
-SELECT DATE_TRUNC(DATE '2008-12-25', MONTH) AS month;
+WITH Dates AS (
+  SELECT DATE_TRUNC(DATE '2008-12-25', YEAR) as d UNION ALL
```
julianhyde, line 1695: can you add an extra column, say `'year' as frame, ` in row 1, 'isoyear' in row 2, just in case the rows get reordered.

babel/src/test/java/org/apache/calcite/test/BabelTest.java
```diff
@@ -189,6 +189,23 @@ private void checkInfixCast(Statement statement, String typeName, int sqlType)
         .fails("'A' is not a valid time frame");
     f2.withSql("SELECT DATE_PART(minute15, " + ts + ")")
         .ok();
+
+    final SqlValidatorFixture f3 =
```
julianhyde, line 193: I think this test should be moved. When DATE_TRUNC supports custom time frames, you will no longer get the parse error "Encountered \"A\" at line 1". Add something to `SqlParserTest.testTimeUnitCodes` and something to `SqlValidatorTest.checkTimeUnitCodes` and I think you're good.

core/src/main/codegen/templates/Parser.jj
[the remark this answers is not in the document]
julianhyde: In CALCITE-5155 comment I wrote:

>  Previously, NANOSECOND and MILLISECOND were allowed in
> EXTRACT but no other functions. Now all functions that
>  accept time frames accept the same time frames (built-in
>  time intervals, identifiers for user-defined time frames,
>   and SQL_TSI_xxx which are defined in the JDBC standard but
>   are treated as identifiers until validation).

I think you should accept all time frames (including custom frames) in the parser. That should make the parser much simpler. We *could* reject certain frames at validate time, but I don't think we will. It's easier if we implement a superset of what BQ implements.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -5050,6 +5050,86 @@ SqlIntervalQualifier TimeUnitOrName() : {
...
+|
+    LOOKAHEAD(4)
+    <WEEK> <LPAREN> <FRIDAY> <RPAREN> { return "WEEK_FRIDAY"; }
+|
+    LOOKAHEAD(4)
+    <WEEK> <LPAREN> <SATURDAY> <RPAREN> { return "WEEK_SATURDAY"; }
+|
+    LOOKAHEAD(1)
+    <WEEK> { return "WEEK"; }
+}
```
julianhyde, line 5120: High lookahead values are expensive, and LOOKAHEAD(4) might be overkill. I would write something like

```
<WEEK>
(
  <LPAREN> w = weekDay() <RPAREN> { return weekDays.get(w); }
|
  /* empty */ { return "WEEK"; }
)
```
and now the parser doesn't have to go so deep before it makes a decision.

core/src/main/java/org/apache/calcite/rel/type/TimeFrames.java
```diff
@@ -100,6 +101,28 @@ private static BuilderImpl addCore(BuilderImpl b) {
 
     b.addRollup(TimeUnit.DAY, TimeUnit.MONTH);
     b.addRollup("ISOWEEK", TimeUnit.ISOYEAR.name());
+    b.addSub("WEEK_SUNDAY", false, 7,
+        "DAY", new TimestampString(1970, 1, 4, 0, 0, 0)
+    );
+    b.addSub("WEEK_MONDAY", false, 7,
```
julianhyde, line 107: no space before ')'

site/_docs/reference.md
```diff
@@ -2616,6 +2625,7 @@ semantics.
 | q | DATEPART(timeUnit, datetime)                   | Equivalent to `EXTRACT(timeUnit FROM  datetime)`
 | b | DATE_FROM_UNIX_DATE(integer)                   | Returns the DATE that is *integer* days after 1970-01-01
 | p | DATE_PART(timeUnit, datetime)                  | Equivalent to `EXTRACT(timeUnit FROM  datetime)`
+| b | DATE_TRUNC(date_expressions, date_part)        | Truncates a *date* value to the granularity of *date_part*. The *date* value is always rounded to the beginning of the *date_part*.
```
julianhyde, line 2628: no underscores, no final period

core/src/main/codegen/templates/Parser.jj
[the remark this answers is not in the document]
olivrlee: Fixed link in comment above! thanks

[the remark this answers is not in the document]
olivrlee: I will change parser to allow custom time frames and have the operands be a super set of what BigQuery enables

babel/src/test/resources/sql/redshift.iq
olivrlee: Re-enabled!

julianhyde: I know Redshift was not one of your specific goals, so I want to thank you for going the extra mile to make this work for Redshift. In Calcite, quality tends to multiply like compound interest, so thanks for playing your part.

PR #3010, opened 2022-12-21
[CALCITE-5408] Return type of PERCENTILE_CONT should be the same as sort expression

zoudan:
Return type of PERCENTILE_CONT should be the same as sort expression

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
@@ -2195,7 +2195,7 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
    */
   public static final SqlAggFunction PERCENTILE_CONT =
       SqlBasicAggFunction
-          .create(SqlKind.PERCENTILE_CONT, ReturnTypes.DOUBLE,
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -2195,7 +2195,7 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
    */
   public static final SqlAggFunction PERCENTILE_CONT =
       SqlBasicAggFunction
-          .create(SqlKind.PERCENTILE_CONT, ReturnTypes.DOUBLE,
```
libenchao, line 2198: the comment should also be changed.

libenchao (commented):
@zoudan Thanks for your PR, I only left one minor comment.

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
zoudan: @libenchao Thanks for reviewing, I have resolved it just now.

libenchao (approved):
LGTM, will merge after the CI passes. Thanks for your contribution @zoudan

PR #3011, opened 2022-12-21
[CALCITE-5452] Add BigQuery LENGTH() as synonym for CHAR_LENGTH()

tanclary:
Add LENGTH() as a library function as an alias for the standard CHAR_LENGTH(). Some dependencies for soon-to-be deprecated standard functions refactored to avoid null pointer exceptions as a result of circular dependencies between the standard and library operators. This decision was made with the help of @mkou .

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -6231,6 +6231,28 @@ public class SqlOperatorTest {
     f0.forEachLibrary(list(SqlLibrary.BIG_QUERY, SqlLibrary.ORACLE), consumer);
   }
 
+  @Test void testLengthFunc() {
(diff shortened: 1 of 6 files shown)


julianhyde:
Can you change the commit message to start with the Jira case. Just like the PR subject.

The comments about deprecated functions above are useful, but would be better as part of the commit message.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6231,6 +6231,28 @@ void assertSubFunReturns(boolean binary, String s, int start,
     f0.forEachLibrary(list(SqlLibrary.BIG_QUERY, SqlLibrary.ORACLE), consumer);
   }
 
+  @Test void testLengthFunc() {
```
julianhyde, line 6234: I'd move this test next to `testCharLengthFunc` because it is so similar.

Use the `f0.forEachLibrary(list(SqlLibrary.BIG_QUERY), consumer);` pattern so that the test can be extended to other libraries in future.

PR #3012, opened 2022-12-22
[CALCITE-5454] Update BigQuery Conformance for != and % operators

tanclary:
Update SqlConformance to include BigQuery as an allowed dialect for != and % Operators.

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -2712,6 +2712,14 @@ public class SqlOperatorTest {
     f1.checkBoolean("1 <> 1", false);
     f1.checkBoolean("1 != 1", false);
     f1.checkBoolean("1 != null", null);
+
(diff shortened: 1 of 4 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -2712,6 +2712,14 @@ private static void checkNullOperand(SqlOperatorFixture f, String op) {
...
     f1.checkBoolean("1 != null", null);
+
+    // "!=" is allowed under BIG_QUERY SQL conformance level
+    final SqlOperatorFixture f2 =
+        f.withConformance(SqlConformanceEnum.BIG_QUERY);
+    f2.checkBoolean("1 <> 1", false);
+    f2.checkBoolean("1 != 1", false);
+    f2.checkBoolean("2 != 1", true);
+    f2.checkBoolean("1 != null", null);
   }
```
julianhyde, line 2723: you have copy-pasted 3 tests from above and added one. can you make the Oracle and BigQuery conformances share code? E.g.
```
```

PR #3022, opened 2023-01-09
[CALCITE-5468] SqlToRelConverter throws if ORDER BY contains IN

dssysolyatin:


core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -4769,4 +4769,18 @@ void checkUserDefinedOrderByOver(NullCollation nullCollation) {
...
+   * SqlToRelConverter should register sub-queries inside ORDER BY clause
+   * for queries without aggregation</a>.
+   */
+  @Test void testOrderByWithSubQuery() {
+    String sql = "SELECT empno\n"
+        + "FROM emp\n"
+        + "ORDER BY\n"
+        + "CASE WHEN empno IN (1,2) THEN 0 ELSE 1 END";
+    sql(sql).ok();
+  }
```
julianhyde, line 4785: Don't put stuff on the end. It causes merge conflicts. Put it in the most logical place.

julianhyde: This IN is not a subquery. Can you rename this test, and add a test for a subquery?

dssysolyatin: Yes, IN is not subquery in terms of SQL. But inside codebase there are places with code like:
```
/**
 * Scalar expression that represents an IN, EXISTS or scalar sub-query.
 */
public class RexSubQuery extends RexCall
```
It is actually quite confusing.

I renamed test and added test for subquery.

dssysolyatin:
@julianhyde Can I merge this PR ? Or there is any objections ?

julianhyde:
@dssysolyatin Looks good. Go ahead and merge to main.

PR #3023, opened 2023-01-11
[CALCITE-5180] Implement BigQuery Date/Time Type Aliases and Constructors

wnob:
This incorporates the type alias work drafted by Julian in https://github.com/julianhyde/calcite/tree/5424-custom-literals and builds on it by incorporating my own first draft to address CALCITE-5180 ([Implement BigQuery date/time constructors](https://github.com/apache/calcite/commit/eb281b1706a6435f791b886ea86dc2aa65faf0f7)), and then adding in the `DATETIME()` function making use of the type aliases ([Add DATETIME() function and switch to the core parser](https://github.com/apache/calcite/commit/6e5b2f22414b1c1617d0e721d6dcf93f1b82d673)), and also switches the parsing logic from the Babel parser to the core parser.


core/src/main/java/org/apache/calcite/sql/SqlUnknownLiteral.java
```diff
@@ -0,0 +1,68 @@
...
+    switch (typeName) {
+    case DATE:
+      return SqlParserUtil.parseDateLiteral(getValue(), pos);
+    case TIME:
+      return SqlParserUtil.parseTimeLiteral(getValue(), pos);
+    case TIMESTAMP:
+      return SqlParserUtil.parseTimestampLiteral(getValue(), pos);
+    case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
+      return SqlParserUtil.parseTimestampWithLocalTimeZoneLiteral(getValue(), pos);
+    default:
```
olivrlee, line 60: did you miss a case for DATETIME? 
https://github.com/apache/calcite/pull/3023/files#diff-e873041549333502af52ece8a1b34301ae5a059ff4719e9bddbaef48929e7047R4617

olivrlee: nvm discussed offline, I think that location should be `return SqlLiteral.createUnknown("TIMESTAMP")` instead

core/src/main/codegen/templates/Parser.jj
```diff
@@ -4609,15 +4610,116 @@ SqlLiteral DateTimeLiteral() :
...
+    final List<? extends SqlNode> args;
+}
+{
+    <DATE> {
+        s = span();
+        qualifiedName = new SqlIdentifier(unquotedIdentifier(), getPos());
+    }
+    args = FunctionParameterList(ExprContext.ACCEPT_SUB_QUERY) {
+        quantifier = (SqlLiteral) args.get(0);
+        args.remove(0);
```
olivrlee, line 4652: what is this line doing?

julianhyde:
There's too much going on here for me to review.

Can you separate it out into pieces that can be understood by an end user as adding some useful functionality? If there refactorings that don't change functionality, keep those separate.

It's confusing that there are type aliases, and there are also functions that are named after types. Basically you need to craft commit messages that are suitable for the release notes. (And each such commit should be a jira case that explains the new functionality.)

babel/src/test/resources/sql/big-query.iq
```diff
@@ -1097,12 +1097,74 @@ select unix_date(datetime '2008-12-25') as d;
...
+| sum_utc    | sum_dst    | sum_std    | win_utc    | win_dst    | win_std    |
++------------+------------+------------+------------+------------+------------+
+| 2008-06-21 | 2008-06-21 | 2008-06-20 | 2008-12-21 | 2008-12-20 | 2008-12-20 |
++------------+------------+------------+------------+------------+------------+
+(1 row)
+
+!ok
+
+#####################################################################
+# DATETIME
```
olivrlee, line 1131: there's a set of DATETIME tests currently disabled around line 650 btw 
https://github.com/apache/calcite/blob/6e5b2f22414b1c1617d0e721d6dcf93f1b82d673/babel/src/test/resources/sql/big-query.iq#L649

core/src/main/codegen/templates/Parser.jj
wnob: If you look at the production for `FunctionParameterList()` you'll see it start like this:
```
{
    <LPAREN>
    (
        qualifier = AllOrDistinct() { list.add(qualifier); }
    |
        { list.add(null); }
    )
    ...
```
Here, `list` is the return value (`args` here in `DateFunctionCall()`). So the first argument is always either `null`, or either `ALL` or `DISTINCT`, since some functions can take those qualifiers. Here, it should always be `null`, although I suppose this means the parser would silently accept something like `DATE(DISTINCT 2023, 01, 18)`, which it probably shouldn't.

PR #3024, opened 2023-01-13
[CALCITE-5469] Implement BigQuery DATETIME_ADD/DATETIME_DIFF

tanclary:
This is a beginning draft for the work to implement BigQuery DATETIME_ADD/DATETIME_DIFF. Currently, DATETIME_ADD is functioning as expected and is passing all operator and quidem tests. DATETIME_DIFF is not currently passing quidem tests for a couple of different reasons: the 'week' time unit is only counting whole number of weeks rather than demarcating each Sunday, as well as 'isoyear' for a similar reason. Any comments on these issues or the rest of the changes would be highly appreciated! More information about the functions may be found below.

DATETIME_ADD(datetime, interval) can accept a timestamp (or a datetime, which is an alias for timestamp) for its first argument and an interval for its second. The output is the datetime that occurs interval after the provided datetime.

DATETIME_DIFF(datetime, datetime2, timeUnit) returns the whole number of timeUnit between datetime and datetime2, with the result being negative if datetime occurs before datetime2

Examples:
DATETIME_ADD(TIMESTAMP '2008-12-25 15:30:00', INTERVAL 5 MINUTE) would return: '2008-12-25 15:35:00'.

DATETIME_DIFF(TIMESTAMP '2008-12-25 15:30:00', TIMESTAMP '2008-12-26 15:30:00', DAY) would return: -1.


core/src/main/codegen/templates/Parser.jj
```diff
@@ -4677,6 +4677,28 @@ SqlNode DatetimeFunctionCall() :
     }
 }
 
+/**
+ * Parses BigQuery's built-in DATETIME_DIFF() function.
```
julianhyde, line 4681: It's not 'BigQuery's'. Just "Parses the DATETIME_DIFF function".

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -764,6 +764,22 @@ private SqlLibraryOperators() {
           ReturnTypes.BIGINT_NULLABLE, OperandTypes.TIMESTAMP,
           SqlFunctionCategory.TIMEDATE);
 
+  /** BigQuery's "DATETIME_ADD(timestamp, interval) function; Behaves similarly
```
julianhyde, line 767: I wouldn't describe it as 'BigQuery's'. Other DBs might support it one day.

The "because in Calcite, datetime is a type alias for timestamp" is confusing, and I'd remove it. The fact is, both accept TIMESTAMP or TIMESTAMP WITH LOCAL TIME ZONE arguments, and both return the same type as their first argument.

site/_docs/reference.md
```diff
@@ -2620,6 +2622,8 @@ semantics.
 | b | DATETIME(date, time)                           | Returns a datetime object, given a date and a time.
 | b | DATETIME(timestamp)                            | Converts a timestamp object to a datetime, assuming UTC.
 | b | DATETIME(timestamp, timezone)                  | Converts a timestamp object to a datetime, in a given time zone.
+| b | DATETIME_ADD(timestamp, interval)              | Synonym for TIMESTAMP_ADD.
```
julianhyde, line 2625: remove '.'

Not really a synonym. The implementation happens to allow both TIMESTAMP and TIMESTAMP WITH LOCAL TIME ZONE, and therefore DATETIME_ADD shares an implementation with TIMESTAMP_ADD. But the user doesn't have to know that.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8122,6 +8122,170 @@ private static void checkArrayConcatAggFuncFails(SqlOperatorFixture t) {
...
+        "2016-03-24 12:42:25",
+        "TIMESTAMP(0) NOT NULL");
+    f.checkScalar("datetime_add(timestamp '2016-02-24 12:42:25', interval 1 year)",
+        "2017-02-24 12:42:25",
+        "TIMESTAMP(0) NOT NULL");
+    f.checkNull("datetime_add(CAST(NULL AS TIMESTAMP), interval 5 minute)");
+
+  }
+
+  /** Tests BigQuery's DATETIME_DIFF(timestamp, timestamp2, timeUnit) function. In Calcite,
```
julianhyde, line 8172: same remarks as for DATETIME_ADD

julianhyde (commented):
Looks good. I think there's only javadoc left to fix.

tanclary:
@julianhyde Thanks for your comments! I added a commit that fixes up the comments. I'll rebase this once the commits it's dependent on have been resolved and merged.

PR #3025, opened 2023-01-13
[CALCITE-5464] Implement BigQuery DATE_ADD/DATE_DIFF

tanclary:
Implement BigQuery DATE_ADD/DATE_DIFF functions. More info on functions may be found below.

Second commit includes additional logic to handle the case of 'WEEK' being the desired time unit as Calcite and BigQuery have differing definitions. By truncating each date to the most recent Sunday, this difference is remedied.

DATE_ADD(date_expression, interval) adds the interval to the date expression.
Example: DATE_ADD(DATE '2008-12-25', INTERVAL 3 DAY) would return '2008-12-28'.
BigQuery docs: https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date_add

DATE_DIFF(date_expression, date_expression2, time_unit) returns the whole number of time_unit between the first and second date expressions, with the result being negative if the first date is earlier than the second.
Example: DATE_DIFF(DATE '2008-12-25', DATE '2008-12-28', DAY) would return -3.
BigQuery docs: https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date_diff


chunweilei (commented):
LGTM

PR #3028, opened 2023-01-14
[CALCITE-5477] Prefer Util.checkArgument over Preconditions.checkArgument

gianm:
Preconditions.checkArgument gained new overloads in Guava 20.0, which prevents code built using Guava 20.0 from running properly with earlier versions of Guava.


gianm:
Closed in favor of https://github.com/apache/calcite/pull/3249.

PR #3030, opened 2023-01-14
[CALCITE-5479] FamilyOperandTypeChecker is not readily composable in sequences

gianm:
By adjusting handling of iFormalOperand in various checkers.

1) CompositeOperandTypeChecker: always send the proper iFormalOperand; never zero.

2) FamilyOperandTypeChecker: always look at the first (and only) family when being used as a SqlSingleOperandTypeChecker via checkSingleOperandType. Continue considering all families when used as a SqlOperandTypeChecker via checkOperandTypes.

3) IntervalOperandTypeChecker: use the SqlNode parameter, not the SqlCallBinding, to get the node to check. Using the SqlCallBinding violates the contract of SqlSingleOperandTypeChecker#checkSingleOperandType.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/type/CompositeOperandTypeChecker.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/CompositeOperandTypeChecker.java
@@ -312,7 +312,7 @@ public class CompositeOperandTypeChecker implements SqlOperandTypeChecker {
         if (!((SqlSingleOperandTypeChecker) rule).checkSingleOperandType(
             callBinding,
             callBinding.getCall().operand(ord.i),
-            rule.getClass() == FamilyOperandTypeChecker.class ? 0 : ord.i,
+            ord.i,
             throwOnFailure)) {
           if (callBinding.isTypeCoercionEnabled()) {
             return coerceOperands(callBinding, false);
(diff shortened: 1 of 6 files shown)


gianm:
There are some Rat failures, but they don't seem related to this patch.

```
Files with unapproved licenses:
 - /home/runner/work/calcite/calcite/C:/Users/runneradmin/.gradle-cache-action/gradle-generated-jars-contents.json
 - /home/runner/work/calcite/calcite/C:/Users/runneradmin/.gradle-cache-action/gradle-generated-jars-info.json
```

Does anyone know if there is something I can do about these?

libenchao:
@gianm There is a discussion in the dev ML: https://lists.apache.org/thread/jx14cyz5dddmmx6mc13g3l6srq2lf2cc about this.

gianm:
Thanks @libenchao for the pointer!

I rebased on main and updated the patch to add a new unit test illustrating the problem. It fails on main, passes in this branch. The linked JIRA has some details on the thought process behind this patch: https://issues.apache.org/jira/browse/CALCITE-5479?focusedCommentId=17690605&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17690605

PR #3031, opened 2023-01-16
[CALCITE-5483] ProjectAggregateMergeRule throws exception if literal is non-numeric

birschick-bq:
Resolve issue with thrown exception if literal is non-numeric.

https://issues.apache.org/jira/browse/CALCITE-5483

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/ProjectAggregateMergeRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/ProjectAggregateMergeRule.java
@@ -33,6 +33,7 @@ import org.apache.calcite.rex.RexShuttle;
 import org.apache.calcite.rex.RexVisitorImpl;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.type.SqlTypeFamily;
 import org.apache.calcite.tools.RelBuilder;
 import org.apache.calcite.util.ImmutableBitSet;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/rel/rules/ProjectAggregateMergeRule.java
```diff
@@ -99,7 +100,8 @@ && kindCount(project.getProjects(), SqlKind.CASE) == 0) {
               && operands.get(1).getKind() == SqlKind.CAST
               && ((RexCall) operands.get(1)).operands.get(0).getKind()
               == SqlKind.INPUT_REF
-              && operands.get(2).getKind() == SqlKind.LITERAL) {
+              && operands.get(2).getKind() == SqlKind.LITERAL
+              && operands.get(2).getType().getFamily() == SqlTypeFamily.NUMERIC) {
```
clesaec, line 104: I wonder where the code `literal.getValueAs(BigDecimal.class)` (line 111) could work as [getValueAs method](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/rex/RexLiteral.java#L1029)  has no case for "clazz == BigDecimal" ... May be not related to this specific case, but it seems there is also an issue on RexLiteral class.

birschick-bq: @clesaec 
I agree with you on this. I believe there could be a conversion for `BigDecimal` from `TIMESTAMP`.

birschick-bq:
cc: @CassieLyu / @sunnie629

PR #3034, opened 2023-01-18
[CALCITE-5357] Implement BigQuery FORMAT_TIMESTAMP, FORMAT_DATE, FORMAT_TIME, and FORMAT_DATETIME

tjbanghart:
Lays the groundwork for [CALCITE-2980](https://issues.apache.org/jira/browse/CALCITE-2980)


core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
```diff
@@ -184,6 +185,46 @@ public MysqlSqlDialect(Context context) {
     return super.getCastSpec(type);
   }
 
+  /** {@inheritDoc}
+   *
+   * <p>MySQL format element reference:
+   * <a href="https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_date-format">
+   * MySQL Date and Time Functions</a>.
+   */
+  @Override public String getFormatElement(FormatElementEnum fmtElement) {
```
tanclary, line 194: Why does this logic go in MySqlSqlDialect?

tjbanghart: Each dialect might have a different set of elements to represent the same type of datetime unit. 

For example an abbreviated month name like `Jan` is `%b` in MySQL and `MON` in Snowflake. Each `SqlDialect` will have to override this method to supply the correct string for the element type. I only did MySQL, BigQuery, and Snowflake in the first commit because I wanted to get some feedback before working on each dialect.

Oracle's format elements were chosen as the "base" set since they map closely to the SQL:2016 standard as described in [CALCITE-2980](https://issues.apache.org/jira/browse/CALCITE-2980).

tjbanghart:
@tanclary yes I will add those. Thanks for reminding me!

core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
tanclary: Makes sense! Looks great.

core/src/main/java/org/apache/calcite/sql/FormatModel.java
```diff
@@ -0,0 +1,154 @@
...
+public class FormatModel extends SqlInternalOperator {
+
+  private List<FormatModelElement> elements;
+  private ImmutableMap<String, FormatModelElement> fmtModelParseMap;
+
+  /**
+   * TODO(CALCITE-2980): This should live elsewhere and be associated with {@link SqlLibrary}
+   * or {@link org.apache.calcite.config.Lex}.
+   */
+  public static final ImmutableMap<String, FormatModelElement> BIG_QUERY_FORMAT_ELEMENT_PARSE_MAP =
```
mkou, line 72: nit:  I know it's everywhere so no need to change it but in general I hake a bit when I see BigQuery since we really want to do "Google SQL"

core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
```diff
@@ -240,6 +281,17 @@ public MysqlSqlDialect(Context context) {
       unparseListAggCall(writer, call, null, leftPrec, rightPrec);
       break;
 
+    case FORMAT_DATE:
+    case FORMAT_TIME:
+    case FORMAT_TIMESTAMP:
+    case FORMAT_DATETIME:
+      writer.print("DATE_FORMAT(");
```
mkou, line 288: This feels a bit weird for me but I might miss some context.
Should this live in the definition of FORMAT_DATETIME that DATE_FORMAT is actually an aliases with the operands in different order + different formats elements?

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -271,6 +272,50 @@ private static TimeUnit validate(TimeUnit timeUnit) {
     }
   }
 
+  /** {@inheritDoc}
+   *
+   * <p>BigQuery format element reference:
+   * <a href="https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements">
+   * BigQuery Standard SQL Format Elements</a>.
+   */
+  @Override public String getFormatElement(FormatElementEnum fmtElement) {
```
mkou, line 281: Yes I like this 🌷

core/src/main/java/org/apache/calcite/util/format/FormatModelElement.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * A format element in a format string. Knows how to parse and unparse itself.
+ */
+public interface FormatModelElement {
```
mkou, line 28: Adding examples for each methods would help take out some of the mental load of figuring out exactly the differences. (Easier to read) even though the names are are self suffiscient that would help

core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
mkou: Like all of them should alias to the same canonical form so you can read the canonical form in every language idk if that makes sense

core/src/main/java/org/apache/calcite/util/format/FormatModelUtil.java
```diff
@@ -0,0 +1,74 @@
...
+
+  private FormatModelUtil() {}
+
+  /**
+   * Parses the {@code fmtString} using element identifiers supplied by {@code fmtModel}.
+   *
+   * TODO(CALCITE-2980): make this configurable for multiple parse maps. Currently this only works
+   * for BigQuery and MySQL style format strings where elements begin with '%'
+   */
+  public static List<FormatModelElement> parse(String fmtString,
```
mkou, line 37: Not sure if this is a good advice or not but it looks like a good case to just use ReGex and look for the formats defined by each dialect

core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
tjbanghart: Yeah that's open for discussion -- how to best swap a `FORMAT_TIMESTAMP` Google SQL style function call to any other dialect. Snowflake for example uses `TO_CHAR` or `TO_VARCHAR`. 

Maybe a generic entry in `SqlKind` called `GOOGLE_SQL_DATETIME_FORMAT`?
Could even be more generic like `DATETIME_FORMAT` and then each dialect would know how to unparse.

core/src/main/java/org/apache/calcite/sql/FormatModel.java
tjbanghart: For sure, I can open a Jira case to swap `GOOGLE_SQL` or `GSQL` with `BIG_QUERY` references -- would have to a non-breaking change for config settings etc.

core/src/main/java/org/apache/calcite/util/format/FormatModelElementAlias.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.StringJoiner;
+
+/**
+ * Represents a format element comprised of one or more {@link FormatElementEnum} entries.
+ */
+public class FormatModelElementAlias implements FormatModelElement {
```
olivrlee, line 29: what does this `FormatModelElementAlias` actually represent? a collection of `FormatModelElement`s?

tjbanghart: Yeah exactly - it maps to one or more FormatModelElementEnum entries.

For example in Google SQL `%R` is the same as `HH24:MI`. I'll add more context/examples in the javadoc

core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
tjbanghart: Or better yet convert these BQ format functions into the standard `CAST(... as FORMAT <str>)` and unparse the `CAST` to `DATE_FORMAT`, `TO_CHAR` accordingly.

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
```diff
@@ -1002,6 +1004,18 @@ protected static void unparseOffset(SqlWriter writer, @Nullable SqlNode offset)
     }
   }
 
+  /**
```
julianhyde, line 1007: There's a complex mapping between format elements and dialects. But I suspect that SqlDialect is not the right place to do it, and functions is closer to the right place.

Hypothetically, someone could be in MySQL dialect and want to enable MySQL's and Postgres's formatting functions. It's a useful hypothetical. SqlDialect is used only for a short part of the query life cycle - when going from AST nodes to SQL text. But functions are around for longer.

I have in mind a map from strings to format elements that is part of the `SqlLibraryOperators.FORMAT_TIMESTAMP` function, and the same or similar map that is used by `FORMAT_TIME` and `PARSE_TIMESTAMP` functions. Parse functions might have an additional parser, because parsing format strings is not always trivial.

Maybe the aforementioned 'map' and 'parser' would evolve into slightly larger classes.

When we tackle the different issue for 'how do I translate a MySQL format string to a BigQuery format string?' - or 'how do I translate a call to MySQL_FORMAT_TIMESTAMP to a call to BigQuery_FORMAT_TIMESTAMP' - then those larger classes could be used to drive the proceedings.

core/src/main/java/org/apache/calcite/sql/fun/SqlBigQueryFormatDatetimeFunction.java
```diff
@@ -0,0 +1,111 @@
...
+ *   <a href="https://cloud.google.com/bigquery/docs/reference/standard-sql/time_functions#format_time">ref</a></li>
+ *   <li>{@code FORMAT_DATE(format_string, date_expr)}
+ *   <a href="https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#format_date">ref</a></li>
+ *   <li>{@code FORMAT_TIMESTAMP(format_string, timestamp[, time_zone])}
+ *  <a href="https://cloud.google.com/bigquery/docs/reference/standard-sql/timestamp_functions#format_timestamp">ref</a></li>
+ *  <li>{@code FORMAT_DATETIME(format_string, timestamp[, time_zone])}
+ *  <a href="https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#format_datetime">ref</a></li>
+ * </ul>
+ */
+public class SqlBigQueryFormatDatetimeFunction extends SqlFunction {
```
julianhyde, line 51: Parsing/formatting functions might have enough commonality to warrant a common class - an extension to `SqlBasicFunction` with an extra field, a map from strings to format elements.

I wouldn't change it now. But come back and refactor if that pattern emerges.

julianhyde: However I would make the constructor private now. A public constructor will prevent you from subclassing later. Use a `create` method instead, and add `withXxx` methods to change the value of parameters that are not frequently used. (Same pattern as `SqlBasicFunction`, and a pattern that is a little nearer to functional programming than object-oriented.)

core/src/main/java/org/apache/calcite/util/format/FormatModelUtil.java
tjbanghart: Took a stab at this in most recent commit.

core/src/main/java/org/apache/calcite/util/format/FormatModelUtil.java
```diff
@@ -30,44 +32,34 @@ private FormatModelUtil() {}
...
-          if (ch == '%') {
-            --idx;
-            break;
-          }
-          buf.append(ch);
-        }
-        elements.push(new FormatModelElementLiteral(buf.toString()));
-        buf.setLength(0);
-        ++idx;
+  public static List<FormatModelElement> parse(String format,
```
mkou, line 36: One thing that I realize is missing is the separator that is available in some dialects like redshift :/

core/src/main/java/org/apache/calcite/util/format/FormatModelUtil.java
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Parses the {@code fmtString} using element identifiers supplied by {@code fmtModel}.
+   */
+  public static List<FormatModelElement> parse(String format,
+      ImmutableMap<String, FormatModelElement> fmtModelMap) {
+    List<FormatModelElement> elements = new ArrayList<>();
+    // TODO(CALCITE-2980): make these regex patterns static and tied to a library or dialect.
+    StringBuilder regex = new StringBuilder();
+    for (String key : fmtModelMap.keySet()) {
+      regex.append("(").append(Pattern.quote(key)).append(")|");
```
mkou, line 42: Make sure you get the longest match

core/src/main/java/org/apache/calcite/util/format/FormatModelUtil.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+    // remove the last '|'
+    regex.setLength(regex.length() - 1);
+    Matcher matcher = Pattern.compile(regex.toString()).matcher(format);
+    int i = 0;
+    while (matcher.find()) {
+      // Add any leading literal text before next element match
+      String literal = format.substring(i, matcher.start());
+      if (!literal.isEmpty()) {
+        elements.add(new FormatModelElementLiteral(literal));
```
mkou, line 52: For parsing and unparsing, litterals might be in specific separators (`"` for example)

mkou (changes requested):

core/src/main/java/org/apache/calcite/sql/FormatModel.java
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * <p>{@link #unparse(SqlWriter, SqlCall, int, int)} calls
+ * {@link SqlDialect#getFormatElement(FormatElementEnum)} for known elements and aliases. Consider
+ * overriding this method if a dialect's format elements differs from those in {@link
+ * FormatElementEnum}
+ */
+public class FormatModel extends SqlInternalOperator {
+
+  private final List<FormatModelElement> elements;
+  private final ImmutableMap<String, FormatModelElement> fmtModelParseMap;
```
julianhyde, line 65: Field type should be `ImmutableMap<String, FormatModelElement>` rather than `Map<String, FormatModelElement>`. (I absolutely agree that the field value should be an `ImmutableMap`.) It's bad form to include implementation details in the field declaration. So, `TreeSet` bad, `SortedSet` good.

This is especially important for public fields and methods. People want to shade guava.

core/src/main/java/org/apache/calcite/sql/FormatModel.java
julianhyde: As discussed in the review, the world knows only BigQuery SQL so it's simpler to call it BigQuery SQL. I agree that Google SQL is more correct but it's more important to be understood than to be 100% correct.

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
julianhyde: I agree with @wnob that this `switch` should go, and it should use a map. Probably the map is bi-directional. Probably the map is the parse map, and that parse map can evolve from an ImmutableMap to a larger class.

core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
julianhyde: As I said previously, the association should be between the FORMAT function (as specified by MySQL) and the format element, not intermediated by the dialect object.

core/src/main/java/org/apache/calcite/util/format/FormatModelElement.java
```diff
@@ -0,0 +1,90 @@
...
```
julianhyde, line 28: I would rename 'FormatModelElement' to 'FormatElement'. It's shorter, and several subclasses follow this convention already.

core/src/main/java/org/apache/calcite/util/format/FormatModelElement.java
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * Whether this element is a {@link FormatModelElementAlias}. An alias comprises one or
+   * more standard elements declared in {@link FormatElementEnum}.
+   *
+   * <p>For example, {@code %R} in Google SQL represents the hour in 24-hour format (e.g., 00..23)
+   * followed by the minute as a decimal number. This method would return {@code true} as "%R" is
+   * an alias for the standard elements {@code HH24} and {@code MI}.
+   */
+  default boolean isAlias() {
```
julianhyde, line 67: `isAlias` makes you think that there are only two subtypes but there are 3 or 4 - built in, literal, composite, and maybe other user-defined.

I would replace both `isAlias` and `getElements` with a method
```
default void flatten(Consumer<FormatModelElement> consumer) {
  consumer.accept(this);
}
```
that is overridden in the composite format element.

It's possible that format elements shouldn't have names. The names are just keys by which they are included in a map. So the 'alias' behavior is better described as 'composite'.

core/src/main/java/org/apache/calcite/util/format/FormatModelElementAlias.java
```diff
@@ -0,0 +1,74 @@
...
```
julianhyde, line 27: As I mentioned previously, I think this should be renamed 'CompositeFormatElement'.

Consider moving into a helper class, `FormatModels`, and access via a method `public static FormatModelElement compositeElement(String, List<FormatModelElement>, String)`. It's possible that the inner class can be private.

core/src/main/java/org/apache/calcite/util/format/FormatModelElementAlias.java
```diff
@@ -0,0 +1,74 @@
...
+
+  private String alias;
+  private String description;
+  private List<FormatModelElement> formatElements;
+
+  private FormatModelElementAlias(String alias, List<FormatModelElement> formatElements,
+      String description) {
+    this.alias = Objects.requireNonNull(alias, "element alias");
+    this.formatElements = Objects.requireNonNull(formatElements, "format elements");
```
julianhyde, line 36: make a copy, using `ImmutableList.copyOf`

core/src/main/java/org/apache/calcite/util/format/FormatModelElementLiteral.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.apache.calcite.util.format;
+
+import java.util.Objects;
+
+/**
+ * Represents literal text in a format string.
+ */
+public class FormatModelElementLiteral implements FormatModelElement {
+
```
julianhyde, line 25: Consider moving into a helper class, `FormatModels.literalElement(String)`.

core/src/main/java/org/apache/calcite/util/format/FormatModelElementLiteral.java
```diff
@@ -0,0 +1,43 @@
...
+  private String literal;
```
julianhyde, line 26: `final`

core/src/main/java/org/apache/calcite/util/format/FormatModelElementAlias.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.StringJoiner;
+
+/**
+ * Represents a format element comprised of one or more {@link FormatElementEnum} entries.
+ */
+public class FormatModelElementAlias implements FormatModelElement {
+
+  private String alias;
```
julianhyde, line 29: all fields should be `final`; once they are `final`, they might as well be package-private

core/src/main/java/org/apache/calcite/util/format/FormatModelUtil.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Utility class used to convert format strings into {@link FormatModelElement}s.
+ */
+public class FormatModelUtil {
```
julianhyde, line 29: following modern naming conventions, I would rename `FormatModelUtil` to `FormatModels`.

I would move various subclasses of `FormatModelElement` into this class as inner classes. With luck - if we have designed the API correctly - these inner classes can be private.

core/src/main/java/org/apache/calcite/util/format/FormatModelUtil.java
```diff
@@ -0,0 +1,66 @@
...
+      ImmutableMap<String, FormatModelElement> fmtModelMap) {
+    List<FormatModelElement> elements = new ArrayList<>();
+    // TODO(CALCITE-2980): make these regex patterns static and tied to a library or dialect.
+    StringBuilder regex = new StringBuilder();
+    for (String key : fmtModelMap.keySet()) {
+      regex.append("(").append(Pattern.quote(key)).append(")|");
+    }
+    // remove the last '|'
+    regex.setLength(regex.length() - 1);
+    Matcher matcher = Pattern.compile(regex.toString()).matcher(format);
```
julianhyde, line 46: Compiling a regex each call is expensive. That implies that after you have created `fmtModelMap` you then build a "parser" object that compiles the regex and stores the `Pattern`.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1859,6 +1859,49 @@ private SqlDialect nonOrdinalDialect() {
...
+    final String formatDate = "select format_date('%b-%d-%Y', date '2012-02-03')\n"
+        + "from \"foodmart\".\"product\"\n";
+    final String formatTimestamp =
+        "select format_timestamp('%b-%d-%Y', timestamp '2012-02-03 12:30:40')\n"
+            + "from \"foodmart\".\"product\"\n";
+    final String formatDatetime = "select format_datetime('%R', timestamp '2012-02-03 12:34:34')\n"
+        + "from \"foodmart\".\"product\"\n";
+
+    final String expectedBqFormatTime =
+        "SELECT FORMAT_TIME('%H', TIME '12:45:30')\nFROM foodmart.product";
```
julianhyde, line 1874: I would break each of these strings after `\n`, so
```
    final String expectedBqFormatTime =
        "SELECT FORMAT_TIME('%H', TIME '12:45:30')\nFROM foodmart.product";
```
becomes
```
    final String expectedBqFormatTime =
        "SELECT FORMAT_TIME('%H', TIME '12:45:30')\n"
            + "FROM foodmart.product";
```

core/src/test/java/org/apache/calcite/util/FormatModelUtilTest.java
```diff
@@ -0,0 +1,81 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Unit test for {@link FormatModelUtil}.
+ */
+public class FormatModelUtilTest {
+
+  private static final ImmutableMap TEST_MAP = FormatModel.BIG_QUERY_FORMAT_ELEMENT_PARSE_MAP;
+
+  private void assertFormatElementParse(String fmtStr, List<String> expected) {
```
julianhyde, line 40: switch the 'expected' argument to a matcher, like this:
```
private void assertThatFormatElementParse(String fmtStr, Matcher<List<String>> matcher) {
```
so that you can write tests so that they can be read left-to-right in English:
```
assertThatFormatElementParse("%j", is(Arrays.asList("DDD")));
```

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
tjbanghart: This doesn't mean we should get rid of the `getFormatElement` method (or at least something like it) entirely right? 

A `SqlDialect` would still need to know how to get an appropriate format string from a BigQuery style one while unparsing. For example, `FORMAT_TIMESTAMP(TIMESTAMP '2012-12-02 12:05:00', '%R')` would need to unparse as`TO_CHAR( TIMESTAMP '2012-12-02 12:05:00', 'HH24:MI')` for Postgres based dialects.

As you mentioned elsewhere,  this translation should not happen in a `FormatModel` _operator_ but rather the formatting function itself. That is, we should reach into the `SqlDialect` for this element mapping while in `SqlBigQueryFormatDatetimeFunction#unparse`

For library specific functions this mapping could also be used as the parse map.

tjbanghart: As a followup to this PR, we can define non-standard element maps per dialect. Then, when we unparse a formatting function we would use the map to find the appropriate element string for the "standard" element. The map would be bi-directional as suggested so that it can be used as both the parse map for functions and for converting elements from one dialect to another.

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
tjbanghart: In addressing this comment I tried to do both things -- formatting/parse functions are given a static `FormatModels` instance for parsing strings. This `FormatModels` receives its element mapping from the `BigQuerySqlDialect` class. 

We could reverse the hierarchy if needed so that the format function holds the element mapping and the SqlDialect refers to that for unparsing.

core/src/main/java/org/apache/calcite/sql/FormatModel.java
```diff
@@ -0,0 +1,154 @@
...
+/** A <a href="https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlqr/Format-Models.html">
+ * format model</a> is a character literal that describes the format of {@code DATETIME} or {@code
+ * NUMBER} data stored in a character string.
+ *
+ * <p>{@link #unparse(SqlWriter, SqlCall, int, int)} calls
+ * {@link SqlDialect#getFormatElement(FormatElementEnum)} for known elements and aliases. Consider
+ * overriding this method if a dialect's format elements differs from those in {@link
+ * FormatElementEnum}
+ */
+public class FormatModel extends SqlInternalOperator {
```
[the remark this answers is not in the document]
tjbanghart: I removed this operator class. Instead, I moved most of the classes into the `FormatModels` utility class as you had suggested elsewhere. It makes sense for them to be in a single place and we no longer try to string a string argument as something other than a string (i.e. no more premature optimization to convert a string into a `FormatModel`)

core/src/main/java/org/apache/calcite/util/format/FormatModelUtil.java
tjbanghart: Formatting functions should be able to create a `FormatModels` instance that will compile the regex from the parse map during initialization. Calls to `parse` should now use the precomputed regex.

PR #3036, opened 2023-01-19
[CALCITE-5424] Custom literals

olivrlee:


core/src/main/java/org/apache/calcite/sql/SqlUnknownLiteral.java
```diff
@@ -0,0 +1,68 @@
...
+    case TIMESTAMP:
+      return SqlParserUtil.parseTimestampLiteral(getValue(), pos);
+    case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
+      return SqlParserUtil.parseTimestampWithLocalTimeZoneLiteral(getValue(), pos);
+    default:
+      throw Util.unexpected(typeName);
+    }
+  }
+
+  @Override public <T> T getValueAs(Class<T> clazz) {
```
olivrlee, line 65: Q: Is this override necessary if it's just calling `super.getValueAs` ?

PR #3041, opened 2023-01-23
[CALCITE-5491] Allow TIME and DATE to be args for TIMESTAMPDIFF

snuyanzin:
The PR allows to have `TIME` and `DATE` types args for `TIMESTAMPDIFF` as it was for 1.32.0 and earlier

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
@@ -1903,8 +1903,23 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
   /** The <code>TIMESTAMPDIFF</code> function. */
   public static final SqlFunction TIMESTAMP_DIFF =
       new SqlTimestampDiffFunction("TIMESTAMPDIFF",
-          OperandTypes.family(SqlTypeFamily.ANY, SqlTypeFamily.TIMESTAMP,
-              SqlTypeFamily.TIMESTAMP));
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -1903,8 +1903,23 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
   /** The <code>TIMESTAMPDIFF</code> function. */
   public static final SqlFunction TIMESTAMP_DIFF =
       new SqlTimestampDiffFunction("TIMESTAMPDIFF",
-          OperandTypes.family(SqlTypeFamily.ANY, SqlTypeFamily.TIMESTAMP,
-              SqlTypeFamily.TIMESTAMP));
+          OperandTypes.family(SqlTypeFamily.ANY, SqlTypeFamily.TIMESTAMP, SqlTypeFamily.TIMESTAMP)
+              .or(
```
julianhyde, line 1907: how about just
```
OperandTypes.family(SqlTypeFamily.ANY, SqlTypeFamily.DATETIME,
            SqlTypeFamily.DATETIME)
```
which is the code that was removed in bc4cfa32 ?

tanclary: I +1 this, I meant to keep it the same in my original change but slipped up and put SqlTypeFamily.TIMESTAMP instead. My mistake.

PR #3042, opened 2023-01-23
Calcite 5418 3

jamesstarr:
Rework RexSubquery to contain the Correlate Id and for the correlate id to be used in subquery expansions.  This fixes bugs with nested correlated subqueries when expand is false.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
@@ -46,6 +46,7 @@ import org.apache.calcite.util.ImmutableBitSet;
 import org.apache.calcite.util.Pair;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 
 import org.immutables.value.Value;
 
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/apache/calcite/rel/logical/LogicalFilter.java
```diff
@@ -47,7 +47,7 @@
  * not targeted at any particular engine or calling convention.
  */
 public final class LogicalFilter extends Filter {
-  private final ImmutableSet<CorrelationId> variablesSet;
+  private final Set<CorrelationId> variablesSet;
```
zinking, line 50: is it possible to keep the immutableSet here? generally Immutable should be preferred.

jamesstarr: Is there any other changes you would like me to make for this?  The set is currently always immutable, just the signature of the member variable is not.

mihaibudiu:
Can we revive this PR?

jamesstarr:
@mihaibudiu
> Can we revive this PR?
I will look at rebasing it next week if I can get a review.

mihaibudiu:
I don't know this part of the code that well, but I will try to supply a review.
This PR is also quite large.
But it looks like it's a bug worth fixing.

PR #3049, opened 2023-01-27
[CALCITE-5501] SqlToRelConverterTest.checkActualAndReferenceFiles fai…

asolimando:
…ls intermittently in Jenkins CI

The PR addresses a race condition over the shared file `SqlToRelConverterTest.xml` by enforcing a lock between `SqlToRelConverterTest` and `SqlToRelConverterExtendedTest` classes.

The tests within the two classes, and with respect to any other test class, remain parallel, so no test runtime drop is expected.


clesaec (approved):
LGTM, didn't know about JUnit5 ResourceLock :+1:

PR #3050, opened 2023-01-27
[CALCITE-5503] Address issue in CheapestPlanReplacer where repeated nodes in a DAG are not reused.

mosche:
When using `CheapestPlanReplacer`, semantics of a `RelNode` tree change if it contains the same node multiple times in case this node has inputs itself that have to be replaced. During replacement such nodes get copied on each occurrence.

Instead `CheapestPlanReplacer` should memoize previously visited nodes and emit the same result again in the that case.

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/volcano/RelSubset.java
+++ b/core/src/main/java/org/apache/calcite/plan/volcano/RelSubset.java
@@ -51,6 +51,7 @@ import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/plan/volcano/RelSubset.java
```diff
@@ -737,6 +745,7 @@ public RelNode visit(
         planner.provenanceMap.put(
             p, new VolcanoPlanner.DirectProvenance(pOld));
       }
+      visited.put(pId, p); // memoize result for pId
```
clesaec, line 748: As it's a recursive method (line 739), it may be not so simple ? at least, should have a unit test with recursion, at most; move the content of method in new private one; and keep memoize in public one :

```java
 public RelNode visit( RelNode p,   int ordinal,  @Nullable RelNode parent) {
      final int pId = p.getId();
      if (visited.containsKey(pId)) {
        // return memoized result if available
        return visited.get(pId);
     }
     RelNode result = visit_inside( p,  ordinal, parent);
     visited.put(pId, result);
}

 private RelNode visit_inside( RelNode p,   int ordinal,  @Nullable RelNode parent) {
   ...
  RelNode input = visit_inside(oldInput, i, p);
  ...
}
```

clesaec (approved):
LGTM

mosche:
Kind ping @julianhyde

libenchao (approved):
+1, merging.

PR #3053, opened 2023-01-29
[CALCITE-5486] SubQuery not support HAVING alias in where condition

JiajunBernoulli:


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4474,6 +4474,21 @@ protected void validateWhereOrOn(
       SqlNode condition,
       String clause) {
     validateNoAggs(aggOrOverOrGroupFinder, condition, clause);
+    //  SqlSelect need to expand alias
+    condition.accept(new SqlShuttle() {
+      @Override public @Nullable SqlNode visit(SqlCall call) {
+        call.getOperandList()
+            .stream()
+            .filter(node -> node != null)
```
zoudan, line 4482: `node -> node != null` -> `Objects::nonNull` seems better, WDYT?

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -5709,6 +5709,19 @@ public boolean isBangEqualAllowed() {
         .withConformance(lenient).ok();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5486">[CALCITE-5486]
+   * SubQuery not support HAVING alias in where condition</a>. */
+  @Test void testHavingAliasInCondition() {
```
zoudan, line 5715: Should we also add a test in `SqlToRelConverterTest` to ensure we get the right plan?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
JiajunBernoulli: Good advice.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
JiajunBernoulli: Yes, done.

zoudan:
LGTM

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4474,6 +4474,21 @@ protected void validateWhereOrOn(
...
+    //  SqlSelect need to expand alias
+    condition.accept(new SqlShuttle() {
+      @Override public @Nullable SqlNode visit(SqlCall call) {
+        call.getOperandList()
+            .stream()
+            .filter(Objects::nonNull)
+            .forEach(node -> node.accept(this));
+        if (call.getKind() == SqlKind.SELECT) {
+          SqlSelect select = (SqlSelect) call;
+          validateHavingClause(select);
```
libenchao, line 4486: `validateWhereOrOn` is used not only to validate `where` clause, is your change also applies to `on` clause, is there any test covers this?

JiajunBernoulli: Thanks for your review, I added one test.

libenchao:
@JiajunBernoulli I've left some comments in the jira, and marked this PR "discussion in jira" label.

PR #3054, opened 2023-01-29
(do not check in)

julianhyde:


babel/src/test/resources/sql/big-query.iq
```diff
@@ -659,6 +659,8 @@ SELECT LENGTH("hello") as length;
 #####################################################################
 # DATE
 #
+# 0. DATE(string)
+#   Shorthand for 'CAST(string AS DATE)'
```
wnob, line 663: These overloads are actually ambiguous due to BigQuery's type system. String literals are implicitly converted to date/time literals if they have the correct format. Here are some queries I tried to illuminate the situation:

This is actually an example of the `DATE(timestamp_expression[, time_zone])` overload. The first argument must be a string literal:
```
SELECT DATE('2023-01-30', 'America/Los_Angeles');

=> 2023-01-29
```

If we get rid of the time zone argument, we can operate on a non-literal string:
```
SELECT DATE(SUBSTRING('x2023-01-30', 2));

=> 2023-01-30
```

However, if we try a non-literal string with the time zone, it's an error:
```
SELECT DATE(SUBSTRING('x2023-01-30', 2), 'America/Los_Angeles');

=> No matching signature for function DATE for argument types: STRING, STRING. Supported signatures: DATE(TIMESTAMP, [STRING]); DATE(DATETIME); DATE(INT64, INT64, INT64); DATE(DATE); DATE(STRING) at [1:8]
```

So, it's true that there is a legitimate `DATE(STRING)` function that I steamrolled because it's not documented, but in general there is some ambiguity in GoogleSQL about what is a string vs. a timestamp or datetime or date literal (a string can be implicitly coerced into any of these types). I think we should proceed with these changes and file a bug to address this implicit coercion throughout the BQ function library later.

wnob: Filed https://issues.apache.org/jira/browse/CALCITE-5509

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2650,6 +2696,200 @@ public static int unixDate(int v) {
...
+    // UTC ("no offset").
+    try {
+      return OffsetDateTime
+          .parse(expression, BIG_QUERY_TIMESTAMP_LITERAL_FORMATTER);
+    } catch (DateTimeParseException e) {
+      try {
+        if (IS_JDK_8) {
+          // JDK 8 has a bug that prevents matching offsets like "+00" but can
+          // match "+00:00".
+          if (expression.endsWith("+00")) {
```
wnob, line 2800: This should probably be a regex: `[-+][0-9][0-9]`.

babel/src/test/resources/sql/big-query.iq
julianhyde: Calcite also allows implicit conversions. But when choosing between several overloads of the same function, it should choose the one with the fewest conversions. I'll respond in 5509.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
julianhyde: Agreed. Or `[-+][0-9]+`, to allow `+0` as well as `+00`?

olivrlee: Would it be `[-+][0-9]?[0-9]` so we prevent `+000`?

julianhyde: I think this is a case for "be liberal in what you accept".

julianhyde: @olivrlee You (and Will) were right. No point matching `+000` because when we translate to `+000:00` it would be invalid.

PR #3055, opened 2023-01-30
[CALCITE-5507] HAVING alias failed when aggregate function in condition

JiajunBernoulli:


NobiGo (commented):
I think we should validate the SQL result in .iq file.

JiajunBernoulli:
> I think we should validate the SQL result in .iq file.

Make sense, but I don't know how to change SqlConformanceEnum in [agg.iq](https://github.com/apache/calcite/blob/main/core/src/test/resources/sql/agg.iq).

Would anyone tell me, thanks~

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -5673,6 +5673,21 @@ public boolean isBangEqualAllowed() {
...
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5507">[CALCITE-5507]
+   * HAVING alias failed when aggregate function in condition</a>. */
+  @Test void testAggregateFunAndAliasInHaving() {
+    final SqlConformanceEnum lenient = SqlConformanceEnum.LENIENT;
+    final SqlConformanceEnum strict = SqlConformanceEnum.STRICT_2003;
+
+    sql("select count(empno) as e from emp having ^e^ > 10 and count(empno) > 10 ")
```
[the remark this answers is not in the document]
JiajunBernoulli: `^` is necessary , because [SqlTests](https://github.com/apache/calcite/blob/ae228f64347cc7620d28eff04f5869583c6bf9d5/testkit/src/main/java/org/apache/calcite/sql/test/SqlTests.java#L238) will check the position. 

If we delete `^`, calcite will throw AssertionError:
```
java.lang.AssertionError: Actual error had a position, but expected error did not. Add error position carets to sql:
select count(empno) as e from emp having ^e^ > 10 and count(empno) > 10 
```

NobiGo:
Hi @JiajunBernoulli 
I know the SET command works on the Prepare class. 
For example:
!set expand false 
works on the THREAD_EXPAND. 
Maybe you can read the code to add this SqlConformanceEnum. 
Or maybe we can validate the RelNode in the SqlToRelConverter?

JiajunBernoulli:
@NobiGo , I added two unit tests in SqlToRelConverter, would you please review it again?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -6888,6 +6891,31 @@ static class ExtendedExpander extends Expander {
...
       return super.visit(literal);
     }
+
+    /**
+     * Returns whether a given node contains a {@link SqlIdentifier}.
+     *
+     * @param parent a SqlNode
+     * @param children a SqlIdentifier
+     */
+    private boolean containIdentifier(SqlNode parent, SqlIdentifier children) {
```
zoudan, line 6901: `containIdentifier` -> `containsIdentifier`?

NobiGo (commented):
LGTM

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
JiajunBernoulli: Good catch, thanks for your review.

zoudan:
LGTM

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -6888,6 +6891,31 @@ static class ExtendedExpander extends Expander {
...
       return super.visit(literal);
     }
+
+    /**
+     * Returns whether a given node contains a {@link SqlIdentifier}.
+     *
+     * @param parent a SqlNode
+     * @param children a SqlIdentifier
+     */
+    private boolean containsIdentifier(SqlNode parent, SqlIdentifier children) {
```
libenchao, line 6901: There is no need to use plural for second parameter. Further more, we even do not need to name them 'parent, child', it's not helpful for readability in my opinion.   
How about name it `containsIdentifier(SqlNode sqlNode, SqlIdentifier target)`?

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -385,6 +385,23 @@ public static void checkActualAndReferenceFiles() {
...
+    sql("select count(empno) as e\n"
+        + "from emp\n"
+        + "having e > 10 and count(empno) > 10")
+        .withConformance(SqlConformanceEnum.LENIENT).ok();
+  }
+
+  @Test void testAggregateFunAndAliasInHaving2() {
+    sql("select count(empno) as e\n"
+        + "from emp\n"
+        + "having e > 10 or count(empno) < 5")
```
libenchao, line 401: I'm not sure if we need to have a different test which has only a minor difference with `testAggregateFunAndAliasInHaving1` (from `AND` to `OR`)

libenchao (commented):
@JiajunBernoulli Thanks for your PR, it looks good generally, I only left some minor comments.

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
JiajunBernoulli: I want to show it because one is reduced and another is not reduced.
- `testAggregateFunAndAliasInHaving1` 
`LogicalFilter(condition=[>($0, 10)])` has no `and`.
- `testAggregateFunAndAliasInHaving2`
`LogicalFilter(condition=[SEARCH($0, Sarg[(-∞..5), (10..+∞)])])` has `or`.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
JiajunBernoulli: Thanks, I renamed them.

libenchao (approved):
+1, merging.

PR #3057, opened 2023-02-02
[CALCITE-5510] Support sort by ordinal when sort by column is an expression in RelToSqlConverter

peng128:
Keep the output sort by column to ordinary when sort by ordinal in dialect is true and the column is an expression.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -1816,6 +1816,13 @@ public abstract class SqlImplementor {
         return true;
       }
 
+      if (rel instanceof Project
+          && clauses.contains(Clause.ORDER_BY)
+          && dialect.getConformance().isSortByOrdinal()) {
+        // Cannot merge a Project that contains sort by ordinal under it.
+        return hasSortByOrdinal();
+      }
+
       if (rel instanceof Aggregate) {
         final Aggregate agg = (Aggregate) rel;
         final boolean hasNestedAgg =
@@ -1833,6 +1840,29 @@ public abstract class SqlImplementor {
         }
       }
 
+      return false;
+    }
+
+    /**
+     * Return whether the current {@link SqlNode} in {@link Result} contains sort by column
+     * in ordinal format.
+     */
+    private boolean hasSortByOrdinal(@UnknownInitialization Result this) {
+      if (node instanceof SqlSelect) {
+        final SqlNodeList orderList = ((SqlSelect) node).getOrderList();
+        if (orderList == null) {
+          return false;
+        }
+        for (SqlNode sqlNode : orderList) {
+          if (sqlNode instanceof SqlBasicCall) {
+            for (SqlNode operand : ((SqlBasicCall) sqlNode).getOperandList()) {
+              if (operand instanceof SqlNumericLiteral) {
+                return true;
+              }
+            }
+          }
+        }
+      }
       return false;
     }
 
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -1122,7 +1122,7 @@ public abstract class SqlImplementor {
         final boolean first =
             field.nullDirection == RelFieldCollation.NullDirection.FIRST;
         SqlNode nullDirectionNode =
-            dialect.emulateNullDirection(orderField(field.getFieldIndex()),
+            dialect.emulateNullDirection(field(field.getFieldIndex()),
                 first, field.direction.isDescending());
         if (nullDirectionNode != null) {
           orderByList.add(nullDirectionNode);
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -1854,11 +1854,12 @@ public abstract class SqlImplementor {
           return false;
         }
         for (SqlNode sqlNode : orderList) {
-          if (sqlNode instanceof SqlBasicCall) {
-            for (SqlNode operand : ((SqlBasicCall) sqlNode).getOperandList()) {
-              if (operand instanceof SqlNumericLiteral) {
-                return true;
-              }
+          if (!(sqlNode instanceof SqlBasicCall)) {
+            return false;
+          }
+          for (SqlNode operand : ((SqlBasicCall) sqlNode).getOperandList()) {
+            if (operand instanceof SqlNumericLiteral) {
+              return true;
             }
           }
         }
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -615,6 +615,13 @@ public abstract class SqlImplementor {
         final String strValue = ((SqlNumericLiteral) node).toValue();
         return SqlLiteral.createCharString(strValue, node.getParserPosition());
       }
+      if (node instanceof SqlCall
+          && dialect.getConformance().isSortByOrdinal()) {
+        // If the filed is expression and sort by ordinal is set in dialect,
+        // convert it to ordinal.
+        return SqlLiteral.createExactNumeric(
+            Integer.toString(ordinal + 1), SqlParserPos.ZERO);
+      }
       return node;
     }
 
@@ -1115,7 +1122,7 @@ public abstract class SqlImplementor {
         final boolean first =
             field.nullDirection == RelFieldCollation.NullDirection.FIRST;
         SqlNode nullDirectionNode =
-            dialect.emulateNullDirection(field(field.getFieldIndex()),
+            dialect.emulateNullDirection(orderField(field.getFieldIndex()),
                 first, field.direction.isDescending());
         if (nullDirectionNode != null) {
           orderByList.add(nullDirectionNode);
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -617,7 +617,7 @@ public abstract class SqlImplementor {
       }
       if (node instanceof SqlCall
           && dialect.getConformance().isSortByOrdinal()) {
-        // If the filed is expression and sort by ordinal is set in dialect,
+        // If the field is expression and sort by ordinal is set in dialect,
         // convert it to ordinal.
         return SqlLiteral.createExactNumeric(
             Integer.toString(ordinal + 1), SqlParserPos.ZERO);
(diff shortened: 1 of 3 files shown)


zoudan:
LGTM

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -759,17 +760,17 @@ private static String toSql(RelNode root, SqlDialect dialect,
     final String expected = "SELECT \"product_class_id\", COUNT(*) AS \"C\"\n"
         + "FROM \"foodmart\".\"product\"\n"
         + "GROUP BY ROLLUP(\"product_class_id\")\n"
-        + "ORDER BY \"product_class_id\", COUNT(*)";
+        + "ORDER BY \"product_class_id\", 2";
     final String expectedMysql = "SELECT `product_class_id`, COUNT(*) AS `C`\n"
         + "FROM `foodmart`.`product`\n"
```
chunweilei, line 765: COUNT(*) is replaced by 2 rather than 1. So the ordinal starts from 1?

peng128: Yes, I think the ordinal is the position of column in select list, it starts from 1.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -809,14 +810,14 @@ private static String toSql(RelNode root, SqlDialect dialect,
...
-        + "ORDER BY \"product_class_id\", \"brand_name\", COUNT(*)";
+        + "ORDER BY \"product_class_id\", \"brand_name\", 3";
     final String expectedMysql = "SELECT `product_class_id`, `brand_name`,"
         + " COUNT(*) AS `C`\n"
         + "FROM `foodmart`.`product`\n"
         + "GROUP BY `product_class_id`, `brand_name` WITH ROLLUP\n"
         + "ORDER BY `product_class_id` IS NULL, `product_class_id`,"
         + " `brand_name` IS NULL, `brand_name`,"
-        + " COUNT(*) IS NULL, COUNT(*)";
+        + " 3 IS NULL, 3";
```
libenchao, line 820: Is `order by 3 is null` valid here? In [Calcite](https://github.com/apache/calcite/blob/56e9deff51942fe7673096aaf5f50544f47f1411/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java#L6677-L6680), we'll take this not a 'field reference'. I'm not sure how other databases behaves in this case.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -615,6 +615,13 @@ public SqlNode orderField(int ordinal) {
         final String strValue = ((SqlNumericLiteral) node).toValue();
         return SqlLiteral.createCharString(strValue, node.getParserPosition());
       }
+      if (node instanceof SqlCall
+          && dialect.getConformance().isSortByOrdinal()) {
+        // If the filed is expression and sort by ordinal is set in dialect,
```
libenchao, line 620: `filed` -> `field`

libenchao (commented):
@peng128 thanks for your PR, I've left some comments.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
peng128: I think it valid here.
I try to make a test to compare the relNode compare '3 is null' sql with the original SQL and the 'count(*) is null' sql.
Using default SqlParser config.

1. For original sql:
```
select "product_class_id", "brand_name",count(*) as c
from "product"
group by rollup("product_class_id", "brand_name")
order by 1, 2, 3
```
relNode is
```
LogicalSort(sort0=[$0], sort1=[$1], sort2=[$2], dir0=[ASC], dir1=[ASC], dir2=[ASC])
  LogicalAggregate(group=[{0, 1}], groups=[[{0, 1}, {0}, {}]], C=[COUNT()])
    LogicalProject(product_class_id=[$0], brand_name=[$2])
      JdbcTableScan(table=[[foodmart, product]])
```


2. For 'count(*) is null' sql:
```
select "product_class_id", "brand_name", count(*) as c
from "product"
group by rollup("product_class_id", "brand_name")
order by "product_class_id", "brand_name", count(*) is null, count(*)
```
relNode is
```
LogicalProject(product_class_id=[$0], brand_name=[$1], C=[$2])
  LogicalSort(sort0=[$0], sort1=[$1], sort2=[$3], sort3=[$2], dir0=[ASC], dir1=[ASC], dir2=[ASC], dir3=[ASC])
    LogicalProject(product_class_id=[$0], brand_name=[$1], C=[$2], EXPR$3=[false])
      LogicalAggregate(group=[{0, 1}], groups=[[{0, 1}, {0}, {}]], C=[COUNT()])
        LogicalProject(product_class_id=[$0], brand_name=[$2])
          JdbcTableScan(table=[[foodmart, product]])
```


3. For '3 is null' sql:
```
select "product_class_id", "brand_name", count(*) as c
from "product"
group by rollup("product_class_id", "brand_name")
order by 1, 2, 3 is null, 3
```
relNode is
```
LogicalProject(product_class_id=[$0], brand_name=[$1], C=[$2])
  LogicalSort(sort0=[$0], sort1=[$1], sort2=[$3], sort3=[$2], dir0=[ASC], dir1=[ASC], dir2=[ASC], dir3=[ASC])
    LogicalProject(product_class_id=[$0], brand_name=[$1], C=[$2], EXPR$3=[false])
      LogicalAggregate(group=[{0, 1}], groups=[[{0, 1}, {0}, {}]], C=[COUNT()])
        LogicalProject(product_class_id=[$0], brand_name=[$2])
          JdbcTableScan(table=[[foodmart, product]])
```

The relNode of '3 is null' sql and 'count(*) is null' sql is almost some. Is my understanding correct?
So I think it's fine here. And I test '3 is null' sql in MySQL, it works.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
peng128: Fixed it. Thanks~

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
libenchao: No, what I mean it that we only takes pure literal as a field reference in Calcite, if it's an expression, it won't be taken as a field reference. Your second case shows what I described above.

To be more precise, `3 is null` equals `false`, not `count(*) is null` in this case.

I'm not saying Calcite is definitely correct for current behavior. We need to investigate what other databases behave when the literal is in a expression, and let Calcite be consistent with the majority of them.

If we prefer to not take literal in expressions as a field reference, then your change now would make it incorrect.

peng128: I got it. We shouldn't take literal in experssion as field reference.
How about keep original expression when we need to sort null values in order by?
For example, the output sql will be 
```
select "product_class_id", "brand_name", count(*) as c
from "product"
group by rollup("product_class_id", "brand_name")
order by 1, 2, count(*) is null, 3
```

libenchao: This sounds good to me.

peng128: I fixed it. Please review again, thanks.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1756,6 +1757,35 @@ private SqlDialect nonOrdinalDialect() {
         + "FROM \"foodmart\".\"employee\"");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5510">[CALCITE-5510]
+   * RelToSqlConverter don't support sort by ordinal when sort by column is an expression</a>.
+   */
+  @Test void testOrderByOrdinalWithExpression() {
+    String query = "select \"product_id\", count(*) as \"c\"\n"
```
libenchao, line 1765: `final`

libenchao (commented):
Looks good to me now, with one minor comment. After fixing that, I'll merge this PR.

libenchao:
reopened, because slow tests fail due to this change: https://github.com/apache/calcite/actions/runs/4223301074/jobs/7332876394

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1809,6 +1816,13 @@ private boolean needNewSubQuery(
         return true;
       }
 
+      if (rel instanceof Project
+          && clauses.contains(Clause.ORDER_BY)
+          && dialect.getConformance().isSortByOrdinal()) {
+        // Cannot merge a Project that contains sort by ordinal under it.
+        return hasSortByOrdinal();
+      }
```
libenchao, line 1824: The [last version we merged](https://github.com/apache/calcite/commit/0e8d6b093d26c291361bfd63ecb14828d1e33bc6) does not have this logic, could you help me to understand why do you need to add this logic now?

peng128: We found test case LatticeTest#testSum is failed in slow test.
These code is to fix it.

The reason of test case failure is that the project will be merged when the sub-relnode is sort(change by [CALCITE-4901](https://issues.apache.org/jira/browse/CALCITE-4901)).

For example,
If the relNode is
```
JdbcProject(C=[$1])
  JdbcSort(sort0=[$1], dir0=[DESC], fetch=[1])
    JdbcAggregate(group=[{0}], C=[$SUM0($7)])
      JdbcTableScan(table=[[foodmart, sales_fact_1997]])
```
The result after merge project is
```
SELECT SUM(`sales_fact_1997`.`unit_sales`) AS `C`
FROM `foodmart`.`sales_fact_1997` AS `sales_fact_1997`
GROUP BY `sales_fact_1997`.`product_id`
ORDER BY 2 DESC
```
it's a incorrect sql, we want
```
SELECT `C` FROM (
SELECT `sales_fact_1997`.`product_id`, SUM(`sales_fact_1997`.`unit_sales`) AS `C`
FROM `foodmart`.`sales_fact_1997` AS `sales_fact_1997`
GROUP BY `sales_fact_1997`.`product_id`
ORDER BY 2 DESC) t
```
So, these code to find out whether there is sort by original under project. 
And don't merge project when there is sort by original under project.

libenchao: Since we agreed that the slow tests failure is not caused by this PR, we'd better handle this in CALCITE-5537.

peng128: There are three test cases failure.
1. LatticeTest#testSum
2. LatticeSuggesterTest#testFoodMartAll
3. LatticeSuggesterTest#testFoodMartAllEvolve

Case 1 is caused by this PR. Maybe we could fix in this PR?
Case 2 & 3 is not caused by this PR. I think we could fix case 2 & 3 in [CALCITE-5537](https://issues.apache.org/jira/browse/CALCITE-5537).

BTW, these code just fix case 1.

libenchao: Thanks for the clarification, I agree with you. I'll review the PR in the coming days.

libenchao (approved):
+1, merging. Thanks @peng128 for your contribution!

PR #3059, opened 2023-02-03
[CALCITE-5514] Add a public RelJson#toRex() instance method

tjbanghart:
Follow up to [CALCITE-5349](https://issues.apache.org/jira/browse/CALCITE-5349) so it is easier to deserialize JSON with a customized `RelJson` instance.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
+++ b/core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
@@ -637,6 +637,20 @@ public class RelJson {
     return map;
   }
 
+  /**
+   * Translates a JSON expression into a RexNode,
+   * using a given {@link InputTranslator} to transform JSON objects that
+   * represent input references into RexNodes.
+   *
+   * @param cluster The optimization environment
+   * @param o JSON object
+   * @return the transformed RexNode
+   */
+  public @PolyNull RexNode toRex(RelOptCluster cluster, @PolyNull Object o) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -857,7 +871,10 @@ private List<RexNode> toRexList(RelInput relInput, List operands) {
    * @param translator Input translator
    * @param o JSON object
    * @return the transformed RexNode
+   *
+   * @deprecated Use {@link #toRex(RelOptCluster, Object)}
```
zoudan, line 875: should we keep them both?

tjbanghart: I don't think having this static method is helpful and could cause confusion during deserialization. With the changes in https://github.com/apache/calcite/commit/02b67eae841c093d966bdb6371381be234c275d6 I think it is too inflexible to keep around.

If I serialize a `RexNode` with a `RelJson` instance with specific library operators set, `readExpression` would throw an error on any non-standard function. Alternatively, we could add a nullable `List<SqlLibrary>` argument to `readExpression` but at that point one might as well create a `RelJson` and set the appropriate libraries. 

In other words, having a single `toRex` instance method helps prevent context mismatch during serialization/deserialization. I am not sure we gain anything by not deprecating the static `readExpression`.

PR #3061, opened 2023-02-06
[CALCITE-5515] Add keyspace parameter to CassandraSchema and CassandraTable

tindzk:
Currently, it is not possible to construct a `CassandraSchema` instance if the keyspace is not set in the CQL session.

Queries generated by `CassandraTable` do not include a keyspace and will fail if the CQL session does not have a keyspace set.


dssysolyatin:
LGTM

PR #3062, opened 2023-02-09
适应presto分页功能

traveler-guoke:
presto分页语句为:
SELECT  * FROM   db1.tb1 offset 100 limit 20;
原版本默认先拼limit再拼offset，不符合presto语法


julianhyde:
Can you please log a jira case and add a test case? Maybe add to `RelToSqlConverterTest.testFetchOffset`.

julianhyde:
Translation:

The presto paging statement is:
SELECT * FROM db1.tb1 offset 100 limit 20;
The original version defaults to spelling limit first and then offset, which does not conform to presto syntax

julianhyde:
I logged https://issues.apache.org/jira/browse/CALCITE-5524. But I discovered that it is already fixed. You just need to use `PrestoSqlDialect`.

PR #3063, opened 2023-02-09
[CALCITE-5521] Remove redundant rowtype double checks in RelSubset#add

asdfgh19:
jira: [CALCITE-5521](https://issues.apache.org/jira/browse/CALCITE-5521)
Double checking relset rowtype and input parameter relNode rowtype may be unnecessary, removing it might make VolcanoPlanner run slightly faster


zabetak (approved):
LGTM, will merge when all checks pass!

PR #3064, opened 2023-02-11
Site: Add Jiajun Xie as committer

JiajunBernoulli:


site/_data/contributors.yml
```diff
@@ -142,6 +142,11 @@
   githubId: jcamachor
   org: Hortonworks
   role: PMC
+- name: Jiajun Xie
```
libenchao, line 145: Would you mind to also add `pronouns` section like https://github.com/apache/calcite/blob/0dcf3ab4185647243255af29f0128b52737c1bf2/site/_data/contributors.yml#L51 ?

libenchao (commented):
Looks great, I left one nit comment.

site/_data/contributors.yml
JiajunBernoulli: Thanks for your review.

Most people didn't add `pronouns`, I prefer to be the same as most people.

libenchao (approved):
+1

PR #3065, opened 2023-02-11
[CALCITE-3870] Change the default value of SqlToRelConverter.Config.expand from true to false

julianhyde:
This is a breaking change. Clients should either call Config.withExpand(true), to keep the old behavior, or ensure that SubQueryRemoveRule is enabled during the planning phase.

We believe that the new behavior - expansion during the planning phase rather than during sql-to-rel phase - is superior because it handles ANY and ALL sub-queries and complex correlated sub-queries.


libenchao (commented):
The PR looks good to me, there is one checker failure, otherwise, I'm +1.

zabetak (approved):
LGTM, left some comments but all are minor treat at your discretion!

libenchao:
It seems that there are still some places which violate the checker rule added by https://github.com/apache/calcite/pull/3065/commits/14ce39829e8b1c4b5865f4c8778f1f0b7890250e

PR #3066, opened 2023-02-12
[CALCITE-5518] RelToSql converter generates invalid order of ROLLUP f…

JiajunBernoulli:
…ields


JiajunBernoulli:
I will merge it tomorrow if there are no other suggestions.

PR #3069, opened 2023-02-16
[WIP] Error message type aliasing - built out aliasTypeMap and in the…

olivrlee:
… middle of testing extraction

### DIFF
--- a/babel/src/test/resources/sql/big-query.iq
+++ b/babel/src/test/resources/sql/big-query.iq
@@ -1937,7 +1937,7 @@ SELECT
 SELECT
   TIMESTAMP "2010-07-07 10:20:00" AS later_timestamp,
   TIMESTAMP "2008-12-25 15:30:00" AS earlier_timestamp,
-  TIMESTAMP_DIFF(TIMESTAMP "2010-07-07 10:20:00", TIMESTAMP "2008-12-25 15:30:00", HOUR) AS hours;
+  TIMESTAMP_DIFF(1, TIMESTAMP "2008-12-25 15:30:00", HOUR) AS hours;
 +---------------------+---------------------+-------+
 | later_timestamp     | earlier_timestamp   | hours |
 +---------------------+---------------------+-------+
--- a/core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
+++ b/core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
@@ -16,8 +16,13 @@
  */
 package org.apache.calcite.prepare;
 
+import com.google.common.collect.BiMap;
+
+import com.google.common.collect.HashBiMap;
+
 import org.apache.calcite.config.CalciteConnectionConfig;
 import org.apache.calcite.jdbc.CalciteSchema;
+import org.apache.calcite.jdbc.CalciteSchema.TypeEntry;
 import org.apache.calcite.jdbc.JavaTypeFactoryImpl;
 import org.apache.calcite.linq4j.function.Hints;
 import org.apache.calcite.model.ModelHandler;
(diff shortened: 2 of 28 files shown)


babel/src/test/resources/sql/big-query.iq
```diff
@@ -1937,7 +1937,7 @@ SELECT
 SELECT
   TIMESTAMP "2010-07-07 10:20:00" AS later_timestamp,
   TIMESTAMP "2008-12-25 15:30:00" AS earlier_timestamp,
-  TIMESTAMP_DIFF(TIMESTAMP "2010-07-07 10:20:00", TIMESTAMP "2008-12-25 15:30:00", HOUR) AS hours;
+  TIMESTAMP_DIFF(1, TIMESTAMP "2008-12-25 15:30:00", HOUR) AS hours;
```
olivrlee, line 1940: Using this babel quidem test to test error messages because CalciteSqlOperatorTest uses MockCatalogReader

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -172,8 +195,14 @@ private Collection<org.apache.calcite.schema.Function> getFunctionsFrom(
     }
     return functions2;
   }
+  // public @Nullable SqlIdentifier getNameFromType(RelDataType type) {
```
olivrlee, line 198: ignore this

core/src/main/java/org/apache/calcite/sql/SqlCallBinding.java
```diff
@@ -398,10 +398,13 @@ public SqlCall permutedCall() {
...
    */
   public CalciteException newValidationSignatureError() {
+    // String type = validator.getCatalogReader().getNamedType(call.getOperator()
+    // .getNameAsId()).getSqlTypeName().getName();
     return validator.newValidationError(call,
         RESOURCE.canNotApplyOp2Type(getOperator().getName(),
             call.getCallSignature(validator, scope),
-            getOperator().getAllowedSignatures()));
+            // getOperator().getAllowedSignatures()));
+            getOperator().getAllowedSignaturesUsingValidator(validator)));
```
olivrlee, line 407: the idea is that this function will pass through validator which we'll use to check for all alias types for this `newValidationSignatureError`

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -108,6 +114,23 @@ protected CalciteCatalogReader(CalciteSchema rootSchema,
             : new LinkedHashSet<>(schemaPaths));
     this.typeFactory = typeFactory;
     this.config = config;
+    this.aliasTypeMap = HashBiMap.create();
+    populateAliasTypeMap(this.aliasTypeMap, rootSchema);
+  }
+
+
+  public static void populateAliasTypeMap(BiMap<TypeEntry, String> aliasTypeMap,
```
olivrlee, line 122: this would be the populating of the type aliases from the schema, it currently doesn't handle non-unique type names yet which needs fixing

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorUtil.java
```diff
@@ -1039,6 +1046,25 @@ public static ImmutableList<ImmutableBitSet> cube(
...
+  //   // SqlIdentifier target = type.;
+  //   Queue<CalciteSchema> queue = new ArrayDeque<>();
+  //   queue.add(rootSchema);
+  //   while(!queue.isEmpty()){
+  //     CalciteSchema currentSchema = queue.remove();
+  //     if currentSchema.getType
+  //   }
+  //   return null;
+  // }
+  public static String getAliasedTypeName(SqlValidator validator, String typeName){
```
olivrlee, line 1060: this is another important function where given a Calcite type we return the dialects reverse aliased type 
(i.e. calcite TIMESTAMP -> bigquery DATETIME)

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
olivrlee: I'm a bit unsure on whats the types I should be indexing in and out so that we have most versatility when retrieving (I'd imagine doing String <-> String we lose a lot of info)

PR #3070, opened 2023-02-16
[CALCITE-5532] CompositeOperandTypeChecker should check operands without type coercion first

dssysolyatin:


core/src/main/java/org/apache/calcite/sql/type/CompositeOperandTypeChecker.java
```diff
@@ -383,7 +383,17 @@ private boolean checkWithoutTypeCoercion(SqlCallBinding callBinding) {
     if (!callBinding.isTypeCoercionEnabled()) {
       return false;
     }
+    SqlCallBinding sqlBindingWithoutTypeCoercion =
```
mihaibudiu, line 386: why do you need to create a new call binding?
`callBinding` does not have type coercion enabled

dssysolyatin: @mihaibudiu That is not true. Even if you look at the if statement above:
```
if (!callBinding.isTypeCoercionEnabled()) {
    return false;
}
```

if callBinding.isTypeCoercionEnabled() were false, as you suggested, then checkWithoutTypeCoercion() would always return false.

mihaibudiu (approved):

PR #3073, opened 2023-02-17
CompositeOperandTypeChecker's java doc uses wrong class name

zoudan:
Fix typo in CompositeOperandTypeChecker: `CompositeOperandsTypeChecking` -> `CompositeOperandTypeChecker`

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/type/CompositeOperandTypeChecker.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/CompositeOperandTypeChecker.java
@@ -46,8 +46,8 @@ import static java.util.Objects.requireNonNull;
  *
  * <blockquote>
  * <pre><code>
- * CompositeOperandsTypeChecking newCompositeRule =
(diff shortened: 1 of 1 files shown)


libenchao:
@zoudan Thanks for the PR! The changes looks good to me. I have two minor comments:

1. We usually do to not use 'Fix xxx' as the commit message, instead, we prefer to use a sentence which describes the symptom of the issue
2. The CI  fails now, it seems unrelated with current change. Just a reminder, you can use `git commit --allow-empty -m "Insert you message"` to add a empty commit to trigger the CI again. You can find these instructions in https://calcite.apache.org/develop/#contributing.

zoudan:
@libenchao Thx for reviewing, I have changed the commit message and the CI was triggered to rerun

libenchao:
@zoudan I don't think `Misspelled an extra letter 's' in CompositeOperandTypeChecker comments` is a good commit message (it's not even right). 

How about `CompositeOperandTypeChecker's java doc uses wrong class name`.

chunweilei:
> 

+1

PR #3075, opened 2023-02-18
Hotfix in RelDistributions

xuzifu666:


xuzifu666:
@libenchao have a review please

core/src/main/java/org/apache/calcite/rel/RelDistributions.java
```diff
@@ -38,19 +38,14 @@
...
+  /** The singleton distribution. */
   public static final RelDistribution SINGLETON =
       new RelDistributionImpl(RelDistribution.Type.SINGLETON, EMPTY);
 
   /** The singleton random distribution. */
   public static final RelDistribution RANDOM_DISTRIBUTED =
       new RelDistributionImpl(RelDistribution.Type.RANDOM_DISTRIBUTED, EMPTY);
 
-  /** The singleton round-robin distribution. */
-  public static final RelDistribution ROUND_ROBIN_DISTRIBUTED =
```
libenchao, line 50: Why are you proposing to remove this? (Notice that, Calcite is a framework used by many third-party projects, hence even some code has no usage in Calcite itself, we cannot remove them safely)

core/src/main/java/org/apache/calcite/rel/RelDistributions.java
```diff
@@ -38,19 +38,14 @@
 public class RelDistributions {
   public static final ImmutableIntList EMPTY = ImmutableIntList.of();
 
-  /** The singleton singleton distribution. */
```
libenchao, line 41: I guess this is not a 'typo'. You can see other comments below, the first 'singleton' means the design pattern is 'singleton', and the second 'singleton' means the 'singleton distribution'.

libenchao (commented):
@xuzifu666 Thanks for contributing to Calcite! However, I see no value in this PR, you can see my comments below. 

Looking forward to your further contributions!

core/src/main/java/org/apache/calcite/rel/RelDistributions.java
xuzifu666: ok，I got it. I would close the pr，thanks~

PR #3076, opened 2023-02-21
 added support for STRING_AGG function with collation

Kowsikraja:


chunweilei (commented):
@Kowsikraja It seems you pushed the wrong project.

zoudan:
It seems that your code is not based on the main branch of calcite, so there are so many changes in this PR

PR #3079, opened 2023-02-22
[CALCITE-5469] Add support for DATETIME_ADD, DATETIME_DIFF, DATE_ADD, DATE_DIFF.

tanclary:
This PR adds implementations for BigQuery's DATETIME_ADD, DATETIME_DIFF, DATE_ADD, and DATE_DIFF functions. These were originally two separate JIRA Issues (links below) but I am opening them in one PR because they both rely on a change I made to the TimestampDiffConvertlet to correctly handle the WEEK time unit. That fix applies to TIME_DIFF, and TIMESTAMP_DIFF as well, which have already been merged. 

DATE_ADD adds an interval to a date. [BigQuery Docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date_add)

DATETIME_ADD adds an interval to a datetime. [BigQuery Docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#datetime_add)

DATE_DIFF returns the whole number of a specified time unit between two dates. [BigQuery Docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date_diff)

DATETIME_DIFF returns the whole number of a specified time unit between two datetimes. [BigQuery Docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#datetime_diff)

[DATE_ADD/DIFF JIRA Case](https://issues.apache.org/jira/browse/CALCITE-5464)

[DATETIME_ADD/DIFF JIRA Case](https://issues.apache.org/jira/browse/CALCITE-5469)

Any comments or suggestions would be much appreciated, thanks!


core/src/main/codegen/default_config.fmpp
```diff
@@ -84,8 +84,10 @@ parser: {
     "DATA"
     "DATABASE"
     "DATE_TRUNC"
+    "DATETIME_DIFF"
     "DATETIME_INTERVAL_CODE"
     "DATETIME_INTERVAL_PRECISION"
+    "DATE_DIFF"
```
olivrlee, line 90: nit: alphabetically next to DATE_TRUNC, or move DATE_TRUNC down here too (visually _ comes before T to me, but maybe ASCII-wise DATE_ should be after DATET?)

tanclary: I was also confused about this, on lines 317-321 of this file, _ is given alphabetic precedence over letters, and then letters over _. I agree that visually it comes before. Maybe @julianhyde has thoughts?

julianhyde: I forget what the rules are. I would defer to UNIX `sort` (using `LC_ALL=C` to override my locale):

```
$ LC_ALL=C sort<<EOF
    "DATE_DIFF"
    "DATETIME_INTERVAL_CODE"
    "DATETIME_INTERVAL_PRECISION"
EOF
    "DATETIME_INTERVAL_CODE"
    "DATETIME_INTERVAL_PRECISION"
    "DATE_DIFF"
```

julianhyde:
Please be sure to reference the jira case in your first commit message. This makes it easier for reviewers juggling many commits.

babel/src/test/resources/sql/big-query.iq
```diff
@@ -1798,21 +1811,20 @@ SELECT
...
-| first_datetime             | second_datetime        | difference             |
-+----------------------------+------------------------+------------------------+
-| 2010-07-07T10:20:00        | 2008-12-25T15:30:00    | 559                    |
-+----------------------------+------------------------+------------------------+
++---------------------+---------------------+------------+
+| first_datetime      | second_datetime     | difference |
++---------------------+---------------------+------------+
+| 2010-07-07 10:20:00 | 2008-12-25 15:30:00 |        558 |
++---------------------+---------------------+------------+
+(1 row)
```
julianhyde, line 1824: difference has changed from 559 to 558. which is the correct value?

Add a comment to the test justifying the value.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -7313,6 +7361,12 @@ SqlNode JdbcFunctionCall() :
         s = span();
     }
     (
+        LOOKAHEAD(1)
+        call = DateDiffFunctionCall() {
+            name = call.getOperator().getName();
+            args = new SqlNodeList(call.getOperandList(), getPos());
+        }
```
julianhyde, line 7368: why have you added a call to `DateDiffFunctionCall` but not `DatetimeDiffFunctionCall`?

with this omission, I don't understand how the tests could pass.

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -1215,6 +1218,7 @@ public enum SqlKind {
                   FILTER, WITHIN_GROUP, IGNORE_NULLS, RESPECT_NULLS, SEPARATOR,
                   DESCENDING, CUBE, ROLLUP, GROUPING_SETS, EXTEND, LATERAL,
                   SELECT, JOIN, OTHER_FUNCTION, POSITION, CAST, TRIM, FLOOR, CEIL,
+                  DATE_ADD, TIMESTAMP_ADD, TIMESTAMP_DIFF, EXTRACT, INTERVAL,
```
julianhyde, line 1221: EXTRACT, INTERVAL and maybe others are now duplicated.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -933,6 +948,22 @@ private SqlLibraryOperators() {
           ReturnTypes.BIGINT_NULLABLE, OperandTypes.TIMESTAMP,
           SqlFunctionCategory.TIMEDATE);
 
+  /** BigQuery's "DATETIME_ADD(timestamp, interval) function; Behaves similarly
```
julianhyde, line 951: Change `BigQuery's "DATETIME_ADD(timestamp, interval) function` to `The "DATETIME_ADD(timestamp, interval) function (BigQuery)`. I don't want to assume that no other library will ever use it.

Center the javadoc in Calcite, not BigQuery. Change "; Behaves similarly to BigQuery's TIMESTAMP_ADD because in Calcite, datetime is a type alias for timestamp." to ". Like {@code TIMESTAMP_ADD}, returns a Calcite {@code TIMESTAMP} (which BigQuery calls a {@code DATETIME})."

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -933,6 +948,22 @@ private SqlLibraryOperators() {
...
+  @LibraryOperator(libraries = {BIG_QUERY})
+  public static final SqlFunction DATETIME_ADD =
+      TIMESTAMP_ADD2.withName("DATETIME_ADD");
+
+  /** BigQuery's "DATETIME_DIFF(timestamp, timestamp, timeUnit) function; Behaves
+   * similarly to BigQuery's TIMESTAMP_DIFF because in Calcite, datetime is a type
+   * alias for timestamp. Returns the whole number of timeUnit between datetime
+   * and datetime2, with the result being negative if datetime occurs before datetime2. */
+  @LibraryOperator(libraries = {BIG_QUERY})
+  public static final SqlFunction DATETIME_DIFF =
```
julianhyde, line 963: As TIMESTAMP_ADD. Also change signature to DATETIME_DIFF(timestamp, timestamp2, timeUnit)" and fix references to "datetime" and "datetime2" in the description.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -2027,13 +2050,38 @@ private static class TimestampDiffConvertlet implements SqlRexConvertlet {
...
-          cx.getTypeFactory().createTypeWithNullability(
-              cx.getTypeFactory().createSqlIntervalType(qualifier),
-              op1.getType().isNullable() || op2.getType().isNullable());
-      final RexCall rexCall = (RexCall) rexBuilder.makeCall(
-          intervalType, SqlStdOperatorTable.MINUS_DATE,
-          ImmutableList.of(op2, op1));
+
+      RelDataType intervalType;
+      RexCall rexCall;
+      /* This additional logic handles the differing definitions of 'WEEK' between BigQuery
```
julianhyde, line 2056: Use Java-style comments (`//`) rather than C-style (`/*` ... `*/`)

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -1974,27 +1982,42 @@ private static class TimestampDiffConvertlet implements SqlRexConvertlet {
       SqlIntervalQualifier qualifier;
       final RexNode op1;
       final RexNode op2;
-      if (call.operand(0).getKind() == SqlKind.INTERVAL_QUALIFIER) {
-        qualifier = call.operand(0);
-        op1 = cx.convertExpression(call.operand(1));
-        op2 = cx.convertExpression(call.operand(2));
-      } else {
+      final boolean isBigQuery = !(call.operand(0).getKind() == SqlKind.INTERVAL_QUALIFIER);
```
julianhyde, line 1985: I don't think the `isBigQuery` variable helps much. You should extend the comment, a few lines previously, about the two variants, and how one only occurs in BigQuery mode.

You should also explain why TIMESTAMP_TRUNC calls are necessary.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -2027,13 +2050,38 @@ private static class TimestampDiffConvertlet implements SqlRexConvertlet {
...
-              op1.getType().isNullable() || op2.getType().isNullable());
-      final RexCall rexCall = (RexCall) rexBuilder.makeCall(
-          intervalType, SqlStdOperatorTable.MINUS_DATE,
-          ImmutableList.of(op2, op1));
+
+      RelDataType intervalType;
+      RexCall rexCall;
+      /* This additional logic handles the differing definitions of 'WEEK' between BigQuery
+      *  and Calcite. BigQuery considers Sunday as the start of the week, so two dates whose
+      *  day difference is <7 may have a week difference of 1 if they occur on two different
```
julianhyde, line 2058: Your comments seem to imply that the difference is due to weeks starting on a different day. I don't think that's true. Calcite considers Sunday to be the start of the week too. (Monday the start of an ISOWEEK.)

I think it's just that this particular function truncates before it subtracts.

site/_docs/reference.md
```diff
@@ -2649,8 +2653,12 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | DATETIME(date)                                 | Converts *date* to a TIMESTAMP value (at midnight)
 | b | DATETIME(date, timeZone)                       | Converts *date* to a TIMESTAMP value (at midnight), in *timeZone*
 | b | DATETIME(year, month, day, hour, minute, second) | Creates a TIMESTAMP for *year*, *month*, *day*, *hour*, *minute*, *second* (all of type INTEGER)
+| b | DATETIME_ADD(datetime, interval)               | Returns the DATETIME value that occurs *interval* after *datetime*
```
julianhyde, line 2656: s/DATETIME/TIMESTAMP

julianhyde (changes requested):

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
tanclary: I will update the comment. What I was trying to say is, Calcite currently computes the number of weeks between two dates by dividing the number of days between them by 7 (and flooring the result). BigQuery however essentially checks how many Sundays are between the two dates, which in certain cases causes a discrepancy.

babel/src/test/resources/sql/big-query.iq
tanclary: I ran this test for timestamp_diff on a fresh branch cut from Calcite main and it also returns 558, although the BigQuery docs have 559 as the answer so it appears unrelated to these changes, I'm looking more into it.

core/src/main/codegen/templates/Parser.jj
tanclary: I mistakenly created DatetimeDiffFunctionCall as a SqlNode instead of a SqlCall so it was circumventing this. I'm not entirely sure how it was passing, but I've updated it to be consistent with the other function calls.

core/src/main/java/org/apache/calcite/sql/SqlKind.java
tanclary: This got messed up during rebase, sorry about that.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
tanclary: I moved and modified my comments a bit, feel free to let me know if you have other suggestions.

PR #3080, opened 2023-02-22
[CALCITE-5537] Slow test case failures in LatticeSuggesterTest

peng128:
This was caused by [CALCITE-3870]. The SqlToRelConverter.Config.expand isn't set to true explicitly in LatticeSuggesterTest when the default value of SqlToRelConverter.Config.expand change from true to false.


julianhyde:
Please use the bug summary as the commit message. Commit messages should almost never contain the word 'fix'. Feel free to force-push to change the message.

peng128:
Thanks for reviewing. I changed the commit message.

PR #3081, opened 2023-02-22
[docs] Clarifying that SQL doesn't support recursive queries yet

gunnarmorling:
Quick draft for the doc change discussed on the mailing list here: https://lists.apache.org/thread/f6zcfsdh5f9dhnknmn2rr4qbf7whycsp

Feedback welcome, happy to adjust this change as needed. E.g. does it need a Jira issue? Thanks!

### DIFF
--- a/site/_docs/algebra.md
+++ b/site/_docs/algebra.md
@@ -265,17 +265,7 @@ or `builder.field(2, 1, 1)`.
 ### Recursive Queries
 
 Warning: The current API is experimental and subject to change without notice.
-A SQL recursive query, e.g. this one that generates the sequence 1, 2, 3, ...10:
-
-{% highlight sql %}
-WITH RECURSIVE aux(i) AS (
-  VALUES (1)
-  UNION ALL
-  SELECT i+1 FROM aux WHERE i < 10
-)
-SELECT * FROM aux
-{% endhighlight %}
-
+A recursive query, e.g. this one that generates the sequence 1, 2, 3, ...10,
 can be generated using a scan on a TransientTable and a RepeatUnion:
 
 {% highlight java %}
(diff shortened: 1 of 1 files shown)


site/_docs/algebra.md
```diff
@@ -309,6 +299,19 @@ LogicalRepeatUnion(all=[true])
         LogicalTableScan(table=[[aux]])
 {% endhighlight %}
 
+Note that currently there is no support for recursive queries in the SQL layer yet.
```
zabetak, line 302: "currently" and "yet" seem a bit superfluous; remove one of them.

site/_docs/algebra.md
```diff
@@ -265,17 +265,7 @@ or `builder.field(2, 1, 1)`.
...
-{% highlight sql %}
-WITH RECURSIVE aux(i) AS (
-  VALUES (1)
-  UNION ALL
-  SELECT i+1 FROM aux WHERE i < 10
-)
-SELECT * FROM aux
-{% endhighlight %}
-
+A recursive query, e.g. this one that generates the sequence 1, 2, 3, ...10,
```
zabetak, line 268: I prefer the SQL example here as I find it easier to follow the example afterwards.

rubenada:
Thanks for the PR @gunnarmorling .

Instead of using "[docs]" as prefix for the PR title and commit message, we usually use a de-facto convention of "Site: ...". And I'd prefer to avoid contractions in there too.
So I'd propose something like: "Site: Clarifying that SQL does not support recursive queries yet" for PR title and final commit message after squashing all potential commits into a single one.

site/_docs/algebra.md
```diff
@@ -309,6 +299,19 @@ LogicalRepeatUnion(all=[true])
...
+
+{% highlight sql %}
+WITH RECURSIVE aux(i) AS (
+  VALUES (1)
+  UNION ALL
+  SELECT i+1 FROM aux WHERE i < 10
+)
+SELECT * FROM aux
+{% endhighlight %}
+
```
zabetak, line 314: Maybe we can keep the note and here just write: "The SQL example above is only for illustrative purposes (`RECURSIVE` keyword not supported)."

site/_docs/algebra.md
rubenada, line 268: nitpick: with the change, IMO the expression "this one" sounds a bit strange. Proposal:
```
A recursive query, e.g. generating the sequence 1, 2, 3, ...10,
can be produced using a scan on a TransientTable and a RepeatUnion:
```

zabetak (approved):
Minor suggestions based on personal preferences.

site/_docs/algebra.md
rubenada: UPDATE: I have checked @zabetak  comments, and I agree, I think it is clearer to keep the SQL example first. 
So I'll resolve this conversation, my comment can be disregarded in favor of @zabetak 's suggestion.

gunnarmorling:
Excellent, thanks a lot!

PR #3084, opened 2023-02-24
[CALCITE-5545] Allow for overriding SqlValidator to enable custom SqlNode validation

olivrlee:


core/src/main/java/org/apache/calcite/jdbc/Driver.java
```diff
@@ -64,6 +64,15 @@ public Driver() {
     this.prepareFactory = createPrepareFactory();
   }
 
+  private Driver(Function0<CalcitePrepare> prepareFactory) {
```
julianhyde, line 67: unify constructors. all constructors except one should call other constructors.

core/src/main/java/org/apache/calcite/jdbc/Driver.java
```diff
@@ -64,6 +64,15 @@ public Driver() {
...
+    super();
+    this.prepareFactory = prepareFactory;
+  }
+  public Driver withPrepareFactory(Function0<CalcitePrepare> prepareFactory) {
```
julianhyde, line 71: add newline and javadoc

core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
```diff
@@ -514,14 +514,35 @@ <T> CalciteSignature<T> prepare_(
...
         return prepare2_(context, query, elementType, maxRowCount,
-            catalogReader, planner);
+            catalogReader, preparingStmt);
       } catch (RelOptPlanner.CannotPlanException e) {
         exception = e;
       }
     }
     throw exception;
   }
+  protected CalcitePreparingStmt getPreparingStmt(
```
julianhyde, line 527: add newline and javadoc

core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
```diff
@@ -590,21 +611,8 @@ <T> CalciteSignature<T> prepare2_(
       Type elementType,
       long maxRowCount,
       CalciteCatalogReader catalogReader,
-      RelOptPlanner planner) {
+      CalcitePreparingStmt preparingStmt) {
```
julianhyde, line 614: Is CalcitePreparingStmt required? Or could parameter be a superclass?

core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
```diff
@@ -937,8 +944,10 @@ public <R> R perform(CalciteServerStatement statement,
         prepareContext.getRootSchema().plus(), statement);
   }
 
-  /** Holds state for the process of preparing a SQL statement. */
-  static class CalcitePreparingStmt extends Prepare
+  /** Holds state for the process of preparing a SQL statement.
+   *  Overload this class and {@link #createSqlValidator} to supply custom validation logic.
+   * */
+  public static class CalcitePreparingStmt extends Prepare
```
julianhyde, line 950: improve javadoc formatting. second sentence should be a new paragraph. use `**/` rather than `* */`.

core/src/main/java/org/apache/calcite/prepare/CalciteSqlValidator.java
```diff
@@ -22,10 +22,12 @@
 import org.apache.calcite.sql.SqlOperatorTable;
 import org.apache.calcite.sql.validate.SqlValidatorImpl;
 
-/** Validator. */
-class CalciteSqlValidator extends SqlValidatorImpl {
+/** Validator.
+ *  Public access-level to allow overriding and custom validation logic.
+ * */
```
julianhyde, line 27: Similar comments to previous javadoc. Second sentence should be a paragraph. Also a sentence.

julianhyde (changes requested):
almost there.

We also need a test that creates a custom validator. Just to make sure that this stays working.

core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
olivrlee: `createSqlValidator` is only available in `CalcitePreparingStmt` so it has to be `CalcitePreparingStmt`

core/src/main/java/org/apache/calcite/jdbc/Driver.java
olivrlee: I've unified the constructors in my last commit: https://github.com/apache/calcite/pull/3084/commits/5314d5d29e5cd6d271626d7b877125278e8292d4#diff-202586edbf28484a3a4c1498d8f61096e7a12c0a9c813260fb4282b5c19a3e96R63

But it doesn't compile due to no explicit assigning of prepareFactory. 
Do you have suggestions here?

julianhyde: You're right there's a problem. The problem is that the constructor calls `createPrepareFactory()` - an instance method - before the constructor is complete, and that's a code smell, because it doesn't allow the class to be subclassed cleanly.

Is there a way to change the design of `Driver` so that it doesn't have that code smell?

I like the idea of adding `Driver withPrepareFactory(Supplier<CalcitePrepare>)` so that people don't have to subclass Driver. But we have to do so compatibly.

olivrlee: The way that its designed seems to imply some conflicting restraints
- we need `createPrepareFactory()` to be override-able (use cases in JdbcTest.java), thus has to stay an instance method 
- we want `prepareFactory` to be `final` 
- all constructors except 1 have to call 1 constructor

Is there a way to satisfy all 3 without conflicts?

olivrlee: Pushed a change so that `private Driver(prepareFactory)` calls the `public Driver()`

core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
olivrlee: I made this public so we can access and assert in `JdbcTest.java`

olivrlee:
> almost there.
> 
> We also need a test that creates a custom validator. Just to make sure that this stays working.

Added a test in `JdbcTest.java`!

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -771,6 +771,11 @@ private void checkTableFunctionInModel(Class<?> clazz) {
     }
   }
 
+  @Test void testCustomValidator() {
+    final Driver driver = new MockDdlDriver().withPrepareFactory(MockPrepareImpl::new);
+    assertTrue(driver.prepareFactory.apply().getClass() == MockPrepareImpl.class);
```
julianhyde, line 776: Use assertThat(...getClass(), is()). Better error message if it fails.

testkit/src/main/java/org/apache/calcite/test/RelOptFixture.java
```diff
@@ -76,8 +76,10 @@
  * <p>A fixture is immutable. If you have two test cases that require a similar
  * set up (for example, the same SQL expression and set of planner rules), it is
  * safe to use the same fixture object as a starting point for both tests.
+ *
+ * <p>The class has public access-level to allow implementing custom fixtures.
```
julianhyde, line 80: I don't think this paragraph adds much. I think you should remove it. (It's not even true; the constructor remains package-private because we don't want people subclassing.)

core/src/main/java/org/apache/calcite/jdbc/Driver.java
```diff
@@ -52,7 +52,7 @@
 public class Driver extends UnregisteredDriver {
   public static final String CONNECT_STRING_PREFIX = "jdbc:calcite:";
 
-  final Function0<CalcitePrepare> prepareFactory;
+  public final Function0<CalcitePrepare> prepareFactory;
```
julianhyde, line 55: This should remain package-private; but change the type to `Supplier`. Add a method

```
public CalcitePrepare createPrepare() {
  return prepareFactory.get();
}
```

julianhyde (changes requested):

olivrlee:
Revised PR with suggested changes, I still don't have a good resolution for unifying the Driver() constructors however.

PR #3085, opened 2023-02-26
[CALCITE-5530] RelToSqlConverter[ORDER BY] generates an incorrect fie…

JiajunBernoulli:
…ld alias when 2 projection fields have the same name


core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1791,6 +1791,14 @@ private boolean needNewSubQuery(
         return false;
       }
       final Clause maxClause = Collections.max(clauses);
+
+      if (rel instanceof Project
+          && clauses.contains(Clause.ORDER_BY)
+          && dialect.getConformance().isSortByOrdinal()) {
```
libenchao, line 1797: As mentioned in the Jira, if the sort-by-ordinal is not enabled, the result is also incorrect. Could you also add a test case which shows that your fix also cover it?

libenchao (commented):
@JiajunBernoulli Thanks for your PR, I've left my comments.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
JiajunBernoulli: Thanks for your review,  I add unit test to covert it.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -879,6 +879,80 @@ private static String toSql(RelNode root, SqlDialect dialect,
...
+    // would be incorrect since the rel is sorting by the field \"EMPNO\" + 1 in which EMPNO
+    // refers to the physical column EMPNO and not the alias
+    String actualSql1 = toSql(base);
+    String expectedSql1 = ""
+        + "SELECT \"EMPNO\"\n"
+        + "FROM (SELECT UPPER(\"ENAME\") AS \"EMPNO\", \"EMPNO\" + 1 AS \"$f1\"\n"
+        + "FROM \"scott\".\"EMP\"\n"
+        + "ORDER BY 2) AS \"t0\"";
+    assertThat(actualSql1, isLinux(expectedSql1));
+    // SqlConformance#isSortByAlias is false for Spark.
```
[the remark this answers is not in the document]
JiajunBernoulli: Thank you for your careful review.

I added a new commit to solve it.

libenchao:
@JiajunBernoulli The tests seem still fail~

JiajunBernoulli:
The PR also solve [CALCITE-5656](https://issues.apache.org/jira/browse/CALCITE-5656) 

Would you please review it again? @libenchao

libenchao:
I'll do a final review

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1791,6 +1791,14 @@ private boolean needNewSubQuery(
...
+        // Cannot merge a Project that contains sort by ordinal under it.
+        return hasSortByOrdinal();
```
libenchao, line 1799: To be honest, I felt uneasy about moving this up. I'm not sure if we will miss some opportunity to add a new sub query. The changes in https://github.com/apache/calcite/pull/3057 was also merged by me, and now I think that is not perfect as well, since it also suffers from this.

For `needNewSubQuery`, we should only bail out when we returns `true`. In that way, we could safely add new conditions into this method, and all the cases are orthogonal.

JiajunBernoulli: Agree, we should `return true` avoid `return false` blocking other conditions.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -879,6 +879,78 @@ private static String toSql(RelNode root, SqlDialect dialect,
         .withPresto().ok(expectedPresto);
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5530">[CALCITE-5530]
+   * RelToSqlConverter[ORDER BY] generates an incorrect field alias
+   * when 2 projection fields have the same name</a>.
+   */
+  @Test void testSortWithAnFieldNotInTheProjectionWithASameAliasAsThatInTheProjection() {
```
libenchao, line 888:
```suggestion
  @Test void testOrderByFieldNotInTheProjectionWithASameAliasAsThatInTheProjection() {
```

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -879,6 +879,78 @@ private static String toSql(RelNode root, SqlDialect dialect,
...
+    assertThat(actualSql1, isLinux(expectedSql1));
+
+    String actualSql2 = toSql(base, nonOrdinalDialect());
+    String expectedSql2 = "SELECT UPPER(ENAME) AS EMPNO\n"
+        + "FROM scott.EMP\n"
+        + "ORDER BY EMPNO";
+    assertThat(actualSql2, isLinux(expectedSql2));
+  }
+
+  @Test void testSortWithAnExpressionNotInTheProjectionThatRefersToUnderlyingFieldWithSameAlias() {
```
libenchao, line 919:
```suggestion
  @Test void testOrderByExpressionNotInTheProjectionThatRefersToUnderlyingFieldWithSameAlias() {
```

JiajunBernoulli: done

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
JiajunBernoulli: done

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1791,6 +1791,15 @@ private boolean needNewSubQuery(
         return false;
       }
       final Clause maxClause = Collections.max(clauses);
+
+      if (rel instanceof Project
```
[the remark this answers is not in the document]
JiajunBernoulli: Moved to L1819-L1825

libenchao (approved):
+1

PR #3087, opened 2023-02-28
[CALCITE-5551]: Support SELECT * Except(column) in parser

hamnis:
Sometimes it is useful to be able to select all columns in a table except one or two columns.

This is a useful feature in for instance bigquery.

I could not figure out how to make the parser deal with dialects, so I added this to the default parser.

I do not think that this is supported by all databases, so not sure where this should live, please advice.

### DIFF
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -1306,6 +1306,7 @@ SqlSelect SqlSelect() :
     final SqlLiteral keyword;
     final SqlNodeList keywordList;
     final List<SqlNode> selectList = new ArrayList<SqlNode>();
+    final SqlNodeList except;
     final SqlNode fromClause;
     final SqlNode where;
     final SqlNodeList groupBy;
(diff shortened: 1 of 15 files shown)


hamnis:
build error seems unrelated to this change

PasunuriSrinidhi:
> Sometimes it is useful to be able to select all columns in a table except one or two columns.
> 
> This is a useful feature in for instance bigquery.
> 
> I could not figure out how to make the parser deal with dialects, so I added this to the default parser.
> 
> I do not think that this is supported by all databases, so not sure where this should live, please advice.

In your parser code, you can use the value of dialect to determine whether to include certain features or syntax. In the case of selecting all columns except one or two, you could check if the dialect is bigquery, and if so, include the necessary syntax for this feature.

By using a configuration option, users can specify which dialect they are using, and the parser can be modified accordingly without affecting other users who may not be using the same dialect.

libenchao:
In Calcite, `SqlConformance` is a place where we used to make parser customizable.  And `SqlDialect` is a place where we customize to convert `RelNode` back to SQL.

This sounds an interesting feature, imo we could add this support for `BIG_QUERY` conformance.

hamnis:
May get around to this later, closing for now

PR #3091, opened 2023-03-02
[CALCITE-5550] Update instructions for requesting Jira account to use self-serve facility

F21:


zoudan:
LGTM

libenchao:
Should we also update `howto#Processing JIRA account requests`? From the procedure described in infra announce email, it seems that we only need to do a final review and approval for jira account requests?

F21:
I have left the procedure for processing requests as is for now, as I'd like to actually process a request before updating the instructions. It is unclear if the form will automatically deny the request if the username is already in use, etc, or if it just simply sends an email to the private list.

libenchao (approved):
make sense to me, then I'm +1 for this change.

rubenada:
LGTM

PR #3092, opened 2023-03-02
[CALCITE-5553] noFlatteningForInput handles single field structs

apilloud:


testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java
```diff
@@ -291,13 +297,10 @@ protected MockCatalogReaderSimple(RelDataTypeFactory typeFactory,
     registerTable(suppliersTable);
 
     // Register "EMP_20" and "EMPNULLABLES_20 views.
-    // Same columns as "EMP" amd "EMPNULLABLES",
-    // but "DEPTNO" not visible and set to 20 by default
-    // and "SAL" is visible but must be greater than 1000,
-    // which is the equivalent of:
+    // Same columns as "EMP" amd "EMPNULLABLES", but "DEPTNO" not visible and set to 20 by default
+    // and "SAL" is visible but must be greater than 1000, which is the equivalent of:
```
libenchao, line 301: Is it necessary to do this comment reformatting?

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -3709,6 +3709,12 @@ void checkCorrelatedMapSubQuery(boolean expand) {
     sql(sql).ok();
   }
 
+  @Test void testNestedStructFieldAccessWhere() {
```
libenchao, line 3712: The test name does not explicitly say that it must be a single field struct.

libenchao (commented):
@apilloud Thanks for your PR, the fix looks good to me generally, I've only two minor comments about the tests.

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java
apilloud: Without the comment reformatting I get the following test failure:
```
/home/runner/work/calcite/calcite/testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java:79:3: Method length is 376 lines (max allowed is 370). [MethodLength]
```
This seemed like the simplest change, but I'm happy to do something else instead.

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
apilloud: Renamed to `testNestedStructSingleFieldAccessWhere`

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java
libenchao: I see, then we'd better split this method into smaller ones in a separate issue, and I'm ok for this workaround in this one.

libenchao (commented):
@apilloud Thanks for the updating, the changes looks good now. There is only one thing left, the tests fails due to the orderness in `SqlToRelConverterTest.xml`.

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java
libenchao: I've logged CALCITE-5574 for this.

libenchao (approved):
+1, will merge after the main branch is open.

PR #3093, opened 2023-03-02
SAFE_CAST (enabled for BigQuery)

olivrlee:
https://issues.apache.org/jira/browse/CALCITE-5557


core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
```diff
@@ -168,7 +182,7 @@ public SqlCastFunction() {
   }
 
   @Override public SqlSyntax getSyntax() {
-    return SqlSyntax.SPECIAL;
+    return this.kind == SqlKind.CAST ? SqlSyntax.SPECIAL : SqlSyntax.FUNCTION;
```
julianhyde, line 185: if SAFE_CAST requires AS, it's not function syntax

core/src/main/codegen/templates/Parser.jj
```diff
@@ -6008,6 +6008,18 @@ SqlNode BuiltinFunctionCall() :
         <RPAREN> {
             return SqlStdOperatorTable.CAST.createCall(s.end(this), args);
         }
+    |
```
julianhyde, line 6011: could you combine with the logic for <CAST>?

core/src/main/java/org/apache/calcite/rex/RexExecutable.java
```diff
@@ -102,6 +102,38 @@ public void reduce(RexBuilder rexBuilder, List<RexNode> constExps,
     Hook.EXPRESSION_REDUCER.run(Pair.of(code, values));
   }
 
+  /**
+   * Same as {@link #reduce(RexBuilder, List, List)} but with a flag to throw on failure.
+   * <p>Catch the exception to return null for SAFE_ functions.</p>
```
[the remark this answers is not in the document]
olivrlee: This was because you had a comment saying public methods shouldnt just be changed, but a new one should be written and the old deprecated -- maybe that doesn't apply here?

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
```diff
@@ -1198,6 +1200,26 @@ public static <T, F extends Function<? extends T>> FunctionExpression<F> lambda(
     return lambda(type, Blocks.toFunctionBlock(body), toList(parameters));
   }
 
+  public static Expression safeExpression(Expression body){
```
olivrlee, line 1203: Using this block I can get it to generate an Expression with the try/catch
```
public Object[] apply(Object root0) {
  return new Object[] {
      (Integer) new org.apache.calcite.linq4j.function.Function0() {
        public Object apply() {
          try {
            return org.apache.calcite.avatica.util.DateTimeUtils.timeStringToUnixDate("12:12:11");
          } catch (Exception e) {
            return null;
          }
        }
      }
    };
}
```

However, in RexExecutable compile line 55 it doesn't like the cast to integer / block : `org.codehaus.commons.compiler.CompileException: Line 3, Column 15: Cannot cast "Reducer$1" to "java.lang.Integer"`

olivrlee:
Hey @julianhyde  if you could take a look at what I've done here: https://github.com/apache/calcite/pull/3093/files#diff-9e0f6e9c4d4f86444e48290b1e613ba700c0854cd614bee4a1912ccb75aed272R1203

@tjbanghart  helped me take a look at your suggestion too with using some `Supplier`

```
public static final Object safe(Supplier r) {
	try {
	 return r.get();
	} catch (RuntimeException e) {
	return null;
	}
}
```

but the typing isn't so compatible with `Expressions.lambda`, which expects a subclass of `Function.class` and we haven't figured it out

core/src/main/java/org/apache/calcite/rex/RexExecutable.java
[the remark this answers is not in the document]
julianhyde: You can usually move the method body into the new method (with the extra parameter) and have the old (deprecated) method call the new method.

If the old method is deprecated you should change all calls to use the new method, and remove its javadoc. Thus the new method's javadoc wouldn't be 'Same as (old method)'.

julianhyde:
In future, please use the jira case number in the first commit message and in the title of the PR. (Yes, that requires that you log a jira case before starting work on the bug. That is a good thing.)

(Too late to change the first commit message now. That would destroy history.)

julianhyde:
In future, please use the jira case number in the first commit message and in the title of the PR. (Yes, that requires that you log a jira case before starting work on the bug. That is a good thing.)

(Too late to change the first commit message now. That would destroy history.)

julianhyde:
I think `Expressions.lambda(BlockStatement, Iterable)` will work. If there are zero parameters `deduceType` will return `Function0.class`, which is equivalent to `java.lang.Supplier`. (This code was written before `Supplier` was added in JDK 8.)

`BlockStatement` will probably contain a single statement, a `GotoStatement` with your expression, which you will create using `Expressions.return_`.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
olivrlee: The original code without adding a lambda + try/catch block looks like this when running `SAFE_CAST("a" AS INT)`

```
public Object[] apply(Object root0) {
  return new Object[] {
      Integer.valueOf(org.apache.calcite.runtime.SqlFunctions.toInt("a"))};
}
```

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
```diff
@@ -1164,6 +1164,7 @@ public static <F extends Function<?>> FunctionExpression<F> lambda(
     return new FunctionExpression<>(type, body, toList(parameters));
   }
 
+
```
zoudan, line 1167: remove this empty line

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3481,6 +3481,7 @@ public static <E> Collection<E> multisetIntersectAll(Collection<E> c1,
     return result;
   }
 
+
```
zoudan, line 3484: remove empty line

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -639,6 +639,10 @@ public enum SqlKind {
    */
   CAST,
 
+  /**
+   * The {@code SAFE_CAST} function. */
```
zoudan, line 643: It is better to add description for what is the difference between `SAFE_CAST` and `CAST`

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -652,8 +714,8 @@ Expression translateCast(
       break;
     case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
       convert =
-          RexImpTable.optimize2(operand,
-              Expressions.call(
+              RexImpTable.optimize2(
```
zoudan, line 717: remove one tab?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -267,10 +268,39 @@ Expression translate(RexNode expr, RexImpTable.NullAs nullAs,
     return nullAs.handle(translated);
   }
 
+  /**
+   * Used for safe operators that return null if an exception is thrown.
+   */
+  Expression expressionHandlingSafe(Expression body, boolean safe) {
```
zoudan, line 274: could be private?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -267,10 +268,39 @@ Expression translate(RexNode expr, RexImpTable.NullAs nullAs,
...
+    if (safe) {
+      return safeExpression(body);
+    } else {
+      return body;
+    }
+  }
+
+  public static Expression safeExpression(Expression body) {
```
zoudan, line 282: could be private?

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
```diff
@@ -83,12 +84,25 @@ public class SqlCastFunction extends SqlFunction {
   //~ Constructors -----------------------------------------------------------
 
   public SqlCastFunction() {
-    super("CAST",
-        SqlKind.CAST,
+    this(SqlKind.CAST);
+  }
+
+  private SqlCastFunction(SqlKind kind) {
```
zoudan, line 90: How about just keep this constructor and remove `withKind`

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
zoudan: +1, SAFE_CAST is a special SqlSyntax

zoudan (changes requested):

core/src/main/java/org/apache/calcite/sql/SqlKind.java
olivrlee: I'll update the comment. I did describe the function in `SqlLibraryOperators` too

olivrlee:
I'm still in the middle of cleaning up the PR, thanks for reviewing @zoudan 


Do you have any thoughts on SqlSyntax.SPECIAL? 
It *should* be SPECIAL, but the operator table lookup currently doesn't look for SPECIAL syntax functions, so that was just a temporary workaround.

babel/src/test/resources/sql/big-query.iq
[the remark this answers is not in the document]
olivrlee: Done.

julianhyde:
@olivrlee Can you review my additional commits? I was able to solve the SPECIAL issue

olivrlee:
I did take a look but communicated to Julian offline*

PR #3094, opened 2023-03-03
[CALCITE-5554] Add DAYOFWEEK and DAYOFYEAR as valid synonyms of DOW, DOY

tanclary:
JIRA Case: https://issues.apache.org/jira/browse/CALCITE-5554
Minor change to parser to accept "DAYOFWEEK" and "DAYOFYEAR" that correlate to the DOW and DOY time units, respectively. 

Note that DAYOFWEEK and DAYOFYEAR are also valid 1-operand functions that return the corresponding DOW or DOY for a provided date. (They are equivalent to EXTRACT(DOW/DOY from DATE)). This is also the case for other time units like HOUR and MINUTE so I tried to mirror how they handled a token that could be both a valid qualifier and function name. 

Please let me know if you have any comments, questions, or suggestions. Thanks!


zoudan:
LGTM

PR #3098, opened 2023-03-07
[CALCITE-5563] Reduce loops to optimize RelSubset#getParents and RelSubset#getParentSubsets

asdfgh19:
jira: [CALCITE-5563](https://issues.apache.org/jira/browse/CALCITE-5563)
Once we have found a matching Relsubset from its parent input, we can immediately end the inner loop


rubenada:
@asdfgh19 Thanks for this improvement!
Could you please squash commits into a single one? As commit message, I'd suggest `[CALCITE-5563] Reduce loops to optimize RelSubset#getParents and RelSubset#getParentSubsets`

asdfgh19:
@rubenada Thanks for reviewing the code!

PR #3099, opened 2023-03-08
[CALCITE-5565] Implement BigQuery LOG Function

tanclary:
Calcite currently supports the LN and LOG10 function. BigQuery offers the LOG function which allows for the base of the logarithm to be specified as a second argument. If the second argument is not provided, the base is defaulted to e (making it identical to the LN function).

Example: LOG(64, 10) would return 2.
Example 2: LOG(10) would return the same as LN(10).

BigQuery docs: https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#log

### DIFF
--- a/babel/src/test/resources/sql/big-query.iq
+++ b/babel/src/test/resources/sql/big-query.iq
@@ -731,6 +731,33 @@ Call to function 'SPLIT' with argument of type 'BINARY(2)' requires extra delimi
 !error
 
 #####################################################################
+# LOG
+#
+# LOG(value1, value2)
+#
+# If only value1 is present, LOG is a synonym of LN. If value2 is also
+# present, LOG computes the logarithm of value1 to base value2.
+SELECT LOG(64, 8) as result;
++--------+
+| result |
++--------+
+|    2.0 |
++--------+
+(1 row)
+
+!ok
+
+SELECT LOG(100) as result;
++-------------------+
+| result            |
++-------------------+
+| 4.605170185988092 |
++-------------------+
+(1 row)
+
+!ok
+
+#####################################################################
 # STRING
 #
 # STRING(timestamp_expression[, time_zone])
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -153,6 +153,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.JSON_SET;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.JSON_STORAGE_SIZE;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.JSON_TYPE;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LEFT;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG;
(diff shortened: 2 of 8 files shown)


core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -473,6 +474,7 @@ Builder populate() {
       defineMethod(EXP, "exp", NullPolicy.STRICT);
       defineMethod(POWER, "power", NullPolicy.STRICT);
       defineMethod(LN, "ln", NullPolicy.STRICT);
+      defineMethod(LOG, "log", NullPolicy.STRICT);
```
julianhyde, line 477: Would this be simpler if done with an implementor that translates 'log(x, y)' to 'ln(x) / ln(y)'? Constant reduction could kick in, we could potentially optimize for bases such as e, 2, 10, and we would not need to add to `SqlFunctions`.

Method implementor is kind of a hammer.

core/src/main/codegen/default_config.fmpp
```diff
@@ -88,6 +88,8 @@ parser: {
     "DATETIME_DIFF"
     "DATETIME_INTERVAL_CODE"
     "DATETIME_INTERVAL_PRECISION"
+    "DAYOFWEEK"
```
zoudan, line 91: this has nothing to do with `LOG`?

babel/src/test/resources/sql/big-query.iq
```diff
@@ -371,6 +371,30 @@ SELECT EXTRACT(DAY FROM DATE '2013-12-25') AS the_day;
 
 !ok
 
+SELECT
+  EXTRACT(DAYOFWEEK FROM DATE '2008-12-25') as day_of_week,
```
zoudan, line 375: This PR includes some changes which are not release to CALCITE-5565

zoudan (changes requested):

babel/src/test/resources/sql/big-query.iq
tanclary: Just rebased, was previously dependent on uncommitted changes. Sorry about that!

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3684,6 +3687,43 @@ private static class LogicalNotImplementor extends AbstractRexCallImplementor {
     }
   }
 
+  /**Implementor for the {@code LN, LOG, and LOG10} operators
```
julianhyde, line 3690: Need space at start of comment. Only the operator names should be monospaced, not “and”.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3684,6 +3687,43 @@ private static class LogicalNotImplementor extends AbstractRexCallImplementor {
...
+        switch (argValueList.size()) {
+        case 1:
+          operand1 = Expressions.constant(Math.exp(1));
+          break;
+        default:
+          operand1 = argValueList.get(1);
+        }
+        break;
+      case "LOG10":
+      default:
```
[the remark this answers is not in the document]
tanclary: Done.

sarah-bienenstock-privitar (approved):
looks good!

PR #3100, opened 2023-03-08
[CALCITE-5548] Add (partial) MSSQL CONVERT support

askarbozcan:
This PR adds partial support for [MSSQL CONVERT function](https://learn.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver16). A function very similar to CAST that has an optional "style" operand to support different kinds of date string formats, float decimal points in string form, etc.

The reason this PR is "partial", is because it simply adds a thin wrapper around CAST operator and does not implement anything related to 'style' operand other than allowing for its parsing.

**Implemented**
* Add lookahead in parser to first <CONVERT> token to make sure there are no clashes with other unrelated CONVERT functions (including MSSQL CONVERT func)
* Add parser rule specifically for CONVERT to add support for its syntax
* SQLOperator (SqlConvertMssqlFunction) which utilizes SqlOperator.rewriteCall() to delegate all the functionality to CAST
* Related Quidem tests in functions.iq

**Not implemented**
* 'style' operand is ignored. It is parsed, however beyond that, any functionality related to it is not implemented.

**Difficulties & notes on implementing 'style' operator**
Almost everywhere in the codebase, the operand order for CAST is hard-coded, which is not a bad assumption at all; yet because of it it seems impossible to fully implement CONVERT unless it's properly added as its own operator.

For reference on CAST-related functionality:
[RexImpTable.CastImplementor](https://github.com/apache/calcite/blob/ee9b80b0b68d442991dfaa142722e3488ec73e79/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java#L2918)

[StandardConvertletTable.convertCast](https://github.com/apache/calcite/blob/ee9b80b0b68d442991dfaa142722e3488ec73e79/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java#L595)

_PS: This is my first contribution, and my first proper "dive" into Calcite internals. So all feedback, no matter how trivial is welcome._


askarbozcan (commented):
LGTM, thanks for rewriting docs (and some comments) in a more concise manner & showing me the more functional style for definition of operators.

PR #3102, opened 2023-03-09
[CALCITE-5571] Remove org.jetbrains.annotations from java source code

zabetak:
The project primarily uses checkerframework annotations (org .checkerframework.checker.nullness.qual) for specifying nullability; not org.jetbrains.annotations.

To keep things consistent and also avoid mixing up annotations from different providers in the future, this commit removes the last references to org.jetbrains.annotations and excludes the org.jetbrains:annotations dependency from the build to avoid accidentally using such annotations in the future.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -334,6 +334,11 @@ allprojects {
     val javaUsed = file("src/main/java").isDirectory
     if (javaUsed) {
         apply(plugin = "java-library")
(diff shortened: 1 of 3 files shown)


libenchao (approved):
+1

snuyanzin (approved):

zoudan:
LGTM

PR #3007, opened 2022-12-17
[CALCITE-5440] Bump gradle from 7.4.2 to 7.6.1

snuyanzin:
`sourceCompatibility` and `targetCompatibility` removed in latest gradle.
Currently there is one warning in output for `./gradlew help --warning-mode=all`
```
> Task :buildSrc:autostyleKotlinCheck NO-SOURCE
Relying on FileTrees for ignoring empty directories when using @SkipWhenEmpty has been deprecated. This is scheduled to be removed in Gradle 8.0. Annotate the property sourceFiles with @IgnoreEmptyDirectories or remove @SkipWhenEmpty. Consult the upgrading guide for further information: https://docs.gradle.org/7.6/userguide/upgrading_version_7.html#empty_directories_file_tree
```

From one side it complains about Autostyle code (https://github.com/autostyle/autostyle), from another it looks like a gradle bug since a corresponding method in autostyle is marked by both annotations:  `@SkipWhenEmpty` and `@IgnoreEmptyDirectories`. 
Anyway it is not related to Calcite code

PR #3107, opened 2023-03-13
[CALCITE-5576] Upgrade kotlin from 1.5.31 to 1.8.10

zabetak:


snuyanzin:
i wonder if it is ok  from compatibility's  point of view...
Compatibility matrix[1] says it should be 1.5.31 for gradle 7.3 and 1.6.21 for 7.5.
I bumped it to 1.7.10 while gradle update to 7.6.1 https://github.com/apache/calcite/pull/3007
[1] https://docs.gradle.org/current/userguide/compatibility.html#kotlin

zabetak:
Thanks for the pointer @snuyanzin . I raised the ticket without checking the compatibility matrix. Indeed it is better to stick to the officially supported versions. I am closing this for now.

PR #3108, opened 2023-03-13
[CALCITE-5577] Map value constructor is unparsed incorrectly for SparkSqlDialect

MasseGuillaume:


libenchao (approved):
+1, will merge after the main branch is open.

PR #3109, opened 2023-03-14
[CALCITE-5580] Add SPLIT() Function (Enabled for BigQuery)

tanclary:
BigQuery offers the `SPLIT()` function which splits a string at an optionally-specified delimiter into a string array. If no delimiter is specified, it is default to a comma. If the string is empty, an array of a single empty string is returned. If the delimiter is not found in the string, an array with a single element (the string) is returned. 

In BigQuery, the function can also accept bytes. In order to implement this, I think some modifications to ByteString.java may be required. I will probably not do this at least for my initial draft. If anyone has any suggestions or guidance on whether or not it should be supported, I would appreciate it. 

Documentation and example cases may be found below.

EXAMPLE: `SPLIT('h,e,l,l,o')` would return: `[h, e, l, l, o]`.
EXAMPLE: `SPLIT('h-e-l-l-o', '-')` would return: `[h, e, l, l, o]`.

[BigQuery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#split) 
[JIRA Case](https://issues.apache.org/jira/browse/CALCITE-5580)

### DIFF
--- a/babel/src/test/resources/sql/big-query.iq
+++ b/babel/src/test/resources/sql/big-query.iq
@@ -625,6 +625,55 @@ SELECT (19 % 19) as result;
 !ok
 
 #####################################################################
+# SPLIT
+#
+# SPLIT(string [, delimiter])
+#
+# Splits string using the delimiter argument. For STRING, the default
+# delimiter is the comma.
+#
+# Returns a STRING array as result.
+WITH letters AS
+  (SELECT '' as letter_group
+  UNION ALL
+  SELECT 'a' as letter_group
+  UNION ALL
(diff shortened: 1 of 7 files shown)


julianhyde:
Not sure how this happened, but the author of the commit is "tanclary <116591231+tanclary@users.noreply.github.com>"; you probably intend it to be "Tanner Clary <tannerclary@google.com>". Can you force-push to fix it?

tanclary:
@julianhyde Oops, switched to new machine and forgot to set configs. Let me know if that looks better now. Thanks!

PR #3113, opened 2023-03-14
[CALCITE-5746] Support jdk19 

snuyanzin:
Enable jdk19 in ci.

~~`-Pwerror=false` since in jdk19 constructors for `Locale` are deprecated~~
Added Locale's constructors to forbidden api signatures


snuyanzin:
@zabetak could you please have a look
sorry for the poke (GitHub just suggested to ask you)

core/src/main/java/org/apache/calcite/sql/parser/SqlParserUtil.java
```diff
@@ -738,17 +741,7 @@ public static ParsedCollation parseCollation(String in) {
...
-    String[] localeParts = localeStr.split("_");
-    Locale locale;
-    if (1 == localeParts.length) {
-      locale = new Locale(localeParts[0]);
-    } else if (2 == localeParts.length) {
-      locale = new Locale(localeParts[0], localeParts[1]);
-    } else if (3 == localeParts.length) {
-      locale = new Locale(localeParts[0], localeParts[1], localeParts[2]);
-    } else {
-      throw RESOURCE.illegalLocaleFormat(localeStr).ex();
```
zabetak, line 750: It seems that after these changes we are never going to throw the `illegalLocaleFormat` exception. Are we going to throw another exception? Are we going to fallback silently to something else? 

If we cannot keep the old behavior then we should document the breaking change in `history.md` file. Also if we never throw `RESOURCE.illegalLocaleFormat` then we maybe we should also drop the respective (dead) code.

core/src/test/java/org/apache/calcite/util/UtilTest.java
```diff
@@ -914,7 +914,7 @@ private List<Integer> makeConsList(int start, int end) {
     }
     // Example locale names in Locale.toString() javadoc.
     String[] localeNames = {
-        "en", "de_DE", "_GB", "en_US_WIN", "de__POSIX", "fr__MAC"
+        "en", "de_DE", "fr_CA"
```
zabetak, line 917: The change implies that we no longer support certain patterns when parsing locales so I think we should mention the change in behavior in `history.md`.

zabetak (changes requested):

core/src/main/java/org/apache/calcite/sql/parser/SqlParserUtil.java
snuyanzin: Since this change makes use of IETF BCP 47 language tag string (used by Locale#forLanguageTag) there could be more parts. Fallback is jdk's fallback

Thanks about mentioning `RESOURCE.illegalLocaleFormat`

core/src/test/java/org/apache/calcite/util/UtilTest.java
```diff
@@ -914,7 +914,7 @@ private List<Integer> makeConsList(int start, int end) {
     }
     // Example locale names in Locale.toString() javadoc.
     String[] localeNames = {
-        "en", "de_DE", "_GB", "en_US_WIN", "de__POSIX", "fr__MAC"
+        "en", "de_DE", "gb_GB", "en_US_WINDOWS", "de__POSIX", "fr__MACOS"
```
snuyanzin, line 917: need to extend `en_US_WIN` to `en_US_WINDOWS` since `Locale#forLanguageTag` assumes that variants contain between 4 and 8 symbols

core/src/test/java/org/apache/calcite/util/UtilTest.java
snuyanzin: added such note

core/src/main/java/org/apache/calcite/sql/parser/SqlParserUtil.java
snuyanzin: First it says (in jdk19 javadoc) to look for `Obtaining Locales` section at https://github.com/openjdk/jdk/commit/526e73498eef5c7608845501ab4ebef0997a5c0d#diff-bae63085ee0310e91011b048b2b59d0986eedf702dee10d7b0ab75c8e20a61eaR728
and then in that section there are 2 ways as mentioned at https://github.com/openjdk/jdk/commit/526e73498eef5c7608845501ab4ebef0997a5c0d#diff-bae63085ee0310e91011b048b2b59d0986eedf702dee10d7b0ab75c8e20a61eaR245
1. `Locale#forLanguageTag` which is used in this PR
2. `Locale#of` which is available only since jdk19 (https://github.com/openjdk/jdk/pull/7947).

>Do all Calcite unit tests run if we use setLanguageTag?

I replaced all occurrences of `Locale` constructor (including tests) then yes...
Or do you mean something else?

 >Are we sure that we do not break some SQL standard specification around COLLATE clause (etc.) by doing this change?
 
 That thing unfortunately I don't know how to check...

zabetak: Before the changes in this PR there was somewhat strict control on what formats are allowed/supported when working with locales and an exception was thrown if the requested locale is now invalid.

Now with the use of `Locale#forLanguageTag` we will never get an error since it appears that we will always fallback to some defaults. 

Instead of using the `Locale#forLanguageTag` we could use the `Locale.Builder.setLanguageTag`, which as far as I understand is not as permissive,  and will thrown an error when the locale is not well formed: https://download.java.net/java/early_access/panama/docs/api/java.base/java/util/Locale.Builder.html#setLanguageTag(java.lang.String)

In the context of this PR I was wondering if the `Locale.Builder.setLanguageTag` is a better alternative to `Locale#forLanguageTag`.

The latest standard can be found here: https://www.iso.org/standard/76584.html 
It is not free but you may find older versions available online (https://modern-sql.com/standard). The Part 2: Foundations covers the `COLLATE` clause.

snuyanzin: I'm ok with `Locale.Builder.setLanguageTag`
the only thing we should probably handle is that current behavior: in case of empty `localeString` it returns `Locale.ROOT` while `Locale.Builder.setLanguageTag` fails

Also added actual `localeStr` in exception message since `IllformedLocaleException` throwing by `setLamguageTag` does not contain it

core/src/main/java/org/apache/calcite/util/Util.java
```diff
@@ -1717,16 +1719,14 @@ private static void appendPosixTime(StringBuilder buf, int millis) {
...
-    case 1:
-      return new Locale(strings[0]);
-    case 2:
-      return new Locale(strings[0], strings[1]);
-    case 3:
-      return new Locale(strings[0], strings[1], strings[2]);
-    default:
-      throw new AssertionError("bad locale string '" + localeString + "'");
+    if (localeString.isEmpty()) {
+      return Locale.ROOT;
```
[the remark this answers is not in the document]
snuyanzin: if we are using `Locale.Builder.setLanguageTag` as was suggested in other comment https://github.com/apache/calcite/pull/3113/files#r1212934318 then yes.

The issue with `Locale.Builder.setLanguageTag` is that it does not allow empty `lcoalStr` and fails
I mentioned it in same tread, probably should do it here  as well

site/_docs/history.md
```diff
@@ -43,6 +43,10 @@ z.
 #### Breaking Changes
 {: #breaking-1-35-0}
 
+The way of Locale parsing changed within [<a href="https://issues.apache.org/jira/browse/CALCITE-5567">CALCITE-5567</a>]
+Now locales are parsed according to IETF BCP 47 language tag string.
+This leads to the fact that some locales not satisfying IETF BCP 47 language tag string.
```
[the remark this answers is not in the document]
snuyanzin: yep, thanks for the catch
I changed it

core/src/main/java/org/apache/calcite/util/Util.java
```diff
@@ -1717,16 +1719,14 @@ private static void appendPosixTime(StringBuilder buf, int millis) {
...
+    }
+    try {
+      return new Locale.Builder().setLanguageTag(
+          UNDERSCORE.matcher(localeString).replaceAll("-")).build();
+    } catch (IllformedLocaleException e) {
+      throw new AssertionError("bad locale string '" + localeString + "'", e);
```
zabetak, line 1729: A `IllegalStateException` or `IllegalArgumentException` would be more appropriate here but since it was an `AssertionError` before I am also fine leaving it as is.
I don't think an `AssertionError` is appropriate cause we may arrive here by user input.

zabetak (approved):
I am fine merging this as soon as `illegalLocaleFormat` is restored.

core/src/main/java/org/apache/calcite/util/Util.java
snuyanzin: makes sense, thanks

snuyanzin:
thanks a lot for having a look @zabetak , @rubenada 

if there is no more objections/suggestions i'm going to merge this weekend

PR #3114, opened 2023-03-14
[CALCITE-5585] Add STRPOS() Function (Enabled for BigQuery)

tanclary:
BigQuery supports the `STRPOS()` function with accepts 2 `STRINGS` as arguments. The function returns the 1-based position of the first occurrence of the second string within the first (example below). If no occurrence is found, the function returns 0. The function also supports ByteStrings with the same behavior.

EXAMPLE: `STRPOS("abc", "bc")` would return: `2`.
EXAMPLE: `STRPOS("abc", "d")` would return: `0`.

[BigQuery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#strpos)
[JIRA Case](https://issues.apache.org/jira/browse/CALCITE-5585)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -255,6 +255,10 @@ private SqlLibraryOperators() {
           OperandTypes.STRING)
           .withFunctionType(SqlFunctionCategory.STRING);
 
+  /** The "STRPOS(value1, value2)" function. */
+  @LibraryOperator(libraries = {BIG_QUERY})
+  public static final SqlFunction STRPOS = new SqlPositionFunction("STRPOS");
```
snuyanzin, line 260: It looks like same function is available in Postgres[1]

[1] https://www.postgresql.org/docs/9.1/functions-string.html

tanclary: Oh yes, sorry about that. Should be fixed now.

PR #3120, opened 2023-03-18
[CALCITE-5593] Support aliases query in Elasticsearch adapters 

coolbeevip:
Jira issue https://issues.apache.org/jira/browse/CALCITE-5593

### DIFF
--- a/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchema.java
+++ b/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchema.java
@@ -38,6 +38,10 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 /**
  * Each table in the schema is an ELASTICSEARCH index.
@@ -121,6 +125,13 @@ public class ElasticsearchSchema extends AbstractSchema {
       }
 
       Set<String> indices = Sets.newHashSet(root.fieldNames());
+      Set<String> aliases = root.findValues("aliases").stream()
+          .map(JsonNode::fieldNames)
+          .flatMap(
+              it -> StreamSupport.stream(
+                  Spliterators.spliteratorUnknownSize(it, Spliterator.ORDERED), false))
+          .collect(Collectors.toSet());
+      indices.addAll(aliases);
       return indices;
     }
   }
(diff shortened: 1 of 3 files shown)


elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchSchema.java
```diff
@@ -120,7 +120,9 @@ private Set<String> indicesFromElastic() throws IOException {
         throw new IllegalStateException(message);
       }
 
+      Set<String> aliases = Sets.newHashSet(root.findPath("aliases").fieldNames());
       Set<String> indices = Sets.newHashSet(root.fieldNames());
+      indices.addAll(aliases);
```
libenchao, line 125: I'm not familiar with es adapter, but is it possible to add a unit test to show that it only works after this change?

coolbeevip: Sure, I need to understand the current ES integration testing method first, I am only familiar with ES testing based on testcontainers

coolbeevip: Hi,   I added a method to create an alias and added unit tests

zips_alias is an alias of zips

```
@Test void testAlias() {
    calciteAssert()
        .query("select state, city from elastic.zips_alias");
  }
```

julianhyde:
The commit message will become the release notes. It needs to describe the feature. Currently it doesn't mention that it concerns the Elasticsearch adatper.

coolbeevip:
> The commit message will become the release notes. It needs to describe the feature. Currently, it doesn't mention that it concerns the Elasticsearch adapter.

thanks for the reminder, I have fixed the commit message

elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/ElasticSearchAdapterTest.java
```diff
@@ -189,6 +191,11 @@ private CalciteAssert.AssertThat calciteAssert() {
         .returnsCount(0);
   }
 
+  @Test void testAlias() {
+    calciteAssert()
+        .query("select state, city from elastic.zips_alias");
```
libenchao, line 196: This test seems asserting nothing, hence it won't be run actually.

libenchao:
@coolbeevip The CI failed now, it seems that you are using an guava feature which is not available in our oldest supported guava version: 16.0.1

coolbeevip:
> @coolbeevip The CI failed now, it seems that you are using an guava feature which is not available in our oldest supported guava version: 16.0.1
Thanks for the tip, I will use Java StreamSupport.stream instead Guava Streams.stream

libenchao (approved):
+1, merging. @coolbeevip Thanks for your contribution.

PR #3125, opened 2023-03-21
[CALCITE-5449] Allow EXTRACT() to accept DAYOFWEEK, DAYOFYEAR, ISOWEEK, WEEK(WEEKDAY)

tanclary:
The `EXTRACT()` function currently does not accept time frames (user-defined, ISOWEEK, WEEK(WEEKDAY), etc). This PR adds support for `ISOWEEK` and `WEEK(WEEKDAY)` by adding a customDateExtract method, similar to `customDateDiff` or `customDateAdd`. 

This implementation utilizes `floorDate()` and `diffDate()` in the `TimeFrameSet` class. 

There is also an issue I discovered while working on this: `EXTRACT(WEEK FROM DATE)` currently returns the `ISOWEEK`. The cause of this lies in Avatica's `DateTimeUtils` class, [here](https://github.com/apache/calcite-avatica/blob/92edd93e1977575f1fa3ceb11f03371b959c6660/core/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java#L802). I plan on filing a separate bug for this because it requires changes to Avatica. If anyone has suggestions on updating the math to correctly return the week, please let me know.

As always, I would appreciate any comments and suggestions on how I could improve this or expand it to include other time frames (such as user-defined custom time frames).


babel/src/test/resources/sql/big-query.iq
```diff
@@ -398,14 +398,12 @@ SELECT
...
-
-!if (false) {
 SELECT
   d,
   EXTRACT(ISOYEAR FROM d) AS isoyear,
-  EXTRACT(ISOWEEK FROM d) AS isoweek,
-  EXTRACT(YEAR FROM d) AS year,
-  EXTRACT(WEEK FROM d) AS week
+  EXTRACT(ISOWEEK FROM d) as isoweek,
+  EXTRACT(WEEK(TUESDAY) FROM d) AS week_tues,
```
zinking, line 405: can you add new columns instead of overriding existing column ?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -5127,8 +5127,6 @@ SqlIntervalQualifier TimeUnit() : {
 |   <MINUTE> { return new SqlIntervalQualifier(TimeUnit.MINUTE, null, getPos()); }
 |   <HOUR> { return new SqlIntervalQualifier(TimeUnit.HOUR, null, getPos()); }
 |   <DAY> { return new SqlIntervalQualifier(TimeUnit.DAY, null, getPos()); }
-|   <DAYOFWEEK> { return new SqlIntervalQualifier(TimeUnit.DOW, null, getPos()); }
```
zinking, line 5130: could you explain why this TOKEN is removed ?

core/src/main/java/org/apache/calcite/rel/type/TimeFrameSet.java
```diff
@@ -287,6 +287,41 @@ public long diffTimestamp(long timestamp, long timestamp2, TimeFrame frame) {
...
 
+  /** For ISOWEEK and WEEK(WEEKDAY), EXTRACT can be rewritten
+   * as the composition of DATE_DIFF and DATE_TRUNC. DAYOFWEEK and DAYOFYEAR
+   * are just translated to DOW and DOY respectively. */
+  public long extractDate(int date, TimeFrame timeFrame) {
+    TimeUnitRange timeUnitRange;
+    int offset = 0;
+    // Date will be truncated to either ISOYEAR or YEAR depending on time frame.
+    switch (timeFrame.name()) {
+    case "DAYOFWEEK":
```
zinking, line 298: I am thinking of a enum class with more comments might be better.

core/src/main/codegen/templates/Parser.jj
tanclary: DAYOFWEEK is a valid time unit for BigQuery. However here (which I added this in a previous commit), it translates it to DOW which is not a valid time unit for BigQuery. Instead, I wanted to treat it as a timeframe that is an alias to DOW (and DOY, respectively).

babel/src/test/resources/sql/big-query.iq
tanclary: Yes I could. I did a poor job of noting this but the reason I didn't do this originally is because WEEK for the EXTRACT function is currently configured incorrectly in Avatica. As it stands, EXTRACT(WEEK ...) actually returns the ISOWEEK. This causes all tests for EXTRACT WEEK in Calcite to be off in some cases where the ISOWEEK and WEEK for a given date happen to be different. I can open a ticket for this issue if you think it is warranted?

PR #3128, opened 2023-03-23
[CALCITE-5606] Add SqlLibrary.ALL and change the library of TANH, COSH, SINH to it

zoudan:
Add a new SqlLibrary named 'ALL' with abbreviation '*', if an operator is marked with this SqlLibrary, it means it's a common one and should belong to all the library. TANH, COSH, SINH are three of these operators.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
@@ -24,6 +24,7 @@ import com.google.common.collect.ImmutableMap;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -68,7 +69,9 @@ public enum SqlLibrary {
   POSTGRESQL("p", "postgresql"),
   /** A collection of operators that are in Apache Spark but not in standard
    * SQL. */
-  SPARK("s", "spark");
+  SPARK("s", "spark"),
+  /** A collection of operators that could be used in all libraries except STANDARD and SPATIAL. */
+  ALL("*", "all");
 
   /** Abbreviation for the library used in SQL reference. */
   public final String abbrev;
@@ -94,11 +97,21 @@ public enum SqlLibrary {
   /** Parses a comma-separated string such as "standard,oracle". */
   public static List<SqlLibrary> parse(String libraryNameList) {
     final ImmutableList.Builder<SqlLibrary> list = ImmutableList.builder();
-    for (String libraryName : libraryNameList.split(",")) {
-      SqlLibrary library =
-          requireNonNull(SqlLibrary.of(libraryName),
-              () -> "library does not exist: " + libraryName);
-      list.add(library);
+    List<String> libList = Arrays.asList(libraryNameList.split(","));
+    if (libList.contains(ALL.abbrev) || libList.contains(ALL.fun)) {
+      // Add all the libraries except ALL, STANDARD, SPATIAL for 'all' and '*'.
+      for (SqlLibrary value : values()) {
+        if (value != ALL && value != STANDARD && value != SPATIAL) {
+          list.add(value);
+        }
+      }
+    } else {
+      for (String libraryName : libraryNameList.split(",")) {
+        SqlLibrary library =
+            requireNonNull(SqlLibrary.of(libraryName),
+                () -> "library does not exist: " + libraryName);
+        list.add(library);
+      }
     }
     return list.build();
   }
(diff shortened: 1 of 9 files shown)


site/_docs/reference.md
```diff
@@ -2648,7 +2648,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | ARRAY_REVERSE(array)                           | Reverses elements of *array*
 | m s | CHAR(integer)                                | Returns the character whose ASCII code is *integer* % 256, or null if *integer* &lt; 0
 | o p | CHR(integer)                                 | Returns the character whose UTF-8 code is *integer*
-| b o | COSH(numeric)                                | Returns the hyperbolic cosine of *numeric*
+| * | COSH(numeric)                                  | Returns the hyperbolic cosine of *numeric*
```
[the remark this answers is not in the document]
zoudan: thanks for reminding

core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
```diff
@@ -68,7 +69,9 @@ public enum SqlLibrary {
   POSTGRESQL("p", "postgresql"),
   /** A collection of operators that are in Apache Spark but not in standard
    * SQL. */
-  SPARK("s", "spark");
+  SPARK("s", "spark"),
+  /** A collection of operators that could be used in all libraries except STANDARD and SPATIAL. */
+  ALL("*", "all");
```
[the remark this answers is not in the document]
zoudan: I'll change it

core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
```diff
@@ -94,11 +97,21 @@ public enum SqlLibrary {
...
+    List<String> libList = Arrays.asList(libraryNameList.split(","));
+    if (libList.contains(ALL.abbrev) || libList.contains(ALL.fun)) {
+      // Add all the libraries except ALL, STANDARD, SPATIAL for 'all' and '*'.
+      for (SqlLibrary value : values()) {
+        if (value != ALL && value != STANDARD && value != SPATIAL) {
+          list.add(value);
+        }
+      }
+    } else {
+      for (String libraryName : libraryNameList.split(",")) {
```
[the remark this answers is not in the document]
zoudan: nice catch

zoudan:
@tanclary I have updated this CR, please have a look when you have time.

tanclary:
@zoudan LGTM! I am not a committer but hopefully someone has time to take a look soon.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperatorTableFactory.java
```diff
@@ -70,18 +71,44 @@ private SqlLibraryOperatorTableFactory(Class... classes) {
   /** A cache that returns an operator table for a given library (or set of
    * libraries). */
   @SuppressWarnings("methodref.receiver.bound.invalid")
-  private final LoadingCache<ImmutableSet<SqlLibrary>, SqlOperatorTable> cache =
+  private final LoadingCache<CacheKey, SqlOperatorTable> cache =
       CacheBuilder.newBuilder().build(CacheLoader.from(this::create));
 
+  /**
+   * Cache key that contains library set and a flag indicate whether include operators under
```
julianhyde, line 78: Can you explain why cacheKey is required? After expanding 'all', the set of SqlLibary is canonical.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
```diff
@@ -94,11 +97,21 @@ public enum SqlLibrary {
   /** Parses a comma-separated string such as "standard,oracle". */
   public static List<SqlLibrary> parse(String libraryNameList) {
     final ImmutableList.Builder<SqlLibrary> list = ImmutableList.builder();
-    for (String libraryName : libraryNameList.split(",")) {
-      SqlLibrary library =
-          requireNonNull(SqlLibrary.of(libraryName),
-              () -> "library does not exist: " + libraryName);
-      list.add(library);
+    List<String> libList = Arrays.asList(libraryNameList.split(","));
```
julianhyde, line 100: Does this do the right thing if someone writes 'all,spatial'? I suspect not, but even if it does, it's not clear.

To simplify, change the method contract so that it may contain duplicates, may contain 'all', and may not be sorted. Then deal with expanding 'all' in the code that calls the method.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperatorTableFactory.java
zoudan: Because I want to handle the special case in `DocumentationTest#testAllFunctionsAreDocumented`, e.g. we do not want to include the operators under 'ALL' when we get operators under 'SPARK'.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
zoudan: I have changed this function, please have another look when you have time.

zoudan: Hi @julianhyde, may I ask if you have time to review this PR, thanks very much

PR #3134, opened 2023-03-29
[CALCITE-5619] Add support for TO_CHAR

tindzk:


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -863,6 +863,16 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
           OperandTypes.STRING.or(OperandTypes.BINARY),
           SqlFunctionCategory.STRING);
 
+  /** The "TO_CHAR(timestamp, format)" function;
+   * converts {@code timestamp} to string according to the given {@code format}.
+   */
+  @LibraryOperator(libraries = {MYSQL, ORACLE, POSTGRESQL})
+  public static final SqlFunction TO_CHAR =
```
dssysolyatin, line 870: PostgreSQL has 3 different to_char:
`to_char ( timestamp, text ) → text`
`to_char ( interval, text ) → text`
`to_char ( numeric_type, text ) → text`

You implemented only it for timestamp type. Can you specify it in jira ticket

babel/src/test/resources/sql/postgresql.iq
```diff
@@ -57,3 +57,8 @@ SELECT * FROM sal_emp;
 NAME, PAY_BY_QUARTER, SCHEDULE
 Bill, [10000, 10000, 10000, 10000], [[meeting, lunch], [training, presentation]]
 !ok
+
+select to_char(timestamp '2022-06-03 12:15:48.678', 'YYYY-MM-DD HH24:MI:SS.MS TZ');
```
dssysolyatin, line 61: Can you add more tests for `toChar(DataContext ctx, long timestamp, String pattern)` function in org.apache.calcite.test.SqlFunctionsTest ?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -583,6 +584,7 @@ Builder populate2() {
       defineMethod(TIME, "time", NullPolicy.STRICT);
 
       // Datetime formatting methods
+      map.put(TO_CHAR, new ToCharImplementor());
```
dssysolyatin, line 587: Use `defineMethod` instead of `ToChatImplementor`

dssysolyatin (changes requested):

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
zoudan, line 1: Shall we add some tests in SqlOperatorTest?

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -165,10 +173,17 @@ public enum FormatElementEnum implements FormatElement {
...
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      return String.format(Locale.ROOT, "%03d", calendar.get(Calendar.MILLISECOND));
+    }
+  },
   TZR("The time zone name") {
     @Override public String format(Date date) {
       // TODO: how to support timezones?
-      throw new UnsupportedOperationException();
+      return "";
```
zoudan, line 186: why we return empty string for this case

tindzk: We do so in order to match PostgreSQL's output when formatting `timestamp`s: https://dbfiddle.uk/VkeR6jUV

dssysolyatin:
LGTM 

If @julianhyde and @zoudan do not have any objection then I will merge it

zoudan:
LGTM

PR #3136, opened 2023-03-30
[CALCITE-4771] add TRY_CAST function (enabled in MSSQL library)

zoudan:
add TRY_CAST function (enabled in MSSQL library)

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -431,48 +431,59 @@ public class SqlOperatorTest {
         true);
   }
 
-  /** Tests that CAST and SAFE_CAST are basically equivalent but SAFE_CAST is
-   * only available in BigQuery library. */
-  @Test void testCastVsSafeCast() {
-    // SAFE_CAST is available in BigQuery library but not by default
+  /** Tests that CAST, SAFE_CAST and TRY_CAST are basically equivalent but SAFE_CAST is
+   * only available in BigQuery library and TRY_CAST is only available in MSSQL library. */
+  @Test void testCastVsSafeCastVsTryCast() {
+    // SAFE_CAST is available in BigQuery library but not by default.
+    // TRY_CAST is available in MSSQL library but not by default.
     final SqlOperatorFixture f0 = fixture();
     f0.checkScalar("cast(12 + 3 as varchar(10))", "15", "VARCHAR(10) NOT NULL");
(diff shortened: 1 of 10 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -431,48 +431,59 @@ protected SqlOperatorFixture fixture() {
         true);
   }
 
-  /** Tests that CAST and SAFE_CAST are basically equivalent but SAFE_CAST is
-   * only available in BigQuery library. */
-  @Test void testCastVsSafeCast() {
```
zinking, line 436: adding a dialect udf shouldn't have touched so many existing tests, can you explain why?

zoudan: Because I want to reuse most of the test code for `CAST` and `SAFE_CAST`, just like what we did in
https://github.com/apache/calcite/pull/3093

zinking: in my understanding parameterized tests are better suited for this.
I'd suggest some refactoring into these tests. let's see what others think

olivrlee: In this case the behavior of the functions (`SAFE_CAST`, `TRY_CAST`) are identical, it is just named differently across dialects, so I am a fan of the code reuse.

zoudan: @zinking All the tests in this class are in form of parameterized test except `testCastVsSafeCastVsTryCast`, such as `testCastToString`, do you mean I also change this test into a parameterized one?

zinking: right, that would be better in my unserstanding

zoudan: @zinking I have modified this, please have another look when you have time

zinking:
I am +1 now, @olivrlee  @libenchao  can some of you let the CI go and merge if no other concerns

libenchao:
@zinking Thanks for your review, I've approved the CI, will go through the PR shortly, and will merge it if everything looks good to me.

libenchao (approved):
+1, merging! 

@zoudan Thanks for your contribution, and @zinking @olivrlee thanks for the review!

PR #3141, opened 2023-04-03
[CALCITE-5624] Add support to parse array function call with Spark

MasseGuillaume:


core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
```diff
@@ -576,4 +576,26 @@ public interface SqlConformance {
    */
   @Experimental
   boolean allowCoercionStringToArray();
+
```
zoudan, line 579: remove a redundant empty line

core/src/main/java/org/apache/calcite/sql/validate/SqlConformanceEnum.java
```diff
@@ -439,4 +446,13 @@ public enum SqlConformanceEnum implements SqlConformance {
       return false;
     }
   }
+
+  @Override public boolean allowArrayFunction() {
+    switch (this) {
```
zoudan, line 451: `return this == SPARK`

MasseGuillaume: I'm following the same coding style as the rest of the file.

https://github.com/apache/calcite/blob/8d502cf421441a791db38ff8c87abc47fa45dd74/core/src/main/java/org/apache/calcite/sql/validate/SqlConformanceEnum.java#L441-L448

if we want to append to the list, the git diff will look better.

MasseGuillaume:
All tests passes, I'm ready for a code review.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -820,6 +820,22 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
           .withOperandTypeInference(InferTypes.RETURN_TYPE)
           .withKind(SqlKind.CONCAT2);
 
+  private static final SqlReturnTypeInference ARRAY_RETURN_TYPE =
```
julianhyde, line 823: these days we use methods rather than lambda constants. see e.g. `transformConvert`.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -4708,14 +4708,40 @@ SqlNode ArrayConstructor() :
 {
     <ARRAY> { s = span(); }
     (
-        LOOKAHEAD(1)
-        <LPAREN>
-        // by sub query "MULTISET(SELECT * FROM T)"
-        e = OrderedQueryOrExpr(ExprContext.ACCEPT_QUERY)
-        <RPAREN>
```
julianhyde, line 4715: That's a lot of 'validation' code in the parser. That's a code smell. Move it to the validator.

My hunch is that you don't need the `conformance.allowArrayFunction()` at all, anywhere. In the validator you can just look up a function called 'array' in the usual way.

core/src/main/java/org/apache/calcite/sql/validate/SqlConformanceEnum.java
```diff
@@ -77,6 +77,10 @@ public enum SqlConformanceEnum implements SqlConformance {
    * consistent with Presto. */
   PRESTO,
 
+  /** Conformance value that instructs Calcite to use SQL semantics
+   * consistent with Spark. */
```
julianhyde, line 81: Apache Spark

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3684,6 +3684,35 @@ private SqlDialect nonOrdinalDialect() {
...
+          + "FROM (VALUES (0)) t (ZERO)");
+  }
+
+  @Test void testArrayFunctionNullary() {
+    sqlSpark("SELECT ARRAY()")
+      .ok("SELECT ARRAY()\n"
+          + "FROM (VALUES (0)) t (ZERO)");
+  }
+
+  @Test void testArrayQueryInvalid() {
```
julianhyde, line 3706: this should be in SqlPArserTest

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7584,6 +7584,20 @@ private static void checkArrayConcatAggFuncFails(SqlOperatorFixture t) {
     f.checkFails("^Array[]^", "Require at least 1 argument", false);
   }
 
+  @Test void testArrayFunctionConstructor() {
+    final SqlOperatorFixture f = fixture();
```
julianhyde, line 7588: add tests where some or all arguments are null

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7584,6 +7584,20 @@ private static void checkArrayConcatAggFuncFails(SqlOperatorFixture t) {
     f.checkFails("^Array[]^", "Require at least 1 argument", false);
   }
 
+  @Test void testArrayFunctionConstructor() {
```
julianhyde, line 7587: rename test to `testArrayFunction`. it's not a constructor. 

add javadoc making clear that it's for the function in the Spark library.

test what happens when the Spark library is not enabled. (as I said elsewhere, I think you probably don't need the conformance setting.)

julianhyde (changes requested):

julianhyde:
Please change the jira subject to 'Add ARRAY function (enabled in Spark library)'. Parsing is only part of that.

core/src/main/codegen/templates/Parser.jj
MasseGuillaume: Ok, I will remove the conformance. I'm not sure I understand the implementation, do you want to always parse a function call? Then in the `SqlValidator`, when the array function is loaded (via the Spark library) validate that all the arguments are non-query expressions.

MasseGuillaume:
This PR is good to go, it just need a re-run of the CI for Julian's last commit (failed to grab build dependencies on window: https://github.com/julianhyde/calcite/actions/runs/4760351969/jobs/8460591740 )

PR #3142, opened 2023-04-06
[CALCITE-5634] Add LEAST, GREATEST for PostgreSQL

dssysolyatin:


tanclary:
I think it could be a good idea to add Postgres to the list of libraries these functions are tested for in `SqlOperatorTest`.

zoudan:
+1 for what @tanclary said, we should improve the test for these functions to ensure they works well for PostgreSQL

dssysolyatin:
It turned out that it is not enough to just enable this function. PostgreSQL's LEAST, GREATEST functions behave differently with NULL values than ORACLE or BigQuery:

From BigQuery documentation [1]:
```
They return NULL if any of the input parameters is NULL.
```

From PostgreSQL documentation [2]:
```
The result will be NULL only if all the expressions evaluate to NULL.
```

[1] https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions
[2] https://www.postgresql.org/docs/15/functions-conditional.html#FUNCTIONS-GREATEST-LEAST

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -383,14 +383,14 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
       SUBSTR.withKind(SqlKind.SUBSTR_POSTGRESQL);
 
   /** The "GREATEST(value, value)" function. */
-  @LibraryOperator(libraries = {BIG_QUERY, ORACLE})
+  @LibraryOperator(libraries = {BIG_QUERY, ORACLE, POSTGRESQL})
```
zoudan, line 386: could you please improve UT in `SqlOperatorTest`?

zoudan (changes requested):

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: @zoudan This looks like a small change and I have it here:
https://github.com/dssysolyatin/calcite/pull/1

What is the best way to move this PR forward?

mihaibudiu (approved):

PR #3143, opened 2023-04-11
[CALCITE-5639] eliminate IS_NOT_NULL check if LIKE expression on the same column is present

LeonidChistov:


chunweilei:
As discussed in the JIRA, can we find a more general solution to handle more cases?

LeonidChistov:
@chunweilei 

I have updated Pull Request with a more general solution. Ready for review.

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -3288,6 +3288,57 @@ private void checkSarg(String message, Sarg sarg,
         not(rexBuilder.makeCall(SqlStdOperatorTable.SIMILAR_TO, ref, literal("%"))));
   }
 
+  @Test void testSimplifyNullCheckInFilter() {
```
asolimando, line 3291: Can we split the test into multiple tests?

Upon failure it should be clear what went wrong, by bundling all in one test it's harder to tell. 

This said, I understand that it will repeat the first part but I still think it's a price worth paying

core/src/test/resources/sql/sub-query.iq
```diff
@@ -1748,7 +1748,7 @@ select sal from "scott".emp e
 (0 rows)
 
 !ok
-EnumerableCalc(expr#0..4=[{inputs}], expr#5=[RAND()], expr#6=[CAST($t5):INTEGER NOT NULL], expr#7=[2], expr#8=[MOD($t6, $t7)], expr#9=[3], expr#10=[=($t8, $t9)], expr#11=[IS NOT NULL($t4)], expr#12=[AND($t4, $t11)], expr#13=[OR($t10, $t12)], SAL=[$t1], $condition=[$t13])
+EnumerableCalc(expr#0..4=[{inputs}], expr#5=[RAND()], expr#6=[CAST($t5):INTEGER NOT NULL], expr#7=[2], expr#8=[MOD($t6, $t7)], expr#9=[3], expr#10=[=($t8, $t9)], expr#11=[OR($t10, $t4)], SAL=[$t1], $condition=[$t11])
```
asolimando, line 1751: I am OK with `IS NOT NULL($t4)` going away due to `AND($t4, $t11)`

`$condition=[$t11]` IIUC makes `$t11` in `AND($t4, $t11)` "redundant", so I understand that `$t11` goes away and that `$t4` can be placed in the `OR`.

What about `$t12`? I don't understand why it's being simplified.

Can you explain a bit the reasoning here, please?

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -1793,16 +1761,51 @@ private <C extends Comparable<C>> RexNode simplifyAnd2ForUnknownAsFalse(
...
-    for (RexNode notDisjunction : notTerms) {
-      terms.add(not(notDisjunction));
-    }
     // The negated terms: only deterministic expressions
     for (RexNode negatedTerm : negatedTerms) {
       if (termsSet.contains(negatedTerm)) {
         return rexBuilder.makeLiteral(false);
       }
     }
+    // Add the NOT disjunctions back in.
```
asolimando, line 1770: Just for my understanding, why you decided to move this part after the for loop above?

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -1793,16 +1761,51 @@ private <C extends Comparable<C>> RexNode simplifyAnd2ForUnknownAsFalse(
...
+      term.accept(collector);
+      for (RexInputRef ref : collector.refs) {
+        if (Strong.isNotTrue(term, ImmutableBitSet.of(ref.index))) {
+          strongOperands.add(ref);
+        }
+      }
+      RexUtil.FieldAccessFinder fieldAccessFinder = new RexUtil.FieldAccessFinder();
+      term.accept(fieldAccessFinder);
+      for (RexFieldAccess rexFieldAccess : fieldAccessFinder.getFieldAccessList()) {
+        if (Strong.of(ImmutableSet.of(rexFieldAccess)).isNotTrue(term)) {
```
asolimando, line 1790: Nit: I personally prefer to move complex `if` conditions out of the `if` itself, not a big deal to leave it but please double check

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -2599,6 +2602,22 @@ default boolean allowedInOr(RelOptPredicateList predicates) {
     }
   }
 
+  /**
+   * Visitor which finds all inputs used by an expressions.
+   */
+  private static class VariableCollector extends RexVisitorImpl<Void> {
+    private final Set<RexInputRef> refs = new LinkedHashSet<>();
```
asolimando, line 2609: Just curious, why specifically a `LinkedHashSet` here?

asolimando (changes requested):
LGTM, few comments but I like the PR

core/src/test/resources/sql/sub-query.iq
LeonidChistov: Since `$t12` is defined as `expr#12=[AND($t4, $t11)]`, and we have just eliminated `expr#11=[IS NOT NULL($t4)]` as redundant (and you agree with that), there is no sense in doing `AND` anymore.

So, what happens here AFAICS, is that we had condition `OR($t10, AND($t4, IS NOT NULL($t4)))` and managed to simplify it to `OR($t10, $t4)`.

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
LeonidChistov: I can do that, but actually it is not very clear for me, which style to follow. I think that majority of the tests in this file are collected to large methods, and also some of them have an honor to reside in their own method.

Are there any guidelines in Calcite regarding this?

core/src/test/resources/sql/sub-query.iq
asolimando: Thanks for the clarification, I agree with your analysis

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
LeonidChistov: That's probably does not matter a lot, I have just moved it a bit closer to the code where we actually need this.
I rely on having both `terms` and `notTerms` included to `terms` list during calculation of `strongOperands`.

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
LeonidChistov: I have just copied it from other place and didn't reflect enough on reasoning. Can be simplified with HashMap.

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
asolimando: AFAIK we don't have strong guidelines around this, we mostly act as you did, we look at existing code and we try to adapt to it.

Concerning the two conflicting styles, I see that older tests (> 5 years ago) are more inclined to put lots of cases inside a single test, newer code seems to be more for "single case - single test", which might be a bit more verbose for the "data preparation", but makes tests more readable and debuggable IMO.

I don't feel strongly about this but I favour the "newer" style. If nobody else chimes in and you prefer the current organization I won't object.

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
asolimando: I don't think it will make much of a difference, so it's fine as-is for me

asolimando (approved):
LGTM, let's give 48h more hours to allow others to check, if no further comments I will merge

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
libenchao: I have the same feeling with @asolimando, usually a `LinkedHashSet` would make me wonder what's its purpose to use it instead of a `HashSet`, which will increase the cost of understanding.

If we have a reason, we'd better write it into the comment. Else, I would prefer to use the most general one for the above reason. WDYT?

LeonidChistov: Agree, fixed.

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -1793,16 +1761,53 @@ private <C extends Comparable<C>> RexNode simplifyAnd2ForUnknownAsFalse(
...
+    for (RexNode term : terms) {
+      if (!RexUtil.isDeterministic(term)) {
+        continue;
+      }
+      final VariableCollector collector = new VariableCollector();
+      term.accept(collector);
+      for (RexInputRef ref : collector.refs) {
+        final boolean strong = Strong.isNotTrue(term, ImmutableBitSet.of(ref.index));
+        if (strong) {
+          strongOperands.add(ref);
```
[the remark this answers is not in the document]
asolimando: That was done on my asking to increase readability, if you think it does more harm than good I don't feel strongly about it @chunweilei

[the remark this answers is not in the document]
chunweilei: Got it! Feel free to move forward.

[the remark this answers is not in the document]
LeonidChistov: Thanks. @asolimando, can we merge the Pull Request then?

PR #3144, opened 2023-04-11
[CALCITE-5636] Consult root schema type map for coercion during validation

wnob:
Always look in the root schema's type map when inserting a `CAST` for automatic type coercion.

Other things to note:
- I modified `SqlFunctions.toTimestampWithLocalTimeZone(String)`, which is only invoked by the rex-to-lix translator to convert strings to `TIMESTAMP WITH LOCAL TIME ZONE` values, such as during coercion or casting in general. The method was implemented to address [CALCITE-1947](https://issues.apache.org/jira/browse/CALCITE-1947) but requires an explicit time zone component in the string literal to avoid a string-index-out-of-bounds exception in the constructor for `TimestampWithTimeZoneString`, which was originally named `TimestampWithLocalTimeZoneString` but renamed after the discussion in the Jira ticket to avoid confusion. It appears the problem with the name was address but the problem of semantics remains; one cannot cast a string without an explicit time zone to a timestamp w/ LTZ. I added a fallback that invokes `DateTimeUtils.timestampStringToUnixDate()`, which is the same function invoked by the rex-to-lix translator to convert strings to regular `TIMESTAMP` values. This effectively assumes UTC for all timestamp w/ LTZ conversions lacking an explicit time zone. Currently looking into getting the proper zone out of the `DataContext` but having trouble figuring out where it gets passed in.
- I set the default time zone for the BQ Quidem test to UTC, which causes `TIMESTAMP WITH LOCAL TIME ZONE` to assume UTC for the coercion of strings without an explicit time zone component (this is consistent with how BigQuery actually behaves).

### DIFF
--- a/babel/src/test/resources/sql/big-query.iq
+++ b/babel/src/test/resources/sql/big-query.iq
@@ -2152,24 +2152,79 @@ SELECT
 
 !ok
 
-# The following example shows the result of DATETIME_DIFF for two days
-# in succession. The first date falls on a Monday and the second date
-# falls on a Sunday. DATETIME_DIFF with the date part WEEK returns 0
-# because this time part uses weeks that begin on
-# Sunday. DATETIME_DIFF with the date part WEEK(MONDAY) returns
-# 1. DATETIME_DIFF with the date part ISOWEEK also returns 1 because
-# ISO weeks begin on Monday.
+# The following example shows the result of DATETIME_DIFF for many nearby dates.
+# 2017-12-18 is a Monday and the other date is between 2017-12-25 (a Friday)
+# and 2017-12-09 (a Saturday). The date part WEEK treats weeks as starting on
+# Sunday, whereas WEEK(MONDAY) and ISOWEEK both treat weeks as starting on
+# Monday.
 
-SELECT
-  DATETIME_DIFF('2017-12-18', '2017-12-17', WEEK) AS week_diff,
-  DATETIME_DIFF('2017-12-18', '2017-12-17', WEEK(MONDAY)) AS week_weekday_diff,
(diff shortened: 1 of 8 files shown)


babel/src/test/resources/sql/big-query.iq
```diff
@@ -2152,24 +2152,79 @@ SELECT
...
-  DATETIME_DIFF('2017-12-18', '2017-12-17', WEEK) AS week_diff,
-  DATETIME_DIFF('2017-12-18', '2017-12-17', WEEK(MONDAY)) AS week_weekday_diff,
-  DATETIME_DIFF('2017-12-18', '2017-12-17', ISOWEEK) AS isoweek_diff;
+# The following example shows the result of DATETIME_DIFF for many nearby dates.
+# 2017-12-18 is a Monday and the other date is between 2017-12-25 (a Friday)
+# and 2017-12-09 (a Saturday). The date part WEEK treats weeks as starting on
+# Sunday, whereas WEEK(MONDAY) and ISOWEEK both treat weeks as starting on
+# Monday.
+
+WITH Diffs AS (
```
julianhyde, line 2161: Let's stop writing this style of test. UNION ALL doesn't guarantee ordering.

Use a CTE with the input values - DATE '2017-12-18', DATE '2017-12-25' etc. - and change the query to print the input value of each row.

julianhyde:
Let's discuss in the Jira case. I am not convinced that this is a bug. I see a bunch of code changes you want to make, but I don't see a high-level rationale.

wnob:
This was barking up the wrong tree. A better solution to this problem (currently a draft) is here: https://github.com/apache/calcite/pull/3476

PR #3145, opened 2023-04-12
[CALCITE-5615] Program to run SQL Logic Tests for Calcite

mihaibudiu:
This PR essentially adds more than 7 million SQL tests to Calcite, from the SQL Logic Test Suite built by sqlite.

This addresses https://issues.apache.org/jira/browse/CALCITE-5615

This is a larger PR but it is for a completely self-contained program with a main entry point.
The README.md contained in this PR documents the program structure and its intended usage.
Using this program will require users to install separately the tests from the SQL Logic Test Suite: https://www.sqlite.org/sqllogictest/doc/trunk/about.wiki  

The testing program finds multiple bugs in the default configuration of the Calcite compiler, both crashes and incorrect outputs. The plan is to file separate issues for each of these bugs as they are found.
However, this program is also intended to be extended to test other Calcite-based compilers (or it could be used for other SQL query engines; there is nothing special about Calcite in the implementation.)


slt/src/main/java/org/apache/calcite/slt/SLTTestFile.java
```diff
@@ -0,0 +1,292 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+
+/**
+ * Represents the data from a .test file from the
+ * SqlLogicTest test framework.
+ */
+/*
```
zabetak, line 42: It's confusing to have javadoc and non-javadoc comment coming one after the other. Please combine them together and use proper javadoc annotations to ensure it displays properly.

slt/src/main/java/org/apache/calcite/slt/SLTTestFile.java
```diff
@@ -0,0 +1,292 @@
...
+ *         query I nosort
+ *         SELECT CASE WHEN c>(SELECT avg(c) FROM t1) THEN a*2 ELSE b*10 END
+ *           FROM t1
+ *          ORDER BY 1
+ *         ----
+ *         30 values hashing to 3c13dee48d9356ae19af2515e05e6b54
+ *
+ */
+public class SLTTestFile {
+  Logger logger = Logger.getLogger("SLTTestFile");
```
zabetak, line 68: Aren't loggers always `private static final`? Do we need to create one per instance?

slt/src/main/java/org/apache/calcite/slt/ExecutionOptions.java
```diff
@@ -0,0 +1,176 @@
...
+import java.io.IOException;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+@SuppressWarnings("CanBeFinal")
+public class ExecutionOptions {
```
zabetak, line 51: All public classes should have at least some top level javadoc. I think there are even checks for enforcing this.

zabetak (commented):
Quick pass leaving some comments along the way while trying to grasp what the contribution is about. I plan to leave also some questions comments under the JIRA ticket.

mbudiu-vmw:
I will only address @zabetak 's comments once we understand how this should be packaged; I have sent an email to the dev mailing list asking for suggestions. One thing I would encourage people to do in the meantime is to actually run the code (first on one or two test files, these are specified on the command-line, e.g., `select1.test`) and look at the bugs found.

zabetak:
@mbudiu-vmw Sure it makes sense to wait a bit. I will actually run the code during the weekend; I am already convinced that it is a very useful contribution :) I am just not sure yet where/how it should land. Thanks again for taking the time to contribute this to the community.

julianhyde:
I think this code could go into the existing `plus` module, not create a new `slt` module. Calcite has too many modules already.

zabetak (changes requested):

slt/src/main/java/org/apache/calcite/slt/Main1.java
```diff
@@ -0,0 +1,77 @@
...
+import net.hydromatic.sqllogictest.executors.HsqldbExecutor;
+import net.hydromatic.sqllogictest.executors.NoExecutor;
+import net.hydromatic.sqllogictest.TestLoader;
+
+import org.apache.calcite.slt.executors.CalciteExecutor;
+
+/**
+ * Execute all SqlLogicTest tests.
+ */
+public class Main1 {
```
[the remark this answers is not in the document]
mihaibudiu: It contains the 'main' entry point.
I am not sure how SLT will be used by Calcite yet, so the plan was to provide an executable that people could run to find bugs. Until the bugs it finds are fixed I don't know exactly how it can be integrated in the test suite.

slt/src/main/java/org/apache/calcite/slt/Main1.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /*
+    URL r = Thread.currentThread().getContextClassLoader().getResource("test");
+    if (r == null) {
+      out.println("Cannot find resources");
+      return 1;
+    }
+     */
+    URL r = new URL("/home/mbudiu/git/sqllogictest/test");
```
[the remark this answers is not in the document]
mihaibudiu: Sorry, that's a bug left over from testing, the commented-out code should be used instead.

settings.gradle.kts
```diff
@@ -83,7 +83,8 @@ include(
     "spark",
     "splunk",
     "testkit",
-    "ubenchmark"
+    "ubenchmark",
+    "slt"
```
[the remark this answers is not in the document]
mihaibudiu: I am not sure what "plus" is. Should I just move the code into the plus project using the org.apache.calcite.slt package?

slt/src/test/java/org/apache/calcite/slt/TestCalcite.java
```diff
@@ -0,0 +1,87 @@
...
+import java.io.IOException;
+import java.io.PrintStream;
+import java.nio.charset.StandardCharsets;
+
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+public class TestCalcite {
```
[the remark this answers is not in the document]
mihaibudiu: This class was created only to make sure that this project correctly extends sql-logic-test. 
If we plan to integrate this tool into the Calcite testsuite we have to devise a better name and design for this class.

mihaibudiu:
@zabetak thank you for the comments, I will address them once sql-logic-test 2.0 is available in my next PR.
But I would appreciate help to decide how to package this tool.

zabetak:
> But I would appreciate help to decide how to package this tool.

@mihaibudiu From my perspective (and Julian's I think) this PR should contain something equivalent to what is now `TestCalcite` inside the `plus` module and allow all SQL logic tests that pass to run. It should probably contain `CalciteExecutor` but nothing more. We just want to run to be able to run the tests as part of the Calcite suite but nothing more for now. If you want I can push the relevant commits following this idea in this PR.

mihaibudiu:
With the latest commit this is testing the CalciteExecutor against the candidate 0.2 release of sql-logic-test.

plus/build.gradle.kts
```diff
@@ -25,9 +25,11 @@ dependencies {
     implementation("com.teradata.tpcds:tpcds")
     implementation("io.prestosql.tpch:tpch")
     implementation("net.hydromatic:chinook-data-hsqldb")
+    implementation("net.hydromatic:sql-logic-test:0.2")
     implementation("net.hydromatic:tpcds")
     implementation("org.apache.calcite.avatica:avatica-server")
     implementation("org.hsqldb:hsqldb")
+    implementation("org.reflections:reflections:0.10.2")
```
zabetak, line 32: Why do we need to declare this? Usually transitive dependencies should be discovered automatically?

plus/build.gradle.kts
```diff
@@ -25,9 +25,11 @@ dependencies {
     implementation("com.teradata.tpcds:tpcds")
     implementation("io.prestosql.tpch:tpch")
     implementation("net.hydromatic:chinook-data-hsqldb")
+    implementation("net.hydromatic:sql-logic-test:0.2")
```
zabetak, line 28: Should this rather be a `testImplementation` dependency?

zabetak: Also probably the version ("0.2") should not be controlled here. That's why we have the gradle.properties file.

plus/src/test/java/org/apache/calcite/slt/TestCalcite.java
```diff
@@ -0,0 +1,86 @@
...
+import java.io.PrintStream;
+import java.nio.charset.StandardCharsets;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+/**
+ * Tests using sql-logic-test suite.
+ */
+public class TestCalcite {
```
zabetak, line 36: The class should be reworked. 

1. We want to run all sql-logic-test that pass and not only `select1.test`. 
2. When there is a regression we want to have a clear way to know which test and which file failed. (optional but useful)
3. We only care about tests using the Calcite executor; anything else is out of context of this pr.

plus/src/test/java/org/apache/calcite/slt/executors/CalciteExecutor.java
```diff
@@ -0,0 +1,170 @@
...
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Set;
+import javax.sql.DataSource;
+
+/**
+ * Executor for SQL logic tests using Calcite's JDBC adapter.
+ */
+public class CalciteExecutor extends SqlSltTestExecutor {
```
zabetak, line 46: I think I left some comments regarding this class in a previous review.

zabetak (changes requested):

plus/src/test/java/org/apache/calcite/slt/TestCalcite.java
mihaibudiu: The main problem is that there are *lots* of failures. Only in these 1000 tests from select1.test there are 153 failures. They are due to a small number of underlying causes. If we run all 5M tests in this call, with this expected failure rate there will be 50K failures, and the output will cause the PrintStreams used in launchSqlLogicTest to run out of memory.

Until we implement https://github.com/hydromatic/sql-logic-test/issues/16 we probably cannot run all the tests.

When all the found Calcite bugs are fixed we can modify the test here to assert that there are no failures, but until that happens I am unsure how to use this test. Right now this code is most useful if it outputs a diagnosis for human consumption.

plus/src/test/java/org/apache/calcite/slt/executors/CalciteExecutor.java
mihaibudiu: I looked everywhere and I could not find other comments about this class. I apologize if I didn't address them.

plus/src/test/java/org/apache/calcite/slt/TestCalcite.java
zabetak: There is nothing preventing us from re-redirecting the streams to files and not keep them in byte buffers. Hopefully by doing this we should avoid running out of memory.

We don't necessarily need to assert that there are no failures but more like that the number of failures doesn't change. Unless the number of failures goes down which I guess is a good thing meaning that we fixed a bug.

As far as I understand, point 2 that I suggested above is hard to implement at this stage given that there are no test files that pass completely so we can definitely defer to another JIRA.

julianhyde:
Rather than checking that the number of failures doesn't change, it's better to check that there are no failures other than the list of 'expected failures'

You don't want the situation where you fix one test and introduce a new failure. It's much better to have a list of known failures rather than just a count.

mihaibudiu:
Currently the "execute" method from Main only returns an integer and writes all diagnostic to the supplied streams.
We have to make it return the actual TestStatistics object.
So this will require another release.

mihaibudiu:
It took about 6 minutes to run 10K tests using CalciteExecutor, so by extrapolation it would take about 2 days to run all tests.
Memory consumption increases slightly with time. The failures are saved in a list which keeps growing. I am not sure whether there could be leaks somewhere else.

zabetak:
> It took about 6 minutes to run 10K tests using CalciteExecutor, so by extrapolation it would take about 2 days to run all tests.

Here are some thoughts regarding duration and frequency:
* Tests runtime ~5 minutes -> could run on every PR;
* Tests runtime between 1h and 3h -> could possibly run once daily;
* Tests runtime under 6h -> could run once per week;
* Tests runtime spanning days -> per request/ not automated (maybe before release)
It's up to us I guess to divide them into meaningful buckets (e..g, those that are likely to break more often should run on every PR) and setup the relevant CI infra.

Obviously we don't have to setup everything as part of this PR but it would be nice if we could at least put the 5minute bucket in place.

> Memory consumption increases slightly with time.

I guess we can find ways to address memory pressure; some quick ideas:
* Run each test file separately; this in principle should reset the memory after each run.
* Provide options/APIs to save failures in files (instead of keeping ever growing in memory structures)

mihaibudiu:
I think having many small tests is probably better. There are 622 test files in the project.
We could write a small script to generate a test for each one (or a group of files).
This could also be beneficial if the tests can run in parallel.

However, JUnit5 has a capability called DynamicTests using the `@TestFactory` annotation, but Calcite does not currently use it: https://www.baeldung.com/junit5-dynamic-tests.
I will try to prototype this solution to see what happens. With DynamicTests we could even choose how many tests to run by sampling the test file list.

mihaibudiu:
I have amended this PR to use DynamicTests. This needs https://github.com/hydromatic/sql-logic-test/pull/36 in order to work, so a new release of sql-logic-tests is necessary. This seems a decent way to run all these 622 tests. On my machine it runs 8 tests at a time, each taking about 5 minutes. The errors discovered are just printed on stderr right now.

julianhyde:
Can you get anything useful committed to Calcite without a new release of sql-logic-test? I have put a lot of effort into this project so far and I want to see some returns before I do another release.

mihaibudiu:
Well, you said "release often". Is there something that prevents a new release of sql-logic-test?
Unfortunately the code was missing some getters for non-public members and I don't know how to work around that.
Only by writing the CalciteExecutor code I figured out the missing features.
(Our own compiler also could use the new features in sql-logic-test; I have converted our implementation to use it.)
I am more worried about this PR: is this way of running the tests useful? These tests take a fairly long time and only print diagnostic on stderr about the encountered bugs, but they can't really fail.

julianhyde:
It takes about an hour of my weekend to make each release.

mihaibudiu:
I apologize, didn't realize it's that involved. Any way I can help?
I hope this will be the last one for a while.

zabetak:
@mihaibudiu In Calcite we haven't used Dynamic Tests so far but we do make use of [Parameterized Tests](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests). The two kinds of tests are very similar so I am fine with any of them; whatever is most convenient for you.

Although a new release may bring in some good enhancements I am not sure if it is strictly mandatory at this point. We know exactly the test files so we could hard code those here and generate a dynamic test for each.

> Rather than checking that the number of failures doesn't change, it's better to check that there are no failures other than the list of 'expected failures'

I will try to elaborate a bit more on the comment by Julian. I think what he had in mind can achieved with the following setup.

For each test file that we decide to run we can create a "golden" file under `src/test/resources/sqllogictest` with the known/expected failures. Once we run the test we create the "actual" file with the failures and compare "golden" with "actual"; if they differ the test fails otherwise it passes. This is very much inline with how `DiffRepository` works so you could try to take advantage of it in these tests. If `DiffRepository` is not pertinent then you could opt for a simple diff or plain string equality comparison between the files.

However, given that we have a large number of failures it might be impractical to create 622 golden files cause we might bloat the repository with test files containing errors. Currently everything under `src/test/resources` (data, golden files, logging configs, etc.) is ~3.5MB.
```
find . -wholename "*/src/test/resources/*" -exec du -ch {} + | grep total
```
If the sqllogic golden files containing the errors are around ~10MB it would be acceptable but going much more beyond that can be problematic and we should probably rethink this. Using compression to keep the file size low should be fine I guess.

julianhyde:
> Any way I can help?

It would be helpful if you merge any pending changes to `main`, then create a branch and PR for Release 0.3. The content would be similar to commit 5195b8d for Release 0.2.

Once the release notes etc. are done I can run the Maven process to publish to Maven Central.

mihaibudiu:
I hope we're close to a working solution for this PR.
You will notice that the CalciteExecutor has been reduced to just a few lines of code. 
Most of the complexity is in the test code - packaging a very large number of tests and failures into a compact representation. The README.md has details about how this tool can be used not only for regression testing but also for diagnosing bugs in Calcite.

zabetak:
@mihaibudiu Thanks for pushing this forward. Indeed I think we are pretty close. 

One basic point that we need to agree on before I do full review of the new part is the duration & frequency of running the new tests. I see that the new tests are tagged as `Slow`. Currently we run `Slow` tests on every commit that lands in master and without the changes under this PR they take ~30minutes.

How much time do the new test take at the moment and how often do we want to run them? We should take into account that things are usually much faster locally than in CI.

As I mentioned earlier (https://github.com/apache/calcite/pull/3145#issuecomment-1580300860) we don't need to address everything as part of this PR. However, I would like to have some of the tests enabled and running as part of the unit tests.

mihaibudiu:
On my machine using 8 parallel threads the tests take about 5-6 hours.
I am not familiar with the tiers of tests in Calcite, so I don't know what tag should be used., but if there is a slower one it should be even easier.

zabetak:
Apart from `Slow` there is no other speed tag currently in Calcite. 

Let's keep it simple for the time being and use three buckets:
* Fast bucket (under a minute) with no special annotation running in unit tests
* Slow bucket (5-10 minutes) tagged with the existing slow annotation
* Very slow bucket (over 10 minutes) disabled for the time being; to be discussed and addressed in a follow-up PR.

WDYT?

mihaibudiu:
"Very slow" sounds fine to me. This could be run just prior to a release or maybe once a month.
The tool is really designed for human-in-the-loop: someone has to find the bugs and file issues.
I want to add some capabilities to the sql-logic-test project to make this task easier (https://github.com/hydromatic/sql-logic-test/issues/22). I hope to do it in the next few days.
The main question is what to do when this tool finds regressions. Probably we need to bisect the commits to find the source of regressions.

zabetak:
The only blocking thing for getting this PR merged is putting in place the "Fast" and "Slow" bucket. The "Very Slow" bucket as well as finding and reporting regressions can be tackled in a follow-up discussion/JIRA ticket.

mihaibudiu:
I wasn't exactly sure how to handle this, so what I have done is to:
- run just one test file as a slow test, takes a couple of minutes
- disable the full testsuite using the `@Disabled` annotation
The idea is to run the testsuite by hand as desired, but not part of regular testing.
Is this an acceptable solution?

plus/src/test/java/org/apache/calcite/slt/SqlLogicTestsForCalciteTests.java
```diff
@@ -0,0 +1,312 @@
...
+    return net.hydromatic.sqllogictest.Main.execute(options, args);
+  }
+
+  TestSummary shortSummary(String file, TestStatistics statistics) {
+    return new TestSummary(file, statistics.getPassedTestCount(), statistics.getFailedTestCount());
+  }
+
+  // The following tests currently timeout during execution.
+  // Technically these are Calcite bugs.
+  Set<String> timeout = new HashSet<String>() {
```
zabetak, line 233: Not worth sub-classing `HashSet` for this. Use `ImmutableSet.of` or `ImmutableSet.builder`. Same in other places where this pattern is used.

plus/src/test/java/org/apache/calcite/slt/SqlLogicTestsForCalciteTests.java
```diff
@@ -0,0 +1,312 @@
...
+    testSummaries.add(summary);
+  }
+
+  @Test @Tag("slow")
+  public void runOneTestFile() throws IOException {
+    runOneTestFile("select1.test");
+  }
+
+  @TestFactory @Disabled("This takes very long, should be run manually")
+  List<DynamicTest> runAllTests() {
```
zabetak, line 275: Since `runAllTests()` is `@Disabled` there is a risk that the code gets broken if it is not run in every day basis. 
I would like to suggest a slight refactoring to ensure that we maximize the code that is tested daily.

```
@TestFactory @Tag("slow")
List<DynamicTest> testSlow() throws IOException {
  return generateTests(ImmutableSet.of("select1.test"));
}

@TestFactory @Disabled("This takes very long, should be run manually")
List<DynamicTest> testAll() throws IOException {
  return generateTests(net.hydromatic.sqllogictest.Main.getTestList());
}

private static List<DynamicTest> generateTests(Set<String> testFiles) {
// This is the code in runAllTests
}
```

plus/src/test/java/org/apache/calcite/slt/SqlLogicTestsForCalciteTests.java
```diff
@@ -0,0 +1,312 @@
...
+
+    void read(InputStream stream) throws IOException {
+      try (BufferedReader reader =
+               new BufferedReader(new InputStreamReader(stream, StandardCharsets.UTF_8))) {
+        reader.lines().forEach(line -> {
+          TestSummary summary = TestSummary.parse(line);
+          if (summary != null) {
+            this.add(summary);
+          } else {
+            System.err.println("Could not parse line " + line);
```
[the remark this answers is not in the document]
zabetak: For instance when tests run in CI we don't want to clutter the console with error messages.

plus/src/test/java/org/apache/calcite/slt/SqlLogicTestsForCalciteTests.java
```diff
@@ -0,0 +1,312 @@
...
+     * @return 'true' if other contains regressions.
+     * @param other  Test results to compare with.
+     *               'other' can contain only a subset of the tests.
+     */
+    boolean regression(AllTestSummaries other) {
+      boolean regression = false;
+      for (TestSummary summary: other.testResults.values()) {
+        TestSummary original = this.testResults.get(summary.file);
+        if (original == null) {
+          System.err.println("No historical data for test " + summary.file);
```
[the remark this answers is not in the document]
mihaibudiu: I will remove the prints, but the hydromatic project also prints stuff, and we currently pass System.out and System.err for the PrintStream values it expects in `launchSqlLogicTest`.

plus/src/test/java/org/apache/calcite/slt/SqlLogicTests.java
```diff
@@ -0,0 +1,337 @@
...
+        return null;
+      }
+    }
+
+    @Override public String toString() {
+      return this.file + ":" + this.failed;
+    }
+
+    /**
+     * Check if the 'other' TestSummaries indicate a regressions
```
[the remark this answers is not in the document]
zabetak: Please ignore, I am applying small fixes myself.

plus/src/test/java/org/apache/calcite/slt/SqlLogicTests.java
```diff
@@ -0,0 +1,337 @@
...
+    // for each error caused by an exception.
+    TestStatistics res = launchSqlLogicTest("-v", "-e", "calcite", testFile);
+    assertThat(res, notNullValue());
+    assertThat(res.getParseFailureCount(), is(0));
+    assertThat(res.getIgnoredTestCount(), is(0));
+    assertThat(res.getTestFileCount(), is(1));
+    res.printStatistics(System.err);  // Print errors found
+    TestSummary summary = this.shortSummary(testFile, res);
+    boolean regression = goldenTestSummaries.regression(summary);
+    Assumptions.assumeFalse(regression, "Regression in " + summary.file);
```
zabetak, line 288: I thought we were always checking the results against the golden file; if that's the case then why do we have the assertThat statements? Do we need both assertThat and assumeFalse, if yes why?

zabetak: OK I understood, the assertThat are basically to ensure that the file was processes correctly.

zabetak (approved):
LGTM, I am doing some final touches locally and will merge this to master in the next few days.

mihaibudiu:
We could also regenerate the golden file to account for the fixes made in the meantime. But I am on vacation for the next week so I can't do it right away.

zabetak:
@mihaibudiu I am now testing everything along with my changes here: https://github.com/zabetak/calcite/tree/slt

If run comes back green I will merge it and we can work on follow-ups after that.

PR #3149, opened 2023-04-14
[CALCITE-5646] JoinDeriveIsNotNullFilterRule incorrectly handles COAESCE in join condition

LeonidChistov:


chunweilei (commented):
The fix looks good to me. Could you please add more tests for it, such as NVL?

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
```diff
@@ -63,9 +63,11 @@ public JoinDeriveIsNotNullFilterRule(Config config) {
 
     final ImmutableBitSet.Builder notNullableKeys = ImmutableBitSet.builder();
     RelOptUtil.conjunctions(join.getCondition()).forEach(node -> {
```
zabetak, line 65: Is there any benefit in splitting the condition in conjunctions and testing strongness individually for each? Maybe we can remove this decomposition.

core/src/main/java/org/apache/calcite/plan/Strong.java
```diff
@@ -149,7 +149,9 @@ public static boolean allStrong(List<RexNode> operands) {
     return operands.stream().allMatch(Strong::isStrong);
   }
 
-  /** Returns whether an expression is definitely not true. */
+  /** Returns whether the analyzed expression will definitely not return true
+   * (equivalently, will definitely not return null or false) if
```
zabetak, line 153: The javadoc here was rather fine. The part that is broken is L82 `(equivalently, will definitely not return null or false)`. It should be `(equivalently, will definitely return null or false)`.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
```diff
@@ -62,9 +62,9 @@ public JoinDeriveIsNotNullFilterRule(Config config) {
     final RelMetadataQuery mq = call.getMetadataQuery();
 
     final ImmutableBitSet.Builder notNullableKeys = ImmutableBitSet.builder();
-    RelOptUtil.conjunctions(join.getCondition()).forEach(node -> {
-      if (Strong.isStrong(node)) {
-        notNullableKeys.addAll(RelOptUtil.InputFinder.bits(node));
+    RelOptUtil.InputFinder.bits(join.getCondition()).forEach(bit -> {
+      if (Strong.of(ImmutableBitSet.of(bit)).isNotTrue(join.getCondition())) {
```
zabetak, line 66: This can also become a bit shorter I guess by using the static util: `Strong.isNotTrue(join.getCondition, ImmutableBitSet.of(bit))`

core/src/main/java/org/apache/calcite/plan/Strong.java
LeonidChistov: Yes, sorry. Thanks for corrections.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveIsNotNullFilterRule.java
LeonidChistov: Agreed.

PR #3151, opened 2023-04-14
CALCITE-5647 Use mq.getRowCount(rel) instead of rel.estimateRowCount(mq)

adamkennedy:
This PR corrects a row count call to use the RelMetadataQuery instead of going directly to rel.estimateRowCount.

Using rel.estimateRowCount directly will behave incorrectly when using a cardinality estimator that is implemented entirely in a MetadataHandler<RowCount> class, as it will evade the metadata handler's implementation of getRowCount and fall through to the underlying default method for Values.

This corrects the problem by using mq.getRowCount instead of rel.estimateRowCount. In the default case this results in the same value but in the custom cardinality model case the expected handler is called.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdPopulationSize.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdPopulationSize.java
@@ -113,7 +113,7 @@ public class RelMdPopulationSize
   public Double getPopulationSize(Values rel, RelMetadataQuery mq,
       ImmutableBitSet groupKey) {
     // assume half the rows are duplicates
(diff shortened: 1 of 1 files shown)


chunweilei (commented):
Could you please add a test to show the difference?

jduo:
Hi @adamkennedy, @libenchao  and others.
I'm continuing the work on this PR at #3632.

(That PR has the unit test and is rebased on main).

mihaibudiu:
Should this PR be closed?

jduo:
> Should this PR be closed?

Yes, please do.

PR #3155, opened 2023-04-15
[CALCITE-5650] Preserve typeSystem field when changing dialect in SqlToRelConverterTest

mbudiu-vmw:


rubenada:
Fix lgtm, there are some conflicts to be handled though.

PR #3159, opened 2023-04-18
[CALCITE-5655] Intermediate table alias should be different to avoid wrong field reference lookup in subquery remove phase

herunkang2018:


core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6252,6 +6286,32 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
         .check();
   }
 
+  @Test void testExpandProjectInWithTwoCorrelatedSubQueries() {
```
herunkang2018, line 6289: These two test cases for expand project in is not affected by this issue, I add these to improve test converage.

rubenada:
The fix looks good. There are some style violations to be corrected.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
libenchao: Thanks for the clarification, I think we can mention this in the commit message when we merging.

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -241,21 +247,21 @@ private static RexNode rewriteSome(RexSubQuery e, Set<CorrelationId> variablesSe
...
                 literalFalse,
                 builder.call(SqlStdOperatorTable.IS_TRUE,
                     builder.call(RexUtil.op(op.comparisonKind),
-                        e.operands.get(0), builder.field("q", "m"))),
+                        e.operands.get(0), builder.field(qAlias, "m"))),
                 literalTrue,
-                builder.greaterThan(builder.field("q", "c"),
-                    builder.field("q", "d")),
+                builder.greaterThan(builder.field(qAlias, "c"),
+                    builder.field(qAlias, "d")),
```
libenchao, line 261: Currently `RelBuilder#field` would return the first one when there is multiple results, I think we'd better to throw in this case. It would be a breaking change, what do you think about this @rubenada ?

herunkang2018:
@rubenada Thanks for reminding it, I will fix it soon.

herunkang2018:
@libenchao @rubenada All fixed, please help to review if you have time, thanks very much.

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
rubenada: Uhm... I think that might be a sensitive, breaking change. I'm not sure about it.
In any case, I think that should be discussed in a separated thread/ticket, and is out of the scope of the current PR.

rubenada:
LGTM. I'd suggest to align the Jira title to the PRs title.

libenchao (approved):
+1

herunkang2018:
@rubenada I have aligned the [Jira](https://issues.apache.org/jira/browse/CALCITE-5655) title to the PRs title. Please help to review if you have time.

rubenada:
Thanks @herunkang2018 . PR looks in a good shape.
@libenchao since you assigned the PR to yourself, I'll let you finalize it and merge it, ok? (needs squashing commits into a single one).

rubenada (approved):

libenchao:
@rubenada Sure, will merge this today. (Besides, since we have a mechanism that we give merits to the committer who merges the PR in our board report,  please feel free to do the merging if the PR is in good shape and everyone reaches consensus, the merit should not be scoped to the first one who assigns it to him/herself)

PR #3162, opened 2023-04-18
CALCITE-5659/ Allow CAST in an INTERVAL literal

olivrlee:


zoudan:
There is a redundant backslash in your commit message

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -5223,6 +5223,8 @@ private static Matcher<SqlNode> isCharLiteral(String s) {
         .ok("INTERVAL '1:2:3' HOUR TO SECOND");
     f.sql("interval \"1:2:3\" hour to second")
         .ok("INTERVAL '1:2:3' HOUR TO SECOND");
+    f.sql("interval cast(1 as INT64) year")
```
zoudan, line 5226: Shall we add more test to ensure the calculation results are within expectations, not only the parse stage

olivrlee: After some consideration I am going to try completing with a different approach, so I'm moving this back to draft. 
Ideally it can handle anything between `INTERVAL` and the final token (i.e. `YEAR`), we don't want to be adding to the parser for every shape that comes in one by one.

PR #3163, opened 2023-04-18
[CALCITE-5402] RelToSql generates invalid code if left and right side field names clash

LeonidChistov:
https://issues.apache.org/jira/browse/CALCITE-5402

The fix changes approach of SQL tree generation for semi-joins.

Let us consider differences of old and new implementations for the example below:

```
SemiJoin [DEPT.DEPTNO = EMP.DEPTNO]
     Filter [DEPTNO > 10]
         Scan[DEPT]
     Scan[EMP]
```

**Old implementation**

1. Construct trees for left side of semi-join & right side of semi-join:
    Left tree SQL: `SELECT * FROM DEPT WHERE DEPTNO > 10` // Declares needed alias `t`
    Right tree SQL: `SELECT * FROM EMP` 
2. Construct condition sub-query based on left & right sides alias contexts:  `t.DEPTNO = EMP.DEPTNO`
3. Replaces prefixed references to left table fields with bare fields: `DEPTNO =  EMP.DEPTNO`
4. Construct filter expression matching semi-join semantics: `EXISTS (SELECT 1 FROM EMP WHERE DEPTNO = EMP.DEPTNO`
5. Construct new filter expression for left side: `DEPTNO > 10 AND EXISTS (SELECT 1 FROM EMP WHERE DEPTNO = EMP.DEPTNO`
6. Returns SQL (with / declared needed alias `t`)
     ```
     SELECT * FROM DEPT 
     WHERE DEPTNO > 10 AND EXISTS (SELECT 1 FROM EMP WHERE DEPTNO = EMP.DEPTNO)
    ``` 

This approach does not work well when field names collide in left and right sides of semi-join (like DEPTNO in our case).
The problem here is that DEPTNO inside EXISTS is being resolved as EMP.DEPTNO, not as DEPT.DEPTNO.

**New implementation**

1. Construct trees for left side of semi-join & right side of semi-join:
    Left tree SQL: `SELECT * FROM (SELECT * FROM DEPT WHERE DEPTNO > 10) AS t`
    Right tree SQL: `SELECT * FROM EMP` 
2. Construct condition sub-query based on left & right sides alias contexts:  `t.DEPTNO = EMP.DEPTNO`
5. Construct new filter expression: `EXISTS (SELECT 1 FROM EMP WHERE DEPTNO = EMP.DEPTNO`
6. Returns SQL
     ```
     SELECT * FROM
     (SELECT * FROM DEPT WHERE DEPTNO > 10) AS t
     WHERE EXISTS (SELECT 1 FROM EMP WHERE t.DEPTNO = EMP.DEPTNO)
    ``` 

**Ideal implementation**

Ideal implementation would create following SQL as the result

     ```
     SELECT * FROM DEPT AS t
     WHERE DEPTNO > 10 AND EXISTS (SELECT 1 FROM EMP WHERE t.DEPTNO = EMP.DEPTNO)
    ``` 

See https://github.com/apache/calcite/pull/3175. It produces required SQL, but Calcite code becomes complicated.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
@@ -80,7 +80,6 @@ import org.apache.calcite.sql.fun.SqlInternalOperators;
 import org.apache.calcite.sql.fun.SqlSingleValueAggFunction;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParserPos;
-import org.apache.calcite.sql.util.SqlShuttle;
 import org.apache.calcite.sql.validate.SqlValidatorUtil;
 import org.apache.calcite.util.ImmutableBitSet;
 import org.apache.calcite.util.Pair;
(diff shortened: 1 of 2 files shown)


zinking:
nicely written up!

```
This approach does not work well when field names collide in left and right sides of semi-join (like DEPTNO in our case).
```
did you mean the old resolution won't work even adding alias to table names ?

LeonidChistov:
@zinking 

Sorry, I don't understand your question. Can you give more details?

zinking:
> @zinking
> 
> Sorry, I don't understand your question. Can you give more details?

in your description, you mentioned before the new approach that old approach wont work. I am just asking why it wont work, does it not work even after adding alias to the subqueries ?

LeonidChistov:
@zinking 

I mean that old approach works wrong when field names from left hand side and right hide collide:

```
SELECT * FROM DEPT 
WHERE DEPTNO > 10 AND EXISTS (SELECT 1 FROM EMP WHERE DEPTNO = EMP.DEPTNO)
```
If we improve it to generate query like 

```
SELECT * FROM DEPT 
WHERE DEPTNO > 10 AND EXISTS (SELECT 1 FROM EMP WHERE DEPT.DEPTNO = EMP.DEPTNO)
```
then everything will be OK.

But my point is that it is not so simple to do this improvement. Initially I didn't understand at all how to do it, but now I have another Draft Pull Request that does required things: https://github.com/apache/calcite/pull/3175/.

But I have concern that new code is complicated and thus more error prone.

PR #3167, opened 2023-04-19
 [CALCITE-5660] Add array subscript operators OFFSET, ORDINAL, SAFE_OFFSET, SAFE_ORDINAL (enabled for BigQuery)

tanclary:
Add support for BigQuery Array Subscript Operators `OFFSET`, `ORDINAL`, `SAFE_OFFSET`, and `SAFE_ORDINAL`.

These operators are similar to the existing `ITEM` operator used to index an array. For this PR, that implementation was parameterized to accept different offsets and out-of-bounds behavior. 

[BigQuery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/operators#array_subscript_operator)

Any comments as always are appreciated, thanks!


core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -678,7 +682,13 @@ Builder populate2() {
       final RexCallImplementor value = new ValueConstructorImplementor();
       map.put(MAP_VALUE_CONSTRUCTOR, value);
       map.put(ARRAY_VALUE_CONSTRUCTOR, value);
-      map.put(ITEM, new ItemImplementor());
+
+      // Each subscript operator has a corresponding offset and 'safe' value.
```
julianhyde, line 686: Better to move `offset` and `safe` from `ItemImplementor` into `SqlItemOperator` (make them public final fields)

Right now you're leaving a little too heavily on object identity (and there's a danger that future maintainers will branch based on the name of the operator, which will be bad).

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3372,11 +3372,16 @@ public static String replace(String s, String search, String replacement) {
 
   /** Helper for "array element reference". Caller has already ensured that
    * array and index are not null. Index is 1-based, per SQL. */
-  public static @Nullable Object arrayItem(List list, int item) {
-    if (item < 1 || item > list.size()) {
-      return null;
+  public static @Nullable Object arrayItem(List list, int item, int offset, boolean safe) {
+
```
julianhyde, line 3376: remove the blank line

julianhyde (commented):
Change the commit message to 'Add array subscript operators OFFSET, ORDINAL (enabled for BigQuery)', just to conform to the usual format.

Are people allowed to call these operators outside of an array? If so, write tests to make sure they can't, even in BigQuery.

Is `array[OFFSET(2)+OFFSET(3)]` valid? Is `array[OFFSET(2 + 3)]` valid?

Good work; this PR looks very close.

tanclary:
> Change the commit message to 'Add array subscript operators OFFSET, ORDINAL (enabled for BigQuery)', just to conform to the usual format.
> 
> Are people allowed to call these operators outside of an array? If so, write tests to make sure they can't, even in BigQuery.
> 
> Is `array[OFFSET(2)+OFFSET(3)]` valid? Is `array[OFFSET(2 + 3)]` valid?
> 
> Good work; this PR looks very close.

@julianhyde 
- Commit message should be changed.
- These operators are solely for array expressions from what I can tell in the docs, so I parameterized the operand type checker as well. I added some test cases for this. 
- OFFSET(2) + OFFSET(3) should not be valid (and it isn't). OFFSET(2+3) should be valid (and it is).

julianhyde:
Should `SELECT OFFSET(1)` be valid? I don't think so, because `OFFSET` is being used outside of an array. But please check in BigQuery and test accordingly.

tanclary:
> Should `SELECT OFFSET(1)` be valid? I don't think so, because `OFFSET` is being used outside of an array. But please check in BigQuery and test accordingly.

@julianhyde No it should not be valid. I've checked in BigQuery and added a quidem test as well.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -678,7 +683,13 @@ Builder populate2() {
       final RexCallImplementor value = new ValueConstructorImplementor();
       map.put(MAP_VALUE_CONSTRUCTOR, value);
       map.put(ARRAY_VALUE_CONSTRUCTOR, value);
+
+      // Each subscript operator has a corresponding offset and 'safe' value.
       map.put(ITEM, new ItemImplementor());
+      map.put(OFFSET, new ItemImplementor());
```
dssysolyatin, line 689: It would be better to use a single instance of ItemImplementor instead of creating separate instances for each operator. Apart from that, everything else looks good.

dssysolyatin:
@tanclary Please fix conflicts and if there aren't objections, I will merge it tomorrow.

tanclary:
> @tanclary Please fix conflicts and if there aren't objections, I will merge it tomorrow.

@dssysolyatin , Just rebased and fixed conflicts, should be good now. Thanks!

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3090,6 +3102,13 @@ private static class ItemImplementor extends AbstractRexCallImplementor {
         final RexCall call, final List<Expression> argValueList) {
       final MethodImplementor implementor =
           getImplementor(call.getOperands().get(0).getType().getSqlTypeName());
+      // If the structure is an array, two additional arguments are added to the argValueList
+      // to denote what the offset is and what the behavior should be if the index is out of boudns.
+      if (implementor.method.getName().equals("arrayItemOptional")) {
```
julianhyde, line 3109: Hard-coded method names are too fragile.

Especially strange given that getImplementor returns BuiltInMethod.ARRAY_ITEM and there's no BuiltInMethod.ARRAY_ITEM_OPTIONAL.

Reading this code, I first thought it was because of the 'optional'. But actually it's because it's an array.

So, figure out a way to make this code more robust and intuitive. It's not sufficient that it works.

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
```diff
@@ -885,6 +885,9 @@ ExInst<CalciteException> illegalArgumentForTableFunctionCall(String a0,
   @BaseMessage("Third argument (pad pattern) for LPAD/RPAD must not be empty")
   ExInst<CalciteException> illegalEmptyPadPattern();
 
+  @BaseMessage("Array index {0,number,#} is out of bounds.")
```
julianhyde, line 888: remove the period

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
tanclary: I added a commit that tries to tidy this up by checking the SqlTypeName of the collection argument instead of checking the implementor method. Let me know what you think.

PR #3168, opened 2023-04-19
CALCITE-5662/ Allow CAST(BOOLEAN as INTEGER)

olivrlee:
This PR does two things:
- Fixes a bug that would cause  `CAST(TRUE as INTEGER)` to throw a `NumberFormatException`. 
- Adds in conformance checking to only allow certain dialects to cast booleans to numeric types, as this is not allowed in the SQL standard. 

Note: No additional tests were added to RexProgramTest.java  because the test `testSimplifyCastLiteral` https://github.com/apache/calcite/blob/main/core/src/test/java/org/apache/calcite/rex/RexProgramTest.java#L2300 loops through the `BOOLEAN` type and checks that it now can be casted into `INTEGER`, `BIGINT`, `SMALLINT`, `TINYINT`

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
+++ b/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
@@ -209,6 +209,19 @@ public interface RelDataTypeFactory {
    */
   @Nullable RelDataType leastRestrictive(List<RelDataType> types);
 
+  /**
+   * Returns the most general of a set of types (that is, one type to which
+   * they can all be cast), or null if conversion is not possible. The result
+   * may be a new type that is less restrictive than any of the input types,
+   * e.g. <code>leastRestrictive(INT, NUMERIC(3, 2))</code> could be
+   * {@code NUMERIC(12, 2)}.
+   *
+   * <p>Accepts a {@link SqlTypeMappingRule} that can be used to change casting behavior.
+   *
+   * @param types input types to be combined using union (not null, not empty)
+   * @param mappingRule mapping rule that determines whether types are convertable
+   * @return canonical union type descriptor
+   */
   @Nullable RelDataType leastRestrictive(List<RelDataType> types, SqlTypeMappingRule mappingRule);
 
   /**
--- a/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
+++ b/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
@@ -21,6 +21,7 @@ import org.apache.calcite.sql.SqlCollation;
 import org.apache.calcite.sql.SqlIntervalQualifier;
 import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.validate.SqlConformance;
 import org.apache.calcite.sql.validate.SqlValidatorUtil;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
@@ -53,6 +54,13 @@ public interface RelDataTypeFactory {
    * @return Type system
    */
   RelDataTypeSystem getTypeSystem();
+
+  /**
+   *  Returns the conformance.
+   *
+   * @return Conformance
+   */
+  SqlConformance getConformance();
 
   /**
    * Creates a type that corresponds to a Java class.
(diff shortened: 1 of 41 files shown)


chunweilei:
Could you please correct the format of commit message?

core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
```diff
@@ -955,6 +961,52 @@ public static boolean canCastFrom(
...
+      RelDataType toType,
+      RelDataType fromType) {
+    // These following types support casting from booleans in some dialects such as BigQuery.
+    List<String> numericTypes =
+            ImmutableList.of(TINYINT,
+            SMALLINT,
+            INTEGER,
+            BIGINT)
+        .stream().map(type -> type.toString()).collect(Collectors.toList());
+
```
chunweilei, line 1005: You can use `SqlTypeName.INT_TYPES` instead.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
```diff
@@ -955,6 +961,52 @@ public static boolean canCastFrom(
...
+    return numericTypes.contains(toType.getSqlTypeName().getName()) && fromType.getSqlTypeName().getName()
+        .equals(SqlType.BOOLEAN.name());
+  }
```
chunweilei, line 1008: Use `SqlTypeName` instead of `String` to compare.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
```diff
@@ -955,6 +961,52 @@ public static boolean canCastFrom(
...
+  /**
+   * Returns whether the attempted cast is from a BOOLEAN to a NUMERIC where
+   * NUMERIC is one of [TINYINT, SMALLINT, INTEGER, BIGINT].
+   * @param toType Target of assignment
+   * @param fromType Source of assignment
+   * @return true iff fromType is boolean and toType is one of [TINYINT, SMALLINT, INTEGER, BIGINT]
+   */
+  public static boolean castingBooleanToNumber(
+      RelDataType toType,
+      RelDataType fromType) {
```
chunweilei, line 997: The name is not clear.  `IsCastBooleanToInt` may be better.

olivrlee:
@julianhyde, commit `5d36106677b705a522dbde324c854e08e284300c` is pretty nasty, let me know if you have other suggestions. 
The goal was so that we have access to `conformance` in SqlTypeUtil.canCastFrom() here: https://github.com/apache/calcite/pull/3168/commits/5d36106677b705a522dbde324c854e08e284300c#diff-2966c59464d13efb7f54aba526317f063365ea01d653683e1257978fe79820bdR193

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
```diff
@@ -208,6 +209,8 @@ RelDataType createTypeWithCharsetAndCollation(
    */
   @Nullable RelDataType leastRestrictive(List<RelDataType> types);
 
+  @Nullable RelDataType leastRestrictive(List<RelDataType> types, SqlTypeMappingRule mappingRule);
```
julianhyde, line 212: javadoc!

core/src/test/java/org/apache/calcite/sql/type/SqlTypeUtilTest.java
```diff
@@ -253,7 +257,17 @@ private void compareTypesIgnoringNullability(
     }
   }
 
+  @Test void testCanCastDifferentConformance() {
```
julianhyde, line 260: What is this test testing? Add a comment. Don't force future maintainers to read the source code of the test.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -10501,7 +10508,8 @@ private void checkCustomColumnResolving(String table) {
     final String sql4 = "insert into EMP_MODIFIABLEVIEW2(\"extra\" INTEGER)"
         + " (empno, ename, job, \"extra\")\n"
         + "values (1, 'Arthur', 'clown', true)";
-    s.withSql(sql4).ok();
+    s.withConformance(SqlConformanceEnum.BIG_QUERY).withTypeCoercion(true).withSql(sql4).ok();
```
julianhyde, line 10511: you've checked that it's ok on BigQuery... but if so, why only BigQuery? should it fail on other conformances?

olivrlee: I've verified for that for other conformance enums these should also be included:
- MySql 5

It should not be included for :
- Oracle 10 
- Oracle 12
- Strict 2003

I wasn't able to determine it for Presto

PR #3169, opened 2023-04-20
[CALCITE-5665] Reducing ineffective matches for MaterializedViewRules

asdfgh19:
jira: [CALCITE-5665](https://issues.apache.org/jira/browse/CALCITE-5665)

1. Reduce the number of MaterializedViewRule matches
2. Reduce RelOptMaterialization metadata recalculation

For example, in the test MaterializedViewRelOptRulesTest#testJoinAggregateMaterializationNoAggregateFuncs9, MaterializedViewRule#perform was triggered 489 times in the original implementation, but in this pr, it was only triggered 23 times.

I did a simple test on my personal computer and added a timer to the MaterializedViewRelOptRulesTest#TESTER#optimize method. Here are the milliseconds this method executes before and after optimization.

before: 237 237 229 250 241 227 258 239 253 267 256 244 228 228 241 245 245 238 273 243
after:   199  202 191 190 243 206 207 204 213 239 239 205 192 200 215 196 195 200 193 217
before avg: 243.95 ms
after    avg: 207.3   ms 

before: 225 239 244 236 240 225 281 239 240 236 238 236 228 239 241 238 237 265 256 233
after:    204 201 187 196 192 213 197 186 200 194 189 200 207 199 210 213 188 214 201 206
before avg: 240.8   ms
after    avg: 199.85 ms

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
@@ -123,7 +123,8 @@ public abstract class MaterializedViewRule<C extends MaterializedViewRule.Config
     MaterializationMetadata materializationMetadata =
         materializationMetadataCache.getUnchecked(mq);
     for (RelOptMaterialization materialization : materializations) {
-      if (materializationMetadata.isValidMaterialization(mq, materialization)) {
+      if (materializationMetadata.isValidMaterialization(
+          mq, materialization, this)) {
         return true;
       }
     }
@@ -228,7 +229,8 @@ public abstract class MaterializedViewRule<C extends MaterializedViewRule.Config
           materializationMetadataCache.getUnchecked(mq);
       for (RelOptMaterialization materialization : materializations) {
         // 3.1. View checks before proceeding
-        if (!materializationMetadata.isValidMaterialization(mq, materialization)) {
+        if (!materializationMetadata.isValidMaterialization(
+            mq, materialization, this)) {
           // Skip it
           continue;
         }
@@ -247,6 +249,7 @@ public abstract class MaterializedViewRule<C extends MaterializedViewRule.Config
         // Extract view table references
         final Set<RelTableRef> viewTableRefs =
             materializationMetadata.viewTableRefsMap.get(materialization);
+        assert viewTableRefs != null;
 
         // Filter relevant materializations. Currently, we only check whether
         // the materialization contains any table that is used by the query
@@ -268,6 +271,7 @@ public abstract class MaterializedViewRule<C extends MaterializedViewRule.Config
         // Extract view predicates
         final RelOptPredicateList viewPredicateList =
             materializationMetadata.viewPredicateListMap.get(materialization);
+        assert viewPredicateList != null;
         final RexNode viewPred =
             simplify.simplifyUnknownAsFalse(
                 RexUtil.composeConjunction(rexBuilder,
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -231,21 +260,11 @@ protected void perform(RelOptRuleCall call, @Nullable Project topProject, RelNod
           continue;
         }
 
-        // 3.1. View checks before proceeding
```
JiajunBernoulli, line 234: Code comments are no longer complete:
3.1 was deleted, but 3.2 was reserved.

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -92,6 +95,16 @@
 public abstract class MaterializedViewRule<C extends MaterializedViewRule.Config>
     extends RelRule<C> {
 
+  /**
+   * Weak-key cache of RelMetadataQuery to MaterializationMetadata. The idea is that when
+   * the RelMetadataQuery goes away, so does the cache entry.
+   */
+  private final LoadingCache<RelMetadataQuery, MaterializationMetadata>
+      materializationMetadataCache =
+      CacheBuilder.newBuilder()
```
JiajunBernoulli, line 104: Should we set some parameters?

For example: initialCapacity, maximumSize, refreshAfterWrite, ...

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
asdfgh19: > Code comments are no longer complete: 3.1 was deleted, but 3.2 was reserved.

fixed

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
asdfgh19: > Should we set some parameters?
> 
> For example: initialCapacity, maximumSize, refreshAfterWrite, ...

Through Gc Roots analysis, I found that weakKeys cannot take effect. Even if I delete MaterializationMetadata#RelMetadataQuery, there is still a strong reference pointing to RelMetadataQuery in MaterializationMetadata#RelOptMaterialization#tableRel, so we should really add two parameters, maximumSize and expireAfterAccess, to materializationMetadataCache. Thanks for your suggestion and review!

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -92,6 +96,16 @@
...
 
+  /**
+   * Cache of RelMetadataQuery to MaterializationMetadata.
+   */
+  private final LoadingCache<RelMetadataQuery, MaterializationMetadata>
+      materializationMetadataCache =
+      CacheBuilder.newBuilder()
+          .expireAfterAccess(5, TimeUnit.MINUTES)
+          .maximumSize(1_024)
+          .build(CacheLoader.from(MaterializationMetadata::new));
```
chunweilei, line 107: Is `1_024` a typo?

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
```diff
@@ -1426,6 +1448,61 @@ protected enum MatchModality {
...
+    private final Map<RelOptMaterialization, Set<RelTableRef>> viewTableRefsMap = new HashMap<>();
+    private final Map<RelOptMaterialization, RelOptPredicateList>
+        viewPredicateListMap = new HashMap<>();
+
+    protected boolean isValidMaterialization(
+        RelMetadataQuery mq, RelOptMaterialization materialization) {
+      if (notValidMaterializations.contains(materialization)) {
+        return false;
+      } else if (viewTableRefsMap.containsKey(materialization)) {
+        return true;
```
chunweilei, line 1466: Should we use `viewTableRefsMap.containsKey(materialization) && viewPredicateListMap.containsKey(materialization)` to be more accurate?

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
asdfgh19: > Is `1_024` a typo?

Not a typo, this is a reference to QuerySqlStatisticProvider#SILENT_CACHING_INSTANCE, in order to be consistent with the existing code style.

core/src/main/java/org/apache/calcite/rel/rules/materialize/MaterializedViewRule.java
asdfgh19: > Should we use `viewTableRefsMap.containsKey(materialization) && viewPredicateListMap.containsKey(materialization)` to be more accurate?

Indeed we should.Thanks for your suggestion and review!

PR #3170, opened 2023-04-21
[CALCITE-5669] Add rule to transform correlate to join when correlation variable is unused

zabetak:
…


core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -3646,6 +3649,104 @@ private void checkRightEmptyAntiJoin(JoinRelType type) {
...
+
+  @Test void testAntiCorrelateWithCorrelationUnchanged() {
+    relFn(correlationPlanBuilder(JoinRelType.ANTI, true))
+        .withRule(CoreRules.CORRELATE_TO_JOIN_RULE)
+        .checkUnchanged();
+  }
+
+  /**
+   * Provides a function that will create a simple plan correlating EMP and DEPT tables with the
+   * provided join type.
```
julianhyde, line 3717: please add `<p>` at start of paragraphs

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -4674,6 +4675,9 @@ private static class CorrelationCollector extends RelHomogeneousShuttle {
     private final VariableUsedVisitor vuv = new VariableUsedVisitor(this);
 
     @Override public RelNode visit(RelNode other) {
+      if (other instanceof HepRelVertex) {
```
julianhyde, line 4678: You might like to cherry-pick in my 5fb70e196ddafaceaf1349378de88e48f9f232f3 commit, which adds the `RelNode.stripped()`. With that commit, this `if` might not be required.

julianhyde (commented):
+1

In the subject, can you capitalize Correlate and Join

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
zabetak: Method doesn't exist in latest patch.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
zabetak: The `RelOptUtil` class remains unchanged in the latest patch so no need to cherry-pick anything.

PR #3171, opened 2023-04-23
[CALCITE-5664] Support CONVERT/TRANSLATE function with USING keyword

ILuffZhe:
In this patch ,we support CONVERT/TRANSLATE(value USING transcodingName).


ILuffZhe:
Sorry about rebasing this wrong, I'll open a new PR for this.

PR #3172, opened 2023-04-23
[CALCITE-5670] Assertion error in SemiJoinJoinTransposeRule when Semi-Join has keys from both tables of the bottom Join

rkondakov:


core/src/main/java/org/apache/calcite/rel/rules/SemiJoinJoinTransposeRule.java
```diff
@@ -102,16 +101,18 @@ public SemiJoinJoinTransposeRule(RelBuilderFactory relBuilderFactory) {
 
     // determine which operands below the semi-join are the actual
     // Rels that participate in the semi-join
+    final ImmutableIntList leftKeys = semiJoin.analyzeCondition().leftKeys;
```
zoudan, line 104: It is unnecessary to move this line here?

rkondakov: The `leftKeys` variable was defined much earlier than the place where it was actually used. I moved it closer to this place to improve code readability.

zoudan: OK, it make sense

core/src/main/java/org/apache/calcite/rel/rules/SemiJoinJoinTransposeRule.java
```diff
@@ -102,16 +101,18 @@ public SemiJoinJoinTransposeRule(RelBuilderFactory relBuilderFactory) {
...
       if (leftKey < nFieldsX) {
         nKeysFromX++;
       }
     }
 
-    // the keys must all originate from either the left or right;
-    // otherwise, a semi-join wouldn't have been created
-    assert (nKeysFromX == 0) || (nKeysFromX == leftKeys.size());
+    if ((nKeysFromX != 0) && (nKeysFromX != leftKeys.size())) {
+      // We can not push semi-join down if join keys come from different tables under the join.
```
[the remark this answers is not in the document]
rkondakov: Fixed. Thank you for your help.

PR #3177, opened 2023-04-26
[CALCITE-5612] Babel parser should support PostgreSQL's SET TRANSACTION command

dssysolyatin:


LakeShen:
Hi @dssysolyatin ,there some conflicts with main branch code,please resolve them.

dssysolyatin:
Apologies for the long inactivity. I've updated the PR.

caicancai:
> Apologies for the long inactivity. I've updated the PR.

I'm not very familiar with this part of the code, I will try to review it

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -358,41 +358,82 @@ private void checkParseInfixCast(String sqlType) {
...
+  }
+
+  @Test void testPostgresSqlReset() {
+    SqlParserFixture f = fixture().withDialect(PostgresqlSqlDialect.DEFAULT);
+
+    f.sql("RESET ALL").same();
+    f.sql("RESET ROLE")
+        .ok("RESET \"role\"");
+    f.sql("RESET SESSION AUTHORIZATION")
+        .ok("RESET \"session_authorization\"");
```
mihaibudiu, line 427: are these also right?

core/src/main/java/org/apache/calcite/sql/SqlBasicFunction.java
```diff
@@ -42,7 +42,7 @@
  * behavior only by providing strategy objects, not by overriding methods in a
  * subclass.
  */
-public class SqlBasicFunction extends SqlFunction {
+public final class SqlBasicFunction extends SqlFunction {
```
mihaibudiu, line 45: this looks like a major change, which may break users of this class

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -358,41 +358,82 @@ private void checkParseInfixCast(String sqlType) {
...
     f.sql("SET SESSION autovacuum = DEFAULT")
-        .ok("ALTER SESSION SET \"autovacuum\" = DEFAULT");
+        .ok("SET \"autovacuum\" = DEFAULT");
     f.sql("SET LOCAL autovacuum TO 'DEFAULT'")
-        .ok("ALTER LOCAL SET \"autovacuum\" = 'DEFAULT'");
+        .ok("SET LOCAL \"autovacuum\" = 'DEFAULT'");
 
     f.sql("SET SESSION TIME ZONE DEFAULT")
-        .ok("ALTER SESSION SET \"timezone\" = DEFAULT");
+        .ok("SET TIME ZONE DEFAULT");
```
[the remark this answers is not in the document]
dssysolyatin: Previous output was wrong. If execute previous query against PostgreSQL
```
ALTER SESSION SET "timezone" = DEFAULT;
```
PostgreSQL throws an exception:
```
psql:commands.sql:1: ERROR:  syntax error at or near "SESSION"
LINE 1: ALTER SESSION SET "timezone" = DEFAULT;
```

The valid sql is SET TIME ZONE DEFAULT

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
dssysolyatin: This is correct.

RESET SESSION AUTHORIZATION is equivalent to RESET \"session_authorization\".

SESSION AUTHORIZATION is simply syntactic sugar.

See https://github.com/postgres/postgres/blob/924e03917d6f7c7536b2a1228b5422f175f58f9e/src/backend/parser/gram.y#L1767

I'll add a comment to the test to clarify.

core/src/main/java/org/apache/calcite/sql/SqlBasicFunction.java
dssysolyatin: I'll remove it. I don’t remember why I added it—thanks for pointing it out.

dssysolyatin: Ok, I remembered why I added it. If you look at the comment in the class:
```
The class is final, and instances are immutable

Since the class is final, you can modify behavior only by providing strategy objects, not by overriding methods in a subclass.
```
This comment was added when the class was introduced, and from the Git history, I can see that making it final was simply overlooked.

Think it is right to make it final.

dssysolyatin:
@mihaibudiu , please recheck the PR. I’ve addressed your comments.

Only `continuous-integration/jenkins/pr-head` is failing, which I believe is unrelated to my changes. It seems to be something about dependencies for :arrow:test.

```
   Execution failed for task ':arrow:test':
        org.gradle.api.internal.artifacts.ivyservice.DefaultLenientConfiguration$ArtifactResolveException: Could not resolve all dependencies for configuration ':arrow:jacocoAgent'.
```

core/src/main/java/org/apache/calcite/sql/SqlBasicFunction.java
mihaibudiu: I agree that the intent was to have the class final, but if someone happens to have inherited from this class in some project that uses Calcite, they will now be stuck.

dssysolyatin: @mihaibudiu This class can’t be inherited from outside because it has a private constructor, which prevents other classes from calling super(). Making it non-final only allows subclasses within SqlBasicFunction itself.

Perhaps this was done to allow adding an inner Builder class in the future, but I’m not sure.

Let’s ask the author of the commit, @julianhyde, about it

mihaibudiu:
I guess if all constructors are private then the change can be made, didn't notice that.

dssysolyatin:
@mihaibudiu Let's keep this change.

1. As mentioned before, the constructor is private, which means that subclasses cannot be created outside of SqlBasicFunction.
2. Due to point 1, adding `final` will not break binary compatibility either ([Java Language Specification, Section 13.4.2](https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.2)).
3. If someone decides to introduce a Builder inside SqlBasicFunction in the future, removing `final` also will not break binary compatibility ([Java Language Specification, Section 13.4.2](https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.2)).

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql/SqlBasicFunction.java
suibianwanwank: Making constructor methods private seems to be a recent change. SqlBasicFunction was extended in the project I'm using, and this seems to be a thing to keep in mind when upgrading calcite.

PR #3178, opened 2023-04-26
[CALCITE-5677] SUBSTR signature incorrect for BigQuery

tanclary:
Prevent BigQuery SUBSTR(value, position[, length]) from being translated to SUBSTRING(value FROM position FOR length). Keep arguments comma-separated by adding onto `BigQuerySqlDialect#unparseCall`.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/ClickHouseSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/ClickHouseSqlDialect.java
@@ -30,11 +30,9 @@ import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlTimeLiteral;
 import org.apache.calcite.sql.SqlTimestampLiteral;
 import org.apache.calcite.sql.SqlWriter;
-import org.apache.calcite.sql.fun.SqlStdOperatorTable;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -185,6 +186,21 @@ public BigQuerySqlDialect(SqlDialect.Context context) {
     case TRIM:
       unparseTrim(writer, call, leftPrec, rightPrec);
       break;
+    case OTHER_FUNCTION:
+      if (call.getOperator() instanceof SqlSubstringFunction) {
+        final SqlWriter.Frame funCallFrame = writer.startFunCall(call.getOperator().getName());
+        call.operand(0).unparse(writer, leftPrec, rightPrec);
+        writer.sep(",", true);
+        call.operand(1).unparse(writer, leftPrec, rightPrec);
+        if (3 == call.operandCount()) {
```
olivrlee, line 195: nit: switch the ordering of these

dssysolyatin:
LGMT. If there is no objections, I will merge it tomorrow

tanclary:
Thanks, @dssysolyatin !

PR #3179, opened 2023-04-26
[CALCITE-5676] Set JDBC `DATA_TYPE`  and `TYPE_NAME` metadata values for `MEASURE` types

tjbanghart:
[CALCITE-5676](https://issues.apache.org/jira/browse/CALCITE-5676)

### DIFF
--- a/core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
+++ b/core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
@@ -838,7 +838,12 @@ public class CalcitePrepareImpl implements CalcitePrepare {
   }
 
   private static int getTypeOrdinal(RelDataType type) {
-    return type.getSqlTypeName().getJdbcOrdinal();
+    switch (type.getSqlTypeName()) {
+    case MEASURE:
+      return type.getMeasureElementType().getSqlTypeName().getJdbcOrdinal();
+    default:
+      return type.getSqlTypeName().getJdbcOrdinal();
+    }
   }
 
   private static String getClassName(@SuppressWarnings("unused") RelDataType type) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
```diff
@@ -837,8 +837,16 @@ private static ColumnMetaData.AvaticaType avaticaType(JavaTypeFactory typeFactor
         : origins.get(origins.size() - 1 - offsetFromEnd);
   }
 
-  private static int getTypeOrdinal(RelDataType type) {
-    return type.getSqlTypeName().getJdbcOrdinal();
+  public static int getTypeOrdinal(RelDataType type) {
```
julianhyde, line 840: formatting looks off. Also, I'd save a few lines by avoiding assigning to the variable, just `return` immediately

core/src/test/java/org/apache/calcite/test/JdbcFrontLinqBackTest.java
```diff
@@ -343,12 +356,13 @@ private CalciteAssert.AssertThat mutable(
         .with(makePostProcessor(employees));
   }
 
-  static AbstractModifiableTable mutable(String tableName,
```
julianhyde, line 346: It doesn't seem right to pass a type factory into a table constructor. I know these are only mock tables for testing. But tables have a lifetime longer than a single statement, therefore longer than a type factory. That's why `getRowType` takes a type factory as an argument.

tjbanghart: Thanks for the review @julianhyde! I apologize for the delay. I got rid of the strange `postProcessor` business in favor of a `withMeasures` flag. May not be ideal but is easier to follow than the passing a type factory for the table to use.

PR #3181, opened 2023-04-28
[CALCITE-5679] HepPlanner#buildFinalPlan: do not clear metadata cache if RelNode has not changed

rubenada:
[CALCITE-5679](https://issues.apache.org/jira/browse/CALCITE-5679) 
Minor optimization:
HepPlanner#buildFinalplan clears the metadata cache info of the impacted RelNode after replacing its inputs, but in the cases where the RelNode does not change (e.g. when it has no inputs) this clear cache should not be necessary (unless I'm missing something).


JiajunBernoulli:
Because [CALCITE-3791](https://issues.apache.org/jira/browse/CALCITE-3791).
When a change occurs, the cache of the parent node will be cleared by another block of code. 

The commit only affects current node, so I approve the PR.

JiajunBernoulli (approved):

libenchao:
+1, there is no meaning to recalculate metadata for unchanged nodes.

PR #3182, opened 2023-04-29
[CALCITE-5680] Wrong plan for multiple IN sub-queries with only literal operands

herunkang2018:


JiajunBernoulli:
I would like to know if `sub-query.iq` can be reproduced.

If it can, we should add unit test.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6448,6 +6448,42 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
...
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5680">[CALCITE-5680]
+   * Wrong field reference lookup due to same intermediate table alias
+   * of multiple sub-queries with only literal operands in subquery remove phase</a>. */
+  @Test void testExpandFilterConstantInCorrelatedWithTwoSubQueries() {
+    final String sql = "select empno from sales.empnullables as e\n"
+        + "where 1 in (\n"
+        + "  select deptno from sales.deptnullables where e.ename = name and deptno > 10)\n"
+        + "or 2 in (\n"
+        + "  select deptno from sales.deptnullables where e.ename = name and deptno < 20)";
+    System.out.println(sql);
```
JiajunBernoulli, line 6461: `System.out.println(sql)` is redundant.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6448,6 +6448,42 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
...
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5680">[CALCITE-5680]
+   * Wrong field reference lookup due to same intermediate table alias
+   * of multiple sub-queries with only literal operands in subquery remove phase</a>. */
+  @Test void testExpandFilterConstantInWithTwoSubQueries() {
+    final String sql = "select empno from sales.empnullables\n"
+        + "where 1 in (\n"
+        + "  select deptno from sales.deptnullables where name = 'dept1')\n"
+        + "or 2 in (\n"
+        + "  select deptno from sales.deptnullables where name = 'dept2')";
+    System.out.println(sql);
```
JiajunBernoulli, line 6479: System.out.println(sql) is redundant.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6448,6 +6448,42 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
...
+    sql(sql)
+        .withSubQueryRules()
+        .withRelBuilderSimplify(false)
```
JiajunBernoulli, line 6482: Is the plan correct  if we enable RelBuilderSimplify?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6448,6 +6448,42 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
...
+    final String sql = "select empno from sales.empnullables as e\n"
+        + "where 1 in (\n"
+        + "  select deptno from sales.deptnullables where e.ename = name and deptno > 10)\n"
+        + "or 2 in (\n"
+        + "  select deptno from sales.deptnullables where e.ename = name and deptno < 20)";
+    System.out.println(sql);
+    sql(sql)
+        .withSubQueryRules()
+        .withRelBuilderSimplify(false)
+        .withTrim(true)
```
JiajunBernoulli, line 6465: Why we need to enable trim?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
herunkang2018: This is for debug only, I have removed it yet.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
herunkang2018: The plan will be more clear to read if we trim unused fields.

herunkang2018:
@JiajunBernoulli Thanks for reviewing, I will check `sub-query.iq`.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
herunkang2018: After this fix, yes. We here don't enable RelBuilderSimplify, to make plan more original, and easier to identify that the plan is correct.

JiajunBernoulli: Ok, I understand.

I suggest you write a note to explain it, otherwise others may mistakenly think that this is a reproduction condition.

herunkang2018: Thanks for your suggestion, I will add it soon.

herunkang2018: Added.

herunkang2018:
@JiajunBernoulli the `sub-query.iq` doesn't cover this case, and I reproduce this bug in [reproduce_issue_5680](https://github.com/apache/calcite/compare/main...herunkang2018:calcite:reproduce_issue_5680#diff-26cbb70deb4731f2c84faf803724dc40a9cdf6e2f2c67f8724e1afbc98fc8950R3547), note that the first sub-query return empty result, and the second sub-query return one row, the final result is empty, which should be 5 rows due to OR predicate and the second sub-query matches.

core/src/test/resources/sql/sub-query.iq
```diff
@@ -3544,4 +3544,33 @@ FROM UNNEST(ARRAY['a', 'b']) AS a(attidentity);
 
 !ok
 
+# Test case for [CALCITE-5680] Wrong plan for multiple IN sub-queries with only literal operands
```
herunkang2018, line 3547: @JiajunBernoulli We only add non-correlated version test case here, not including a correlated one.
This is because the correlated one will throw exception now, see [CALCITE-5390](https://issues.apache.org/jira/browse/CALCITE-5390), after we fix CALCITE-5390, the correlated one should be added here.

JiajunBernoulli: Ok, I get it.

zoudan:
LGTM

libenchao (approved):
@herunkang2018 Thanks for your PR! The changes looks good to me.

@JiajunBernoulli Regarding the Quidem tests, since the plan could clearly describe the problem and the fix, I think they are nice-to-have, not a must-have. And CALCITE-5390 seems not expected to be fixed very soon, so it shouldn't be a blocker for merging this one.

JiajunBernoulli (approved):

JiajunBernoulli:
LGTM

PR #3184, opened 2023-05-01
[CALCITE-3959] Implement INSTR function (Enabled for BigQuery, MySql, Oracle)

jhugomoore:
[BiqQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#instr) and [Oracle](https://docs.oracle.com/cd/B13789_01/server.101/b10759/functions058.htm) both support functionally identical INSTR(source_value, search_value[, position[, occurrence]]) functions which accepts 2 (character strings or binary strings), 1 optional int representing position, and 1 optional int representing occurrence.

Occurrence and position are assigned a default value of 1 if not specified.

The function returns the 1-based position of the nth occurrence of the 2nd operand in the 1st operand where n is defined by the 4th operand. The function begins searching at the 1-based position specified in the 3rd operand.

The function also supports negative position values, with -1 indicating the last character, and will search backwards from the position specified in that case. 

Returns 0 if:

No match is found.
If occurrence is greater than the number of matches found.
If position is greater than the length of source_value.
Returns NULL if:

Any input argument is NULL.
Returns an error if:

position is 0.
occurrence is 0 or negative.
EXAMPLE: INSTR("abc", "bc") would return 2.

EXAMPLE: INSTR("abcabc", "bc", 3) would return 5.

EXAMPLE: INSTR("abcabc", "bc", -1, 1) would return 5.

EXAMPLE: INSTR("abcabc", "bc", -1, 2) would return 2.

 

MySQL also has an  [INSTR](https://www.w3schools.com/sql/func_mysql_instr.asp) function, the functionality of which is a subset of the INSTR present in BQ and Oracle. MySQL INSTR only takes 2 parameters and returns the first occurrence of the search value in the source value.

[Jira Case](https://issues.apache.org/jira/browse/CALCITE-3959)

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
+++ b/core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
@@ -1020,10 +1020,10 @@ public interface CalciteResource {
   @BaseMessage("No operator for ''{0}'' with kind: ''{1}'', syntax: ''{2}'' during JSON deserialization")
   ExInst<CalciteException> noOperator(String name, String kind, String syntax);
 
-  @BaseMessage("Invalid input for position function from operand value must not be zero")
+  @BaseMessage("Invalid input for POSITION function: from operand value must not be zero")
   ExInst<CalciteException> fromNotZero();
 
-  @BaseMessage("Invalid input for position function occurrence operand value must not be zero")
+  @BaseMessage("Invalid input for POSITION function: occurrence operand value must be positive")
   ExInst<CalciteException> occurrenceNotZero();
 
   @BaseMessage("Only tables with set semantics may be partitioned. Invalid PARTITION BY clause in the {0,number,#}-th operand of table function ''{1}''")
@@ -1034,6 +1034,4 @@ public interface CalciteResource {
 
   @BaseMessage("A table function at most has one input table with row semantics. Table function ''{0}'' has multiple input tables with row semantics")
   ExInst<SqlValidatorException> multipleRowSemanticsTables(String funcName);
-
-
 }
--- a/core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
+++ b/core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
@@ -1020,6 +1020,12 @@ public interface CalciteResource {
   @BaseMessage("No operator for ''{0}'' with kind: ''{1}'', syntax: ''{2}'' during JSON deserialization")
   ExInst<CalciteException> noOperator(String name, String kind, String syntax);
 
+  @BaseMessage("Invalid input for position function from operand value must not be zero")
+  ExInst<CalciteException> fromNotZero();
+
+  @BaseMessage("Invalid input for position function occurrence operand value must not be zero")
+  ExInst<CalciteException> occurrenceNotZero();
+
   @BaseMessage("Only tables with set semantics may be partitioned. Invalid PARTITION BY clause in the {0,number,#}-th operand of table function ''{1}''")
   ExInst<SqlValidatorException> invalidPartitionKeys(int idx, String funcName);
 
@@ -1028,4 +1034,6 @@ public interface CalciteResource {
 
   @BaseMessage("A table function at most has one input table with row semantics. Table function ''{0}'' has multiple input tables with row semantics")
   ExInst<SqlValidatorException> multipleRowSemanticsTables(String funcName);
+
+
 }
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -3104,88 +3104,94 @@ public class SqlFunctions {
 
   /** SQL {@code POSITION(seek IN string FROM integer)} function. */
   public static int position(String seek, String s, int from) {
+    if (from == 0) throw RESOURCE.fromNotZero().ex();
+    // Case when from is positive
     if (from > 0) {
       final int from0 = from - 1; // 0-based
-      if (from0 > s.length() || from0 < 0) {
+      if (from0 >= s.length() || from0 < 0) {
         return 0;
       }
 
       return s.indexOf(seek, from0) + 1;
     }
+    // Case when from is negative
     final int rightIndex = from + s.length(); // negative position to positive index
     if (rightIndex <= 0) {
       return 0;
     }
-    return s.substring(0, rightIndex).lastIndexOf(seek) + 1;
+    return s.substring(0, rightIndex+1).lastIndexOf(seek) + 1;
   }
 
   /** SQL {@code POSITION(seek IN string FROM integer)} function for byte
    * strings. */
   public static int position(ByteString seek, ByteString s, int from) {
+    if (from == 0) throw RESOURCE.fromNotZero().ex();
+    // Case when from is positive
     if (from > 0) {
       final int from0 = from - 1; // 0-based
-      if (from0 > s.length() || from0 < 0) {
+      if (from0 >= s.length() || from0 < 0) {
         return 0;
       }
-
       return s.indexOf(seek, from0) + 1;
     }
+    // Case when from is negative
     final int rightIndex = from + s.length();
     if (rightIndex <= 0) {
       return 0;
     }
-    return -1;
-    //s.lastIndexOf(seek, rightIndex) + 1;
+    int lastIndex = 0;
+    while (lastIndex < rightIndex) {
+      int indexOf = s.substring(lastIndex, rightIndex + 1).indexOf(seek) + 1;
+      if (indexOf == 0) {
+        break;
+      }
+      lastIndex += indexOf;
+    }
+    return lastIndex;
   }
 
   /** SQL {@code POSITION(seek, string, from, occurrence)} function. */
   public static int position(String seek, String s, int from, int occurrence) {
-    if (from > 0){
-      int rollingFrom = from;
-      for (int i = 0; i< occurrence; i++) {
-        rollingFrom = position(seek, s, rollingFrom);
-        if (rollingFrom == 0) {
+    if (occurrence == 0) throw RESOURCE.occurrenceNotZero().ex();
+    for (int i = 0; i < occurrence; i++){
+      if (from > 0){
+        from = position(seek, s, from + (i == 0 ? 0 : 1));
+        if (from == 0) {
           return 0;
         }
+      } else {
+        from = position(seek, s, from);
+        if (from == 0) {
+          return 0;
+        }
+        from -= (s.length() + 2);
       }
-      return rollingFrom;
     }
-    int rollingFromNeg = from;
-    int rollingFromPos = 0;
-    for (int i = 0; i< occurrence; i++) {
-      rollingFromPos = position(seek, s, rollingFromNeg);
-      if (rollingFromPos == 0) {
-        return 0;
-      }
-      rollingFromNeg = rollingFromPos - s.length();
-    }
-    return rollingFromPos;
-
+    if (from < 0) from += s.length() + 2;
+    return from;
   }
 
   /** SQL {@code POSITION(seek, string, from, occurrence)} function for byte
    * strings. */
   public static int position(ByteString seek, ByteString s, int from, int occurrence) {
-    if (from > 0){
-      int rollingFrom = from;
-      for (int i = 0; i< occurrence; i++) {
-        rollingFrom = position(seek, s, rollingFrom);
-        if (rollingFrom == 0) {
+    if (occurrence == 0) throw RESOURCE.occurrenceNotZero().ex();
+    for (int i = 0; i < occurrence; i++){
+      if (from > 0){
+        from = position(seek, s, from + (i == 0 ? 0 : 1));
+        if (from == 0) {
           return 0;
         }
       }
-      return rollingFrom;
-    }
-    int rollingFromNeg = from;
-    int rollingFromPos = 0;
-    for (int i = 0; i< occurrence; i++) {
-      rollingFromPos = position(seek, s, rollingFromNeg);
-      if (rollingFromPos == 0) {
-        return 0;
+      else {
+        from = position(seek, s, from);
+        if (from == 0) {
+          return 0;
+        }
+        from -= (s.length() + 2);
       }
-      rollingFromNeg = rollingFromPos - s.length();
     }
-    return rollingFromPos;
+    if (from < 0) from += s.length() + 2;
+    return from;
 
   }
 
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -3104,23 +3104,89 @@ public class SqlFunctions {
 
   /** SQL {@code POSITION(seek IN string FROM integer)} function. */
   public static int position(String seek, String s, int from) {
-    final int from0 = from - 1; // 0-based
-    if (from0 > s.length() || from0 < 0) {
+    if (from > 0) {
+      final int from0 = from - 1; // 0-based
+      if (from0 > s.length() || from0 < 0) {
+        return 0;
+      }
+
+      return s.indexOf(seek, from0) + 1;
+    }
+    final int rightIndex = from + s.length(); // negative position to positive index
+    if (rightIndex <= 0) {
       return 0;
     }
-
-    return s.indexOf(seek, from0) + 1;
+    return s.substring(0, rightIndex).lastIndexOf(seek) + 1;
   }
 
   /** SQL {@code POSITION(seek IN string FROM integer)} function for byte
    * strings. */
   public static int position(ByteString seek, ByteString s, int from) {
-    final int from0 = from - 1;
-    if (from0 > s.length() || from0 < 0) {
+    if (from > 0) {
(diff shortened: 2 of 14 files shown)


core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
```diff
@@ -1028,4 +1034,6 @@ ExInstWithCause<CalciteException> failedToAccessField(
 
   @BaseMessage("A table function at most has one input table with row semantics. Table function ''{0}'' has multiple input tables with row semantics")
   ExInst<SqlValidatorException> multipleRowSemanticsTables(String funcName);
+
```
olivrlee, line 1037: These can be removed

core/src/main/java/org/apache/calcite/sql/fun/SqlPositionFunction.java
```diff
@@ -53,12 +57,23 @@ public SqlPositionFunction(String name) {
       int leftPrec,
       int rightPrec) {
     final SqlWriter.Frame frame = writer.startFunCall(getName());
-    call.operand(0).unparse(writer, leftPrec, rightPrec);
-    writer.sep("IN");
-    call.operand(1).unparse(writer, leftPrec, rightPrec);
-    if (3 == call.operandCount()) {
-      writer.sep("FROM");
+    if (2 == call.operandCount() || 3 == call.operandCount()) {
```
olivrlee, line 60: nit: switch the ordering like `call.operandCount() == 2`

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3104,23 +3104,105 @@ public static int position(ByteString seek, ByteString s) {
...
+  }
 
-    return s.indexOf(seek, from0) + 1;
+  /** SQL {@code POSITION(seek, string, from, occurrence)} function for byte
+   * strings. */
+  public static int position(ByteString seek, ByteString s, int from, int occurrence) {
+    if (occurrence == 0) {
+      throw RESOURCE.occurrenceNotZero().ex();
+    }
+    for (int i = 0; i < occurrence; i++) {
```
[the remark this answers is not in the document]
tjbanghart: ^ I didn't think about overlapping patterns. I am not sure if those suggestions will work.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -402,6 +407,32 @@ private static RexNode convertNvl(SqlRexContext cx, SqlCall call) {
...
 
+  /** Converts a call to the INSTR function. */
+  private static RexNode convertInstr(SqlRexContext cx, SqlCall call) {
+    final RexBuilder rexBuilder = cx.getRexBuilder();
+    final List<RexNode> operands =
+        convertOperands(cx, call, SqlOperandTypeChecker.Consistency.NONE);
+    final RelDataType type =
+        cx.getValidator().getValidatedNodeType(call);
+    final List<RexNode> exprs = new ArrayList<>();
+    if (call.operandCount() == 2) {
```
olivrlee, line 418: consider a 
```
switch (call.operandCount())
case 2 
case 3
...
```

olivrlee:
Left some minor comments, great if someone else takes a look too.
Awesome job and first PR!

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -3686,7 +3686,7 @@ static void checkRlikeFails(SqlOperatorFixture f) {
     f.checkScalarExact("position(x'bb' in x'aabbccaabbcc' FROM 3)", 5);
     f.checkScalarExact("position(x'bb' in x'aabbccaabbcc' FROM 5)", 5);
     f.checkScalarExact("position(x'bb' in x'aabbccaabbcc' FROM 6)", 0);
-    f.checkScalarExact("position(x'bb' in x'aabbccaabbcc' FROM -5)", 0);
```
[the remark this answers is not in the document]
jhugomoore: same reasoning^ also had me confused for a while

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
jhugomoore: Definitely had a sneaking feeling there's a much more elegant solution than what I implemented, will keep brainstorming.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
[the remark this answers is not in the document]
jhugomoore: added some but might've gone overboard, let me know if it's too verbose

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3104,23 +3104,106 @@ public static int position(ByteString seek, ByteString s) {
...
+        return s.indexOf(seek, from0) + 1;
+      }
+    }
+    // Case when from is negative
+    final int rightIndex = from + s.length();
+    if (rightIndex <= 0) {
       return 0;
     }
+    int lastIndex = 0;
+    while (lastIndex < rightIndex) {
```
julianhyde, line 3148: Is there a method we could add to `ByteString` to allow you to write the same code in this method as for `position(String, String, int)`? (`ByteString` is in Avatica, so we would have to wait a release.)

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3104,23 +3104,106 @@ public static int position(ByteString seek, ByteString s) {
...
+    return lastIndex;
+  }
 
-    return s.indexOf(seek, from0) + 1;
+  /** SQL {@code POSITION(seek, string, from, occurrence)} function. */
+  public static int position(String seek, String s, int from, int occurrence) {
+    if (occurrence == 0) {
+      throw RESOURCE.occurrenceNotZero().ex();
+    }
+    for (int i = 0; i < occurrence; i++) {
```
julianhyde, line 3163: `from` is constant and `i` is not. So would it make sense to move the `if` outside the `for`, so that it's only checked once?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
julianhyde: The call to `position` isn't recursive, thank goodness. 

I'd be inclined to inline the `position` function, and move the `if` outside the `for`. I think you can skip a lot of checks.

I was wondering how to implement this without calling `String.substring`, but I can't find an easy solution. So let's leave it as is.

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
```diff
@@ -1054,6 +1055,54 @@ private void thereAndBack(byte[] bytes) {
       // ok
     }
   }
+  @Test void testPosition() {
```
julianhyde, line 1058: add blankline before

site/_docs/reference.md
```diff
@@ -2693,6 +2693,8 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b o | GREATEST(expr [, expr ]*)                    | Returns the greatest of the expressions
 | b h s | IF(condition, value1, value2)              | Returns *value1* if *condition* is TRUE, *value2* otherwise
 | b | IFNULL(value1, value2)                         | Equivalent to `NVL(value1, value2)`
+| b o | INSTR(string, substring[, from, occurrence]) | Returns the position of substring in string, searching starting at from, and until locating the nth occurrence of substring.
```
julianhyde, line 2696: s/substring/\*substring\*/ etc.

s/from/from (default 1)/

s/occurrence/occurrence (default 1)/

remove '.' at end of line

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
```diff
@@ -1054,6 +1055,54 @@ private void thereAndBack(byte[] bytes) {
...
+    assertThat(3, is(position("c", "abcdec")));
+    assertThat(3, is(position("c", "abcdec", 2)));
```
julianhyde, line 1060: These tests are all backwards! For example, rather than
```
assertThat(3, is(position("c", "abcdec", 2)));
```
you should write
```
assertThat(position("c", "abcdec", 2), is(3)));
```
Here, `position("c", "abcdec", 2)` is the thing you are testing, and `3` is the expected result. In fact `is(3)` is the assertion about the expected result.

Why does it matter? If the test fails - say the position call returns 57 - then the framework should print 'got 57, expected 3'. Also, the test reads from left-to-right like an English sentence.

Maybe you're familiar with `assertEquals`, and earlier method where you were SUPPOSED to write the arguments backwards. `assertEquals(3, position("c", "abcdec", 2));` would be the correct way to write this test using `assertEquals`.

core/src/main/resources/org/apache/calcite/runtime/CalciteResource.properties
```diff
@@ -336,4 +336,6 @@ InvalidPartitionKeys=Only tables with set semantics may be partitioned. Invalid
 InvalidOrderBy=Only tables with set semantics may be ordered. Invalid ORDER BY clause in the {0,number,#}-th operand of table function ''{1}''
 MultipleRowSemanticsTables=A table function at most has one input table with row semantics. Table function ''{0}'' has multiple input tables with row semantics
 NoOperator=No operator for ''{0}'' with kind: ''{1}'', syntax: ''{2}'' during JSON deserialization
+FromNotZero=Invalid input for position function: from operand value must not be zero
```
julianhyde, line 339: s/position/POSITION/

Other error messages have the function in upper case. It reinforces that it's a function name.

julianhyde (changes requested):
Nearly there, just a few tweaks left. Nice work. Please ping me when you've revised.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
jhugomoore: Yes! Java strings have a nice [.lastIndexOf()](https://www.w3schools.com/java/ref_string_lastindexof.asp) function that isn't present in Avatica Bytestrings

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
jhugomoore: Not sure I understand what you mean. From gets updated every time the for loop runs, so we have to check whether to use the logic for from being positive or negative each time. I think checking if from is 0 after assignment is necessary because it will error if we continue running

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
```diff
@@ -1020,6 +1020,12 @@ ExInstWithCause<CalciteException> failedToAccessField(
   @BaseMessage("No operator for ''{0}'' with kind: ''{1}'', syntax: ''{2}'' during JSON deserialization")
   ExInst<CalciteException> noOperator(String name, String kind, String syntax);
 
+  @BaseMessage("Invalid input for position function: from operand value must not be zero")
+  ExInst<CalciteException> fromNotZero();
+
+  @BaseMessage("Invalid input for position function: occurrence operand value must not be zero")
```
wnob, line 1026: `occurence` must be positive. Update the message and condition.

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
```diff
@@ -1020,6 +1020,12 @@ ExInstWithCause<CalciteException> failedToAccessField(
   @BaseMessage("No operator for ''{0}'' with kind: ''{1}'', syntax: ''{2}'' during JSON deserialization")
   ExInst<CalciteException> noOperator(String name, String kind, String syntax);
 
+  @BaseMessage("Invalid input for position function: from operand value must not be zero")
```
wnob, line 1023: BQ and Oracle both call this `position`, not `from`. Is there another dialect that calls it `from`?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3104,23 +3104,106 @@ public static int position(ByteString seek, ByteString s) {
...
+    if (from > 0) {
+      final int from0 = from - 1; // 0-based
+      if (from0 >= s.length() || from0 < 0) {
+        return 0;
+      } else {
+        return s.indexOf(seek, from0) + 1;
+      }
+    }
+    // Case when from is negative
+    final int rightIndex = from + s.length(); // negative position to positive index
```
wnob, line 3120: I don't think this is the correct semantics. BQ will actually search backward when `position` is negative. To see what I mean, try running `SELECT INSTR("fersher dude", "dude", -3, 1), INSTR("fersher dude", "dude", -4, 1), INSTR("fersher dude", "dude", -5, 1)` in http://pantheon/bigquery

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
jhugomoore: Ah great point! I had thought this felt a little off, thanks!

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
jhugomoore: oooh boy great catch! fixing that now.

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
jhugomoore: in the standard syntax in calcite it was Position(a IN b FROM c) so the c parameter was referred to as from in a couple spots. Decided to keep that especially in the function logic so there isn't a bunch of position variables in the position function.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
jhugomoore: Added some tests and a fix? It's definitely not particularly pretty though.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
julianhyde: Ah, good to know. Please log an Avatica bug (similar to https://issues.apache.org/jira/browse/CALCITE-5415) for that. No urgency to fix.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
julianhyde: Ok, `from` isn't constant. But is `from > 0` constant? Or, if it isn't, should it be? That is, if we go into the `else` branch of the `if` on i = 0, we also want to go into the `else` branch for i = 1 and i = 2.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
jhugomoore: https://issues.apache.org/jira/browse/CALCITE-5682?filter=-2
Done! let me know if anything isn't correct there

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
```diff
@@ -1020,10 +1020,10 @@ ExInstWithCause<CalciteException> failedToAccessField(
   @BaseMessage("No operator for ''{0}'' with kind: ''{1}'', syntax: ''{2}'' during JSON deserialization")
   ExInst<CalciteException> noOperator(String name, String kind, String syntax);
 
-  @BaseMessage("Invalid input for position function from operand value must not be zero")
+  @BaseMessage("Invalid input for POSITION function: from operand value must not be zero")
   ExInst<CalciteException> fromNotZero();
 
-  @BaseMessage("Invalid input for position function occurrence operand value must not be zero")
+  @BaseMessage("Invalid input for POSITION function: occurrence operand value must be positive")
```
julianhyde, line 1026: Don't put stuff at the end of a file. Open source maintainers will hate you.

core/src/main/resources/org/apache/calcite/runtime/CalciteResource.properties
```diff
@@ -336,6 +336,6 @@ InvalidPartitionKeys=Only tables with set semantics may be partitioned. Invalid
 InvalidOrderBy=Only tables with set semantics may be ordered. Invalid ORDER BY clause in the {0,number,#}-th operand of table function ''{1}''
 MultipleRowSemanticsTables=A table function at most has one input table with row semantics. Table function ''{0}'' has multiple input tables with row semantics
 NoOperator=No operator for ''{0}'' with kind: ''{1}'', syntax: ''{2}'' during JSON deserialization
-FromNotZero=Invalid input for position function: from operand value must not be zero
-OccurrenceNotZero=Invalid input for position function: occurrence operand value must not be zero
+FromNotZero=Invalid input for POSITION function: from operand value must not be zero
+OccurrenceNotZero=Invalid input for POSITION function: occurrence operand value must be positive
```
julianhyde, line 340: Don't put stuff at the end of a file. Open source maintainers will hate you.

PR #3185, opened 2023-05-02
[CALCITE-5401] Rule fired by HepPlanner can return Volcano's RelSubset

rubenada:
See problem description in:
[CALCITE-5401](https://issues.apache.org/jira/browse/CALCITE-5401) Rule fired by HepPlanner can return Volcano's RelSubset

Correct the issue via:
- Add new (default) methods `convert` in `RelOptPlanner` (code taken from `RelOptRule#convert`).
- `RelOptRule#convert` methods will just call `RelOptPlanner#convert`.
- All calls to `convert` inside a rule's `onMatch` method will have to go through `RelOptRuleCall.getPlanner().convert(...)` to ensure that the right planner conversion is applied (planner taken from the rule call, not from the RelNode cluster).

Notice that 3rd point is the key to avoid this issue, and it aligns the `onMatch` with other planner-related methods. For example, right now rules that need to call  `RelOptPlanner#prune` inside  `onMatch`  always call `call.getPlanner().prune(...)`  (and not `relNode.getCluster().getPlanner().prune(...)` ). Similarly, rules calling `RelOptPlanner#onCopy` inside `onMatch` do so via `call.getPlanner().onCopy(...)`  (and not via `relNode.getCluster().getPlanner().onCopy(...)` ). Therefore, it seems more consistent that any call to `convert` inside `onMatch` would go via `call.getPlanner().convert(...)` .

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/RelOptPlanner.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptPlanner.java
@@ -350,4 +350,49 @@ public interface RelOptPlanner {
       super(message);
     }
   }
+
+  /**
+   * Converts a relation expression to a given set of traits, if it does not
+   * already have those traits.
+   *
+   * @param rel      Relational expression to convert
+   * @param toTraits desired traits
+   * @return a relational expression with the desired traits; never null
+   */
+  default RelNode convert(RelNode rel, RelTraitSet toTraits) {
+    RelTraitSet outTraits = rel.getTraitSet();
+    for (int i = 0; i < toTraits.size(); i++) {
+      RelTrait toTrait = toTraits.getTrait(i);
+      if (toTrait != null) {
+        outTraits = outTraits.replace(i, toTrait);
+      }
+    }
+
+    if (rel.getTraitSet().matches(outTraits)) {
+      return rel;
+    }
+
+    return this.changeTraits(rel, outTraits);
+  }
+
+  /**
+   * Converts one trait of a relational expression, if it does not
+   * already have that trait.
+   *
+   * @param rel      Relational expression to convert
+   * @param toTrait  Desired trait
+   * @return a relational expression with the desired trait; never null
+   */
+  default RelNode convert(RelNode rel, @Nullable RelTrait toTrait) {
+    RelTraitSet outTraits = rel.getTraitSet();
+    if (toTrait != null) {
+      outTraits = outTraits.replace(toTrait);
+    }
+
+    if (rel.getTraitSet().matches(outTraits)) {
+      return rel;
+    }
+
+    return this.changeTraits(rel, outTraits.simplify());
+  }
 }
--- a/core/src/main/java/org/apache/calcite/plan/RelOptPlanner.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptPlanner.java
@@ -350,49 +350,4 @@ public interface RelOptPlanner {
       super(message);
     }
   }
-
-  /**
-   * Converts a relation expression to a given set of traits, if it does not
-   * already have those traits.
-   *
-   * @param rel      Relational expression to convert
-   * @param toTraits desired traits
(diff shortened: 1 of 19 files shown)


core/src/main/java/org/apache/calcite/plan/RelOptPlanner.java
```diff
@@ -350,4 +350,49 @@ public CannotPlanException(String message) {
       super(message);
     }
   }
+
+  /**
+   * Converts a relation expression to a given set of traits, if it does not
```
zabetak, line 355: I am wondering if it is a good idea to add yet another `convert` method or if it would be better to put this logic into the existing `changeTraits` method inside the planner and change rules to call `changeTraits` directly.

The `VolcanoPlanner#changeTraits` already has some assertions that the traits are different so it may make sense to relax it and bail out instead. 

The `HepPlanner#changeTraits` does not care much about traits so we should be good.

rubenada: I'm a bit concerned of potential side-effects on the existing callers of planner `changeTraits`, which currently don't go through this logic, if they might get impacted if they start going through it...

zabetak: I think the risk is small given that we have the assertions about traits in the Volcano planner. I feel that having less APIs for extension and maintenance is positive for the future of the project but maybe other have a different opinion; feel free to open the discussion to the dev list if you like.

If we decide to go for new methods then I guess I like more your suggestion about adding a new `RelOptRule#convert` with an additional planner parameter.

core/src/main/java/org/apache/calcite/plan/RelOptPlanner.java
```diff
@@ -350,4 +350,49 @@ public CannotPlanException(String message) {
...
+  /**
+   * Converts one trait of a relational expression, if it does not
+   * already have that trait.
+   *
+   * @param rel      Relational expression to convert
+   * @param toTrait  Desired trait
+   * @return a relational expression with the desired trait; never null
+   */
+  default RelNode convert(RelNode rel, @Nullable RelTrait toTrait) {
+    RelTraitSet outTraits = rel.getTraitSet();
```
zoudan, line 387: How about use `return convert(rel, RelTraitSet.createEmpty().plus(toTrait))`, so that we could reuse most of the code.

core/src/main/java/org/apache/calcite/plan/RelOptPlanner.java
rubenada: @zabetak I have applied your idea on the second commit:
- Move convert logic out of `RelOptPlanner` back to `RelOptRule`
- Inside rule `onMatch`, call directly planner `changeTraits` instead of convert
- Relax assertions inside `VolcanoPlanner#changeTraits`

wdyt?

rubenada: Thanks for the feedback @zabetak !
I'm looking again into this, and currently there are a lot of ConverterRules that still call `RelOptRule#convert`, so in order to deprecate it, we should refactor those too (not covered by the PR as it is right now). 
In any case, I think we might want to refactor them and avoid calling `RelOptRule#convert` in their case too, because theoretically the same problem could happen on them (maybe it is a bit more far-fetched to see it in a real production env, but I have been able to force it in a UT). I don't have much time atm to re-work this, I'll try to circle back to it in the coming days...

rubenada: @zabetak  I've created a new PR (https://github.com/apache/calcite/pull/3243) with the idea of new RelOptRule#convert with an additional planner parameter, which perhaps is the most pragmatic and less disruptive solution to correct this problem, wdyt?

rubenada:
Closing this in favor of https://github.com/apache/calcite/pull/3243

PR #3189, opened 2023-05-05
[CALCITE-5674] Cast expr to target type should respect nullable when …

liuyongvs:
Cast expr to target type should respect nullable when it is complex type

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
@@ -43,11 +43,19 @@ import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.SetMultimap;
 
 import java.text.Collator;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import static org.apache.calcite.sql.type.SqlTypeUtil.isArray;
+import static org.apache.calcite.sql.type.SqlTypeUtil.isCollection;
+import static org.apache.calcite.sql.type.SqlTypeUtil.isMap;
+import static org.apache.calcite.sql.type.SqlTypeUtil.isRow;
 import static org.apache.calcite.util.Static.RESOURCE;
+
+import static java.util.Objects.requireNonNull;
 
 /**
  * SqlCastFunction. Note that the std functions are really singleton objects,
@@ -122,8 +130,62 @@ public class SqlCastFunction extends SqlFunction {
   /** Derives the type of "CAST(expression AS targetType)". */
   public static RelDataType deriveType(RelDataTypeFactory typeFactory,
       RelDataType expressionType, RelDataType targetType, boolean safe) {
-    return typeFactory.createTypeWithNullability(targetType,
-        expressionType.isNullable() || safe);
+    return createTypeWithNullabilityFromExpr(typeFactory, expressionType, targetType, safe);
+  }
+
+  private static RelDataType createTypeWithNullabilityFromExpr(RelDataTypeFactory typeFactory,
+      RelDataType expressionType, RelDataType targetType, boolean safe) {
+    boolean isNullable = expressionType.isNullable() || safe;
+
+    if (isCollection(expressionType)) {
+      RelDataType expressionElementType = expressionType.getComponentType();
+      RelDataType targetElementType = targetType.getComponentType();
+      requireNonNull(expressionElementType, () -> "componentType of " + expressionType);
+      requireNonNull(targetElementType, () -> "componentType of " + targetType);
+      RelDataType newElementType =
+          createTypeWithNullabilityFromExpr(
+              typeFactory, expressionElementType, targetElementType, safe);
+      return isArray(expressionType)
+          ? SqlTypeUtil.createArrayType(typeFactory, newElementType, isNullable)
+          : SqlTypeUtil.createMultisetType(typeFactory, newElementType, isNullable);
+    }
+
+    if (isRow(expressionType)) {
+      final int fieldCount = expressionType.getFieldCount();
+      List<RelDataType> typeList = new ArrayList<>();
+      for (int i = 0; i < fieldCount; ++i) {
+        RelDataType expressionElementType = expressionType.getFieldList().get(i).getType();
+        RelDataType targetElementType = targetType.getFieldList().get(i).getType();
+        typeList.add(
+            createTypeWithNullabilityFromExpr(typeFactory, expressionElementType,
+            targetElementType, safe));
+      }
+      return typeFactory.createTypeWithNullability(
+          typeFactory.createStructType(
+              typeList,
+              targetType.getFieldNames()), isNullable);
+    }
+
+    if (isMap(expressionType)) {
+      RelDataType expressionKeyType =
+          requireNonNull(expressionType.getKeyType(), () -> "keyType of " + expressionType);
+      RelDataType expressionValueType =
+          requireNonNull(expressionType.getValueType(), () -> "valueType of " + expressionType);
+      RelDataType targetKeyType =
+          requireNonNull(targetType.getKeyType(), () -> "keyType of " + targetType);
+      RelDataType targetValueType =
+          requireNonNull(targetType.getValueType(), () -> "valueType of " + targetType);
+
+      RelDataType keyType =
+          createTypeWithNullabilityFromExpr(
+              typeFactory, expressionKeyType, targetKeyType, safe);
+      RelDataType valueType =
+          createTypeWithNullabilityFromExpr(
+              typeFactory, expressionValueType, targetValueType, safe);
+      SqlTypeUtil.createMapType(typeFactory, keyType, valueType, isNullable);
+    }
+
+    return typeFactory.createTypeWithNullability(targetType, isNullable);
   }
 
   @Override public String getSignatureTemplate(final int operandsCount) {
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
@@ -152,13 +152,13 @@ public class SqlCastFunction extends SqlFunction {
 
     if (isRow(expressionType)) {
       final int fieldCount = expressionType.getFieldCount();
-      List<RelDataType> typeList = new ArrayList<>();
+      final List<RelDataType> typeList = new ArrayList<>(fieldCount);
       for (int i = 0; i < fieldCount; ++i) {
         RelDataType expressionElementType = expressionType.getFieldList().get(i).getType();
         RelDataType targetElementType = targetType.getFieldList().get(i).getType();
         typeList.add(
             createTypeWithNullabilityFromExpr(typeFactory, expressionElementType,
-            targetElementType, safe));
+                targetElementType, safe));
       }
       return typeFactory.createTypeWithNullability(
           typeFactory.createStructType(
--- a/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
@@ -7036,7 +7036,7 @@ public class SqlValidatorTest extends SqlValidatorTestCase {
     sql("select cast(a as row(f0 int, f1 varchar)) from COMPLEXTYPES.CTC_T1")
         .withExtendedCatalog()
         .columnType("RecordType(INTEGER NOT NULL F0, VARCHAR NOT NULL F1) NOT NULL");
-    sql("select cast(b as row(f0 int, f1 varchar))\n"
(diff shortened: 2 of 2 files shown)


liuyongvs:
hi @JiajunBernoulli do you also have time to review it?

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
```diff
@@ -122,8 +128,50 @@ static SqlReturnTypeInference returnTypeInference(boolean safe) {
...
+
+  public static RelDataType createTypeWithNullabilityFromExpr(RelDataTypeFactory typeFactory,
+      RelDataType expressionType, RelDataType targetType, boolean safe) {
+    boolean isn = expressionType.isNullable() || safe;
+
+    if (isCollection(expressionType)) {
+      RelDataType expressionElementType = expressionType.getComponentType();
+      RelDataType targetElementType = targetType.getComponentType();
+      RelDataType newElementType =
+          createTypeWithNullabilityFromExpr(typeFactory, expressionElementType, targetElementType, false);
```
libenchao, line 142: The third parameter should be `safe`?

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
```diff
@@ -122,8 +128,50 @@ static SqlReturnTypeInference returnTypeInference(boolean safe) {
...
+      if (isArray(expressionType)) {
+        return SqlTypeUtil.createArrayType(typeFactory, newElementType, isn);
+      } else {
+        return SqlTypeUtil.createMultisetType(typeFactory, newElementType, isn);
+      }
+    }
+
+    if (isRow(expressionType)) {
+      int fieldCount = expressionType.getFieldCount();
```
libenchao, line 151: We usually add `final` to variables when it's possible.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -6984,9 +6984,9 @@ void testGroupExpressionEquivalenceParams() {
 
   @Test void testCastAsCollectionType() {
     sql("select cast(array[1,null,2] as int array) from (values (1))")
-        .columnType("INTEGER NOT NULL ARRAY NOT NULL");
+        .columnType("INTEGER ARRAY NOT NULL");
```
libenchao, line 6987: Could you also try to add tests for map/row/multiset types?

libenchao (commented):
@liuyongvs Thanks for your contribution, I've left my comments.

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
liuyongvs: it should not. because the try_cast/safe_cast(different vendors, flink/spark former, mssql latter) will return null when it can not cast to targettype instead of throw exception like cast.

it only changes the outer nullable

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
```diff
@@ -122,8 +128,50 @@ static SqlReturnTypeInference returnTypeInference(boolean safe) {
...
   public static RelDataType deriveType(RelDataTypeFactory typeFactory,
       RelDataType expressionType, RelDataType targetType, boolean safe) {
-    return typeFactory.createTypeWithNullability(targetType,
-        expressionType.isNullable() || safe);
+    return createTypeWithNullabilityFromExpr(typeFactory, expressionType, targetType, safe);
+  }
+
+  public static RelDataType createTypeWithNullabilityFromExpr(RelDataTypeFactory typeFactory,
+      RelDataType expressionType, RelDataType targetType, boolean safe) {
+    boolean isn = expressionType.isNullable() || safe;
```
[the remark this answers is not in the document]
liuyongvs: i does it, becasue other place does it too, i will change it.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
liuyongvs: add test for multiset types and row.

because calcite doesn't support map type in parser, so i doesn't add.

```
java.lang.RuntimeException: Error while parsing query: select cast(map['washington', 1, null, 44] as map[CHAR(10), INTEGER NOT NULL>] from (values (1))
	at org.apache.calcite.sql.test.AbstractSqlTester.parseAndValidate(AbstractSqlTester.java:157)
	at org.apache.calcite.sql.test.AbstractSqlTester.validateAndThen(AbstractSqlTester.java:248)
	at org.apache.calcite.sql.test.AbstractSqlTester.checkColumnType(AbstractSqlTester.java:177)
	at org.apache.calcite.test.SqlValidatorFixture.columnType(
```

liuyongvs:
hi @libenchao  fix all your reviews, please have a look again.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
liuyongvs: add this grammer should not support select cast(b as row(f0 int not null, f1 varchar null) , i will fix it a new pr latter

sql standard below
```
<cast specification> ::=
CAST <left paren> <cast operand> AS <cast target> <right paren>
<cast operand> ::=
<value expression>
| <implicitly typed value specification>
<cast target> ::=
<domain name>
| <data type>

<data type> ::=
<predefined type>
| <row type>
| <path-resolved user-defined type name>
| <reference type>
| <collection type>

<row type> ::=
ROW <row type body>
<row type body> ::=
<left paren> <field definition> [ { <comma> <field definition> }... ] <right paren>

<field definition> ::=
<field name> <data type>
```

liuyongvs:
hi @libenchao how do i format code with command to fix checkstyle problem 
[Task :core:checkstyleMain] [LineLength] Line is longer than 100 characters (found 118).

i does it but not take effect
$ ./gradlew style # update code formatting (for auto-correctable cases) and verify style

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
```diff
@@ -122,8 +128,50 @@ static SqlReturnTypeInference returnTypeInference(boolean safe) {
...
+
+    if (isCollection(expressionType)) {
+      RelDataType expressionElementType = expressionType.getComponentType();
+      RelDataType targetElementType = targetType.getComponentType();
+      RelDataType newElementType =
+          createTypeWithNullabilityFromExpr(typeFactory, expressionElementType, targetElementType, false);
+      if (isArray(expressionType)) {
+        return SqlTypeUtil.createArrayType(typeFactory, newElementType, isNullable);
+      } else {
+        return SqlTypeUtil.createMultisetType(typeFactory, newElementType, isNullable);
```
chunweilei, line 146: Maybe we can use `isArray(expressionType) ?  xxx: yyy` to be more concise.

liuyongvs: fixed @chunweilei

libenchao:
> hi @libenchao how do i format code with command to fix checkstyle problem automatic like flink using spotless plugin [Task :core:checkstyleMain] [LineLength] Line is longer than 100 characters (found 118).
> 
> i does it but not take effect in https://calcite.apache.org/docs/howto.html#setting-up-an-ide-for-contributing $ ./gradlew style # update code formatting (for auto-correctable cases) and verify style

`./gradlew autostyleApply` should work.

liuyongvs:
@libenchao thanks

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
libenchao: I'm not sure about this, if the inner type is `not null`, then safe cast should be `nullable` because the cast may produce `null`?

liuyongvs: yes @libenchao  
for exmpale:
a type is varchar not null
cast(a as timestamp)  throw exception
safe_cast/try_cast(a as timestamp) return null, the return type is varchar, which can be nullable.

libenchao: This is not what I meant to say. 

Take a example, `ARRAY<VARCHAR NOT NULL> NOT NULL`, safe_cast it to `ARRAY<INT NULLABLE> NULLABLE`, what should the result be? IMO, the inner type should be `NULLABLE` because it may produce `null`s.

liuyongvs:
hi @libenchao @chunweilei  ci passed, do you have time to look again

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
liuyongvs: @libenchao why?
1. cast target type should not contains nullable like your example ARRAY<INT NULLABLE> NULLABLE
2. the array type and element type is different 
    array type nullable( outer nullable), the value is null
    element type nullable (inner nullable), the value is [null]

liuyongvs:
@libenchao i am sorry for do not see your comments. i response to it just now.

liuyongvs:
squash the commits

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
libenchao: because in safe mode, we'll produce `null`s, maybe a more concrete example explains this better: "['123', '456', 'abc']", if we cast it to `array<int>`, what should the result be? "[123, 456, null]", or "null" ? I don't know what other databases behaves in this case, but in my opinion, the former one is better. If this is also what other databases which provide safe cast be behaves, then the type for element type should also be nullable.

liuyongvs: fixed and add some test for try_cast and safe_cast

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
```diff
@@ -122,8 +130,62 @@ static SqlReturnTypeInference returnTypeInference(boolean safe) {
...
+          createTypeWithNullabilityFromExpr(
+              typeFactory, expressionElementType, targetElementType, safe);
+      return isArray(expressionType)
+          ? SqlTypeUtil.createArrayType(typeFactory, newElementType, isNullable)
+          : SqlTypeUtil.createMultisetType(typeFactory, newElementType, isNullable);
+    }
+
+    if (isRow(expressionType)) {
+      final int fieldCount = expressionType.getFieldCount();
+      List<RelDataType> typeList = new ArrayList<>();
```
libenchao, line 155: NIT: add `final`, and initialize `ArrayList` with `fieldCount`

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
```diff
@@ -122,8 +130,62 @@ static SqlReturnTypeInference returnTypeInference(boolean safe) {
...
+      for (int i = 0; i < fieldCount; ++i) {
+        RelDataType expressionElementType = expressionType.getFieldList().get(i).getType();
+        RelDataType targetElementType = targetType.getFieldList().get(i).getType();
+        typeList.add(
+            createTypeWithNullabilityFromExpr(typeFactory, expressionElementType,
+            targetElementType, safe));
```
libenchao, line 161: NIT: add indent.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -7033,7 +7036,7 @@ void testGroupExpressionEquivalenceParams() {
     sql("select cast(a as row(f0 int, f1 varchar)) from COMPLEXTYPES.CTC_T1")
         .withExtendedCatalog()
         .columnType("RecordType(INTEGER NOT NULL F0, VARCHAR NOT NULL F1) NOT NULL");
-    sql("select cast(b as row(f0 int not null, f1 varchar null))\n"
+    sql("select cast(b as row(f0 int, f1 varchar))\n"
```
libenchao, line 7039: why are you changing this test and the below one?

libenchao (commented):
@liuyongvs Thanks for the updating, I've left some more comments.

Besides, please do not squash your commits before a reviewer asks you because it will make the reviewer harder to tracking changes.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
liuyongvs: just for test row and i fallback it.
then i will submit new pr to fix cast( type not null/ null) latter according sql standard, what do you think?

libenchao: sounds good to me, one commit should focus on one thing

libenchao (approved):
+1, will merge after the CI passes

PR #3191, opened 2023-05-08
[CALCITE-5693] AssertionError with UNION using CTE

kramerul:


core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -135,6 +135,14 @@ class JdbcAdapterTest {
             + "WHERE \"product_id\" = 1");
   }
 
+  @Test void testFilterUnionIncludingWithPlan() {
+    CalciteAssert.model(FoodmartSchema.FOODMART_MODEL)
+        .query("  ( with a as (select * from \"sales_fact_1997\")  select * from a)\n"
```
zoudan, line 140: It is better to re-beautify the format, e.g. remove inappropriate spaces

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4021,19 +4021,30 @@ private void validateModality(SqlNode query) {
...
-          throw newValidationError(operand,
-              Static.RESOURCE.streamSetOpInconsistentInputs());
-        }
-        validateModality(operand);
+        validateModality(operand, modality);
       }
     }
   }
 
+  private  void validateModality(SqlNode operand, SqlModality modality) {
```
zoudan, line 4040: remove one space

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -135,6 +135,14 @@ class JdbcAdapterTest {
             + "WHERE \"product_id\" = 1");
   }
 
+  @Test void testFilterUnionIncludingWithPlan() {
```
zoudan, line 138: I do no see any `filter` in the query, maybe we should rename the method name?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4021,19 +4021,30 @@ private void validateModality(SqlNode query) {
...
+    if (deduceModality(operand) != modality) {
+      throw newValidationError(operand,
+          Static.RESOURCE.streamSetOpInconsistentInputs());
```
libenchao, line 4043: The exception is no longer accurate since you are also validating with list.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
libenchao, line 138: Since the exception is thrown in the validating phase, it's better to add the test to `SqlValidatorTest`

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4021,19 +4021,30 @@ private void validateModality(SqlNode query) {
       default:
         break;
       }
+    } else if (query.getKind() == SqlKind.WITH) {
```
libenchao, line 4024: This would open the door to support `WITH` in streaming, then it would be great to have some tests for this.

libenchao (commented):
@kramerul Thanks for the contribution! I've left my comments.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
kramerul: Done

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
kramerul: Done

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
kramerul: But I did not change the logic. I only extracted existing code into this new method.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
kramerul: My intention is not to open support for `WITH` in streaming. I would like to fix the bug that `WITH` is not working together with `UNION`.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
kramerul: I added more checks to make this test fit into this file.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
kramerul: I am not really familiar with Calcite core development. So it will take me quite a bit of time to add tests for streaming here.

libenchao: It's a nice-to-have, not a must-have request, I'm ok with this.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
libenchao: I don't agree, you actually enlarged the scope of this code: it is only used for set query before, but now it also handles with query.

mihaibudiu:

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
kramerul: I added a new message.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -180,6 +180,20 @@ class JdbcAdapterTest {
             + "WHERE \"product_id\" = 1) AS \"t3\"");
   }
 
+  @Test void testUnionIncludingCte() {
+    CalciteAssert.model(FoodmartSchema.FOODMART_MODEL)
```
NobiGo, line 184: This test case can be executed without any modification. Please reconfirm it.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4439,20 +4439,29 @@ private void validateModality(SqlNode query) {
      // SQL: WITH STREAMTABLE AS (SELECT STREAM * FROM KAFKA.MOCKTABLE)
      //      SELECT STREAM * FROM STREAMTABLE
      // The modality should be STREAM.
-      validateModality(((SqlWith) query).body);
+      SqlWith with = (SqlWith) query;
```
NobiGo, line 4442: I hope the code comments here can also provide some explanations.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
kramerul: I also tested it a few minutes ago. I can confirm that this PR is no longer required.

PR #3192, opened 2023-05-08
[CALCITE-5602] Implement CSC and SEC function

zoudan:


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1219,6 +1219,20 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
           OperandTypes.NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
+  @LibraryOperator(libraries = {ALL})
+  public static final SqlFunction CSC =
+      SqlBasicFunction.create("CSC",
+          ReturnTypes.DOUBLE_NULLABLE,
```
chunweilei, line 1225: Do all the libraries have CSE and SEC function？

rubenada (approved):

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
```diff
@@ -440,6 +440,8 @@ public enum BuiltInMethod {
   LOG(SqlFunctions.class, "log", long.class, long.class),
   TANH(SqlFunctions.class, "tanh", long.class),
   SINH(SqlFunctions.class, "sinh", long.class),
+  CSC(SqlFunctions.class, "csc", long.class),
```
rubenada, line 443: @zoudan if I am not mistaken, this change in BuiltInMethod is not required: these `CSC` and `SEC` are not used anywhere, so they should be removed to avoid confusion.
If we check for example a similar PR: https://github.com/apache/calcite/pull/3130 ; in that case BuiltInMethod .java was not impacted, so I think it should be the same in this case.

zoudan: Yes, you are right, I add these by mistake. I will remove them.

zoudan: updated

rubenada:
Thanks @zoudan , LGTM.
Could you please squash commits into a single one? Also may I ask you a minor correction on the final commit message: `[CALCITE-5602] Implement CSC and SEC functions`  (instead of "function", singular).
I'll proceed with the merge after that.

PR #3193, opened 2023-05-09
[CALCITE-5683] Two level nested correlated subquery throws an excepti…

HanumathRao:
…on during decorrelation.

Currently two level correlated subqueries with more than one correlated variables are throwing exception, the fix is to track the correlated variables and then only consider those variables from the level which a logicalCorrelate node is created.

Added two test cases and checked for the expected plan in RelOptRulesTest.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
@@ -866,6 +866,8 @@ public class SubQueryRemoveRule
           LogicVisitor.find(RelOptUtil.Logic.TRUE, ImmutableList.of(c), e);
       final Set<CorrelationId>  variablesSet =
           RelOptUtil.getVariablesUsed(e.rel);
+      /* Only consider the correlated variables which originated from this sub-query level*/
+      variablesSet.retainAll(filter.getVariablesSet());
       final RexNode target =
           rule.apply(e, variablesSet, logic,
               builder, 1, builder.peek().getRowType().getFieldCount(), count);
--- a/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
@@ -866,10 +866,9 @@ public class SubQueryRemoveRule
           LogicVisitor.find(RelOptUtil.Logic.TRUE, ImmutableList.of(c), e);
       final Set<CorrelationId>  variablesSet =
           RelOptUtil.getVariablesUsed(e.rel);
-      // Only consider the correlated variables which originated from this sub-query level.
+      // Only retain when filter has non-empty correlated variables set to avoid removing
+      // all the correlated variables from parent scope.
       if (!filter.getVariablesSet().isEmpty()) {
-        // Only retain when filter has non-empty correlated variables set to avoid removing
-        // all the correlated variables from parent scope.
         variablesSet.retainAll(filter.getVariablesSet());
       }
       final RexNode target =
--- a/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
@@ -866,7 +866,7 @@ public class SubQueryRemoveRule
           LogicVisitor.find(RelOptUtil.Logic.TRUE, ImmutableList.of(c), e);
       final Set<CorrelationId>  variablesSet =
           RelOptUtil.getVariablesUsed(e.rel);
-      /* Only consider the correlated variables which originated from this sub-query level*/
+      // Only consider the correlated variables which originated from this sub-query level
       variablesSet.retainAll(filter.getVariablesSet());
       final RexNode target =
           rule.apply(e, variablesSet, logic,
--- a/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
@@ -866,8 +866,10 @@ public class SubQueryRemoveRule
           LogicVisitor.find(RelOptUtil.Logic.TRUE, ImmutableList.of(c), e);
       final Set<CorrelationId>  variablesSet =
           RelOptUtil.getVariablesUsed(e.rel);
-      // Only consider the correlated variables which originated from this sub-query level
+      // Only consider the correlated variables which originated from this sub-query level.
       if (!filter.getVariablesSet().isEmpty()) {
+        // Only retain when filter has non-empty correlated variables set to avoid removing
+        // all the correlated variables from parent scope.
         variablesSet.retainAll(filter.getVariablesSet());
       }
       final RexNode target =
--- a/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
@@ -866,11 +866,8 @@ public class SubQueryRemoveRule
           LogicVisitor.find(RelOptUtil.Logic.TRUE, ImmutableList.of(c), e);
       final Set<CorrelationId>  variablesSet =
           RelOptUtil.getVariablesUsed(e.rel);
-      // Only retain when filter has non-empty correlated variables set to avoid removing
-      // all the correlated variables from parent scope.
-      if (!filter.getVariablesSet().isEmpty()) {
-        variablesSet.retainAll(filter.getVariablesSet());
-      }
+      // Only consider the correlated variables which originated from this sub-query level.
+      variablesSet.retainAll(filter.getVariablesSet());
       final RexNode target =
           rule.apply(e, variablesSet, logic,
               builder, 1, builder.peek().getRowType().getFieldCount(), count);
--- a/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
@@ -867,7 +867,9 @@ public class SubQueryRemoveRule
       final Set<CorrelationId>  variablesSet =
           RelOptUtil.getVariablesUsed(e.rel);
       // Only consider the correlated variables which originated from this sub-query level
-      variablesSet.retainAll(filter.getVariablesSet());
+      if (!filter.getVariablesSet().isEmpty()) {
+        variablesSet.retainAll(filter.getVariablesSet());
+      }
       final RexNode target =
           rule.apply(e, variablesSet, logic,
               builder, 1, builder.peek().getRowType().getFieldCount(), count);
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -3549,7 +3549,17 @@ public class SqlToRelConverter {
     // implement HAVING (we have already checked that it is non-trivial)
     relBuilder.push(bb.root());
     if (havingExpr != null) {
-      relBuilder.filter(havingExpr);
+      /* Set the correlation variables used in this sub-query to the filter node,
+       * same logic is being used for the filter generated in where clause.
+       */
+      Set<CorrelationId> variableSet = new HashSet<>();
+      if (havingExpr instanceof RexSubQuery) {
+        CorrelationUse p = getCorrelationUse(bb, ((RexSubQuery) havingExpr).rel);
+        if (p != null) {
+          variableSet.add(p.id);
+        }
+      }
+      relBuilder.filter(variableSet, havingExpr);
     }
 
     // implement the SELECT list
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -3549,9 +3549,8 @@ public class SqlToRelConverter {
     // implement HAVING (we have already checked that it is non-trivial)
     relBuilder.push(bb.root());
     if (havingExpr != null) {
-      /* Set the correlation variables used in this sub-query to the filter node,
-       * same logic is being used for the filter generated in where clause.
-       */
+      // Set the correlation variables used in this sub-query to the filter node,
+      // same logic is being used for the filter generated in where clause.
       Set<CorrelationId> variableSet = new HashSet<>();
       if (havingExpr instanceof RexSubQuery) {
         CorrelationUse p = getCorrelationUse(bb, ((RexSubQuery) havingExpr).rel);
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -3552,8 +3552,9 @@ public class SqlToRelConverter {
       // Set the correlation variables used in this sub-query to the filter node,
       // same logic is being used for the filter generated in where clause.
       Set<CorrelationId> variableSet = new HashSet<>();
-      if (havingExpr instanceof RexSubQuery) {
-        CorrelationUse p = getCorrelationUse(bb, ((RexSubQuery) havingExpr).rel);
+      RexSubQuery subQ = RexUtil.SubQueryFinder.find(havingExpr);
+      if (subQ != null) {
+        CorrelationUse p = getCorrelationUse(bb, subQ.rel);
         if (p != null) {
           variableSet.add(p.id);
         }
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -133,6 +133,7 @@ import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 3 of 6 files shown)


core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6823,6 +6824,65 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
         .checkUnchanged();
   }
 
+  /** Test case for CALCITE-5683 for two level nested decorrelate with standard program
+   * failing during the decorrelation phase. */
+  @Test void testTwoLevelDecorrelate() {
```
libenchao, line 6829: The test style is different from existing ones, could you change it to be the same style with existing testings?

HanumathRao: @libenchao Thanks for reviewing the PR. Yeah this test case is different in style because the issue is showing up only with standard program. Let me know if there is a better place(file) to add these test cases.

HanumathRao: @libenchao Just want to circle back on the review comment.Please let me know if you have any suggestions. Thanks

rubenada: A possible alternative might be putting the tests inside a `*.iq` file (I guess `sub-query.iq` ?). What do you guys think?

libenchao: Since the exception is thrown in `RelDecorrelator`, I would try to enable `withSubQueryRules().withLateDecorrelate(true)` to reproduce the problem. 

`QuidemTest` (*.iq) is also a place we can put the tests, usually I will add tests there if the problem occurs in the 'runtime', or the problem occurs in earlier phase but we also want to verify the result because the plan is not easy to verify correctness.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -3549,7 +3549,17 @@ protected final void createAggImpl(
     // implement HAVING (we have already checked that it is non-trivial)
     relBuilder.push(bb.root());
     if (havingExpr != null) {
-      relBuilder.filter(havingExpr);
+      /* Set the correlation variables used in this sub-query to the filter node,
+       * same logic is being used for the filter generated in where clause.
+       */
+      Set<CorrelationId> variableSet = new HashSet<>();
```
rubenada, line 3555: Would it possible to have some new test(s) in `SqlToRelConverterTest.java` showing the problem that this piece of code fixes?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
rubenada: Thanks @HanumathRao , I think it looks good. I have added another comment to see if we can provide more tests for the SqlToRel case.
@libenchao what do you think?

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -866,6 +866,8 @@ private static void matchFilter(SubQueryRemoveRule rule,
           LogicVisitor.find(RelOptUtil.Logic.TRUE, ImmutableList.of(c), e);
       final Set<CorrelationId>  variablesSet =
           RelOptUtil.getVariablesUsed(e.rel);
+      /* Only consider the correlated variables which originated from this sub-query level*/
+      variablesSet.retainAll(filter.getVariablesSet());
```
rubenada, line 870: Interestingly, when I apply this fix in my local environment and run CoreQuidemTest, I get a regression in `misc.iq` :
```
< +-----+---+
< | KEY | C |
< +-----+---+
< |   2 | 1 |
< +-----+---+
< (1 row)
< 
---
> java.lang.AssertionError: Correlation variable $cor0 should be defined
> 	at org.apache.calcite.adapter.enumerable.EnumerableRelImplementor.getCorrelVariableGetter(EnumerableRelImplementor.java:476)
> 	at org.apache.calcite.adapter.enumerable.RexToLixTranslator.visitFieldAccess(RexToLixTranslator.java:1393)
> 	at org.apache.calcite.adapter.enumerable.RexToLixTranslator.visitFieldAccess(RexToLixTranslator.java:101)
> 	at org.apache.calcite.rex.RexFieldAccess.accept(RexFieldAccess.java:94)
...
```
Not sure why this is not shown on the PR checks...

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
HanumathRao: Without this fix there is an existing test case failure (I think it is in CoreQuidemTest).

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
rubenada: Yep, sorry, my bad. I confirm if I apply the fix in both `SubQueryRemoveRule` *and* `SqlToRelConverter`, then `misc.iq` test run successfully.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
rubenada: Yes, correct, `misc.iq` would fail without this code, as described on the other conversation. Thanks for the clarification.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6823,6 +6823,48 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
...
+        + "                         e1.deptno = e2.deptno and"
+        + "                         d1.deptno < e2.deptno))"
+        + " FROM dept d1";
+
+    sql(sql)
+        .withExpand(false)
+        .withLateDecorrelate(true)
+        .withTrim(true)
+        .withRule()
+        .checkUnchanged();
```
libenchao, line 6845: This seems weird that you are asserting the plan is not changed, but the content in `RelOptRulesTest.xml` does contain `planAfter` section, and the tests still pass.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -3549,7 +3549,17 @@ protected final void createAggImpl(
...
+      if (havingExpr instanceof RexSubQuery) {
```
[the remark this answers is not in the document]
HanumathRao: Agree that this might be taking care of limited cases, I didn't want to increase the scope of this PR. Please let me know if you foresee any use cases that is missing then I can take care of it in this PR (otherwise I think it might be better to handle it in another PR).

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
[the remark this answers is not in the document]
HanumathRao: Agree that this fix is using the existing mechanism to set the variable in Filter operator, I did have questions like what happens when there other operator nodes etc. I want to handle the specific case for this PR(as you mentioned going in the right direction) and handle as we have more use cases (please let me know if it makes sense).

Agree with you that this code might result in a throwing an exception in cases where previous code was generating a plan. However, I encountered a scenario while debugging the JIRA([CALCITE-5716](https://issues.apache.org/jira/browse/CALCITE-5716)) wherein it was not reporting an error but generating a wrong plan. With the fix provided in this PR it is generating a right result (when running a generated plan). Updated the JIRA with the analysis

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
HanumathRao: Yeah, please let me know if it makes sense to remove these tests (as they actually are redundant) as sub-query.iq is having these test cases and are reporting an error without the fix.

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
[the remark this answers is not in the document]
libenchao: An important case would be the simple one, `Filter` with a correlated condition, and no correlation variable in `Filter`. This case would break I guess.

There is a trick way to avoid this, we only do this when `filter.getVariablesSet()` is not empty, then we can avoid this. What do you guys think?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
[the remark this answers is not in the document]
libenchao: Take an example: `select count(*) from emp group by ename having count(*) > 1 and (your correlated sub-query)`. This would not match this condition. We can easily handle this using `RexUtil.SubQueryFinder.find`, what do you think?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
libenchao: I think it's worth to keep these tests, my question is about why `checkUnchanged` test contains `planAfter`, I'll check this and give my further comment.

libenchao: Now I know the reason, and I've logged CALCITE-5727 for further work. For this case, you can delete the `planAfter` section directly in `RelOptRulesTest.xml`, that does not affect the test, and there is no need to add this section.

HanumathRao: OK, thank you for working on it.

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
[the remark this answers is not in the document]
HanumathRao: Yeah, it makes sense to add that changes to be on the safe side. I have changed the code. thanks

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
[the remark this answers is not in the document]
HanumathRao: Added the code changes to find the SubQuery and also a test case to exercise this code path. thanks

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -866,6 +866,10 @@ private static void matchFilter(SubQueryRemoveRule rule,
           LogicVisitor.find(RelOptUtil.Logic.TRUE, ImmutableList.of(c), e);
       final Set<CorrelationId>  variablesSet =
           RelOptUtil.getVariablesUsed(e.rel);
+      // Only consider the correlated variables which originated from this sub-query level
+      if (!filter.getVariablesSet().isEmpty()) {
```
libenchao, line 870: NIT: how about add a comment for this condition to mention about what we've discussed, I think it's valuable for others to understand it.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -14683,6 +14683,115 @@ LogicalProject(DEPTNO=[$0], DEPTNO0=[$1])
...
+  LogicalProject(SAL=[$5])
+    LogicalFilter(condition=[AND(=($cor1.SAL, $5), =($cor1.DEPTNO, $7), <($cor0.DEPTNO, $7))])
+      LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+})))], variablesSet=[[$cor1]])
+    LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+}))])
+  LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
+]]>
+    </Resource>
+    <Resource name="planAfter">
```
libenchao, line 14723: As mentioned before, you can delete this 'planAfter'.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -14683,6 +14683,115 @@ LogicalProject(DEPTNO=[$0], DEPTNO0=[$1])
...
+  LogicalProject(SAL=[$5])
+    LogicalFilter(condition=[AND(=($cor1.SAL, $5), =($cor1.DEPTNO, $7), <($cor0.DEPTNO, $7))])
+      LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+})))], variablesSet=[[$cor1]])
+    LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+}))])
+  LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
+]]>
+    </Resource>
+    <Resource name="planMid">
```
libenchao, line 14708: Although you are using `withLateDecorrelate(true)`, actually it does nothing because it handles `Correlate` instread `RexSubQuery`. Could you add `withSubQueryRules()` to make decorrelation work?

libenchao (commented):
A general tip: please do not use squash/force-push before the reviewer asks you to, this will make the reviewer hard to track incremental changes.

HanumathRao:
@libenchao  and @rubenada  thanks for the review comments. I have fixed all the comments, can you please review and let me know.

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -866,6 +866,12 @@ private static void matchFilter(SubQueryRemoveRule rule,
           LogicVisitor.find(RelOptUtil.Logic.TRUE, ImmutableList.of(c), e);
       final Set<CorrelationId>  variablesSet =
           RelOptUtil.getVariablesUsed(e.rel);
+      // Only consider the correlated variables which originated from this sub-query level.
+      if (!filter.getVariablesSet().isEmpty()) {
+        // Only retain when filter has non-empty correlated variables set to avoid removing
+        // all the correlated variables from parent scope.
```
libenchao, line 872: This comment explains why the `if (...)` exists, how about moving it above the `if (...)`.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6823,6 +6823,50 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
...
+        + "                         e1.deptno = e2.deptno and"
+        + "                         d1.deptno < e2.deptno))"
+        + " FROM dept d1";
+
+    sql(sql)
+        .withExpand(false)
+        .withLateDecorrelate(true)
+        .withSubQueryRules()
+        .withTrim(true)
+        .withRule()
```
libenchao, line 6845: `withSubQueryRules` will be overridden by `withRule`, so the test actually has no rules (the content in the *.xml file also shows that the SubQuery is actually not removed). Can you change this to something like `withSubQueryRules().withLateDecorrelate().withTrim().check()`.

NIT: `withExpand(false)` is not needed now since it's false by default now.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6823,6 +6823,50 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
...
+        .checkUnchanged();
+  }
+
+  /** Test case for CALCITE-5683 for two level nested decorrelate with standard program
+   * failing during the decorrelation phase. */
+  @Test void testTwoLevelDecorrelateSkipInBetween() {
```
libenchao, line 6851: The only difference of this test from `testTwoLevelDecorrelate` is you removed `d1.deptno = e1.deptno` condition, and I'm not sure what is the meaning for `SkipInBetween` in the test name, and the reason why we need this test, can you explain it a bit?

core/src/test/resources/sql/sub-query.iq
```diff
@@ -3573,4 +3573,48 @@ or 20 in (
 
 !ok
 
+# Test case for CALCITE-5683 which throws an exception during the de-correlation phase
+SELECT d1.dname, d1.deptno +
+        (   SELECT max(e1.empno) FROM emp e1
```
libenchao, line 3578: Usually we start a new line for `FROM` clause.

libenchao (commented):
@HanumathRao Sorry about the late reply. The changes almost looks good, I've left some comments about the tests.

Besides, you mentioned that CALCITE-5716 is also being fixed by this, could you add some tests about this?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
HanumathRao: By SkipInBetween, I actually mean't that the correlation variable is not used at the first level and directly used at the deeper level (here second level depth). I wanted to make sure that even if there is no correlation used in the intermediate levels, the solution works.

libenchao:
@HanumathRao I saw you pushed a few new commits, is this ready for review now? If yes, I'll do it in the following days.

HanumathRao:
@libenchao Thank you for tracking the PR. Yeah I have made the changes for addressing review comments.I added the test case which reproduces the issue mentioned in CALCITE-5716. Please review it and let me know. 

The if check which was done to safeguard against possible regressions is removed because with that check in place it will not fix CALCITE-5716.

I am travelling in the next 5 days and I will have limited wifi connectivity, so please expect some delay in my response.

libenchao:
@HanumathRao Thanks for letting me know.

I've gone through the PR again, I find that the `if (filter.getVariablesSet().size() > 0)` has been reverted, so I added it back, and added a test for it. After that, it seems that the last test in `sub-query.iq` is broken.

The changes I made is in https://github.com/libenchao/calcite/commits/5683-nested-correlated-sub-query-throws-2, it also contains some fixups for coding style. It seems that there is still something that does not work well.

rubenada:
Closing this in favor of https://github.com/apache/calcite/pull/3277 , which integrates the changes from the current PR too.

PR #3196, opened 2023-05-11
[CALCITE-5691] IN sub-query inside FILTER clause throws IndexOutOfBoundsException

herunkang2018:


herunkang2018:
There is a failure in code check in Window JDK8, which is not related to this PR. Would someone please tell me how to re-run the test?
> FAILURE   0.1sec,    1 completed,   1 failed,   0 skipped, org.apache.calcite.adapter.geode.rel.GeodeBookstoreTest

gianm:
I've re-run it. If you're a committer you can do it by clicking on the test, then going to the "attempt" drop down in the upper right and selecting "Re-run failed tests".

herunkang2018:
@julianhyde @libenchao Would you like to review this PR if you have time?

core/src/main/java/org/apache/calcite/plan/RelOptPredicateList.java
```diff
@@ -233,7 +234,8 @@ public boolean isEffectivelyNotNull(RexNode e) {
         return true;
       }
     }
-    if (SqlKind.COMPARISON.contains(e.getKind())) {
+    // Skip IN RexSubQuery because it is not literal we want to handle here
+    if (SqlKind.COMPARISON.contains(e.getKind()) && !(e instanceof RexSubQuery)) {
```
rubenada, line 238: Would it be perhaps more clear and flexible to express the new condition in a more generic way? (just in case at some point in time this exceptional "1 operand comparison" is not limited to RexSubQuery), e.g.:
```
    if (SqlKind.COMPARISON.contains(e.getKind())) {
      // A comparison with a (non-null) literal, such as 'ref < 10', is not null if 'ref'
      // is not null.
      List<RexNode> operands = ((RexCall) e).getOperands();
      // We can have one operand in case e.g. of a RexSubQuery with IN operator.
      if (operands.size() > 1 && operands.get(1) instanceof RexLiteral
          && !((RexLiteral) operands.get(1)).isNull()) {
        return isEffectivelyNotNull(operands.get(0));
      }
    }
```

herunkang2018: This is more clear and can handle more illegal cases besides RexSubQuery, I will change it.

herunkang2018: @rubenada updated.

rubenada:
Thanks @herunkang2018  LGTM.
Could you please squash commits into a single one? I'll proceed with the merge afterwards if no further comment arrives.

herunkang2018:
@rubenada I have squashed commits. Thanks a lot for reviewing.

PR #3199, opened 2023-05-12
[CALCITE-5698] EXTRACT from INTERVAL partially does not follow the SQL standard

zstan:


babel/src/test/resources/sql/big-query.iq
```diff
@@ -371,6 +371,76 @@ SELECT EXTRACT(DAY FROM DATE '2013-12-25') AS the_day;
 
 !ok
 
+SELECT EXTRACT(MONTH FROM INTERVAL -1 MONTHS) as extr;
```
zoudan, line 374: make sure your first commit message(the CR title) is begin with [`${jira_ID}`]

zstan: @zoudan thanks, all done i hope.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -2890,12 +2891,13 @@ private static class ExtractImplementor extends AbstractRexCallImplementor {
     }
   }
 
-  private static Expression mod(Expression operand, long factor) {
+  private static Expression mod(Expression operand, long factor, boolean floorMod) {
     if (factor == 1L) {
       return operand;
     } else {
-      return Expressions.call(BuiltInMethod.FLOOR_MOD.method,
```
julianhyde, line 2897: convert '?' to 'else if' to be consistent with existing method

also revert your change removing the line-break; short lines are better

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -2890,12 +2891,13 @@ private static class ExtractImplementor extends AbstractRexCallImplementor {
     }
   }
 
-  private static Expression mod(Expression operand, long factor) {
```
julianhyde, line 2893: add javadoc explaining the purpose of the `floorMod` parameter. (If true, negative arguments will yield a negative result.)

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
zstan: thanks, done.

PR #3201, opened 2023-05-14
[CALCITE-5626] Sub-query with fully qualified table name throws table not found exception in validation phase

herunkang2018:


core/src/main/java/org/apache/calcite/sql/validate/IdentifierNamespace.java
```diff
@@ -258,6 +258,9 @@ public SqlIdentifier getId() {
   }
 
   @Override public @Nullable SqlValidatorTable getTable() {
+    if (resolvedNamespace == null) {
+      validator.validateNamespace(this, validator.unknownType);
+    }
     return resolvedNamespace == null ? null : resolve().getTable();
```
JiajunBernoulli, line 264: Above processed `resolvedNamespace == null`, 
do we still `resolvedNamespace == null ?` here?

julianhyde: I had the same thought.  And when I removed the `resolvedNamespace == null` check, 4 tests in `SqlAdvisorTest` broke. 

We seem to be violating the efforts of `AbstractNamespace.validate` to ensure that a particular namespace aren't under validation more than once, and that after validation has finished, `IdentifierNamespace.resolvedNamespace` is correctly set.

herunkang2018: Thanks for pointing out, I'll check it

herunkang2018: The reason why test cases failed in `SqlAdvisorTest` is that, there is table not found in catalog, so `resolvedNamespace` is still null, for example:
> select a.empno, b.deptno from ^dummy a, sales.dummy b

The table `dummy` is not found in catalog, so when validate `IdentifierNamespace` of `dummy`, it throws validation error in `IdentifierNamespace.resolveImpl`, but outter `SqlAdvisorValidator.validateWhereClause` catches it and **ignores** it, note that this is different from the behaviour of `SqlValidatorImpl`, which is just throws exception here. So in this case, the program continues executing and `resolvedNamespace` is still null.

herunkang2018: In thie PR, `resolvedNamespace == null` check seems not good enough, I want a clearer way to indicate that the namespace has been validated, I check that `resolvedNamespace` and `rowType` should be set after validation, but in `SqlAdvisorValidator`, these two variables remains null in unresolved table's namespace, which is hard to check whether namespace has been validated or not.

herunkang2018: Due to SqlAdvisorValidator's special logic with table not found, I think keep the current logic is also reasonable. @julianhyde @JiajunBernoulli looking forward to hearing your opinions.

julianhyde: There are 3 states to validation - not validated, under validation, and validated. `AbstractNamespace.validate` and `SqlNamespace.validateImpl` work in a rather subtle way. Your change has broken that, I think.

julianhyde: I acknowledge the problems with `SqlAdvisorValidator`.

However.

The logic in your current fix is so subtle that a future maintainer will not understand it.

You need to find a way to make the logic self-evidently correct.

herunkang2018: In current fix, the call path is `validator.validateNamespace` -> `namespace.validate` -> `AbstractNamespace.validate`, so the validation status has been protected yet. I find this is the same way used in `AbstractNamespace.getRowType`, so I don't understand your point very clear.

herunkang2018: @julianhyde would you like to give some advice?

PR #3203, opened 2023-05-16
[CALCITE-5703] Reduce amount of generated runtime code

zstan:


linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
```diff
@@ -1507,9 +1507,6 @@ public static TernaryExpression makeTernary(ExpressionType ternaryType,
     case Conditional:
       if (e1 instanceof ConstantUntypedNull) {
         type = Primitive.box(e2.getType());
-        if (e1.getType() != type) {
```
rubenada, line 1510: Why is this "if" block removed, but not the one just below (lines 1515..1517), which is the same but with `e2` instead of `e1`? Could that one be removed too?

zstan: thanks, seems you are right, i change it, will wait new test run completed.

DonnyZone:
A nice finding to reduce the generated code. From my perspective, the change in `RexImpTable` loss the internal state information during code generation phase, probably leading to some potential risks. Optimizing the code in **linq4j** (`ExpressionWriter` & `DeclarationStatement` & `TernaryExpression`) may be a better way. How about your opinion?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -1164,10 +1164,9 @@ public Expression handle(Expression x) {
 
   static Expression getDefaultValue(Type type) {
     Primitive p = Primitive.of(type);
-    if (p != null) {
-      return Expressions.constant(p.defaultValue, type);
-    }
-    return Expressions.constant(null, type);
+    return p != null
+        ? Expressions.constant(p.defaultValue, type)
+        : Expressions.constant(null);
```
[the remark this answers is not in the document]
zstan: thanks, why we need type info for null ? BTW all tests are green, probably you can suggest additional test for your assumption ?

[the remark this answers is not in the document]
rubenada: @DonnyZone , you mean e.g. if the expression is used in an overloaded method that requires a null parameter to be cast to avoid ambiguity?

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
zstan: I force push suggested change and it`s all ok from CI side.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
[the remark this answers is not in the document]
DonnyZone: @rubenada @DonnyZone Yes, just concern, but I can't come up with a UT to trigger this code path right now. This function is also called by some complex WinAggs. I'm personally incline to perform such common optimizations at the Linq4J side when generating the **code string**, rather than dropping type information of expressions.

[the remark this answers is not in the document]
DonnyZone: On the other hand, this PR changes two static functions which can be called by users. But the functions are not only limited to the scope of this ticket. For example, the original `Expressions.makeTernary` is intended to infer the untyped null, but this PR breaks this behavior.

DonnyZone:
We can conduct such optimizations in Optimizer. FYI:(https://github.com/apache/calcite/blob/b3e16dffa84b904f92413f8e3503fc23dc597165/linq4j/src/test/java/org/apache/calcite/linq4j/test/OptimizerTest.java#L158)

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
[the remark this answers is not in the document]
zstan: @DonnyZone ok i see, i will prepare new approach soon, thanks.

zstan:
@DonnyZone @rubenada new approach is force pushed, plz take a look?

DonnyZone (commented):
The `accept` function is responsible for generating code string with `ExpressionWriter`. Instead of rewriting the fields in `BinaryExpression` and `TernaryExpression`, we only need to rewrite the `expression1` and `expression2` to narrow the scope.
```
   ...
   // Comments ...
   if (..) {}
    new_expression0.accept(writer, lprec, nodeType.lprec);
    writer.append(nodeType.op);
    new_expression1.accept(writer, nodeType.rprec, nodeType.lprec);
    writer.append(nodeType.op2);
    new_expression2.accept(writer, nodeType.rprec, rprec);
```

linq4j/src/test/java/org/apache/calcite/linq4j/test/OptimizerTest.java
```diff
@@ -165,6 +165,17 @@ class OptimizerTest {
             NULL)));
   }
 
+  @Test void testOptimizeTernaryNullCasting() {
```
DonnyZone, line 168: How about the `v? "a": null`?

snuyanzin:
As mentioned in jira
removal of such cast could lead to Janino code compiliation problems like https://github.com/janino-compiler/janino/issues/188
So I would suggest to wait for Janino fix for that first

zstan:
@DonnyZone 3rd reincarnation ) is ready for review, 

@snuyanzin if i understand also if PR will be accepted we can`t merge it, do i need to fill the issue for bumping janino version in calcite or you can do it yourself ?

snuyanzin:
I remember a discussion about that and it was a desicion to not create jira issues just for version updates...
Anyway since I'm involved in the mentioned Janino's issue I could make an update once a release with fix is available (and passes all the tests like Calcite and Flink) and will notify here about that

zstan:
@DonnyZone thanks for review, but i still not hear any contradictions against this PR can be merged. If it so - is it possible to merge it ?

DonnyZone:
> @DonnyZone thanks for review, but i still not hear any contradictions against this PR can be merged. If it so - is it possible to merge it ?

Sorry for the late reply. I make some tests in my local environment. The optimization for `BinaryExpression` seems to be incorrect. 
The code after optimization throws compilation error.
```
        Boolean a = true || null;
        Boolean b = null || true;
```
While the code below works well.
```
        Boolean c = true || (Boolean) null;
        Boolean d = false || (Boolean) null;
        Boolean e = (Boolean) null && true;
```

zstan:
> Sorry for the late reply. I make some tests in my local environment. The optimization for `BinaryExpression` seems to be incorrect. The code after optimization throws compilation error.
> 
> ```
>         Boolean a = true || null;
>         Boolean b = null || true;
> ```

Ok i understand, but how can we build such expressions ?
```
    final ParameterExpression x_ = Expressions.parameter(Boolean.class, "x");
    BinaryExpression y1 = Expressions.orElse(Expressions.constant(false), Expressions.constant(null));

    BinaryExpression exp0 = Expressions.assign(x_, y1);
```
will produce code you mention, but it hand made code i have no ideas how it could be generated after optimization phase (OptimizeShuttle) . If optimization will be called it returns:
return x = null; Or return x = true; In constant(true) case.

DonnyZone:
> > Sorry for the late reply. I make some tests in my local environment. The optimization for `BinaryExpression` seems to be incorrect. The code after optimization throws compilation error.
> > ```
> >         Boolean a = true || null;
> >         Boolean b = null || true;
> > ```
> 
> @DonnyZone thanks for reply
> 
> I can`t obtain such a code after optimization ( can you show me how can i make it ?
> 
> final ParameterExpression x_ = Expressions.parameter(Object.class, "x"); BinaryExpression y1 = Expressions.orElse(Expressions.constant(true), Expressions.constant(null)); DeclarationStatement exp0 = Expressions.declare(Modifier.PUBLIC, x_, y1);
> 
> assertEquals("{\n public Object x = true;\n}\n", optimize(exp0))
> 
> after optimization shows: public Object x = true;

I just pick the code from your UT.
```
  @Test void testOptimizeBinaryNullCasting() {
    // (v ? null : "a") == 1
    assertEquals("{\n  return (v || null) == 1;\n}\n",
        optimize(
            Expressions.equal(
                Expressions.makeBinary(ExpressionType.OrElse,
                    Expressions.parameter(boolean.class, "v"),
                    new ConstantExpression(String.class, null)),
                ONE)));
  }
```
By the way, is it necessary to apply `skipNullCast` for the `BinaryExpression`?

zstan:
@DonnyZone got it, i refactor a code according to your suggestions and change the tests.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/OptimizeShuttle.java
```diff
@@ -164,6 +169,9 @@ && eq(cmp.expression1, expression2)) {
       Expression expression1) {
     //
     Expression result;
+    if (expression0 instanceof ParameterExpression) {
```
DonnyZone, line 172: Could you elaborate on why we need this optimization here? I just removed them and find `testOptimizeBinaryNullCasting` can still pass

zstan: this is not a "nitpicking" this is my fail, append additional test. thanks !

DonnyZone, line 172: I get your point now. The pattern to optimize is `Integer x = (Integer) null`. But it is not safe to check the `ParameterExpression` here. For example (`x || null `compilation error):
```
@Test void testOptimizeBinaryNullCasting2() {
    // Boolean x;
    ParameterExpression x = Expressions.variable(Boolean.class, "x");
    ParameterExpression y = Expressions.variable(Boolean.class, "y");
    // x = Boolean.valueOf(true)
    BinaryExpression xt = Expressions.assign(x,
        new ConstantExpression(Boolean.class, true));
    System.out.println(xt);
    // Boolean y = x || (Boolean) null;
    BinaryExpression yt = Expressions.assign(y,
        Expressions.orElse(x,
            new ConstantExpression(Boolean.class, null)));
    assert(optimize(yt).contains("x || (Boolean) null"));
}
```
I suggest to check the node type.
```
    if (binary.getNodeType() == ExpressionType.Assign) {
      expression1 = skipNullCast(expression1);
    }
```

DonnyZone:
Thanks for the work. Please address the final comment and squash the commits. I will merge it.

zstan:
@DonnyZone thanks for suggestions, all done.

DonnyZone (approved):
LGTM

PR #3206, opened 2023-05-16
Exploring ORDER BY ordinals fail to regenerate in cases with Numeric Literals

jhugomoore:


core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1119,6 +1119,34 @@ public List<SqlNode> fieldList() {
       };
     }
 
+    /**
+     * Checks if a generated ORDER BY node is a literal value.
+     * @param node to check
+     * @return true if node is representing a literal value, false otherwise.
+     */
+    boolean isLiteral(SqlNode node) {
```
julianhyde, line 1127: method should be private, or name of method should indicate that tests whether a node is a literal for purposes of the ORDER BY clause

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1119,6 +1119,34 @@ public List<SqlNode> fieldList() {
...
+      // In dialects that use ordinals, NumericLiterals are ordinals and should be kept.
+      if (dialect.getConformance().isSortByOrdinal() && node instanceof SqlNumericLiteral) {
+        return false;
+      }
+      if (node instanceof SqlLiteral) {
+        return true;
+      } else {
+        // The literal value may be wrapped in SqlBasicCalls applying postfix operators.
+        while (node instanceof SqlBasicCall) {
```
julianhyde, line 1136: testing for SqlBasicCall is too broad. `isLiteral('a' || x)` would currently return true, and that's wrong.

I think you're only interested in stripping away DESC, NULLS FIRST, NULLS LAST. Is that true?

If that is true, then you might do better checking for literal in the `RelFieldCollation` (i.e. before you call `toSql(field)`)

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1838,11 +1838,11 @@ private SqlDialect nonOrdinalDialect() {
   }
 
   /** Test case for
-   * <a href="https://issues.apache.org/jira/browse/CALCITE-5044">[CALCITE-5044]
-   * JDBC adapter generates integer literal in ORDER BY, which some dialects
-   * wrongly interpret as a reference to a field</a>. */
```
julianhyde, line 1843: You seem to have stomped on a unit test for CALCITE-5044. Does that mean that you think that the functionality added by CALCITE-5044 is wrong, and should be reverted? 

If so, you should reference CALCITE-5044 in bug CALCITE-5724 and explain why they are in conflict.

(My gut is that they are not in conflict, and you should keep the existing test method and add a new one.)

piglet/src/main/java/org/apache/calcite/piglet/PigRelToSqlConverter.java
```diff
@@ -102,7 +102,8 @@ public class PigRelToSqlConverter extends RelToSqlConverter {
 
       final List<SqlNode> orderList = Expressions.list();
       for (RelFieldCollation orderKey : winGroup.collation().getFieldCollations()) {
-        orderList.add(builder.context.toSql(orderKey));
+        builder.addOrderItem(orderList, orderKey);
+
```
julianhyde, line 106: remove this blank line

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1854,19 +1854,25 @@ private SqlDialect nonOrdinalDialect() {
...
-            + "ORDER BY '1', '23', '12', \"ENAME\", '34' DESC NULLS LAST")
-        .dialect(nonOrdinalDialect())
-        .ok("SELECT JOB, ENAME\n"
-            + "FROM scott.EMP\n"
-            + "ORDER BY 1, '23', 12, ENAME, 34 DESC NULLS LAST");
+            + "ORDER BY \"ENAME\"");
+    relFn(relFn2)
+        .ok("SELECT \"JOB\", \"ENAME\"\n"
+            + "FROM \"scott\".\"EMP\"");
+
```
julianhyde, line 1875: remove this blank line

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1854,19 +1854,25 @@ private SqlDialect nonOrdinalDialect() {
...
   }
```
julianhyde, line 1877: add a test case where the ORDER BY clause has one literal, that literal is skipped by your code, and the generated query has no ORDER BY clause

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
julianhyde: Maybe they can coexist. Add a method `SqlConformance.isSortByLiteralAllowed()`, default true, false in BigQuery. The test would have the old behavior if `isSortByLiteralAllowed` is true, the new behavior if it is false.

julianhyde (changes requested):

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
jhugomoore: Such a good call, in testing it looks like postfix operators aren't applied to the node created from field(fieldCollation.getFieldIndex()), so testing for literals directly is possible.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
jhugomoore: Applied, let me know if its along the lines of what you had in mind

core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
```diff
@@ -303,6 +303,17 @@ public interface SqlConformance {
    */
   boolean isInsertSubsetColumnsAllowed();
 
+  /**
+   * Whether the dialect is able to ORDER BY literal values.
```
julianhyde, line 307: 2nd ('This flag') and 3rd ('Returns false') paragraphs need to start with `<p>`.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1120,6 +1120,11 @@ public List<SqlNode> fieldList() {
     }
 
     void addOrderItem(List<SqlNode> orderByList, RelFieldCollation field) {
+      // If the field being ordered on is a Literal value, we can safely skip it.
```
julianhyde, line 1123: does this work if the field is an aliased literal, e.g. `date '2000-01-01' as d`? I'd add that as a scenario to the test.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1877,6 +1885,19 @@ private SqlDialect nonOrdinalDialect() {
...
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5724">[CALCITE-5724]
+   * Test the ORDER BY is properly removed when all literal nodes are filtered out
+   * in a dialect that doesn't support ORDER BY literals</a>.
+   */
+  @Test void testRemoveEmptyOrderByClause() {
+    String query = "select 5, \"product_id\" from \"product\"\n"
+        + "order by 1";
+    final String expected = "SELECT 5, product_id\n"
+        + "FROM foodmart.product";
+    sql(query).withBigQuery().ok(expected);
```
julianhyde, line 1898: i'd chain `.withMySql().ok(expectedMySql)` just so we see what happens in the 'normal' case.

core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
```diff
@@ -303,6 +303,17 @@ public interface SqlConformance {
...
+   *
+   * This flag is used to filter out literal values in the
+   * order by clause of dialects that don't support them.
+   *
+   * Returns false if the dialect errors when a literal
+   * value is present in ORDER BY, true otherwise.
+   */
```
julianhyde, line 314: Add a 'Default is false; true in BigQuery' paragraph similar to the other methods in this class.

julianhyde (changes requested):

core/src/main/java/org/apache/calcite/sql/validate/SqlConformanceEnum.java
```diff
@@ -397,6 +397,15 @@ public enum SqlConformanceEnum implements SqlConformance {
     }
   }
 
+  @Override public boolean isSortByLiteralAllowed() {
+    switch (this) {
+    case BIG_QUERY:
+      return false;
```
tanclary, line 403: is it easy to check if other dialects allow this?

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1133,7 +1138,8 @@ void addOrderItem(List<SqlNode> orderByList, RelFieldCollation field) {
                   RelFieldCollation.NullDirection.UNSPECIFIED);
         }
       }
-      orderByList.add(toSql(field));
+      SqlNode orderByNode = toSql(field);
```
tanclary, line 1141: is this expanded just for readability

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1120,6 +1120,11 @@ public List<SqlNode> fieldList() {
     }
 
     void addOrderItem(List<SqlNode> orderByList, RelFieldCollation field) {
+      // If the field being ordered on is a Literal value, we can safely skip it.
+      if (!dialect.getConformance().isSortByLiteralAllowed()
+          && field(field.getFieldIndex()) instanceof SqlLiteral) {
```
tanclary, line 1125: just a question I don't know the answer to: which ordering of these checks is more "right"? would you rather check if sortByLiteral is allowed and then see if it's a literal? or does it make more sense to see if it's a literal and then check if it's allowed? 🤷

jhugomoore: I think especially because isSortByLiteralAllowed() is default true for all non-bq languages it's probably better to check in this order. Also not totally sure though.

core/src/main/java/org/apache/calcite/sql/validate/SqlConformanceEnum.java
jhugomoore: I'm not aware of a way to test this one a wide array of dialects, but if someone has a good resource for that I'd be happy to

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
jhugomoore: Think it's a lingering artifact of previous debugging lol, i'll get that changed up

julianhyde:
CALCITE-5529 would make it possible to change dialects optimistically (i.e. assuming that it's OK to sort by a literal), run the test on all dialects, and see which ones the SQL fails. But CALCITE-5529 is a long way from done.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1122,6 +1122,11 @@ public List<SqlNode> fieldList() {
...
+        return;
+      }
```
wnob, line 1129: See my comment on the JIRA case. I think a better solution will live in `SqlImplementor.Context.orderField`.

PR #3207, opened 2023-05-17
[CALCITE-5707] Add ARRAY_CONTAINS function (enabled in Spark library)

liuyongvs:
ARRAY_CONTAINS - Returns true if the array contains the value

For more details
https://spark.apache.org/docs/latest/sql-ref-functions-builtin.html

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -3882,8 +3882,7 @@ public class SqlFunctions {
 
   /** Support the ARRAY_CONTAINS function. */
   public static boolean contains(List list, Object element) {
-    final Set set = new HashSet(list);
-    return set.contains(element);
+    return list.contains(element);
   }
 
   /** Support the ARRAY_DISTINCT function. */
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -3880,6 +3880,12 @@ public class SqlFunctions {
     return result;
   }
 
+  /** Support the ARRAY_CONTAINS function. */
+  public static boolean contains(List list, Object element) {
+    final Set set = new HashSet(list);
+    return set.contains(element);
+  }
+
   /** Support the ARRAY_DISTINCT function. */
   public static List distinct(List list) {
     Set result = new LinkedHashSet<>(list);
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -3881,7 +3881,7 @@ public class SqlFunctions {
   }
 
   /** Support the ARRAY_CONTAINS function. */
-  public static boolean contains(List list, Object element) {
+  public static boolean arrayContains(List list, Object element) {
     return list.contains(element);
   }
 
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -886,7 +886,14 @@ public abstract class SqlLibraryOperators {
           ReturnTypes.LEAST_RESTRICTIVE,
           OperandTypes.AT_LEAST_ONE_SAME_VARIADIC);
 
-  /** The "ARRAY_DISTINCT(array)" function (Spark). */
+  /** The "ARRAY_CONTAINS(array, element)" function. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction ARRAY_CONTAINS =
+      SqlBasicFunction.create(SqlKind.ARRAY_CONTAINS,
+          ReturnTypes.BOOLEAN_NULLABLE_IF_ARG0_NULLABLE,
+          OperandTypes.ARRAY_ELEMENT);
+  
+  /** The "ARRAY_DISTINCT(array)" function. */
   @LibraryOperator(libraries = {SPARK})
   public static final SqlFunction ARRAY_DISTINCT =
       SqlBasicFunction.create(SqlKind.ARRAY_DISTINCT,
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -890,9 +890,9 @@ public abstract class SqlLibraryOperators {
   @LibraryOperator(libraries = {SPARK})
   public static final SqlFunction ARRAY_CONTAINS =
       SqlBasicFunction.create(SqlKind.ARRAY_CONTAINS,
-          ReturnTypes.BOOLEAN_NULLABLE_IF_ARG0_NULLABLE,
+          ReturnTypes.BOOLEAN_NULLABLE,
           OperandTypes.ARRAY_ELEMENT);
-  
+
   /** The "ARRAY_DISTINCT(array)" function. */
   @LibraryOperator(libraries = {SPARK})
   public static final SqlFunction ARRAY_DISTINCT =
--- a/core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
@@ -1,0 +1,75 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.sql.type;
+
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.sql.SqlCallBinding;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlOperandCountRange;
+import org.apache.calcite.sql.SqlOperator;
+
+import com.google.common.collect.ImmutableList;
+
+import static org.apache.calcite.sql.type.NonNullableAccessors.getComponentTypeOrThrow;
+import static org.apache.calcite.util.Static.RESOURCE;
+
+/**
+ * Parameter type-checking strategy where types must be Array and Array element type.
+ */
+public class ArrayElementOperandTypeChecker implements SqlOperandTypeChecker {
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public boolean checkOperandTypes(
+      SqlCallBinding callBinding,
+      boolean throwOnFailure) {
+    final SqlNode op0 = callBinding.operand(0);
+    if (!OperandTypes.ARRAY.checkSingleOperandType(
+        callBinding,
+        op0,
+        0,
+        throwOnFailure)) {
+      return false;
+    }
+
+    RelDataType arrayComponentType =
+        getComponentTypeOrThrow(SqlTypeUtil.deriveType(callBinding, op0));
+    final SqlNode op1 = callBinding.operand(1);
+    RelDataType aryType1 = SqlTypeUtil.deriveType(callBinding, op1);
+
+    RelDataType biggest =
+        callBinding.getTypeFactory().leastRestrictive(
+            ImmutableList.of(arrayComponentType, aryType1));
+    if (null == biggest) {
+      if (throwOnFailure) {
+        throw callBinding.newError(
+            RESOURCE.typeNotComparable(
+                arrayComponentType.toString(), aryType1.toString()));
+      }
+
+      return false;
+    }
+    return true;
+  }
+
+  @Override public SqlOperandCountRange getOperandCountRange() {
+    return SqlOperandCountRanges.of(2);
+  }
+
+  @Override public String getAllowedSignatures(SqlOperator op, String opName) {
+    return "<ARRAY> " + opName + " <ARRAY>";
+  }
+}
--- a/core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
@@ -53,7 +53,7 @@ public class ArrayElementOperandTypeChecker implements SqlOperandTypeChecker {
     RelDataType biggest =
         callBinding.getTypeFactory().leastRestrictive(
             ImmutableList.of(arrayComponentType, aryType1));
-    if (null == biggest) {
+    if (biggest == null) {
       if (throwOnFailure) {
         throw callBinding.newError(
             RESOURCE.typeNotComparable(
--- a/core/src/main/java/org/apache/calcite/sql/type/SqlTypeTransforms.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/SqlTypeTransforms.java
@@ -97,6 +97,21 @@ public abstract class SqlTypeTransforms {
       };
 
   /**
+   * Parameter type-inference transform strategy where a derived type is
+   * transformed into the same type, but nullable if and only if the type
+   * of a call's operand #0 (0-based) is nullable.
+   */
+  public static final SqlTypeTransform ARG0_NULLABLE =
+      (opBinding, typeToTransform) -> {
+        RelDataType arg0 = opBinding.collectOperandTypes().get(0);
+        if (arg0.isNullable()) {
+          return opBinding.getTypeFactory()
+              .createTypeWithNullability(typeToTransform, true);
+        }
+        return typeToTransform;
+      };
+
+  /**
    * Type-inference strategy whereby the result type of a call is VARYING the
    * type given. The length returned is the same as length of the first
    * argument. Return type will have same nullability as input type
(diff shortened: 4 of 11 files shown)


liuyongvs:
hi @JiajunBernoulli would you also help review this?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3791,6 +3791,12 @@ private static AtomicLong getAtomicLong(String key) {
     return atomic;
   }
 
+  /** Support the ARRAY_CONTAINS function. */
+  public static boolean contains(List list, Object element) {
+    final Set set = new HashSet(list);
```
JiajunBernoulli, line 3796: Is this for faster search?

core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
```diff
@@ -0,0 +1,75 @@
...
+        op0,
+        0,
+        throwOnFailure)) {
+      return false;
+    }
+
+    RelDataType arrayComponentType =
+        getComponentTypeOrThrow(SqlTypeUtil.deriveType(callBinding, op0));
+    final SqlNode op1 = callBinding.operand(1);
+    RelDataType aryType1 = SqlTypeUtil.deriveType(callBinding, op1);
```
JiajunBernoulli, line 51: Why not use `getComponentTypeOrThrow`?

core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
```diff
@@ -0,0 +1,75 @@
...
+
+import com.google.common.collect.ImmutableList;
+
+import static org.apache.calcite.sql.type.NonNullableAccessors.getComponentTypeOrThrow;
+import static org.apache.calcite.util.Static.RESOURCE;
+
+/**
+ * Parameter type-checking strategy where types must be Array and Array element type.
+ */
+public class ArrayElementOperandTypeChecker implements SqlOperandTypeChecker {
```
JiajunBernoulli, line 33: Many codes are same as `MultisetOperandTypeChecker`, Can we extract them?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -844,7 +844,14 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
           SqlLibraryOperators::arrayReturnType,
           OperandTypes.SAME_VARIADIC);
 
-  /** The "ARRAY_DISTINCT(array)" function (Spark). */
+  /** The "ARRAY_CONTAINS(array)" function. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction ARRAY_CONTAINS =
+      SqlBasicFunction.create(SqlKind.ARRAY_CONTAINS,
+          ReturnTypes.BOOLEAN_NULLABLE_IF_ARG0_NULLABLE,
```
JiajunBernoulli, line 851: I don't understand why is `ARG0`.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
liuyongvs: yeap, it is no different with for loop, may cost some space. if you think it is need, i will change it to for loop

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
liuyongvs: because 
1) array it null, result is null
2) element is null, it is not . for example array_contains(array[null, 1] , null) -> true . type is not null

flink implementation 
https://github.com/apache/flink/blob/06688f345f6793a8964ec00002175f44cda13c33/flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/BuiltInFunctionDefinitions.java#L195
```
public static final BuiltInFunctionDefinition ARRAY_CONTAINS =
            BuiltInFunctionDefinition.newBuilder()
                    .name("ARRAY_CONTAINS")
                    .kind(SCALAR)
                    .inputTypeStrategy(
                            sequence(
                                    Arrays.asList("haystack", "needle"),
                                    Arrays.asList(
                                            logical(LogicalTypeRoot.ARRAY), ARRAY_ELEMENT_ARG)))
                    .outputTypeStrategy(
                            nullableIfArgs(
                                    ConstantArgumentCount.of(0), explicit(DataTypes.BOOLEAN())))
                    .runtimeClass(
                            "org.apache.flink.table.runtime.functions.scalar.ArrayContainsFunction")
                    .build();

```

core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
liuyongvs: i am using it  RelDataType arrayComponentType =
        getComponentTypeOrThrow(SqlTypeUtil.deriveType(callBinding, op0));

core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
liuyongvs: yes refer it.
do not find a good idea to abstract it.
MultisetOperandTypeChecker for two Multiset
ArrayElementOperandTypeChecker for Array and element type

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
JiajunBernoulli: In fact, set has not been reused, and I don't think it can improve speed.

Because the construction time of the set may be longer than the search time of the list.

core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
JiajunBernoulli: Why `op1` not need?
```    
RelDataType biggest =
        callBinding.getTypeFactory().leastRestrictive(
            ImmutableList.of(
                getComponentTypeOrThrow(SqlTypeUtil.deriveType(callBinding, op0)),
                getComponentTypeOrThrow(SqlTypeUtil.deriveType(callBinding, op1))));
```
Here is code in MultisetOperandTypeChecker.
https://github.com/apache/calcite/blob/3e95a13b8c5d715298f469051f792286cea704fa/core/src/main/java/org/apache/calcite/sql/type/MultisetOperandTypeChecker.java#L66

liuyongvs: op1 it not array type, it is element type. for example array_contains(a[1, 2, 3], 1) 

and why we should  find biggest type , for exmplae array_contains(a[1, 2, 3], 'a')
array type is integer not null array not null -> element type is integer not null
arg1 type is char[1] not null, we can not find  biggest type , so it will throw exception
```
    RelDataType biggest =
        callBinding.getTypeFactory().leastRestrictive(
            ImmutableList.of(arrayComponentType, aryType1));
```

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
liuyongvs: yeap, fixed @JiajunBernoulli

liuyongvs:
hi @JiajunBernoulli   fix conflict and all your reviews, and thanks for your review so much

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
julianhyde: I don't think you need this method at all. The code generator can just call `java.util.List.contains`. You will need to add it to `BuiltInMethod`.

liuyongvs: @julianhyde I have considered this， but it can't. because the array_contains have 2 args instead of 1 in java.util.List.contains

liuyongvs:
when you review in weekend, you can review it first because other functions depends it(type ARRAY_ELEMENT_ARG), such as array_position/array_remove/array_append/array_prepend and so on

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5324,6 +5324,27 @@ private static void checkIf(SqlOperatorFixture f) {
...
+        "BOOLEAN NOT NULL");
+    f.checkScalar("array_contains(array[1, null], cast(null as integer))", true,
+        "BOOLEAN NOT NULL");
+    f.checkScalar("array_contains(array[map[1, 'a'], map[2, 'b']], map[1, 'a'])", true,
+        "BOOLEAN NOT NULL");
+    f.checkFails("^array_contains(array[1, 2], true)^",
+        "INTEGER is not comparable to BOOLEAN", false);
+    f.checkNull("array_contains(cast(null as integer array), 1)");
+    f.checkType("array_contains(cast(null as integer array), 1)", "BOOLEAN");
+  }
```
MasseGuillaume, line 5346:
```suggestion
      f.checkScalar("array_contains(array[1], 1)", true,
        "BOOLEAN NOT NULL");
      f.checkScalar("array_contains(array(), 1)", false,
        "BOOLEAN NOT NULL");
  }
```

liuyongvs: @MasseGuillaume added

liuyongvs:
fix conflicts and all your reviews @MasseGuillaume @JiajunBernoulli

liuyongvs:
hi @JiajunBernoulli could you review it because other functions depends it(type ARRAY_ELEMENT_ARG), such as array_position/array_remove/array_append/array_prepend and so on, i will support them in one pr

core/src/main/java/org/apache/calcite/sql/type/SqlTypeTransforms.java
```diff
@@ -96,6 +96,21 @@ public abstract class SqlTypeTransforms {
...
       };
 
+  /**
+   * Parameter type-inference transform strategy where a derived type is
+   * transformed into the same type, but nullable if and only if the type
+   * of a call's operand #0 (0-based) is nullable.
+   */
+  public static final SqlTypeTransform ARG0_NULLABLE =
+      (opBinding, typeToTransform) -> {
+        RelDataType arg0 = opBinding.collectOperandTypes().get(0);
```
tanclary, line 106: similar comment to what i left on your other PR opBinding.getOperandType()

site/_docs/reference.md
```diff
@@ -2654,6 +2654,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | s | ARRAY(expr [, expr ]*)                         | Construct an array in Apache Spark
 | s | ARRAY_COMPACT(array)                           | Removes null values from the *array*
 | b | ARRAY_CONCAT(array [, array ]*)                | Concatenates one or more arrays. If any input argument is `NULL` the function returns `NULL`
+| s | ARRAY_CONTAINS(array, element)                 | Returns true if the *array* contains the value
```
tanclary, line 2657: description arguments should match function arguments (replace value with element or vice versa)

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5364,6 +5364,31 @@ private static void checkIf(SqlOperatorFixture f) {
...
 
+  /** Tests {@code ARRAY_CONTAINS} function from Spark. */
+  @Test void testArrayContainsFunc() {
+    final SqlOperatorFixture f0 = fixture();
+    f0.setFor(SqlLibraryOperators.ARRAY_CONTAINS);
+    f0.checkFails("^array_contains(array[1, 2], 1)^",
+        "No match found for function signature "
+            + "ARRAY_CONTAINS\\(<INTEGER ARRAY>, <NUMERIC>\\)", false);
+
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
```
tanclary, line 5375: again curious what the behavior is/should be if you search an array of type X for a value of type Y, obviously it would return false but should it be allowed in the first place?

julianhyde: Good point, @tanclary. The validator should give an error if you - say - search for a BOOLEAN in a DATE ARRAY. We should add a test case to this test method.

liuyongvs: hi @julianhyde @tanclary , there is the unit test in the end
    f.checkFails("^array_contains(array[1, 2], true)^",
        "INTEGER is not comparable to BOOLEAN", false);

core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
```diff
@@ -0,0 +1,75 @@
...
+
+    RelDataType arrayComponentType =
+        getComponentTypeOrThrow(SqlTypeUtil.deriveType(callBinding, op0));
+    final SqlNode op1 = callBinding.operand(1);
+    RelDataType aryType1 = SqlTypeUtil.deriveType(callBinding, op1);
+
+    RelDataType biggest =
+        callBinding.getTypeFactory().leastRestrictive(
+            ImmutableList.of(arrayComponentType, aryType1));
+    if (null == biggest) {
```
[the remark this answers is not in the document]
liuyongvs: done

site/_docs/reference.md
liuyongvs: fixed

core/src/main/java/org/apache/calcite/sql/type/SqlTypeTransforms.java
liuyongvs: fixed

liuyongvs:
hi @tanclary @JiajunBernoulli @julianhyde @MasseGuillaume fix all your reviews, do you have time to look again?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5364,6 +5364,31 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f.checkScalar("array_contains(array[1, 2], 1)", true,
+        "BOOLEAN NOT NULL");
+    f.checkScalar("array_contains(array[1, null], cast(null as integer))", true,
+        "BOOLEAN NOT NULL");
```
MasseGuillaume, line 5379: Do we want to type check exactly as Apache Spark does?

```
spark.sql("select array_contains(array(1, null), null)").show()
org.apache.spark.sql.AnalysisException: cannot resolve 'array_contains(array(1, CAST(NULL AS INT)), NULL)' due to data type mismatch: Null typed values cannot be used as arguments; line 1 pos 7;
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5364,6 +5364,31 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f.checkScalar("array_contains(array[1], 1)", true,
+        "BOOLEAN NOT NULL");
+    f.checkScalar("array_contains(array(), 1)", false,
+        "BOOLEAN NOT NULL");
+    f.checkScalar("array_contains(array[map[1, 'a'], map[2, 'b']], map[1, 'a'])", true,
+        "BOOLEAN NOT NULL");
```
MasseGuillaume, line 5385:
```
spark.sql("""select array_contains(array(map(1, "1"), map(2, "2")), map(2, "2"))""").show()
org.apache.spark.sql.AnalysisException: cannot resolve 'array_contains(array(map(1, '1'), map(2, '2')), map(2, '2'))' due to data type mismatch: function array_contains does not support ordering on type map<int,string>; line 1 pos 7;
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
liuyongvs: you should use array_contains(array[1, null], cast(null as integer) 
and for my side, spark the behavior is not good the second arg is null, also return null.
so i does flink way, in flink array_contains(array[1, null], cast(null as integer)  it return true, while in spark return null

public static final BuiltInFunctionDefinition ARRAY_CONTAINS =
            BuiltInFunctionDefinition.newBuilder()
                    .name("ARRAY_CONTAINS")
                    .kind(SCALAR)
                    .inputTypeStrategy(
                            sequence(
                                    Arrays.asList("haystack", "needle"),
                                    Arrays.asList(
                                            logical(LogicalTypeRoot.ARRAY), ARRAY_ELEMENT_ARG)))
                    .outputTypeStrategy(
                            nullableIfArgs(
                                    ConstantArgumentCount.of(0), explicit(DataTypes.BOOLEAN())))
                    .runtimeClass(
                            "org.apache.flink.table.runtime.functions.scalar.ArrayContainsFunction")
                    .build();

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
liuyongvs: Due to implementation limitation, currently Spark can't compare or do equality check between map types. As a result, map values can't appear in EQUAL or comparison expressions, can't be grouping key, etc.()
while calcite Map runtime implementation using java collection Map, which supports equality check. while spark not
```
/**
 * This is an internal data representation for map type in Spark SQL. This should not implement
 * `equals` and `hashCode` because the type cannot be used as join keys, grouping keys, or
 * in equality tests. See SPARK-9415 and PR#13847 for the discussions.
 */
abstract class MapData extends Serializable {
```


https://github.com/apache/spark/pull/23045

JiajunBernoulli:
@liuyongvs Please resolve conflict files.

liuyongvs:
hi @JiajunBernoulli @julianhyde fix conflicts and align with spark instead of flink's more reasonable behavior

julianhyde (commented):
@liuyongvs @tanclary @JiajunBernoulli @MasseGuillaume Thanks to all who reviewed. I think this is in good shape. Do you all agree? If so I'll merge.

I would change only two things:
* Remove SqlFunctions.arrayContains and use java.util.List.contains directly.
* Add some comments to the test about Spark vs Flink behavor.

JiajunBernoulli (approved):

liuyongvs:
@julianhyde thanks so much.  I will submit others quickly when it is merged because others in my local commit

PR #3208, opened 2023-05-17
[CALCITE-4555] Invalid zero literal value is used for TIMESTAMP WITH LOCAL TIME ZONE type in RexBuilder

snuyanzin:
The PR makes `org.apache.calcite.rex.RexBuilder#makeZeroLiteral` returning `RexLiteral` for `TIMESTAMP_WITH_LOCAL_TIME_ZONE` based on `TimestampString(0, 1, 1, 0, 0, 0)` since month and day number should be positive


zoudan:
LGTM

snuyanzin:
thanks for having a look @zoudan

PR #3209, opened 2023-05-17
[CALCITE-4334] LITERAL_AGG, an aggregate function that returns a constant value

julianhyde:
Branch contains fixes for [CALCITE-4334](https://issues.apache.org/jira/browse/CALCITE-4334) (LITERAL_AGG), [CALCITE-5697](https://issues.apache.org/jira/browse/CALCITE-5697) (a bug in `RelBuilder.convert`) and a few refactoring commits.

It is a step on the road to [CALCITE-4496](https://issues.apache.org/jira/browse/CALCITE-4496) (measures). See https://github.com/julianhyde/calcite/tree/4496-measure.


core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -1688,6 +1692,27 @@ static class GroupingImplementor implements AggImplementor {
...
+    }
+
+    @Override public void implementAdd(AggContext info, AggAddContext add) {
+    }
+
+    @Override public Expression implementResult(AggContext info,
+        AggResultContext result) {
+      checkArgument(info.aggregation().kind == SqlKind.LITERAL_AGG);
+      checkArgument(result.call().rexList.size() == 1);
+      final RexNode rexNode = result.call().rexList.get(0);
```
rubenada, line 1711: Should we check here that this rexNode is a RexLiteral?

julianhyde: I don't feel strongly about it. There's some benefit to retaining flexibility.

For example, in `Sort` we didn't validate that `limit` and `offset` were literals and it turned out to be useful - because people wanted to use parameters.

Clearly the expression should not contain `RexInputRef`. But calls to parameter-less functions such as `CURRENT_DATE` seem like reasonable use cases.

PR #3210, opened 2023-05-17
[CALCITE-5564] Add parsing and validation for PERCENTILE_CONT/PERCENTILE_DISC functions (enabled in BigQuery)

tanclary:
Calcite currently supports parsing and validation for the standard PERCENTILE_CONT and PERCENTILE_DISC functions. BigQuery has their own form of these functions with different syntax (further explained on [CALCITE-5564](https://issues.apache.org/jira/browse/CALCITE-5564). This PR allows for the BigQuery syntax to be parsed and validated as well. 

I tried to test these changes as thoroughly as I could given that the actual implementation has not be done yet (that is the task of [CALCITE-4666](https://issues.apache.org/jira/browse/CALCITE-4666).

Any comments, suggestions, or questions are welcome as always.

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
@@ -6536,6 +6536,64 @@ public class SqlValidatorTest extends SqlValidatorTestCase {
         .fails("'PERCENTILE_DISC' requires precisely one ORDER BY key");
   }
 
+  @Test void testPercentileFunctionsBigQuery() {
+    final SqlOperatorTable opTable = operatorTableFor(SqlLibrary.BIG_QUERY);
+    final String sql = "select\n"
+        + " percentile_cont(sal, 0.25) over() as c,\n"
+        + " percentile_disc(sal, 0.5) over() as d\n"
+        + "from emp";
+    sql(sql)
+        .withConformance(SqlConformanceEnum.BIG_QUERY)
+        .withOperatorTable(opTable)
+        .type("RecordType(DOUBLE NOT NULL C, INTEGER NOT NULL D) NOT NULL");
+  }
+
+  @Test void testPercentileContBigQueryFraction() {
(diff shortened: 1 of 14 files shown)


core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -6540,6 +6540,64 @@ void testGroupExpressionEquivalenceParams() {
         .fails("'PERCENTILE_DISC' requires precisely one ORDER BY key");
   }
 
+  @Test void testPercentileFunctionsBigQuery() {
+    final SqlOperatorTable opTable = operatorTableFor(SqlLibrary.BIG_QUERY);
```
julianhyde, line 6544: the examples over `unnest(array)` are appropriate for `big-query.iq` but are not typical of actual usage. for these tests can you switch to queries that use `emp`. Compute the percentile of sal of each employee within their department and within the whole org.

julianhyde (changes requested):
looks good. a few cosmetic changes required.

did you add tests to big-query.iq? If you can't execute the queries, at least you could validate them using `!type`.

If you need examples, you should use `emp` and `sal` and `deptno`. It's a lot more meaningful than `unnest(array ...)`.

tanclary:
> looks good. a few cosmetic changes required.
> 
> did you add tests to big-query.iq? If you can't execute the queries, at least you could validate them using `!type`.
> 
> If you need examples, you should use `emp` and `sal` and `deptno`. It's a lot more meaningful than `unnest(array ...)`.

Hey @julianhyde , I amended my commit to include most of your fixes. The only one I couldn't quite do was your suggestion to add tests to big-query.iq. Even just checking the type throws the following error: 

`Unable to get aggregate implementation for aggregate PERCENTILE_CONT in window context`

Of course this makes sense given that there is no aggregate implementation, but if there's a workaround for me to still include these tests please let me know.

Thanks for the review!

PR #3211, opened 2023-05-18
[CALCITE-5708] Change SUBSTRING result if either of parameters is NULL

zstan:


core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -1042,6 +1042,10 @@ void testDyadicCollateOperator() {
         .columnType("VARCHAR(1) NOT NULL");
     expr("substring('a', 1, '3')")
         .columnType("VARCHAR(1) NOT NULL");
+
+    expr("SUBSTRING(NULL FROM 1 FOR 2)").ok();
```
herunkang2018, line 1046: this test case is not effected by this issue, would you like to add a comment to clarify it?

core/src/test/resources/sql/functions.iq
```diff
@@ -464,4 +464,19 @@ limit 3;
 
 !ok
 
+# SUBSTRING
+-- returns 'null'
+select SUBSTRING(NULL FROM 1 FOR 2);
```
herunkang2018, line 469: can these test cases in quidem file be put together? I'm not sure if this works for later test cases

zstan: yes, they can.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
zstan: I just expand the tests coverage.

herunkang2018: I know it, if you add comment to clarify it, it would be better I think.

zstan: @herunkang2018 fixed, can you make a review plz ? I update description and starting code point in appropriate issue.

zstan:
@herunkang2018 can you make a review plz ? I fill additional description in issue.

herunkang2018:
@zstan I leave two comments in Jira, after that, I will review this PR again.

zstan:
@herunkang2018 thanks ! i check 
```
  @Test
  void testStrcmpFunc() {
    final SqlOperatorFixture f = fixture()
        .setFor(SqlLibraryOperators.STRCMP)
        .withLibrary(SqlLibrary.MYSQL);

    f.checkNull("STRCMP(null, null)");
```
and what kind of runtine it generates for null literals
`EnumerableCalc#implement`
and i check my PR change for SUBSTRING and as for me it looks the same (for now) i.e. did the same as from java doc :

```
If any of the other operands are nullable the
   * returned type will also be nullable. First Arg must be of string type.
   */
  public static final SqlReturnTypeInference ARG0_NULLABLE_VARYING =
      ARG0.andThen(SqlTypeTransforms.TO_NULLABLE)
          .andThen(SqlTypeTransforms.TO_VARYING);
```
i.e. for null literals appropriate runtime implementation will never been called, seems it correct.

zstan:
@herunkang2018 kindly remind help with review if possible, thanks !

core/src/main/java/org/apache/calcite/sql/type/SqlTypeFamily.java
```diff
@@ -244,7 +244,7 @@ public Collection<SqlTypeName> getTypeNames() {
     case EXACT_NUMERIC:
       return SqlTypeUtil.getMaxPrecisionScaleDecimal(factory);
     case INTEGER:
-      return factory.createSqlType(SqlTypeName.BIGINT);
+      return factory.createSqlType(SqlTypeName.INTEGER);
```
gianm, line 247: Why this change? I would think that `BIGINT` is a good default concrete type for the family `INTEGER`, since it's the least restrictive in the family. Seems like this may have unwanted effects, if it causes inference for the `INTEGER` family to use a too-narrow type.

core/src/test/resources/sql/functions.iq
```diff
@@ -938,5 +938,19 @@ table(AUX.TBLFUN_IDENTITY(3)) as t3(v);
 
 !ok
 
+# SUBSTRING
```
gianm, line 941: I suggest including a test for the 2-arg form too. Like:

```
select SUBSTRING('text' FROM NULL)
```

It should return null.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -1040,6 +1040,11 @@ void testDyadicCollateOperator() {
         .columnType("VARCHAR(1) NOT NULL");
     expr("substring('a', 1, '3')")
         .columnType("VARCHAR(1) NOT NULL");
+
+    // Correctly processed null string and params.
```
gianm, line 1044: I suggest including a test for the 2-arg form too. Like:

```
expr("SUBSTRING('text' FROM NULL)").ok();
```

zstan: done

core/src/test/resources/sql/functions.iq
zstan: done

core/src/main/java/org/apache/calcite/sql/type/SqlTypeFamily.java
zstan: yep, thanks, seems it redundant change

core/src/test/resources/sql/functions.iq
```diff
@@ -938,5 +938,20 @@ table(AUX.TBLFUN_IDENTITY(3)) as t3(v);
...
+-- returns 'null'
+select SUBSTRING(NULL FROM 1 FOR 2);
+select SUBSTRING('text' FROM 1 FOR NULL);
+select SUBSTRING('text' FROM NULL FOR 2);
+select SUBSTRING(s from i for l) from (values ('abc', null, 2)) as t (s, i, l);
```
herunkang2018, line 946: minor comment, better to unify the usage of uppercase and lowercase. You could follow the test cases above.

herunkang2018:
@gianm Thanks for your review, I will review again in 24h.

core/src/test/resources/sql/functions.iq
zstan: done

herunkang2018:
@zstan @gianm LGTM

herunkang2018 (approved):

zstan:
thanks a lot, i will squash all commits in final PR soon

zstan:
seems it ready for merging, thanks a lot

PR #3212, opened 2023-05-18
[CALCITE-5710] Add ARRAY_MAX, ARRAY_MIN function (enabled in Spark li…

liuyongvs:
ARRAY_MAX - Returns the maximum value in the array

ARRAY_MIN - Returns the minmum value in the array

 

For more details
https://spark.apache.org/docs/latest/api/sql/index.html

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -3886,6 +3886,34 @@ public class SqlFunctions {
     return new ArrayList<>(result);
   }
 
+  /** Support the ARRAY_MAX function. */
+  public static @Nullable <T extends Object & Comparable<? super T>> T arrayMax(
+      List<? extends T> list) {
+
+    T max = null;
+    for (int i = 0; i < list.size(); i++) {
+      T item = list.get(i);
+      if (item != null && (max == null || item.compareTo(max) > 0)) {
+        max = item;
+      }
+    }
+    return max;
+  }
+
+  /** Support the ARRAY_MIN function. */
+  public static @Nullable <T extends Object & Comparable<? super T>> T arrayMin(
+      List<? extends T> list) {
+
+    T min = null;
+    for (int i = 0; i < list.size(); i++) {
+      T item = list.get(i);
+      if (item != null && (min == null || item.compareTo(min) < 0)) {
+        min = item;
+      }
+    }
+    return min;
+  }
+
   /** Support the ARRAY_REPEAT function. */
   public static @Nullable List<Object> repeat(Object element, Object count) {
     if (count == null) {
--- a/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
@@ -488,6 +488,21 @@ public abstract class ReturnTypes {
   };
 
   /**
+   * Returns the element type of a ARRAY or MULTISET.
+   *
+   * <p>For example, given <code>INTEGER ARRAY or MULTISET ARRAY</code>, returns
+   * <code>INTEGER</code>.
+   */
+  public static final SqlReturnTypeInference TO_COLLECTION_ELEMENT =
+      ARG0.andThen(SqlTypeTransforms.TO_COLLECTION_ELEMENT_TYPE);
+
+  public static final SqlReturnTypeInference TO_COLLECTION_ELEMENT_NULLABLE =
+      TO_COLLECTION_ELEMENT.andThen(SqlTypeTransforms.TO_NULLABLE);
+
+  public static final SqlReturnTypeInference TO_COLLECTION_ELEMENT_FORCE_NULLABLE =
+      TO_COLLECTION_ELEMENT.andThen(SqlTypeTransforms.FORCE_NULLABLE);
+
+  /**
    * Returns a MULTISET type.
    *
    * <p>For example, given <code>INTEGER</code>, returns
@@ -500,7 +515,7 @@ public abstract class ReturnTypes {
    * Returns the element type of a MULTISET.
    */
   public static final SqlReturnTypeInference MULTISET_ELEMENT_NULLABLE =
-      MULTISET.andThen(SqlTypeTransforms.TO_MULTISET_ELEMENT_TYPE);
+      MULTISET.andThen(SqlTypeTransforms.TO_COLLECTION_ELEMENT_TYPE);
 
   /**
    * Same as {@link #MULTISET} but returns with nullability if any of the
--- a/core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
+++ b/core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
@@ -630,6 +630,8 @@ public enum BuiltInMethod {
       Collection.class),
   ARRAY_COMPACT(SqlFunctions.class, "compact", List.class),
   ARRAY_DISTINCT(SqlFunctions.class, "distinct", List.class),
+  ARRAY_MAX(SqlFunctions.class, "arrayMax", List.class),
+  ARRAY_MIN(SqlFunctions.class, "arrayMin", List.class),
   ARRAY_REPEAT(SqlFunctions.class, "repeat", Object.class, Integer.class),
   ARRAY_REVERSE(SqlFunctions.class, "reverse", List.class),
   MAP_ENTRIES(SqlFunctions.class, "mapEntries", Map.class),
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -5376,6 +5376,37 @@ public class SqlOperatorTest {
     f.checkNull("array_distinct(null)");
   }
 
+  /** Tests {@code ARRAY_MAX} function from Spark. */
+  @Test void testArrayMaxFunc() {
+    final SqlOperatorFixture f0 = fixture();
+    f0.setFor(SqlLibraryOperators.ARRAY_MAX);
+    f0.checkFails("^array_max(array[1, 2])^",
+        "No match found for function signature ARRAY_MAX\\(<INTEGER ARRAY>\\)", false);
+
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
+    f.checkScalar("array_max(array[1, 2])", "2", "INTEGER");
+    f.checkScalar("array_max(array[1, 2, null])", "2", "INTEGER");
+    f.checkScalar("array_max(array[1])", "1", "INTEGER");
+    f.checkType("array_max(array())", "UNKNOWN");
+    f.checkNull("array_max(array())");
+    f.checkNull("array_max(cast(null as integer array))");
+  }
+
+  /** Tests {@code ARRAY_MIN} function from Spark. */
+  @Test void testArrayMinFunc() {
+    final SqlOperatorFixture f0 = fixture();
+    f0.setFor(SqlLibraryOperators.ARRAY_MIN);
+    f0.checkFails("^array_min(array[1, 2])^",
+        "No match found for function signature ARRAY_MIN\\(<INTEGER ARRAY>\\)", false);
+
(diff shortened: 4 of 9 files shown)


liuyongvs:
comments:

spark docs:

array_max Returns the maximum value in the array. NaN is greater than any non-NaN elements for double/float type. NULL elements are skipped.

i add NaN is greater than any non-NaN elements for double/float type it to calcite docs

because the NAN compareTo others always return true in java so the result is same with spark.

but it will confuse user if it is in the doc

liuyongvs:
hi @JiajunBernoulli will you also review it if you have time?

liuyongvs:
hi @NobiGo do you have time to help review?

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -487,6 +487,18 @@ public static SqlCall stripSeparator(SqlCall call) {
...
   };
 
+  /**
+   * Returns the element type of a ARRAY or MULTISET.
+   *
+   * <p>For example, given <code>INTEGER ARRAY or MULTISET ARRAY</code>, returns
+   * <code>INTEGER</code>.
+   */
+  public static final SqlReturnTypeInference TO_COLLECTION_ELEMENT =
+      ARG0.andThen(SqlTypeTransforms.TO_COLLECTION_ELEMENT_TYPE);
```
MasseGuillaume, line 497: Will fail with an empty array:
https://github.com/apache/calcite/pull/3218#discussion_r1209622163

MasseGuillaume (changes requested):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5336,6 +5336,36 @@ private static void checkIf(SqlOperatorFixture f) {
...
 
+  /** Tests {@code ARRAY_MAX} function from Spark. */
+  @Test void testArrayMaxFunc() {
+    final SqlOperatorFixture f0 = fixture();
+    f0.setFor(SqlLibraryOperators.ARRAY_MAX);
+    f0.checkFails("^array_max(array[1, 2])^",
+        "No match found for function signature ARRAY_MAX\\(<INTEGER ARRAY>\\)", false);
+
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
+    f.checkScalar("array_max(array[1, 2])", "2",
```
MasseGuillaume, line 5347: minimum case:
`f.checkScalar("array_max(array[1])", "1", "INTEGER NOT NULL");`

empty case:
`f.checkScalar("array_max(array())", "NULL", "UNKNOWN");`

liuyongvs: @MasseGuillaume  thanks, add the test.

and the type should always nullable, which you can see here 
```
case class ArrayMax(child: Expression)
  extends UnaryExpression with ImplicitCastInputTypes with NullIntolerant {

  override def nullable: Boolean = true
```

https://github.com/apache/spark/blob/014dd357656106cc8dc0fcfb30cbe067a714916c/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/collectionOperations.scala#L2126

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
liuyongvs: @MasseGuillaume  
it doesn't faill, because array() type is UNKNOWN NOT NULL ARRAY NOT NULL,
so the TO_COLLECTION_ELEMENT will return UNKNOWN NOT NULL

liuyongvs:
hi @tanclary @MasseGuillaume @JiajunBernoulli fix all your reviews

liuyongvs:
hi @tanclary do you also have time to review it?

MasseGuillaume (approved):

JiajunBernoulli (approved):

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -500,7 +515,7 @@ public static SqlCall stripSeparator(SqlCall call) {
    * Returns the element type of a MULTISET.
    */
   public static final SqlReturnTypeInference MULTISET_ELEMENT_NULLABLE =
-      MULTISET.andThen(SqlTypeTransforms.TO_MULTISET_ELEMENT_TYPE);
+      MULTISET.andThen(SqlTypeTransforms.TO_COLLECTION_ELEMENT_TYPE);
```
JiajunBernoulli, line 518: Do you think SEQ or LIST better than COLLECTION?

JiajunBernoulli (changes requested):

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
liuyongvs: @JiajunBernoulli i modify the name because it can be use not only multiset type, but also array type

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
```diff
@@ -629,6 +629,8 @@ public enum BuiltInMethod {
   SUBMULTISET_OF(SqlFunctions.class, "submultisetOf", Collection.class,
       Collection.class),
   ARRAY_DISTINCT(SqlFunctions.class, "distinct", List.class),
+  ARRAY_MAX(SqlFunctions.class, "max",  List.class),
```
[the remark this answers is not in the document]
liuyongvs: i does that because i refer to array_reverse implmented by snuyanzin, the function also names reverse. so i do it too @JiajunBernoulli do you think it is need? i can modify all

liuyongvs:
@JiajunBernoulli fix all your reviews and squash the commits

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
JiajunBernoulli: `COLLECTION` usually also includes `MAP` and `SET`, does them also work?

liuyongvs: @JiajunBernoulli COLLECTION includes array and multiset, both have getComponentType, but doesn't include map type. map type has getKeyType and getValueType

liuyongvs:
@JiajunBernoulli  git rebase to fix conflicts with main branch

JiajunBernoulli (approved):

liuyongvs:
hi @JiajunBernoulli could it be merged otherwise it will conflict with SORT_ARRAY when ci passed. i just fix conflict

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3870,7 +3870,7 @@ private static AtomicLong getAtomicLong(String key) {
   }
 
   /** Support the ARRAY_COMPACT function. */
-  public static List compact(List list) {
+  public static List arrayCompact(List list) {
```
snuyanzin, line 3873: why are we going to rename public methods?
that could impact downstream dependencies

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3881,13 +3881,41 @@ public static List compact(List list) {
   }
 
   /** Support the ARRAY_DISTINCT function. */
-  public static List distinct(List list) {
+  public static List arrayDistinct(List list) {
```
snuyanzin, line 3884: why are we going to rename public methods?
that could impact downstream dependencies

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3881,13 +3881,41 @@ public static List compact(List list) {
...
+      if (item != null && (min == null || item.compareTo(min) < 0)) {
+        min = item;
+      }
+    }
+    return min;
+  }
+
   /** Support the ARRAY_REPEAT function. */
-  public static @Nullable List<Object> repeat(Object element, Object count) {
+  public static @Nullable List<Object> arrayRepeat(Object element, Object count) {
```
snuyanzin, line 3918: why are we going to rename public methods?
that could impact downstream dependencies

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4038,7 +4066,7 @@ public static Collection multisetUnionAll(Collection collection1,
   }
 
   /** Support the ARRAY_REVERSE function. */
-  public static List reverse(List list) {
+  public static List arrayReverse(List list) {
```
snuyanzin, line 4069: why are we going to rename public methods?
that could impact downstream dependencies

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
liuyongvs: @snuyanzin this method merged in the same version, just a few days, do not have released

liuyongvs:
hi @snuyanzin thanks for your review, this method justs merged a few days and also in the 1.35 version(snapshot), which don't have released, we think the method name is not good for search, which builtin function belongs to

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
snuyanzin: ok then the naming became more confusing: `arrayCompact` says about array however returns a `List`

liuyongvs: @snuyanzin just for unify with others, i use removeNulls before, but change it to arrayCompact when reviewed.

snuyanzin: I didn't get it

I see a review comment here https://github.com/apache/calcite/pull/3233#discussion_r1210553037
to use `compact` name and from my point of view this is fine since with such name there is no confusion about `List` and `array`
Where in review `arrayCompact` has appeared?

liuyongvs: @snuyanzin 
"this could also be named compact so it's easier to find, removeNulls is also ok", this is in the comments you refer.
originally i use removeNulls then changed to compact, and i changed to arrayCompact just because the spark function name is arrayCompact, which others do it too. for search easy.
https://github.com/apache/calcite/pull/3212#discussion_r1213835525

snuyanzin: my concern is not about search, moreover i don't think search is an issue since you have a comment with sql version
my concern is about name confusion which I raised in this thread
> ok then the naming became more confusing: `arrayCompact` says about `array` however returns a `List`

liuyongvs: @snuyanzin i know you meaning. List is just array type's runtime implementation. from my side, the name is ok. if you think it is not good, i can change all of them to revert.

liuyongvs: hi @JiajunBernoulli what is your opinion, should it revert? if do don't oppose it, i will revert according to @snuyanzin

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
JiajunBernoulli: 1.35 has not been released yet, it will not affect downstream dependencies.

The old name is not accurate enough, we should solve this problem before the release version.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
JiajunBernoulli:
```
  ARRAY_MAX(SqlFunctions.class, "arrayMax", List.class),
  ARRAY_MIN(SqlFunctions.class, "arrayMin", List.class),
```
should be save.

Other function name can be revert.

liuyongvs: @JiajunBernoulli has reverted, thanks for your suggestion @JiajunBernoulli @snuyanzin

PR #3214, opened 2023-05-19
[CALCITE-5608] Implement ASINH, ACOSH, ATANH functions

zoudan:


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1854,6 +1854,21 @@ public static double acos(double b0) {
...
+  // ACOSH
+  /** SQL <code>ACOSH</code> operator applied to BigDecimal values. */
+  public static double acosh(BigDecimal b0) {
+    return acosh(b0.doubleValue());
+  }
+
+  /** SQL <code>ACOSH</code> operator applied to double values. */
+  public static double acosh(double b0) {
+    if (b0 < 1) {
+      SQLException e = new SQLException("Input parameter of acosh cannot be small than 1!");
```
rubenada, line 1866: I would lean towards throwing an IllegalArgumentException in this case (same applies to the `atanh`method).
Also, I think it would me more correct if the message says "... cannot be less than 1".

zoudan: Thanks, I will update it soon

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1263,6 +1263,32 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
...
+
+  /** The {@code ASINH(numeric)} function; returns the inverse hyperbolic sine of {@code value}. */
+  @LibraryOperator(libraries = {ALL})
+  public static final SqlFunction ASINH =
+      SqlBasicFunction.create("ASINH",
+          ReturnTypes.DOUBLE_NULLABLE,
+          OperandTypes.NUMERIC,
+          SqlFunctionCategory.NUMERIC);
+
+  /** The {@code ATANH(numeric)} function. returns the inverse hyperbolic tangent
```
rubenada, line 1283: minor: I think you meant to have a `;` instead of `.` after "function"

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1854,6 +1854,20 @@ public static double acos(double b0) {
...
+  public static double acosh(BigDecimal b0) {
+    return acosh(b0.doubleValue());
+  }
+
+  /** SQL <code>ACOSH</code> operator applied to double values. */
+  public static double acosh(double b0) {
+    if (b0 < 1) {
+      throw new IllegalArgumentException("Input parameter of acosh cannot be less than 1!");
+    }
+    return Math.log(Math.sqrt(b0 * b0 - 1.0) + b0);
```
rubenada, line 1868: For the newbies on trigonometric functions :) , is there any link you can provide to show where these formulas come from?

zoudan: https://en.wikipedia.org/wiki/Inverse_hyperbolic_functions

rubenada, line 1868: Super-nitpick: for consistency with the other methods, I think we should use `1.0d` instead of  `1.0` here

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1897,6 +1930,29 @@ public static double atan2(double b0, double b1) {
...
 
+  // ATANH
+  /** SQL <code>ATANH</code> operator applied to BigDecimal values. */
+  public static double atanh(BigDecimal b) {
+    return atanh(b.doubleValue());
+  }
+
+  /** SQL <code>ATANH</code> operator applied to double values. */
+  public static double atanh(double b) {
+    if (b > 1 || b < -1) {
```
rubenada, line 1941: According to [this](https://en.wikipedia.org/wiki/Inverse_hyperbolic_functions) and [this](https://support.microsoft.com/en-us/office/atanh-function-3cd65768-0de7-4f1d-b312-d01c8c930d90), I think that 1 and -1 should also be forbidden:
```
if (b => 1 || b <= -1) {
  throw new IllegalArgumentException(...);
}
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5539,6 +5539,72 @@ private static void checkIf(SqlOperatorFixture f) {
...
+      f.checkType("acosh(case when false then 1 else null end)", "DOUBLE");
+      f.checkType("acosh('abc')", "DOUBLE NOT NULL");
+      f.checkScalarApprox("acosh(1)", "DOUBLE NOT NULL",
+          isWithin(0d, 0.0001d));
+      f.checkScalarApprox("acosh(cast(10 as decimal))", "DOUBLE NOT NULL",
+          isWithin(2.9932d, 0.0001d));
+      f.checkNull("acosh(cast(null as integer))");
+      f.checkNull("acosh(cast(null as double))");
+      f.checkFails("acosh(0.1)",
+          "Input parameter of acosh cannot be small than 1!",
```
rubenada, line 5559: How is this test passing? Shouldn't the expected error message be adapted?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5539,6 +5539,72 @@ private static void checkIf(SqlOperatorFixture f) {
...
+      f.checkType("atanh(cast(0 as Decimal))", "DOUBLE NOT NULL");
+      f.checkType("atanh(case when false then 0 else null end)", "DOUBLE");
+      f.checkType("atanh('abc')", "DOUBLE NOT NULL");
+      f.checkScalarApprox("atanh(0.76159416)", "DOUBLE NOT NULL",
+          isWithin(1d, 0.0001d));
+      f.checkScalarApprox("atanh(cast(-0.1 as decimal))", "DOUBLE NOT NULL",
+          isWithin(-0.1003d, 0.0001d));
+      f.checkNull("atanh(cast(null as integer))");
+      f.checkNull("atanh(cast(null as double))");
+      f.checkFails("atanh(10)",
```
rubenada, line 5601: Could you please add tests for the cases 1 and -1? (which should fail according to the links I provided on the separate comment)

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
zoudan: I found it is because we only check this expression will be successful during parse and validate. The error message has been ignored, you could see more details in https://github.com/apache/calcite/blob/f32289870be80e69dabc5b7e24bd6c777d4eb688/testkit/src/main/java/org/apache/calcite/test/SqlOperatorFixtureImpl.java#L161

Maybe we should improve this?

rubenada: Ah, ok. Thanks for the info. It is certainly counter-intuitive and probably there's room for improvement, but IMO this is out of the scope of the current PR.
I would say, please just align this "expected error message" with the actual error that is thrown, just for the sake of correctness (and in case the fixture mechanism gets improved in the future).

rubenada:
LGTM

rubenada:
@tanclary I see you have participated on other trigonometric-related PRs, would you like to take a look at this one?
I think it is in a good shape and I will proceed to merge (after squashing commits)

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1897,6 +1930,29 @@ public static double atan2(double b0, double b1) {
...
   }
 
+  // ATANH
+  /** SQL <code>ATANH</code> operator applied to BigDecimal values. */
+  public static double atanh(BigDecimal b) {
+    return atanh(b.doubleValue());
+  }
+
+  /** SQL <code>ATANH</code> operator applied to double values. */
+  public static double atanh(double b) {
```
tanclary, line 1940: nit: could just check abs(b) <= 1

tanclary (commented):
@rubenada @zoudan  left one small comment, LGTM!

tanclary (commented):
@rubenada @zoudan  left one small comment, LGTM!

rubenada:
Thanks @tanclary !

@zoudan could you please squash commits into a single one? I'll proceed with the merge afterwards.

zoudan:
@rubenada Thanks, I have squashed the commits

PR #3219, opened 2023-05-23
[CALCITE-5711] Implement the SINGLE_VALUE aggregation in PostgreSQL D…

NobiGo:
…ialect

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/PostgresqlSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/PostgresqlSqlDialect.java
@@ -22,20 +22,27 @@ import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
 import org.apache.calcite.rel.type.RelDataTypeSystemImpl;
 import org.apache.calcite.sql.SqlAlienSystemTypeNameSpec;
+import org.apache.calcite.sql.SqlBasicCall;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/PostgresqlSqlDialect.java
```diff
@@ -97,6 +104,39 @@ public PostgresqlSqlDialect(Context context) {
         SqlParserPos.ZERO);
   }
 
+  @Override public SqlNode rewriteSingleValueExpr(SqlNode aggCall, RelDataType relDataType) {
+    final SqlNode operand = ((SqlBasicCall) aggCall).operand(0);
+    final SqlLiteral nullLiteral = SqlLiteral.createNull(SqlParserPos.ZERO);
+    final SqlNode unionOperand =
```
ILuffZhe, line 110: Out of curiosity, why do we cast null here. What if we use **select null union all select null**.

ILuffZhe (commented):
Just one small question about this PR.

core/src/main/java/org/apache/calcite/sql/dialect/PostgresqlSqlDialect.java
NobiGo: When we use `select null union all select null`, The PostgreSQL will throw "ERROR: CASE types text and bigint(The input type) cannot be matched", So we need to tell it the NULL's type.

ILuffZhe (approved):
LGTM!

PR #3222, opened 2023-05-24
[CALCITE-5674] CAST expr to target type should respect nullable when it is complex type (follow-up)

rubenada:
Fixes minor regression introduced by https://github.com/apache/calcite/pull/3189


rubenada:
@libenchao could you please take a quick look at this patch (as we agree on a separate discussion)

libenchao (approved):
+1

PR #3223, opened 2023-05-24
[CALCITE-5721] Capture build scans on ge.apache.org to benefit from deep build insights

clayburn:
This PR publishes a build scan for every CI build on Jenkins and GitHub Actions and for every local build from an authenticated Apache committer. The build will not fail if publishing fails.

The build scans of the Apache Calcite project are published to the Gradle Enterprise instance at [ge.apache.org](https://ge.apache.org/), hosted by the Apache Software Foundation and run in partnership between the ASF and Gradle. This Gradle Enterprise instance has all features and extensions enabled and is freely available for use by the Apache Calcite project and all other Apache projects.

Currently, the Apache Calcite build is configured to publish build scans to the publicly available [scans.gradle.com](https://scans.gradle.com/). This pull request enhances that functionality by instead publishing build scans to [ge.apache.org](https://ge.apache.org/). On this Gradle Enterprise instance, Apache Calcite will have access not only to all of the published build scans but other aggregate data features such as:

- Dashboards to view all historical build scans, along with performance trends over time
- Build failure analytics for enhanced investigation and diagnosis of build failures
- Test failure analytics to better understand trends and causes around slow, failing, and flaky tests

If interested in exploring a fully populated Gradle Enterprise instance, please explore the Maven builds already connected to [ge.apache.org](https://ge.apache.org/), the [Spring project’s instance](https://ge.spring.io/scans?search.relativeStartTime=P28D&search.timeZoneId=Europe/Zurich), or any number of other [OSS projects](https://gradle.com/enterprise-customers/oss-projects/) for which we sponsor instances of Gradle Enterprise.

Please let me know if there are any questions about the value of Gradle Enterprise or the changes in this pull request and I’d be happy to address them.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -21,6 +21,8 @@ pluginManagement {
         fun String.v() = extra["$this.version"].toString()
         fun PluginDependenciesSpec.idv(id: String, key: String = id) = id(id) version key.v()
 
+        idv("com.gradle.enterprise")
+        idv("com.gradle.common-custom-user-data-gradle-plugin")
         idv("com.autonomousapps.dependency-analysis")
         idv("org.checkerframework")
         idv("com.github.autostyle")
@@ -53,8 +55,8 @@ pluginManagement {
 }
 
 plugins {
-    id("com.gradle.enterprise") version "3.13.2"
-    id("com.gradle.common-custom-user-data-gradle-plugin") version "1.10"
+    id("com.gradle.enterprise")
+    id("com.gradle.common-custom-user-data-gradle-plugin")
     id("com.github.burrunan.s3-build-cache")
 }
 
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -14,6 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import com.gradle.enterprise.gradleplugin.internal.extension.BuildScanExtensionWithHiddenFeatures
+
 pluginManagement {
     plugins {
         fun String.v() = extra["$this.version"].toString()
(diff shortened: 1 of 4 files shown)


settings.gradle.kts
```diff
@@ -100,12 +104,18 @@ fun property(name: String) =
 
 val isCiServer = System.getenv().containsKey("CI")
 
-if (isCiServer) {
-    gradleEnterprise {
-        buildScan {
-            termsOfServiceUrl = "https://gradle.com/terms-of-service"
-            termsOfServiceAgree = "yes"
-            tag("CI")
+gradleEnterprise {
```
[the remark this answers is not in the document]
clayburn: Really, you have three scenarios here:

1. CI builds - With this PR, nothing changes regarding publishing of CI builds, except that they go to https://ge.apache.org. Your builds will pick up credentials automatically in Jenkins and GitHub.
2. Authenticated Local builds - Apache committers _can_ publish local build scans to https://ge.apache.org by using LDAP credentials to login and create an access token. If that is done, then local scans are published.
3. Unauthenticated Local builds - For these, the plugin will be applied, but since the user is not authenticated to https://ge.apache.org, nothing actually happens.

So, to summarize, you won't see any real differences to _when_ a build scan gets published, unless an Apache committer explicitly opts in to the local scans.

settings.gradle.kts
```diff
@@ -51,7 +53,8 @@ pluginManagement {
 }
 
 plugins {
-    `gradle-enterprise`
+    id("com.gradle.enterprise") version "3.13.2"
+    id("com.gradle.common-custom-user-data-gradle-plugin") version "1.10"
```
zabetak, line 57: Usually we keep versions centralized in `gradle.properties` file. Is it possible to use the `pluginManagement` section and the `idv` function to handle this?

zabetak (commented):
Overall LTGM. Is there a way to test that it works as expected before merging?

settings.gradle.kts
clayburn: Done. Since this is a settings plugin, it still needs to be applied here, but the versioning is not handled as you handle other plugins.

zabetak (approved):
I tested the patch by publishing authenticated scans from my local environment and things look good. I left a minor question and will merge this as soon as I get an answer.

PR #3224, opened 2023-05-24
[CALCITE-5722] RangeSets.isPoint incorrectly gives false for closed ranges with endpoints which are equal under `.compareTo` but not under `.equals` 

ian-bertolacci:
This PR fixes a peculiar issue where RangeSets.isPoint will return false when a closed range with endpoints that are equal under `Comparable.compareTo`, but not `T.equals`.

Using equality as determined by `Comparable.compareTo` is the most correct notion of equality in this situation.

### DIFF
--- a/core/src/main/java/org/apache/calcite/util/RangeSets.java
+++ b/core/src/main/java/org/apache/calcite/util/RangeSets.java
@@ -129,7 +129,7 @@ public class RangeSets {
   public static <C extends Comparable<C>> boolean isPoint(Range<C> range) {
     return range.hasLowerBound()
         && range.hasUpperBound()
-        && range.lowerEndpoint().equals(range.upperEndpoint())
+        && range.lowerEndpoint().compareTo(range.upperEndpoint()) == 0
         && !range.isEmpty();
   }
 
--- a/core/src/test/java/org/apache/calcite/util/RangeSetTest.java
+++ b/core/src/test/java/org/apache/calcite/util/RangeSetTest.java
@@ -155,6 +155,9 @@ class RangeSetTest {
     assertThat(RangeSets.isPoint(Range.atMost(0)), is(false));
     assertThat(RangeSets.isPoint(Range.greaterThan(0)), is(false));
     assertThat(RangeSets.isPoint(Range.atLeast(0)), is(false));
+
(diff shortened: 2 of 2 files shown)


ian-bertolacci:
Pertinent line from the [`Comparable` documents](https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html#:~:text=Virtually%20all%20Java%20core%20classes%20that%20implement%20Comparable%20have%20natural%20orderings%20that%20are%20consistent%20with%20equals.%20One%20exception%20is%20java.math.BigDecimal%2C%20whose%20natural%20ordering%20equates%20BigDecimal%20objects%20with%20equal%20values%20and%20different%20precisions%20(such%20as%204.0%20and%204.00).):

> Virtually all Java core classes that implement Comparable have natural orderings that are consistent with equals. One exception is java.math.BigDecimal, whose natural ordering equates BigDecimal objects with equal values and different precisions (such as 4.0 and 4.00).

core/src/test/java/org/apache/calcite/util/RangeSetTest.java
```diff
@@ -155,6 +155,9 @@ class RangeSetTest {
     assertThat(RangeSets.isPoint(Range.atMost(0)), is(false));
     assertThat(RangeSets.isPoint(Range.greaterThan(0)), is(false));
     assertThat(RangeSets.isPoint(Range.atLeast(0)), is(false));
+
+    // Test situation where endpoints of closed range are equal under `Comparable.compareTo` but not `T.equals`
```
zoudan, line 159: line too long, please check the code style

core/src/main/java/org/apache/calcite/util/RangeSets.java
```diff
@@ -129,7 +129,7 @@ public static <C extends Comparable<C>> int hashCode(RangeSet<C> rangeSet) {
   public static <C extends Comparable<C>> boolean isPoint(Range<C> range) {
     return range.hasLowerBound()
         && range.hasUpperBound()
-        && range.lowerEndpoint().equals(range.upperEndpoint())
+        && range.lowerEndpoint().compareTo(range.upperEndpoint()) == 0
```
zoudan, line 132: Maybe it is better to add comment for why we use `compareTo` here

PR #3225, opened 2023-05-25
[CALCITE-5723] Oracle dialect generates SQL that cannot be recognized by lower version Oracle Server(<12) when unparsing OffsetFetch

ILuffZhe:


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7089,7 +7105,18 @@ Sql withMysql8() {
...
+      return withOracle(12);
+    }
+
+    Sql withOracle(int majorVersion) {
+      final SqlDialect oracleDialect = DatabaseProduct.ORACLE.getDialect();
+      return dialect(
+          new OracleSqlDialect(OracleSqlDialect.DEFAULT_CONTEXT
+              .withDatabaseProduct(DatabaseProduct.ORACLE)
+              .withDatabaseMajorVersion(majorVersion)
+              .withIdentifierQuoteString(oracleDialect.quoteIdentifier("")
```
JiajunBernoulli, line 7117: Why need `withIdentifierQuoteString` and `withNullCollation`?

ILuffZhe: In this case, we don't need those two configuration. But I'm wandering if they might be reused in the future cases? And I also see that both `withMssql` and `withMysql8` do the same thing.

core/src/main/java/org/apache/calcite/sql/dialect/OracleSqlDialect.java
```diff
@@ -188,4 +191,13 @@ public OracleSqlDialect(Context context) {
...
     }
   }
+
+  @Override public void unparseOffsetFetch(SqlWriter writer, @Nullable SqlNode offset,
+       @Nullable SqlNode fetch) {
+    // majorVersion in SqlDialect.EMPTY_CONTEXT is -1 by default
+    if (this.majorVersion != -1 && this.majorVersion < 12) {
+      throw new RuntimeException("Lower Oracle version(<12) doesn't support offset/fetch syntax!");
+    }
+    unparseFetchUsingAnsi(writer, offset, fetch);
```
JiajunBernoulli, line 201: `super.unparseOffsetFetch(writer, offset, fetch))` is better for me.

ILuffZhe: Make sense, I've updated the PR.

JiajunBernoulli (approved):

ILuffZhe:
@JiajunBernoulli  Thanks for your review!

ILuffZhe:
Squash the commits into one.

PR #3227, opened 2023-05-27
[CALCITE-5727] RelOptFixture#checkUnchanged should assert planAfter i…

libenchao:
…s not present


gianm (commented):
Makes sense to me. I suggest replacing the `assert` with the following, so people adding new tests will get better error messages, if they get this wrong:

```
      if (!"${planAfter}".equals(expandedPlanAfter)) {
        throw new AssertionError("Expected planAfter must not be present when using unchanged=true " 
            + "or calling checkUnchanged.");
      }
```

Additionally, the current code has `expandedPLanAfter`, but it should be `expandedPlanAfter` (capitalization).

libenchao:
@gianm Thanks for the review, I've address your comment.

gianm:
Thank you @libenchao!

PR #3228, opened 2023-05-29
[CALCITE-5730] Initial null values can be dropped by EnumerableLimitSort with offset

gf2121:


rubenada:
Thanks for the PR @gf2121 .
First of all, as a general comment, the commit message should describe the problem, and not being with "Fix ...". I propose to use as commit message (together with the ticket number) the title in Jira (which I also have adjusted slightly).
There seems to be some issue with CheckStyle and CheckerFramework, could you please take a look?

linq4j/src/main/java/org/apache/calcite/linq4j/EnumerableDefaults.java
```diff
@@ -2732,6 +2732,7 @@ public static <TSource, TKey> Enumerable<TSource> orderBy(
           // search the key up to (but excluding) which we have to remove entries from the map
           int skipped = 0;
           TKey until = null;
+          boolean hasResult = false;
```
rubenada, line 2735: I wonder if it would be clearer to keep the same approach in the code, but just using another "unset value" instead of `null`, e.g. the `DUMMY` object that already exists in the class:
```
TKey until = (TKey) DUMMY;
...
if (until == DUMMY)
```

gf2121:
> First of all, as a general comment, the commit message should describe the problem, and not being with "Fix ...". I propose to use as commit message (together with the ticket number) the title in Jira (which I also have adjusted slightly).

Thank you very much for your suggestion! Jira's adjusted title is much clearer. Sorry for my poor english. I have changed the PR title and commit message to be consistent with Jira ticket.

rubenada:
> Thank you very much for your suggestion! Jira's adjusted title is much clearer. Sorry for my poor english. I have changed the PR title and commit message to be consistent with Jira ticket.

No problem :) 
I have added a comment in Jira with a test class with a test that systematically reproduces the problem, could you please include it in the PR? Also, since we are adding this new test class for `EnumerableLimitSortTest`, I think it would be a nice bonus to include a couple of extra tests (even if they are unrelated to the current bug) just to show other sql queries scenarios for `EnumerableLimitSort` (nulls last, order on several columns, limit without offset, ...); the {{HrSchemaBig}} that I used on the bug test should be enough to provide all these scenarios.

gf2121:
@rubenada Thanks a lot for your patient guidance!
I have added some more tests around `EnumerableLimitSort`.

rubenada:
@gf2121 overall looks good. I left you one minor comment to improve the new tests (to make them more varied).

Apart from that, a couple of general comments (it is OK for this PR, but you might want to keep them in mind for your next PR):
- Generally we launch the PR from a source branch which is created expressly for it (usually named after the Jira ticket number, or the description of the feature/bug).
- Generally, it is better to refrain from doing force push during the review process, it is better to keep the commit history to facilitate an incremental review. At the end of the review, when the PR looks good to be merged, we squash commits into a single one and force push before the final merge.

gf2121:
@rubenada Very appreciated for your great advice!  I'll follow them in next PRs.

`I left you one minor comment to improve the new tests (to make them more varied).`
Sorry but I could not find any comments in the code. Did i misunderstand something ?

thomasrebele:
I'm sorry to have introduced this bug. Thanks for fixing it! Changes LGTM.

rubenada:
Thanks for your review, @thomasrebele ! And no problem at all, I also missed this during the review of EnumerableLimitSort. .

@gf2121 what's the status here? Do you have time for the final minor changes that I requested on the new test class? 

> Minor request: could you make the non-bug-related tests a bit more varied: different offset/limit values (not always 1), use asc/desc on order by, use different columns (not always commission+salary)

gf2121:
> Minor request: could you make the non-bug-related tests a bit more varied: different offset/limit values (not always 1), use asc/desc on order by, use different columns (not always commission+salary)

@rubenada Sorry, for some reason I could not see this comment on github. Just got that from your latest comment. I have added a new commit to fix it. will squash commits when ready to merge :)

PR #3229, opened 2023-05-29
[CALCITE-5425] Should not pushdown Filter through Aggregate without group keys

zoudan:


libenchao (approved):
+1, the comment in this PR is really helpful, thanks for the work!

PR #3230, opened 2023-05-29
[CALCITE-5568] When there are LogicalValues, the DecorrelateQuery continues

guestli:


core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -488,7 +488,7 @@ protected RexNode removeCorrelationExpr(
 
   public @Nullable Frame decorrelateRel(Values rel, boolean isCorVarDefined) {
     // There are no inputs, so rel does not need to be changed.
-    return null;
+    return decorrelateRel((RelNode) rel, isCorVarDefined);
```
zoudan, line 491: Perhaps we should improve the commit message and also the title of this PR because it did not clearly describe the problem

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -7901,4 +7901,22 @@ private void checkJoinAssociateRuleWithTopAlwaysTrueCondition(boolean allowAlway
         .check();
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-5568">[CALCITE-5568]
+   * The effect of the DecorrelateQuery when there is LogicalValues
```
zoudan, line 7907: This should be consistent with abstract of the Jira.

PR #3231, opened 2023-05-30
[CALCITE-5701] Add NAMED_STRUCT function (enabled in Spark library)

MasseGuillaume:


site/_docs/reference.md
```diff
@@ -2728,6 +2728,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b o | LTRIM(string)                                | Returns *string* with all blanks removed from the start
 | b m p | MD5(string)                                | Calculates an MD5 128-bit checksum of *string* and returns it as a hex string
 | m | MONTHNAME(date)                                | Returns the name, in the connection's locale, of the month in *datetime*; for example, it returns '二月' for both DATE '2020-02-10' and TIMESTAMP '2020-02-10 10:10:10'
+| s | NAMED_STRUCT(k_1, v_1, [, k_i, v_i]*)          | Creates a struct with the given field names and values.
```
liuyongvs, line 2731: remove end '.'

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -827,6 +828,47 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
+    final RelDataTypeFactory typeFactory = operatorBinding.getTypeFactory();
+    return typeFactory.createStructType(values, keys);
+  }
+
+  /** Creates a struct with the given field names and values. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction NAMED_STRUCT =
+      SqlBasicFunction.create("NAMED_STRUCT",
+          SqlLibraryOperators::namedStructReturnType,
+          OperandTypes.SAME_VARIADIC);
```
liuyongvs, line 870: put "NAMED_STRUCT" in SqlKind

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -827,6 +828,47 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
 
+  private static RelDataType namedStructReturnType(SqlOperatorBinding operatorBinding) {
+    if (operatorBinding.getOperandCount() < 2) {
+      throw operatorBinding.newError(Static.RESOURCE.namedStructRequiresTwoOrMoreArgs());
+    }
+    if (operatorBinding.getOperandCount() % 2 != 0) {
+      throw operatorBinding.newError(Static.RESOURCE.namedStructRequiresEvenNumberOfArgs());
+    }
+    List<String> keys = new ArrayList<String>();
+    List<RelDataType> values = new ArrayList<RelDataType>();
```
liuyongvs, line 839: calcite try best to using final,
final List<String> keys 
final List<RelDataType> values

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7909,6 +7909,45 @@ private static void checkArrayConcatAggFuncFails(SqlOperatorFixture t) {
         "17357000", "BIGINT NOT NULL");
   }
```
liuyongvs, line 7911: only has the definition of function, where is the implementation of the function

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7909,6 +7909,45 @@ private static void checkArrayConcatAggFuncFails(SqlOperatorFixture t) {
...
 
+  @Test void testNamedStruct() {
+    final SqlOperatorFixture f = fixture().withLibrary(SqlLibrary.SPARK);
+    f.setFor(SqlLibraryOperators.NAMED_STRUCT, VmName.EXPAND);
+
+    f.checkFails(
```
liuyongvs, line 7916: need passitive test to test the result

site/_docs/reference.md
julianhyde: also remove spurious ','.

might be clearer to write 'name1, value1' rather than 'k_1, v_1'.

julianhyde:
Looks close. But I'd be more confident if this was executable in a Quidem test.

See how `functions.iq` has `!use mysqlfunc` and maybe define a "scott-spark" in CoreQuidemTest that behaves similarly.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
MasseGuillaume: Make sense, I think I will name it `STRUCT_CONSTRUCTOR`
based on:
```
* <p>Only commonly-used nodes have their own type; other nodes are of type
 * {@link #OTHER}. Some of the values, such as {@link #SET_QUERY}, represent
 * aggregates.</p>
```

In my PR for CALCITE-5624, the `array` Apache Spark function should have been of kind `ARRAY_VALUE_CONSTRUCTOR` I think.

So if we implement BigQuery's struct constructor we could share the Kind `STRUCT_CONSTRUCTOR`:
https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#constructing_a_struct

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
MasseGuillaume: We should use Scala 🤣 .

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
MasseGuillaume: I will add a Quidem test https://github.com/apache/calcite/pull/3231#issuecomment-1569341799

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7909,6 +7909,45 @@ private static void checkArrayConcatAggFuncFails(SqlOperatorFixture t) {
...
+    // in Apache Spark, we do not support it for simplicity
+    // scala> spark.sql("select named_struct('k' || '1', 1) as s").printSchema
+    // root
+    //  |-- s: struct (nullable = false)
+    //  |    |-- k1: integer (nullable = false)
+    f.checkFails(
+        "^named_struct('k1', 1, 'k' || '1', 2)^",
+        "named_struct key is not a string literal, got an expression",
+        false);
+  }
```
MasseGuillaume, line 7949: Is there a way to get the position of the parameters via `SqlOperatorBinding`? 

For example, I was trying to get more precision positions on the test failures:

```
f.checkFails(
        "named_struct('k1', 1, ^2^, 3)",
        "named_struct key is not a string literal, found type 'BigDecimal'",
        false);

f.checkFails(
        "named_struct('k1', 1, ^'k' || '1'^, 2)",
        "named_struct key is not a string literal, got an expression",
        false);
```

MasseGuillaume:
I am struggling with the implementation.

`named_struct` looks really similar to `MAP_VALUE_CONSTRUCTOR`:

https://github.com/apache/calcite/blob/58120b2d84a47d4703b26df7c2d91e16b8c754e7/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java#L3125

How would you build a Row with the expression dsl? What's the underlying class that I would use in the blockBuilder?

julianhyde:
While `named_struct(a, 1, b, true)` is similar to `MAP`, it's even more similar to `ROW`. The value at runtime needs to be a Java list, just like `ROW`. The only difference between `named_struct` and `ROW` is the SQL type.

Could you implement `named_struct` the same way as `ROW`?

MasseGuillaume:
@julianhyde I implemented named_struct with a runtime of Array Object.

https://github.com/apache/calcite/pull/3231/commits/d7eb559244ca3bbd9992b2d87eb3057e72a8482c#diff-2095e662d66d6c1b2851915fa0d73934c3a847550545396835e02d88a9f5e2daR2518

However, if you try to access a field with the `item` operator it fails since there is no field name information at runtime:

https://github.com/apache/calcite/blob/d1a12bb6f4d4f617234dfbc32cb0813c473b0029/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java#L4482

Should I create a new runtime type that can keep both the field name and the value and add another case to the `structAccess` function?

MasseGuillaume:
Take 2: https://github.com/apache/calcite/pull/3231

PR #3233, opened 2023-05-30
[CALCITE-5734] Add ARRAY_COMPACT function (enabled in Spark library)

liuyongvs:
array_compact(array) - Removes null values from the array.
Examples:

> SELECT array_compact(array(1, 2, 3, null));

[1,2,3]

> SELECT array_compact(array("a", "b", "c"));

 ["a","b","c"]

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -3869,6 +3869,17 @@ public class SqlFunctions {
     return atomic;
   }
 
+  /** Support the ARRAY_COMPACT function. */
+  public static List compact(List list) {
+    final List result = new ArrayList();
+    for (Object element : list) {
+      if (element != null) {
+        result.add(element);
+      }
+    }
+    return result;
+  }
+
   /** Support the ARRAY_DISTINCT function. */
   public static List distinct(List list) {
     Set result = new LinkedHashSet<>(list);
--- a/core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
+++ b/core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
@@ -628,6 +628,7 @@ public enum BuiltInMethod {
   IS_EMPTY(Collection.class, "isEmpty"),
   SUBMULTISET_OF(SqlFunctions.class, "submultisetOf", Collection.class,
       Collection.class),
+  ARRAY_COMPACT(SqlFunctions.class, "compact", List.class),
   ARRAY_DISTINCT(SqlFunctions.class, "distinct", List.class),
   ARRAY_REPEAT(SqlFunctions.class, "repeat", Object.class, Integer.class),
   ARRAY_REVERSE(SqlFunctions.class, "reverse", List.class),
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -5328,6 +5328,25 @@ public class SqlOperatorTest {
     f.checkScalar("rand_integer(2, 11)", 1, "INTEGER NOT NULL");
   }
 
+  /** Tests {@code ARRAY_COMPACT} function from Spark. */
+  @Test void testArrayCompactFunc() {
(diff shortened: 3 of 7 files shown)


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3869,6 +3869,17 @@ private static AtomicLong getAtomicLong(String key) {
...
+    List result = new ArrayList();
+    for (int i = 0; i < list.size(); i++) {
+      if (list.get(i) != null) {
+        result.add(list.get(i));
+      }
+    }
+    return result;
+  }
+
   /** Support the ARRAY_DISTINCT function. */
```
liuyongvs, line 3883: the reason why i don't use this because
1) the spark array(null) throw exception java.lang.UnsupportedOperationException: remove
    because the spark array implementation using Arrays.asList(); which make the exception happends
2) the standard array[null] pass
```
list.removeIf(Objects::isNull);
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5328,6 +5328,25 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f.checkScalar("array_compact(array[null, 1, null, 2])", "[1, 2]",
+        "INTEGER ARRAY NOT NULL");
+    f.checkScalar("array_compact(array[null])", "[]",
+        "NULL ARRAY NOT NULL");
+    f.checkScalar("array_compact(array(null))", "[]",
+        "NULL ARRAY NOT NULL");
+    f.checkScalar("array_compact(array())", "[]",
+        "UNKNOWN NOT NULL ARRAY NOT NULL");
+    f.checkNull("array_compact(null)");
+  }
```
liuyongvs, line 5348: array() which only supports spark array constructor, while standard not.

liuyongvs:
hi @rubenada do you have time to help review? thanks

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: Is there a reason to iterate by index rather than something like `for(Object obj : list)` ? Could help with readability a bit. Just an idea.

liuyongvs: fixed @tanclary

liuyongvs:
hi @tanclary fix your reviews, do you have time to look it again

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
```diff
@@ -628,6 +628,7 @@ public enum BuiltInMethod {
   IS_EMPTY(Collection.class, "isEmpty"),
   SUBMULTISET_OF(SqlFunctions.class, "submultisetOf", Collection.class,
       Collection.class),
+  ARRAY_COMPACT(SqlFunctions.class, "removeNulls", List.class),
```
MasseGuillaume, line 631: this could also be named `compact` so it's easier to find, removeNulls is also ok.

MasseGuillaume (approved):
LGTM

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
tanclary: +1, think the BuiltInMethod name normally matches the SqlFunctions name

tanclary (commented):
Aside from the +1 on @MasseGuillaume 's suggestion, LGTM!

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
liuyongvs: fixed @MasseGuillaume @tanclary

liuyongvs:
@tanclary @MasseGuillaume fix all your reviews and squash the commits

tanclary:
LGTM, if there are no other comments or improvements I can merge soon.

liuyongvs:
the ci passed @tanclary

tanclary (approved):
About to merge, thanks for the PR @liuyongvs !

PR #3234, opened 2023-05-31
[CALCITE-5735] Add SAFE_MULTIPLY function (enabled in BigQuery)

tanclary:
Add support for SAFE_MULTIPLY in BigQuery.
SAFE_MULTIPLY: Equivalent to the multiplication operator '*', but returns NULL if overflow occurs.

Example 1: SELECT SAFE_MULTIPLY(3, 4) would return 12.
Example 2: SELECT SAFE_MULTIPLY(2^63, 2) would return NULL because of overflow.

Any comments, questions, or suggestions are welcome as always!

### DIFF
--- a/babel/src/test/resources/sql/big-query.iq
+++ b/babel/src/test/resources/sql/big-query.iq
@@ -601,6 +601,75 @@ FROM t;
 !}
 
 #####################################################################
+# SAFE_MULTIPLY
+#
+# SAFE_MULTIPLY(value1, value2)
+#
+# Equivalent to the mulitply operator (*), but returns NULL if overflow/underflow occurs.
+SELECT SAFE_MULTIPLY(5, 4) as result;
++--------+
+| result |
++--------+
+|     20 |
++--------+
+(1 row)
+
+!ok
+
+# Overflow occurs if result is greater than 2^63 - 1
+SELECT SAFE_MULTIPLY(9223372036854775807, 2) as overflow_result;
++-----------------+
+| overflow_result |
++-----------------+
+|                 |
++-----------------+
+(1 row)
+
+!ok
+
+# Underflow occurs if result is less than -2^63
+SELECT SAFE_MULTIPLY(-9223372036854775806, 3) as underflow_result;
++------------------+
+| underflow_result |
++------------------+
+|                  |
++------------------+
+(1 row)
+
+!ok
+
+SELECT SAFE_MULTIPLY(CAST(1.7e308 as DOUBLE), CAST(3 as BIGINT)) as double_overflow;
++-----------------+
+| double_overflow |
++-----------------+
+|                 |
++-----------------+
+(1 row)
+
+!ok
+
+SELECT SAFE_MULTIPLY(CAST(-3.5e75 AS DECIMAL(76, 0)), CAST(10 AS BIGINT)) as decimal_overflow;
++------------------+
+| decimal_overflow |
++------------------+
+|                  |
++------------------+
+(1 row)
+
+!ok
+
+# NaN arguments should return NaN
+SELECT SAFE_MULTIPLY(CAST('NaN' AS DOUBLE), CAST(3 as BIGINT)) as NaN_result;
++------------+
+| NaN_result |
++------------+
+|        NaN |
++------------+
+(1 row)
+
+!ok
+
+#####################################################################
 # NOT EQUAL Operator (value1 != value2)
 #
 # Compares value1 and value2 and returns TRUE if not equal.
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -216,6 +216,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.RIGHT;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.RLIKE;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.RPAD;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SAFE_CAST;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.SAFE_MULTIPLY;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SAFE_OFFSET;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SAFE_ORDINAL;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SEC;
@@ -602,6 +603,8 @@ public class RexImpTable {
       defineMethod(TANH, "tanh", NullPolicy.STRICT);
       defineMethod(TRUNC, "struncate", NullPolicy.STRICT);
       defineMethod(TRUNCATE, "struncate", NullPolicy.STRICT);
+
+      map.put(SAFE_MULTIPLY, new SafeArithmeticImplementor());
 
       map.put(PI, new PiImplementor());
       return populate2();
@@ -2365,6 +2368,30 @@ public class RexImpTable {
             dateMethod.method.getName(), operand);
       default:
         throw new AssertionError("unknown type " + type);
+      }
+    }
+  }
+
+  /** Implementor for the {@code SAFE_MULTIPLY} function. */
+  private static class SafeArithmeticImplementor extends MethodNameImplementor {
+    SafeArithmeticImplementor() {
+      super("safeMultiply", NullPolicy.STRICT, false);
+    }
+
+    @Override Expression implementSafe(final RexToLixTranslator translator,
+        final RexCall call, final List<Expression> argValueList) {
+      Expression arg0 = convertType(argValueList.get(0), call.operands.get(0));
+      Expression arg1 = convertType(argValueList.get(1), call.operands.get(1));
+      return Expressions.call(SqlFunctions.class, "safeMultiply", arg0, arg1);
+    }
+
+    // Because BigQuery treats all int types as aliases for BIGINT (Java's long)
+    // they can all be converted to LONG to minimize entries in the SqlFunctions class.
+    private Expression convertType(Expression arg, RexNode node) {
+      if (SqlTypeName.INT_TYPES.contains(node.getType().getSqlTypeName())) {
+        return Expressions.convert_(arg, long.class);
+      } else {
+        return arg;
       }
     }
   }
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -1592,6 +1592,85 @@ public class SqlFunctions {
     throw notArithmetic("*", b0, b1);
   }
 
+  /** SQL <code>SAFE_MULTIPLY</code> function applied to long values. */
+  public static @Nullable Long safeMultiply(long b0, long b1) {
+    try {
+      return Math.multiplyExact(b0, b1);
+    } catch (ArithmeticException e) {
+      return null;
+    }
+  }
+
+  /** SQL <code>SAFE_MULTIPLY</code> function applied to long and BigDecimal values. */
+  public static @Nullable BigDecimal safeMultiply(long b0, BigDecimal b1) {
+    BigDecimal ans = BigDecimal.valueOf(b0).multiply(b1);
+    return safeDecimal(ans) ? ans : null;
+  }
+
+  /** SQL <code>SAFE_MULTIPLY</code> function applied to BigDecimal and long values. */
+  public static @Nullable BigDecimal safeMultiply(BigDecimal b0, long b1) {
+    BigDecimal ans = b0.multiply(BigDecimal.valueOf(b1));
+    return safeDecimal(ans) ? ans : null;
+  }
+
+  /** SQL <code>SAFE_MULTIPLY</code> function applied to BigDecimal values. */
+  public static @Nullable BigDecimal safeMultiply(BigDecimal b0, BigDecimal b1) {
+    BigDecimal ans = b0.multiply(b1);
+    return safeDecimal(ans) ? ans : null;
+  }
+
+  /** SQL <code>SAFE_MULTIPLY</code> function applied to double and long values. */
+  public static @Nullable Double safeMultiply(double b0, long b1) {
+    double ans = b0 * b1;
+    return safeDouble(ans) || !Double.isFinite(b0) ? ans : null;
+  }
+
+  /** SQL <code>SAFE_MULTIPLY</code> function applied to long and double values. */
+  public static @Nullable Double safeMultiply(long b0, double b1) {
+    double ans = b0 * b1;
+    return safeDouble(ans) || !Double.isFinite(b1) ? ans : null;
+  }
+
+  /** SQL <code>SAFE_MULTIPLY</code> function applied to double and BigDecimal values. */
+  public static @Nullable Double safeMultiply(double b0, BigDecimal b1) {
+    double ans = b0 * b1.doubleValue();
+    return safeDouble(ans) || !Double.isFinite(b0) ? ans : null;
+  }
+
+  /** SQL <code>SAFE_MULTIPLY</code> function applied to BigDecimal and double values. */
+  public static @Nullable Double safeMultiply(BigDecimal b0, double b1) {
+    double ans = b0.doubleValue() * b1;
+    return safeDouble(ans) || !Double.isFinite(b1) ? ans : null;
+  }
+
+  /** SQL <code>SAFE_MULTIPLY</code> function applied to double values. */
+  public static @Nullable Double safeMultiply(double b0, double b1) {
+    double ans = b0 * b1;
+    boolean isFinite = Double.isFinite(b0) && Double.isFinite(b1);
+    return safeDouble(ans) || !isFinite ? ans : null;
+  }
+
+  /** Returns whether a BigDecimal value is safe (that is, has not overflowed).
+   * According to BigQuery, BigDecimal overflow occurs if the precision is greater
+   * than 76 or the scale is greater than 38. */
+  private static boolean safeDecimal(BigDecimal b) {
+    return b.scale() <= 38 && b.precision() <= 76;
+  }
+
+  /** Returns whether a double value is safe (that is, has not overflowed). */
+  private static boolean safeDouble(double d) {
+    // If the double is positive and falls between the MIN and MAX double values,
+    // overflow has not occurred. If the double is negative and falls between the
+    // negated MIN and MAX double values, overflow has not occurred. Otherwise,
+    // overflow has occurred. Important to note that 'Double.MIN_VALUE' refers to
+    // minimum positive value.
+    if (d < Double.MAX_VALUE && d > Double.MIN_VALUE) {
+      return true;
+    } else {
+      return d > -Double.MAX_VALUE && d < -Double.MIN_VALUE;
+    }
+  }
+
   private static RuntimeException notArithmetic(String op, Object b0,
       Object b1) {
     return RESOURCE.invalidTypesForArithmetic(b0.getClass().toString(),
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2779,6 +2779,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b o | RPAD(string, length[, pattern ])             | Returns a string or bytes value that consists of *string* appended to *length* with *pattern*
 | b o | RTRIM(string)                                | Returns *string* with all blanks removed from the end
 | b | SAFE_CAST(value AS type)                       | Converts *value* to *type*, returning NULL if conversion fails
+| b | SAFE_MULTIPLY(numeric1, numeric2)              | Returns *numeric1* * *numeric2*, or NULL on overflow
 | b | SAFE_OFFSET(index)                             | Similar to `OFFSET` except null is returned if *index* is out of bounds
 | b | SAFE_ORDINAL(index)                            | Similar to `OFFSET` except *index* begins at 1 and null is returned if *index* is out of bounds
 | * | SEC(numeric)                                   | Returns the secant of *numeric* in radians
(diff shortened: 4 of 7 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6111,6 +6111,31 @@ private static void checkIf(SqlOperatorFixture f) {
     f.checkNull("truncate(cast(null as double))");
   }
 
+  @Test void testSafeMultiplyFunc() {
+    final SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.SAFE_MULTIPLY);
+    // Bounds of int64
+    final String max = "9223372036854775807";
```
zoudan, line 6117: use `Long.MAX_VALUE` may be better

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6111,6 +6111,31 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f.checkScalar("safe_multiply(2, -1)", "-2", "INTEGER");
+    f.checkScalar("safe_multiply(2.5, 3.5)", "8.75", "DECIMAL(4, 2)");
+    f.checkScalar("safe_multiply(cast(2 as float), 3)", "6.0", "FLOAT");
+
+    // Test that a result larger than INTEGER but less than overflow will return BIGINT type
+    f.checkScalar("safe_multiply(1, " + max + ")", max, "BIGINT");
+    f.checkScalar("safe_multiply(1, " + min + ")", min, "BIGINT");
+
+    // Test that overflow/underflow/null operand correctly returns null
+    f.checkNull("safe_multiply(2, " + max + ")");
```
zoudan, line 6134: shall we add more test cases for overflow of other types such as safe_multiply(int b0, int b1), which should return null if the result is greater than int max.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1524,6 +1524,19 @@ public static int multiply(int b0, int b1) {
...
   }
 
+  /** Helper method for safe arithmetic functions that checks if overflow has occurred. */
+  private static boolean isOverflow(BigDecimal b0) {
+    return b0.compareTo(BigDecimal.valueOf(Long.MIN_VALUE)) < 0
+        || b0.compareTo(BigDecimal.valueOf(Long.MAX_VALUE)) > 0;
+  }
+
+  /** SQL {@code SAFE_MULTIPLY(<value0>, <value1>)} function. */
+  public static @Nullable Double safeMultiply(@PolyNull BigDecimal b0,
```
zoudan, line 1534: we do not have to use `@PolyNull` here？

zoudan, line 1534: We just have to handle BigDecimal * BigDecimal？May be we should also handle all the combination of input types?

tanclary: Will remove!

tanclary: @zoudan I think I am going to mirror my commit from [LOG function](https://github.com/apache/calcite/commit/74bc025fbb5af4f43762926e0bec75acf6ac3025). This function has the same operand type // return type relationship so should be okay. [BigQuery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#log)

tanclary:
@zoudan  I have amended my commit, if you would like to take a look. I added more robust support for all of the possible operand combinations, and I adjusted how I check for overflow as well. Feel free to leave any comments.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1515,6 +1515,7 @@ public static int multiply(int b0, int b1) {
       @PolyNull Object b1) {
     if (b0 == null || b1 == null) {
       return castNonNull(null);
+
```
jhugomoore, line 1518: nit: remove blank line

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1524,6 +1524,139 @@ public static int multiply(int b0, int b1) {
...
+    boolean overflow =
+        isOverflow(ans, BigDecimal.valueOf(Float.MAX_VALUE), BigDecimal.valueOf(-Float.MAX_VALUE));
+    return overflow ? null : ans;
+  }
+
+  /** SQL <code>SAFE_MULTIPLY</code> function applied to BigDecimal values. */
+  public static @Nullable BigDecimal safeMultiply(BigDecimal b0, BigDecimal b1) {
+    BigDecimal ans = multiply(b0, b1);
+    boolean overflow =
+        isOverflow(ans, BigDecimal.valueOf(Float.MAX_VALUE), BigDecimal.valueOf(-Float.MAX_VALUE));
```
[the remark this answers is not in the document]
julianhyde: I'm also surprised that you use  `Float` (rather than `Double`). Each SQL type should use the overflow checking for its corresponding Java data type: SMALLINT = short, INTEGER = int, BIGINT = long, REAL = float, DOUBLE = double. (`TINYINT` is tricky because it is signed whereas `byte` is unsigned.)

Note that BigQuery `FLOAT64` matches SQL `DOUBLE` and Java `double`.

I think you should have a version of `safeMultiply` for each java type, e.g. `(short, short)`, `(int, int)`, `(long, long)`, `(float, float)`, `(double, double)`. I don't think you need `BigDecimal`.

For `int` and `long` call `Math.multiplyExact` and catch the exception.

For `short`, call the `int` version and check whether the result is in range.

For `float` and `double` I don't know how you detect overflow. Maybe do ordinary multiplication for now.

julianhyde (changes requested):

tanclary:
Hey @julianhyde , just got around to fixing this. Considering that BigQuery treats all `INT` types as aliases for `INT64` (Java's `LONG`) , should we just convert all `TINYINT`, `SMALLINT` arguments etc to `LONG` and then check their result against the `LONG` bounds?

julianhyde:
@tanclary It seems like you're changing the specification (that SAFE_MULTIPLY should have arguments BIGINT, BIGINT and return BIGINT, and not support any other integer types). That's fine, and makes sense, but you should make that proposal in the Jira.

tanclary:
@julianhyde Well, sort of.

The query `SAFE_MULTIPLY(CAST(3 as TINYINT), CAST(2 AS TINYINT))` should be a valid query, but the return type is a bit tricky given the aliasing I mentioned. The overflow values are the same as Java's `LONG` so perhaps it should return a `BIGINT`? I am not sure.

julianhyde:
Two options. You can choose. Whichever you choose needs to be documented in the Jira.

If SAFE_MULTIPLY applied to two TINYINT values returns a BIGINT there's no point having a SAFE_MULTIPLY(TINYINT, TINYINT) function. You might as well call SAFE_MULTIPLY(BIGINT, BIGINT) and use implicit arg conversion.

If you want to have `SAFE_MULTIPLY(TINYINT, TINYINT)` return `TINYINT` (and similar variations for other integer types) that's fine too. Just make sure that you throw iff the result doesn't fit in a `TINYINT`. You could probably implement it by multiplying in the `int` domain.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
tanclary: I agree with both of you, `Double` makes more sense than `Float`. I've changed it.

tanclary:
@zoudan @julianhyde I have pushed a new commit that addresses several things. I have also updated the JIRA case to detail some of the decisions this PR includes.

There are 9 combinations in `SqlFunctions`, from pairings of `Long`, `BigDecimal` and `Double`. I also added checks for `Infinity` and adjusted the way I calculated overflow for `BigDecimal` and `Double` types. The testing is also more thorough now, which each combination being tested for both normal and overflow behavior.

If either of you have time to take another look, please feel free to leave more comments. Thank you!

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6114,35 +6114,62 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f.checkScalar("safe_multiply(cast(20 as bigint), cast(20 as bigint))",
+                  "400", "BIGINT");
+    f.checkScalar("safe_multiply(cast(20 as bigint), cast(1.2345 as decimal(5,4)))",
+                  "24.6900", "DECIMAL(19, 4)");
+    f.checkScalar("safe_multiply(cast(1.2345 as decimal(5,4)), cast(20 as bigint))",
+                  "24.6900", "DECIMAL(19, 4)");
+    f.checkScalar("safe_multiply(cast(1.2345 as decimal(5,4)), "
+                          + "cast(2.0 as decimal(2, 1)))", "2.46900",
+                  "DECIMAL(7, 5)");
+    f.checkScalar("safe_multiply(cast(3 as float), cast(3 as bigint))",
```
julianhyde, line 6130: Are you sure? 

SQL's `FLOAT` datatype is more similar to Java `double` than Java `float`.

In SQL, 32-bit floating point is called `REAL`.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6114,35 +6114,62 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f.checkNull("safe_multiply(cast(1.7e308 as float), cast(3 as bigint))");
+    f.checkNull("safe_multiply(cast(-1.7e308 as float), cast(3 as bigint))");
+    f.checkNull("safe_multiply(cast(3 as bigint), cast(1.7e308 as float))");
+    f.checkNull("safe_multiply(cast(3 as bigint), cast(-1.7e308 as float))");
+    f.checkNull("safe_multiply(cast(3 as float), cast(1.7e308 as float))");
+    f.checkNull("safe_multiply(cast(3 as float), cast(-1.7e308 as float))");
+    // Check that null argument retuns null
+    f.checkNull("safe_multiply(cast(null as float), cast(3 as bigint))");
+    f.checkNull("safe_multiply(cast(3 as float), cast(null as bigint))");
   }
```
julianhyde, line 6173: add blank line after `}`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6114,35 +6114,62 @@ private static void checkIf(SqlOperatorFixture f) {
...
+                        + "cast(9223372036854775807 as bigint))");
+    f.checkNull("safe_multiply(cast(20 as bigint), "
+                        + "cast(-9223372036854775807 as bigint))");
+    f.checkNull("safe_multiply(cast(10 as bigint), cast(3.5e75 as DECIMAL(76, 0)))");
+    f.checkNull("safe_multiply(cast(10 as bigint), cast(-3.5e75 as DECIMAL(76, 0)))");
+    f.checkNull("safe_multiply(cast(3.5e75 as DECIMAL(76, 0)), cast(10 as bigint))");
+    f.checkNull("safe_multiply(cast(-3.5e75 as DECIMAL(76, 0)), cast(10 as bigint))");
+    f.checkNull("safe_multiply(cast(3.5e75 as DECIMAL(76, 0)), "
+                        + "cast(1.5 as DECIMAL(2, 1)))");
+    f.checkNull("safe_multiply(cast(-3.5e75 as DECIMAL(76, 0)), "
```
julianhyde, line 6158: indent `+ "cast(1.5 as DECIMAL(2, 1)))");` only 4 spaces

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1524,136 +1524,89 @@ public static int multiply(int b0, int b1) {
...
+  public static @Nullable Double safeMultiply(double b0, double b1) {
+    double ans = b0 * b1;
+    boolean isInfinite = Double.isInfinite(b0) || Double.isInfinite(b1);
+    return doubleOverflow(ans) || isInfinite ? ans : null;
+  }
+
+  /** Helper for the safe arithmetic functions to determine
+   * overflow for a BigDecimal value. According to BigQuery, BigDecimal
+   * overflow occurs if the precision is greater than 76 or the precision
+   * is greater than 38. */
```
julianhyde, line 1592: The implementation of this function would return true if b.scale is 2 and b.precision is 5. That doesn't seem right.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: These should have been `Double` that's a mistake on my part, good catch.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: The behavior is correct, the naming is just a bit confusing. I changed it to `safeDouble` and `safeDecimal` which return `TRUE` if there is not overflow and `FALSE` if there is. Let me know if you agree with this.

tanclary:
@julianhyde Amended to address your suggestions, thanks for taking a look.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -2305,6 +2308,30 @@ private static class LastDayImplementor extends MethodNameImplementor {
...
+    @Override Expression implementSafe(final RexToLixTranslator translator,
+        final RexCall call, final List<Expression> argValueList) {
+      Expression arg0 = toLong(argValueList.get(0), call.operands.get(0));
+      Expression arg1 = toLong(argValueList.get(1), call.operands.get(1));
+      return Expressions.call(SqlFunctions.class, "safeMultiply", arg0, arg1);
+    }
+
+    // Because BigQuery treats all int types as aliases for INT64 (Java's LONG)
+    // they can all be converted to LONG to minimize entries in the SqlFunctions class.
+    private Expression toLong(Expression arg, RexNode node) {
```
zoudan, line 2326: As we only convert arg to long as when it is int type, but return the origin `arg` for other cases, it would be better to use a more proper method name to describe what this method doas.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1524,6 +1524,91 @@ public static int multiply(int b0, int b1) {
...
+  private static boolean safeDecimal(BigDecimal b) {
+    return b.scale() <= 38 && b.precision() <= 76;
+  }
+
+  /** Helper for the safe arithmetic functions to determine
+   * overflow for a double. */
+  private static boolean safeDecimal(double d) {
```
zoudan, line 1599: It would be better to use `safeDouble` here?

babel/src/test/resources/sql/big-query.iq
```diff
@@ -600,6 +600,44 @@ FROM t;
...
+SELECT SAFE_MULTIPLY(-9223372036854775806, 3) as underflow_result;
++------------------+
+| underflow_result |
++------------------+
+|                  |
++------------------+
+(1 row)
+
+!ok
+
```
zoudan, line 640: should we add overflow cases for decimal and double?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
olivrlee: Is that sentence correct?
`if the precision is greater than 76 or the precision is greater than 38.`

Should one of them be scale?

babel/src/test/resources/sql/big-query.iq
tanclary: Done.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: Good call. Done.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
tanclary: Good point. Changed to `convertType`, let me know if this works for you.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: Oops, good catch. Fixed now.

julianhyde:
If I were to make `safeDouble` always return true, would any tests fail?

If I wrote
```
boolean safeBoolean(boolean b) {
  return b == true || b == false;
}
```
it's obvious that it will always return true. But I don't see how `safeDouble` is any different, because you've already dealt with inf and -inf. Maybe you just need to check for NaN. `java.lang.Double` has methods for that kind of thing.

tanclary:
> If I were to make `safeDouble` always return true, would any tests fail?
> 
> If I wrote
> 
> ```
> boolean safeBoolean(boolean b) {
>   return b == true || b == false;
> }
> ```
> 
> it's obvious that it will always return true. But I don't see how `safeDouble` is any different, because you've already dealt with inf and -inf. Maybe you just need to check for NaN. `java.lang.Double` has methods for that kind of thing.

The answer to your first question is yes. Following BQ behavior, multiplying two non-infinite numbers whose result is > ~1.7e308 should return `NULL`. However, if either operand is infinite, the result should not be `NULL`. To match this, I check if either number is +/- inf, and then continue accordingly.

For example, `(+inf * 1)` and `(1.7e308 * 2)` should return different results (the first: `+inf`, the second: `NULL`) despite the fact that Java would tell you the answer to both is `+inf`.

You are right though that I need to address `NaN`.

tanclary:
@zoudan @julianhyde I modified the checks s.t. if either operand is non-finite (infinite or NaN), the result should be returned. I added some tests as well.

zoudan:
LGTM

tanclary:
@julianhyde Do you have any more comments before I merge this?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6843,6 +6843,79 @@ private static void checkIf(SqlOperatorFixture f) {
...
 
+  @Test void testSafeMultiplyFunc() {
+    final SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.SAFE_MULTIPLY);
+    f0.checkFails("^safe_multiply(2, 3)^",
+        "No match found for function signature "
+        + "SAFE_MULTIPLY\\(<NUMERIC>, <NUMERIC>\\)", false);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.BIG_QUERY);
+    // Basic test for each of the 9 2-permutations of BIGINT, DECIMAL, and FLOAT
+    f.checkScalar("safe_multiply(cast(20 as bigint), cast(20 as bigint))",
+                  "400", "BIGINT");
```
julianhyde, line 6854: method arguments should be indented 4, not aligned with open paren. please fix throughout this method

site/_docs/reference.md
```diff
@@ -2779,6 +2779,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b o | RPAD(string, length[, pattern ])             | Returns a string or bytes value that consists of *string* appended to *length* with *pattern*
 | b o | RTRIM(string)                                | Returns *string* with all blanks removed from the end
 | b | SAFE_CAST(value AS type)                       | Converts *value* to *type*, returning NULL if conversion fails
+| b | SAFE_MULTIPLY(numeric1, numeric2)              | Returns *numeric1* * *numeric2*, null is returned if overflow occurs
```
julianhyde, line 2782: passive voice should be avoided

don't use comma where grammar requires a semicolon

I would write "Returns *numeric1* * *numeric2*, or null on overflow"

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1592,6 +1592,87 @@ public static int multiply(int b0, int b1) {
...
+
+  /** Helper for the safe arithmetic functions to determine
+   * overflow for a BigDecimal value. According to BigQuery, BigDecimal
+   * overflow occurs if the precision is greater than 76 or the scale
+   * is greater than 38. */
+  private static boolean safeDecimal(BigDecimal b) {
+    return b.scale() <= 38 && b.precision() <= 76;
+  }
+
+  /** Helper for the safe arithmetic functions to determine
```
julianhyde, line 1661: Comment is still misleading. The method doesn't 'determine overflow'. It determines that the result has *not* overflowed.

Keep it simple. 'Returns whether a double value is safe (that is, has not overflowed).'

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -2369,6 +2372,30 @@ private static class LastDayImplementor extends MethodNameImplementor {
...
+    }
+
+    @Override Expression implementSafe(final RexToLixTranslator translator,
+        final RexCall call, final List<Expression> argValueList) {
+      Expression arg0 = convertType(argValueList.get(0), call.operands.get(0));
+      Expression arg1 = convertType(argValueList.get(1), call.operands.get(1));
+      return Expressions.call(SqlFunctions.class, "safeMultiply", arg0, arg1);
+    }
+
+    // Because BigQuery treats all int types as aliases for BIGINT (Java's LONG)
```
julianhyde, line 2388: s/LONG/long/

(Uppercase type names connote SQL types.)

julianhyde (changes requested):
I suggested a few cosmetic changes but +1 to merge without further review when they are done.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: Done

site/_docs/reference.md
tanclary: Done

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: Done

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
tanclary: Done

PR #3237, opened 2023-06-02
[CALCITE-5728] Add ARRAY_TO_STRING function (enabled in BigQuery library)

zoudan:


zoudan:
@gf2121 It seems you have push something that has nothing to do with this issue, I do not known what happened.

zoudan:
@rubenada It seems a unrelated commit has been pushed to my branch. Do you know what happened? Should I force push to my branch to recover it?

rubenada:
I think there was a force push on Calcite main, and somehow you got caught up in between. I can see in this PR the commit 295df907d126ffa059756e2e0cc8b69051ff6da7 , which was my original commit to fix CALCITE-5730, but now it is detached (and the commit to fix that issue is now e34caf70ac97996035b8fd8fc494a64022b07f91 ).
Perhaps the simplest would be to close this PR and start a new one from scratch, using a new branch (making sure that you have the main branch up to date).

zoudan:
@rubenada Maybe this could be resolved after I update my branch and force push it?

rubenada:
@zoudan update your branch how? What git commands do you intend to execute exactly?
I'm afraid the problem is that [your branch contains the commit](https://github.com/zoudan/calcite/commits/CALCITE-5728) that was detached from Calcite master after the force-push (295df907d126ffa059756e2e0cc8b69051ff6da7), I'm not sure what's the best way to proceed in this case

zoudan:
@rubenada This problem has been resolved after I rebase my branch and force push it, Thanks. Btw, it would be great if you could help review this PR

rubenada:
Glad to hear that @zoudan !
I'll take a look at the PR when I have some time.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -937,6 +937,14 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
           ReturnTypes.LEAST_RESTRICTIVE,
           OperandTypes.AT_LEAST_ONE_SAME_VARIADIC);
 
+  /** The "ARRAY_TO_STRING(array, delimiter [, nullText ])" function. */
+  @LibraryOperator(libraries = {BIG_QUERY})
+  public static final SqlFunction ARRAY_TO_STRING =
+      SqlBasicFunction.create(SqlKind.ARRAY_TO_STRING,
+          ReturnTypes.explicit(SqlTypeName.VARCHAR)
```
rubenada, line 944: There are already 5 places (this would be the 6th) in SqlLibraryOperators (and another one in SqlRegexpReplaceFunction) that use the code
`ReturnTypes.explicit(SqlTypeName.VARCHAR).andThen(SqlTypeTransforms.TO_NULLABLE)`

Why don't we just create in `ReturnTypes`:
```
public static final SqlReturnTypeInference VARCHAR_NULLABLE = explicit(SqlTypeName.VARCHAR).andThen(SqlTypeTransforms.TO_NULLABLE)
```
and use simply `ReturnTypes.VARCHAR_NULLABLE` in all those places (include the new ARRAY_TO_STRING definition)?

site/_docs/reference.md
```diff
@@ -2661,6 +2661,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | s | ARRAY_REPEAT(element, count)                   | Returns the array containing element count times.
 | b | ARRAY_REVERSE(array)                           | Reverses elements of *array*
 | s | ARRAY_SIZE(array)                              | Synonym for `CARDINALITY`
+| b | ARRAY_TO_STRING(array, delimiter [, nullText ])| Returns a concatenation of the elements in *array* as a STRING and take *delimiter* as the delimiter. If the *nullText* parameter is used, the function replaces any `NULL` values in the array with the value of *nullText*. If the *nullText* parameter is not used, the function omits the `NULL` value and its preceding delimiter.
```
rubenada, line 2664: Nitpick:
```
and take *delimiter* as the delimiter ...
=>
and takes *delimiter* as the delimiter ...
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5468,6 +5468,27 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f.checkScalar("array_to_string(array['aa', 'b', 'c'], '-')", "aa-b-c",
+        "VARCHAR NOT NULL");
+    f.checkScalar("array_to_string(array[null, 'aa', null, 'b', null], '-', 'empty')",
+        "empty-aa-empty-b-empty", "VARCHAR NOT NULL");
+    f.checkScalar("array_to_string(array[null, 'aa', null, 'b', null], '-')", "aa-b",
+        "VARCHAR NOT NULL");
+    f.checkScalar("array_to_string(array[null, x'aa', null, x'bb', null], '-')", "aa-bb",
+        "VARCHAR NOT NULL");
+    f.checkNull("array_to_string(null, '-')");
+    f.checkNull("array_to_string(array['a', 'b', null], null)");
```
rubenada, line 5487: Is this expected? Having a `null` delimiter shall lead to `null` result?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4080,6 +4080,46 @@ public static List reverse(List list) {
...
+  public static String arrayToString(List list, String delimiter) {
+    return arrayToString(list, delimiter, null);
+  }
+
+  /** SQL {@code ARRAY_TO_STRING(array, delimiter, nullText)} function. */
+  public static String arrayToString(List list, String delimiter, @Nullable String nullText) {
+    StringBuilder sb = new StringBuilder();
+    int size = list.size();
+    for (int i = 0; i < size; i++) {
+      boolean shouldAddDelimiter = i != 0 && sb.length() > 0;
```
rubenada, line 4093: Would it be enough to just use this?
```
boolean addDelimiter = sb.length() > 0;
```
In which case, we could remove the `size` and for `i`  and simply do:
```
for (Object item : list) {
```
Would that work?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4080,6 +4080,46 @@ public static List reverse(List list) {
...
+        str = null;
+      } else if (item instanceof String) {
+        str = (String) item;
+      } else if (item instanceof ByteString) {
+        str = item.toString();
+      } else {
+        throw new IllegalStateException(
+            "Only String or ByteString are supported but get " + item.getClass().getName());
+      }
+      if (str == null) {
```
rubenada, line 4106: Perhaps we can simplify this code and avoid duplication by doing something like
```
if (str == null) {
  if (nullText == null) {
    continue;
  }
  else {
    str = nullText;
  }
}

if (addDelimiter) {
  sb.append(delimiter);
}
sb.append(str);
```
Wdyt?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4080,6 +4080,46 @@ public static List reverse(List list) {
...
+      String str;
+      if (item == null) {
+        str = null;
+      } else if (item instanceof String) {
+        str = (String) item;
+      } else if (item instanceof ByteString) {
+        str = item.toString();
+      } else {
+        throw new IllegalStateException(
+            "Only String or ByteString are supported but get " + item.getClass().getName());
```
rubenada, line 4104: Nitpick:
```
but get ...
=>
but got ...
```

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
zoudan: good idea

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
zoudan: Yes, this is the expected behavior.

zoudan:
@rubenada Thanks for your valuable advice, I have updated my branch, please take a look when you have time.

rubenada:
Thanks @zoudan . There is one check style violation (Line longer than 100 characters), could you please fix it?

zoudan:
@rubenada Of course, updated

zoudan:
@rubenada updated, please have a look when you have time

rubenada:
@zoudan it seems some recent commits on Calcite master led to conflicts here, could you please resolve them and squash commits into a single one? I'll proceed with the merge afterwards.

zoudan:
@rubenada updated

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4102,6 +4102,43 @@ public static List reverse(List list) {
...
+      }
+      if (str == null) {
+        if (nullText == null) {
+          continue;
+        } else {
+          str = nullText;
+        }
+      }
+
+      if (sb.length() > 0) {
```
rubenada, line 4134: Sorry @zoudan , one last detail (rather an edge case): what is the expected behavior if the item (`str` in this point) is the empty string? Will it be appended like any other value with the corresponding delimiter before the precious value? 
For example:
```
array_to_string(array['', 'b'], '-') => '-b'    ???
array_to_string(array['', ''], '-') => '-'    ???
```
If that is the case, I have the impression that this condition is wrong and we would need to find another mechanism, e.g.
```
boolean first = true;
for (Object item : list) {
  ...
  if (!first) {
    sb.append(delimiter);
  }
  sb.append(str);
  first = false;
} 
```
wdyt?

zoudan: @rubenada Thank you for your careful observation. Yes, this is the expected behavior:
`array_to_string(array['', 'b'], '-') => '-b'`
`array_to_string(array['', ''], '-') => '-' `
I have updated my code and also squashed the commits.

rubenada: Thanks @zoudan ! I saw that you also added some unit tests for this scenario, that's great.
I'll merge as soon as the CI checks are finished

site/_docs/reference.md
```diff
@@ -2663,6 +2663,7 @@ BigQuery's type system uses confusingly different names for types and functions:
...
```
liuyongvs, line 2666: remove end with '.'

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5474,6 +5474,29 @@ private static void checkIf(SqlOperatorFixture f) {
...
+        "VARCHAR NOT NULL");
+    f.checkScalar("array_to_string(array[null, x'aa', null, x'bb', null], '-')", "aa-bb",
+        "VARCHAR NOT NULL");
+    f.checkScalar("array_to_string(array['', 'b'], '-')", "-b", "VARCHAR NOT NULL");
+    f.checkScalar("array_to_string(array['', ''], '-')", "-", "VARCHAR NOT NULL");
+    f.checkNull("array_to_string(null, '-')");
+    f.checkNull("array_to_string(array['a', 'b', null], null)");
+    f.checkFails("array_to_string(array[1, 2, 3], '-')",
+        "arrayToString supports only String or ByteString, but got java.lang.Integer", true);
+  }
```
liuyongvs, line 5498: sorry for late review, it may not be right here "arrayToString supports only String or ByteString, but got java.lang.Integer", it should throw exception in validate type process, instead of runtime process

zoudan: good idea

zoudan:
@liuyongvs Thanks for you comments, I have updated my code, please take a look. cc @rubenada

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -471,6 +471,57 @@ public static SqlOperandTypeChecker variadic(
           .or(OperandTypes.family(SqlTypeFamily.MAP))
           .or(OperandTypes.family(SqlTypeFamily.ANY));
 
+  public static final SqlOperandTypeChecker ARRAY_OF_STRING_OR_BINARY_STRING_STRING_OPTIONAL_STRING
```
rubenada, line 474: I'd suggst to try to reuse the FamilyOperandTypeChecker (adding just the relevant code to check the array component type). Proposal (unverified):
```
  public static final SqlOperandTypeChecker STRING_ARRAY_OR_BINARY_ARRAY_STRING_OPTIONAL_STRING =
      new FamilyOperandTypeChecker(
          ImmutableList.of(SqlTypeFamily.ARRAY, SqlTypeFamily.STRING, SqlTypeFamily.STRING),
          i -> i == 2) {
            SqlCallBinding callBinding,
            boolean throwOnFailure) {
          if (!super.checkOperandTypes(callBinding, throwOnFailure)) {
            return false;
          }
          RelDataType arrayType = callBinding.getOperandType(0).getComponentType();
          if (arrayType == null || (!SqlTypeFamily.STRING.contains(arrayType)
              && !SqlTypeFamily.BINARY.contains(arrayType))) {
            if (throwOnFailure) {
              throw callBinding.newValidationSignatureError();
            }
            return false;
          }
          return true;
        }
      };
```

Regarding the checker framework error, if you cannot get rid of it, you can use `@SuppressWarnings("argument.type.incompatible")` .
If I am not mistaken, you should be able to run it locally in your dev environment via
`gradlew build -x test -PenableCheckerframework`

zoudan: Thank you for your guidance, I have updated my code

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -471,6 +471,30 @@ public static SqlOperandTypeChecker variadic(
...
+          RelDataType elementType = callBinding.getOperandType(0).getComponentType();
+          if (elementType == null || (!SqlTypeFamily.STRING.contains(elementType)
+              && !SqlTypeFamily.BINARY.contains(elementType))) {
+            if (throwOnFailure) {
+              throw callBinding.newValidationSignatureError();
+            }
+            return false;
+          }
+          return true;
+        }
```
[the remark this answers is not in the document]
zoudan: updated

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
rubenada: @liuyongvs  what do you think about the newer version (including the validation check)?

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -427,6 +427,13 @@ public static SqlCall stripSeparator(SqlCall call) {
   public static final SqlReturnTypeInference VARCHAR_2000_NULLABLE =
       VARCHAR_2000.andThen(SqlTypeTransforms.TO_NULLABLE);
 
+  /**
+   * Type-inference strategy that always returns "VARCHAR" with nulls
+   * allowed if any of the operands allow nulls.
+   */
+  public static final SqlReturnTypeInference VARCHAR_NULLABLE =
```
JiajunBernoulli, line 434: Can we declare another `VARCHAR`?

Because
`VARCHAR_4` -> `VARCHAR_4_NULLABLE`;
`VARCHAR_2000` -> `VARCHAR_2000_NULLABLE`;

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4102,6 +4102,45 @@ public static List reverse(List list) {
     return list;
   }
 
+  /** SQL {@code ARRAY_TO_STRING(array, delimiter)} function. */
+  public static String arrayToString(List list, String delimiter) {
+    return arrayToString(list, delimiter, null);
+  }
+
+  /** SQL {@code ARRAY_TO_STRING(array, delimiter, nullText)} function. */
+  public static String arrayToString(List list, String delimiter, @Nullable String nullText) {
```
JiajunBernoulli, line 4111: Maybe `stream` is more simplify.
```
    list.stream()
        .map(item -> {
          if (item == null) {
            return nullText;
          } else if (item instanceof String) {
            return (String) item;
          } else if (item instanceof ByteString) {
            return item.toString();
          } else {
            throw new IllegalStateException(
                "arrayToString supports only String or ByteString, but got "
                    + item.getClass().getName());
          }
        }).collect(Collectors.joining(delimiter));
  }
```

zoudan: Considering performance, I tend to not use lambda. WDYT?

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
zoudan: Thanks for reviewing, updated

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -471,6 +471,34 @@ public static SqlOperandTypeChecker variadic(
...
+        @Override public boolean checkOperandTypes(
+            SqlCallBinding callBinding,
+            boolean throwOnFailure) {
+          if (!super.checkOperandTypes(callBinding, throwOnFailure)) {
+            return false;
+          }
+          RelDataType elementType = callBinding.getOperandType(0).getComponentType();
+          if (elementType == null || (!SqlTypeFamily.STRING.contains(elementType)
+              && !SqlTypeFamily.BINARY.contains(elementType))) {
+            if (throwOnFailure) {
```
liuyongvs, line 488: if i understand correct, it can use SqlTypeUtil.isString.
string familiy has 4 type
  public static final List<SqlTypeName> CHAR_TYPES =
      ImmutableList.of(CHAR, VARCHAR);
  public static final List<SqlTypeName> BINARY_TYPES =
      ImmutableList.of(BINARY, VARBINARY);

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -471,6 +471,34 @@ public static SqlOperandTypeChecker variadic(
           .or(OperandTypes.family(SqlTypeFamily.MAP))
           .or(OperandTypes.family(SqlTypeFamily.ANY));
 
+  public static final SqlOperandTypeChecker STRING_OR_BINARY_ARRAY_STRING_OPTIONAL_STRING =
+      new FamilyOperandTypeChecker(
```
liuyongvs, line 475: personaly, the name STRING_OR_BINARY_ARRAY_STRING_OPTIONAL_STRING is not good. because the string contains BINARY family's type

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
zoudan: make sense, updated

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
liuyongvs: @rubenada i have responsed

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
JiajunBernoulli: Ok

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
liuyongvs: @rubenada it can not be. refers to the bigquery's doc https://cloud.google.com/bigquery/docs/reference/standard-sql/array_functions#array_to_string, the array's element type should be check in compiler process.

zoudan: @liuyongvs, If I understand correctly, what @rubenada meant was we should use `SqlTypeFamily.CHARACTER` instead of `SqlTypeFamily.STRING` for the second and third parameters. That make sense, I will updated it

liuyongvs: ok, i understand, thanks @zoudan

rubenada:
@liuyongvs @liuyongvs thanks for your comments in the PR, do you have further remarks? Shall we consider squash commits and merge?

liuyongvs (approved):
LGTM

zoudan:
> @liuyongvs @JiajunBernoulli thanks for your comments in the PR, do you have further remarks? Shall we consider squash commits and merge?

I have squashed the commits.

JiajunBernoulli (approved):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4102,6 +4102,45 @@ public static List reverse(List list) {
...
+      } else if (item instanceof String) {
+        str = (String) item;
+      } else if (item instanceof ByteString) {
+        str = item.toString();
+      } else {
+        throw new IllegalStateException(
+            "arrayToString supports only String or ByteString, but got "
+                + item.getClass().getName());
+      }
+      if (str == null) {
```
rubenada, line 4127: sorry @zoudan , just one last thing before the merge. I realized we could remove this `if` block and move its content to the first `if`, in order to make the code more compact:
```
for (Object item : list) {
  String str;
  if (item == null) {
    if (nullText == null) {
      continue;
    } else {
      str = nullText;
    }
  } else if (item instanceof String) { 
  ...
```
wdyt?

rubenada:
Thanks @zoudan , I will merge as soon as the CI checks are finished.

PR #3238, opened 2023-06-02
[CALCITE-5744] Add MAP_FROM_ARRAYS, STR_TO_MAP function (enabled in Spark library)

liuyongvs:
STR_TO_MAP(string[, stringDelimiter[, keyValueDelimiter]])

Returns a map after splitting the string into key/value pairs using delimiters. Default delimiters are ',' for stringDelimiter and ':' for keyValueDelimiter Both pairDelim and keyValueDelim are treated as regular expressions.

```
Examples:

> SELECT str_to_map('a:1,b:2,c:3', ',', ':');

- {"a":"1","b":"2","c":"3"}

> SELECT str_to_map('a');
 {"a":null} 
```

MAP_FROM_ARRAYS (array1, array2)

Returns a map created from an array1 and *array2. Note that the lengths of two arrays should be the same

> SELECT map_from_arrays(array(1.0, 3.0), array('2', '4'));
 {1.0:"2",3.0:"4"}

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -181,6 +181,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_AND;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_OR;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LPAD;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_ENTRIES;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_FROM_ARRAYS;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_KEYS;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_VALUES;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAX_BY;
@@ -215,6 +216,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.SPACE;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SPLIT;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.STARTS_WITH;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.STRCMP;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.STR_TO_MAP;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.TANH;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.TIME;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.TIMESTAMP;
@@ -428,6 +430,8 @@ public class RexImpTable {
       Expressions.constant(true);
   public static final ConstantExpression COMMA_EXPR =
       Expressions.constant(",");
+  public static final ConstantExpression COLON_EXPR =
+      Expressions.constant(":");
   public static final MemberExpression BOXED_FALSE_EXPR =
       Expressions.field(null, Boolean.class, "FALSE");
   public static final MemberExpression BOXED_TRUE_EXPR =
@@ -714,6 +718,8 @@ public class RexImpTable {
       defineMethod(MAP_ENTRIES, BuiltInMethod.MAP_ENTRIES.method, NullPolicy.STRICT);
       defineMethod(MAP_KEYS, BuiltInMethod.MAP_KEYS.method, NullPolicy.STRICT);
       defineMethod(MAP_VALUES, BuiltInMethod.MAP_VALUES.method, NullPolicy.STRICT);
+      defineMethod(MAP_FROM_ARRAYS, BuiltInMethod.MAP_FROM_ARRAYS.method, NullPolicy.ANY);
+      map.put(STR_TO_MAP, new StringToMapImplementor());
       map.put(ARRAY_CONCAT, new ArrayConcatImplementor());
       map.put(SORT_ARRAY, new SortArrayImplementor());
       final MethodImplementor isEmptyImplementor =
@@ -3136,6 +3142,37 @@ public class RexImpTable {
                         BuiltInMethod.COLLECTION_ADDALL.method, expression))));
       }
       return list;
+    }
+  }
+
+  /** Implementor for str_to_map. */
+  private static class StringToMapImplementor extends AbstractRexCallImplementor {
+    StringToMapImplementor() {
+      super("str_to_map", NullPolicy.STRICT, false);
+    }
+
+    @Override Expression implementSafe(RexToLixTranslator translator,
+        RexCall call, List<Expression> argValueList) {
+      switch (call.getOperands().size()) {
+      case 1:
+        return Expressions.call(
+            BuiltInMethod.STR_TO_MAP.method,
+            argValueList.get(0),
+            COMMA_EXPR,
+            COLON_EXPR);
+      case 2:
+        return Expressions.call(
+            BuiltInMethod.STR_TO_MAP.method,
+            argValueList.get(0),
+            argValueList.get(1),
+            COLON_EXPR);
+      case 3:
+        return Expressions.call(
+            BuiltInMethod.STR_TO_MAP.method,
+            argValueList);
+      default:
+        throw new AssertionError();
+      }
     }
   }
 
--- a/core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
+++ b/core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
@@ -897,6 +897,9 @@ public interface CalciteResource {
   @BaseMessage("Substring error: negative substring length not allowed")
   ExInst<CalciteException> illegalNegativeSubstringLength();
 
+  @BaseMessage("Illegal arguments: The length of the keys array {0,number,#} is not equal to the length of the values array {1,number,#} in MAP_FROM_ARRAYS function")
+  ExInst<CalciteException> illegalArgumentsInMapFromArraysFunc(int arg0, int arg1);
+
   @BaseMessage("Trim error: trim character must be exactly 1 character")
   ExInst<CalciteException> trimError();
 
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -92,6 +92,7 @@ import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
@@ -4001,6 +4002,33 @@ public class SqlFunctions {
   /** Support the MAP_VALUES function. */
   public static List mapValues(Map map) {
     return new ArrayList<>(map.values());
+  }
+
+  /** Support the MAP_FROM_ARRAYS function. */
+  public static Map mapFromArrays(List keysArray, List valuesArray) {
+    if (keysArray.size() != valuesArray.size()) {
+      throw RESOURCE.illegalArgumentsInMapFromArraysFunc(keysArray.size(), valuesArray.size()).ex();
+    }
+    final Map map = new LinkedHashMap<>();
+    for (int i = 0; i < keysArray.size(); i++) {
+      map.put(keysArray.get(i), valuesArray.get(i));
+    }
+    return map;
+  }
+
+  /** Support the STR_TO_MAP function. */
+  public static Map strToMap(String string, String stringDelimiter, String keyValueDelimiter) {
+    final Map map = new LinkedHashMap();
+    final String[] keyValues = string.split(stringDelimiter, -1);
+    for (String s : keyValues) {
+      String[] keyValueArray = s.split(keyValueDelimiter, 2);
+      String key = keyValueArray[0];
+      String value = keyValueArray.length < 2
+          ? null
+          : keyValueArray[1];
+      map.put(key, value);
+    }
+    return map;
   }
 
   /** Support the SLICE function. */
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -1041,6 +1041,31 @@ public abstract class SqlLibraryOperators {
           ReturnTypes.TO_MAP_VALUES_NULLABLE,
           OperandTypes.MAP);
 
+  private static RelDataType deriveTypeMapFromArrays(SqlOperatorBinding opBinding) {
+    final RelDataType keysArrayType = opBinding.getOperandType(0);
+    final RelDataType valuesArrayType = opBinding.getOperandType(1);
+    final boolean nullable = keysArrayType.isNullable() || valuesArrayType.isNullable();
+    return SqlTypeUtil.createMapType(
+        opBinding.getTypeFactory(),
+        requireNonNull(keysArrayType.getComponentType(), "inferred key type"),
+        requireNonNull(valuesArrayType.getComponentType(), "inferred value type"),
+        nullable);
+  }
+
+  /** The "MAP_FROM_ARRAYS(keysArray, valuesArray)" function. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction MAP_FROM_ARRAYS =
+      SqlBasicFunction.create(SqlKind.MAP_FROM_ARRAYS,
+          SqlLibraryOperators::deriveTypeMapFromArrays,
+          OperandTypes.ARRAY_ARRAY);
+
+  /** The "STR_TO_MAP(string[, stringDelimiter[, keyValueDelimiter]])" function. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction STR_TO_MAP =
+      SqlBasicFunction.create(SqlKind.STR_TO_MAP,
+          ReturnTypes.IDENTITY_TO_MAP_NULLABLE,
+          OperandTypes.STRING_OPTIONAL_STRING_OPTIONAL_STRING);
+
   @LibraryOperator(libraries = {BIG_QUERY, MYSQL})
   public static final SqlFunction REVERSE =
       SqlBasicFunction.create(SqlKind.REVERSE,
--- a/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
@@ -598,6 +598,22 @@ public abstract class ReturnTypes {
       ARG0.andThen(SqlTypeTransforms.TO_MAP);
 
   /**
+   * Returns a MAP type.
+   *
+   * <p>For example, given {@code STRING}, returns
+   * {@code (STRING, STRING) MAP}.
+   */
+  public static final SqlReturnTypeInference IDENTITY_TO_MAP =
+      ARG0.andThen(SqlTypeTransforms.IDENTITY_TO_MAP);
+
+  /**
+   * Same as {@link #IDENTITY_TO_MAP} but returns with nullability if any of the
+   * operands is nullable.
+   */
+  public static final SqlReturnTypeInference IDENTITY_TO_MAP_NULLABLE =
+      IDENTITY_TO_MAP.andThen(SqlTypeTransforms.TO_NULLABLE);
+
+  /**
    * Returns a ROW type.
    *
    * <p>For example, given {@code (INTEGER, DATE) MAP}, returns
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2743,6 +2743,8 @@ BigQuery's type system uses confusingly different names for types and functions:
 | s | MAP_ENTRIES(map)                               | Returns the entries of the *map* as an array, the order of the entries is not defined
 | s | MAP_KEYS(map)                                  | Returns the keys of the *map* as an array, the order of the entries is not defined
 | s | MAP_VALUES(map)                                | Returns the values of the *map* as an array, the order of the entries is not defined
+| s | MAP_FROM_ARRAYS(array1, array2)                | Returns a map created from an *array1* and *array2*. Note that the lengths of two arrays should be the same
+| s | STR_TO_MAP(string [, stringDelimiter [, keyValueDelimiter]]) | Returns a map after splitting the *string* into key/value pairs using delimiters. Default delimiters are ',' for *stringDelimiter* and ':' for *keyValueDelimiter*
 | b m p | MD5(string)                                | Calculates an MD5 128-bit checksum of *string* and returns it as a hex string
 | m | MONTHNAME(date)                                | Returns the name, in the connection's locale, of the month in *datetime*; for example, it returns '二月' for both DATE '2020-02-10' and TIMESTAMP '2020-02-10 10:10:10'
 | o | NVL(value1, value2)                            | Returns *value1* if *value1* is not null, otherwise *value2*
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -5688,6 +5688,82 @@ public class SqlOperatorTest {
         "INTEGER ARRAY NOT NULL");
   }
 
+  /** Tests {@code MAP_FROM_ARRAYS} function from Spark. */
+  @Test void testMapFromArraysFunc() {
+    final SqlOperatorFixture f0 = fixture();
+    f0.setFor(SqlLibraryOperators.MAP_FROM_ARRAYS);
+    f0.checkFails("^map_from_arrays(array[1, 2], array['foo', 'bar'])^",
+        "No match found for function signature MAP_FROM_ARRAYS\\(<INTEGER ARRAY>, "
+            + "<CHAR\\(3\\) ARRAY>\\)", false);
+
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
+    f.checkScalar("map_from_arrays(array[1, 2], array['foo', 'bar'])", "{1=foo, 2=bar}",
+        "(INTEGER NOT NULL, CHAR(3) NOT NULL) MAP NOT NULL");
+    f.checkScalar("map_from_arrays(array[1, 1, null], array['foo', 'bar', 'name'])",
+        "{1=bar, null=name}", "(INTEGER, CHAR(4) NOT NULL) MAP NOT NULL");
+    f.checkScalar("map_from_arrays(array(), array())",
+        "{}", "(UNKNOWN NOT NULL, UNKNOWN NOT NULL) MAP NOT NULL");
+    f.checkType("map_from_arrays(cast(null as integer array), array['foo', 'bar'])",
+        "(INTEGER NOT NULL, CHAR(3) NOT NULL) MAP");
+    f.checkNull("map_from_arrays(cast(null as integer array), array['foo', 'bar'])");
+
+    f.checkFails("^map_from_arrays(array[1, 2], 2)^",
+        "Cannot apply 'MAP_FROM_ARRAYS' to arguments of type 'MAP_FROM_ARRAYS\\(<INTEGER ARRAY>,"
+            + " <INTEGER>\\)'. Supported form\\(s\\): 'MAP_FROM_ARRAYS\\(<ARRAY>, <ARRAY>\\)'",
+        false);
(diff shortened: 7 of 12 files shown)


liuyongvs:
hi @JiajunBernoulli @tanclary @MasseGuillaume  do you have time to review this ,just review STR_TO_MAP and after
the former commits merged, i will rebase it.otherwise, it will lots of conflict

liuyongvs:
hi @JiajunBernoulli close the map_from_array pr , and one pr for the two commits

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -944,6 +944,31 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
...
+  private static RelDataType mapReturnType(SqlOperatorBinding opBinding) {
+    final RelDataType keysArrayType = opBinding.collectOperandTypes().get(0);
+    final RelDataType valuesArrayType = opBinding.collectOperandTypes().get(1);
+    final boolean nullable = keysArrayType.isNullable() || valuesArrayType.isNullable();
+    return SqlTypeUtil.createMapType(
+        opBinding.getTypeFactory(),
+        requireNonNull(keysArrayType.getComponentType(), "inferred key type"),
+        requireNonNull(valuesArrayType.getComponentType(), "inferred value type"),
+        nullable);
+  }
```
liuyongvs, line 956: comments:
        final boolean nullable = keysArrayType.isNullable() || valuesArrayType.isNullable();
 this should be ,so i add the unit test
    f.checkType("str_to_map(cast(null as varchar))",
        "(VARCHAR, VARCHAR) MAP");

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3917,6 +3918,38 @@ public static List mapValues(Map map) {
...
+    for (int i = 0; i < keysArray.size(); i++) {
+      map.put(keysArray.get(i), valuesArray.get(i));
+    }
+    return map;
+  }
+
+  /** Support the STR_TO_MAP function. */
+  public static Map strToMap(String string, String stringDelimiter, String keyValueDelimiter) {
+    final Map map = new LinkedHashMap();
+    final String[] keyValues = string.split(stringDelimiter, -1);
```
liuyongvs, line 3941: the limit should be -1, which aligns with spark.
what is the difference with limit -1, you can se here
https://stackabuse.com/how-to-split-a-string-in-java/

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3917,6 +3918,38 @@ public static List mapValues(Map map) {
     return new ArrayList<>(map.values());
   }
 
+  /** Support the MAP_FROM_ARRAYS function. */
+  public static Map mapFromArrays(List keysArray, List valuesArray) {
+    if (keysArray.size() != valuesArray.size()) {
+      throw new IllegalArgumentException(
```
liuyongvs, line 3924: the length check is need, and i add a negative test for it 
```
    f.checkFails("map_from_arrays(array[1, 2], array['foo'])",
        "Invalid function MAP_FROM_ARRAYS call:\n"
            + "The length of the keys array 2 is not equal to the length of the values array 1",
        true);
```

liuyongvs:
hi @tanclary @JiajunBernoulli @MasseGuillaume do you have time to help review ?

liuyongvs:
Hi @tanclary do you also have time to review this pr?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3092,6 +3096,36 @@ private static class ArrayConcatImplementor extends AbstractRexCallImplementor {
...
+  private static class StringToMapImplementor extends AbstractRexCallImplementor {
+    StringToMapImplementor() {
+      super("str_to_map", NullPolicy.STRICT, false);
+    }
+
+    @Override Expression implementSafe(RexToLixTranslator translator,
+        RexCall call, List<Expression> argValueList) {
+      String defaultStringDelimiter = ",";
+      String defaultKeyValueDelimiter = ":";
+      if (argValueList.size() == 1) {
```
tanclary, line 3109: i personally find switch (operand count) to look better here but it's very subjective i suppose.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3092,6 +3096,36 @@ private static class ArrayConcatImplementor extends AbstractRexCallImplementor {
...
 
+  /** Implementor for str_to_map. */
+  private static class StringToMapImplementor extends AbstractRexCallImplementor {
+    StringToMapImplementor() {
+      super("str_to_map", NullPolicy.STRICT, false);
+    }
+
+    @Override Expression implementSafe(RexToLixTranslator translator,
+        RexCall call, List<Expression> argValueList) {
+      String defaultStringDelimiter = ",";
```
tanclary, line 3107: is there a way to use the "SPLIT" implementation offered by Bigquery here? Some of this looks very similar.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3092,6 +3096,36 @@ private static class ArrayConcatImplementor extends AbstractRexCallImplementor {
...
+            BuiltInMethod.STR_TO_MAP.method,
+            argValueList.get(0),
+            Expressions.constant(defaultStringDelimiter),
+            Expressions.constant(defaultKeyValueDelimiter));
+      } else if (argValueList.size() == 2) {
+        return Expressions.call(
+          BuiltInMethod.STR_TO_MAP.method,
+              argValueList.get(0),
+              argValueList.get(1),
+              Expressions.constant(defaultKeyValueDelimiter));
```
tanclary, line 3120: if you only use the default Strings for Expressions.constant, you could just initialize them as expressions instead?

tanclary: i am wondering whether the return statement could be extracted out, and the if-else-if (or switch, if you choose to use it) could just be used for assigning operands and whatnot. Right now it looks like a lot of duplicate code.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3917,6 +3918,38 @@ public static List mapValues(Map map) {
     return new ArrayList<>(map.values());
   }
 
+  /** Support the MAP_FROM_ARRAYS function. */
```
tanclary, line 3921: is this different/unrelated to STR_TO_MAP? if so, the commit message etc should be updated to reflect everything the PR includes.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary, line 3924: can this exception be added as a constant to Resources like the other errors thrown from this class?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -944,6 +944,31 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
           ReturnTypes.TO_MAP_VALUES_NULLABLE,
           OperandTypes.MAP);
 
+  private static RelDataType mapReturnType(SqlOperatorBinding opBinding) {
```
tanclary, line 947: I am not sure this belongs here. Maybe somewhere like ReturnTypes.java would be more appropriate.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -944,6 +944,31 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
...
+    return SqlTypeUtil.createMapType(
+        opBinding.getTypeFactory(),
+        requireNonNull(keysArrayType.getComponentType(), "inferred key type"),
+        requireNonNull(valuesArrayType.getComponentType(), "inferred value type"),
+        nullable);
+  }
+
+  /** The "MAP_FROM_ARRAYS(keysArray, valuesArray)" function. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction MAP_FROM_ARRAYS =
```
tanclary, line 960: just reiterating that if this PR includes multiple functions, the JIRA case/commit message should reflect that

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -944,6 +944,31 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
...
+      SqlBasicFunction.create(SqlKind.MAP_FROM_ARRAYS,
+          SqlLibraryOperators::mapReturnType,
+          OperandTypes.ARRAY_ARRAY);
+
+  /** The "STR_TO_MAP(string[, stringDelimiter[, keyValueDelimiter]])" function. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction STR_TO_MAP =
+      SqlBasicFunction.create(SqlKind.STR_TO_MAP,
```
tanclary, line 968: I think there is an entry in OperandTypes called something like "STRING_OPTIONAL_STRING" or something along those lines, either use that or add an entry to fit this case, so future functions can use it easily if need be.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -536,6 +536,18 @@ public static SqlCall stripSeparator(SqlCall call) {
...
+  /**
+   * Returns a MAP type.
+   *
+   * <p>For example, given {@code STRING}, returns
+   * {@code (STRING, STRING) MAP}.
+   */
+  public static final SqlReturnTypeInference IDENTITY_TO_MAP =
+      ARG0.andThen(SqlTypeTransforms.IDENTITY_TO_MAP);
+
+  public static final SqlReturnTypeInference IDENTITY_TO_MAP_NULLABLE =
```
tanclary, line 548: nit: many *_NULLABLE return types have a JAVADOC

tanclary (commented):
Left some comments, hope this helps!

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
liuyongvs: fix pr info and issue description

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
liuyongvs: just reuse, not define a new OperandTypes like STRING_OPTIONAL_STRING because i don't find a good way to express it

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
liuyongvs: added

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
liuyongvs: change to switch

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
liuyongvs: but not same

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
liuyongvs: done

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
liuyongvs: use Resources

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
liuyongvs: other function also puts here like split you support in bigquery

liuyongvs:
hi @tanclary thanks for your review and your several valuable suggestion +1

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -965,6 +965,31 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
           ReturnTypes.TO_MAP_VALUES_NULLABLE,
           OperandTypes.MAP);
 
+  private static RelDataType deriveTypeMapFromArrays(SqlOperatorBinding opBinding) {
+    final RelDataType keysArrayType = opBinding.collectOperandTypes().get(0);
```
tanclary, line 969: can you do opBinding.getOperandType? if so, that's a little cleaner I think

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3119,6 +3125,37 @@ private static class ArrayConcatImplementor extends AbstractRexCallImplementor {
...
+  private static class StringToMapImplementor extends AbstractRexCallImplementor {
+    StringToMapImplementor() {
+      super("str_to_map", NullPolicy.STRICT, false);
+    }
+
+    @Override Expression implementSafe(RexToLixTranslator translator,
+        RexCall call, List<Expression> argValueList) {
+      switch (call.getOperands().size()) {
+      case 1:
+        return Expressions.call(
```
tanclary, line 3138: nit: I'm still wondering if the return can be extracted out of the switch statement. You notice that in each case it's "return Expressions.call(...)". What do you think?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5543,6 +5543,65 @@ private static void checkIf(SqlOperatorFixture f) {
...
+  /** Tests {@code MAP_FROM_ARRAYS} function from Spark. */
+  @Test void testMapFromArraysFunc() {
+    final SqlOperatorFixture f0 = fixture();
+    f0.setFor(SqlLibraryOperators.MAP_FROM_ARRAYS);
+    f0.checkFails("^map_from_arrays(array[1, 2], array['foo', 'bar'])^",
+        "No match found for function signature MAP_FROM_ARRAYS\\(<INTEGER ARRAY>, "
+            + "<CHAR\\(3\\) ARRAY>\\)", false);
+
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
+    f.checkScalar("map_from_arrays(array[1, 2], array['foo', 'bar'])", "{1=foo, 2=bar}",
```
tanclary, line 5555: do all elements of the array have to be of the same type ? could you have [1, "a", true]?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
liuyongvs: However, personally, I think the difference is not significant, and many do the same 
for example ItemImplementor/LogImplementor/DatetimeArithmeticImplementor/...

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
liuyongvs: @tanclary "do all elements of the array have to be of the same type"  yeap, the args should be find leastRestrictive, otherwise it will throw exception like "SqlValidatorException: Parameters must be of the same type"

tanclary: Ok that could be a good test, then.

liuyongvs:
@tanclary squash the commits

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
liuyongvs: @tanclary thanks for your quick response, thank you so much

liuyongvs:
@tanclary @MasseGuillaume will you help look again?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5543,6 +5543,65 @@ private static void checkIf(SqlOperatorFixture f) {
...
+        "(INTEGER NOT NULL, CHAR(3) NOT NULL) MAP NOT NULL");
+    f.checkScalar("map_from_arrays(array[1, 1, null], array['foo', 'bar', 'name'])",
+        "{1=bar, null=name}", "(INTEGER, CHAR(4) NOT NULL) MAP NOT NULL");
```
MasseGuillaume, line 5558:
```
spark.sql("""select map_from_arrays(array(1, 1, null), array('foo', 'bar', 'name'))""").show
java.lang.RuntimeException: Duplicate map key 1 was found, please check the input data. If you want to remove the duplicated keys, you can set spark.sql.mapKeyDedupPolicy to LAST_WIN so that the key inserted at last takes precedence.
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5543,6 +5543,65 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f.checkScalar("str_to_map('a:1&b:2', '&')", "{a=1, b=2}",
+        "(CHAR(7) NOT NULL, CHAR(7) NOT NULL) MAP NOT NULL");
+    f.checkScalar("str_to_map('k:2%v:3', '%')", "{k=2, v=3}",
+        "(CHAR(7) NOT NULL, CHAR(7) NOT NULL) MAP NOT NULL");
+    f.checkScalar("str_to_map('a')", "{a=null}",
+        "(CHAR(1) NOT NULL, CHAR(1) NOT NULL) MAP NOT NULL");
+    f.checkType("str_to_map(cast(null as varchar))",
+        "(VARCHAR, VARCHAR) MAP");
+    f.checkNull("str_to_map(cast(null as varchar))");
+  }
```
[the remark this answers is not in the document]
liuyongvs: added test
```
    f.checkScalar("str_to_map('a,b,c')", "{a=null, b=null, c=null}",
        "(CHAR(5) NOT NULL, CHAR(5) NOT NULL) MAP NOT NULL");
    f.checkScalar("str_to_map('a-b--c', '--')", "{a-b=null, c=null}",
        "(CHAR(6) NOT NULL, CHAR(6) NOT NULL) MAP NOT NULL");
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
liuyongvs: this is because spark's map is different from calcite is map implementation.
spark suppprts two way, while calcite is LAST_WIN. if we 100% match spark, it should break the behavior of all map functions
```
  object MapKeyDedupPolicy extends Enumeration {
    val EXCEPTION, LAST_WIN = Value
  }
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5543,6 +5543,65 @@ private static void checkIf(SqlOperatorFixture f) {
...
+        "(INTEGER NOT NULL, CHAR(3) NOT NULL) MAP NOT NULL");
+    f.checkScalar("map_from_arrays(array[1, 1, null], array['foo', 'bar', 'name'])",
+        "{1=bar, null=name}", "(INTEGER, CHAR(4) NOT NULL) MAP NOT NULL");
+    f.checkScalar("map_from_arrays(array(), array())",
+        "{}", "(UNKNOWN NOT NULL, UNKNOWN NOT NULL) MAP NOT NULL");
+    f.checkType("map_from_arrays(cast(null as integer array), array['foo', 'bar'])",
+        "(INTEGER NOT NULL, CHAR(3) NOT NULL) MAP");
+    f.checkNull("map_from_arrays(cast(null as integer array), array['foo', 'bar'])");
+
+    f.checkFails("^map_from_arrays(array[1, 2], 2)^",
```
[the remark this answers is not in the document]
liuyongvs: added

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5543,6 +5543,65 @@ private static void checkIf(SqlOperatorFixture f) {
...
+        "Cannot apply 'MAP_FROM_ARRAYS' to arguments of type 'MAP_FROM_ARRAYS\\(<INTEGER ARRAY>,"
+            + " <INTEGER>\\)'. Supported form\\(s\\): 'MAP_FROM_ARRAYS\\(<ARRAY>, <ARRAY>\\)'",
+        false);
+    f.checkFails("map_from_arrays(array[1, 2], array['foo'])",
+        "Illegal arguments: The length of the keys array 2 is not equal to the length "
+            + "of the values array 1 in MAP_FROM_ARRAYS function",
+        true);
+  }
```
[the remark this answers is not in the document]
liuyongvs: no need. there is already exist
```
    f.checkType("map_from_arrays(cast(null as integer array), array['foo', 'bar'])",
        "(INTEGER NOT NULL, CHAR(3) NOT NULL) MAP");
    f.checkNull("map_from_arrays(cast(null as integer array), array['foo', 'bar'])");
```

liuyongvs:
because both of you @tanclary @MasseGuillaume says add array with different element type, it will throw exception.
i add

```
   f.checkFails("map_from_arrays(^array[1, '1', true]^, array['a', 'b', 'c'])",
        "Parameters must be of the same type",
        false);
    f.checkFails("map_from_arrays(array['a', 'b', 'c'], ^array[1, '1', true]^)",
        "Parameters must be of the same type",
        false);
```

and from my side, it should be test in array_value_constructor function instead of this. i also add test in array_value_constructor
```
    f.checkFails("^array[1, '1', true]^", "Parameters must be of the same type", false);

```

liuyongvs:
hi @tanclary @MasseGuillaume fix all your reviews, will you have a look again?

site/_docs/reference.md
```diff
@@ -2737,6 +2737,8 @@ BigQuery's type system uses confusingly different names for types and functions:
 | s | MAP_ENTRIES(map)                               | Returns the entries of the *map* as an array, the order of the entries is not defined
 | s | MAP_KEYS(map)                                  | Returns the keys of the *map* as an array, the order of the entries is not defined
 | s | MAP_VALUES(map)                                | Returns the values of the *map* as an array, the order of the entries is not defined
+| s | MAP_FROM_ARRAYS(array1, array2)                | Returns a map created from an *array1* and *array2*. Note that the lengths of two arrays should be the same
+| s | STR_TO_MAP(string[, stringDelimiter[, keyValueDelimiter]]) | Returns a map after splitting the *string* into key/value pairs using delimiters. Default delimiters are ',' for *stringDelimiter* and ':' for *keyValueDelimiter*
```
tanclary, line 2741: nit: should there be spaces here? check other examples with optional args

liuyongvs: @tanclary does it too, for example

```
| m o | REGEXP_REPLACE(string, regexp, rep [, pos [, occurrence [, matchType]]]) | Replaces all substrings of *string* that match *regexp* with *rep* at the starting *pos* in expr (if omitted, the default is 1), *occurrence* means which occurrence of a match to search for (if omitted, the default is 1), *matchType* specifies how to perform matching

```

liuyongvs:
Hi @tanclary @MasseGuillaume @JiajunBernoulli @julianhjulidoes anyone make it further. My other pr depends on it.  Thanks for all reviews very much

liuyongvs:
hi @tanclary @MasseGuillaume do you have time to look again? thank you

liuyongvs:
@tanclary @MasseGuillaume will you have time to look again?

liuyongvs:
Hi @tanclary do you have any other comment.  If not,  i will squalsh the commits

MasseGuillaume (approved):

site/_docs/reference.md
tanclary: In the example you provided, there are spaces between the words and the brackets (rep and [ has a space in between) we should probably stay consistent w/ that

tanclary (approved):
LGTM aside from one nit

site/_docs/reference.md
liuyongvs: fixed @tanclary

liuyongvs:
hi @tanclary thanks for your review, fix the nit.

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
```diff
@@ -897,6 +897,9 @@ ExInst<CalciteException> illegalArgumentForTableFunctionCall(String a0,
   @BaseMessage("Substring error: negative substring length not allowed")
   ExInst<CalciteException> illegalNegativeSubstringLength();
 
+  @BaseMessage("Illegal arguments: The length of the keys array {0,number,#} is not equal to the length of the values array {1,number,#} in MAP_FROM_ARRAYS function")
```
NobiGo, line 900: How about we change this message to 'The key array and value array of parameters must have the same length'?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4003,6 +4004,33 @@ public static List mapValues(Map map) {
...
+  public static Map mapFromArrays(List keysArray, List valuesArray) {
+    if (keysArray.size() != valuesArray.size()) {
```
NobiGo, line 4009: the parameter keysArray and values Array has confirmed is Not NULL?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5688,6 +5688,79 @@ private static void checkIf(SqlOperatorFixture f) {
...
+        "(CHAR(7) NOT NULL, CHAR(7) NOT NULL) MAP NOT NULL");
+    f.checkScalar("str_to_map('a')", "{a=null}",
+        "(CHAR(1) NOT NULL, CHAR(1) NOT NULL) MAP NOT NULL");
+    f.checkScalar("str_to_map('a,b,c')", "{a=null, b=null, c=null}",
+        "(CHAR(5) NOT NULL, CHAR(5) NOT NULL) MAP NOT NULL");
+    f.checkScalar("str_to_map('a-b--c', '--')", "{a-b=null, c=null}",
+        "(CHAR(6) NOT NULL, CHAR(6) NOT NULL) MAP NOT NULL");
+    f.checkType("str_to_map(cast(null as varchar))",
+        "(VARCHAR, VARCHAR) MAP");
+    f.checkNull("str_to_map(cast(null as varchar))");
```
NobiGo, line 5761: Add tests like(Any parameters is NULL, the result will be NULL):

`SELECT str_to_map(null, ',', ':');

SELECT str_to_map('a:1,b:2,c:3', null, ':');

SELECT str_to_map('a:1,b:2,c:3', ',',null);
`

NobiGo (commented):
LGTM. Only need to add some tests.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
liuyongvs: @NobiGo yeap. the calcite supports NullPolicy(6 policy), which will insert code in runtime, spark  using NullIntolerant(only one policy), which is equal to NullPolicy.

for example：
if the NullPolicy.ANY, the calcite will return null  If any of the arguments are null. so the user code doesn't need check whether the input is null

```
defineMethod(MAP_FROM_ARRAYS, BuiltInMethod.MAP_FROM_ARRAYS.method, NullPolicy.ANY);
```

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
liuyongvs: 1) this info is more  concrete, it will make user know what is the problem?
2) the error info aligns with flink. because i think it is better than spark's info.

```
spark error info
  "_LEGACY_ERROR_TEMP_2128" : {
    "message" : [
      "The key array and value array of MapData must have the same length."
    ]
```

```
fink error info
        if (keysArray.size() != valuesArray.size()) {
            throw new FlinkRuntimeException(
                    "Invalid function MAP_FROM_ARRAYS call:\n"
                            + "The length of the keys array "
                            + keysArray.size()
                            + " is not equal to the length of the values array "
                            + valuesArray.size());


https://github.com/apache/flink/blob/65217c35e7c8f19acb1c7761d2a513ec9c0d316d/flink-table/flink-table-runtime/src/main/java/org/apache/flink/table/runtime/functions/scalar/MapFromArraysFunction.java#L30
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
liuyongvs: @NobiGo added the unit tests

NobiGo (approved):
LGTM

PR #3239, opened 2023-06-02
[CALCITE-5739] Add MAP_FROM_ARRAYS function (enabled in Spark library)

liuyongvs:
MAP_FROM_ARRAYS (array1, array2)

Returns a map created from an array1 and *array2. Note that the lengths of two arrays should be the same

```
> SELECT map_from_arrays(array(1.0, 3.0), array('2', '4'));
 {1.0:"2",3.0:"4"} 
```


liuyongvs:
hi @JiajunBernoulli @tanclary @MasseGuillaume do you have time to review this ,just review  MAP_FROM_ARRAYS and after
the MAP_ENTRIES merged, i will rebase it.otherwise, it will lots of conflict

liuyongvs:
as @JiajunBernoulli  saied in the issue, commits with one pr. so close it

PR #3240, opened 2023-06-03
[CALCITE-5748] Upgrade guava from 31.1-jre to 32.0.0-jre

snuyanzin:
regarding NPE on Windows it is just a consequence...
the real reason is fixing of CVE-2020-8908 on guava level[1] and as a result Files#createTempDir stopped working on Windows
I raised an issue for that [2]

Files#createTempDir is used inside embedded redis server and that's why it started to fail on Windows.
The most distracting thing is that this redis embedded server from one side looks not supported since 2018 and from the other side is not easily extensible...
For that reason need to reimplement 3 classes in PR to override behavior

[1] https://github.com/google/guava/commit/feb83a1c8fd2e7670b244d5afd23cba5aca43284
[2] https://github.com/google/guava/issues/6535


core/src/main/java/org/apache/calcite/runtime/Hook.java
```diff
@@ -159,15 +159,6 @@ public <T> Closeable addThread(final Consumer<T> handler) {
...
   }
 
-  // CHECKSTYLE: IGNORE 1
-  /** @deprecated Use {@link #addThread(Consumer)}. */
-  @SuppressWarnings("Guava")
-  @Deprecated // to be removed before 2.0
-  public <T, R> Closeable addThread(
-      final com.google.common.base.Function<T, R> handler) {
-    return addThread((Consumer<T>) handler::apply);
-  }
```
snuyanzin, line 169: With guava 32 it starts failing on errorprone checks with `CheckReturnValue`
Since it is already deprecated with  a comment what to use instead probably it makes sense to remove it

testkit/src/main/java/org/apache/calcite/test/CalciteAssert.java
```diff
@@ -1736,15 +1736,6 @@ public AssertQuery queryContains(Consumer<List> predicate1) {
...
     }
 
-    // CHECKSTYLE: IGNORE 1
-    /** @deprecated Use {@link #queryContains(Consumer)}. */
-    @SuppressWarnings("Guava")
-    @Deprecated // to be removed before 2.0
-    public final AssertQuery queryContains(
-        com.google.common.base.Function<List, Void> predicate1) {
-      return queryContains((Consumer<List>) predicate1::apply);
-    }
```
snuyanzin, line 1746: With guava 32 it starts failing on errorprone checks with `CheckReturnValue`
Since it is already deprecated with  a comment what to use instead probably it makes sense to remove it

PR #3241, opened 2023-06-03
[CALCITE-5745] CONCAT2 function (enabled in Oracle library) gets wrong result when one of the arguments is NULL

ILuffZhe:
In Oracle, CONCAT function returns null only when both two string arguments are null, otherwise it treats null as empty string.

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -768,6 +768,19 @@ public class SqlFunctions {
     return s0 + s1;
   }
 
+  /** Returns null only when both s0 and s1 are null,
+   * otherwise null is treated as empty string. */
+  public static @Nullable String concatWithNull(String s0, String s1) {
+    if (s0 == null && s1 == null) {
+      return null;
+    } else if (s0 == null) {
+      return s1;
+    } else if (s1 == null) {
+      return s0;
+    }
+    return s0 + s1;
+  }
+
   /** SQL {@code binary || binary} operator. */
   public static ByteString concat(ByteString s0, ByteString s1) {
     return s0.concat(s1);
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -768,6 +768,19 @@ public static String concat(String s0, String s1) {
     return s0 + s1;
   }
 
+  /** Return null only when both s0 and s1 is null,
+   * otherwise null is treated as empty string. */
+  public static String concatWithNull(String s0, String s1) {
```
snuyanzin, line 773:
```suggestion
  public static @Nullable String concatWithNull(String s0, String s1) {
```
in case it can return `null`

ILuffZhe: Done.

core/src/test/resources/sql/functions.iq
```diff
@@ -549,4 +549,47 @@ limit 3;
...
++--------+
+| EXPR$0 |
++--------+
+| a      |
++--------+
+(1 row)
+
+!ok
+
+select concat('a', cast(null as varchar));
```
rubenada, line 575: Nitpick: for the sake of completeness, I think we should add also the test scenarios of `null` as first parameter, and `non-null` as second.

ILuffZhe:
Hi, @snuyanzin @rubenada . Do you have any other comments on this patch? Shall I need to squash the commits?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -844,12 +844,18 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
   /** The "CONCAT(arg0, arg1)" function that concatenates strings.
    * For example, "CONCAT('a', 'bc')" returns "abc".
    *
+   * If one of the argument is null, it will be treated as empty string.
```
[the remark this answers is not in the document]
ILuffZhe: Whoops, missed it. Done!

ILuffZhe:
Thanks for the review! @snuyanzin @rubenada @julianhyde

PR #3243, opened 2023-06-05
[CALCITE-5401] Rule fired by HepPlanner can return Volcano's RelSubset

rubenada:
See problem description in:
[CALCITE-5401](https://issues.apache.org/jira/browse/CALCITE-5401) Rule fired by HepPlanner can return Volcano's RelSubset.

Issue solved by  adding new `RelOptRule#convert` with an additional planner parameter, and use it in potentially sensitive locations for this issue to occur (basically rules' `onMatch`)

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/RelOptRule.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptRule.java
@@ -588,8 +588,10 @@ public abstract class RelOptRule {
    * @return a relational expression with the desired traits; never null
    */
   public static RelNode convert(RelNode rel, RelTraitSet toTraits) {
-    RelOptPlanner planner = rel.getCluster().getPlanner();
+    return convert(rel.getCluster().getPlanner(), rel, toTraits);
+  }
 
+  public static RelNode convert(RelOptPlanner planner, RelNode rel, RelTraitSet toTraits) {
     RelTraitSet outTraits = rel.getTraitSet();
     for (int i = 0; i < toTraits.size(); i++) {
       RelTrait toTrait = toTraits.getTrait(i);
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/apache/calcite/plan/RelOptRule.java
```diff
@@ -614,7 +616,10 @@ public static RelNode convert(RelNode rel, RelTraitSet toTraits) {
    * @return a relational expression with the desired trait; never null
    */
   public static RelNode convert(RelNode rel, @Nullable RelTrait toTrait) {
```
zabetak, line 618: Should we deprecate the old methods so that we avoid the problem from re-appearing?

zabetak (approved):
LGTM, left a minor comment for deprecation but feel free to ignore it if you don't think its relevant.

core/src/main/java/org/apache/calcite/plan/RelOptRule.java
rubenada: Thanks for the feedback @zabetak .
At this stage we cannot do that, because the "old" `convert` method is still used by *many* converter rules inside their `public RelNode convert(RelNode rel)` implementations. The problem is that, inside that method, we do not have the `RelOptRuleCall call` to get the planner (it is not passed by `ConverterRule#onMatch` into `#convert`, so in order deprecate this method, we would need a huge refactoring on ConverterRule and all the implementations (which would also impact the downstream ones)... all to fix this rare issue.
On top of that, this problem ("Rule fired by HepPlanner can return Volcano's RelSubset") should normally never occur inside a ConverterRule, because HepPlanner does not care about Conventions, so normally Converter rules would not be applied in there, would they?
All things considered, I think we should correct the problem where it can reasonably happen (Calcite's non-converter rules onMatch implems), going beyond that seems too much disruption for a very, very unlikely problem. Wdyt?

PR #3245, opened 2023-06-05
[CALCITE-5747] Conflicting FLOOR return type between Calcite and BigQ…

tanclary:
…uery

Add another pathway in the Parser, if the conformance is set to BigQuery, use BigQuery FLOOR and CEIL functions so return type can be properly inferred.

Any comments, questions, or suggestions are appreciated as always!

### DIFF
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -114,6 +114,7 @@ import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.parser.SqlParserUtil;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.validate.SqlConformance;
+import org.apache.calcite.sql.validate.SqlConformanceEnum;
 import org.apache.calcite.util.Glossary;
 import org.apache.calcite.util.Pair;
 import org.apache.calcite.util.SourceStringReader;
@@ -7335,9 +7336,7 @@ SqlNode StandardFloorCeilOptions(Span s, boolean floorFlag) :
         }
     )?
     <RPAREN> {
-        SqlOperator op = floorFlag
-            ? SqlStdOperatorTable.FLOOR
-            : SqlStdOperatorTable.CEIL;
+        SqlOperator op = SqlStdOperatorTable.floorCeil(floorFlag, (SqlConformanceEnum) this.conformance);
         function =  op.createCall(s.end(this), args);
     }
     (
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlFloorFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlFloorFunction.java
@@ -16,6 +16,7 @@
  */
 package org.apache.calcite.sql.fun;
 
+import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlFunctionCategory;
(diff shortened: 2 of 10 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -149,6 +149,16 @@ public BigQuerySqlDialect(SqlDialect.Context context) {
   @Override public void unparseCall(final SqlWriter writer, final SqlCall call, final int leftPrec,
       final int rightPrec) {
     switch (call.getKind()) {
+    case CEIL_BIG_QUERY:
+      final SqlWriter.Frame ceilFrame = writer.startFunCall("CEIL");
+      call.operand(0).unparse(writer, leftPrec, rightPrec);
+      writer.endFunCall(ceilFrame);
+      break;
+    case FLOOR_BIG_QUERY:
```
tjbanghart, line 157: Can the unparse logic be pushed to the operator itself? Otherwise I think each dialect will have to override the `BIG_QUERY` style function calls (e.g. BQ incoming query targeting a Postgres backend).

core/src/main/java/org/apache/calcite/sql/fun/SqlFloorFunction.java
```diff
@@ -41,22 +45,23 @@
...
-    Preconditions.checkArgument(kind == SqlKind.FLOOR || kind == SqlKind.CEIL);
+    Preconditions.checkArgument(kind == SqlKind.FLOOR || kind == SqlKind.CEIL
+        || kind == SqlKind.FLOOR_BIG_QUERY || kind == SqlKind.CEIL_BIG_QUERY);
   }
 
   //~ Methods ----------------------------------------------------------------
 
   @Override public SqlMonotonicity getMonotonicity(SqlOperatorBinding call) {
-    // Monotonic iff its first argument is, but not strict.
+    // Monotonic if its first argument is, but not strict.
```
tjbanghart, line 64: Might be intentional for "if and only if"

core/src/main/java/org/apache/calcite/sql/fun/SqlFloorFunction.java
```diff
@@ -41,22 +45,23 @@
...
     return call.getOperandMonotonicity(0).unstrict();
   }
```
tjbanghart, line 67: Consider moving the unparsing logic for the BQ style functions here.

tjbanghart (commented):
Nice! Maybe a link to the BQ docs where it shows this conversion between ints -> floats would be helpful - perhaps in SqlKind so we know why they are being treated differently.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -7335,9 +7336,13 @@ SqlNode StandardFloorCeilOptions(Span s, boolean floorFlag) :
         }
     )?
     <RPAREN> {
-        SqlOperator op = floorFlag
-            ? SqlStdOperatorTable.FLOOR
-            : SqlStdOperatorTable.CEIL;
+        SqlOperator op;
+        boolean isBigQuery = this.conformance.semantics() == SqlLibrary.BIG_QUERY;
```
tanclary, line 7340: look at way to make more scaleable

tanclary:
@tjbanghart Thanks for the review! I think I've adjusted for all of your comments. Feel free to take another look.

core/src/main/java/org/apache/calcite/sql/fun/SqlFloorFunction.java
julianhyde: yes it was intentional.

When podcasters say "If you liked this episode, leave a review on Apple Music" I'm always jumping up and down, "You mean if and only if!"

julianhyde (changes requested):
SqlKind.FLOOR is used in 29 places in the code. By introducing SqlKind.FLOOR_BIG_QUERY I suspect that you broke some of those places. It seems that your new functions are still the same 'species' as FLOOR and CEIL, just with a few differences in their DNA.

It's probably safer to keep FLOOR and CEIL and add an extra field to the SqlOperator (maybe it's sufficient to change the return type strategy and name)?

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -1853,12 +1853,14 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
   /**
    * The <code>FLOOR</code> function.
    */
-  public static final SqlFunction FLOOR = new SqlFloorFunction(SqlKind.FLOOR);
+  public static final SqlFunction FLOOR =
+      new SqlFloorFunction(SqlKind.FLOOR, ReturnTypes.ARG0_OR_EXACT_NO_SCALE);
```
julianhyde, line 1857: I would keep the same constructor, and add methods `withName`, `withReturnTypeInference`. (We have discovered that for functions/operators, subclassing doesn't scale, and adding constructor parameters doesn't scale.)

core/src/main/codegen/templates/Parser.jj
julianhyde: can you create a function in SqlStdOperatorTable similar to `SqlStdOperatorTable.like(boolean,boolean)`? Its args should be boolean and conformance.

site/_docs/reference.md
```diff
@@ -2661,6 +2661,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | s | ARRAY_SIZE(array)                              | Synonym for `CARDINALITY`
 | * | ASINH(numeric)                                 | Returns the inverse hyperbolic sine of *numeric*
 | * | ATANH(numeric)                                 | Returns the inverse hyperbolic tangent of *numeric*
+| b | CEIL(value)                                   | Similar to standard `CEIL(value)` except if *value* is an integer type, the return type is a double
```
julianhyde, line 2664: add space before `|`

julianhyde (changes requested):
two changes requested. when you've done those, +1 to go ahead and merge to main.

core/src/main/codegen/templates/Parser.jj
tanclary: What does the boolean argument represent in this case? Whether you want it to be like/unlike the conformance you pass in?

Or do you mean you pass two conformances in (the one being used and the one you're checking against) and you return `BOOLEAN` (whether they are the same or not)?

tanclary:
> two changes requested. when you've done those, +1 to go ahead and merge to main.

Those changes should be added now.

PR #3248, opened 2023-06-06
[CALCITE-5757] BigQuery DATE_TRUNC return type should be ARG0 and TIMESTAMP_TRUNC/DATETIME_TRUNC should return TIMESTAMP for DATE/TIMESTAMPs and TIMESTAMP_LTZ for TIMESTAMP_LTZ

tanclary:
BigQuery TRUNC functions follow the return type behavior described in [CALCITE-5757](https://issues.apache.org/jira/browse/CALCITE-5757). This PR adjusts the return type for those functions such that it matches the described behavior. 

Notable changes include converting a `DATE` to a `TIMESTAMP` for the `TIMESTAMP_TRUNC` and `DATETIME_TRUNC` functions. It also ensures that `DATE_TRUNC` return type is equivalent to the operand type (aka `ReturnTypes.ARG0`). 

Prior to this PR, discrepancies included `DATETIME_TRUNC` incorrectly returning a `TIMESTAMP_LTZ` if given a `TIMESTAMP` argument and `DATE_TRUNC` having a `DATE` return type even if the argument was a `TIMESTAMP` or `TIMESTAMP_LTZ`.

@julianhyde I opted to just include the TRUNC functions in this PR to keep the review and scope fairly easy (hopefully).

### DIFF
--- a/babel/src/test/resources/sql/big-query.iq
+++ b/babel/src/test/resources/sql/big-query.iq
@@ -2462,6 +2462,14 @@ FROM Dates;
 
 !ok
 
+# Show that the return type matches the operand type
+SELECT
+    DATE_TRUNC(TIMESTAMP "2008-12-25 15:30:00", MONTH) AS timestamp_result,
+    DATE_TRUNC(DATETIME "2008-12-25 15:30:00", MONTH) AS datetime_result;
+timestamp_result TIMESTAMP_WITH_LOCAL_TIME_ZONE NOT NULL
+datetime_result TIMESTAMP NOT NULL
+!type
+
 # In the following example, the original date falls on a
 # Sunday. Because the date_part is WEEK(MONDAY), DATE_TRUNC returns
 # the DATE for the preceding Monday.
@@ -2549,6 +2557,15 @@ SELECT
 (1 row)
 
 !ok
+
+# Show that the return type matches the operand type, unless it is a date
+# in which case a TIMESTAMP is returned.
+SELECT
+    DATETIME_TRUNC(TIMESTAMP "2008-12-25 15:30:00", MONTH) AS timestamp_result,
+    DATETIME_TRUNC(DATETIME "2008-12-25 15:30:00", MONTH) AS datetime_result;
+timestamp_result TIMESTAMP_WITH_LOCAL_TIME_ZONE NOT NULL
+datetime_result TIMESTAMP NOT NULL
+!type
 
 
 # In the following example, the original DATETIME falls on a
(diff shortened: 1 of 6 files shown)


babel/src/test/resources/sql/big-query.iq
```diff
@@ -2462,6 +2462,14 @@ FROM Dates;
 
 !ok
 
+# Show that the return type matches the operand type
+SELECT
+    DATE_TRUNC(TIMESTAMP "2008-12-25 15:30:00", MONTH) AS timestamp_result,
+    DATE_TRUNC(DATETIME "2008-12-25 15:30:00", MONTH) AS datetime_result;
+timestamp_result TIMESTAMP_WITH_LOCAL_TIME_ZONE NOT NULL
+datetime_result TIMESTAMP NOT NULL
+!type
```
olivrlee, line 2471: neat, I didnt know about `!type` in quidem tests

tanclary: yeah super useful!!

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -302,6 +302,29 @@ public static SqlCall stripSeparator(SqlCall call) {
   public static final SqlReturnTypeInference DATE_NULLABLE =
       DATE.andThen(SqlTypeTransforms.TO_NULLABLE);
 
+  /**
+   * Type-inference strategy that returns the type of the first operand,
+   * unless it is a DATE, in which case the return type is TIMESTAMP.
+   */
+  public static final SqlReturnTypeInference ARG0_EXCEPT_DATE = opBinding -> {
```
olivrlee, line 309: Might be nice to add a comment referencing [CALCITE-5757] to remind that this is for BQ typing behavior

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -2049,6 +2054,21 @@ private static class TimestampAddConvertlet implements SqlRexConvertlet {
...
   }
 
+  /** Convertlet that handles the BigQuery {@code DATETIME_TRUNC} and
+   * {@code TIMESTAMP_TRUNC} functions. */
+  private static class TruncConvertlet implements SqlRexConvertlet {
+    @Override public RexNode convertCall(SqlRexContext cx, SqlCall call) {
+      final RexBuilder rexBuilder = cx.getRexBuilder();
+      RexNode op1 = cx.convertExpression(call.operand(0));
+      RexNode op2 = cx.convertExpression(call.operand(1));
+      if (op1.getType().getSqlTypeName() == SqlTypeName.DATE) {
```
olivrlee, line 2064: Maybe also a comment here discussion why DATE is handled separately

julianhyde:
The commit message is currently as follows:

> Correct DATE_TRUNC return type and convert DATEs to TIMESTAMPs for DATETIME_TRUNC and TIMESTAMP_TRUNC for BigQuery

That contains 'correct', which is a synonym for 'fix', and is superfluous. Just state the desired behavior. Would the following be better?

> BigQuery functions DATETIME_TRUNC and TIMESTAMP_TRUNC should return same type as their first argument, and DATE_TRUNC should return DATE

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
tanclary: Updated, let me know if it's clear enough

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
tanclary: Same as above

julianhyde:
After you've removed & inlined `ARG0_EXCEPT_DATE_NULLABLE`, +1, go ahead and merge.

tanclary:
@julianhyde I updated it, also, DATE_TRUNC /should/ return ARG0 and not DATE (in case you wonder why commit message differs from your suggestion)

PR #3249, opened 2023-06-07
[CALCITE-5477] Build with Guava 19.0.

gianm:
Calcite aims to support Guava down to 16.0.1 (https://issues.apache.org/jira/browse/CALCITE-5428), but we are foiled by the fact that `Preconditions.checkArgument` gained new overloads in Guava 20.0. So, when building against Guava 20.0+, the Java compiler generates calls to methods that don't exist in those earlier versions.

On the linked Jira ticket, we discussed three possible solutions:

- Adding an additional level of indirection (implemented in #3028).
- Copying the latest version of Preconditions from Guava into Calcite itself.
- Making release builds of Calcite using Guava 19.

Consensus was to implement the third option for some number of releases, then raise the minimum Guava version to at least 20.0.


gianm:
Note: this PR removes the `errorprone` action, leaving only `errorprone-guava31`. Rationale:

- Running errorprone on guava 19 is problematic, since it flags APIs that are `@Beta` in 19 but have stabilized by 31. It's OK to use these APIs, even on 19, because they're future-proof by virtue of the `@Beta` tag being removed in a later version.
- I don't see a reason to run errorprone on guava 19 at all, given that we are running it on guava 31. Before this patch, both the `errorprone` and `errorprone-guava31` actions ran on guava 31 anyway, because the default version to build was 31.

gianm:
Thank you for the review @julianhyde. I have pushed up these changes. I'll commit it once CI passes.

gianm:
Merging per discussion on https://issues.apache.org/jira/browse/CALCITE-5477. We should revert this prior to 2023-09-01 per https://issues.apache.org/jira/browse/CALCITE-5763.

PR #3252, opened 2023-06-07
[CALCITE-5766] Add SAFE_NEGATE function (enabled for BigQuery library)

tanclary:
PR to contribute to https://issues.apache.org/jira/browse/CALCITE-5591

Any comments, questions, or suggestions appreciated as always!

[BigQuery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#safe_negate)

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -7480,7 +7480,42 @@ public class SqlOperatorTest {
 
   /** Test case for
    * <a href="https://issues.apache.org/jira/browse/CALCITE-5770">[CALCITE-5770]
-   * Add SAFE_SUBTRACT function (enabled in BigQuery library).</a>.
+   * Add SAFE_NEGATE function (enabled in BigQuery library)</a>.
+   */
+  @Test void testSafeNegateFunc() {
+    final SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.SAFE_NEGATE);
+    f0.checkFails("^safe_negate(2)^",
+                  "No match found for function signature "
+                    + "SAFE_NEGATE\\(<NUMERIC>\\)", false);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.BIG_QUERY);
+    f.checkScalar("safe_negate(cast(20 as bigint))", "-20",
+        "BIGINT");
+    f.checkScalar("safe_negate(cast(-20 as bigint))", "20",
+        "BIGINT");
+    f.checkScalar("safe_negate(cast(1.5 as decimal(2, 1)))", "-1.5",
+        "DECIMAL(2, 1)");
+    f.checkScalar("safe_negate(cast(-1.5 as decimal(2, 1)))", "1.5",
+        "DECIMAL(2, 1)");
+    f.checkScalar("safe_negate(cast(12.3456 as double))", "-12.3456",
+        "DOUBLE");
+    f.checkScalar("safe_negate(cast(-12.3456 as double))", "12.3456",
+        "DOUBLE");
+    // Infinity and NaN tests
+    f.checkScalar("safe_negate(cast('Infinity' as double))",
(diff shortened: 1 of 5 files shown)


tanclary:
@zoudan Here's SAFE_NEGATE, feel free to take a look!

zoudan:
There are some checkstyle issues, could you please fix them?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1524,6 +1524,43 @@ public static int multiply(int b0, int b1) {
...
+  /** SQL <code>SAFE_NEGATE</code> function applied to a long value. */
+  public static @Nullable Long safeNegate(long b) {
+    BigDecimal ans = multiply(BigDecimal.valueOf(b), BigDecimal.valueOf(-1));
+    boolean overflow =
+        isOverflow(ans, BigDecimal.valueOf(Long.MAX_VALUE), BigDecimal.valueOf(Long.MIN_VALUE));
+    return overflow ? null : ans.longValue();
+  }
+
+  /** SQL <code>SAFE_NEGATE</code> function applied to a BigDecimal value. */
+  public static @Nullable BigDecimal safeNegate(BigDecimal b) {
```
zoudan, line 1557: Should we add methods to handle other data types such as `float`, `smallint`?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5243,6 +5243,23 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.BIG_QUERY);
+    f.checkScalar("safe_negate(3)", "-3", "INTEGER");
+    f.checkScalar("safe_negate(-3)", "3", "INTEGER");
+    f.checkScalar("safe_negate(0)", "0", "INTEGER");
+    f.checkScalar("safe_negate(cast(3.0 as double))", "-3.0", "DOUBLE");
+    f.checkScalar("safe_negate(cast(3.456 as float))", "-3.456", "FLOAT");
+    f.checkScalar("safe_negate(cast(3.1234 as decimal(5, 4)))",
+        "-3.1234", "DECIMAL(5, 4)");
+    // check overflow returns null
+    f.checkNull("safe_negate(-9223372036854775808)");
```
zoudan, line 5260: Maybe we should add more overflow test cases for all kinds of numerical types？

julianhyde (commented):
I think you need more tests - one for each numeric data type.

The implementation is actually straightforward:
For INTEGER, safe_negate(x) is `x == INTEGER.MIN_VALUE ? null : -x`

You probably don't need to add functions to `SqlFunctions`, just add logic in the implementor.

Or you could just implement `SAFE_NEGATE(x)` as `SAFE_MULTIPLY(x, -1)`.

tanclary:
> I think you need more tests - one for each numeric data type.
> 
> The implementation is actually straightforward: For INTEGER, safe_negate(x) is `x == INTEGER.MIN_VALUE ? null : -x`
> 
> You probably don't need to add functions to `SqlFunctions`, just add logic in the implementor. Or you could just implement `SAFE_NEGATE(x)` as `SAFE_MULTIPLY(x, -1)`.
> 
> The key thing is a comprehensive set of tests.

@julianhyde I revisited this PR and made some of the changes you suggested. As far as tests go, I am not sure how to add tests for non-integer types given that the absolute value of the minimum and maximum values for other types are equivalent, so overflow, to my knowledge, is not possible. Let me know if you feel otherwise or have other comments.

tanclary:
@julianhyde I rebased to include your recent changes and no longer use strings to identify the safe functions' methods. Let me know what you think.

julianhyde:
Two cosmetic things in  testSafeNegateFunc:
 * There are too many '.' in the javadoc. The jira summary does not include a '.'
 * Function arguments should be indented 4, not aligned with the '('

+1 to merge without further review when you've fixed those

PR #3257, opened 2023-06-10
[CALCITE-5759] 'SEARCH(1, Sarg[IS NOT NULL])' should be simplified to 'TRUE'

herunkang2018:


core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -1768,6 +1768,68 @@ private void checkExponentialCnf(int n) {
     checkSimplify(expr, simplified);
   }
 
+  /** Unit test for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5759">[CALCITE-5759]
+   * SEARCH operator with special sarg is not fully simplified</a>. */
+  @Test void testSimplifySearchWithSpecialSargIsNotNull() {
+    RexNode intExpression =
```
libenchao, line 1775: How about name it `intLiteral`?

libenchao (commented):
@herunkang2018 Thanks for your contribution, the changes looks great generally. I only have 1 nit comment.

Besides, for the commit message / Jira title, we can describe it more specifically, such as `SEARCH(1, Sarg[IS NOT NULL]) should be simplified to TRUE`, what do you think?

NobiGo (changes requested):
LGTM. Only a small piece of advice. Need to add code annotation. Like others
`   // "x < x" simplifies to "false"`

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
herunkang2018: I will refine it.

herunkang2018:
@libenchao @NobiGo Thanks a lot for reviewing.
I have changed the commit message / Jira title to `SEARCH(1, Sarg[IS NOT NULL]) should be simplified to TRUE` to make it more clear.

PR #3258, opened 2023-06-11
[CALCITE-5771] Apply two different NULL semantics for CONCAT function(enabled in MySQL, Postgresql, BigQuery and MSSQL)

ILuffZhe:
This patch applies two NULL semantics for CONCAT function(so we need implement two functions):

- for MySQL and BigQUery: returns null if any argument is null, otherwise return string, call it CONCAT_FUNCTION
- for Postgresql and MSSQL: always return string, null is treated as empty string, call it CONCAT_FUNCTION_WITH_NULL
- both of the above functions requires at least 1 argument

For more information, see [ISSUE](https://issues.apache.org/jira/browse/CALCITE-5771).

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -792,6 +792,16 @@ public class SqlFunctions {
     return String.join("", args);
   }
 
+  /** SQL {@code CONCAT(arg0, ...)} function which can accept null
+   * but never return null. Always treats null as empty string. */
+  public static String concatMultiWithNull(String... args) {
+    StringBuilder sb = new StringBuilder();
+    for (String arg : args) {
+      sb.append(arg == null ? "" : arg);
+    }
+    return sb.toString();
+  }
+
   /** SQL {@code CONVERT(s, src_charset, dest_charset)} function. */
   public static String convertWithCharset(String s, String srcCharset,
       String destCharset) {
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -831,15 +831,44 @@ public abstract class SqlLibraryOperators {
       new SqlLikeOperator("NOT RLIKE", SqlKind.RLIKE, true, true);
 
   /** The "CONCAT(arg, ...)" function that concatenates strings.
-   * For example, "CONCAT('a', 'bc', 'd')" returns "abcd". */
-  @LibraryOperator(libraries = {MYSQL, POSTGRESQL, BIG_QUERY})
+   * For example, "CONCAT('a', 'bc', 'd')" returns "abcd".
+   *
+   * <p>It accepts at least 1 argument and returns null if any of
+   * the arguments is null. */
+  @LibraryOperator(libraries = {MYSQL, BIG_QUERY})
   public static final SqlFunction CONCAT_FUNCTION =
       SqlBasicFunction.create("CONCAT",
           ReturnTypes.MULTIVALENT_STRING_SUM_PRECISION_NULLABLE,
-          OperandTypes.repeat(SqlOperandCountRanges.from(2),
+          OperandTypes.repeat(SqlOperandCountRanges.from(1),
               OperandTypes.STRING),
           SqlFunctionCategory.STRING)
           .withOperandTypeInference(InferTypes.RETURN_TYPE);
+
+  /** The "CONCAT(arg, ...)" function that concatenates strings,
+   * which never returns null.
+   * For example, "CONCAT('a', 'bc', 'd')" returns "abcd".
+   *
+   * <p>If one of the arguments is null, it will be treated as empty string.
+   * "CONCAT('a', null)" returns "a".
+   * "CONCAT('a', null, 'b')" returns "ab".
+   *
+   * <p>Returns empty string only when all arguments are null
+   * or the empty string.
+   * "CONCAT(null)" returns "".
+   * "CONCAT(null, '')" returns "".
+   * "CONCAT(null, null, null)" returns "".
+   *
+   * <p>It differs from {@link #CONCAT_FUNCTION} when processing
+   * null values. */
+  @LibraryOperator(libraries = {MSSQL, POSTGRESQL})
+  public static final SqlFunction CONCAT_FUNCTION_WITH_NULL =
+      SqlBasicFunction.create("CONCAT",
+          ReturnTypes.MULTIVALENT_STRING_SUM_PRECISION_NOT_NULLABLE,
+          OperandTypes.repeat(SqlOperandCountRanges.from(1),
+              OperandTypes.STRING),
+          SqlFunctionCategory.STRING)
+          .withOperandTypeInference(InferTypes.RETURN_TYPE)
+          .withKind(SqlKind.CONCAT_WITH_NULL);
 
   /** The "CONCAT(arg0, arg1)" function that concatenates strings.
    * For example, "CONCAT('a', 'bc')" returns "abc".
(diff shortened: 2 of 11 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -831,16 +831,43 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
+   * null values. */
+  @LibraryOperator(libraries = {MSSQL, POSTGRESQL})
+  public static final SqlFunction CONCAT_FUNCTION_WITH_NULL =
+      SqlBasicFunction.create("CONCAT",
+          ReturnTypes.MULTIVALENT_STRING_SUM_PRECISION_NOT_NULLABLE,
+          OperandTypes.repeat(SqlOperandCountRanges.from(1),
+              OperandTypes.STRING),
+          SqlFunctionCategory.STRING)
+          .withOperandTypeInference(InferTypes.RETURN_TYPE)
+          .withKind(SqlKind.CONCAT_WITH_NULL);
```
tanclary, line 869: is there a SqlKind.CONCAT? What is the advantage of having multiply SqlKind entries?

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
```diff
@@ -149,6 +151,24 @@ static <E> List<E> list() {
...
+    // tests show. Not the desired behavior for SQL.
+    assertThat(concatMulti((String) null), is("null"));
+    assertThat(concatMulti((String) null, null), is("nullnull"));
+    assertThat(concatMulti("a", null, "b"), is("anullb"));
+  }
+
+  @Test void testConcatMultiWithNull() {
+    assertThat(concatMultiWithNull("a b", "cd", "e"), is("a bcde"));
+    // Null value could be passed in which is treated as empty string.
+    assertThat(concatMultiWithNull((String) null), is(""));
```
tanclary, line 167: Just to clarify, which of these tests mimics the behavior that BigQuery follows? My understanding is that BigQuery does the following:

`CONCAT(null, null)` returns `NULL` while
`CONCAT("", "")` returns `""`

tanclary (commented):
Just left a couple comments/questions

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -792,6 +792,16 @@ public static String concatMulti(String... args) {
     return String.join("", args);
   }
 
+  /** SQL {@code CONCAT(arg0, ...)} function which can accept null
```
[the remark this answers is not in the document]
ILuffZhe: Whoops, it's a typo.

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
ILuffZhe: For BigQuery, you should see testConcatMulti(), and you are right about the behavior.(but the code generator will ensure nulls won't be passed in)
testConcatMultiWithNull() is a test for CONCAT function that enabled in Postgresql and MSSQL.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
ILuffZhe: Those two functions(CONCAT_FUNCTION and CONCAT_FUNCTION_WITH_NULL) have the same operator name and operands type/count, so we need a new SqlKind when overloading functions.
You can also see CONCAT2 function that enabled in Oracle, and we use SqlKind.CONCAT2 to make it different.

ILuffZhe:
The failed CI seems not relevant to this PR, so I'll leave it as it is.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -831,16 +831,43 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
 
+  /** The "CONCAT(arg, ...)" function that concatenates strings,
+   * which never returns null.
+   * For example, "CONCAT('a', 'bc', 'd')" returns "abcd".
+   *
+   * <p>If one of the arguments is null, it will be treated as empty string.
+   * "CONCAT('a', null)" returns "a".
+   * "CONCAT('a', null, 'b')" returns "ab".
+   *
+   * <p>Returns empty string only when all arguments are null.
```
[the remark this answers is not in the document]
ILuffZhe: Good point! I've updated the PR.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1959,6 +1960,24 @@ private static void checkConcatFunc(SqlOperatorFixture f) {
     f.checkFails("^concat()^", INVALID_ARGUMENTS_NUMBER, false);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5771">[CALCITE-5771]
+   * Apply two different NULL semantics for CONCAT function(enabled in MySQL、
```
rubenada, line 1965: I think it would be better to use here the "regular comma":  `,`  instead of `、`
The same applies for the similar messages in functions.iq, final commit message, PR title and Jira title.

rubenada:
@ILuffZhe  I left you a minor remark about the comma character to be used in the comments / titles of this issue.
Apart from that, patch looks good.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
ILuffZhe: Done.

core/src/test/resources/sql/functions.iq
```diff
@@ -499,6 +510,120 @@ select CONVERT(DATE, '05/01/2000', 103);
...
++--------+
+(1 row)
+
+!ok
+
+select concat('', '');
++--------+
+| EXPR$0 |
++--------+
+|        |
```
rubenada, line 538: Stupid question: how can we distinguish if the expected result is the empty string (as in here) or null (as in the "concat in BigQuery" tests a few lines below)? Both cases look the same at first glance....

ILuffZhe: As far as I know, we cannot distinguish them in functions.iq(null and "" behave the same since we don't use (null) for NULL result here). But we can tell the difference in SqlOperatorTest.

rubenada: What if we do something like:
```
select v, v is null from (select concat(null, '') as v);
```
On the "ambiguous" tests?

ILuffZhe: Agreed. I add two test cases in functions.iq which help to tell the difference of the result is null or empty string.

rubenada:
Thanks @ILuffZhe  lgtm.

@tanclary do you want to take another look? Do you have further comments before the final squash+merge?

core/src/test/resources/sql/functions.iq
```diff
@@ -499,6 +510,144 @@ select CONVERT(DATE, '05/01/2000', 103);
...
+select concat(null, null, null);
++--------+
+| EXPR$0 |
++--------+
+|        |
++--------+
+(1 row)
+
+!ok
+
```
julianhyde, line 638: Quidem's output format isn't helpful here, because it prints empty strings the same as null.

May I suggest that you make each test print the value and also print whether that value is null, e.g.
```
with t as (select concat(null, null, null) as c)
select c, c is null as c_is_null
from t;
```

tanclary:
> Thanks @ILuffZhe lgtm.
> 
> @tanclary do you want to take another look? Do you have further comments before the final squash+merge?

LGTM!

core/src/test/resources/sql/functions.iq
ILuffZhe: Done. I also add such test for CONCAT2 function in Oracle by the way.

tanclary:
This looks good to me, if you squash your commits I can merge

ILuffZhe:
@tanclary @rubenada @julianhyde Thanks for your review!

PR #3260, opened 2023-06-11
[CALCITE-5751] Add ARRAY_APPEND, ARRAY_POSITION, ARRAY_REMOVE, ARRAY_…

liuyongvs:
…PREPEND function (enabled in Spark library)


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3931,6 +3939,34 @@ public static List distinct(List list) {
     return min;
   }
```
liuyongvs, line 3941: comments:
it can not use 

```
list.add
return list;
```
because it will throw exception
java.lang.UnsupportedOperationException
	at java.util.AbstractList.add(AbstractList.java:148)
	at java.util.AbstractList.add(AbstractList.java:108)
	at org.apache.calcite.runtime.SqlFunctions.arrayAppend(SqlFunctions.java:3899)

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -945,6 +964,27 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
...
+      SqlBasicFunction.create(SqlKind.ARRAY_PREPEND,
+          SqlLibraryOperators::arrayAppendPrependReturnType,
+          OperandTypes.ARRAY_ELEMENT);
+
+  /** The "ARRAY_REMOVE(array, element)" function. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction ARRAY_REMOVE =
+      SqlBasicFunction.create(SqlKind.ARRAY_REMOVE,
+          ReturnTypes.ARG0_NULLABLE,
+          OperandTypes.ARRAY_ELEMENT);
```
liuyongvs, line 986: comments: 
ARRAY_REMOVE's behavior  is also different in spark and flink, just like array_contains. and i  also
add a unit test.
```
 // Flink and Spark differ on the following. The expression
    //   array_remove(array[1, null], cast(null as integer))
    // returns [1] in Flink, and returns null in Spark. The current
    // function has Spark behavior, but if we supported a Flink function
    // library (i.e. "fun=flink") we could add a function with Flink behavior.
    f.checkNull("array_remove(array[1, null], cast(null as integer))");
```

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -945,6 +964,27 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
           ReturnTypes.TO_COLLECTION_ELEMENT_FORCE_NULLABLE,
           OperandTypes.ARRAY);
 
+  /** The "ARRAY_POSITION(array, element)" function. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction ARRAY_POSITION =
+      SqlBasicFunction.create(SqlKind.ARRAY_POSITION,
+          ReturnTypes.BIGINT_NULLABLE,
+          OperandTypes.ARRAY_ELEMENT);
+
```
liuyongvs, line 973: comments:
ARRAY_POSITION's return type is bigint. some of other vendor (for example flink)is int. 
and some of vendor returns result is 0-based index. this i add in the docs(reference.md)

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -878,6 +878,26 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
...
+    RelDataType type = arrayType.getComponentType();
+    final RelDataType elementType = opBinding.collectOperandTypes().get(1);
+
+    if (elementType.isNullable()) {
+      type = opBinding.getTypeFactory().createTypeWithNullability(type, true);
+    }
+    requireNonNull(type, "inferred array element type");
+    return SqlTypeUtil.createArrayType(opBinding.getTypeFactory(), type, arrayType.isNullable());
+  }
+
```
liuyongvs, line 893: comments: 
the array's internal nullable should be changed when element type is nullable

liuyongvs:
hi @JiajunBernoulli @julianhyde @tanclary @rubenada will you also help review this pr, the important part i have added the comments, which will be convenient to review.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -878,6 +878,26 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
           SqlLibraryOperators::arrayReturnType,
           OperandTypes.SAME_VARIADIC);
```
liuyongvs, line 880: comments:
so it should add @SuppressWarnings("argument.type.incompatible") like others
```
[Task :core:compileJava] [argument.type.incompatible] incompatible argument for parameter type of createTypeWithNullability. type = opBinding.getTypeFactory().createTypeWithNullability(type, true);
```

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3894,6 +3894,14 @@ public static List compact(List list) {
     return result;
   }
 
+  /** Support the ARRAY_APPEND function. */
+  public static List arrayAppend(List list, Object element) {
+    final List result = new ArrayList(list.size() + 1);
```
JiajunBernoulli, line 3899: Can you resolve it?

image

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3931,6 +3939,34 @@ public static List distinct(List list) {
...
+    result.add(element);
+    result.addAll(list);
+    return result;
+  }
+
+  /** Support the ARRAY_POSITION function. */
+  public static Long arrayPosition(List list, Object element) {
+    final int index = list.indexOf(element);
+    if (index != -1) {
```
JiajunBernoulli, line 3953: Can you resolve it?
image

site/_docs/reference.md
```diff
@@ -2652,6 +2652,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | m | expr1 <=> expr2                                | Whether two values are equal, treating null values as the same, and it's similar to `IS NOT DISTINCT FROM`
 | * | ACOSH(numeric)                                 | Returns the inverse hyperbolic cosine of *numeric*
 | s | ARRAY(expr [, expr ]*)                         | Construct an array in Apache Spark
+| s | ARRAY_APPEND(array, element)                   | Appends an *element* to the end of the *array* and returns the result. If the *array* itself is null, the function will return null. If an *element* to add is null, the null *element* will be added to the end of the *array*
```
JiajunBernoulli, line 2655: 1. `If the *array* itself is null,` `itself` maybe redundant.
2. `Type of *element* should be similar to type of the elements of the array`. 
3. `If an *element* that is null, the null *element* will be added to the end of the *array*`

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
liuyongvs: fixed @JiajunBernoulli

site/_docs/reference.md
```diff
@@ -2661,6 +2662,9 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | ARRAY_LENGTH(array)                            | Synonym for `CARDINALITY`
 | s | ARRAY_MAX(array)                               | Returns the maximum value in the *array*
 | s | ARRAY_MIN(array)                               | Returns the minimum value in the *array*
+| s | ARRAY_POSITION(array, element)                 | Returns the (1-based) index of the first *element* of the *array* as long
+| s | ARRAY_REMOVE(array, element)                   | Remove all elements that equal to *element* from the *array*
+| s | ARRAY_PREPEND(array, element)                  | Appends an *element* to the beginning of the *array* and returns the result. If the *array* itself is null, the function will return null. If an *element* to add is null, the null *element* will be added to the beginning of the *array*
```
JiajunBernoulli, line 2667: Same suggestion as `ARRAY_APPEND`.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5334,6 +5334,33 @@ private static void checkIf(SqlOperatorFixture f) {
...
+            + "<INTEGER ARRAY>, <NUMERIC>\\)", false);
+
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
+    f.checkScalar("array_append(array[1], 2)", "[1, 2]",
+        "INTEGER NOT NULL ARRAY NOT NULL");
+    f.checkScalar("array_append(array[1], null)", "[1, null]",
+        "INTEGER ARRAY NOT NULL");
+    f.checkScalar("array_append(array(null), null)", "[null, null]",
+        "NULL ARRAY NOT NULL");
+    f.checkScalar("array_append(array(), null)", "[null]",
```
JiajunBernoulli, line 5352: Can we append integer or string to `array_append(array(), null)`?

(The `UNKNOWN` type maybe trick.）

site/_docs/reference.md
liuyongvs: fixed

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
liuyongvs: good catch! it should and i add the test array_append(array(), 1) i find a bug , the return type should return leastRestrictive type. i fixed.

liuyongvs: the result is aligns with spark. some different is return type not null. which is not print in spark.
```
spark-sql (default)> select array_append(array(), 1);
[1]
spark-sql (default)> select typeof(array_append(array(), 1));
array<int>

spark-sql (default)> select array_append(array(), null);
[null]
spark-sql (default)> select typeof(array_append(array(), null));
array<void>

```

liuyongvs:
@JiajunBernoulli thanks for your review so much and fix all your reviews

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -878,6 +880,28 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
...
 
+  @SuppressWarnings("argument.type.incompatible")
+  private static RelDataType arrayAppendPrependReturnType(SqlOperatorBinding opBinding) {
```
JiajunBernoulli, line 884: Is there no more suitable class to save it?

`SqlLibraryOperators` only should  defines functions and operators.

julianhyde: @JiajunBernoulli, I think it's OK. There are already a few methods like this that help define the return type (and are used via  `::` method references to become lambdas)

JiajunBernoulli: Thanks for your explanation.

JiajunBernoulli (approved):

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -177,6 +177,14 @@ public static SqlCall stripSeparator(SqlCall call) {
   public static final SqlReturnTypeInference ARG0_NULLABLE =
       ARG0.andThen(SqlTypeTransforms.TO_NULLABLE);
 
+  /**
+   * Type-inference strategy whereby the result type of a call is the type of
+   * the operand #0 (0-based). If the operand #0 (0-based) is nullable, the
+   * returned type will also be nullable.
+   */
+  public static final SqlReturnTypeInference ARG0_NULLABLE_IF_ARG0_NULLABLE =
```
JiajunBernoulli, line 185: Does it be used?

liuyongvs: i think put it here for others to use @JiajunBernoulli

liuyongvs:
hi @JiajunBernoulli @julianhyde thanks your for review, squash the commits

liuyongvs:
Hi @JiajunBernoulli will it be merged now?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
liuyongvs: The ci failed with this.  So i can see it

PR #3262, opened 2023-06-13
[CALCITE-5772] Add MAP_CONCAT, MAP_FROM_ENTRIES function (enabled in …

liuyongvs:
[map_concat](https://spark.apache.org/docs/latest/api/sql/index.html#map_concat)

map_concat(map, ...) - Returns the union of all the given maps

Examples:
```
> SELECT map_concat(map(1, 'a', 2, 'b'), map(3, 'c'));
 {1:"a",2:"b",3:"c"} 
```
[map_from_entries](https://spark.apache.org/docs/latest/api/sql/index.html#map_from_entries)

map_from_entries(arrayOfEntries) - Returns a map created from the given array of entries.

Examples:

```
> SELECT map_from_entries(array(struct(1, 'a'), struct(2, 'b')));
 {1:"a",2:"b"}
```


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1005,6 +1012,23 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
...
+  private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding) {
+    final RelDataType entriesType = opBinding.collectOperandTypes().get(0);
+    final RelDataType entryType = entriesType.getComponentType();
+    return SqlTypeUtil.createMapType(
+        opBinding.getTypeFactory(),
+        requireNonNull(entryType.getFieldList().get(0).getType(), "inferred key type"),
+        requireNonNull(entryType.getFieldList().get(1).getType(), "inferred value type"),
+        entriesType.isNullable() || entryType.isNullable());
+  }
+
```
liuyongvs, line 1024: comments:
```
comments:

because my other prs will conflict it, so i don't submit it now. so i add an summary:

the spark MAP_FROM_ENTRIES nullable have 2 case:

1) input is nullable

2) array, which contains null entry like this ,it will also return null

so we should take this into consideration, type and result, i will add comment in the pr when you review you will see

// spark
spark-sql> select map_from_entries(array(struct(1, 'a'), null));
NULL 

// calcite should test this
f.checkNull("map_from_entries(array[row(1, 'a'), null])");
f.checkType("map_from_entries(array[row(1, 'a'), null])",
    "(INTEGER, CHAR(1)) MAP");
 

case class MapFromEntries(child: Expression)
  extends UnaryExpression
  with NullIntolerant
  with QueryErrorsBase {

  @transient
  private lazy val dataTypeDetails: Option[(MapType, Boolean, Boolean)] = child.dataType match {
    case ArrayType(
      StructType(Array(
        StructField(_, keyType, keyNullable, _),
        StructField(_, valueType, valueNullable, _))),
      containsNull) => Some((MapType(keyType, valueType, valueNullable), keyNullable, containsNull))
    case _ => None
  }

  @transient private lazy val nullEntries: Boolean = dataTypeDetails.get._3

  override def nullable: Boolean = child.nullable || nullEntries
```

liuyongvs:
hi @rubenada do you have time to review it?

liuyongvs:
Hi @JiajunBernoulli do you have time to review this?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3984,6 +3985,13 @@ public static List sortArray(List list, boolean ascending) {
     return list;
   }
 
+  /** Support the MAP_CONCAT function. */
+  public static Map mapConcat(Map... maps) {
+    final Map result = new LinkedHashMap();
+    Arrays.stream(maps).forEach(result::putAll);
```
zoudan, line 3991: I prefer not to use lambda expressions for performance reasons, WDYT?

liuyongvs: @zoudan i have a question, why  lambda expressions has poor performance

MasseGuillaume (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5646,6 +5646,33 @@ private static void checkIf(SqlOperatorFixture f) {
...
+        "(CHAR(3) NOT NULL, INTEGER NOT NULL) MAP NOT NULL");
+    f.checkScalar("map_concat(map[null, 1], map[null, 2])", "{null=2}",
+        "(NULL, INTEGER NOT NULL) MAP NOT NULL");
+    f.checkScalar("map_concat(map[1, 2], map[1, null])", "{1=null}",
+        "(INTEGER NOT NULL, INTEGER) MAP NOT NULL");
+    f.checkNull("map_concat(map['foo', 1], null)");
+    f.checkType("map_concat(map['foo', 1], null)", "(CHAR(3) NOT NULL, INTEGER NOT NULL) MAP");
+    f.checkNull("map_concat(null, map['foo', 1])");
+    f.checkType("map_concat(null, map['foo', 1])", "(CHAR(3) NOT NULL, INTEGER NOT NULL) MAP");
+    f.checkFails("^map_concat()^", INVALID_ARGUMENTS_NUMBER, false);
```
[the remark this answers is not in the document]
liuyongvs: @MasseGuillaume  good catch! spark supports zero arg.
```
  @transient override lazy val dataType: MapType = {
    if (children.isEmpty) {
      MapType(StringType, StringType)
    } else {
      super.dataType.asInstanceOf[MapType]
    }
  }
```
and i added the test

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1020,12 +1022,32 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
...
+      requireNonNull(type, "type");
+      return SqlTypeUtil.createMapType(typeFactory, type, type, true);
+    } else {
+      final List<RelDataType> operandTypes = opBinding.collectOperandTypes();
+      for (RelDataType operandType: operandTypes) {
+        if (!SqlTypeUtil.isMap(operandType)) {
+          throw opBinding.newError(
+              RESOURCE.typesShouldAllBeMap(
+                  opBinding.getOperator().getName(),
+                  operandType.getFullTypeString()));
```
liuyongvs, line 1038: comments:
we need check every arg whether it is map. and i also add the tests. but we can have 2 different implementation here.
1) check operand type in return type inference. the implementation is simple
2) check operand type in operand type inference.

```
  public static final SqlFunction MAP_CONCAT =
      SqlBasicFunction.create(SqlKind.MAP_CONCAT,
          SqlLibraryOperators::deriveTypeMapConcat,
          OperandTypes.SAME_VARIADIC.and(OperandTypes.ALL_ARGS_MAP));

  private static class AllArgsMapOperandTypeChecker
      implements SqlSingleOperandTypeChecker {
    @Override public boolean checkSingleOperandType(SqlCallBinding callBinding,
        SqlNode node, int iFormalOperand, boolean throwOnFailure) {
      RelDataType type = SqlTypeUtil.deriveType(callBinding, node);
      boolean valid = false;
      if (type.getSqlTypeName() == SqlTypeName.MAP) {
        valid = true;
      }
      if (!valid && throwOnFailure) {
        throw callBinding.newError(
            RESOURCE.typesShouldAllBeMap(
                callBinding.getOperator().getName(),
                type.getFullTypeString()));
      }
      return valid;
    }

      return SqlOperandCountRanges.any();
    }

    @Override public String getAllowedSignatures(SqlOperator op, String opName) {
      ....
    }
  }
```



```
spark-sql (default)> select map_concat(map('foo', 1), null);
[DATATYPE_MISMATCH.MAP_CONCAT_DIFF_TYPES] Cannot resolve "map_concat(map(foo, 1), NULL)" due to data type mismatch: The `map_concat` should all be of type map, but it's ["MAP<STRING, INT>", "VOID"].; line 1 pos 7;
'Project [unresolvedalias(map_concat(map(foo, 1), null), None)]
+- OneRowRelation

spark-sql (default)> select map_concat(map('foo', 1), cast(null as map<string, int>));
NULL
```

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
zoudan: I have an impression that lambda expressions have performance issues in certain scenarios in jdk8, but I haven't found clear evidence yet, so it is ok if you want to keep it.

liuyongvs:
rebase to fix conflicts with main

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
liuyongvs: @zoudan thanks for your suggestion. i think it makes code clear, so i will not change it

site/_docs/reference.md
```diff
@@ -2740,9 +2740,11 @@ BigQuery's type system uses confusingly different names for types and functions:
 | m | TO_BASE64(string)                              | Converts the *string* to base-64 encoded form and returns a encoded string
 | b m | FROM_BASE64(string)                          | Returns the decoded result of a base-64 *string* as a string
 | b o | LTRIM(string)                                | Returns *string* with all blanks removed from the start
+| s | MAP_CONCAT(map [, map]*)                       | Concatenates one or more maps. If any input argument is `NULL` the function returns `NULL`
```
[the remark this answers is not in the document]
liuyongvs: @MasseGuillaume add the comments for all the spark map constuct function

PR #3263, opened 2023-06-14
[CALCITE-5778] Add ARRAY_JOIN, ARRAYS_OVERLAP, ARRAYS_ZIP function (e…

liuyongvs:
[array_join](https://spark.apache.org/docs/latest/api/sql/index.html#array_join)

array_join(array, delimiter[, nullReplacement]) - Concatenates the elements of the given array using the delimiter and an optional string to replace nulls. If no value is set for nullReplacement, any null value is filtered.

Examples:
```
 > SELECT array_join(array('hello', 'world'), ' ');
 hello world
> SELECT array_join(array('hello', null ,'world'), ' ');
 hello world
> SELECT array_join(array('hello', null ,'world'), ' ', ',');
 hello , world
```

[arrays_overlap](https://spark.apache.org/docs/latest/api/sql/index.html#arrays_overlap)

arrays_overlap(a1, a2) - Returns true if a1 contains at least a non-null element present also in a2. If the arrays have no common element and they are both non-empty and either of them contains a null element null is returned, false otherwise.

Examples:
```
> SELECT arrays_overlap(array(1, 2, 3), array(3, 4, 5));
 true 
```

[arrays_zip](https://spark.apache.org/docs/latest/api/sql/index.html#arrays_zip)

arrays_zip(a1, a2, ...) - Returns a merged array of structs in which the N-th struct contains all N-th values of input arrays.

Examples:
```
> SELECT arrays_zip(array(1, 2, 3), array(2, 3, 4));
 [{"0":1,"1":2},{"0":2,"1":3},{"0":3,"1":4}]
> SELECT arrays_zip(array(1, 2), array(2, 3), array(3, 4));
 [{"0":1,"1":2,"2":3},{"0":2,"1":3,"2":4}]
```


core/src/main/java/org/apache/calcite/sql/type/SqlTypeTransforms.java
```diff
@@ -111,6 +112,26 @@ public abstract class SqlTypeTransforms {
...
+          final RelDataType componentType = requireNonNull(arrayType.getComponentType());
+          argComponentTypes.add(componentType);
+        }
+
+        if (argComponentTypes.stream().anyMatch(RelDataType::isNullable)) {
+          return opBinding.getTypeFactory()
+              .createTypeWithNullability(typeToTransform, true);
+        }
+        return typeToTransform;
+      };
```
liuyongvs, line 133: comments:
nullable has 4 case:
so here does it too
```
// spark
case class ArraysOverlap(left: Expression, right: Expression)
  override def nullable: Boolean = {
    left.nullable || right.nullable || leftArrayElementNullable || rightArrayElementNullable
  }
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5594,6 +5619,86 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f0.checkFails("^arrays_zip(array[1, 2], array[2])^",
+        "No match found for function signature ARRAYS_ZIP\\("
+            + "<INTEGER ARRAY>, <INTEGER ARRAY>\\)", false);
+
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
+    f.checkScalar("arrays_zip(array[1, 2], array[2, 3], array[3, 4])", "[{1, 2, 3}, {2, 3, 4}]",
+        "RecordType(INTEGER NOT NULL 0, INTEGER NOT NULL 1, INTEGER NOT NULL 2) "
+            + "NOT NULL ARRAY NOT NULL");
+    f.checkScalar("arrays_zip(array[1, 2], array[2])", "[{1, 2}, {2, null}]",
+        "RecordType(INTEGER NOT NULL 0, INTEGER NOT NULL 1) NOT NULL ARRAY NOT NULL");
```
liuyongvs, line 5668: comments: why the result is different . 
```
-- spark
spark-sql> select arrays_zip(array(1, 2), array(2));
[{"0":1,"1":2},{"0":2,"1":null}]

-- calcite
f.checkScalar("arrays_zip(array[1, 2], array[2])", "[{1, 2}, {2, null}]",

```

the reason is row print string is different with calcite for example, but it does impact the logical of this function

```
-- spark 
spark-sql> select struct(1, 2);
{"col1":1,"col2":2}

-- calcite 
select row(1, 'a');
[{1, 2}]
```

liuyongvs:
hi @julianhyde @JiajunBernoulli @tanclary will you also have time to review it , the important port i add the comments

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3883,6 +3883,59 @@ private static AtomicLong getAtomicLong(String key) {
     return atomic;
   }
 
+  /** Support the ARRAYS_OVERLAP function. */
+  public static @Nullable Boolean arraysOverlap(List list1, List list2) {
+    final List bigger = list1.size() > list2.size() ? list1 : list2;
+    final List smaller = list1.size() > list2.size() ? list2 : list1;
+    boolean hasNull = false;
```
tanclary, line 3890: could this be simplified using `.retainsAll()` or some other set operation like an `intersect`? I guess the `NULL` behavior could make that more difficult, just wondering.

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
```diff
@@ -639,6 +639,8 @@ public enum BuiltInMethod {
   ARRAY_INTERSECT(SqlFunctions.class, "arrayIntersect", List.class, List.class),
   ARRAY_UNION(SqlFunctions.class, "arrayUnion", List.class, List.class),
   ARRAY_REVERSE(SqlFunctions.class, "reverse", List.class),
+  ARRAYS_OVERLAP(SqlFunctions.class, "arraysOverlap", List.class, List.class),
```
tanclary, line 642: Why no entry for `ARRAY_JOIN`?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
julianhyde: `retainAll` won't work because it mutates one list.

However, we have `Util.intersects(Collection, Collection)`. It's even tested. However it assumes that there are no null values.

Maybe you could quickly check for nulls and then call `Util.intersects` when you've determined that there are no nulls.

liuyongvs: hi @julianhyde @tanclary 
1) why Util.intersects is fast, it has O(n * m) time space, and my implementation just has O(n+m) time space.
```
  public static <E> boolean intersects(Collection<E> c0, Collection<E> c1) {
    for (E e : c1) {
      if (c0.contains(e)) {
        return true;
      }
    }
    return false;
  }
```

2)  about null
Returns true if *array1 contains at least a non-null element present also in *array2*. If the arrays have no common element and they are both non-empty and either of them contains a null element null is returned, false otherwise
  
if you look the three unit test, you will know why, and the implementation is align with spark
f.checkScalar("arrays_overlap(array[1, null], array[1])", true,
        "BOOLEAN");
f.checkNull("arrays_overlap(array[1], array[2, null])");
f.checkScalar("arrays_overlap(array[1, 2], array[3])", false,
        "BOOLEAN NOT NULL");

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
liuyongvs: @tanclary       defineMethod(ARRAY_JOIN, "arrayToString", NullPolicy.STRICT), which  aligns with array_to_string in bigquery

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
julianhyde: Yes, your algorithm is probably more efficient.

Note that `intersects` uses `contains`, which is possibly optimized using intrinsics.

You could assign smaller and larger using one test rather than two. In fact, to keep it simple, I'd probably write
```
if (list1.size() > list2.size()) {
  return arraysOverlap(list2, list1);
}
```

Can you add a unit test to `SqlFunctionsTest`? There are at least 16 combinations to check: each list can be empty, contain only nulls, contain a mixture of nulls and non-nulls, and contain only non-nulls, which gives 4 * 4. That's easier done in a Java unit test.

julianhyde: Also,
```
final Set smallestSet = new HashSet(smaller);
hasNull = smallestSet.remove(null);
```
is probably more efficient because it allows the set to right-size the table.

liuyongvs: hi @julianhyde added the unit test, thanks!

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
```diff
@@ -96,6 +97,35 @@ static <E> List<E> list() {
     return ImmutableList.of();
   }
 
+  @Test void testArraysOverlap() {
+    final List<Object> listWithOnlyNull = new ArrayList<>();
+    listWithOnlyNull.add(null);
+
```
liuyongvs, line 103: comments:
```
spark-sql (default)> select arrays_overlap(array(1, null), array());
false

```

liuyongvs:
hi @julianhyde @tanclary thanks for your review very much, do you have time to look again?

liuyongvs:
@JiajunBernoulli do you also have time to help review?

liuyongvs:
hi @JiajunBernoulli @MasseGuillaume will you also have time to help review this pr?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5415,6 +5415,31 @@ private static void checkIf(SqlOperatorFixture f) {
...
+        + " signature ARRAY_JOIN\\(<CHAR\\(2\\) ARRAY>, <CHARACTER>\\)", false);
+
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
+    f.checkScalar("array_join(array['aa', 'b', 'c'], '-')", "aa-b-c",
+        "VARCHAR NOT NULL");
+    f.checkScalar("array_join(array[null, 'aa', null, 'b', null], '-', 'empty')",
+        "empty-aa-empty-b-empty", "VARCHAR NOT NULL");
+    f.checkScalar("array_join(array[null, 'aa', null, 'b', null], '-')", "aa-b",
+        "VARCHAR NOT NULL");
+    f.checkScalar("array_join(array[null, x'aa', null, x'bb', null], '-')", "aa-bb",
```
NobiGo, line 5431: What do you mean x'aa' ? Looks like some skill I don't know.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5594,6 +5619,90 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f0.checkFails("^arrays_overlap(array[1, 2], array[2])^",
+        "No match found for function signature ARRAYS_OVERLAP\\("
+            + "<INTEGER ARRAY>, <INTEGER ARRAY>\\)", false);
+
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
+    f.checkScalar("arrays_overlap(array[1, 2], array[2])", true,
+        "BOOLEAN NOT NULL");
+    f.checkScalar("arrays_overlap(array[1, 2], array[3])", false,
+        "BOOLEAN NOT NULL");
+    f.checkScalar("arrays_overlap(array[1, null], array[1])", true,
```
NobiGo, line 5635: I want to confirm this `arrays_overlap(array[1, null], array[1])` the datatype is Boolean not BOOLEAN NOT NULL?

NobiGo (commented):
LGTM. Only some minor problems need to be confirmed.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
liuyongvs: @NobiGo  
 Hexadecimal Literals
https://docs.oracle.com/cd/E17952_01/mysql-8.0-en/hexadecimal-literals.html

for example
```
spark-sql (default)> select x'61';
a
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
liuyongvs: yeap @NobiGo 
1) 
array[1, null] -> type is INT ARRAY NOT NULL
array[1] -> type is INT NOT NULL ARRAY NOT NULL
so the type is boolean

```
case class ArraysOverlap(left: Expression, right: Expression)
  override def nullable: Boolean = {
    left.nullable || right.nullable || leftArrayElementNullable || rightArrayElementNullable
  }
```

2) spark/postgres's typeof doesn't display not null/null info, but calcite/flink does. so the spark display it not full.
we should confirm the behavior by code

 you can look here https://issues.apache.org/jira/browse/FLINK-31906

```
spark-sql (default)> select typeof(arrays_overlap(array(1, null), array(1)));
boolean

spark-sql (default)> select typeof(arrays_overlap(array(1), array(1)));
boolean // don't display not null info.
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
NobiGo: I get it. Thanks.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
NobiGo: @liuyongvs Maybe we can test in Spark By:
`arrays_overlap(array[1, null], array[1]) IS NOT NULL` ?

liuyongvs: @NobiGo can not.
IS NOT NULL just judge whether the value is null
```
   for example arrays_overlap(array[1, null], array[1])  is true, so it is not null. then the result arrays_overlap(array[1, null], array[1]) IS NOT NULL is true
```
while the nullable is a constraint, just for whether it can store not null/null value.
the semantic is different

liuyongvs:
hi @NobiGo Thanks for your review, squash the commits.

NobiGo (approved):
LGTM

PR #3264, opened 2023-06-14
[CALCITE-5756] Expand ProjectJoinRemoveRule to support inner join remove by the foreign-unique constraint in the catalog

JingDas:
This feature is mainly about expanding ProjectJoinRemoveRule to support inner join remove
by the foreign-unique constraint in the catalog.

The main steps are as follows:
1. Analyse the column used by project, and then split them to left and right side.
2. Acccording to the project info above and outer join type, bail out in some scene.
3. Get join info such as join keys.
4. For inner join check foreign and unique keys, these may use
RelMetadataQuery#getForeignKeys(newly add, similar to RelMetadataQuery#getUniqueKeys),
RelOptTable#getReferentialConstraints.
5. Check removing side join keys are areColumnsUnique both for outer join and inner join.
6. If all done, calculate the fianl project and transform. 

Test cases are added to RelOptRulesTest and RelMetadataTest.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
@@ -17,6 +17,7 @@
 package org.apache.calcite.rel.metadata;
 
 import org.apache.calcite.plan.RelOptCost;
+import org.apache.calcite.plan.RelOptForeignKey;
 import org.apache.calcite.plan.RelOptPlanner;
 import org.apache.calcite.plan.RelOptPredicateList;
 import org.apache.calcite.plan.volcano.VolcanoPlanner;
@@ -100,6 +101,43 @@ public abstract class BuiltInMetadata {
           boolean ignoreNulls);
 
       @Override default MetadataDef<UniqueKeys> getDef() {
+        return DEF;
+      }
+    }
+  }
+
+  /**
+   * Metadata about which columns have foreign keys.
+   */
+  public interface ForeignKeys extends Metadata {
+    MetadataDef<ForeignKeys> DEF =
+        MetadataDef.of(ForeignKeys.class, ForeignKeys.Handler.class,
+            BuiltInMethod.FOREIGN_KEYS.method);
+
+    /**
+     * Extract foreign keys from {@link org.apache.calcite.rel.RelNode}.
+     * Foreign keys are represented as an {@link org.apache.calcite.util.ImmutableBitSet},
+     * where each bit position represents a 0-based output column ordinal.
+     *
+     * @param ignoreNulls if true, allow containing null values when determining
+     *                     whether the keys are foreign keys
+     *
+     * @return bit set of foreign keys, or empty if not enough information is
+     * available to make that determination
+     */
+    Set<RelOptForeignKey> getForeignKeys(boolean ignoreNulls);
+
+    /**
+     * Handler API.
+     */
+    @FunctionalInterface
+    interface Handler extends MetadataHandler<ForeignKeys> {
+      Set<RelOptForeignKey> getForeignKeys(
+          RelNode rel,
+          RelMetadataQuery mq,
+          boolean ignoreNulls);
+
+      @Override default MetadataDef<ForeignKeys> getDef() {
         return DEF;
       }
     }
@@ -837,6 +875,6 @@ public abstract class BuiltInMetadata {
   interface All extends Selectivity, UniqueKeys, RowCount, DistinctRowCount,
       PercentageOriginalRows, ColumnUniqueness, ColumnOrigin, Predicates,
       Collation, Distribution, Size, Parallelism, Memory, AllPredicates,
-      ExpressionLineage, TableReferences, NodeTypes {
+      ExpressionLineage, TableReferences, NodeTypes, ForeignKeys {
   }
 }
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
@@ -1,0 +1,339 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.metadata;
+
+import org.apache.calcite.plan.RelOptForeignKey;
+import org.apache.calcite.plan.RelOptTable;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.RelReferentialConstraint;
+import org.apache.calcite.rel.SingleRel;
+import org.apache.calcite.rel.core.Aggregate;
+import org.apache.calcite.rel.core.Calc;
+import org.apache.calcite.rel.core.Correlate;
+import org.apache.calcite.rel.core.Filter;
+import org.apache.calcite.rel.core.Join;
+import org.apache.calcite.rel.core.Project;
+import org.apache.calcite.rel.core.SetOp;
+import org.apache.calcite.rel.core.Sort;
+import org.apache.calcite.rel.core.TableModify;
+import org.apache.calcite.rel.core.TableScan;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rex.InferredRexTableInputRef;
+import org.apache.calcite.rex.RexInputRef;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexProgram;
+import org.apache.calcite.util.ImmutableBitSet;
+import org.apache.calcite.util.Pair;
+import org.apache.calcite.util.Util;
+import org.apache.calcite.util.mapping.IntPair;
+
+import com.google.common.collect.ImmutableListMultimap;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * RelMdForeignKeys supplies a default implementation of
+ * {@link RelMetadataQuery#getForeignKeys} for the standard logical algebra.
+ * The relNodes supported are same to {@link RelMetadataQuery#getUniqueKeys(RelNode)}
+ */
+public class RelMdForeignKeys
+    implements MetadataHandler<BuiltInMetadata.ForeignKeys> {
+  protected static final Set<RelOptForeignKey> EMPTY_BIT_SET = new HashSet<>();
+  public static final RelMetadataProvider SOURCE =
+      ReflectiveRelMetadataProvider.reflectiveSource(
+          new RelMdForeignKeys(), BuiltInMetadata.ForeignKeys.Handler.class);
+
+//~ Constructors -----------------------------------------------------------
+
+  private RelMdForeignKeys() {}
+
+//~ Methods ----------------------------------------------------------------
+
+  @Override public MetadataDef<BuiltInMetadata.ForeignKeys> getDef() {
+    return BuiltInMetadata.ForeignKeys.DEF;
+  }
+
+  public Set<RelOptForeignKey> getForeignKeys(Filter rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+    return mq.getForeignKeys(rel.getInput(), ignoreNulls);
+  }
+
+  public Set<RelOptForeignKey> getForeignKeys(Sort rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+    return mq.getForeignKeys(rel.getInput(), ignoreNulls);
+  }
+
+  public Set<RelOptForeignKey> getForeignKeys(Correlate rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+    return mq.getForeignKeys(rel.getLeft(), ignoreNulls);
+  }
+
+  public Set<RelOptForeignKey> getForeignKeys(TableModify rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+    return mq.getForeignKeys(rel.getInput(), ignoreNulls);
+  }
+
+  public Set<RelOptForeignKey> getForeignKeys(Join rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+    final RelNode left = rel.getLeft();
+    final RelNode right = rel.getRight();
+    if (!rel.getJoinType().projectsRight()) {
+      // only return the foreign keys from the LHS since a semi or anti join only
+      // returns the LHS
+      return mq.getForeignKeys(left, ignoreNulls);
+    }
+    int nLeftColumns = rel.getLeft().getRowType().getFieldList().size();
+    final Set<RelOptForeignKey> foreignKeys = new HashSet<>();
+    final Set<RelOptForeignKey> leftInputForeignKeys =
+        mq.getForeignKeys(left, ignoreNulls);
+    final Set<RelOptForeignKey> rightInputForeignKeys =
+        mq.getForeignKeys(right, ignoreNulls);
+
+    if (leftInputForeignKeys.isEmpty() && rightInputForeignKeys.isEmpty()) {
+      return EMPTY_BIT_SET;
+    }
+    // shift right index
+    Set<RelOptForeignKey> shiftedRightInputForeignKeys = rightInputForeignKeys.stream()
+        .map(
+            foreignKey -> foreignKey.shift(nLeftColumns,
+            RelOptForeignKey.ShiftSide.INFERRED_FOREIGN_SOURCE,
+            RelOptForeignKey.ShiftSide.INFERRED_UNIQUE_TARGET))
+        .collect(Collectors.toSet());
+    if (!rel.getJoinType().generatesNullsOnLeft() || ignoreNulls) {
+      foreignKeys.addAll(
+          tryMerge(leftInputForeignKeys, shiftedRightInputForeignKeys));
+    }
+    if (!rel.getJoinType().generatesNullsOnRight() || ignoreNulls) {
+      foreignKeys.addAll(
+          tryMerge(shiftedRightInputForeignKeys, leftInputForeignKeys));
+    }
+    return foreignKeys;
+  }
+
+  private Set<RelOptForeignKey> tryMerge(Set<RelOptForeignKey> foreignKeys,
+      Set<RelOptForeignKey> uniqueKeys) {
+    ImmutableMap.Builder<Set<InferredRexTableInputRef>, RelOptForeignKey>
+        inferredUniqueKeyMapBuilder = ImmutableMap.builder();
+    // The upper-level relational algebra may use constraints,
+    // so keep the unmerged foreign keys
+    final Set<RelOptForeignKey> mixedForeignKeys = new HashSet<>(foreignKeys);
+    mixedForeignKeys.addAll(uniqueKeys);
+    if (foreignKeys.isEmpty() || uniqueKeys.isEmpty()) {
+      return mixedForeignKeys;
+    }
+    // Build unique keys map, key -> unique table set, value -> unique relOptForeignKey
+    for (RelOptForeignKey uniqueKey : uniqueKeys) {
+      if (!uniqueKey.getConstraints().isEmpty()
+          && uniqueKey.isInferredUniqueKey()) {
+        inferredUniqueKeyMapBuilder.put(
+            Sets.newHashSet(RelOptForeignKey.constraintsRight(uniqueKey.getConstraints())),
+            uniqueKey);
+      }
+    }
+    ImmutableMap<Set<InferredRexTableInputRef>, RelOptForeignKey> inferredUniqueKeyMap =
+        inferredUniqueKeyMapBuilder.build();
+    for (RelOptForeignKey foreignKey : foreignKeys) {
+      if (foreignKey.getConstraints().isEmpty()
+          || !foreignKey.isInferredForeignKey()) {
+        continue;
+      }
+      // try merge
+      Set<InferredRexTableInputRef> foreignSideNeededUniqueSet =
+          Sets.newHashSet(RelOptForeignKey.constraintsRight(foreignKey.getConstraints()));
+      RelOptForeignKey inferredUniqueKey =
+          inferredUniqueKeyMap.get(foreignSideNeededUniqueSet);
+      if (inferredUniqueKey != null) {
+        mixedForeignKeys.add(
+            RelOptForeignKey.of(
+                foreignKey.getConstraints().stream()
+                    .map(
+                        constraint -> Pair.of(
+                        constraint.left.copy(true),
+                        constraint.right.copy(true)))
+                    .collect(Collectors.toList()),
+                ImmutableBitSet.of(foreignKey.getForeignColumns()),
+                ImmutableBitSet.of(inferredUniqueKey.getUniqueColumns())));
+      }
+    }
+    return mixedForeignKeys;
+  }
+
+  public Set<RelOptForeignKey> getForeignKeys(Aggregate rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+    final ImmutableBitSet groupSet = rel.getGroupSet();
+    if (groupSet.isEmpty()) {
+      return EMPTY_BIT_SET;
+    }
+    final Set<RelOptForeignKey> inputForeignKeys =
+        mq.getForeignKeys(rel.getInput(), ignoreNulls);
+    return inputForeignKeys.stream()
+        .filter(foreignKey -> filterValidateColumns(groupSet, foreignKey))
+        .collect(Collectors.toSet());
+  }
+
+  public Set<RelOptForeignKey> getForeignKeys(Project rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+    return getProjectForeignKeys(rel, mq, ignoreNulls, rel.getProjects());
+  }
+
+  public Set<RelOptForeignKey> getForeignKeys(Calc rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+    RexProgram program = rel.getProgram();
+    return getProjectForeignKeys(rel, mq, ignoreNulls,
+        Util.transform(program.getProjectList(), program::expandLocalRef));
+  }
+
+  private static Set<RelOptForeignKey> getProjectForeignKeys(SingleRel rel,
+      RelMetadataQuery mq,
+      boolean ignoreNulls,
+      List<RexNode> projExprs) {
+
+    // Single input can be mapped to multiple outputs
+    final ImmutableListMultimap.Builder<Integer, Integer> inToOutIndexBuilder =
+        ImmutableListMultimap.builder();
+    final ImmutableBitSet.Builder inColumnsBuilder = ImmutableBitSet.builder();
+    for (int i = 0; i < projExprs.size(); i++) {
+      RexNode projExpr = projExprs.get(i);
+      if (projExpr instanceof RexInputRef) {
+        int inputIndex = ((RexInputRef) projExpr).getIndex();
+        inToOutIndexBuilder.put(inputIndex, i);
+        inColumnsBuilder.set(inputIndex);
+      }
+    }
+    final ImmutableBitSet inColumnsUsed = inColumnsBuilder.build();
+    if (inColumnsUsed.isEmpty()) {
+      return EMPTY_BIT_SET;
+    }
+    final Map<Integer, List<Integer>> mapInToOutPos =
+        Maps.transformValues(inToOutIndexBuilder.build().asMap(), Lists::newArrayList);
+    final Set<RelOptForeignKey> inputForeignKeys =
+        mq.getForeignKeys(rel.getInput(), ignoreNulls);
+    if (inputForeignKeys.isEmpty()) {
+      return EMPTY_BIT_SET;
+    }
+    return inputForeignKeys.stream()
+        .filter(foreignKey -> filterValidateColumns(inColumnsUsed, foreignKey))
+        .flatMap(foreignKey -> foreignKey.permute(mapInToOutPos, mapInToOutPos).stream())
+        .collect(Collectors.toSet());
+  }
+
+  private static boolean filterValidateColumns(ImmutableBitSet columnsUsed,
+      RelOptForeignKey foreignKey) {
+    return (!foreignKey.getForeignColumns().isEmpty()
+        && columnsUsed.contains(foreignKey.getForeignColumns()))
+        || (!foreignKey.getUniqueColumns().isEmpty()
+        && columnsUsed.contains(foreignKey.getUniqueColumns()));
+  }
+
+  public Set<RelOptForeignKey> getForeignKeys(TableScan rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+    final RelOptTable table = rel.getTable();
+    final BuiltInMetadata.ForeignKeys.Handler handler =
+        table.unwrap(BuiltInMetadata.ForeignKeys.Handler.class);
+    if (handler != null) {
+      return handler.getForeignKeys(rel, mq, ignoreNulls);
+    }
+
+    final List<RelReferentialConstraint> referentialConstraints =
+        table.getReferentialConstraints();
+    final List<ImmutableBitSet> keys = table.getKeys();
+
+    final Set<RelOptForeignKey> foreignKeys = new HashSet<>();
+    if (referentialConstraints != null) {
+      foreignKeys.addAll(referentialConstraints.stream()
+          .map(constraint -> {
+            List<Pair<InferredRexTableInputRef, InferredRexTableInputRef>> constraints =
+                constraint.getColumnPairs().stream()
+                    .map(
+                        intPair -> Pair.of(
+                        InferredRexTableInputRef.of(constraint.getSourceQualifiedName(),
+                            intPair.source,
+                            false),
+                        InferredRexTableInputRef.of(constraint.getTargetQualifiedName(),
+                            intPair.target,
+                            false)))
+                    .collect(Collectors.toList());
+            return RelOptForeignKey.of(constraints,
+                ImmutableBitSet.of(IntPair.left(constraint.getColumnPairs())),
+                ImmutableBitSet.of());
+          })
+          .collect(Collectors.toSet()));
+    }
+    if (keys != null) {
+      foreignKeys.addAll(
+          keys.stream()
+              .map(keyBitSet -> {
+                List<Pair<InferredRexTableInputRef, InferredRexTableInputRef>> constraints =
+                    keyBitSet.asList().stream()
+                        .map(
+                            index -> Pair.of(
+                            InferredRexTableInputRef.of(),
+                            InferredRexTableInputRef.of(
+                                table.getQualifiedName(),
+                                index,
+                                false)))
+                        .collect(Collectors.toList());
+                return RelOptForeignKey.of(constraints, ImmutableBitSet.of(), keyBitSet);
+              })
+              .collect(Collectors.toList()));
+    }
+    if (!ignoreNulls) {
+      final List<RelDataTypeField> fieldList = rel.getRowType().getFieldList();
+      return foreignKeys.stream()
+          .filter(foreignKey -> foreignKey.getForeignColumns().asSet().stream()
+              .noneMatch(index -> fieldList.get(index).getType().isNullable())
+              && foreignKey.getUniqueColumns().asSet().stream()
+              .noneMatch(index -> fieldList.get(index).getType().isNullable()))
+          .collect(Collectors.toSet());
+    }
+    return foreignKeys;
+  }
+
+  /**
+   * The foreign keys of SetOp are precisely the intersection of its every
+   * input foreign keys.
+   */
+  public Set<RelOptForeignKey> getForeignKeys(SetOp rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+
+    Set<RelOptForeignKey> foreignKeys = new HashSet<>();
+    for (RelNode input : rel.getInputs()) {
+      Set<RelOptForeignKey> inputForeignKeys = mq.getForeignKeys(input, ignoreNulls);
+      if (inputForeignKeys.isEmpty()) {
+        return EMPTY_BIT_SET;
+      }
+      foreignKeys = foreignKeys.isEmpty()
+          ? inputForeignKeys : Sets.intersection(foreignKeys, inputForeignKeys);
+    }
+    return foreignKeys;
+  }
+
+  /** Catch-all rule when none of the others apply. */
+  public Set<RelOptForeignKey> getForeignKeys(RelNode rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+    // no information available
+    return EMPTY_BIT_SET;
+  }
+}
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMetadataQuery.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMetadataQuery.java
@@ -17,6 +17,7 @@
 package org.apache.calcite.rel.metadata;
 
 import org.apache.calcite.plan.RelOptCost;
+import org.apache.calcite.plan.RelOptForeignKey;
 import org.apache.calcite.plan.RelOptPredicateList;
 import org.apache.calcite.plan.RelOptTable;
 import org.apache.calcite.plan.volcano.VolcanoPlanner;
@@ -40,6 +41,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import static org.apache.calcite.linq4j.Nullness.castNonNull;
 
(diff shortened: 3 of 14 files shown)


core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
```diff
@@ -105,6 +105,44 @@ interface Handler extends MetadataHandler<UniqueKeys> {
...
+   */
+  public interface ForeignKeys extends Metadata {
+    MetadataDef<ForeignKeys> DEF =
+        MetadataDef.of(ForeignKeys.class, ForeignKeys.Handler.class,
+            BuiltInMethod.FOREIGN_KEYS.method);
+
+    /**
+     * Determines the list of foreign keys for this expression. Foreign keys are
+     * represented as an {@link org.apache.calcite.util.ImmutableBitSet}, where
+     * each bit position represents the column ordinal is foreign key.
```
asolimando, line 119: I am not sure I understand this sentence, is it something like "where each bit position represents the column ordinal for the foreign key" what you mean or similar?

I mean, I get that the bit position is encoding the ordinal position of each foreign key, but I am not sure about the wording.

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
```diff
@@ -105,6 +105,44 @@ interface Handler extends MetadataHandler<UniqueKeys> {
...
+     *
+     * @param ignoreNulls if true, ignore null values when determining
```
asolimando, line 121: Can you explain why we need this?

Maybe I am bit rusty on FK relationships, but I guess that `unknownAsFalse` semantics will apply (like for `WHERE` clauses), so there is no way a row can "point" to a `NULL` value in the parent table.

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
```diff
@@ -105,6 +105,44 @@ interface Handler extends MetadataHandler<UniqueKeys> {
...
+     *                    whether the keys are foreign keys
+     *
+     * @return set of foreign keys, or empty set if this information
+     * cannot be determined (whereas null indicates definitely
```
asolimando, line 125: Can you sketch an example for the "empty vs null" case here?

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
```diff
@@ -0,0 +1,246 @@
...
+import com.google.common.collect.ImmutableListMultimap;
+import com.google.common.collect.Maps;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * RelMdReferentialConstraits supplies a default implementation of
```
asolimando, line 50: Apart from the typo, I don't see any class with this name, can you clarify?

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
```diff
@@ -0,0 +1,246 @@
...
+    ImmutableBitSet foreignKeys = ImmutableBitSet.of();
+    for (RelNode input : rel.getInputs()) {
+      ImmutableBitSet inputForeignKeys = mq.getForeignKeys(input, ignoreNulls);
+      if (inputForeignKeys.isEmpty()) {
+        return EMPTY_BIT_SET;
+      }
+      foreignKeys = foreignKeys.isEmpty()
+          ? inputForeignKeys : foreignKeys.intersect(inputForeignKeys);
+    }
+    if (foreignKeys.isEmpty()) {
```
asolimando, line 221: I think this is redundant, if `isEmpty()` returns true, no need to return an empty bitset

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
```diff
@@ -0,0 +1,246 @@
...
+        table.unwrap(BuiltInMetadata.ForeignKeys.Handler.class);
+    if (handler != null) {
+      return handler.getForeignKeys(rel, mq, ignoreNulls);
+    }
+
+    final List<RelReferentialConstraint> referentialConstraints =
+        table.getReferentialConstraints();
+    if (referentialConstraints == null || referentialConstraints.size() == 0) {
+      return EMPTY_BIT_SET;
+    }
```
asolimando, line 193:
```suggestion
    if (referentialConstraints == null || referentialConstraints.isEmpty()) {
      return EMPTY_BIT_SET;
    }
```

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
```diff
@@ -0,0 +1,246 @@
...
+    final List<IntPair> foreignUniquePair = referentialConstraints.stream()
```
asolimando, line 194: If you really want unique, you have to remove duplicates (using a `Set<IntPair>` possibly).

If you don't need unique, please update the field name not to confuse readers.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
```diff
@@ -0,0 +1,246 @@
...
+        .map(RelReferentialConstraint::getColumnPairs)
+        .flatMap(Collection::stream)
+        .collect(Collectors.toList());
+
+    List<Integer> foreignKeys = IntPair.left(foreignUniquePair);
+    if (!ignoreNulls) {
```
asolimando, line 200: Looks the opposite to me, if `ignoreNulls` is true, then you want to filter them out (executing the body of the if), no?

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    ImmutableBitSet allOutColumns = ImmutableBitSet.of();
+    for (int index : inputForeignKeys.asList()) {
+      ImmutableBitSet outColumns = mapInToOutPos.get(index);
+      if (outColumns != null && !outColumns.isEmpty()) {
+        allOutColumns = allOutColumns.union(outColumns);
+      }
+    }
+    return allOutColumns.isEmpty() ? EMPTY_BIT_SET : allOutColumns;
```
asolimando, line 177: Isn't it equivalent to:
```suggestion
    return allOutColumns;
```
?

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
```diff
@@ -0,0 +1,246 @@
...
+      ImmutableBitSet.Builder rightOutForeignKeys = ImmutableBitSet.builder();
+      for (int index : rightInputForeignKeys.asList()) {
+        rightOutForeignKeys.set(index + nLeftColumns);
+      }
+      outForeignKeys = outForeignKeys.union(rightOutForeignKeys.build());
+    }
+    return outForeignKeys;
+  }
+
+  public ImmutableBitSet getForeignKeys(Aggregate rel, RelMetadataQuery mq,
```
asolimando, line 116: What if in the aggregation we have something other than grouping sets?

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
```diff
@@ -0,0 +1,246 @@
...
+
+  public ImmutableBitSet getForeignKeys(TableModify rel, RelMetadataQuery mq,
+      boolean ignoreNulls) {
+    return mq.getForeignKeys(rel.getInput(), ignoreNulls);
+  }
+
+  public ImmutableBitSet getForeignKeys(Join rel, RelMetadataQuery mq, boolean ignoreNulls) {
+    final RelNode left = rel.getLeft();
+    final RelNode right = rel.getRight();
+    if (!rel.getJoinType().projectsRight()) {
```
asolimando, line 92: `projectRight()` is true for (left?) anti joins too, would this be ok for them as well?

If so, maybe we can make the comment more generic.

core/src/main/java/org/apache/calcite/rel/metadata/RelMetadataQuery.java
```diff
@@ -505,6 +509,29 @@ public static RelMetadataQuery instance() {
...
+   * Returns the
+   * {@link BuiltInMetadata.ForeignKeys#getForeignKeys(boolean)}
+   * statistic.
+   *
+   * @param rel         the relational expression
+   * @param ignoreNulls if true, ignore null values when determining
+   *                    whether the keys are foreign keys
+   *
+   * @return            set of keys, or empty set if this information cannot be determined
+   *                    (whereas empty set indicates definitely no keys at all)
```
asolimando, line 522: In `BuiltInMetadata` you have a different javadoc for return, please check and align them

core/src/main/java/org/apache/calcite/rel/metadata/RelMetadataQuery.java
```diff
@@ -505,6 +509,29 @@ public static RelMetadataQuery instance() {
...
+   */
+  public ImmutableBitSet getForeignKeys(RelNode rel, boolean ignoreNulls) {
+    for (;;) {
+      try {
+        return foreignKeysHandler.getForeignKeys(rel, this, ignoreNulls);
+      } catch (MetadataHandlerProvider.NoHandler e) {
+        foreignKeysHandler =
+            revise(BuiltInMetadata.ForeignKeys.Handler.class);
```
asolimando, line 530: No need to split the line I think:

```suggestion
        foreignKeysHandler = revise(BuiltInMetadata.ForeignKeys.Handler.class);
```

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java
```diff
@@ -86,6 +89,12 @@ private void registerTableEmp(MockTable empTable, Fixture fixture) {
     empTable.addColumn("COMM", fixture.intType);
     empTable.addColumn("DEPTNO", fixture.intType);
     empTable.addColumn("SLACKER", fixture.booleanType);
+    empTable.setReferentialConstraints(
+        Lists.newArrayList(
+            RelReferentialConstraintImpl.of(
+        Lists.newArrayList(DEFAULT_CATALOG, DEFAULT_SCHEMA, "EMP"),
+        Lists.newArrayList(DEFAULT_CATALOG, DEFAULT_SCHEMA, "DEPT"),
+        Lists.newArrayList(IntPair.of(7, 0)))));
```
asolimando, line 97: Might help to state what columns are linked for clarity (in a comment)

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java
```diff
@@ -99,6 +108,12 @@ private void registerTableEmpNullables(MockTable empNullablesTable, Fixture fixt
     empNullablesTable.addColumn("COMM", fixture.intTypeNull);
     empNullablesTable.addColumn("DEPTNO", fixture.intTypeNull);
     empNullablesTable.addColumn("SLACKER", fixture.booleanTypeNull);
+    empNullablesTable.setReferentialConstraints(
+        Lists.newArrayList(
+            RelReferentialConstraintImpl.of(
+        Lists.newArrayList(DEFAULT_CATALOG, DEFAULT_SCHEMA, "EMPNULLABLES"),
+        Lists.newArrayList(DEFAULT_CATALOG, DEFAULT_SCHEMA, "DEPT"),
+        Lists.newArrayList(IntPair.of(7, 0)))));
```
asolimando, line 116: Same here, a comment mentioning the column name from both tables would help clarify

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -20,27 +20,35 @@
...
 import java.util.stream.Collectors;
 
 /**
  * Planner rule that matches an {@link Project}
- * on a {@link Join} and removes the join provided that the join is a left join
- * or right join and the join keys are unique.
+ * on a {@link Join} and removes the join provided that the join is a left or
+ * right join and the join keys are unique, and removes the join provided that
+ * the join is inner join and the join keys are foreign and unique key,
+ * foreign key is not nullable.
```
asolimando, line 51: Nit: I am no native speaker so I don't have a strong opinion on this, but the following reads "better" to me, feel free to adopt or not:
```suggestion
 * and the foreign key is not nullable.
```

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -20,27 +20,35 @@
...
  *
  * <p>For instance,
```
asolimando, line 53: We could add another example showing your new feature here, I think it would help understand the textual description (same as the original example did for the original description)

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -78,40 +86,56 @@ public ProjectJoinRemoveRule(
...
 
-    final List<Integer> leftKeys = new ArrayList<>();
-    final List<Integer> rightKeys = new ArrayList<>();
-    RelOptUtil.splitJoinCondition(join.getLeft(), join.getRight(),
-        join.getCondition(), leftKeys, rightKeys,
-        new ArrayList<>());
+    // Check project range
+    ImmutableBitSet projectBits = RelOptUtil.InputFinder.bits(project.getProjects(), null);
+    boolean onlyUseLeft = projectBits.asList().stream()
+        .allMatch(i -> i >= 0 && i < join.getLeft().getRowType().getFieldCount());
```
asolimando, line 95: Please extract the common expressions like `join.getLeft().getRowType().getFieldCount()`, the lambda is already long enough

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -78,40 +86,56 @@ public ProjectJoinRemoveRule(
...
+          leftKeys, rightKeys, leftForeignKeys, mq, join.getLeft(), join.getRight())) {
+        return;
+      }
+      if (onlyUseRight && !areForeignKeysValid(
+          rightKeys, leftKeys, rightForeignKeys, mq, join.getRight(), join.getLeft())) {
+        return;
+      }
+    }
+
+    BooleanSupplier isLeftSideReserved = () -> isLeftJoin || (isInnerJoin && onlyUseLeft);
```
asolimando, line 129: I don't get the need for a `BooleanSupplier` here, can't we achieve the same with a plain `boolean`?

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -78,40 +86,56 @@ public ProjectJoinRemoveRule(
...
+    final List<Integer> joinKeys = isLeftSideReserved.getAsBoolean() ? rightKeys : leftKeys;
     if (!Boolean.TRUE.equals(
```
asolimando, line 131:
```suggestion
    if (Boolean.FALSE.equals(
```

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -78,40 +86,56 @@ public ProjectJoinRemoveRule(
...
+            ImmutableBitSet.of(joinKeys)))) {
       return;
     }
 
     RelNode node;
-    if (isLeftJoin) {
+    if (isLeftSideReserved.getAsBoolean()) {
       node = project
           .copy(project.getTraitSet(), join.getLeft(), project.getProjects(),
               project.getRowType());
```
asolimando, line 141: Style check doesn't complain but it feels weird to indent like that fluent expressions with a single call, what about the following?

```suggestion
      node = project.copy(project.getTraitSet(), join.getLeft(), project.getProjects(), 
          project.getRowType());
```

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -127,6 +151,49 @@ public ProjectJoinRemoveRule(
     call.transformTo(node);
   }
 
+  /**
+   * Check as following:
+   * 1. Make sure that every foreign column always be foreign key.
+   * 2. The target of foreign key is the correct unique key correspondingly.
```
asolimando, line 157: Same as per the other "linguistic" comments:

```suggestion
   * 2. The target of foreign key is the correct corresponding unique key.
```

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -127,6 +151,49 @@ public ProjectJoinRemoveRule(
...
+   */
+  private static boolean areForeignKeysValid(List<Integer> foreignColumns,
+      List<Integer> uniqueColumns,
+      ImmutableBitSet foreignKeys,
+      RelMetadataQuery mq,
+      RelNode foreignSideRel,
+      RelNode uniqueSideRel) {
```
asolimando, line 164: As long as stylecheck passes I am OK, but I have rarely seen in Calcite formatting each parameter on a newline like that, I think the following could also work:

```suggestion
  private static boolean areForeignKeysValid(List<Integer> foreignColumns,
      List<Integer> uniqueColumns, ImmutableBitSet foreignKeys, RelMetadataQuery mq,
      RelNode foreignSideRel, RelNode uniqueSideRel) {
```

Not necessary to change if nobody else complains and you like this style.

testkit/src/main/java/org/apache/calcite/test/RelMetadataFixture.java
```diff
@@ -264,6 +265,24 @@ public RelMetadataFixture assertPercentageOriginalRows(Matcher<Double> matcher)
     return this;
   }
 
+  public RelMetadataFixture assertForeignKeys(
+      Matcher<ImmutableBitSet> constraintMatcher) {
+    RelNode rel = toRel();
+    RelMetadataQuery mq = rel.getCluster().getMetadataQuery();
+    ImmutableBitSet foreignKeys = mq.getForeignKeys(rel, false);
```
asolimando, line 272: It seems that we are not covering cases with `ignoreNulls` to `true`, can we add such tests?

testkit/src/main/java/org/apache/calcite/test/RelMetadataFixture.java
```diff
@@ -264,6 +265,24 @@ public RelMetadataFixture assertPercentageOriginalRows(Matcher<Double> matcher)
     return this;
   }
 
+  public RelMetadataFixture assertForeignKeys(
+      Matcher<ImmutableBitSet> constraintMatcher) {
```
asolimando, line 269:
```suggestion
  public RelMetadataFixture assertForeignKeys(Matcher<ImmutableBitSet> constraintMatcher) {
```

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -273,6 +273,104 @@ final RelMetadataFixture sql(String sql) {
                 / (DEPT_SIZE + EMP_SIZE)));
   }
 
+  // ----------------------------------------------------------------------
+  // Tests for getForeignKeys
+  // ----------------------------------------------------------------------
+
+  @Test void testForeignKeysAggregateCallEmpty() {
```
asolimando, line 280: What's the difference with the test right below this one?

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -273,6 +273,104 @@ final RelMetadataFixture sql(String sql) {
...
+    sql("select ename, job from emp")
+        .assertForeignKeysAreEmpty();
+  }
+
+  @Test void testForeignKeysExpressionEmpty() {
+    sql("select deptno + 1, ename from emp")
+        .assertForeignKeysAreEmpty();
+  }
+
+  @Test void testForeignKeysFilter() {
```
asolimando, line 310: Maybe add a test using the FK column in the filter too?

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -273,6 +273,104 @@ final RelMetadataFixture sql(String sql) {
...
+        .assertForeignKeys(equalTo(bitSetOf(1, 2)));
+  }
+
+  @Test void testForeignKeysJoinLeftOuter() {
+    sql("select name as dname, emp.deptno, dept.name from emp left outer join dept"
+        + " on emp.deptno = dept.deptno")
+        .assertForeignKeys(equalTo(bitSetOf(1)));
+  }
+
+  @Test void testForeignKeysJoinRightOuter() {
```
asolimando, line 331:
```suggestion
  @Test void testForeignKeysRightOuterJoin() {
```

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -273,6 +273,104 @@ final RelMetadataFixture sql(String sql) {
...
+    sql("select name as dname, emp.deptno, dept.name, emp.deptno from dept right outer join emp"
```
asolimando, line 332: Can we also add left/right outer join but with the opposite tables' order?

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -273,6 +273,104 @@ final RelMetadataFixture sql(String sql) {
...
+        + " on emp.deptno = dept.deptno")
+        .assertForeignKeys(equalTo(bitSetOf(1, 3)));
+  }
+
+  @Test void testForeignKeysJoinOuterEmpty() {
```
asolimando, line 337:
```suggestion
  @Test void testForeignKeysOuterJoinEmpty() {
```

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -273,6 +273,104 @@ final RelMetadataFixture sql(String sql) {
...
+    sql("select name as dname, emp.deptno, dept.name from dept left outer join emp"
+        + " on emp.deptno = dept.deptno")
+        .assertForeignKeysAreEmpty();
+  }
+
+  @Test void testForeignKeysJoinFullOuterEmpty() {
```
asolimando, line 343:
```suggestion
  @Test void testForeignKeysFullOuterJoinEmpty() {
```

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6019,6 +6019,87 @@ private HepProgram getTransitiveProgram() {
         .check();
   }
```
asolimando, line 6021: Can we add some queries where the simplification happens having also some filters?

asolimando (changes requested):
LGTM, lots of minor comments and some missing tests but the PR is in good shape and the contribution is very valuable, will get there in few iterations, thanks @JingDas for your work!

asolimando:
Almost forgot: the commit message must match the title of the Jira ticket, at the moment they are different

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
JingDas: You are right, the getForeignKeys method is similar to getUniqueKeys.   The result of getForeignKeys method represent the foreign keys pos above the current relNode, and the index is "0-based".   My comment was misunderstood, I'll edit it.

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
JingDas: In our inner join removing scene. It will make a mistake when forein key is nullable. because after removing inner join side which contains unique keys, If foreign key is nullable, the result will contain some null foreign key result wrongly. 
In other scene, it maybe want foreign keys containg null. 
And BuiltInMetadata.ForeignKeys is a common component, so i add "ignoreNulls"option for getForeignKeys method. 
BuiltInMetadata.ForeignKeys can be used for different scene, and in our scene it should not contains null.

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
JingDas: I make a mistake. It should return empty ImmutableBitSet but never null If the current relNode(mainly underlying tableScan relNode correspondingly) does't contains foreign keys. I will fix it.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
JingDas: Yeah, you are right. "RelMdReferentialConstraits" should be "RelMdForeignKeys". I will fix it.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
JingDas: Yeah, it is also redundant somewhere else. I will fix them

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
JingDas: OK, got your point. 
I think we can represent multi foreign keys in a  bit set if a table has multi referential constraints. 
And retain the right index position. 
For example:
`SELECT ename, deptno, deptno
FROM emp `
where deptno is foreign key. 
We should returen {1, 2} when call getForeignKeys method.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
JingDas: I also consider it some times. Actually I get the idea from `UniqueKeys#(boolean ignoreNulls)`. 
My understanding is that if `ignoreNulls ` is true, we allow result contains null. 
The logical is consistent to `UniqueKeys#(boolean ignoreNulls)`. 
Maybe it's better to use "containNulls". 
If "containNulls" is true, we allow result contains null.
If "containNulls" is false, we do not allow result contains null.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
JingDas: ForieignKeys info are available in a table which is often the bottom part fo a whole relNode.
The foreign key may be invalid when propagate derivation from the bottom to up.
I think that a foreign key propagate from the bottom to up in aggregate,
if it is in the groupset, it can maintain the validity，
if it is in the other position except groupset, it can't maintain the validity, such as count(), sum() aggregate function.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
JingDas: It‘s exactly same, i will fix it.

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java
```diff
@@ -468,6 +483,9 @@ private void registerStructTypeTables(Fixture fixture) {
             "customBigInt"),
         typeFactory -> typeFactory.createSqlType(SqlTypeName.BIGINT));
 
+    // Register "DEPT" table.
```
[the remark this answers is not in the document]
JingDas: For this case,
my consideration is that we should need to know the referenced table and the unique key
before creating the foreign key.
It dosn't cause a exception when we register dept table after emp table in the test.
Out of habit, I prefer to register the dept table first and then the emp table.
It also feels OK to keep the previous registration order.

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
JingDas: This assertion is used in three places and `BooleanSupplier` is added primarily for reuse,
maybe it's better. What do you think?

testkit/src/main/java/org/apache/calcite/test/RelMetadataFixture.java
JingDas: As the [discussion](https://github.com/apache/calcite/pull/3264#discussion_r1230251674) above, maybe we don't need to add a test where `ignoreNulls` equals `true`.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
JingDas:
```
 public boolean projectsRight() {
    return this != SEMI && this != ANTI;
 }
```
This code above is in `JoinRelType`.
If we want to get right side projects from join, it can't be anti or semi join.
So if `projectsRight` is true, we can get right project,
if `projectsRight` is false, we can't get right projects, can only get left projects.
Only semi join is mentioned in this comment, I will also add anti join to the comment.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
JingDas: Add testProjectJoinRemove17 and testProjectJoinRemove18 case.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
JingDas: Add testForeignKeysFilter and testForeignKeysFilterEmpty case.

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
JingDas: Add "Another instance" comment.

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java
JingDas: Add the comment.

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java
JingDas: Yeah, add it.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -273,6 +273,104 @@ final RelMetadataFixture sql(String sql) {
...
+        + "where emp_agg.ename = 'job'")
+        .assertForeignKeys(equalTo(bitSetOf(1)));
+  }
+
+  @Test void testForeignKeysValuesEmpty() {
+    sql("values(1,2,3)")
+        .assertForeignKeysAreEmpty();
+  }
+
+  @Test void testForeignKeysUnion() {
```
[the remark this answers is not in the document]
JingDas: Add it

core/src/main/java/org/apache/calcite/rel/metadata/RelMetadataQuery.java
JingDas: Adjust them to be the same.

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
JingDas: I am not a native speaker, Thanks for your comment grammar check.

asolimando:
@JingDas thanks for addressing the comments, I am on a business trip at the moment, I should be able to review again by the end of this week.

FYI: when you are done with comments and you want the reviewers to take another look you can also use the "request review" button close to the reviewers' name in the top-right part of the GitHub page

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
```diff
@@ -174,72 +176,88 @@ private static ImmutableBitSet getProjectForeignKeys(SingleRel rel, RelMetadataQ
...
       foreignKeys = foreignKeys.isEmpty()
           ? inputForeignKeys : foreignKeys.intersect(inputForeignKeys);
     }
-    if (foreignKeys.isEmpty()) {
-      return EMPTY_BIT_SET;
+    return foreignKeys;
+  }
+
+  /**
+   * The foreign keys of Intersect are precisely the union set of its every
```
asolimando, line 233: I think it's the same as per `Union`, that is, the intersection of the FKs from the operands. Because you can apply this set operations on whatever pair of relations having the "same" schema type, but you can very well mix up columns with different FK relationship for the same position.

The same applies to `Minus`, so you probably want to revert back to a single implementation for `SetOp` and revise the tests.

In any case, `areForeignKeysValid` will bail out on such cases, as it will try to track down the original column via `RelMetadataQuery.getColumnOrigin()`, and if a mix is found, more origins will be detected, and it will return `null`.

Since it's a bit tricky I suggest to add at least some basic tests to `RelOptRulesTest` per each set operator to confirm our understanding (simple case and tricky one like a mix of column origin under the same position).

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -127,6 +163,46 @@ public ProjectJoinRemoveRule(
...
+    if (!foreignKeys.contains(ImmutableBitSet.of(foreignColumns))) {
+      return false;
+    }
+    List<RelReferentialConstraint> referentialConstraints;
+    for (IntPair foreignUniqueKey : IntPair.zip(foreignColumns, uniqueColumns)) {
+      RelColumnOrigin foreignOrigin = mq.getColumnOrigin(foreignSideRel, foreignUniqueKey.source);
+      RelColumnOrigin uniqueOrigin = mq.getColumnOrigin(uniqueSideRel, foreignUniqueKey.target);
+      if (foreignOrigin == null || uniqueOrigin == null) {
+        return false;
+      }
```
asolimando, line 186: You want to bail out as soon as possible in rules, and doing so also spares a call to `RelMetadataQuery` which are generally expensive:
```suggestion
      RelColumnOrigin foreignOrigin = mq.getColumnOrigin(foreignSideRel, foreignUniqueKey.source);
      if (foreignOrigin == null) {
        return false;
      }
      RelColumnOrigin uniqueOrigin = mq.getColumnOrigin(uniqueSideRel, foreignUniqueKey.target);
      if (uniqueOrigin == null) {
        return false;
      }
```

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -78,47 +99,62 @@ public ProjectJoinRemoveRule(
...
+      final ImmutableBitSet leftForeignKeys = mq.getForeignKeys(join.getLeft(), false);
+      final ImmutableBitSet rightForeignKeys = mq.getForeignKeys(join.getRight(), false);
+      if (onlyUseLeft && !areForeignKeysValid(
+          leftKeys, rightKeys, leftForeignKeys, mq, join.getLeft(), join.getRight())) {
+        return;
+      }
+      if (onlyUseRight && !areForeignKeysValid(
+          rightKeys, leftKeys, rightForeignKeys, mq, join.getRight(), join.getLeft())) {
+        return;
+      }
```
asolimando, line 139: Bail out fast and avoids calling `getForeignKeys` for the right side if you should already bail out on left:
```suggestion
            final ImmutableBitSet leftForeignKeys = mq.getForeignKeys(join.getLeft(), false);
      if (onlyUseLeft && !areForeignKeysValid(
          leftKeys, rightKeys, leftForeignKeys, mq, join.getLeft(), join.getRight())) {
        return;
      }
      final ImmutableBitSet rightForeignKeys = mq.getForeignKeys(join.getRight(), false);
      if (onlyUseRight && !areForeignKeysValid(
          rightKeys, leftKeys, rightForeignKeys, mq, join.getRight(), join.getLeft())) {
        return;
      }
```

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -78,47 +99,62 @@ public ProjectJoinRemoveRule(
...
-    RelOptUtil.splitJoinCondition(join.getLeft(), join.getRight(),
-        join.getCondition(), leftKeys, rightKeys,
-        new ArrayList<>());
+    // Check project range
+    ImmutableBitSet projectBits = RelOptUtil.InputFinder.bits(project.getProjects(), null);
+    final int leftFieldsNum = join.getLeft().getRowType().getFieldCount();
+    final boolean onlyUseLeft = projectBits.asList().stream()
+        .allMatch(i -> i >= 0 && i < leftFieldsNum);
+    final boolean onlyUseRight = projectBits.asList().stream().allMatch(i -> i >= leftFieldsNum
+            && i < join.getRowType().getFieldCount());
```
asolimando, line 111: To match the style you used above:
```suggestion
    final boolean onlyUseRight = projectBits.asList().stream()
        .allMatch(i -> i >= leftFieldsNum && i < join.getRowType().getFieldCount());
```

testkit/src/main/java/org/apache/calcite/test/RelMetadataFixture.java
asolimando: I see your point, but we are still supporting a "mode" we aren't testing, and which comes with bits of code, so it needs proper testing like any other part of our codebase, we should have tests for `ignoreNulls = true` too.

asolimando (changes requested):
I have reviewed your changes, I have some further comments but we are converging.

I'd ask you the favour of not resolving comments on your side, as it makes it very tricky for me to track if what was asked has been addressed or not, I will mark them as resolved myself, thanks!

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
JingDas: OK, I get it.

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
JingDas: You are right, I get it, I will fix it.

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
```diff
@@ -128,21 +141,20 @@ public ProjectJoinRemoveRule(
...
     }
 
     RelNode node;
     if (isLeftSideReserved.getAsBoolean()) {
-      node = project
-          .copy(project.getTraitSet(), join.getLeft(), project.getProjects(),
-              project.getRowType());
+      node =
+          project.copy(project.getTraitSet(), join.getLeft(),
+              project.getProjects(), project.getRowType());
```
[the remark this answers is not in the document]
JingDas: I think you are right, I also agree with you. 
This newline operation is edited by check style.
I'll give it a try as we agreed.

core/src/main/java/org/apache/calcite/rel/rules/ProjectJoinRemoveRule.java
JingDas: OK, I agree with you. Thanks for the suggestion.

testkit/src/main/java/org/apache/calcite/test/RelMetadataFixture.java
JingDas: OK, got your point, I will add it.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdForeignKeys.java
JingDas: OK, I would fall back to using `ignoreNulls`.

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
```diff
@@ -114,18 +114,18 @@ public interface ForeignKeys extends Metadata {
...
-     *                    whether the keys are foreign keys
+     * @param containNulls if true, allow containing null values when determining
+     *                     whether the keys are foreign keys
      *
-     * @return set of foreign keys, or empty set if this information
-     * cannot be determined (whereas null indicates definitely
-     * no constraint at all)
+     * @return bit set of foreign keys, or empty if not enough information is
+     * available to make that determination (whereas empty indicates definitely
+     * no foreign keys at all)
```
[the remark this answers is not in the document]
JingDas: OK, I will remove it.

asolimando (approved):
LGTM, can you squash your changes and make sure the only commit has `[CALCITE-5756] Expand ProjectJoinRemoveRule to support inner join removal by using the foreign-unique constraints` matching the commit message like the associated Jira ticket?

I will merge in the next 24 hours if no objections arise.

As a side note, avoid rebasing (or squashing) until the very end of the review as it makes it harder to check changes incrementally.

asolimando:
Adding "request review" from another committer given the comments in the associated Jira ticket.

JingDas:
> > The `RelOptForeignKey` component has been added to represent the composite or single foreign key and the unique constraint that foreign key references. Sorry to bother you, @asolimando @julianhyde would you help me to review the code?
> 
> Thanks @JingDas, I will review again as soon as I have some time.
> 
> In the meantime, I have approved the pending workflows, but I have also noticed that there are 18 code smells detected by SonarCloud, could you do a pass on this list and fix whatever can be reasonably fixed?
> 
> As for any automatic tool not everything makes sense and there might even be false positives.

JingDas:
Sorry, it was closed by mistake，I reOpen it and fixed the most necessary SonarCloud detected code.

JingDas:
Aha! It seems calcite main code change, And the CI seems to use the new [main logic](https://github.com/apache/calcite/blob/3ab8003106507b1664ad2c42860789776a816cd3/testkit/src/main/java/org/apache/calcite/test/RelOptFixture.java#L394C1-L400C45). Should I try to rebase the calcite upstream main code and commit it again?

JingDas:
> Aha! It seems calcite main code change, And the CI seems to use the new [main logic](https://github.com/apache/calcite/blob/3ab8003106507b1664ad2c42860789776a816cd3/testkit/src/main/java/org/apache/calcite/test/RelOptFixture.java#L394C1-L400C45). I will fix it according to main code usage.

Have fixed it.

mihaibudiu:
This PR was approved, but never merged. What's the plan?

asolimando (changes requested):
@mihaibudiu, discussion in the associated Jira ticket led to creating [CALCITE-5881](https://issues.apache.org/jira/browse/CALCITE-5881), which needs to be addressed before getting this in, unfortunately there was no progress on it for a long time now. I will remove the approval to make the situation clearer, thanks for the reminder.

PR #3268, opened 2023-06-15
[CALCITE-5783] Make the TableFunctionScan hintable

Aitozi:
This PR is meant to make the `TableFunctionScan` hintable

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/SqlHintsConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlHintsConverterTest.java
@@ -38,6 +38,7 @@ import org.apache.calcite.rel.core.Filter;
 import org.apache.calcite.rel.core.Join;
 import org.apache.calcite.rel.core.JoinInfo;
 import org.apache.calcite.rel.core.Snapshot;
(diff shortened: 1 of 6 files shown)


core/src/test/java/org/apache/calcite/test/SqlHintsConverterTest.java
```diff
@@ -931,6 +937,12 @@ private static class HintCollector extends RelShuttleImpl {
           if (snapshot.getHints().size() > 0) {
             this.hintsCollect.add("Snapshot:" + snapshot.getHints());
           }
+        } else if (other instanceof LogicalTableFunctionScan) {
```
libenchao, line 940: Prefer to use the base class `TableFunctionScan`.

libenchao (commented):
@Aitozi Thanks for your PR, it looks great, I only have one minor comment.

core/src/test/java/org/apache/calcite/test/SqlHintsConverterTest.java
Aitozi: Updated

libenchao (approved):
+1

PR #3269, opened 2023-06-16
[CALCITE-5784] Generate the same correlationId for the same query

Aitozi:
Add a map to cache the correlarionId in the `SqlToRelConverter` to deduplicate correlationId for the same scope of identifier

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -278,6 +278,13 @@ public class SqlToRelConverter {
       new HashMap<>();
 
   /**
+   * The scope and identifier to the correlationId to deduplicate the creation of the
+   * correlationId.
+   */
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -277,6 +277,13 @@ public class SqlToRelConverter {
   private final Map<CorrelationId, DeferredLookup> mapCorrelToDeferred =
       new HashMap<>();
 
+  /**
+   * The scope and identifier to the correlationId to deduplicate the creation of the
+   * correlationId.
+   */
+  private final Map<Pair<SqlValidatorScope, String>, CorrelationId> correlationIdMap =
```
Aitozi, line 284: This map will rely on the object identifier of the scope, in the test case I posted, It can work well. But I'm not sure whether it's a good enough solution to reply on this object identifier here. 

Looking forward to some suggestion, thanks.

PR #3270, opened 2023-06-16
[CALCITE-5779] Implicit column alias for single-column table function should work

dssysolyatin:


core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -7598,6 +7598,18 @@ void testGroupExpressionEquivalenceParams() {
         .fails("Cannot call table function here: 'RAMP'");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5779">[CALCITE-5779]
+   * Implicit column alias for single-column table function should work</a>. */
+  @Test void testTableFunctionSingleColumnAlias() {
+    final SqlValidatorFixture s = fixture()
+        .withOperatorTable(MockSqlOperatorTable.standard().extend());
+    s.withSql("select rmp from table(ramp(3)) as rmp").ok();
```
JiajunBernoulli, line 7607: Should we need `SqlToRelConverterTest` or `QuidemTest` to confirm plan is right?

dssysolyatin: More tests never hurt. I will add couple of tests to SqlToRelConverterTest and QuidemTest

JiajunBernoulli (approved):

PR #3271, opened 2023-06-17
[CALCITE-5741] Add support for CONCAT_WS function(enabled in MSSQL, MySQL and Postgres)

ILuffZhe:
CONCAT_WS function returns a string resulting from the concatenation, or joining, of two or more string values in an end-to-end manner.
It separates those concatenated string values with the delimiter specified in the **first function argument**. e.g:

- CONCAT_WS(',', 'c', 'h', 'a', 'r') returns 'c,h,a,r'

One thing need to be noted: this function has two NULL semantics for the separator argument in different Database products. Let's take CONCAT_WS(null, 'a', 'b') as an example:
- Postgres and MySQL: **it returns null when the separator is null**, otherwise treats null as empty string. So the above SQL returns null.

- MSSQL: never returns null and treats null as empty string, the above SQL returns "ab".


ILuffZhe:
@julianhyde Thanks for your review! LGTM.
I've changed the PR title to make it consistent with JIRA, and in functions.iq(line 106) the JIRA title should be changed too.

PR #3273, opened 2023-06-18
[CALCITE-5786] Update QuidemTest and DiffRepository to write intermediate test files to directories other than build/resources/test. This addresses an incremental builds issue with the tests

akshayabd:
Writing to build/resources/test causes issues with incremental build support, ideally these directories shouldn't be written to once the tests have started.

With this change the intermediate files used by tests are written to other directories in the build folder. These locations are passed to the tests through system properties which are marked internal - so they don't factor into input hash key creation for the tests.

I don't know if there was a specific reason why the files were being written to build/resources/test - I'm assuming it was an oversight. Please review this change with care to ensure there wasn't a reason why the behavior was this way.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -812,6 +812,10 @@ allprojects {
                         passProperty(e)
                     }
                 }
+                jvmArgumentProviders.add(
+                    AdditionalTestSystemPropertiesProvider(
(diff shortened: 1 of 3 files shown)


akshayabd:
cc @clayburn @runningcode

akshayabd:
The Jira issue has screenshots and more details on the issue:
https://issues.apache.org/jira/browse/CALCITE-5786

PR #3188, opened 2023-05-05
[CALCITE-5688] Support TRUNCATE TABLE DDL statement in Babel parser

sumeetgajjar:
The PR adds support for TRUNCATE TABLE statement to the babel parser.

PR #3274, opened 2023-06-18
[CALCITE-5688] Support TRUNCATE TABLE DDL statement in server module

JiajunBernoulli:
Close #3188

### DIFF
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -1137,6 +1137,10 @@ SqlNode SqlStmt() :
         stmt = SqlDrop()
     |
 </#if>
+<#if (parser.truncateStatementParserMethods!default.parser.truncateStatementParserMethods)?size != 0>
+        stmt = SqlTruncate()
+    |
+</#if>
         stmt = OrderedQueryOrExpr(ExprContext.ACCEPT_QUERY)
     |
         stmt = SqlExplain()
@@ -4350,6 +4354,30 @@ SqlDrop SqlDrop() :
     )
     {
         return drop;
+    }
+}
+</#if>
+
+<#if (parser.truncateStatementParserMethods!default.parser.truncateStatementParserMethods)?size != 0>
+/**
+ * Parses a TRUNCATE statement.
+ */
+SqlTruncate SqlTruncate() :
+{
+    final Span s;
+    final SqlTruncate truncate;
+}
+{
+    <TRUNCATE> { s = span(); }
+    (
+<#-- additional literal parser methods are included here -->
+<#list (parser.truncateStatementParserMethods!default.parser.truncateStatementParserMethods) as method>
+        truncate = ${method}(s)
+        <#sep>|</#sep>
+</#list>
+    )
+    {
+        return truncate;
     }
 }
 </#if>
--- a/core/src/main/java/org/apache/calcite/sql/SqlTruncate.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlTruncate.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.sql;
+
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+/**
+ * Base class for an TRUNCATE statements parse tree nodes. The portion of the statement covered by
+ * this class is "TRUNCATE". Subclasses handle whatever comes afterwards.
+ */
+public abstract class SqlTruncate extends SqlDdl {
+
+  /**
+   * Creates a SqlTruncate.
+   */
+  protected SqlTruncate(SqlOperator operator, SqlParserPos pos) {
+    super(operator, pos);
+  }
+}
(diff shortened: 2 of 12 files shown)


babel/src/main/java/org/apache/calcite/sql/babel/SqlTruncateTable.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+
+/**
+ * Parse tree for {@code TRUNCATE TABLE} statement.
+ */
+public class SqlTruncateTable extends SqlDdl {
+
+  private static final SqlOperator OPERATOR =
```
ILuffZhe, line 37: What if we extend one SqlKind like `TRUNCATE_TABLE` for this?
I noticed that `OTHER_DDL` is for short term usage in the javadoc.

ILuffZhe (commented):
Looks good! Just one minor question.

babel/src/main/java/org/apache/calcite/sql/babel/SqlTruncateTable.java
JiajunBernoulli: Because I don't want to break the core module for now.

JiajunBernoulli: Let us discuss in JIRA: https://issues.apache.org/jira/browse/CALCITE-5688

core/src/main/java/org/apache/calcite/sql/SqlTruncate.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.apache.calcite.sql;
+
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+/**
+ * Base class for an TRUNCATE statements parse tree nodes. The portion of the statement covered by
+ * this class is "TRUNCATE". Subclasses handle whatever comes afterwards.
+ */
+public abstract class SqlTruncate extends SqlDdl {
```
ILuffZhe, line 25: I have one small question, except `TRUNCATE TABLE`, is there any other truncate syntax in the SQL standard?

core/src/main/java/org/apache/calcite/sql/ddl/SqlTruncateTable.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.calcite.sql.SqlTruncate;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+
+/**
+ * Parse tree for {@code TRUNCATE TABLE} statement.
```
ILuffZhe, line 33: Nitpick: maybe `Parses` is better based on the following java doc. It's your call.

server/src/test/resources/sql/table.iq
```diff
@@ -215,4 +215,8 @@ drop table if exists t;
 
 !update
 
+truncate table t;
```
ILuffZhe, line 218: Can we have a positive test here? To tell the effectiveness of TRUNCATE

core/src/main/java/org/apache/calcite/sql/SqlTruncate.java
JiajunBernoulli: As far as I know, there's nothing else.

The abstract class is reserved for the future.

server/src/test/resources/sql/table.iq
```diff
@@ -50,6 +50,21 @@ select * from t;
 
 !ok
 
+
+truncate table t;
```
JiajunBernoulli, line 54: `TRUNCATE` is DDL like `DROP`, so `0 rows modified`.

core/src/main/java/org/apache/calcite/sql/ddl/SqlTruncateTable.java
JiajunBernoulli: I like to maintain a unified style.😊
image

server/src/test/resources/sql/table.iq
JiajunBernoulli: I use more efficient methods to clear data and add one positive test.

server/src/main/java/org/apache/calcite/server/ServerDdlExecutor.java
```diff
@@ -182,6 +183,22 @@ static SqlNode renameColumns(@Nullable SqlNodeList columnList,
...
+    final CalciteSchema calciteSchema = pair.left;
+    final String tblName = pair.right;
+    final Table table = calciteSchema
+        .getTable(tblName, context.config().caseSensitive())
+        .getTable();
+    if (table instanceof MutableArrayTable) {
+      MutableArrayTable mutableArrayTable = (MutableArrayTable) table;
+      mutableArrayTable.rows.clear();
+    } else {
+      throw new UnsupportedOperationException("Only MutableArrayTable support truncate");
```
ILuffZhe, line 198: May I ask why only MutableArrayTable is supported? ^_^

ILuffZhe (commented):
I think the PR is in a good shape, just two minor comments.

server/src/test/resources/sql/table.iq
ILuffZhe: Good to know, I also tried in MySQL and confirmed your statement.

server/src/main/java/org/apache/calcite/server/ServerDdlExecutor.java
JiajunBernoulli: - The DELETE statement cannot delete data in the server module.
   I initially thought it was a [bug](https://issues.apache.org/jira/browse/CALCITE-5796). But after analysis, it  is limitation of the server module.
- `MutableArrayTable` is enough, because `CREATE` table all are `MutableArrayTable`. Here is code link: https://github.com/apache/calcite/blob/301f770a66ef3d26c9d4e8d274eae0f92a819e00/server/src/main/java/org/apache/calcite/server/ServerDdlExecutor.java#L515

ILuffZhe: OK, got it. Maybe some java doc is better for other developers, WDYT?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -4353,6 +4357,30 @@ SqlDrop SqlDrop() :
 }
 </#if>
 
+<#if (parser.truncateStatementParserMethods!default.parser.truncateStatementParserMethods)?size != 0>
+/**
+ * Parses a TRUNCATE TABLE statement.
```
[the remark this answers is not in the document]
JiajunBernoulli: Good catch.

server/src/main/java/org/apache/calcite/server/ServerDdlExecutor.java
JiajunBernoulli: Make sense, I added some explanations.

ILuffZhe:
+1

ILuffZhe (approved):
+1. LGTM!

JiajunBernoulli:
`rebase main` and `squash commits`, I will merge it tomorrow if no other suggestions.

PR #3276, opened 2023-06-20
[CALCITE-5793] Use NULLS FIRST/LAST syntax for BigQuery

wnob:
This may not be a complete solution to CALCITE-5775 but it fixes the issue for BigQuery by using the `NULLS FIRST` / `NULLS LAST` syntax introduced in 2020.

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -7058,40 +7058,6 @@ class RelToSqlConverterTest {
         .withBigQuery().ok(expected);
   }
 
-  /**
-   * Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-5775">[CALCITE-5775]</a>;
-   * make sure BigQuery uses {@code NULLS LAST} syntax.
-   */
-  @Test void testBigQueryUsesNullsLastInsteadOfEmulation() {
-    final Function<RelBuilder, RelNode> relFn = b ->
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7058,6 +7058,40 @@ private void checkLiteral2(String expression, String expected) {
         .withBigQuery().ok(expected);
   }
 
+  /**
+   * Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-5775">[CALCITE-5775]</a>;
+   * make sure BigQuery uses {@code NULLS LAST} syntax.
+   */
+  @Test void testBigQueryUsesNullsLastInsteadOfEmulation() {
+    final Function<RelBuilder, RelNode> relFn = b ->
```
julianhyde, line 7066: Could you extend an existing test case, e.g. `testSelectOrderByAscNullsLast` rather than adding a new one?

Or at least add it next to the existing related test cases. (Adding at the end causes conflicts.)

Could the input be SQL rather than RelNode?

Follow the conventional format for javadoc comments. 80 char lines, no semi after `</a>`, include the bug summary inside the `<a>` element.

wnob: I'm thinking now that the 2 existing tests that were affected by this change are enough. They test a `NULLS LAST` and a `DESC NULLS FIRST`. This new test will be for CALCITE-5775 as opposed to CALCITE-5793.

PR #3277, opened 2023-06-21
[CALCITE-5789] Query with two nested subqueries where the inner-most references the outer-most table returns wrong result

rubenada:
See details in CALCITE-5789

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
@@ -850,6 +850,7 @@ public class SubQueryRemoveRule
   private static void matchFilter(SubQueryRemoveRule rule,
       RelOptRuleCall call) {
     final Filter filter = call.rel(0);
+    final Set<CorrelationId> filterVariablesSet = filter.getVariablesSet();
     final RelBuilder builder = call.builder();
     builder.push(filter.getInput());
     int count = 0;
@@ -865,6 +866,12 @@ public class SubQueryRemoveRule
           LogicVisitor.find(RelOptUtil.Logic.TRUE, ImmutableList.of(c), e);
       final Set<CorrelationId>  variablesSet =
           RelOptUtil.getVariablesUsed(e.rel);
+      // Filter without variables could be handled before this change, we do not want
+      // to break it yet for compatibility reason.
+      if (!filterVariablesSet.isEmpty()) {
+        // Only consider the correlated variables which originated from this sub-query level.
+        variablesSet.retainAll(filterVariablesSet);
+      }
       final RexNode target =
           rule.apply(e, variablesSet, logic,
               builder, 1, builder.peek().getRowType().getFieldCount(), count);
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -3512,7 +3512,17 @@ public class SqlToRelConverter {
 
     // implement HAVING (we have already checked that it is non-trivial)
     relBuilder.push(bb.root());
-    relBuilder.filter(havingExpr);
+    // Set the correlation variables used in this sub-query to the filter node,
+    // same logic is being used for the filter generated in where clause.
+    Set<CorrelationId> variableSet = new HashSet<>();
+    RexSubQuery subQ = RexUtil.SubQueryFinder.find(havingExpr);
+    if (subQ != null) {
+      CorrelationUse p = getCorrelationUse(bb, subQ.rel);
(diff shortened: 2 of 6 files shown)


rubenada:
@libenchao sorry to disturb, you have recently reviewed other PRs concerning `SubQueryRemoveRule`, do you mind taking a look at this one?

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -865,6 +866,9 @@ private static void matchFilter(SubQueryRemoveRule rule,
           LogicVisitor.find(RelOptUtil.Logic.TRUE, ImmutableList.of(c), e);
       final Set<CorrelationId>  variablesSet =
           RelOptUtil.getVariablesUsed(e.rel);
+      if (!filterVariablesSet.isEmpty()) {
+        variablesSet.retainAll(filterVariablesSet);
```
libenchao, line 870: The change looks good to me, and this was also introduced in https://github.com/apache/calcite/pull/3193, (it seems that the author reverts this change in the latest version).   

But I found after this change, there is one test that produces incorrect result, see my latest comment(https://github.com/apache/calcite/pull/3193#issuecomment-1596028987). I'm not sure whether it's a regression, if not, I think we can safely merge this one.

rubenada: @libenchao I have the impression that `if (!filterVariablesSet.isEmpty()) { variablesSet.retainAll(filterVariablesSet); }` it's better than the current code, since fixes some issues (like the current one), but the proper fix would be just `variablesSet.retainAll(filterVariablesSet);` (without the `if` check).
When applying the fix without the `if` I got one regression in `misc.iq`, but this is something that already appeared in https://github.com/apache/calcite/pull/3193 and that PR contains the fix for it (see [this comment](https://github.com/apache/calcite/pull/3193#discussion_r1202473255))

I also found one new test case that is currently broken, and will only be fixed if we apply the patch without the ìf`: 
```
select deptno from dept d1 where exists (
 select 1 from dept d2 where d2.deptno = d1.deptno and exists (
  select 1 from dept d3 where d3.dname = d1.dname));
```
It is a bit "forced", because the query can be easily re-written to remove the inner-most subquery (which does not require to be 2-level nested); but it is a valid (although rare) scenario, which should ideally work (but since it is a bit far-fectched and it has a straightforward workaround by rewriting the query, I didn't put too much attention on it).

The thing is, and it seems you and I both reached the same conclusion (as you explain [here](https://github.com/apache/calcite/pull/3193#discussion_r1204993079)), if we apply the patch without the `if`, I fear we can lead to some issues on downstream projects, because the logic to initialize the variableSet of Filter is a bit fragile, and downstream projects using RelBuilder API can create a (wrong, incomplete) Filter that has a condition with RexSubQuery with a correlation variable, without setting this variable in the Filter's variableSet. Thus, when the `filterVariablesSet` is not empty, we know that is was correctly initialized, so we can exploit that info; but when it is empty, we can never know if it is correct (i.e. it is empty because it is supposed to be empty), or it is incorrect (it is supposed to be **not** empty, but the Filter was "wrongly" created without a variableSet). So, at that point, if we exploit that info (e.g. by applying the fix without the `if`), things can't stop working in the second case.
That's why I thought it was safer to include the "partial fix" (retain only if `filterVariablesSet` is not empty), even though it will not fix certain cases (like the SQL query that I mentioned earlier, or the "last test" that you mention on your branch, which actually comes from a different ticket [CALCITE-5716](https://issues.apache.org/jira/browse/CALCITE-5716) ).

I would propose (given than 1.35 RC phase is approaching) a conservative approach:
- I can close this PR & ticket, because it is superseded by https://github.com/apache/calcite/pull/3193 (we could perhaps add my new test in `sub-query.iq` from this PR into that PR, just for completeness).
- To avoid regressions on downstream projects, https://github.com/apache/calcite/pull/3193 would apply the fix with the `if`; that should fix [CALCITE-5683](https://issues.apache.org/jira/browse/CALCITE-5683) (and my ticket [CALCITE-5789](https://issues.apache.org/jira/browse/CALCITE-5789)); but it shall not contain the "last test in sub-query.iq` (which would be broken because AFAIK it requires the fix without the `if`).
- In any case, that test actually belongs to a different (related) issue: [CALCITE-5716](https://issues.apache.org/jira/browse/CALCITE-5716). That ticket was closed, but IMO it should be re-open; and we would need to re-evaluate whether or not fix it (and the risks it may provoke) by removing the `if` from the patch (in a future version).

wdyt? (sorry for the long message)

libenchao: Agreed, thanks for the detailed explanation, we're on the same page.

PR #3278, opened 2023-06-23
[CALCITE-5773] Gradle show tasks fails when creating javadocAggregateIncludingTests

zabetak:
1. Exclude "bom" project from the javadoc aggregation since it does not have "main" and "test" objects causing "tasks" to fail.

2. Include "tasks" when building with newest & oldest JDK in CI to ensure it remains functional.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -202,6 +202,7 @@ val javadocAggregateIncludingTests by tasks.registering(Javadoc::class) {
     description = "Generates aggregate javadoc for all the artifacts"
 
     val sourceSets = subprojects
+        .filter { it.name != "bom" }
         .mapNotNull { it.extensions.findByType<SourceSetContainer>() }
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -202,6 +202,7 @@ val javadocAggregateIncludingTests by tasks.registering(Javadoc::class) {
     description = "Generates aggregate javadoc for all the artifacts"
 
     val sourceSets = subprojects
+        .filter { it.name != "bom" }
```
libenchao, line 205: `javadocAggregate` section uses `.filter { it.names.contains("main") }`, we'd better to keep them consistent? (Both way looks good to me).

NobiGo (approved):
LGTM

build.gradle.kts
zabetak: I prefer to keep it as is. Thanks for taking a look!

PR #3281, opened 2023-06-28
Draft pr to simplify and fix ORDER BY ordinal issues

jhugomoore:
Exploring what would have to change in ReltoSqlConverterTest in order to simplify SqlImplementor->Context->orderField. 

Current findings
Tests that seem to be caused by bugs
- testConvertWindowToSql
Put in fix for egregious issue of (group by 10) in query 1. Ignore ordinals when in an aliasContext was fix. Don’t know enough about what it’s doing in the rest to be sure if they’re correct. 6 & 8 definitely seem wrong
- testMySqlUnparseListAggCall
listagg() order by shouldn't be ordinals, seems needlessly confusing
- PigRelOpTest -> testForEachNested
Confusing change, doesn't seem to match anything else.
- PigRelOpTest -> testRank()
Inserts Ordinals that don't exist 

Tests that insert a pretty crazy select clause, not sure if it’s what we want
- testOrderByNotInSelectList
- testRewriteOrderByWithNumericConstants
- testSelectQueryWithAscDescOrderByClause
- testSelectQueryWithLimitOffsetClause
- testSelectQueryWithTwoOrderByClause

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -610,17 +610,7 @@ public abstract class SqlImplementor {
      */
     public SqlNode orderField(int ordinal) {
       final SqlNode node = field(ordinal);
-      if (node instanceof SqlNumericLiteral
-          && dialect.getConformance().isSortByOrdinal()) {
-        // An integer literal will be wrongly interpreted as a field ordinal.
-        // Convert it to a character literal, which will have the same effect.
-        final String strValue = ((SqlNumericLiteral) node).toValue();
-        return SqlLiteral.createCharString(strValue, node.getParserPosition());
-      }
-      if (node instanceof SqlCall
-          && dialect.getConformance().isSortByOrdinal()) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -610,17 +610,10 @@ protected Context(SqlDialect dialect, int fieldCount, boolean ignoreCast) {
...
-        final String strValue = ((SqlNumericLiteral) node).toValue();
-        return SqlLiteral.createCharString(strValue, node.getParserPosition());
-      }
-      if (node instanceof SqlCall
-          && dialect.getConformance().isSortByOrdinal()) {
-        // If the field is expression and sort by ordinal is set in dialect,
-        // convert it to ordinal.
+      if (this instanceof SqlImplementor.AliasContext) {
+        return node;
+      }
```
wnob, line 615: Just override `AliasContext.orderField()` instead of checking `instanceof` in the parent class here.

jhugomoore: Good call, done

PR #3286, opened 2023-06-29
[CALCITE-5810] Prevent overflow in substring length computation

mihaibudiu:


mihaibudiu:
This is a very simple change, maybe it can make it in the next release.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -577,16 +577,18 @@ public static String substring(String c, int s) {
   /** SQL SUBSTRING(string FROM ... FOR ...) function. */
   public static String substring(String c, int s, int l) {
     int lc = c.length();
-    int e = s + l;
+    long e = (long) s + (long) l;
```
tanclary, line 580: Should we adjust the other `substring()` signatures in this class as well for consistency?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7347,6 +7347,8 @@ private static void checkSubstringFunction(SqlOperatorFixture f) {
         "ab", "VARCHAR(3) NOT NULL");
     f.checkString("substring(x'aabbcc' from 1 for 2)",
         "aabb", "VARBINARY(3) NOT NULL");
+    f.checkString("substring('abc' from 2 for 2147483646)",
```
tanclary, line 7350: What are the bounds of the `FROM` and `FOR` arguments? Might be worth adding a test for that case as well.

tanclary (commented):
Just left a couple comments, thank you for finding this bug and opening a fix!

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: This is a trick question, it may depend on the SQL dialect.
This implementation supports any `int` values for these arguments.
So you are suggesting adding negative tests that fail if the values are out of the `int` range?
I will look into it.

tanclary: That's a good point.

For example, BigQuery's arguments have `LONG` bounds, meaning a query like the following is valid:

`SELECT SUBSTR("hello", 2, 9223372036854775807);`

The scope of your bug might not necessarily include this but feel free to make additional adjustments to make the behavior more accurate, if you would like.

tanclary: That makes sense, I appreciate you looking into it. I think your fix for the original bug you filed looks good, I'll give it a little while to see if anyone else has comments but it LGTM.

mihaibudiu: Here's what Postgres says when you try it:
`Query Error: error: function substring(unknown, integer, bigint) does not exist`
This is a reasonable solution.

This is tied with the way Calcite does type inference and implicit cast insertion, which I can't claim I understand well.
An alternative approach would be to implicitly cast the bigint to an int, and if that requires truncation probably a warning should be issued.

Currently it looks to me like the constant folding fails but the error that causes the failure is lost when handling the exception, probably because of the use of Janino which entails dynamic code generation.

tanclary: There are already dialect-specific implementations for `SUBSTRING`, perhaps we could take advantage of these to specify the operand type to match the dialect behavior, and then have different signatures in `SqlFunctions` to handle all of the cases. I'll post a similar comment on the case you opened and we can keep the discussion going there. Thanks for looking into this!

NobiGo (approved):
LGTM

PR #3288, opened 2023-06-30
[CALCITE-5740] support aggregate semi-join rule

walterddr:
support a variance of the JOIN_TO_SEMI_JOIN rule that allows aggregate node on top of the targeted join node

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
@@ -67,13 +67,15 @@ public abstract class SemiJoinRule
     super(config);
   }
 
-  protected void perform(RelOptRuleCall call, @Nullable Project project,
+  protected void perform(RelOptRuleCall call, @Nullable RelNode topRel,
       Join join, RelNode left, Aggregate aggregate) {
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/apache/calcite/rel/rules/SemiJoinRule.java
```diff
@@ -67,13 +67,12 @@ protected SemiJoinRule(Config config) {
...
-  protected void perform(RelOptRuleCall call, @Nullable Project project,
+  protected void perform(RelOptRuleCall call, @Nullable RelNode topRel,
       Join join, RelNode left, Aggregate aggregate) {
     final RelOptCluster cluster = join.getCluster();
     final RexBuilder rexBuilder = cluster.getRexBuilder();
-    if (project != null) {
-      final ImmutableBitSet bits =
-          RelOptUtil.InputFinder.bits(project.getProjects(), null);
+    if (topRel != null) {
+      final ImmutableBitSet bits = findBits(topRel);
```
chunweilei, line 75: It should return directly If `findBits` returns empty BitSet.

walterddr:
@chunweilei could you share any additional follow up comments or if this is ready for merge? thanks

PR #3290, opened 2023-07-03
[CALCITE-5782] Add TO_HEX and FROM_HEX functions (enabled in BigQuery library)

zoudan:


site/_docs/reference.md
```diff
@@ -2752,6 +2752,8 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | FROM_BASE32(string)                            | Returns the decoded result of a base-32 *string* as a string
 | m | TO_BASE64(string)                              | Converts the *string* to base-64 encoded form and returns a encoded string
 | b m | FROM_BASE64(string)                          | Returns the decoded result of a base-64 *string* as a string
+| b | TO_HEX(binary)                                 | Converts *binary* into a hexadecimal varchar
+| b | FROM_HEX(varchar)                              | Converts a hexadecimal-encoded *varchar* into bytes
```
tanclary, line 2756: why `VARCHAR` and not `STRING`?

tanclary (commented):
Left a small question

site/_docs/reference.md
zoudan: Should we use types in the related libs or in calcite? I think it is the latter.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1247,6 +1247,20 @@ private static RelDataType deriveTypeMapFromArrays(SqlOperatorBinding opBinding)
           OperandTypes.STRING,
           SqlFunctionCategory.STRING);
 
+  @LibraryOperator(libraries = {BIG_QUERY})
```
ILuffZhe, line 1250: How about adding some java doc for those two functions? It helps others to understand better without browsing the JIRA description.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4174,6 +4174,37 @@ static void checkRlikeFails(SqlOperatorFixture f) {
...
+    final SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.TO_HEX);
+    f0.checkFails("^to_hex(x'')^",
+        "No match found for function signature TO_HEX\\(<BINARY>\\)",
+        false);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.BIG_QUERY);
+    f.checkString("to_hex(x'00010203AAEEEFFF')",
+        "00010203aaeeefff",
+        "VARCHAR NOT NULL");
+    f.checkString("to_hex(x'')", "", "VARCHAR NOT NULL");
+    f.checkNull("to_hex(cast (null as varbinary))");
```
ILuffZhe, line 4187: I noticed that there is a blank space after `cast`, shall we remove it? So as the following test.

ILuffZhe (commented):
Overall looks good! Left some minor comments.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
zoudan: @ILuffZhe Thanks for your suggestion, I have updated my branch. Please take a look when you have time.

site/_docs/reference.md
ILuffZhe: Agree with Dan, we should use Calcite's type names when describing the func.

ILuffZhe (approved):
+1.
Hi @tanclary , do you have any other comments? I think this PR is in a good shape now.

site/_docs/reference.md
tanclary: That makes sense to me, do you know why many of the functions use `string` in their documentation?

tanclary:
> +1. Hi @tanclary , do you have any other comments? I think this PR is in a good shape now.

I think it looks good as well, the only thing I am confused about is the documentation question but I ask more out of curiosity.

site/_docs/reference.md
zoudan: @tanclary IMO, people can easily confuse `string` and `varchar`, as we described in [CALCITE-5777](https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-5777?filter=myopenissues). 
STRING_TYPES not only contains CHAR_TYPES, but also BINARY_TYPES.
`public static final List<SqlTypeName> STRING_TYPES =
      combine(CHAR_TYPES, BINARY_TYPES);`

tanclary: That makes sense, thanks for explaining

ILuffZhe:
Hi @zoudan, could you please rebase and squash the commits? I think the PR is ready to merge.

PR #3291, opened 2023-07-03
[CALCITE-5816] Only return left-hand table columns when validate LEFT SEMI JOIN query

macroguo-ghy:


NobiGo (commented):
Almost looks good to me. Only small changes need to confirm.

NobiGo (approved):
LGTM

PR #3295, opened 2023-07-05
[CALCITE-5701] Add NAMED_STRUCT function (enabled in Spark library)

MasseGuillaume:


core/src/main/java/org/apache/calcite/interpreter/NamedStruct.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Set;
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+import org.apache.calcite.avatica.util.StructImpl;
+
+import java.sql.Struct;
+import java.sql.SQLException;
+
+public class NamedStruct extends StructImpl {
```
MasseGuillaume, line 29: It was necessary to subclass `StructImpl` in order to pass the Quidem tests

https://github.com/apache/calcite-avatica/blob/be8b99aeede2883e5692124edd187eaeffabed2c/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java#L1479

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableInterpretable.java
```diff
@@ -126,7 +126,8 @@ public static Bindable toBindable(Map<String, Object> parameters,
       }
     } catch (Exception e) {
       throw Helper.INSTANCE.wrap("Error while compiling generated Java code:\n"
-          + s, e);
```
NobiGo, line 129: This is unrelated changes about this ISSUE, If you think this message needs to improve, Please log a new issue in Jira.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -656,6 +658,7 @@ Builder populate2() {
           new PeriodNameImplementor("monthName",
               BuiltInMethod.MONTHNAME_WITH_TIMESTAMP,
               BuiltInMethod.MONTHNAME_WITH_DATE));
+
```
NobiGo, line 661: Unrelated changes, Please remove them.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3270,6 +3276,31 @@ private static class ValueConstructorImplementor
                       Expressions.box(value))));
         }
         return lyst;
+
+      case STRUCT_CONSTRUCTOR:
+
```
NobiGo, line 3281: In order to keep the code style unified, Please remove empty lines.

NobiGo (changes requested):
Some comments need to change.

MasseGuillaume:
For some reason, GitHub is not showing all the commit range: https://github.com/MasseGuillaume/calcite/commits/CALCITE-5701-2

ILuffZhe:
Hi @MasseGuillaume . I noticed that your PR contains someone else codes(maybe it's due to the force push), can you adjust your patch？ So we can proceed to the next step.

gabrywu:
any update here?

MasseGuillaume:
@gabrywu 

I'm stuck on the runtime implementation.

in JIRA: https://issues.apache.org/jira/browse/CALCITE-5701?focusedCommentId=17741811&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17741811
> You should not be using SqlFunctions.item. The javadoc says "Implements item on an object whose type is not known until runtime". But you know the type is STRUCT(k1 VARCHAR, K2 INTEGER). You know that "k1" is item zero of the Object[]. You should translate "item" call to a field access. And then none of your changes to ItemImplementor will be necessary.

I'm not too sure where that `translate "item" call to a field access` should happen. Maybe during the validation. (eg. `performUnconditionalRewrites`).

PR #3296, opened 2023-07-06
[CALCITE-5769] Optimizing 'CAST(e AS t) IS NOT NULL' to 'e IS NOT NULL'

NobiGo:


core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -2458,6 +2458,63 @@ private SqlOperator getNoDeterministicOperator() {
...
+  }
+
+  /** Unit test for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5769">[CALCITE-5769]
+   * Optimizing 'CAST(e AS t) IS NOT NULL' to 'e IS NOT NULL'</a>. */
+  @Test void testSimplifyCastIsNotNull2() {
+    // "(cast A as bigint) IS NOT NULL" when A is int and A is nullable
+    // ==>
+    // "A IS NOT NULL"
+    checkSimplify(isNotNull(cast(vInt(), tBigInt(true))), "IS NOT NULL(?0.int0)");
```
JiajunBernoulli, line 2510: How about `cast(varchar as binary) is not null`?

> More test between types is better.

NobiGo: @JiajunBernoulli I add this test. Now I use RexUtil.isLosslessCast to decide whether remove the cast in this expression. So now this implementation is secure and conservative(Because cast(varchar(8) as varchar(4)) actually can return value in SQL, But this Pr still keeps this cast. Only handle cast(varchar(4) as varchar(8))). About the type change test, we can find it in RexLosslessCastTest.

JiajunBernoulli:
Maybe we need to rerun CI?
image

core/src/test/resources/sql/misc.iq
```diff
@@ -2400,4 +2400,124 @@ TS
...
+| false  |
++--------+
+(14 rows)
+
+!ok
+EnumerableCalc(expr#0..7=[{inputs}], expr#8=[IS NULL($t3)], EXPR$0=[$t8])
+  EnumerableTableScan(table=[[scott, EMP]])
+!plan
+
+# Test CAST(e as t) IS NULL when CAST is not lossless And e is nullable
```
ILuffZhe, line 2496: We won't simplify the CAST that is not lossless, may I ask what's this test for?

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -2458,6 +2458,79 @@ private SqlOperator getNoDeterministicOperator() {
     checkSimplifyUnchanged(isNull(cast(vVarcharNotNull(), tInt(false))));
   }
 
+  /** Unit test for
```
ILuffZhe, line 2461: I noticed that there are many duplicated java doc for those tests, it would be much better if we propose a suitable way to organize them. But I'm OK with it if remains still.

ILuffZhe (changes requested):
Left some comments.

NobiGo:
> Maybe we need to rerun CI? image

Yes. This is because my commit info not following the format specifications(Capitalizes the first letter). So If the PR is good, I will rerun it.

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -2458,6 +2458,79 @@ private SqlOperator getNoDeterministicOperator() {
...
+    checkSimplifyUnchanged(isNotNull(cast(vVarcharNotNull(), tVarbinary(false))));
+  }
+
+  /** Unit test for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5769">[CALCITE-5769]
+   * Optimizing 'CAST(e AS t) IS NOT NULL' to 'e IS NOT NULL'</a>. */
+  @Test void testSimplifyCastIsNull4() {
+    // "(cast A as bigint) IS NULL" when A is int and A is nullable
+    // ==>
+    // "A IS NULL"
```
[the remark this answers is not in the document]
NobiGo: Because A is nullable(Possible include NULL value). For example:
When A is a 5, the expression value will be false. 
If A is NULL, the expression value will be true.

core/src/test/resources/sql/misc.iq
NobiGo: This tests have two reasons:
1. Compare with other test cases
2. Make sure the code change is right.

core/src/test/resources/sql/misc.iq
```diff
@@ -2400,4 +2400,124 @@ TS
...
+select cast(comm as integer) IS NOT NULL from "scott".emp;
```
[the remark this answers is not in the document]
NobiGo: Nice catch. I will fix it.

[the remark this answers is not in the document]
NobiGo: Done.

julianhyde:
Sorry about the CI failure. Yes it's because of the commit message. Unfortunately `LintTest` cannot distinguish between commits in a PR and commits on main.

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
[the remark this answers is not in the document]
ILuffZhe: Thanks for your explanation.

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -2458,6 +2458,79 @@ private SqlOperator getNoDeterministicOperator() {
...
   }
 
+  /** Unit test for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5769">[CALCITE-5769]
+   * Optimizing 'CAST(e AS t) IS NOT NULL' to 'e IS NOT NULL'</a>. */
+  @Test void testSimplifyCastIsNull3() {
+    // "(cast A as bigint) IS NULL" when A is int and A is not null
+    // ==>
+    // "false"
+    checkSimplify(isNull(cast(vIntNotNull(), tBigInt(false))), "false");
```
ILuffZhe, line 2468: I'm a little confused about this `vIntNotNull()`. Per its java doc, it returns a non-nullable int variable, however, in `vParamNotNull()`, whose java doc says this method returns nullable varchar variable.

NobiGo: Looks like the java doc has some problem. We can fix it in another issue.

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
JiajunBernoulli: @NobiGo  Thanks for your detailed explanation.

I will approve if the CI success.

ILuffZhe (approved):

JiajunBernoulli (approved):

PR #3299, opened 2023-07-08
[MINOR] Adjust nullableType and java doc for RexProgramBuilderBase

ILuffZhe:
Adjust some java docs and object references in RexProgramBuilderBase, which helps others to understand better.

### DIFF
--- a/core/src/test/java/org/apache/calcite/rex/RexProgramBuilderBase.java
+++ b/core/src/test/java/org/apache/calcite/rex/RexProgramBuilderBase.java
@@ -147,6 +147,8 @@ public abstract class RexProgramBuilderBase {
     });
   }
 
+  // Operators
+
   protected RexNode isNull(RexNode node) {
     return rexBuilder.makeCall(SqlStdOperatorTable.IS_NULL, node);
   }
@@ -341,6 +343,7 @@ public abstract class RexProgramBuilderBase {
   }
 
   // Types
+
   protected RelDataType nullable(RelDataType type) {
     if (type.isNullable()) {
       return type;
@@ -367,7 +370,6 @@ public abstract class RexProgramBuilderBase {
     }
     return sqlType;
   }
-
 
   protected RelDataType tChar(int precision) {
     return tChar(false, precision);
@@ -437,6 +439,7 @@ public abstract class RexProgramBuilderBase {
   protected RelDataType tArray(RelDataType elemType) {
     return typeFactory.createArrayType(elemType, -1);
   }
+
   // Literals
 
   /**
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/apache/calcite/rex/RexProgramBuilderBase.java
```diff
@@ -756,8 +759,8 @@ protected RexNode vParam(String name, RelDataType type) {
    */
   protected RexNode vParam(String name, int arg, RelDataType type) {
     assertArgValue(arg);
-    RelDataType nonNullableType = typeFactory.createTypeWithNullability(type, false);
-    return rexBuilder.makeFieldAccess(getDynamicParam(nonNullableType, name), arg);
+    RelDataType nullableType = typeFactory.createTypeWithNullability(type, true);
```
JiajunBernoulli, line 762: The logic of the function has also been changed.

So I think the commit message is not accurate enough.

JiajunBernoulli (commented):
`[Minor]` should be `[MINOR]`.

Because `[CALCITE-xxx]`, Not `[Calcite]`.

core/src/test/java/org/apache/calcite/rex/RexProgramBuilderBase.java
JiajunBernoulli: How about `Adjust nullableType  and java doc for RexProgramBuilderBase`?

ILuffZhe: Good point, I'll update the commit info.

JiajunBernoulli (approved):

core/src/test/java/org/apache/calcite/rex/RexProgramBuilderBase.java
NobiGo, line 762: This line code is necessary? Because we have changed the code logic, The Test still passes, So this is an invalid code？

NobiGo (changes requested):

core/src/test/java/org/apache/calcite/rex/RexProgramBuilderBase.java
ILuffZhe: I checked `getDynamicParam()` in `vParam`, which creates nullableType internally. So we don't need to change these codes.

NobiGo (approved):
LGTM

NobiGo (approved):
LGTM

ILuffZhe:
@JiajunBernoulli @NobiGo  Thanks for the review!

PR #3300, opened 2023-07-09
[CALCITE-5820] Add PARSE_URL function (enabled in Hive and Spark library)

herunkang2018:


tanclary (commented):
Left some comments

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -875,6 +878,70 @@ public static String translateWithCharset(String s, String transcodingName) {
...
+    if (partToExtract.equals("USERINFO")) {
+      return uri.getRawUserInfo();
+    }
+
+    return null;
+  }
+
+  public static @Nullable String parseUrl(@Nullable String urlStr,
+      @Nullable String partToExtract, @Nullable String keyToExtract) {
+    if (!partToExtract.equals("QUERY")) {
```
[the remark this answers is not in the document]
herunkang2018: I will fix it soon.

site/_docs/reference.md
```diff
@@ -2769,6 +2769,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | PARSE_DATETIME(format, string)                 | Uses format specified by *format* to convert *string* representation of datetime to a TIMESTAMP value
 | b | PARSE_TIME(format, string)                     | Uses format specified by *format* to convert *string* representation of time to a TIME value
 | b | PARSE_TIMESTAMP(format, string[, timeZone])    | Uses format specified by *format* to convert *string* representation of timestamp to a TIMESTAMP WITH LOCAL TIME ZONE value in *timeZone*
+| h s | PARSE_URL(urlString, partToExtract [, keyToExtract]) | Returns the specified part from the URL. Valid values for partToExtract include HOST, PATH, QUERY, REF, PROTOCOL, AUTHORITY, FILE, and USERINFO
```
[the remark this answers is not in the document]
herunkang2018: Thanks for reviewing, I will fix it soon.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -394,6 +394,14 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
   public static final SqlFunction SUBSTR_POSTGRESQL =
       SUBSTR.withKind(SqlKind.SUBSTR_POSTGRESQL);
 
+  /** The "PARSE_URL(urlString, partToExtract [, keyToExtract])" function. */
```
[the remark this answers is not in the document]
herunkang2018: I found that both with and without spaces were used in the previous code, so I want to know which one is recommended and would you like to explain the reason?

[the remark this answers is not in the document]
tanclary: Yeah that's true there are inconsistencies. I think it makes sense to have spaces on either both sides or neither. Since nearly all examples have a space (`foo [, bar`) at the beginning it makes sense to match. But you are right it is not consistent across all documentation.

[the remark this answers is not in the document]
herunkang2018: Fixed.

site/_docs/reference.md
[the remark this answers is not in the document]
herunkang2018: Fixed.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
herunkang2018: Fixed.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -875,6 +878,70 @@ public static String translateWithCharset(String s, String transcodingName) {
...
+    }
+
+    URI uri;
+    try {
+      uri = new URI(urlStr);
+    } catch (URISyntaxException e) {
+      return null;
+    }
+
+    if (partToExtract.equals("HOST")) {
```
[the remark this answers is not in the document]
herunkang2018: I have refactor it. Thanks for your advice.

herunkang2018:
@tanclary all comments resolved, please help to review if you have time.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -875,6 +878,82 @@ public static String translateWithCharset(String s, String transcodingName) {
...
+      }
+      break;
+    case AUTHORITY:
+      extractValue = uri.getRawAuthority();
+      break;
+    case USERINFO:
+      extractValue = uri.getRawUserInfo();
+      break;
+    default:
+      return null;
```
tanclary, line 933: Is this the right behavior? Or should it throw an error?

site/_docs/reference.md
```diff
@@ -2769,6 +2769,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | PARSE_DATETIME(format, string)                 | Uses format specified by *format* to convert *string* representation of datetime to a TIMESTAMP value
 | b | PARSE_TIME(format, string)                     | Uses format specified by *format* to convert *string* representation of time to a TIME value
 | b | PARSE_TIMESTAMP(format, string[, timeZone])    | Uses format specified by *format* to convert *string* representation of timestamp to a TIMESTAMP WITH LOCAL TIME ZONE value in *timeZone*
+| h s | PARSE_URL(urlString, partToExtract [, keyToExtract] ) | Returns the specified part from the *urlString*. Valid values for *partToExtract* include HOST, PATH, QUERY, REF, PROTOCOL, AUTHORITY, FILE, and USERINFO
```
tanclary, line 2772: I would make mention of `keyToExtract` here

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: Oh I see you have a test for so I guess it is right

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -875,6 +878,82 @@ public static String translateWithCharset(String s, String transcodingName) {
...
+
+    String query = parseUrl(urlStr, partToExtract);
+    if (query == null) {
+      return null;
+    }
+
+    Pattern p = Pattern.compile("(&|^)" + keyToExtract + "=([^&]*)");
+    Matcher m = p.matcher(query);
+    if (m.find()) {
+      return m.group(2);
```
tanclary, line 952: can this be simplified: `return m.find() ? m.group(2) : null;`?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -3946,6 +3946,87 @@ static void checkRlikeFails(SqlOperatorFixture f) {
...
+              + " 'AUTHORITY')",
+          "calcite.apache.org",
+          "VARCHAR NOT NULL");
+
+      // test with invalid partToExtract
+      f.checkNull("parse_url('http://calcite.apache.org/path1/p.php?k1=v1&k2=v2#Ref1',"
+              + " 'INVALID_PART_TO_EXTRACT')");
+
+      // test with operands with null values
+      f.checkNull("parse_url('http://calcite.apache.org/path1/p.php?k1=v1&k2=v2#Ref1',"
```
NobiGo, line 4019: Please add a test when we have an invalid URL.

site/_docs/reference.md
NobiGo: I agree with this too.

NobiGo: I agree with this too.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -875,6 +878,82 @@ public static String translateWithCharset(String s, String transcodingName) {
     }
   }
 
+  /** SQL {@code PARSE_URL(urlStr, partToExtract [, keyToExtract] )} function. */
```
NobiGo, line 881: Looks like the Java Doc is more suitable for placing on the method below.

NobiGo, line 881: Looks like the Java Doc is more suitable for placing on the method below.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -3946,6 +3946,87 @@ static void checkRlikeFails(SqlOperatorFixture f) {
     f.checkNull("ASCII(cast(null as varchar(1)))");
   }
 
+  @Test void testParseUrl() {
```
NobiGo, line 3949: I will put this method in parseTime after, making it more reasonable. But it is Ok to put it here.

NobiGo (commented):
Left some comments that need to confirm.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: Thanks for remind, I think this can be improved to `break;`, I will update it soon.
Note that since we have handled `IllegalArgumentException` before, here we don't need to handle mismatched enum value again, and can safely use `break;`.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
herunkang2018: Thanks for reviewing. I have considerd this option, but these tests are related to date/time/timestmap processing, not string processing, so I place it here.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: I have adjusted the places of these two method, and the doc.

herunkang2018:
@tanclary @NobiGo Thanks a lot for reviewing.
All comments resolved, please help to review if you have time.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -875,6 +878,80 @@ public static String translateWithCharset(String s, String transcodingName) {
...
+      break;
+    case AUTHORITY:
+      extractValue = uri.getRawAuthority();
+      break;
+    case USERINFO:
+      extractValue = uri.getRawUserInfo();
+      break;
+    default:
+      extractValue = null;
+      break;
```
tanclary, line 950: I don't think you need a `break` in the default case.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4565,4 +4642,16 @@ public enum FlatProductInputType {
     SCALAR, LIST, MAP
   }
 
+  /** Type of part to extract passed into {@link #parseUrl}. */
+  public enum PartToExtract {
+    HOST,
```
tanclary, line 4647: Is there a reason for this specific ordering? If not maybe we could alphabetize this and the switch statement? This is just a nit so if you feel strongly one way or the other it's probably fine.

site/_docs/reference.md
```diff
@@ -2769,6 +2769,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | PARSE_DATETIME(format, string)                 | Uses format specified by *format* to convert *string* representation of datetime to a TIMESTAMP value
 | b | PARSE_TIME(format, string)                     | Uses format specified by *format* to convert *string* representation of time to a TIME value
 | b | PARSE_TIMESTAMP(format, string[, timeZone])    | Uses format specified by *format* to convert *string* representation of timestamp to a TIMESTAMP WITH LOCAL TIME ZONE value in *timeZone*
+| h s | PARSE_URL(urlString, partToExtract [, keyToExtract] ) | Returns the specified part from the *urlString*. Valid values for *partToExtract* include HOST, PATH, QUERY, REF, PROTOCOL, AUTHORITY, FILE, and USERINFO. *keyToExtract* specifies which query to extract
```
tanclary, line 2772: Does it make more sense to say `Returns specified \*partToExtract\* from the \*urlString\*`?

tanclary: Backslashes are just to avoid formatting in github

tanclary (commented):
Thanks for fixing all the comments! I left a couple more small ones but I think this is nearly mergeable, what do you think @NobiGo ?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: IMO the enum type doesn't need to keep alphabetize order, like `SqlKind`.
In this case, this is related to the order of the extracted parts of URL, like HOST/PATH/QUERY/REF. Although the latter part is not strictly in accordance with the order of the URL, it may be used by users less, so it is placed later.
And this way is keep the same as Hive's and Spark's implementation, so I think it's probably fine.

site/_docs/reference.md
herunkang2018: Thanks for review. I will improve it soon.

herunkang2018:
@tanclary All comments resolved, please help to review if you have time.

NobiGo:
> Thanks for fixing all the comments! I left a couple more small ones but I think this is nearly mergeable, what do you think @NobiGo ?

LGTM. We can continue.

herunkang2018:
@tanclary @NobiGo would you like to help to review again if you have time? Thanks very much.

tanclary (approved):
LGTM, Calcite main is frozen right now for the new release but once that is resolved I think this is ready to merge.

herunkang2018:
@tanclary would like to have a look if you have time, since the 1.35.0 is released and the main branch may open again.

herunkang2018:
@tanclary I have squashed to the final commit. Thanks for the reviewing work very kindly.

PR #3302, opened 2023-07-09
[CALCITE-5821] Add FORMAT_NUMBER function (enabled in Hive and Spark library)

herunkang2018:


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -618,6 +618,102 @@ public static ByteString substring(ByteString c, int s, int l) {
...
+    if (decimalVal > 0) {
+      pattern.append(".");
+      for (int i = 0; i < decimalVal; i++) {
+        pattern.append("0");
+      }
+    }
+    return pattern.toString();
+  }
+
+  public static DecimalFormat getNumberFormat(String pattern) {
```
tanclary, line 705: If this is just a helper you can make it private

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -618,6 +618,102 @@ public static ByteString substring(ByteString c, int s, int l) {
...
+    return NumberUtil.decimalFormat(pattern);
+  }
+
+  public static DecimalFormat getNumberFormat(int decimalVal) {
```
tanclary, line 709: Same comment as above

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
```diff
@@ -417,6 +417,7 @@ public enum BuiltInMethod {
   INITCAP(SqlFunctions.class, "initcap", String.class),
   SUBSTRING(SqlFunctions.class, "substring", String.class, int.class,
       int.class),
+  FORMAT_NUMBER(SqlFunctions.class, "formatNumber", int.class, int.class),
```
tanclary, line 420: How do the functions with `(int, string)` signature get called when you specify `(int.class, int.class)` here? Not sure if it causes problems, just curious.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -618,6 +618,102 @@ public static ByteString substring(ByteString c, int s, int l) {
...
+
+  public static String formatNumber(BigDecimal value, String format) {
+    DecimalFormat numberFormat = getNumberFormat(format);
+    return numberFormat.format(value);
+  }
+
+  public static String getFormatPattern(int decimalVal) {
+    StringBuilder pattern = new StringBuilder();
+    pattern.append("#,###,###,###,###,###,##0");
+
```
tanclary, line 695: I'm kind of surprised there's not a method somewhere that does this already.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -618,6 +618,102 @@ public static ByteString substring(ByteString c, int s, int l) {
...
+    DecimalFormat numberFormat = getNumberFormat(decimalVal);
+    return numberFormat.format(value);
+  }
+
+  public static String formatNumber(int value, int decimalVal) {
+    DecimalFormat numberFormat = getNumberFormat(decimalVal);
+    return numberFormat.format(value);
+  }
+
+  public static String formatNumber(long value, int decimalVal) {
```
tanclary, line 637: do you need a method for all of these types? There's other functions that use implicit type conversion (see `COSH` or one of the trig functions for a good example). If they're necessary that's fine, just make sure they're all tested.

tanclary (commented):
Thanks for the PR, left a couple small comments

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: Yes, I'm also curious about that. But currently there's not a method here to reuse.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: Thanks for your advice, I agree with you because it does not affect the correctness of the result, and it can make code more concise.

herunkang2018:
@tanclary Thanks for your review very kindly. And all comments resolved.

site/_docs/reference.md
```diff
@@ -2724,6 +2724,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | FLOOR(value)                                   | Similar to standard `FLOOR(value)` except if *value* is an integer type, the return type is a double
 | b | FORMAT_DATE(string, date)                      | Formats *date* according to the specified format *string*
 | b | FORMAT_DATETIME(string, timestamp)             | Formats *timestamp* according to the specified format *string*
+| h s | FORMAT_NUMBER(value, decimalOrFormat)        | Formats the number *value* like '#,###,###.##', rounded to decimal places. If *decimalOrFormat* is 0, the result has no decimal point or fractional part. The *decimalOrFormat* also accept a user specified format. This is supposed to function like MySQL's FORMAT
```
tanclary, line 2727: I am wondering if it makes sense to split this into two entries, one for `FORMAT_NUMBER(value, decimal)` and another for `FORMAT_NUMBER(value, format)`. I think it would make writing the descriptions easier.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7714,6 +7714,277 @@ void assertSubFunReturns(boolean binary, String s, int start,
...
+          "VARCHAR NOT NULL");
+      f.checkString("format_number(cast(1 as tinyint), 4)", "1.0000",
+          "VARCHAR NOT NULL");
+      f.checkString("format_number(cast(-1 as tinyint), 4)", "-1.0000",
+          "VARCHAR NOT NULL");
+      f.checkString("format_number(cast(12 as tinyint), 2)", "12.00",
+          "VARCHAR NOT NULL");
+      f.checkString("format_number(cast(123 as tinyint), 5)", "123.00000",
+          "VARCHAR NOT NULL");
+      f.checkString("format_number(cast(1 as tinyint), '#,###,###,###,###,###,##0.0000')",
```
tanclary, line 7734: I think we could make the testing here a bit more concise, if you check out the `testAbsFunc` test or some others, I don't think adding the same few tests for each type is adding much value, maybe just one per type or something just to verify they're all compatible. What do you think?

tanclary (commented):
left a couple comments, thank you!

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
herunkang2018: You are right. I will reduce the test cases for each types. And a reasonable way is to split the test cases into there general types: bigint/double/decimal, and add more test cases to cover. For other types, like tinyint/smallint/int, just add one test case is enough.

site/_docs/reference.md
herunkang2018: I agree with you. It will be more readable for user.

herunkang2018:
@tanclary All comments resolved, please help to review if you have time.

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
herunkang2018: The declared parameters type is only used to check whether there is a matched implemented method in the declared class, the related code line is [link](https://github.com/apache/calcite/blob/b2cdab28ad154f20dfb5ce19ec7fd4d772d6ea2c/core/src/main/java/org/apache/calcite/util/BuiltInMethod.java#L766). It doesn't check all implemented methods, it just validates only one, and if not found, it throws `NoSuchMethodException`. So this is a weak check here.

IMO, there are two phases to validate function's parameters type and arguments type:

In validate phase, the function's parameters type will be validated.

In execution phase, the generated code for expression will be compiled to class, if there is no matched method, including not matched parameters type, it will throws compilation error. Notice the code generation phase doesn't check the matchness of actual arguments type and function's parameters type.

Please point out if there is something wrong.

tanclary (commented):
I think this looks good, thanks for responding to all of the comments. I think you might have a merge conflict but if you could resolve that and squash your commits, I will merge later today assuming no one else has comments. Thanks!

herunkang2018:
@tanclary Thanks very kindly for you reviewing. I have rebased and squashed to one final commit.

herunkang2018:
@tanclary would you like to help review again? Or this is ready to be merged if it looks good.

tanclary:
Yep, looks good. Will merge once main branch is open again. Thanks!

herunkang2018:
@tanclary Since the 1.35.0 is released and the main branch may open again, would like to have a look if you have time?

tanclary:
I think @tjbanghart was tagged on the JIRA case so I will ask if he has any comments, otherwise this is good. Sorry for the delay in review I have been out the past few days.

herunkang2018:
@tjbanghart would you like to review this PR if you have time?

herunkang2018:
@tanclary Since he has not responded for 48h, I think we can continue. Please let me know if you have other consideration.

herunkang2018:
@tanclary Thanks a lot for reviewing this PR. Since @tjbanghart have not responded to review request for one month, could we continue on this PR?

tjbanghart (approved):
I thought we might be able to reuse some of the format utils added for CALCITE-5357 but on second take, I don't think that is applicable here. 

LGTM, thank you!

PR #3303, opened 2023-07-10
[CALCITE-5827] Add IS_INF and IS_NAN functions (enabled in BigQuery l…

tanclary:
…ibrary)

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -2096,6 +2096,36 @@ public class SqlFunctions {
     return Math.toDegrees(b0);
   }
 
+  /** SQL <code>IS_INF</code> operator applied to BigDecimal values. */
+  public static boolean isInf(BigDecimal b0) {
+    return Double.isInfinite(b0.doubleValue());
+  }
+
+  /** SQL <code>IS_INF</code> operator applied to double values. */
+  public static boolean isInf(double b0) {
+    return Double.isInfinite(b0);
+  }
+
+  /** SQL <code>IS_INF</code> operator applied to float values. */
+  public static boolean isInf(float b0) {
+    return Float.isInfinite(b0);
+  }
+
+  /** SQL <code>IS_NAN</code> operator applied to BigDecimal values. */
+  public static boolean isNaN(BigDecimal b0) {
+    return Double.isNaN(b0.doubleValue());
+  }
+
+  /** SQL <code>IS_NAN</code> operator applied to double values. */
+  public static boolean isNaN(double b0) {
+    return Double.isNaN(b0);
+  }
+
+  /** SQL <code>IS_NAN</code> operator applied to float values. */
+  public static boolean isNaN(float b0) {
+    return Float.isNaN(b0);
+  }
+
   // RADIANS
   /** SQL <code>RADIANS</code> operator applied to BigDecimal values. */
   public static double radians(BigDecimal b0) {
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2729,10 +2729,12 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b o | GREATEST(expr [, expr ]*)                    | Returns the greatest of the expressions
 | b h s | IF(condition, value1, value2)              | Returns *value1* if *condition* is TRUE, *value2* otherwise
 | b | IFNULL(value1, value2)                         | Equivalent to `NVL(value1, value2)`
-| b o | INSTR(string, substring [, from [, occurrence ] ]) | Returns the position of *substring* in *string*, searching starting at *from* (default 1), and until locating the nth *occurrence* (default 1) of *substring*
-| m | INSTR(string, substring)                       | Equivalent to `POSITION(substring IN string)`
 | p | string1 ILIKE string2 [ ESCAPE string3 ]       | Whether *string1* matches pattern *string2*, ignoring case (similar to `LIKE`)
 | p | string1 NOT ILIKE string2 [ ESCAPE string3 ]   | Whether *string1* does not match pattern *string2*, ignoring case (similar to `NOT LIKE`)
+| b o | INSTR(string, substring [, from [, occurrence ] ]) | Returns the position of *substring* in *string*, searching starting at *from* (default 1), and until locating the nth *occurrence* (default 1) of *substring*
+| m | INSTR(string, substring)                       | Equivalent to `POSITION(substring IN string)`
+| b | IS_INF(value)                                  | Returns whether *value* is infinite
+| b | IS_NAN(value)                                  | Returns whether *value* is NaN
 | m | JSON_TYPE(jsonValue)                           | Returns a string value indicating the type of *jsonValue*
 | m | JSON_DEPTH(jsonValue)                          | Returns an integer value indicating the depth of *jsonValue*
 | m | JSON_PRETTY(jsonValue)                         | Returns a pretty-printing of *jsonValue*
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -6618,6 +6618,64 @@ public class SqlOperatorTest {
     f.checkNull("pow(cast(null as integer), 2)");
   }
 
+  @Test void testInfinity() {
+    final SqlOperatorFixture f = fixture();
+    f.checkScalar("cast('Infinity' as double)", "Infinity",
+        "DOUBLE NOT NULL");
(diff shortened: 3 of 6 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6618,6 +6618,34 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f.checkBoolean("is_inf(1.2345)", false);
+    f.checkBoolean("is_inf(cast('NaN' as double))", false);
+    f.checkBoolean("is_inf(cast('Infinity' as double))", true);
+    f.checkBoolean("is_inf(cast('-Infinity' as double))", true);
+    f.checkNull("is_inf(cast(null as double))");
+  }
+
+  @Test void testIsNaNFunc() {
+    final SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.IS_NAN);
+    f0.checkFails("^is_nan(3)^",
```
olivrlee, line 6637: Is this test to test that it's only enabled for BQ?

tanclary: Yep!

site/_docs/reference.md
```diff
@@ -2731,6 +2731,8 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | IFNULL(value1, value2)                         | Equivalent to `NVL(value1, value2)`
 | b o | INSTR(string, substring [, from [, occurrence ] ]) | Returns the position of *substring* in *string*, searching starting at *from* (default 1), and until locating the nth *occurrence* (default 1) of *substring*
 | m | INSTR(string, substring)                       | Equivalent to `POSITION(substring IN string)`
+| b | IS_INF(value)                                  | Returns whether *value* is infinite
+| b | IS_NAN(value)                                  | Returns whether *value* is NaN
```
julianhyde, line 2735: alphabetically, should be after `ILIKE`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6618,6 +6618,34 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    final SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.IS_INF);
+    f0.checkFails("^is_inf(3)^",
+        "No match found for function signature IS_INF\\(<NUMERIC>\\)",
+        false);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.BIG_QUERY);
+    f.checkBoolean("is_inf(3)", false);
+    f.checkBoolean("is_inf(1.2345)", false);
+    f.checkBoolean("is_inf(cast('NaN' as double))", false);
+    f.checkBoolean("is_inf(cast('Infinity' as double))", true);
+    f.checkBoolean("is_inf(cast('-Infinity' as double))", true);
```
julianhyde, line 6631: did we ever test `CAST('Infinity' as DOUBLE)`?

is this standard SQL behavior?

julianhyde: also check on REAL values (implicit cast of 32-bit float to 64-bit float)

site/_docs/reference.md
tanclary: Done

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: Added tests for REAL and Infinity/NaN

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2106,6 +2106,11 @@ public static boolean isInf(double b0) {
     return Double.isInfinite(b0);
   }
 
+  /** SQL <code>IS_INF</code> operator applied to float values. */
```
julianhyde, line 2109: do the tests pass if you remove the `isInf(float)` and isNaN(float)` methods? I suspect they will. If so, remove the methods. less is more.

tanclary: Done.

tanclary:
@julianhyde Responded to your comments, thanks for the review.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: This was actually not right. The tests fail because if you remove the `float` methods, they use the BigDecimal functions which do not recognize `Infinity` and `NaN`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6618,6 +6618,60 @@ private static void checkIf(SqlOperatorFixture f) {
     f.checkNull("pow(cast(null as integer), 2)");
   }
 
+  @Test void testInfinity() {
+    final SqlOperatorFixture f = fixture();
+    f.checkScalar("CAST('Infinity' AS DOUBLE)", "Infinity",
```
herunkang2018, line 6623: Would it be better to align the case of keywords, such as keep lower case as other test cases?

tanclary: Done, let me know if you have other suggestions!

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: Could you add a test case for float value? IMO this may go to the function applied to double value, if we remove the function applied to float value. I am not trying it, so please let me know if I'm wrong.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6618,6 +6618,60 @@ private static void checkIf(SqlOperatorFixture f) {
...
+  }
+
+  @Test void testIsNaNFunc() {
+    final SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.IS_NAN);
+    f0.checkFails("^is_nan(3)^",
+        "No match found for function signature IS_NAN\\(<NUMERIC>\\)",
+        false);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.BIG_QUERY);
+    f.checkBoolean("is_nan(3)", false);
+    f.checkBoolean("is_nan(1.2345)", false);
```
herunkang2018, line 6665: Could you add a test case when decimal is Infinity or NaN?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: Done

tanclary:
@herunkang2018 @julianhyde I responded to your comments, let me know if you have any other suggestions.

herunkang2018:
@tanclary I will check it soon.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6618,6 +6618,64 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f.checkBoolean("is_nan(cast('Infinity' as double))", false);
+    f.checkBoolean("is_nan(cast('Infinity' as float))", false);
+    f.checkBoolean("is_nan(cast('Infinity' as real))", false);
+    f.checkBoolean("is_nan(cast('-Infinity' as double))", false);
+    f.checkBoolean("is_inf(cast('-Infinity' as float))", false);
```
herunkang2018, line 6672: Is this a typo(is_inf -> is_nan)?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
herunkang2018: When cast Infinity or NaN to decimal, I find it throws `NumberFormatException: Infinite or NaN`, because BigDecimal does not support Infinity or NaN.
So currently we are not able to add a test case for these special decimal value.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: Yes thanks!

tanclary:
@herunkang2018 Could you elaborate on your comment about the decimal testing? is that blocking this PR?

herunkang2018:
@tanclary I checked that decimal type does not support NaN and Infinite, so the comment will be resolved and we can continue.

herunkang2018:
@tanclary LGTM

herunkang2018 (approved):

PR #3307, opened 2023-07-10
[CALCITE-5831] Add SOUNDEX function (enabled in Spark library)

chucheng92:

Returns Soundex code of the string.

> SELECT soundex('Miller');
M460
 
details pls see: https://issues.apache.org/jira/browse/CALCITE-5831
note: spark soundex not throw exception but return original string if the input string can't be mapped.

# Brief change log
Add soundex function (enabled in Spark Library)

# Verifying this change

SqlOperatorTests#testSoundex2Func()

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -228,6 +228,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.SHA512;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SINH;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SORT_ARRAY;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SOUNDEX;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.SOUNDEX_SPARK;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SPACE;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SPLIT;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.STARTS_WITH;
@@ -540,6 +541,7 @@ public class RexImpTable {
       defineMethod(SPACE, BuiltInMethod.SPACE.method, NullPolicy.STRICT);
       defineMethod(STRCMP, BuiltInMethod.STRCMP.method, NullPolicy.STRICT);
       defineMethod(SOUNDEX, BuiltInMethod.SOUNDEX.method, NullPolicy.STRICT);
+      defineMethod(SOUNDEX_SPARK, BuiltInMethod.SOUNDEX_SPARK.method, NullPolicy.STRICT);
       defineMethod(DIFFERENCE, BuiltInMethod.DIFFERENCE.method, NullPolicy.STRICT);
       defineMethod(REVERSE, BuiltInMethod.REVERSE.method, NullPolicy.STRICT);
       defineMethod(SPLIT, "split", NullPolicy.STRICT);
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -717,6 +717,15 @@ public class SqlFunctions {
     return SOUNDEX.soundex(s);
   }
 
+  /** SQL SOUNDEX(string) function but return original s when not mapped. */
+  public static String soundexSpark(String s) {
+    try {
+      return SOUNDEX.soundex(s);
+    } catch (IllegalArgumentException ignore) {
+      return s;
+    }
+  }
+
   /** SQL DIFFERENCE(string, string) function. */
   public static int difference(String s0, String s1) {
     String result0 = soundex(s0);
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -821,6 +821,12 @@ public abstract class SqlLibraryOperators {
           OperandTypes.CHARACTER,
           SqlFunctionCategory.STRING);
 
+  /** The variant of the SOUNDEX operator. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction SOUNDEX_SPARK =
+      ((SqlBasicFunction) SOUNDEX).withKind(SqlKind.SOUNDEX_SPARK)
+          .withReturnTypeInference(ReturnTypes.VARCHAR_NULLABLE);
+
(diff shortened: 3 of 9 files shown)


core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -227,6 +227,7 @@
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SINH;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SORT_ARRAY;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.SOUNDEX;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.SOUNDEX2;
```
tanclary, line 230: I feel like SOUNDEX2 implies it takes 2 arguments, maybe SOUNDEX_SPARK would be a better name? Similar to the SUBSTR or FLOOR/CEIL functions.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -821,6 +821,13 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
           OperandTypes.CHARACTER,
           SqlFunctionCategory.STRING);
 
+  @LibraryOperator(libraries = {SPARK})
```
tanclary, line 824: could you just do `= SOUNDEX.withName('SOUNDEX_SPARK')` (or whatever name you choose)

tanclary (commented):
Looks pretty good I think. Is the behavior for the other dialects correct? I would imagine so but on the off chance it's not, we could avoid adding a whole extra operator. Otherwise just left a couple small comments.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
chucheng92: Maybe the behavior is still a bit different, such as calcite mysql library(it limits 4 chars) now
```
mysql> SELECT SOUNDEX('Quadratically');
         -> 'Q363'
```

And mysql:
```
mysql> SELECT SOUNDEX('Quadratically');
         -> 'Q36324'
```

But this behavior is also different from spark, so we may need to deal with each dialect separately later. I will create an issue describing the syntax difference of this function in mysql when i'm free. WDYT?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
chucheng92: good point. updated it.

NobiGo:
@chucheng92 Please make the commit info stay the same as the Jira summary.

chucheng92:
> @chucheng92 Please make the commit info stay the same as the Jira summary.

@NobiGo thanks. I added a commit for letting commiters to review. should i merge to one commit and force-push it?

chucheng92:
> Looks pretty good I think. Is the behavior for the other dialects correct? I would imagine so but on the off chance it's not, we could avoid adding a whole extra operator. Otherwise just left a couple small comments.

Maybe the behavior is still a bit different, such as calcite mysql library(it limits 4 chars) now
```
mysql> SELECT SOUNDEX('Quadratically');
         -> 'Q363'
```

And mysql:
```
mysql> SELECT SOUNDEX('Quadratically');
         -> 'Q36324'
```

But this behavior is also different from spark, so we may need to deal with each dialect separately later. I will create an issue describing the syntax difference of this function in mysql when i'm free. WDYT?

chucheng92:
@tanclary thanks for reviewing. I have updated this pr.

tanclary (commented):
I think this is looking good, I just left one comment about the operator table entry. If you wouldn't mind changing the commit message from Soundex->SOUNDEX whenever you squash your commits that'd be great. Thank you!

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -717,6 +717,17 @@ public static String soundex(String s) {
     return SOUNDEX.soundex(s);
   }
 
+  /** SQL SOUNDEX(string) function but return original s when not mapped. */
+  public static String soundexSpark(String s) {
+    String res;
+    try {
```
olivrlee, line 723: I'd probably rewrite this as 
```
 try {
      return SOUNDEX.soundex(s);
  } catch (IllegalArgumentException e) {
      return s
  }
  ```

chucheng92: thanks, good point. fixed

chucheng92:
@tanclary @olivrlee hi, I have addressed the problems and rebased calcite-main, PTAL. thanks.

chucheng92:
@tanclary Hi, tanner. Main Code freeze has been closed. Can you help to merge this?

tanclary:
> @tanclary Hi, tanner. Main Code freeze has been closed. Can you help to merge this?

I saw you opened an issue for making an addition to SqlBasicFunction and I believe you referenced this case. Should we resolve that so this case can use that logic?

chucheng92:
> > @tanclary Hi, tanner. Main Code freeze has been closed. Can you help to merge this?
> 
> I saw you opened an issue for making an addition to SqlBasicFunction and I believe you referenced this case. Should we resolve that so this case can use that logic?

yes. can i add that method in this PR? or separate 2 PRs?

chucheng92:
@tanclary hi, tanner. I have add a method of `withReturnTypeInference` and use it, just like Julian suggested in the issue, we should squash it in this PR instead of open another PR. PTAL~  thanks a lot.

tanclary (commented):
I think this looks good and is just about ready. Would you mind writing a RelToSqlConverter test and ensure that it gets unparsed as SOUNDEX(..) and not SOUNDEX_SPARK(..)?  Sorry for the delay in review I have been out the past few days.

chucheng92:
> I think this looks good and is just about ready. Would you mind writing a RelToSqlConverter test and ensure that it gets unparsed as SOUNDEX(..) and not SOUNDEX_SPARK(..)? Sorry for the delay in review I have been out the past few days.

Hi, @tanclary Never mind, I got your point. I have added it. PTAL~

tanclary:
> > I think this looks good and is just about ready. Would you mind writing a RelToSqlConverter test and ensure that it gets unparsed as SOUNDEX(..) and not SOUNDEX_SPARK(..)? Sorry for the delay in review I have been out the past few days.
> 
> Hi, @tanclary I got your point. I have added it. PTAL.

Looks good. Squash your commits please and I will merge soon.

chucheng92:
> > > I think this looks good and is just about ready. Would you mind writing a RelToSqlConverter test and ensure that it gets unparsed as SOUNDEX(..) and not SOUNDEX_SPARK(..)? Sorry for the delay in review I have been out the past few days.
> > 
> > 
> > Hi, @tanclary I got your point. I have added it. PTAL.
> 
> Looks good. Squash your commits please and I will merge soon.

Done. thanks again.

PR #3308, opened 2023-07-11
Style:fix build mapping rule function formal parameter name

ForestLH:
The name of the parameter of this function is inconsistent with the actual meaning


chunweilei (commented):
I think you're right. So I am +1 for this change. But the commit message should be more clear.

ForestLH:
> I think you're right. So I am +1 for this change. But the commit message should be more clear.

Thank you reply   :)

ForestLH:
Does anyone can review my code?
It's best to merge,Or is there any problem, you can say.
Thank you very much.
@chunweilei

libenchao:
+1 too, and propose to rename the commit message to "Minor: the parameter names of SqlTypeMappingRules.Builder#add are misleading"

PR #3312, opened 2023-07-12
[CALCITE-5839] EnumerableInterpretable#StaticFieldDetector can overwrite its flag and return an incorrect result

rubenada:
See details in CALCITE-5839


thomasrebele:
LGTM

PR #3313, opened 2023-07-12
[CALCITE-5841] Improve the logic of creating JDBCMeta singleton in calcite plus

chucheng92:

Improve the logic of creating JDBCMeta singleton for potential multi-thread problem.

# Brief change log
Add double-checking logic for creating instance.

# Verifying this change

A small work, java behavior ensure its correctness. And also current total cases could cover it.

### DIFF
--- a/plus/src/main/java/org/apache/calcite/chinook/ChinookAvaticaServer.java
+++ b/plus/src/main/java/org/apache/calcite/chinook/ChinookAvaticaServer.java
@@ -62,16 +62,18 @@ public class ChinookAvaticaServer {
     private static final CalciteConnectionProvider CONNECTION_PROVIDER =
         new CalciteConnectionProvider();
 
-    private static JdbcMeta instance = null;
+    private static volatile JdbcMeta instance = null;
 
     private static JdbcMeta getInstance() {
       if (instance == null) {
-        try {
-          instance =
-              new JdbcMeta(CalciteConnectionProvider.DRIVER_URL,
-                  CONNECTION_PROVIDER.provideConnectionInfo());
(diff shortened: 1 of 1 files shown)


plus/src/main/java/org/apache/calcite/chinook/ChinookAvaticaServer.java
```diff
@@ -62,16 +62,18 @@ public static class CalciteChinookMetaFactory implements Meta.Factory {
...
 
     private static JdbcMeta getInstance() {
       if (instance == null) {
-        try {
-          instance =
-              new JdbcMeta(CalciteConnectionProvider.DRIVER_URL,
-                  CONNECTION_PROVIDER.provideConnectionInfo());
-        } catch (SQLException | IOException e) {
-          throw new RuntimeException(e);
+        synchronized (JdbcMeta.class) {
```
libenchao, line 69: How about just synchronizing on `CalciteChinookMetaFactory.class`? (Same to `RawChinookMetaFactory`)

libenchao (commented):
@chucheng92 Thanks for your PR, it looks great. I'll left one minor comment.

plus/src/main/java/org/apache/calcite/chinook/ChinookAvaticaServer.java
chucheng92: @chucheng92 hi, benchao. thanks for reviewing. updated.

libenchao (approved):
+1, merging

libenchao:
reopened, because we're in the code freeze period for releasing 1.35.0

PR #3314, opened 2023-07-13
[CALCITE-5843] Constant expression with nested casts causes a compiler crash

mihaibudiu:
I wasn't entirely sure how to handle conversions to other types than the numeric ones, so I left the previous behavior unchanged, but there may be other problems lurking.

### DIFF
--- a/linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
+++ b/linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
@@ -570,7 +570,15 @@ public abstract class Expressions {
           value = new BigInteger(stringValue);
         }
         if (primitive != null) {
-          value = primitive.parse(stringValue);
+          if (value instanceof Number) {
+            Number valueNumber = (Number) value;
+            value = primitive.numberValue(valueNumber);
+            if (value == null) {
+              value = primitive.parse(stringValue);
+            }
+          } else {
+            value = primitive.parse(stringValue);
+          }
         }
       }
     }
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -687,6 +687,17 @@ public class SqlOperatorTest {
         "654342432412312");
   }
 
+  /**
+   * Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-5843">
+   * Constant expression with nested casts causes a compiler crash</a>. */
+  @Test public void testConstantCast() {
+    SqlOperatorFixture f = fixture();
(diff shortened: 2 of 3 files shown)


mihaibudiu:
Can someone please review this bug fix?
It remedies a Calcite crash.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
```diff
@@ -570,7 +570,37 @@ public static ConstantExpression constant(@Nullable Object value, Type type) {
           value = new BigInteger(stringValue);
         }
         if (primitive != null) {
-          value = primitive.parse(stringValue);
+          if (value instanceof Number) {
```
rubenada, line 573: Would it make sense to put this logic inside `Primitive` class, in a new method (something similar to the existing `Primitive#parse`) e.g.:
```
public enum Primitive {
...
  public Object numberValue(Number number) {
    switch (this) {
    case BYTE:
      return number.byteValue();
    case CHAR:
      return (char) number.intValue();
    case DOUBLE:
      return number.doubleValue();
    case FLOAT:
      return number.floatValue();
    case INT:
      return number.intValue();
    case LONG:
      return number.longValue();
    case SHORT:
      return number.shortValue();
    default:
      throw new AssertionError(number);
    }
  }
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -687,6 +687,17 @@ void testCastToExactNumeric(CastType castType, SqlOperatorFixture f) {
         "654342432412312");
   }
 
+  /**
+   * Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-5843">
+   * Legal constant SQL expression with cast causes a parse error</a>. */
```
rubenada, line 692: Please align the javadoc with the new Jira/commit title

mihaibudiu:
@rubenada hopefully I have addressed your comments. The commits are already squashed.

rubenada:
Thanks @mihaibudiu , IMO the PR looks in a good shape.
I'll merge it in the next days if no other remarks appear.

BTW, as a general note, it's not a big deal but usually we prefer to keep the commit history in the PR (with several commits) to facilitate the work of reviewers who may want to take a look at the evolution and different steps taken during the review process. Once all remarks have been addressed and the PR is in a good shape and ready to merge, we ask to squash commits into a single one right before the merge.

rubenada:
@mihaibudiu I think we can move on and merge this patch.
One final detail, could you please amend the commit message so that it starts with the ticket number? `[CALCITE-5843] ...`

mihaibudiu:
> @mihaibudiu I think we can move on and merge this patch. One final detail, could you please amend the commit message so that it starts with the ticket number? `[CALCITE-5843] ...`

Done.

PR #3317, opened 2023-07-14
[CALCITE-5826] Add FIND_IN_SET function (enabled in Hive and Spark libraries)

herunkang2018:


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1076,6 +1076,32 @@ public static int levenshtein(String string1, String string2) {
     return LEVENSHTEIN_DISTANCE.apply(string1, string2);
   }
 
+  /** SQL FIND_IN_SET(string, stringArray) function. */
+  public static int findInSet(String matchStr, String textStr) {
+    if (matchStr.contains(",")) {
```
LakeShen, line 1081: Why you add this check? Could you give the code comments to explain?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1076,6 +1076,32 @@ public static int levenshtein(String string1, String string2) {
     return LEVENSHTEIN_DISTANCE.apply(string1, string2);
   }
 
+  /** SQL FIND_IN_SET(string, stringArray) function. */
+  public static int findInSet(String matchStr, String textStr) {
```
LakeShen, line 1080: Because this method is public, I think we should handle the cases where matchStr and textStr are null

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1076,6 +1076,32 @@ public static int levenshtein(String string1, String string2) {
...
+    if (matchStr.contains(",")) {
+      return 0;
+    }
+    int textStrLen = textStr.length();
```
LakeShen, line 1084: final int textStrLen = textStr.length();

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1076,6 +1076,32 @@ public static int levenshtein(String string1, String string2) {
...
+    int matchStrLen = matchStr.length();
```
LakeShen, line 1085: final int matchStrLen = matchStr.length();

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1076,6 +1076,32 @@ public static int levenshtein(String string1, String string2) {
...
+    int n = 1;
+    int lastComma = -1;
```
LakeShen, line 1087: lastComma -> lastCommaIndex or lastCommaPosition maybe better.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1076,6 +1076,32 @@ public static int levenshtein(String string1, String string2) {
...
+    for (int i = 0; i < textStrLen; i++) {
+      if (textStr.charAt(i) == ',') {
```
LakeShen, line 1089: Now ',' is hard coding,are there some character constants that represent this.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1076,6 +1076,32 @@ public static int levenshtein(String string1, String string2) {
...
+        if (i - (lastComma + 1) == matchStrLen
+            && textStr.substring(lastComma + 1, i).equals(matchStr)) {
+          return n;
+        }
+        lastComma = i;
+        n++;
+      }
+    }
+    if (textStrLen - (lastComma + 1) == matchStrLen
+        && textStr.substring(lastComma + 1, textStrLen).equals(matchStr)) {
```
LakeShen, line 1099: If there are more code comments would help people understand better,I think you could give more code comments in this method:)

LakeShen:
@herunkang2018  ,thank you for opening this PR, I left a couple of comments,overall looks good.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: This check is from the function specification, which refered to [Spark's doc](https://spark.apache.org/docs/latest/api/sql/index.html#find_in_set). I will add function annotation to address the meaning of this function.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: For null check issue, we can define the null policy of the function, and null policy will handle null properly in code generation phase, so I think there is no need to implement null check in each function implementation additionally.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: Thanks for catching this, I will improve that.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: Thanks for the comment. Updated.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: Updated.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: Resolved.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: Resolved.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: The function annotatio is added.

herunkang2018:
@LakeShen Thanks a lot for the review. All comments resolved, please help to review again if you have time.

LakeShen:
@herunkang2018 ,LGTM.

And need the second people to review it:)

site/_docs/reference.md
```diff
@@ -2729,6 +2729,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | o | EXTRACT(xml, xpath, [, namespaces ])           | Returns the XML fragment of the element or elements matched by the XPath expression. The optional namespace value that specifies a default mapping or namespace mapping for prefixes, which is used when evaluating the XPath expression
 | o | EXISTSNODE(xml, xpath, [, namespaces ])        | Determines whether traversal of a XML document using a specified xpath results in any nodes. Returns 0 if no nodes remain after applying the XPath traversal on the document fragment of the element or elements matched by the XPath expression. Returns 1 if any nodes remain. The optional namespace value that specifies a default mapping or namespace mapping for prefixes, which is used when evaluating the XPath expression.
 | m | EXTRACTVALUE(xml, xpathExpr))                  | Returns the text of the first text node which is a child of the element or elements matched by the XPath expression.
+| h s | FIND_IN_SET(string, stringArray)             | Returns the index (1-based) of the given *string* in the comma-delimited *stringArray*. Returns 0, if the given *string* was not found or if *string* contains a comma
```
mihaibudiu, line 2732: From this description I infer that stringArray is of type `ARRAY STRING`, however it seems to be just a string which may contain commas. (I also think that the function name is not a very good choice, but that cannot be changed.)

herunkang2018: Good advice, the parameter name can be improved to a more concise one, like `matchStr` and `textStr`, to avoid ambiguity.

herunkang2018: I think we can add an example to clarify it, like `FIND_IN_SET('bc', 'a,bc,def')` returns TRUE.

herunkang2018: Updated.

site/_docs/reference.md
```diff
@@ -2729,6 +2729,7 @@ BigQuery's type system uses confusingly different names for types and functions:
...
 | o | EXISTSNODE(xml, xpath, [, namespaces ])        | Determines whether traversal of a XML document using a specified xpath results in any nodes. Returns 0 if no nodes remain after applying the XPath traversal on the document fragment of the element or elements matched by the XPath expression. Returns 1 if any nodes remain. The optional namespace value that specifies a default mapping or namespace mapping for prefixes, which is used when evaluating the XPath expression.
 | m | EXTRACTVALUE(xml, xpathExpr))                  | Returns the text of the first text node which is a child of the element or elements matched by the XPath expression.
+| h s | FIND_IN_SET(string, stringArray)             | Returns the index (1-based) of the given *string* in the comma-delimited *stringArray*. Returns 0, if the given *string* was not found or if *string* contains a comma. For example, FIND_IN_SET('bc', 'a,bc,def') returns 2
```
JiajunBernoulli, line 2732: `*stringArray*` is string not array. Perhaps it's confusing.

How about `*strings*`, `mkStrings` or `*joinStrings*`?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
JiajunBernoulli: Agree with @LakeShen.

Check null for public method always valuable. 

And your unit test need have null parameter case.

site/_docs/reference.md
herunkang2018: Thanks for the advice. Updated.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: I have added the null check. Thanks for the review.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1076,6 +1078,41 @@ public static int levenshtein(String string1, String string2) {
...
+    }
+    if (matchStr.contains(String.valueOf(COMMA_DELIMITER))) {
+      return 0;
+    }
+    final int textStrLen = textStr.length();
+    final int matchStrLen = matchStr.length();
+    int n = 1;
+    int lastCommaIndex = -1;
+    for (int i = 0; i < textStrLen; i++) {
+      if (textStr.charAt(i) == COMMA_DELIMITER) {
```
JiajunBernoulli, line 1100: Why not use `textStr.split(",")` ?
It will be easier to understand.

herunkang2018: Thanks for the advice. Current implementation is consistent with Spark's. For better code readability, change to use `split` is good, but some performance may be reduced since for loop can early exit when matched and `split` cannot.

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: Updated.

libenchao:
This PR has received 1 approval from @mihaibudiu . Do you have more concerns @LakeShen @JiajunBernoulli ? Let's try to get it in 1.36.0

LakeShen (approved):

LakeShen:
> This PR has received 1 approval from @mihaibudiu . Do you have more concerns @LakeShen @JiajunBernoulli ? Let's try to get it in 1.36.0

LGTM.

JiajunBernoulli (approved):

PR #3318, opened 2023-07-14
[CALCITE-5825] Add URL_ENCODE and URL_DECODE function (enabled in Spark library)

herunkang2018:


core/src/main/java/org/apache/calcite/runtime/UrlFunctions.java
```diff
@@ -0,0 +1,58 @@
...
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
+import static org.apache.calcite.util.Static.RESOURCE;
+
+/**
+ * A collection of functions used in Url processing.
```
LakeShen, line 28: Is UrlFunctions only available in Spark?

If so, I think the comments could be more detailed in the description

core/src/main/java/org/apache/calcite/runtime/UrlFunctions.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  private static final Charset UTF_8 = StandardCharsets.UTF_8;
+
+  /** The "URL_DECODE(string)" function. */
+  public static String urlDecode(String value) {
+    String url;
+    try {
+      url = URLDecoder.decode(value, UTF_8.name());
+    } catch (UnsupportedEncodingException e) {
```
LakeShen, line 42: Besides UnsupportedEncodingException, are there any other exceptions to handle?

Like some RuntimeException,such as NullPointerException.

core/src/main/java/org/apache/calcite/runtime/UrlFunctions.java
```diff
@@ -0,0 +1,58 @@
...
+      throw RESOURCE.charsetEncoding(value, UTF_8.name()).ex();
+    }
+    return url;
+  }
+
+  /** The "URL_ENCODE(string)" function. */
+  public static String urlEncode(String url) {
```
LakeShen, line 49: Same as above

core/src/main/java/org/apache/calcite/runtime/UrlFunctions.java
herunkang2018: This is an util class for all the general Url processing funcitons, not only for Spark.

core/src/main/java/org/apache/calcite/runtime/UrlFunctions.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public class UrlFunctions {
+
+  private UrlFunctions() {
+  }
+
+  private static final Charset UTF_8 = StandardCharsets.UTF_8;
+
+  /** The "URL_DECODE(string)" function. */
+  public static String urlDecode(String value) {
```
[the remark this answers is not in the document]
herunkang2018: Thanks for catching this. The correct behavior is to return empty string. I will add two unit tests to cover it.

core/src/main/java/org/apache/calcite/runtime/UrlFunctions.java
herunkang2018: Thanks for remind. `RuntimeException` is not need to explicitly throw, so easy to forget to handle it. I check the`RuntimeException` in `URLDecoder.decode`, there will be `IllegalArgumentException` when encounter the unencoded `%`.

The next I will do are two things:
1. Keep current exception handling way. Since `IllegalArgumentException` exception messages are very diverse, I think  that not handling with it explicitly here is reasonable, and just throws `IllegalArgumentException` when it encountered.
2. Add exception test cases to cover `IllegalArgumentException`.

core/src/main/java/org/apache/calcite/runtime/UrlFunctions.java
herunkang2018: There will be `RuntimeException`, including `NullPointerException`/`IllegalCharsetNameException`/`UnsupportedCharsetException`, since we use UTF-8 charset, and this charset always exist. So these `RuntimeException` will never appear. I think we can ignore them.

core/src/main/java/org/apache/calcite/runtime/UrlFunctions.java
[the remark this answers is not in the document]
herunkang2018: For null check, the `ARG0` of NullPolicy will do null check in code generation phase, so I think there is no need to implement null check in each function implementation additionally.

core/src/main/java/org/apache/calcite/runtime/UrlFunctions.java
herunkang2018: The test cases to cover IllegalArgumentException is added.

herunkang2018:
@LakeShen Thanks a lot for the review. All comments resolved, please help to review again if you have time.

core/src/main/java/org/apache/calcite/runtime/UrlFunctions.java
LakeShen: Get it

LakeShen:
@herunkang2018 LGTM.

And need the second people to review it:)

site/_docs/reference.md
```diff
@@ -2854,6 +2854,8 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | UNIX_MILLIS(timestamp)                         | Returns the number of milliseconds since 1970-01-01 00:00:00
 | b | UNIX_SECONDS(timestamp)                        | Returns the number of seconds since 1970-01-01 00:00:00
 | b | UNIX_DATE(date)                                | Returns the number of days since 1970-01-01
+| s | URL_DECODE(string)                             | Decodes a *string* in 'application/x-www-form-urlencoded' format using a specific encoding scheme
```
mihaibudiu, line 2857: You should specify what happens if the function fails to encode/decode. Is the result NULL or a runtime error?

mihaibudiu (approved):

site/_docs/reference.md
herunkang2018: Thanks for remind, I will check it soon.

herunkang2018: It will return original string when decoded error. I have fixed this and added the behaviour to doc.

JiajunBernoulli (approved):
The PR looks good, would you please squash your commits?

herunkang2018:
@JiajunBernoulli Sure, the commit is squashed.

LakeShen (approved):

PR #3319, opened 2023-07-15
[CALCITE-5848] Add BIT_GET and GETBIT function (enabled in Spark library)

herunkang2018:


herunkang2018:
@tanclary would you like to help review if you have time? Thanks very much.

site/_docs/reference.md
```diff
@@ -2728,6 +2729,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | FORMAT_DATETIME(string, timestamp)             | Formats *timestamp* according to the specified format *string*
 | b | FORMAT_TIME(string, time)                      | Formats *time* according to the specified format *string*
 | b | FORMAT_TIMESTAMP(string timestamp)             | Formats *timestamp* according to the specified format *string*
+| s | GETBIT(value, position)                        | Returns the bit (0 or 1) value at the specified *position* of *value*. The positions are numbered from right to left, starting at zero. The *position* argument cannot be negative
```
zoudan, line 2732: Could be simplify to 'Equivalent to `BIT_GET(value, position)`' ?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -3947,6 +3947,56 @@ static void checkRlikeFails(SqlOperatorFixture f) {
     f1.checkNull("BIT_LENGTH(cast(null as binary))");
   }
 
+  @Test void testBitGetFunc() {
```
zoudan, line 3950: Maybe we could reuse the test code for BIT_GET and GETBIT, as the only difference between them is the function name.

site/_docs/reference.md
```diff
@@ -2678,6 +2678,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | * | ATANH(numeric)                                 | Returns the inverse hyperbolic tangent of *numeric*
 | s | BIT_LENGTH(binary)                             | Returns the bit length of *binary*
 | s | BIT_LENGTH(string)                             | Returns the bit length of *string*
+| s | BIT_GET(value, position)                       | Returns the bit (0 or 1) value at the specified *position* of *value*. The positions are numbered from right to left, starting at zero. The *position* argument cannot be negative
```
zoudan, line 2681: What types are supported for `value`?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -815,6 +815,39 @@ public static int bitLength(ByteString s) {
     return s.length() * 8;
   }
 
+  /** SQL BIT_GET(value, position) function. */
+  public static int bitGet(long value, int position) {
```
zoudan, line 819: Shall we take other number types like Decimal into consideration?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1842,4 +1842,17 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
           ReturnTypes.INTEGER_NULLABLE,
           OperandTypes.or(OperandTypes.CHARACTER, OperandTypes.BINARY),
           SqlFunctionCategory.NUMERIC);
+
+  /** The "BIT_GET(value, position)" function. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlBasicFunction BIT_GET =
+      SqlBasicFunction.create("BIT_GET",
+          ReturnTypes.INTEGER_NULLABLE,
```
zoudan, line 1850: Are you sure that the return type is `INTEGER_NULLABLE`?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -815,6 +815,39 @@ public static int bitLength(ByteString s) {
...
+
+  /** SQL BIT_GET(value, position) function. */
+  public static int bitGet(byte value, int position) {
+    checkPosition(position, java.lang.Byte.SIZE);
+    return (value >> position) & 1;
+  }
+
+  private static void checkPosition(int position, int size) {
+    if (position < 0) {
+      throw RESOURCE.illegalNegativeBitGetPosition().ex();
```
zoudan, line 844: It would be better to add `position` in the error message

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -815,6 +815,39 @@ public static int bitLength(ByteString s) {
...
+    }
+    if (size <= position) {
+      throw RESOURCE.illegalBitGetPositionExceedsLimit().ex();
```
zoudan, line 847: It would be better to add `size` and `position` in the error message

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -815,6 +815,39 @@ public static int bitLength(ByteString s) {
...
+  /** SQL BIT_GET(value, position) function. */
+  public static int bitGet(long value, int position) {
+    checkPosition(position, java.lang.Long.SIZE);
+    return (int) ((value >> position) & 1);
+  }
+
+  /** SQL BIT_GET(value, position) function. */
+  public static int bitGet(int value, int position) {
+    checkPosition(position, java.lang.Integer.SIZE);
+    return (value >> position) & 1;
```
zoudan, line 827: Is this expected behavior when it is the signed bit?

site/_docs/reference.md
herunkang2018: It would be more concise. I will fix it soon.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
herunkang2018: Would you mean to reuse the test code by parameterized test?
By the way, currently `testGetBitFunc` only contains basic test, and `testBitGetFunc` contains the full test cases.

site/_docs/reference.md
herunkang2018: The supported type is `NUMERIC` type.

herunkang2018: I will update the doc to address this.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
herunkang2018: According to Spark's [PR](https://github.com/apache/spark/pull/31198/files) to support `BIT_GET` function, the return type should be `TINYINT`. I will fix this soon.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: According to Spark's [PR](https://github.com/apache/spark/pull/31198/files) to implement BIT_GET function, currently we don't support decimal type. For consistency, I think we should not take other number types like Decimal into consideration.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: Good advice, I will update it soon.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
zoudan: Yes, in that way we could test all cases for both of them.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
herunkang2018: I checked this is the expected behavior. This implementation is the same as Spark's and just returns the bit of the position, not specially considering the meaning of signed bit. I find that currently the related test case is missing in Spark, and I will add it in this PR.
Besides, the test cases for negative value are missing. I will add them to improve code coverage.

herunkang2018:
@zoudan Thanks for the review.
All comments resolved, please help to review if you have time.

zoudan:
LGTM, would you mind rebase and squash your commits?

herunkang2018:
@zoudan The commits is squashed. Thanks for the valuable comments in the review.

zoudan:
Sorry, one minor comment use `functions` instead of `function` in the commit message.

herunkang2018:
@zoudan Good catch. This is fixed.

PR #3320, opened 2023-07-16
[CALCITE-5830] Add ARRAY_INSERT function(enabled in Spark library)

chucheng92:

Add array_insert(x, pos, val) function(enabled in Spark library).

Places val into index pos of array x. Array indices start at 1, or start from the end if index is negative. Index above array size appends the array, or prepends the array if index is negative, with 'null' elements

issue: https://issues.apache.org/jira/browse/CALCITE-5830

Examples:

> SELECT array_insert(array(1, 2, 3, 4), 5, 5); [1,2,3,4,5]

> SELECT array_insert(array(5, 3, 2, 1), -3, 4); [5,4,3,2,1] 

https://spark.apache.org/docs/latest/api/sql/index.html#array_insert

# Brief change log
Add array_insert(x, pos, val) function(enabled in Spark library).

# Verifying this change

SqlOperatorTests#testArrayInsertFunc()

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -147,6 +147,12 @@ public class SqlFunctions {
 
   private static final Base32 BASE_32 = new Base32();
 
+  // Some JVMs can't allocate arrays of length Integer.MAX_VALUE; actual max is somewhat smaller.
+  // Be conservative and lower this value a little.
+  // @see http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/util/ArrayList.java#l229
+  // Note: this variable handling is inspired by Apache Spark
+  private static final int MAX_ARRAY_LENGTH = Integer.MAX_VALUE - 15;
+
   private static final Function1<List<Object>, Enumerable<Object>> LIST_AS_ENUMERABLE =
       a0 -> a0 == null ? Linq4j.emptyEnumerable() : Linq4j.asEnumerable(a0);
 
@@ -4105,6 +4111,94 @@ public class SqlFunctions {
     final Set result = new LinkedHashSet<>(list1);
     result.removeAll(list2);
     return new ArrayList<>(result);
+  }
+
+  /** Support the ARRAY_INSERT function. */
+  public static @Nullable List arrayInsert(List baselist, Object pos, Object val) {
+    if (baselist == null || pos == null) {
+      return null;
+    }
+    int posInt = (int) pos;
+    Object[] baseArray = baselist.toArray();
+    if (posInt == 0 || posInt >= MAX_ARRAY_LENGTH || posInt <= -MAX_ARRAY_LENGTH) {
+      throw new IllegalArgumentException("The index 0 is invalid. "
+          + "An index shall be either < 0 or > 0 (the first element has index 1) "
+          + "and not exceeds the allowed limit.");
+    }
+
+    boolean usePositivePos = posInt > 0;
+
+    if (usePositivePos) {
+      int newArrayLength = Math.max(baseArray.length + 1, posInt);
+
+      if (newArrayLength > MAX_ARRAY_LENGTH) {
+        throw new IndexOutOfBoundsException(
+            String.format(Locale.ROOT, "The new array length %s exceeds the allowed limit.",
+                newArrayLength));
+      }
+
+      Object[] newArray = new Object[newArrayLength];
+
+      int posIndex = posInt - 1;
+      if (posIndex < baseArray.length) {
+        System.arraycopy(baseArray, 0, newArray, 0, posIndex);
+        newArray[posIndex] = val;
+        System.arraycopy(baseArray, posIndex, newArray, posIndex + 1, baseArray.length - posIndex);
+      } else {
+        System.arraycopy(baseArray, 0, newArray, 0, baseArray.length);
+        newArray[posIndex] = val;
+      }
+
+      return Arrays.asList(newArray);
+    } else {
+      int posIndex = posInt;
+
+      boolean newPosExtendsArrayLeft = baseArray.length + posIndex < 0;
+
+      if (newPosExtendsArrayLeft) {
+        // special case, if the new position is negative but larger than the current array size
+        // place the new item at start of array, place the current array contents at the end
+        // and fill the newly created array elements in middle with a null
+        int newArrayLength = -posIndex + 1;
+
+        if (newArrayLength > MAX_ARRAY_LENGTH) {
+          throw new IndexOutOfBoundsException(
+              String.format(Locale.ROOT, "The new array length %s exceeds the allowed limit.",
+                  newArrayLength));
+        }
+
+        Object[] newArray = new Object[newArrayLength];
+        System.arraycopy(baseArray, 0, newArray, Math.abs(posIndex + baseArray.length) + 1,
+            baseArray.length);
+        newArray[0] = val;
+
+        return Arrays.asList(newArray);
+      } else {
+        posIndex = posIndex + baseArray.length;
+
+        int newArrayLength = Math.max(baseArray.length + 1, posIndex + 1);
+
+        if (newArrayLength > MAX_ARRAY_LENGTH) {
+          throw new IndexOutOfBoundsException(
+              String.format(Locale.ROOT, "The new array length %s exceeds the allowed limit.",
+                  newArrayLength));
+        }
+
+        Object[] newArray = new Object[newArrayLength];
+
+        if (posIndex < baseArray.length) {
+          System.arraycopy(baseArray, 0, newArray, 0, posIndex);
+          newArray[posIndex] = val;
+          System.arraycopy(baseArray, posIndex, newArray, posIndex + 1,
+              baseArray.length - posIndex);
+        } else {
+          System.arraycopy(baseArray, 0, newArray, 0, baseArray.length);
+          newArray[posIndex] = val;
+        }
+
+        return Arrays.asList(newArray);
+      }
+    }
   }
 
   /** Support the ARRAY_INTERSECT function. */
--- a/core/src/main/java/org/apache/calcite/sql/type/ArrayInsertOperandTypeChecker.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ArrayInsertOperandTypeChecker.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.sql.type;
+
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.sql.SqlCallBinding;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlUtil;
+
+import com.google.common.collect.ImmutableList;
+
+import static org.apache.calcite.sql.type.NonNullableAccessors.getComponentTypeOrThrow;
+import static org.apache.calcite.util.Static.RESOURCE;
+
+/**
+ * Parameter type checking strategy, where the operand size is 3, and the first type is array,
+ * the second type is integer, and the third type is the component type of the first array (
+ * strictly matched without leastRestrictive conversion).
+ */
+public class ArrayInsertOperandTypeChecker extends SameOperandTypeChecker {
+
+  public ArrayInsertOperandTypeChecker() {
+    super(3);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public boolean checkOperandTypes(
+      SqlCallBinding callBinding,
+      boolean throwOnFailure) {
+    // all operands can't be null (without cast)
+    for (SqlNode node : callBinding.operands()) {
+      if (SqlUtil.isNullLiteral(node, false)) {
+        if (throwOnFailure) {
+          throw callBinding.getValidator().newValidationError(node,
+              RESOURCE.argumentMustNotBeNull(callBinding.getOperator().getName()));
+        } else {
+          return false;
+        }
+      }
+    }
+
+    // op0 is an array type
+    final SqlNode op0 = callBinding.operand(0);
+    if (!OperandTypes.ARRAY.checkSingleOperandType(
+        callBinding,
+        op0,
+        0,
+        throwOnFailure)) {
+      return false;
+    }
+
+    // op1 is an Integer type
+    final SqlNode op1 = callBinding.operand(1);
+    RelDataType op1Type = SqlTypeUtil.deriveType(callBinding, op1);
+    SqlTypeName op1TypeName = op1Type.getSqlTypeName();
+    if (op1TypeName != SqlTypeName.INTEGER) {
+      if (throwOnFailure) {
+        throw callBinding.newError(
+            RESOURCE.typeNotComparable(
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4107,6 +4113,97 @@ public static List arrayExcept(List list1, List list2) {
...
+
+      if (newArrayLength > MAX_ARRAY_LENGTH) {
+        throw new IndexOutOfBoundsException(
+            String.format(Locale.ROOT, "The new array length %s exceeds the allowed limit.",
+                newArrayLength));
+      }
+
+      Object[] newArray = new Object[newArrayLength];
+
+      int posIndex = posInt - 1;
```
NobiGo, line 4140: Can we use `list.add(pos,element)` ?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4107,6 +4113,97 @@ public static List arrayExcept(List list1, List list2) {
...
+        }
+      }
+
+      newArray[posIndex] = val;
+
+      return Arrays.asList(newArray);
+    } else {
+      int posIndex = posInt;
+
+      boolean newPosExtendsArrayLeft = -posIndex > baselist.size();
```
NobiGo, line 4155: How about we use 'posIndex+baselist.size()' >0 or <0? `-posIndex` maybe having a number overflow exception.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
chucheng92: @NobiGo I'm afraid not. 

because our logic here is to move the element on the right of the insertion point to the right, leave the left side unchanged, leaving a space for the target element. If we use `list.add` to insert the left element, there is no problem. When inserting the right element, because we want to reserve a position, it will go out of bounds at this time.

A simple case can explain:
List<String> list = new ArrayList<>();
list.add(0, "A");
list.add(1, "A"); 
list.add(3, "A"); // exception

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4107,6 +4113,97 @@ public static List arrayExcept(List list1, List list2) {
...
+    int posInt = (int) pos;
+    if (posInt == 0) {
+      throw new IllegalArgumentException("The index 0 is invalid. "
+          + "An index shall be either < 0 or > 0 (the first element has index 1)");
+    }
+
+    boolean usePositivePos = posInt > 0;
+
+    if (usePositivePos) {
+      int newArrayLength = Math.max(baselist.size() + 1, posInt);
```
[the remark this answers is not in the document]
chucheng92: and here posInt is positive, if posInt < 0, it will from left to insert. 

And the logic of inserting null is the same, as long as `pos` is greater than the length of the array, then calculate the new array length, and fill in the middle with `null`

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4107,6 +4113,97 @@ public static List arrayExcept(List list1, List list2) {
     return new ArrayList<>(result);
   }
 
+  /** Support the ARRAY_INSERT(array, pos, val) function. */
+  public static @Nullable List arrayInsert(List baselist, Object pos, Object val) {
```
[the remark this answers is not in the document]
chucheng92: @NobiGo e.g. array_insert(array(1,2,3), 1, 2) calcite will parse and search `arrayInsert[interface java.util.List, int, int]` signature, but it can't match `arrayInsert(List baselist, Integer pos, Object val)`. Java method reflection limits boxing and primitive types can't be matched.

we can other array function. i.e. array_position has same behavior.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4107,6 +4113,97 @@ public static List arrayExcept(List list1, List list2) {
     return new ArrayList<>(result);
   }
 
+  /** Support the ARRAY_INSERT(array, pos, val) function. */
```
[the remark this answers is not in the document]
chucheng92: fixed it.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
chucheng92: another case, if second pos is a null value. e.g. array_insert(array(1,2,3), cast(null as Integer), 2), then searched method is `arrayInsert[interface java.util.List, Integer, int]`.  We want to use `Integer` to cover `int and Integer`, but it can't be work.

chucheng92:
@NobiGo hi, sorry to ping you. I have solved some comments. If you have time, pls take a look. thanks.

site/_docs/reference.md
```diff
@@ -2658,6 +2658,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | s | ARRAY_CONTAINS(array, element)                 | Returns true if the *array* contains the *element*
 | s | ARRAY_DISTINCT(array)                          | Removes duplicate values from the *array* that keeps ordering of elements
 | s | ARRAY_EXCEPT(array1, array2)                   | Returns an array of the elements in *array1* but not in *array2*, without duplicates
+| s | ARRAY_INSERT(array, pos, element)              | Places *element* into index *pos* of *array*. Array index start at 1, or start from the end if index is negative. Index above array size appends the array, or prepends the array if index is negative, with 'null' elements.
```
NobiGo, line 2661: `NULL` value is better?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
NobiGo: I get it. Thanks for your explanation. How about we use the `System.arraycopy()` and `Arrays.fill()` to implement this? Others are good to me.

site/_docs/reference.md
chucheng92: yes. fixed

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
chucheng92: @NobiGo yes. you are right. `System.arraycopy()` may has better performance when array is large. I have fixed it. PTAL.

chucheng92:
@NobiGo hi, I have solved all your comments. If you have time, pls help to re-check it. thank you.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4107,6 +4113,93 @@ public static List arrayExcept(List list1, List list2) {
...
+    } else {
+      int posIndex = posInt;
+
+      boolean newPosExtendsArrayLeft = baseArray.length + posIndex < 0;
+
+      if (newPosExtendsArrayLeft) {
+        // special case, if the new position is negative but larger than the current array size
+        // place the new item at start of array, place the current array contents at the end
+        // and fill the newly created array elements in middle with a null
+        int newArrayLength = -posIndex + 1;
```
NobiGo, line 4161: same question -posIndex maybe having a number overflow exception? Maybe we can use:
`int newArrayLength = Integer.MAX_VALUE - (Integer.MAX_VALUE + posIndex);`
Then the next copy code will be :
`System.arraycopy(baseArray, 0, newArray, newArrayLength - baseArray.length + 1, baseArray.length);`

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
```diff
@@ -658,6 +658,7 @@ public enum BuiltInMethod {
   ARRAYS_OVERLAP(SqlFunctions.class, "arraysOverlap", List.class, List.class),
   ARRAYS_ZIP(SqlFunctions.class, "arraysZip", List.class, List.class),
   SORT_ARRAY(SqlFunctions.class, "sortArray", List.class, boolean.class),
+  ARRAY_INSERT(SqlFunctions.class, "arrayInsert", List.class, Integer.class, Object.class),
```
NobiGo, line 661: Put this line to `https://github.com/apache/calcite/pull/3320/files#diff-5207c1dd4af072b3ecbebf9bf87ab0677171bd47309eda809013ef66ae9f77d6R655`. Alphabetic order.

site/_docs/reference.md
```diff
@@ -2658,6 +2658,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | s | ARRAY_CONTAINS(array, element)                 | Returns true if the *array* contains the *element*
 | s | ARRAY_DISTINCT(array)                          | Removes duplicate values from the *array* that keeps ordering of elements
 | s | ARRAY_EXCEPT(array1, array2)                   | Returns an array of the elements in *array1* but not in *array2*, without duplicates
+| s | ARRAY_INSERT(array, pos, element)              | Places *element* into index *pos* of *array*. Array index start at 1, or start from the end if index is negative. Index above array size appends the array, or prepends the array if index is negative, with 'NULL' elements.
```
NobiGo, line 2661: Usually, We use `` to bracket the keyword. Like: https://github.com/apache/calcite/pull/3320/files#diff-9cda2b29a1b9206e0daa6e6d722eb476575f83eadea1a6e302225afd08d9f0d2R2663

NobiGo (approved):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
chucheng92: thanks for reminding me. however, we can not use
`int newArrayLength = Integer.MAX_VALUE - (Integer.MAX_VALUE + posIndex);`
if posIndex = Integer.MIN_VALUE, it will be 
`int newArrayLength = Integer.MAX_VALUE + 1;` and overflow.

In fact, pos cannot exceed the maximum array limit, we can check in advance just like pos cannot be 0. and -pos will not overflow in subsequent code logic.

I have updated the PR and add overflow test cases. If you are free, PTAL.

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
chucheng92: fixed

site/_docs/reference.md
chucheng92: fixed

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
mihaibudiu: why not just compute using long values?

chucheng92: > why not just compute using long values?

thanks for suggestions. the `posIndex` should < Integer.MAX_VALUE - 15, we can simply use Integer to cover it and without overflow. Otherwise, i think it's a good way to check the parameters in advance.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5845,6 +5845,84 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    f1.checkScalar("array_insert(array[1, 2, 3], 3, 4)",
+        "[1, 2, 4, 3]", "INTEGER NOT NULL ARRAY NOT NULL");
+    f1.checkScalar("array_insert(array[1, 2, 3], 3, cast(null as integer))",
+        "[1, 2, null, 3]", "INTEGER ARRAY NOT NULL");
+    f1.checkScalar("array_insert(array[2, 3, 4], 1, 1)",
+        "[1, 2, 3, 4]", "INTEGER NOT NULL ARRAY NOT NULL");
+    f1.checkScalar("array_insert(array[1, 3, 4], -2, 2)",
+        "[1, 2, 3, 4]", "INTEGER NOT NULL ARRAY NOT NULL");
+    f1.checkScalar("array_insert(array[2, 3, null, 4], -5, 1)",
+        "[1, null, 2, 3, null, 4]", "INTEGER ARRAY NOT NULL");
```
JiajunBernoulli, line 5923: Why not `[1, 2, 3, null, 4]`?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5845,6 +5845,84 @@ private static void checkIf(SqlOperatorFixture f) {
...
+        "INTEGER is not comparable to CHAR", false);
+
+    f1.checkScalar("array_insert(array[1, 2, 3], 3, 4)",
+        "[1, 2, 4, 3]", "INTEGER NOT NULL ARRAY NOT NULL");
+    f1.checkScalar("array_insert(array[1, 2, 3], 3, cast(null as integer))",
+        "[1, 2, null, 3]", "INTEGER ARRAY NOT NULL");
+    f1.checkScalar("array_insert(array[2, 3, 4], 1, 1)",
+        "[1, 2, 3, 4]", "INTEGER NOT NULL ARRAY NOT NULL");
+    f1.checkScalar("array_insert(array[1, 3, 4], -2, 2)",
+        "[1, 2, 3, 4]", "INTEGER NOT NULL ARRAY NOT NULL");
```
JiajunBernoulli, line 5921: Why not `[1, 3, 2, 4]`?

NobiGo (approved):
LGTM

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
chucheng92: when index is negative, it means 'prepends'. 

array_insert(array[1, 3, 4], -2, 2) => [1, 2, 3, 4]
when we reversed the input array(appends):
array_insert(array[4, 3, 1], 2, 2) => [4, 3, 2, 1]

spark make sure the insert behavior can be equivalent from left or right.

however, [1, 3, 2, 4] is wrong. it's 'appends' behavior.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
chucheng92: same as the question below, i get your point. and I have replied some details in the issue. thanks for reviewing.

JiajunBernoulli (approved):

core/src/main/java/org/apache/calcite/sql/type/ArrayInsertOperandTypeChecker.java
```diff
@@ -0,0 +1,95 @@
...
+    if (!OperandTypes.ARRAY.checkSingleOperandType(
+        callBinding,
+        op0,
+        0,
+        throwOnFailure)) {
+      return false;
+    }
+
+    // op1 is an Integer type
+    final SqlNode op1 = callBinding.operand(1);
```
liuyongvs, line 66: op1 is an Integer type  +1 great, which aligns with spark

core/src/main/java/org/apache/calcite/sql/type/ArrayInsertOperandTypeChecker.java
```diff
@@ -0,0 +1,95 @@
...
+      return false;
+    }
+
+    // op2 must has same type with op0 component type strictly
+    final RelDataType op0ComponentType =
+        getComponentTypeOrThrow(SqlTypeUtil.deriveType(callBinding, op0));
+    SqlNode op2 = callBinding.operand(2);
+    RelDataType op2Type = SqlTypeUtil.deriveType(callBinding, op2);
+    if (op2Type.getSqlTypeName() != op0ComponentType.getSqlTypeName()) {
+      if (throwOnFailure) {
```
liuyongvs, line 84: op2Type.getSqlTypeName() != op0ComponentType.getSqlTypeName())
it may not enougn, when the type is also complex type, like array/row/spark. for example row(int) and row(int , bigint)

liuyongvs, line 84: op2Type.getSqlTypeName() != op0ComponentType.getSqlTypeName())
it may not enougn, when the type is also complex type, like array/row/spark. for example row(int) and row(int , bigint)

liuyongvs:
looks great, only left one comments, which should be changed @chucheng92

chucheng92:
> looks great, only left one comments, which should be changed @chucheng92

thanks Lau, you are right. here we need also using leastRestrictive type just like other array function. I have updated the PR and added the complex type cases. If you have time, pls check it. thanks.

liuyongvs (approved):

chucheng92:
Thank you all for patient reviewing, all comments are resolved. @NobiGo hi, xiong, If you help time, could you help to merge it?

PR #3322, opened 2023-07-18
Remove mentions of binary distribution from README

zabetak:
The project does not make binary releases at the moment so mentioning binary distribution in the README is misleading.

### DIFF
--- a/README
+++ b/README
@@ -1,15 +1,14 @@
 Apache Calcite release 1.34.0
 
-This is a source or binary distribution of Apache Calcite.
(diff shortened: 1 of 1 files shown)


chucheng92:
+1 for the patch. The old description is indeed very misleading.

chucheng92 (commented):
Maybe we should rebase the main? current calcite is released 1.35.0

zabetak:
Thanks for the review @chucheng92 !

PR #3324, opened 2023-07-19
[CALCITE-5862] Incorrect semantics of ARRAY function (Spark library) when elements have Numeric and Character types

chucheng92:

https://issues.apache.org/jira/browse/CALCITE-5862

# Brief change log
Fix incorrect spark semantics of array function when array elements have Numeric and Character types.

# Verifying this change

SqlOperatorTests#testArrayFunc()

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -1011,10 +1011,46 @@ public abstract class SqlLibraryOperators {
           .withKind(SqlKind.CONCAT_WS_MSSQL);
 
   private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
-    RelDataType type =
-        opBinding.getOperandCount() > 0
-            ? ReturnTypes.LEAST_RESTRICTIVE.inferReturnType(opBinding)
-            : opBinding.getTypeFactory().createUnknownType();
+    final List<RelDataType> operandTypes = opBinding.collectOperandTypes();
+
+    // only numeric & character types check
+    boolean hasNumeric = false;
+    boolean hasCharacter = false;
+    boolean hasOthers = false;
+    for (RelDataType type : operandTypes) {
+      SqlTypeFamily family = type.getSqlTypeName().getFamily();
+      requireNonNull(family, "array element type family");
+      switch (family) {
+      case NUMERIC:
+        hasNumeric = true;
+        break;
+      case CHARACTER:
+        hasCharacter = true;
+        break;
+      case NULL:
+        // skip it becase we allow null
+        break;
+      default:
+        hasOthers = true;
+        break;
+      }
+    }
+
+    RelDataType type;
+    boolean useCharacterTypes = hasNumeric && hasCharacter && !hasOthers;
+    if (useCharacterTypes) {
+      List<RelDataType> characterTypes =
+          // may include NULL literal
+          operandTypes.stream().filter(
+              t -> t.getSqlTypeName().getFamily() != SqlTypeFamily.NUMERIC)
+              .collect(Collectors.toList());
+      type = opBinding.getTypeFactory().leastRestrictive(characterTypes);
+    } else {
+      type =
+          opBinding.getOperandCount() > 0
+              ? ReturnTypes.LEAST_RESTRICTIVE.inferReturnType(opBinding)
+              : opBinding.getTypeFactory().createUnknownType();
+    }
     requireNonNull(type, "inferred array element type");
     return SqlTypeUtil.createArrayType(opBinding.getTypeFactory(), type, false);
   }
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -9941,6 +9941,14 @@ public class SqlOperatorTest {
         "[null, foo]", "CHAR(3) ARRAY NOT NULL");
     f2.checkScalar("array(null)",
         "[null]", "NULL ARRAY NOT NULL");
+    f2.checkScalar("array(1, 2, 'Hi')",
+        "[1, 2, Hi]", "CHAR(2) NOT NULL ARRAY NOT NULL");
+    f2.checkScalar("array(1, 2, 'Hi', 'Hello')",
+        "[1, 2, Hi, Hello]", "CHAR(5) NOT NULL ARRAY NOT NULL");
+    f2.checkScalar("array(1, 2, 'Hi', null)",
+        "[1, 2, Hi, null]", "CHAR(2) ARRAY NOT NULL");
+    f2.checkScalar("array(1, 2, 'Hi', cast(null as char(10)))",
+        "[1, 2, Hi, null]", "CHAR(10) ARRAY NOT NULL");
   }
 
   /**


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -955,10 +955,38 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
-        opBinding.getOperandCount() > 0
-            ? ReturnTypes.LEAST_RESTRICTIVE.inferReturnType(opBinding)
-            : opBinding.getTypeFactory().createUnknownType();
+    // only numeric & character types check
+    List<RelDataType> numericTypes = new ArrayList<>();
+    List<RelDataType> characterTypes = new ArrayList<>();
+    List<RelDataType> nullTypes = new ArrayList<>();
+    for (int i = 0; i < opBinding.getOperandCount(); i++) {
+      SqlTypeFamily family = opBinding.getOperandType(i).getSqlTypeName().getFamily();
+      if (family == SqlTypeFamily.NUMERIC) {
```
tanclary, line 964: can this be rewritten as a switch case?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -955,10 +955,38 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
+      } else if (family == SqlTypeFamily.CHARACTER) {
+        characterTypes.add(opBinding.getOperandType(i));
+      } else if (family == SqlTypeFamily.NULL) {
+        nullTypes.add(opBinding.getOperandType(i));
+      } else {
+        break;
+      }
+    }
+
+    boolean onlyNumericCharacterTypes = opBinding.getOperandCount() > 0
```
tanclary, line 975: I could be misunderstanding, but if you only care about the size of the lists and not the elements themselves, you could just use boolean flags to keep track of which types have appeared. I think it will simplify the expression below.

tanclary (commented):
Left a couple of comments. Thanks for filing and opening a fix for this bug!

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
tanclary:
```     final List<RelDataType> types = opBinding.collectOperandTypes();
    boolean hasNumeric = false;
    boolean hasCharacter = false;
    types.forEach(type -> {
			// null check here maybe
			switch (type.getSqlTypeName().getFamily())  {
			 case x:
			 		hasNumeric = true;
			 		break;
			 case y:
			 		hasCharacter = true;
			 		break;
			 default:
		 }
		 if hasNumeric || hasCharacter || c {....} ``` you could do something like this maybe? (Excuse the pseudocode it's just an example)

**Comment by chucheng92** [comment]
@tanclary thanks for reviewing. I will address these comments later.

File: core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java (hunk 2 above)
Comment (chucheng92) [reply core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java]: good point. I have fixed this. PTAL.

**Review by JiajunBernoulli — CHANGES_REQUESTED** [review]
I think array must have same type.

You should use `TypeCoercionImpl` to solve your problem.

**Comment by chucheng92** [comment]
> I think array must have same type.
> 
> You should use `TypeCoercionImpl` to solve your problem.

Hi, Jiajun. thanks for reviewing. 

However, this PR does not modify the `SameOperandTypeChecker` constraint of array. In fact, the current `SameOperandTypeChecker` allows character and numeric to exist at the same time and passes this restriction. It's just that inferReturnType does not handle the simultaneous existence of character and numeric. please see:

```
  public static final SqlFunction ARRAY =
      SqlBasicFunction.create("ARRAY",
          OperandTypes.SAME_VARIADIC);
```

https://github.com/apache/calcite/blob/bdfb17029f7e205f895dc3dfd0f37c8ff2520823/core/src/main/java/org/apache/calcite/sql/type/SameOperandTypeChecker.java#L92

The `SameOperandTypeChecker` will handle the special case of character, which itself allows character and numeric to coexist.

so currently run `select array(1, 2, 'Hi');` can pass the OperandTypeChecker but with NPE in return type checker. This pr is used to fix the inconsistent behavior of operandTypeChecker and returnTypeInference. WDYT?

**Comment by chucheng92** [comment]
btw, the `SameOperandTypeChecker` is a common class(If we must limit the strict same type, we should modify the OperandTypeChecker), however the `arrayReturnType` is a private method just for spark semantics. This PR only modify the spark `arrayReturnType` to solve this bug, I think it's reasonable solution. @tanclary @JiajunBernoulli looking forward to your opinion.

**Comment by chucheng92** [comment]
@tanclary @JiajunBernoulli hi, Tanner, Jiajun. Sorry to ping you, can you help to review again?

**Comment by chucheng92** [comment]
@JiajunBernoulli @tanclary Sorry to ping you. It has been a long time for this PR. 

Do you have any other concerns? If you have any comments, I hope to receive your feedback. If there are some problems in my implementation, maybe you can give some specific opinions. I will appreciate it. Having no one to review is more frustrating than modifying code. thanks.

**Comment by tanclary** [comment]
So to clarify: you can have an array with multiple types, but only if they are NUMERIC/CHARACTER? What if you have another combination like [1, Date '2008-12-25', 'a']?

**Comment by chucheng92** [comment]
> So to clarify: you can have an array with multiple types, but only if they are NUMERIC/CHARACTER? What if you have another combination like [1, Date '2008-12-25', 'a']?

@tanclary thanks for reviewing. I understand your point. however, array function use `SameOperandTypeChecker`, it will cause exception firstly and can not enter arrayReturnTypeInference when use `[1, Date '2008-12-25', 'a']`.(varchar is a special case, it will pass SameOperandTypeChecker in SqlTypeUtil.isComparable). So, can returnTypeInference rely on OperandTypeChecker?  if yes, then here we do not need other types check. and if no, we need other types check here. from your professional perspective, WDYT?

**Comment by chucheng92** [comment]
hi, @tanclary It’s been a long time since this PR was created, could we continue on this PR?

**Comment by tanclary** [comment]
@chucheng92 Yes I think this looks good but I will wait for @JiajunBernoulli to approve/give + 1 because they also had some comments if I remember correctly.

**Comment by JiajunBernoulli** [comment]
@chucheng92 , Compared to NPE, the is indeed an improvement.

LGTM, remember to squash your commits before merging.

**Comment by JiajunBernoulli** [comment]
The commit message should contain some info like `java. lang. NullPointerException: referred array element type`.

It can be easily found by other users.

**Review by JiajunBernoulli — APPROVED** [review]

**Comment by chucheng92** [comment]
@tanclary @JiajunBernoulli thanks for your patient review. all comments resolved. @tanclary If you have time, could you help to merge this?

**Comment by tanclary** [comment]
@chucheng92 Hi yes, would you mind editing the commit message to add a space between "function" and "(Spark library)"? Apologies for the nit. Re-ping me after and I will help merge 👍

**Comment by chucheng92** [comment]
> @chucheng92 Hi yes, would you mind editing the commit message to add a space between "function" and "(Spark library)"? Apologies for the nit. Re-ping me after and I will help merge 👍

Thanks, fixed it.

**Comment by MasseGuillaume** [comment]
This PR introduce a regression when building an array with structs.

I defined in my codebase `named_struct` that can return a struct type:

```java
public class NamedStructUtils {

  private NamedStructUtils() {
    throw new UnsupportedOperationException("This is a utility class and cannot be instantiated");
  }

  public static SqlBasicFunction create() {
    return SqlBasicFunction.create(
      "NAMED_STRUCT",
      NamedStructUtils::returnType,
      OPERAND_TYPE_CHECKER);
  }

  private static String namedStructRequiresLiteralStringKeysGotExpression() {
    return "named_struct key is not a string literal, got an expression";
  }

  private static String namedStructRequiresLiteralStringKeysGotOtherType(String tpe) {
    return String.format("named_struct key is not a string literal, found type '%s'", tpe);
  }

  private static String namedStructRequiresEvenNumberOfArgs() {
    return "named_struct requires an even number for arguments";
  }
  
  private static RelDataType returnType(SqlOperatorBinding operatorBinding) {
    final List<String> keys = new ArrayList<String>();
    final List<RelDataType> values = new ArrayList<RelDataType>();
    int i = 0;
    for (RelDataType type : operatorBinding.collectOperandTypes()) {
      if (i % 2 == 0) {
        Object key = operatorBinding.getOperandLiteralValue(i, Object.class);
        if (key == null) {
          throw new CalciteException(
            namedStructRequiresLiteralStringKeysGotExpression(), null);
        }
        if (key instanceof NlsString) {
          keys.add(((NlsString) key).getValue());
        } else {
          String tpe = key.getClass().getSimpleName();
          throw new CalciteException(
            namedStructRequiresLiteralStringKeysGotOtherType(tpe), null);
        }

      } else {
        values.add(type);
      }
      i++;
    }
    final RelDataTypeFactory typeFactory = operatorBinding.getTypeFactory();
    return typeFactory.createStructType(values, keys);
  }

  private static final SqlOperandTypeChecker TWO_OR_MORE =
      OperandTypes.variadic(SqlOperandCountRanges.from(2));

  private static final SqlOperandTypeChecker OPERAND_TYPE_CHECKER =
      new SqlOperandTypeChecker() {
            SqlCallBinding callBinding,
            boolean throwOnFailure) {
          final boolean isRangeOk = TWO_OR_MORE.checkOperandTypes(callBinding, throwOnFailure);
          final int operandCount = callBinding.operands().size();
          final boolean isEvenNumberOfArgs = operandCount % 2 != 0;

          if (throwOnFailure && isEvenNumberOfArgs) {
            throw new CalciteException(namedStructRequiresEvenNumberOfArgs(), null);
          }

          return isRangeOk && isEvenNumberOfArgs;
        }

          return TWO_OR_MORE.getOperandCountRange();
        }

        @Override public String getAllowedSignatures(SqlOperator op, String opName) {
          return opName + "(...)";
        }
      };
}
```

Since `SqlTypeName.ROW` has type family null: `ROW(PrecScale.NO_NO, false, Types.STRUCT, null),`
it fails the check `requireNonNull(family, "array element type family");`

eg 
```
array(named_struct('k', 'v'))
```

chucheng92:
@MasseGuillaume thanks for reporting this. I will try to reproduce/fix this.

PR #3325, opened 2023-07-19
[CALCITE-5863] Calcite rejects valid query with multiple ORDER BY columns and constant RANGE bounds in window functions

itiels:
Allow multiple ORDER BY columns with RANGE bounds in window functions, if both of the bounds are constant bounds (CURRENT ROW, UNBOUNDED PRECEDING, UNBOUNDED FOLLOWING)

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlWindow.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlWindow.java
@@ -610,7 +610,9 @@ public class SqlWindow extends SqlCall {
       // SQL03 7.10 Rule 11a
       if (orderList.size() > 0) {
         // if order by is a compound list then range not allowed
-        if (orderList.size() > 1 && !isRows()) {
+        if (orderList.size() > 1
+            && !isRows()
+            && !onlySymbolBounds(lowerBound, upperBound)) {
           throw validator.newValidationError(isRows,
               RESOURCE.compoundOrderByProhibitsRange());
         }
@@ -625,7 +627,9 @@ public class SqlWindow extends SqlCall {
         // requires an ORDER BY clause if frame is logical(RANGE)
         // We relax this requirement if the table appears to be
         // sorted already
-        if (!isRows() && !SqlValidatorUtil.containsMonotonic(scope)) {
+        if (!onlySymbolBounds(lowerBound, upperBound)
+            && !isRows()
+            && !SqlValidatorUtil.containsMonotonic(scope)) {
           throw validator.newValidationError(this,
               RESOURCE.overMissingOrderBy());
         }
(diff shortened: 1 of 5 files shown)


LakeShen:
Hi,could you make the title CALCITE-5863 to [CALCITE-5863]

core/src/main/java/org/apache/calcite/sql/SqlWindow.java
```diff
@@ -610,8 +610,10 @@ public boolean isAllowPartial() {
       if (orderList.size() > 0) {
         // if order by is a compound list then range not allowed
         if (orderList.size() > 1 && !isRows()) {
-          throw validator.newValidationError(isRows,
-              RESOURCE.compoundOrderByProhibitsRange());
+          if (!onlyLiteralBounds(lowerBound, upperBound)) {
```
LakeShen, line 613: Can you add a sql case to the misc.iq file to test the results？

itiels: Added test case

core/src/main/java/org/apache/calcite/sql/SqlWindow.java
```diff
@@ -660,6 +664,12 @@ public boolean isAllowPartial() {
     }
   }
 
+  private boolean onlyLiteralBounds(@Nullable SqlNode lowerBound, @Nullable SqlNode upperBound) {
```
mihaibudiu, line 667: This does not seem to check for what the compiler calls "literal", which is essentially a constant value.
You are checking for specific kinds of bounds. So perhaps this function is not named properly.

itiels: Naming is hard.. My thought was literal in the sense of `SqlLiteral`, but I can see why it's confusing. 
Changed it to `onlySymbolBounds`.

mihaibudiu (approved):

JiajunBernoulli:
Nice work, can we add some tests in RelOptRulesTest?
> Use some window rule like `CoreRules.WINDOW_REDUCE_EXPRESSIONS`.

itiels:
@JiajunBernoulli I'm assuming you've wanted a test to see that the `WINDOW_REDUCE_EXPRESSIONS` rule can handle the queries that were rejected by the validator before, as I haven't changed any rules code, So I've added such a test. please let me know if that's not what you've meant.

libenchao:
This has been reviewed by 3 committers, and with 1 approval. Could we merge it within 1.36.0?

LakeShen:
Hi @itiels ,please keep the same title with jira's title.

I find the jira's title is `Calcite rejects valid query with multiple ORDER BY columns and constant RANGE bounds in window functions`

LakeShen:
At the same time,the commits' information should be same with jira's title.

JiajunBernoulli:
I will approve if commit message  same with JIRA summary.

itiels:
@JiajunBernoulli @LakeShen should I squash the commits and rewrite the commit message?

itiels:
@LakeShen done. Thank you!

LakeShen (approved):

PR #3326, opened 2023-07-19
CALCITE-5860 Set scale when casting Decimal

pfzhan:


core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -774,10 +779,15 @@ public static Expression translateLiteral(
...
-          Expressions.constant(
+      Expression expression =
+          Expressions.new_(
+              BigDecimal.class, constant(
               requireNonNull(bd,
                   () -> "value for " + literal).toString()));
+      if (type.getScale() != bd.scale()) {
+        expression = adjustDecimalScale(type.getScale(), expression);
+      }
+      return expression;
```
zabetak, line 790: If we have a literal why do we need to adjust the scale? Does the `BigDecimal(String)` constructor pick a wrong scale for certain inputs? If that's the case can you please provide some examples with such inputs?

Which of the existing unit tests are affected by this change?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -625,7 +627,10 @@ private static Expression checkExpressionPadTruncate(
       default:
         return operand;
       }
-
+    case DECIMAL:
+      return sourceType.getScale() == targetType.getScale()
+          ? operand
+          : adjustDecimalScale(targetType.getScale(), operand);
```
zabetak, line 633: Most of the numeric conversions are happening inside `EnumUtils.convert` method so I am wondering if `checkExpressionPadTruncate` is the right place to apply the changes. Can you please elaborate on why the changes should be done here and not for instance inside `EnumUtils`?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -3315,6 +3315,13 @@ private void checkReduceNullableToNotNull(ReduceExpressionsRule<?> rule) {
     checkReduceNullableToNotNull(rule);
   }
 
+  @Test void testReduceCastDecimal() {
+    final String sql = "SELECT CAST( ( (2) / SQRT(3.0) ) AS NUMERIC(18, 0) ) * SQRT(3.0)";
+    sql(sql).withRule(CoreRules.PROJECT_REDUCE_EXPRESSIONS)
+        .withRelBuilderSimplify(true)
+        .check();
+  }
+
```
zabetak, line 3324: Usually we add tests in this class to demonstrate that certain rules will apply and have the desired result. I don't see any changes in `PROJECT_REDUCE_EXPRESSIONS` rule so this test seems slightly misplaced/redundant.

The problem seems to be around the CAST function so a more appropriate place would be the `operator.iq` file and/or `SqlOperatorTest`.

Also I am wondering if the expression `CAST( ( (2) / SQRT(3.0) ) AS NUMERIC(18, 0) ) * SQRT(3.0)` is minimal. Is there a smaller case that demonstrates the problem? If yes then we should use that one.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6421,7 +6421,11 @@ private static void checkIf(SqlOperatorFixture f) {
       f.checkType("atanh('abc')", "DOUBLE NOT NULL");
       f.checkScalarApprox("atanh(0.76159416)", "DOUBLE NOT NULL",
           isWithin(1d, 0.0001d));
+      f.checkScalarApprox("atanh(cast(-0.1 as decimal(19, 0)))", "DOUBLE NOT NULL",
```
zabetak, line 6424: Do we need `atanh` and `ceil` to hit the problem or it is sufficient to use `CAST`? If the problem is reproable simply with `CAST` then we should simply add test cases for that.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -838,6 +848,13 @@ public static Expression translateLiteral(
     return Expressions.constant(value2, javaClass);
   }
 
+  private static MethodCallExpression adjustDecimalScale(int scale,
+      Expression expression) {
+    assert expression.getType() == BigDecimal.class;
+    return Expressions.call(expression, BuiltInMethod.BIG_DECIMAL_SET_SCALE.method,
+        Expressions.constant(scale), Expressions.constant(RoundingMode.HALF_UP));
```
zabetak, line 855: Why did we pick the `RoundingMode.HALF_UP`? Is it something mandated by the SQL standard? Is it something commonly used by other DBMS? Do we need to document this behavior in the website?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
pfzhan: The method `EnumUtils#convert(Expression operand, Type fromType, Type toType)` indeed transfers the type, but for BigDecimal it just new a instance of the operand.Since existing methods do not pass target scale info into this method, there is no way to deal with scale in `EnumUtils#convert`. 
`Expression translateCast( RelDataType sourceType, RelDataType targetType, Expression operand,  boolean safe) {
    Expression convert = getConvertExpression(sourceType, targetType, operand);
    Expression convert2 = checkExpressionPadTruncate(convert, sourceType, targetType);
    Expression convert3 = expressionHandlingSafe(convert2, safe);
    return scaleIntervalToNumber(sourceType, targetType, convert3);
  }`
This method deals with type casting, the expression returned by `getConvertExpression` can be seen as not dealing with scale, and `checkExpressionPadTruncate` similarly deals with scale for char/timestamp etc., so it seems a good place to deal with the scale of BigDecimal. Besides, we know the scale information of targetType and sourceType at this point, so it's easy to compare and set the target scale.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
pfzhan: when invoke RexBuilder#canRemoveCastFromLiteral the call, for example `cast(-0.1 as decimal(19, 0))` returns true, then it will be translate to -0.1:BigDeciaml(19, 0) as a literal. 
image
However, the BigDecimal gives scale=1. So we should change the scale to the correct one.
If we run the `CalciteSqlOperatorTest.java` we will found this problem, but the agg.iq, misc.iq, and the expected values of `SqlOperatorTest.java` are wrong, so this issue is being ignored.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
pfzhan: I've been thinking about this point for quite a while, `RoundingMode.HALF_UP` might not be the best, maybe `RoundingMode.HALF_EVEN` would be better, but I'm not really sure, so I refer to the
`org.apache.calcite.util.NumberUtil#rescaleBigDecimal` to keep the behavior consistent with the original. If we plan to tweak the use of `RoundingMode`, I think we'll probably need to look at all the places where it is already being used and refer to more DBMSs to make the tradeoffs.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
pfzhan: Got it, It's not necessary, I will put this test to the SqlOperatorTest.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
pfzhan: Just compare with the line 6425.

pfzhan:
@zabetak Would you mind to help review this pr again?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
zabetak: The fact that we have a literal that is `-0.1:DECIMAL(19,0)` seems like a bug to me that we may need to address first.

These kind of literals indicate that there is a problem at the logical layer but this PR attacks the problem at the physical layer which is not very intuitive.

Moreover, we want to fix the root cause so that users who are not using at all Calcite's physical layer (Enumerable, Bindable, code generation, etc.) can benefit from the fix.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
zabetak: As long as we are consistent with other places in Calcite that's fine with me.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6422,6 +6422,10 @@ private static void checkIf(SqlOperatorFixture f) {
       f.checkScalarApprox("atanh(0.76159416)", "DOUBLE NOT NULL",
           isWithin(1d, 0.0001d));
       f.checkScalarApprox("atanh(cast(-0.1 as decimal))", "DOUBLE NOT NULL",
+          isWithin(-0.0d, 0.0001d));
```
zabetak, line 6425: The SqlOperatorTest has a lot of test cases with casts to decimals etc that are currently disabled due to `SqlOperatorTest#DECIMAL` field. Are we planning to fix those as part of this PR? It looks like we should.

Note that currently there is a bug (tracked under CALCITE-5923) and due to that `CalciteSqlOperatorTest` are not running as expected.

pfzhan:
retest this please

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
zabetak: Agreed. The name of the method (`checkExpressionPadTruncate`) is a bit misleading but indeed it is a good place for the conversion.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
zabetak: I think we can keep the original proposal and deal with `-0.1:DECIMAL(19,0)` in a follow-up.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
zabetak: @pfzhan This question is still open. How we should treat the test cases guarded by SqlOperatorTest#DECIMAL?

zabetak: If we enable the tests under `SqlOperatorTest#DECIMAL` do we need the new tests that were added in this class by this PR? Do they offer additional test coverage?

pfzhan: I run the test found that SqlOperatorTest does not compare with expected result, but the CalciteSqlOperatorTest class does. Some cases were add to explain slight changes, the test in .iq was implicit, but this is explicit.

I checked the pr of [CALCITE-5923](https://issues.apache.org/jira/browse/CALCITE-5923). It seems that handles date/timestamp cases rather than decimal type.

core/src/test/resources/sql/operator.iq
```diff
@@ -258,6 +258,19 @@ X1 INTEGER(10) NOT NULL
...
+!type
++----+
+| X1 |
++----+
+|  1 |
++----+
+(1 row)
+
+!ok
+
```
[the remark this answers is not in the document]
pfzhan: ok, this can be removed from the test case.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
zabetak: The `SqlOperatorTest` class contains many tests casting various types to decimals. These tests are all disabled since `SqlOperatorTest#DECIMAL` is `false`.

Since in this PR you are basically fixing CAST to DECIMAL type, I was thinking that it makes sense to enable those that are affected by this fix. If the existing tests are sufficient to test the changes in this PR then it is not really necessary to add new ones (using `atanh` etc.).

Basically, another way to put my question is the following.
* Can we set SqlOperatorTest#DECIMAL to true ?
* Can we remove some `if (DECIMAL)`, `if (!DECIMAL)` statements?
* Does this PR fix any of the above?

pfzhan: Got it, this will be considered in this pr.

zabetak (commented):
Overall the changes LGTM, however when I apply locally the patch for CALCITE-5923, which makes all the tests under SqlOperatorTest run as expected, I get some errors.

```
./gradlew cleanTest :core:test --tests CalciteSqlOperatorTest
```
@pfzhan Can you please check if those errors are related with the changes here?

pfzhan:
> Overall the changes LGTM, however when I apply locally the patch for [CALCITE-5923](https://issues.apache.org/jira/browse/CALCITE-5923), which makes all the tests under SqlOperatorTest run as expected, I get some errors.
> 
> ```
> ./gradlew cleanTest :core:test --tests CalciteSqlOperatorTest
> ```
> 
> @pfzhan Can you please check if those errors are related with the changes here?

I pulled the latest code locally and ran this command and it was OK.

BUILD SUCCESSFUL in 56s
74 actionable tasks: 19 executed, 3 from cache, 52 up-to-date
The remote build cache was disabled during the build due to errors.
(base) ➜  calcite git:(CALCITE-5860)

zabetak:
Due to the bug reported under CALCITE-5923 the tests that rely on `safeParameters` method are not running as expected. Basically, without the fix in CALCITE-5923, the tests will only pass from the validator; executable code is not generated so basically the actual results are not checked.

If you want to test your changes properly you will have to apply "temporarily" the proposed fix in CALCITE-5923 and then run `CalciteSqlOperatorTest`.

zabetak:
@pfzhan Now that CALCITE-5923 is committed you can revert https://github.com/apache/calcite/pull/3326/commits/c976ef4ff24966418553807b9e576a0d3c433cd7 and continue the work as usual. Ping me again once you address the test failures. Thanks!

PR #3327, opened 2023-07-19
[CALCITE-5865] ClassCastException with FLOOR and CEIL on conformances that are not builtin

gianm:
[CALCITE-5747](https://issues.apache.org/jira/browse/CALCITE-5747) introduced some BigQuery-specific logic for FLOOR and CEIL that is keyed off the conformance being `SqlConformanceEnum.BIG_QUERY`. However, it was implemented in such a way that implementations of `SqlConformance` that are not `SqlConformanceEnum` would throw `ClassCastException`.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
@@ -2701,11 +2701,10 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
 
   /** Returns the operator for {@code FLOOR} and {@code CEIL} with given floor flag
    * and library. */
-  public static SqlOperator floorCeil(boolean floor, SqlConformanceEnum conformance) {
-    switch (conformance) {
-    case BIG_QUERY:
+  public static SqlOperator floorCeil(boolean floor, SqlConformance conformance) {
+    if (SqlConformanceEnum.BIG_QUERY.equals(conformance)) {
       return floor ? SqlLibraryOperators.FLOOR_BIG_QUERY : SqlLibraryOperators.CEIL_BIG_QUERY;
-    default:
+    } else {
       return floor ? SqlStdOperatorTable.FLOOR : SqlStdOperatorTable.CEIL;
     }
   }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -2701,11 +2701,10 @@ public static SqlOperator like(boolean negated, boolean caseSensitive) {
 
   /** Returns the operator for {@code FLOOR} and {@code CEIL} with given floor flag
    * and library. */
-  public static SqlOperator floorCeil(boolean floor, SqlConformanceEnum conformance) {
-    switch (conformance) {
-    case BIG_QUERY:
+  public static SqlOperator floorCeil(boolean floor, SqlConformance conformance) {
+    if (SqlConformanceEnum.BIG_QUERY.equals(conformance)) {
```
tanclary, line 2705: Is there a test you can add for this?

tanclary (commented):
Thanks again for opening this fix

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
gianm: I thought a little about this, but I wasn't sure what kind of test to add, or where.

I could certainly add a test for this specific method verifying that a user-defined conformance works OK. But it doesn't seem likely to be a high-value test. It would be better to somehow write a test that verifies _everything_ works with user-defined conformances. But I didn't see a way to do that after a brief look.

I'm open to thoughts here. If you think it's valuable to write a test for this specific method, LMK and I will add one. Even better, if you have an idea about how to write a more comprehensive test, LMK.

rubenada:
LGTM.
About the test discussion: considering we are a bit on a rush in the middle of a release process, the fix seems simple and clear; and creating a specific unit test can be cumbersome (or even an overkill), I'd agree with merging the fix as it is, without test.

tanclary:
@rubenada Thanks for your input, I agree with that. LGTM

gianm:
Thanks all. I have committed the patch to `main`.

PR #3328, opened 2023-07-20
[CALCITE-5861] Optimization rules do not constant-fold expressions in window bounds

mihaibudiu:
With this change the PROJECT_REDUCE_EXPRESSIONS rule will also optimize constants that appear in window bound expressions. This makes it possible to compile programs that use (constant) expressions in window bounds.

However, I suspect that the rule PROJECT_TO_LOGICAL_PROJECT_AND_WINDOW is still buggy, since it will move window bound expressions into separate projections, as described in the JIRA case. Unfortunately I could not figure out how to fix that rule - the code is very generic and treats all expressions in the same way, regardless where they appear in the RexNode. Perhaps I should file another JIRA issue about this?

### DIFF
--- a/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
@@ -4206,6 +4206,22 @@ LogicalProject(SUM1=[SUM($7) OVER (PARTITION BY $7 ORDER BY $5)], SUM2=[SUM(+($7
 ]]>
     </Resource>
   </TestCase>
+  <TestCase name="testExpressionPreceding">
+    <Resource name="sql">
+      <![CDATA[select COUNT(*) over (
+ORDER BY empno
+ROWS BETWEEN 5 + 5 PRECEDING AND 1 PRECEDING) AS w_count from emp
+]]>
+    </Resource>
+    <Resource name="planBefore">
+      <![CDATA[
+LogicalProject($0=[$1])
+  LogicalWindow(window#0=[window(order by [0] rows between $1 PRECEDING and $2 PRECEDING aggs [COUNT()])])
+    LogicalProject(EMPNO=[$0])
+      LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+]]>
+    </Resource>
+  </TestCase>
   <TestCase name="testExpressionSimplification1">
     <Resource name="sql">
       <![CDATA[select * from emp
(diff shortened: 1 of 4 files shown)


asolimando:
> However, I suspect that the rule PROJECT_TO_LOGICAL_PROJECT_AND_WINDOW is still buggy, since it will move window bound expressions into separate projections, as described in the JIRA case. Unfortunately I could not figure out how to fix that rule - the code is very generic and treats all expressions in the same way, regardless where they appear in the RexNode. Perhaps I should file another JIRA issue about this?

I agree on separating the two problems and filing a separate Jira about moving the window bound expressions in a projection.

asolimando:
@mihaibudiu, if you file a different ticket for the projection part, I guess the test you added will look different and involve only `ReduceExpressionsRule`, right?

In addition, in general, we tend to phrase the Jira ticket titles a bit differently though, something like "ReduceExpressionsRule rules should constant-fold expressions in window bounds", WDYT?

mihaibudiu:
In the JIRA ticket discussion @julianhyde said this:

> The SQL spec doesn't 'require' that bounds are constant. It says that a conforming implementation shall allow bounds that are constant. It doesn't say that a conforming implementation shall not allow bounds that are not constant.
> If I recall correctly, Oracle allows expressions. It's a nice feature. We should continue to allow expressions, even if Postgres doesn't.

So I think based on this interpretation a second issue will not be necessary. I will rename the issue as you suggested.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -4203,6 +4203,22 @@ LogicalProject($0=[$3], $1=[$4])
...
+    <Resource name="sql">
+      <![CDATA[select COUNT(*) over (
+ORDER BY empno
+ROWS BETWEEN 5 + 5 PRECEDING AND 1 PRECEDING) AS w_count from emp
+]]>
+    </Resource>
+    <Resource name="planBefore">
+      <![CDATA[
+LogicalProject($0=[$1])
+  LogicalWindow(window#0=[window(order by [0] rows between $1 PRECEDING and $2 PRECEDING aggs [COUNT()])])
```
NobiGo, line 4219: It looks like we miss something in the plan. In general, we can extract the original SQL from the plan. But in this plan, we don't know the $1 and $2 values. We should create a new ISSUE to improve it? @mihaibudiu @asolimando

julianhyde: It's possible that `$1` and `$2` reference constants inside `LogicalWindow`, and these are not being printed in the plan. See field `Window.constants`.

NobiGo: Thanks for the tips. I create a new ISSUE [https://issues.apache.org/jira/browse/CALCITE-5929](https://issues.apache.org/jira/browse/CALCITE-5929) to improve it.

NobiGo (approved):
LGTM

asolimando (approved):
LGTM too

mihaibudiu:
Maybe someone can merge this?

PR #3329, opened 2023-07-20
[CALCITE-5864] Convert interval weeks and quarters into the correct millisecond total

jsternberg:
The week and quarter intervals were added but did not correctly convert
into the correct number of milliseconds when evaluated.

These now correctly translate a week to 7 days and a quarter to 3
months.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlIntervalQualifier.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlIntervalQualifier.java
@@ -143,6 +143,10 @@ public class SqlIntervalQualifier extends SqlNode {
           "SQL_TSI_QUARTER",
           "SQL_TSI_YEAR");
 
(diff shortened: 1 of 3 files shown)


jsternberg:
Figured out how to add some unit tests for this. I also have a comment on the original issue about why the numbers are intended to be these numbers: https://issues.apache.org/jira/browse/CALCITE-5864?focusedCommentId=17746626&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17746626

njriasan:
@jsternberg Wanted to comment an issue with this approach I found in case you still care about this. Inside `RexLiteral` the `intervalString` method uses the typeName information directly instead of the actual qualifier in `getTimeUnits()`. As a result, if you try convert from Rex to SQL, you will accidentally multiply the underlying value (by 7 for week and 3 for quarter).

PR #3334, opened 2023-07-23
[CALCITE-985] Validate MERGE

JiajunBernoulli:


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -2256,11 +2256,16 @@ protected void registerNamespace(
       @Nullable String alias,
       SqlValidatorNamespace ns,
       boolean forceNullable) {
-    namespaces.put(requireNonNull(ns.getNode(), () -> "ns.getNode() for " + ns), ns);
+    final SqlNode sqlNode = requireNonNull(ns.getNode(), () -> "ns.getNode() for " + ns);
```
julianhyde, line 2259: Consider removing the `sqlNode` variable. After you've called `requireNonNull(ns.getNode())` the validator will no longer complain.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -5269,6 +5274,20 @@ private static SqlNode getNthExpr(SqlNode query, int ordinal, int sourceCount) {
     SqlInsert insertCallAfterValidate = call.getInsertCall();
     if (insertCallAfterValidate != null) {
       validateInsert(insertCallAfterValidate);
+      // Validate NULL
```
julianhyde, line 5277: I didn't understand the 'Validate NULL' comment at first. 'Throw if select list contains NULL literal and target is NOT NULL' might be better.

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -3076,16 +3076,14 @@ void checkCorrelatedMapSubQuery(boolean expand) {
...
   @Test void testMerge() {
-    final String sql = "merge into emp as target\n"
-        + "using (select * from emp where deptno = 30) as source\n"
-        + "on target.empno = source.empno\n"
-        + "when matched then\n"
-        + "  update set sal = sal + source.sal\n"
-        + "when not matched then\n"
-        + "  insert (empno, deptno, sal)\n"
-        + "  values (source.empno, source.deptno, source.sal)";
+    final String sql = "merge into empnullables e "
```
julianhyde, line 3080: Can you end each line of SQL with "\n" (and no space). It makes the SQL easier to read inside `SqlToRelConverterTest.xml`.

julianhyde (changes requested):
I requested some cosmetic changes. With those, +1 to merge without further review.

Thank you for doing this -- the case has been open for many years!

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
JiajunBernoulli: Good idea.

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
JiajunBernoulli: Done

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
JiajunBernoulli: Removed `sqlNode`.

JiajunBernoulli:
Closed #3283

PR #3336, opened 2023-07-24
[CALCITE-5869] LEAST_RESTRICTIVE does not use inner type of MEASURE f…

tanclary:
…or comparisons

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlOperator.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlOperator.java
@@ -22,6 +22,7 @@ import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.type.MeasureSqlType;
 import org.apache.calcite.sql.type.SqlOperandTypeChecker;
 import org.apache.calcite.sql.type.SqlOperandTypeInference;
 import org.apache.calcite.sql.type.SqlReturnTypeInference;
@@ -535,6 +536,11 @@ public abstract class SqlOperator {
         throw new IllegalArgumentException("Cannot infer return type for "
             + opBinding.getOperator() + "; operand types: "
             + opBinding.collectOperandTypes());
+      }
+
+      // MEASURE wrapper should be removed, e.g. MEASURE<DOUBLE> should just be DOUBLE
+      if (returnType instanceof MeasureSqlType) {
+        returnType = returnType.getMeasureElementType();
       }
 
       if (operandTypeInference != null
--- a/core/src/main/java/org/apache/calcite/sql/SqlOperator.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlOperator.java
@@ -22,7 +22,6 @@ import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParserPos;
-import org.apache.calcite.sql.type.MeasureSqlType;
 import org.apache.calcite.sql.type.SqlOperandTypeChecker;
 import org.apache.calcite.sql.type.SqlOperandTypeInference;
 import org.apache.calcite.sql.type.SqlReturnTypeInference;
@@ -49,6 +48,7 @@ import java.util.Objects;
 import java.util.function.Supplier;
 
 import static org.apache.calcite.linq4j.Nullness.castNonNull;
+import static org.apache.calcite.sql.type.SqlTypeUtil.isMeasure;
 import static org.apache.calcite.util.Static.RESOURCE;
 
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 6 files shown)


core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
```diff
@@ -153,6 +153,15 @@ private static final class CustomTypeSystem extends RelDataTypeSystemImpl {
     assertEquals(SqlTypeName.DOUBLE, dataType.getSqlTypeName());
   }
 
+  @Test void testLeastRestrictiveUsesMeasureElementType() {
+    RelDataType innerType = TYPE_FACTORY.createSqlType(SqlTypeName.DOUBLE);
+    RelDataType operand1 = TYPE_FACTORY.createMeasureType(innerType);
+    RelDataType operand2 = TYPE_FACTORY.createSqlType(SqlTypeName.INTEGER);
+    RelDataType dataType = SqlLibraryOperators.IFNULL
+        .inferReturnType(TYPE_FACTORY, Lists.newArrayList(operand1, operand2));
+    assertEquals(dataType, operand1);
```
tjbanghart, line 162: `assertEquals(dataType, operand1)` ensures we have a `MEASURE<DOUBLE>`. Is that desired? I think yes, but want to consider if the return type should be the `innerType` of the measure?

tanclary: I think it makes sense for the return type to be a `MEASURE`, for instance if you had `IFNULL(AVG(x), 0)` and `AVG(x)` is not null, the return type should be a `MEASURE<DOUBLE>`. What do you think?

julianhyde: Please use `assertThat` rather than `assertEquals`.

I think the measure wrapper should be removed. The type should be `DOUBLE`.

Add a comment explaining what the test does, e.g. "IFNULL(MEASURE<DOUBLE>, INTEGER) should have type "DOUBLE". Much easier for future maintainers to review. Also add an `<a>` link.

core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
```diff
@@ -153,6 +154,15 @@ private static final class CustomTypeSystem extends RelDataTypeSystemImpl {
     assertEquals(SqlTypeName.DOUBLE, dataType.getSqlTypeName());
   }
 
+  @Test void testLeastRestrictiveUsesMeasureElementType() {
+    RelDataType innerType = TYPE_FACTORY.createSqlType(SqlTypeName.DOUBLE);
```
julianhyde, line 158: Not your fault, but TYPE_FACTORY, TYPE_FIXTURE, CUSTOM_FACTORY  should not be static. Static fields in tests cause memory leaks (and sometimes race conditions).

Can you rework (in a separate pre-factoring commit in the same PR) so that each test starts with `final SqlTypeFixture f = new SqlTypeFixture();`.

tanclary: Done (also did not know that so thanks for the knowledge).

core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
tanclary: Done

tanclary:
@tjbanghart @julianhyde Do either of you have additional comments? Thanks for your suggestions so far!

core/src/main/java/org/apache/calcite/sql/SqlOperator.java
```diff
@@ -537,6 +538,11 @@ public RelDataType inferReturnType(
             + opBinding.collectOperandTypes());
       }
 
+      // MEASURE wrapper should be removed, e.g. MEASURE<DOUBLE> should just be DOUBLE
+      if (isMeasure(returnType) && returnType.getMeasureElementType() != null) {
+        returnType = Objects.requireNonNull(returnType.getMeasureElementType());
```
barrkel, line 543: I am leaving this comment to highlight relevant code to ask @tanclary about it.

This strips off the `MEASURE<>`-ness from `SELECT`ed columns when the `AS` operator is used - aliases are modeled in the AST as an operator.

This means that if

```
SELECT m FROM table
```

has `m` of type `MEASURE<INTEGER>`, then

```
SELECT m AS m FROM table
```

will have `m` of type `INTEGER`. This then causes breakage when the AST validator has a different opinion about the type of the row than the conversion into rel nodes.

I want to understand more about the requirement to strip of `MEASURE`. I think it shouldn't be done for `AS`, but if I can understand why it does need to be done, then there might be more cases where it shouldn't be done.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
```diff
@@ -896,6 +900,12 @@ public static boolean canCastFrom(
     if (toType.equals(fromType)) {
       return true;
     }
+    // If fromType is a measure, you should compare the inner type otherwise it will
+    // always return TRUE because the SqlTypeFamily of MEASURE is ANY
+    if (isMeasure(fromType)) {
```
barrkel, line 905: @tanclary If I understand this code correctly, this is testing for implicit (I think) conversions presumably as part of type checking on behalf of the caller. But it's not wholly clear to me that measures ought to be implicitly convertible to anything without applying `AGGREGATE`. Is it meaningful to compute anything with measures in themselves, outside an aggregation context? At the very least, there's an assumption of naked measures and implicit aggregation, right?

core/src/main/java/org/apache/calcite/sql/SqlOperator.java
tanclary: This is originally what I thought too. See the above comment chain between me, Julian, and TJ.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
tanclary: Maybe you're right. This code allows the inner type to be compared to the toType.

PR #3338, opened 2023-07-25
[CALCITE-5873] Add REGEXP_CONTAINS function (enabled in BigQuery library)

Anthrino:


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -394,6 +394,15 @@ private static int makeRegexpFlags(@Nullable String stringFlags) {
     return flags;
   }
 
+  /** SQL {@code REGEXP_CONTAINS(value, regexp)} function. */
+  public static boolean regexpContains(String value, String regex) {
+    try {
+      return Pattern.compile(regex, Pattern.CASE_INSENSITIVE).matcher(value).matches();
+    } catch (Exception exception) {
```
tanclary, line 401: nit: change this to 'e' to match other caught exception names

tanclary: This is inconsistent though because there is 'e', 'ex', etc

tanclary (commented):
Thanks for the PR! Left a few comments.

Would you also mind updating the commit message/JIRA ticket to follow the same format as other functions? e.g. Add REGEXP_CONTAINS function (enabled for BigQuery)'

Also you can just put the ticket # of the subticket you don't need to include the parent

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4438,6 +4438,31 @@ private static void checkIf(SqlOperatorFixture f) {
...
+
+    f.checkQuery("select regexp_contains('abc def ghi', 'abc')");
+    f.checkQuery("select regexp_contains('foo@bar.com', '@[a-zA-Z0-9-]+\\\\.[a-zA-Z0-9-.]+')");
+    f.checkQuery("select regexp_contains('55as56664as422', '\\d{10}')");
+
+    f.checkFails("regexp_contains('abc def ghi', '(abc')",
+        "Invalid regex input for REGEXP_CONTAINS: '(abc'", true);
+    f.checkFails("regexp_contains('abc def ghi', '{3}')",
+        "Invalid regex input for REGEXP_CONTAINS: '{3}'", true);
+    f.checkFails("regexp_contains('abc def ghi', '[z-a]')",
```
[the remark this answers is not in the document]
Anthrino: Quick clarification @tanclary , f.checkNull() verifies if we get back a SQL NULL value from the expression right? What should be done in case we are not expecting a NULL return, as in this function always returns a boolean value if the required arguments are passed or an exception for invalid regex.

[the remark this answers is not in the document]
tanclary: That's right. It's still helpful to add .checkNull() tests to ensure that if either (or both) arguments is null that, that null is returned. This is useful because not all functions follow that behavior. In this case however, based on a couple of quick queries I ran against BigQuery, if either or both arguments is null the function should return null. Hope this helps.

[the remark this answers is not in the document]
tanclary: In other cases where you are expecting a non-non value you should use .checkBoolean (at least for this function given that the return type is boolean).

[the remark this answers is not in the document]
Anthrino: Got it, thanks Tanner! The documentation didn't mention NULL returns so I assumed it returns bools, will run some test queries for additional NULL scenarios.

tanclary (commented):
Left a couple more (small) comments!

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -401,6 +402,16 @@ private static int makeRegexpFlags(@Nullable String stringFlags) {
     return flags;
   }
 
+  /** SQL {@code REGEXP_CONTAINS(value, regexp)} function. */
```
julianhyde, line 405: Can you briefly document what happens if the regex is invalid.

julianhyde (changes requested):
I've made a few cosmetic suggestions. Hopefully they are easy to fix.

Also, you need to add to `SqlOperatorTest`. There are currently no tests for scenarios where REGEXP_CONTAINS returns null (null first or second argument).

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -401,6 +402,16 @@ private static int makeRegexpFlags(@Nullable String stringFlags) {
...
+  public static boolean regexpContains(String value, String regex) {
+    try {
+      Pattern regexp = Pattern.compile(regex, Pattern.CASE_INSENSITIVE);
```
mihaibudiu, line 408: Does this pattern handle Java regular expressions or BigQuery regular expressions?

Anthrino: Hi @mihaibudiu, this Pattern object is an instance of the java.util.regex library, it has almost complete overlap with BQ regex except for a few negative scenarios as highlighted on this ticket: [CALCITE-5910](https://issues.apache.org/jira/browse/CALCITE-5910)

site/_docs/reference.md
```diff
@@ -2776,6 +2776,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | PARSE_TIMESTAMP(format, string[, timeZone])    | Uses format specified by *format* to convert *string* representation of timestamp to a TIMESTAMP WITH LOCAL TIME ZONE value in *timeZone*
 | h s | PARSE_URL(urlString, partToExtract [, keyToExtract] ) | Returns the specified *partToExtract* from the *urlString*. Valid values for *partToExtract* include HOST, PATH, QUERY, REF, PROTOCOL, AUTHORITY, FILE, and USERINFO. *keyToExtract* specifies which query to extract
 | b | POW(numeric1, numeric2)                        | Returns *numeric1* raised to the power *numeric2*
+| b | REGEXP_CONTAINS(value, regexp)                 | Returns TRUE if *value* is a partial match for the regular expression, *regexp*
```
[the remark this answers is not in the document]
Anthrino: Sure @julianhyde, that sounds more clear I just thought it would be best to align it to BQ documentation as they use *value* everywhere.

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
```diff
@@ -234,6 +235,34 @@ static <E> List<E> list() {
     assertThat(posixRegex("abcq", "[[:xdigit:]]", false), is(true));
   }
 
+  @Test void testRegexpContains() {
+    try {
+      regexpContains("abc def ghi", "(abc");
+      fail("'regexp_contains' on an invalid regex input '(abc' is not possible");
+    } catch (CalciteException e) {
+      assertThat(e.getMessage(),
+          is("Invalid regex input for REGEXP_CONTAINS: 'Unclosed group near index 4\n(abc'"));
```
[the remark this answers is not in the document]
Anthrino: > Break strings after each `\n`. It makes them easier to read.

Thanks for the pointer @julianhyde, the \n is included only as it is part of the exception message returned, I can break the string for legibility but this was easier to match with the returned string.



> I'm surprised that this test works. Does `regexpContains` throw a `CalciteException`? Looks like it throws a `RuntimeException`.

Yes we were returning a `CalciteException` previously, saw your comment on the other file will modify it to a `RuntimeException`.

site/_docs/reference.md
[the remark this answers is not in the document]
julianhyde: I understand your dilemma. I think it's better to make the Calcite descriptions uniform, even if they come from many different DBs.

In `big-query.iq`, feel free to use the doc copy-pasted from BigQuery.

babel/src/test/resources/sql/big-query.iq
```diff
@@ -3520,5 +3520,71 @@ FROM items;
 
 !ok
 
+#####################################################################
```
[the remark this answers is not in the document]
Anthrino: Thanks for pointing out Julian, moved the tests to appropriate location.

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
```diff
@@ -1005,6 +1005,9 @@ ExInstWithCause<CalciteException> failedToAccessField(
   @BaseMessage("Invalid input for JSON_STORAGE_SIZE: ''{0}''")
   ExInst<CalciteException> invalidInputForJsonStorageSize(String value);
 
+  @BaseMessage("Invalid regex input for REGEXP_CONTAINS: ''{0}''")
```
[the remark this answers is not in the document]
Anthrino: Updated error statement as suggested.

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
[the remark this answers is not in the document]
Anthrino: Converted into Runtime Exceptions and corrected \n formatting.

site/_docs/reference.md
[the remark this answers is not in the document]
Anthrino: Updated ref docs as per suggestions.

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
```diff
@@ -1005,6 +1005,9 @@ ExInstWithCause<CalciteException> failedToAccessField(
...
+  ExInst<CalciteException> invalidInputForRegexpContains(String value);
```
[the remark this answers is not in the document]
Anthrino: Changed into RuntimeException, thanks for the correction.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
Anthrino: Added a statement, please check if this is what you had in mind.

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
```diff
@@ -234,6 +235,44 @@ static <E> List<E> list() {
...
   }
 
+  @Test void testRegexpContains() {
+    try {
+      regexpContains("abc def ghi", "(abc");
+      fail("'regexp_contains' on an invalid regex input '(abc' is not possible");
+    } catch (RuntimeException e) {
+      assertThat(e.getMessage(),
+          is("Invalid regular expression for REGEXP_CONTAINS: 'Unclosed group near index 4\n"
+              +
```
julianhyde, line 245: no newline after `+`

site/_docs/reference.md
```diff
@@ -2776,6 +2776,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | PARSE_TIMESTAMP(format, string[, timeZone])    | Uses format specified by *format* to convert *string* representation of timestamp to a TIMESTAMP WITH LOCAL TIME ZONE value in *timeZone*
 | h s | PARSE_URL(urlString, partToExtract [, keyToExtract] ) | Returns the specified *partToExtract* from the *urlString*. Valid values for *partToExtract* include HOST, PATH, QUERY, REF, PROTOCOL, AUTHORITY, FILE, and USERINFO. *keyToExtract* specifies which query to extract
 | b | POW(numeric1, numeric2)                        | Returns *numeric1* raised to the power *numeric2*
+| b | REGEXP_CONTAINS(string, regexp)                | Returns whether *string* is a partial match for the regular expression, *regexp*
```
julianhyde, line 2779: remove 'the regular expression, '. It's clear from the variable name 'regexp'

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -475,6 +475,14 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
   @LibraryOperator(libraries = {MYSQL, ORACLE})
   public static final SqlFunction REGEXP_REPLACE = new SqlRegexpReplaceFunction();
 
+  /** The "REGEXP_CONTAINS(value, regexp)" function.
+   * Returns TRUE if value is a partial match for the regular expression, regexp. */
+  @LibraryOperator(libraries = {BIG_QUERY})
```
wnob, line 480: BigQuery [uses re2][1] for their regex engine. Google has an [implementation for Java][2]. We should use that here, both as a matter of syntax conformance (since this is a BigQuery-specific function) and for the security reasons outlined in the RE2J readme. This could be a follow-up, but I don't think we should close the bug until that's done.

[1]: https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#regexp_contains
[2]: https://github.com/google/re2j

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -345,6 +346,17 @@ public static String sha512(ByteString string)  {
     return DigestUtils.sha512Hex(string.getBytes());
   }
 
+  /** SQL {@code REGEXP_CONTAINS(value, regexp)} function.
+   * Throws a runtime exception for invalid regular expressions.*/
+  public static boolean regexpContains(String value, String regex) {
+    try {
+      Pattern regexp = Pattern.compile(regex);
```
tanclary, line 353: If you're going to stick with Java library, I think it would help to add a comment here or elsewhere linking to the case where that decision was made for future users/devs

Anthrino: Thanks for the pointer Tanner, added a note with issue link to the method description.

julianhyde: I wouldn't link to jira from the source code. Comments in the source code should be self-contained.

(I know we add jira links in bugs. But that's because we *removed* a bug, and we don't want to leave a trail of mistakes in the source code.)

This is a decision affects the specification of the function, not just its implementation. So I think the right place for the comment is in `big-query.iq`. You could have another comment in the body of `regexpContains` if you think maintainers might be confused why we're using Java regex.

tanclary: I think it makes sense to have one in both place because If someone were to write a query using regex that was valid in the re2 library and not Java, if it used Calcite's implementation then the error would get thrown from `regexpContains` and the user/developer would be able to reference that comment for clarification on the library decision.

Anthrino: Sure I think it makes sense to add a comment in the method body too, just wanted to point out anything that works in RE2 should be processed identically by java regex as well, which should cover 99.9% of queries. Only some invalid regex errors caught by RE2 aren't caught here so Calcite might produce an incorrect output for a few corner cases, which are technically incorrect regex inputs so they shouldn't be used in a query anyway.

PR #3339, opened 2023-07-27
[CALCITE-5875] Remove unnecessary NULL checks in Redis adapter

chucheng92:

code cleanup, remove some unnecessary null checks for redis adapter, because the static variable must be initialized successfully unless there is an exception.

# Brief change log
Remove some unnecessary null checks in redis adapter.

# Verifying this change

It's minor fix can be covered by current total cases.

### DIFF
--- a/redis/src/test/java/org/apache/calcite/adapter/redis/RedisCaseBase.java
+++ b/redis/src/test/java/org/apache/calcite/adapter/redis/RedisCaseBase.java
@@ -113,7 +113,7 @@ public abstract class RedisCaseBase {
 
   @AfterAll
   public static void stopRedisContainer() {
-    if (REDIS_CONTAINER != null && REDIS_CONTAINER.isRunning()) {
+    if (REDIS_CONTAINER.isRunning()) {
       REDIS_CONTAINER.stop();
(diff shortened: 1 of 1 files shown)


redis/src/test/java/org/apache/calcite/adapter/redis/RedisDataCaseBase.java
```diff
@@ -99,8 +99,6 @@ public void makeData() {
 
   @AfterEach
   public void shutDown() {
-    if (null != pool) {
```
NobiGo, line 102: I don't think so. When we init the pool throw exception, if, remove the condition, the code will throw NPE.

NobiGo (changes requested):

redis/src/test/java/org/apache/calcite/adapter/redis/RedisDataCaseBase.java
chucheng92: @NobiGo Thanks for reviewing, sorry for my mistake, @AfterEach will still execute if @BeforeEach fails. I have fixed it.

PR #3341, opened 2023-07-28
[CALCITE-5877] AssertionError during MOD operation if result scale is greater than maximum numeric scale

mihaibudiu:


rubenada:
LGTM.
The fix looks reasonable, considering how the precision is handled just a few lines below, and it is consistent with the description on the javadoc of the method.

JiajunBernoulli:
LGTM

rubenada:
@mihaibudiu  before merging the fix, could you please amend the commit message (and the Jira ticket's title) to make it more concise? No need to specify "Calcite crashes...", just describe the issue, I'd propose something along the lines: "AssertionError during MOD operation if result scale is greater than maximum numeric scale". Wdyt?

mihaibudiu:
> @mihaibudiu before merging the fix, could you please amend the commit message (and the Jira ticket's title) 

Done

mihaibudiu:
My mistake, should have done this right away. Hopefully now it's fine.

rubenada:
Thanks @mihaibudiu , just one final, minor detail: the comment of the new unit test has the form of comment lines, could you please change it into javadoc with a proper link? (just to keep consistency with other similar comments in the same test class).
I can see that the comment of the next test is also using comment lines instead of javadoc (probably this is where the confusion comes from), but this seems an anomaly, and should also be changed to keep consistency within the class, so I'd suggest to take advantage of this PR and fix that comment too:
```
  /** Test case for
   * <a href="https://issues.apache.org/jira/browse/CALCITE-5877">[CALCITE-5877]
   * AssertionError during MOD operation if result scale is greater than maximum numeric scale</a>. */
  @Test void testNumericScaleMod() {
    final String sql = "SELECT MOD(CAST(2 AS DECIMAL(39, 20)), 2)";
    final String expected = "SELECT MOD(2, 2)\nFROM (VALUES (0)) AS \"t\" (\"ZERO\")";
    sql(sql).withPostgresqlModifiedDecimalTypeSystem()
        .ok(expected);
  }

  /** Test case for
   * <a href="https://issues.apache.org/jira/browse/CALCITE-5651">[CALCITE-5651]
   * Inferred scale for decimal should not exceed maximum allowed scale</a>. */
  @Test void testNumericScale() {
  ...
  ```

PR #3344, opened 2023-07-28
[CALCITE-5879] Calcite crashes with AssertionFailure during evaluation of constant SPLIT expression

mihaibudiu:
@tanclary


rubenada:
Thanks for the patch, @mihaibudiu. The fix LGTM.
I assume the same error would happen with `select split(NULL)`, if that is the case, could you please add that test too in `RelOptRulesTest.java`?
Also, could you please adjust the JIRA title (and hence the PR title, commit message, and new unit tests javadoc) to describe a bit more clearly the issue, I'd suggest something along the lines: `[CALCITE-5879] AssertionError during constant reduction of SPLIT expression that returns NULL`

rubenada:
Thanks @mihaibudiu . IMO the patch looks in a good shape.
@tanclary , if I am not mistaken, you were the one that introduced the SPLIT function for BigQuery, would you like to take a look at this fix?

tanclary (commented):
Thanks for filing and opening a fix. The only thing I'm curious about like I noted on your other PR is why this test belongs in RelOptRules?

mihaibudiu:
I actually wrote a long email on the dev mailing list about this problem. That's the interesting thing: you actually *have* added tests for these cases for SPLIT in SqlOperatorTest, but the tests didn't hit these bugs. That's because that compilation path take in that line of tests does not apply the reduction rules which cause the incorrect expansions. Every test should be evaluate in two ways: compile time evaluation, and run-time evaluation, and the two results should be identical. Currently SqlOperatorTest only tests the "run-time" evaluation.

I think there should be a fixture to do this kind of testing; it is pointless to write each test twice. I bet it will catch quite a few bugs, I have seen this happen with many other functions.

tanclary:
Ohhh I understand much better now. Thanks for the explanation.

rubenada:
Just to confirm, @tanclary do you agree with the patch? Do you have any other remark or I just can move on with the merge?

PR #3345, opened 2023-08-01
[CALCITE-5881] Support to get foreign keys metadata in RelMetadataQuery

JingDas:
We can get constraints by RelOptTable#getReferentialConstraints method, but maybe can't get appropriate constraints at top relNode.
Foreign keys metadata is very useful in many optimize scenes. Such as it can be used in join eliminate when join type is inner join and some other star schema query optimize.
So support to get foreign keys metadata in RelMetadataQuery, it support to get composite foreign keys and  get the corresponding unique key on the arbitrary relNode if you want.
Tests are added to `RelMetadataTest`, the test code takes up a large part of the PR.

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/RelOptForeignKey.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptForeignKey.java
@@ -1,0 +1,411 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.plan;
+
+import org.apache.calcite.linq4j.Linq4j;
+import org.apache.calcite.rel.metadata.RelMetadataQuery;
+import org.apache.calcite.rex.InferredConstraintKey;
+import org.apache.calcite.util.ImmutableBitSet;
+import org.apache.calcite.util.Pair;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * ForeignKey represents the foreign and unique key constraint relationship
+ * on the current {@link org.apache.calcite.rel.RelNode}.
+ *
+ * <p><b>constraints</b> field {@link #constraints} are
+ * constraints that foreign key and unique key relationships in bottom-up derivation.
+ *
+ * <p><b>foreignColumns</b> field {@link #foreignColumns} indicates the
+ * position of the foreign key on the current {@link org.apache.calcite.rel.RelNode}
+ * if not or not be confirmed, it is an empty set.
+ *
+ * <p><b>uniqueColumns</b> field {@link #uniqueColumns} indicates the position of
+ * the unique key on the current {@link org.apache.calcite.rel.RelNode},
+ * if not or not be confirmed, it is an empty set, the position in uniqueColumns
+ * corresponds to foreignColumns.
+ *
+ * <p>The element positions in {@code uniqueColumns} and {@code foreignColumns}
+ * correspond to each other. The order of elements in {@code uniqueColumns}
+ * and {@code foreignColumns} is consistent with the order of constraints in
+ * the constraints list.
+ *
+ * <p>For instance,
+ * <blockquote>
+ * <pre>select e.deptno, e.ename, d.deptno
+ * from emp as e
+ * inner join dept as d
+ * on e.deptno = d.deptno</pre></blockquote>
+ *
+ * <p>the foreign key is the DEPTNO column of CATALOG.SALES.EMP table,
+ * reference the DEPTNO unique column of CATALOG.SALES.DEPT table.
+ *
+ * <p>Invoke the {@link RelMetadataQuery#getConfirmedForeignKeys} method which
+ * input param is the top {@link org.apache.calcite.rel.core.Project},
+ * the following results can be obtained.
+ *
+ * <p>{@code constraints} is
+ * [{left: [CATALOG, SALES, EMP].$7.#true, right: [CATALOG, SALES, DEPT].$0.#true}]
+ * {@code foreignColumns} is {0}
+ * {@code uniqueColumns} is {2}
+ *
+ * <p>For instance,
+ * <blockquote>
+ * <pre>select name, deptno
+ * from dept</pre></blockquote>
+ *
+ * <p>{@code constraints} is
+ * [{left: null.null.#false, right: [CATALOG, SALES, DEPT].$0.#false}]
+ * {@code foreignColumns} is {}
+ * {@code uniqueColumns} is {1}
+ *
+ * <p>For instance,
+ * <blockquote>
+ * <pre>select ename, sal, deptno
+ * from emp</pre></blockquote>
+ *
+ * <p>{@code constraints} is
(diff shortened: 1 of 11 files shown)


LakeShen:
Hi @JingDas ,your pr's workflow has some abnormal information,please check it.
You could run this cmd in your local idea:
```sql
./gradlew check
```

core/src/main/java/org/apache/calcite/plan/RelOptForeignKey.java
```diff
@@ -0,0 +1,411 @@
...
+ * @see org.apache.calcite.plan.RelOptForeignKey
+ */
+public class RelOptForeignKey {
+
+  /** Foreign key and unique key relationships in bottom-up derivation. */
+  private final List<Pair<InferredConstraintKey, InferredConstraintKey>> constraints;
+  /** Position of the foreign key on the current {@link org.apache.calcite.rel.RelNode}. */
+  private final ImmutableBitSet foreignColumns;
+  /** Position of the unique key on the current {@link org.apache.calcite.rel.RelNode}. */
+  private final ImmutableBitSet uniqueColumns;
```
LakeShen, line 110: Why is there information about a unique key in the RelOptForeignKey class?

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
```diff
@@ -105,6 +106,43 @@ interface Handler extends MetadataHandler<UniqueKeys> {
...
+     * Foreign keys are represented as an {@link org.apache.calcite.util.ImmutableBitSet},
+     * where each bit position represents a 0-based output column ordinal.
+     *
+     * @param ignoreNulls if true, allow containing null values when determining
+     *                     whether the keys are foreign keys
+     *
+     * @return bit set of foreign keys(contains intermediate inferred foreign keys),
+     * or empty if not enough information is available to make that determination
+     */
+    Set<RelOptForeignKey> getForeignKeys(boolean ignoreNulls);
```
LakeShen, line 128: Is it @Nullable?

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
```diff
@@ -105,6 +106,43 @@ interface Handler extends MetadataHandler<UniqueKeys> {
...
+
+    /**
+     * Handler API.
+     */
+    @FunctionalInterface
+    interface Handler extends MetadataHandler<ForeignKeys> {
+      Set<RelOptForeignKey> getForeignKeys(
+          RelNode rel,
+          RelMetadataQuery mq,
```
LakeShen, line 137: Is it @Nullable,like the method:
```sql
@Nullable Set<ImmutableBitSet> getUniqueKeys(RelNode r, RelMetadataQuery mq,
          boolean ignoreNulls);
```

core/src/main/java/org/apache/calcite/plan/RelOptForeignKey.java
LakeShen: I think a foreign key is a reference to the Primary key of another table,make uniqueColumns to referredPrimaryKeys.

Make the name reflect its true meaning

WDYT?

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
LakeShen, line 128: Is it @Nullable ? Like this method:
```sql
@Nullable Set<ImmutableBitSet> getUniqueKeys(boolean ignoreNulls);

```

core/src/main/java/org/apache/calcite/plan/RelOptForeignKey.java
```diff
@@ -0,0 +1,411 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
```
LakeShen, line 40: Good code comments are very important. Can you make your comments concise and clear so that others can understand them easily？

core/src/main/java/org/apache/calcite/plan/RelOptForeignKey.java
```diff
@@ -0,0 +1,411 @@
...
+   *   },
+   *   {
+   *     foreignColumns: {6, 7}
+   *     uniqueColumns: {8}
+   *   }
+   * ]
+   *
+   * @param foreignMapping foreignKey mapping relationship
+   * @param uniqueMapping uniqueKey mapping relationship
+   * @return mapped relOptForeignKey set
```
LakeShen, line 197: What is it meaning?Give it a better description.

LakeShen: Any other instructions about @return should have a good explanation

testkit/src/main/java/org/apache/calcite/test/RelMetadataFixture.java
```diff
@@ -264,6 +265,18 @@ public RelMetadataFixture assertPercentageOriginalRows(Matcher<Double> matcher)
     return this;
   }
 
+  public RelMetadataFixture assertForeignKeys(Matcher<Set<RelOptForeignKey>> matcher,
+      Matcher<Set<RelOptForeignKey>> confirmedMatcher,
+      boolean ignoreNulls) {
+    RelNode rel = toRel();
+    RelMetadataQuery mq = rel.getCluster().getMetadataQuery();
+    Set<RelOptForeignKey> foreignKeys = mq.getForeignKeys(rel, ignoreNulls);
```
LakeShen, line 273: You could write more test cases, just like you have getForeignKeys method in RelMdForeignKeys for different relNodes

core/src/main/java/org/apache/calcite/plan/RelOptForeignKey.java
```diff
@@ -0,0 +1,411 @@
...
+
+  public ImmutableBitSet getUniqueColumns() {
+    return uniqueColumns;
+  }
+
+  public List<Pair<InferredConstraintKey, InferredConstraintKey>> getConstraints() {
+    return constraints;
+  }
+
+  /** Returns the left side of a list of constraints. */
```
LakeShen, line 140: What is meaning of left side,I think It should be described with clearer comments

core/src/main/java/org/apache/calcite/plan/RelOptForeignKey.java
```diff
@@ -0,0 +1,411 @@
...
+      List<Pair<InferredConstraintKey, InferredConstraintKey>> constraints) {
+    if (constraints.isEmpty()) {
+      return new ArrayList<>();
+    }
+    return constraints.stream()
+        .map(Pair::getKey)
+        .collect(Collectors.toList());
+  }
+
+  /** Returns the right side of a list of constraints. */
```
LakeShen, line 151: same as above

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -971,6 +989,1025 @@ final RelMetadataFixture sql(String sql) {
...
+                    InferredConstraintKey.of(DEPT_QNAME, 0, false))),
+            bitSetOf(1), ImmutableBitSet.of()));
+    Set<RelOptForeignKey> confirmedForeignKeys =
+        Sets.newHashSet(
+            RelOptForeignKey.of(
+            Lists.newArrayList(
+                Pair.of(
+                    InferredConstraintKey.of(EMP_QNAME, 7, true),
+                    InferredConstraintKey.of(DEPT_QNAME, 0, true))),
+            bitSetOf(1), bitSetOf(3)));
```
LakeShen, line 1364: These two variables are used in almost every single test I see. They have too much duplicate code. You can define two methods that take different parameter types and then construct those two variables. This reduces code duplication.

LakeShen:
Another point,Do you need add a class GeneratedMetadata_ForeignXXX like GeneratedMetadata_CollationHandler.java?
image

core/src/main/java/org/apache/calcite/plan/RelOptForeignKey.java
JingDas: Add the referred prefix is more accessible, I will fix it.

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
JingDas: It is reasonable to be consistent in behavior, get it.

LakeShen:
Maybe we could close this PR,because we would add the functional dependency,the jira is https://issues.apache.org/jira/browse/CALCITE-5913

LakeShen:
IMO,a foreign key is one of the table constraints, the current in Calcite, we could use Statistic# getReferentialConstraints method to get the table's foreign key.

So in an SQL plan(RelNode Tree),for example,if a RelNode is Project and its child is a TableScan,we want to use the foreign key information of TableScan. Instead of getting the foreign key information of Project, we should use functional dependencies.

In other words, only tables have foreign keys.The above is my understanding,If I understand this wrong,pls correct me,thanks.

PR #3348, opened 2023-08-04
[CALCITE-5840] Incorrect SOUNDEX function semantics in MySQL library

taoran92:

https://issues.apache.org/jira/browse/CALCITE-5840

The SqlFunctions#soundexMySQL rewrite the mysql's c++ soundex implementation. pls see:
https://github.com/mysql/mysql-server/blob/ea1efa9822d81044b726aab20c857d5e1b7e046a/sql/item_strfunc.cc#L2054

# Brief change log
Add a new mysql soundex implementation to correct current wrong behavior in calcite. 

# Verifying this change

SqlOperatorTests#testMysqlSoundex()


chucheng92:
hi @tanclary could you also help review? it's same like a previous `soundex_spark` function.

PR #3355, opened 2023-08-08
[CALCITE-5893] Wrong NULL operand behavior of ARRAY_CONTAINS/ARRAY_EXCEPT/ARRAY_INTERSECT In Spark Library

chucheng92:
https://issues.apache.org/jira/browse/CALCITE-5893

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
@@ -21,6 +21,7 @@ import org.apache.calcite.sql.SqlCallBinding;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlOperandCountRange;
 import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.SqlUtil;
 
 import com.google.common.collect.ImmutableList;
 
@@ -31,11 +32,41 @@ import static org.apache.calcite.util.Static.RESOURCE;
  * Parameter type-checking strategy where types must be Array and Array element type.
  */
 public class ArrayElementOperandTypeChecker implements SqlOperandTypeChecker {
+  //~ Instance fields --------------------------------------------------------
+
+  private final boolean allowNullCheck;
+  private final boolean allowCast;
+
+  //~ Constructors -----------------------------------------------------------
+
+  public ArrayElementOperandTypeChecker() {
+    this.allowNullCheck = false;
+    this.allowCast = false;
+  }
+
+  public ArrayElementOperandTypeChecker(boolean allowNullCheck, boolean allowCast) {
+    this.allowNullCheck = allowNullCheck;
+    this.allowCast = allowCast;
+  }
+
   //~ Methods ----------------------------------------------------------------
 
   @Override public boolean checkOperandTypes(
       SqlCallBinding callBinding,
       boolean throwOnFailure) {
+    if (allowNullCheck) {
+      // no operand can be null for type-checking to succeed
+      for (SqlNode node : callBinding.operands()) {
+        if (SqlUtil.isNullLiteral(node, allowCast)) {
+          if (throwOnFailure) {
+            throw callBinding.getValidator().newValidationError(node, RESOURCE.nullIllegal());
+          } else {
+            return false;
+          }
+        }
+      }
+    }
+
     final SqlNode op0 = callBinding.operand(0);
     if (!OperandTypes.ARRAY.checkSingleOperandType(
         callBinding,
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -6419,6 +6419,11 @@ public class SqlOperatorTest {
     f.checkType("array_contains(array[1, null], cast(null as integer))", "BOOLEAN");
     f.checkFails("^array_contains(array[1, 2], true)^",
         "INTEGER is not comparable to BOOLEAN", false);
+
+    // check null without cast
+    f.checkFails("array_contains(array[1, 2], ^null^)", "Illegal use of 'NULL'", false);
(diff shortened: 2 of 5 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6015,6 +6037,16 @@ private static void checkIf(SqlOperatorFixture f) {
     f.checkNull("array_intersect(cast(null as integer array), array[1])");
     f.checkNull("array_intersect(array[1], cast(null as integer array))");
     f.checkNull("array_intersect(cast(null as integer array), cast(null as integer array))");
+
+    // check null without cast
+    f.checkFails("^array_intersect(array[1, 2], null)^",
+        "Cannot apply 'ARRAY_INTERSECT' to arguments of type 'ARRAY_INTERSECT\\(<INTEGER ARRAY>, "
+            + "<NULL>\\)'\\. Supported form\\(s\\): 'ARRAY_INTERSECT\\(<EQUIVALENT_TYPE>, "
+            + "<EQUIVALENT_TYPE>\\)'", false);
+    f.checkFails("^array_intersect(null, array[1, 2])^",
```
mihaibudiu, line 6046: how about a test where both are null

core/src/main/java/org/apache/calcite/sql/type/NullOperandTypeChecker.java
```diff
@@ -0,0 +1,57 @@
...
+  public NullOperandTypeChecker(final int nOperands, final boolean allowCast) {
+    super(nOperands);
+    this.allowCast = allowCast;
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public boolean checkOperandTypes(final SqlCallBinding callBinding,
+      final boolean throwOnFailure) {
+    // all operands can't be null
```
mihaibudiu, line 45: I think it's clearer to say "no operand can be null for typechecking to succeed".

core/src/main/java/org/apache/calcite/sql/type/NullOperandTypeChecker.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.calcite.sql.SqlCallBinding;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlUtil;
+
+import static org.apache.calcite.util.Static.RESOURCE;
+
+/**
+ * Parameter type-checking strategy type must not be a NULL (including <code>NULL</code>,
+ * <code>CAST(NULL as ...)</code> but not <code>CAST(CAST(NULL as ...) AS ...)</code>).
+ */
```
mihaibudiu, line 28: This comment is not describing a type, it is describing an expression.
The only expression whose type is NULL is in fact a NULL literal.
Expressions of the form CAST(NULL as X) will have a different type.
So I think it's much simpler to check that the type is not SqlTypeName.NULL

mihaibudiu (changes requested):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
chucheng92: yes, added it.

core/src/main/java/org/apache/calcite/sql/type/NullOperandTypeChecker.java
chucheng92: yes, updated it.

core/src/main/java/org/apache/calcite/sql/type/NullOperandTypeChecker.java
chucheng92: updated it, pls review it.

mihaibudiu: This class should be called NotNullOperandTypeChecker.

core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
```diff
@@ -31,11 +32,41 @@
  * Parameter type-checking strategy where types must be Array and Array element type.
  */
 public class ArrayElementOperandTypeChecker implements SqlOperandTypeChecker {
+  //~ Instance fields --------------------------------------------------------
+
+  private final boolean allowNullCheck;
+  private final boolean allowCast;
```
mihaibudiu, line 38: if this is always "false", maybe it shouldn't be actually a field until there is some use for it.

chucheng92: Here, by adding member variables, these two parameters can be passed in through different constructors. For functions that do not require null checking, use the default constructor. For scenarios that require null checking or allow cast, it will be more versatile to pass them in through different constructors. what do you think?

core/src/main/java/org/apache/calcite/sql/type/NullOperandTypeChecker.java
chucheng92: yes, agree with you. renamed it.

core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
mihaibudiu: I understand why, but I am pointing out that all uses of this class now have allowCast = false.
Do you foresee a use case where allowCast = true?

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
chucheng92: Not currently, but since it is a public operand checker class, it may be used for other purposes in the future. I would be ok if I set it directly to false.

thanks for patient reviewing.

chucheng92:
thanks Mihai for reviewing. I will merge it later.

chucheng92:
Given that the PR has been approved and there have been no new other comments for a week, will merge it now ...

PR #3364, opened 2023-08-13
[CALCITE-5923] Some test cases in SqlOperatorTest violates the test fixture's design principle

herunkang2018:


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -441,8 +443,8 @@ protected SqlOperatorFixture fixture() {
   }
 
   /** Generates parameters to test both regular and safe cast. */
-  static Stream<Arguments> safeParameters() {
-    SqlOperatorFixture f = SqlOperatorFixtureImpl.DEFAULT;
+  private Stream<Arguments> safeParameters() {
```
herunkang2018, line 446: We remove static here to allow use of non-static method `fixture()` to use subclass's overridden implementation, and this needs to add `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` to this test class to allow this. See more at https://github.com/junit-team/junit5/issues/984

herunkang2018:
There are test cases fails in JDK11 with avatica main, I will check the reason.

herunkang2018:
The avatica error in CI seems related to CALCITE-5678, I will check it. The reason why this error appears is that it doesn't check result before, and after this fix, the test case fails.

zabetak:
Is there a JIRA/PR for fixing the problem in Avatica? Ideally we want to merge the changes in both repos roughly at the same time otherwise the CI will remain broken till this happens and we don't want that.

herunkang2018:
@zabetak It seems related to the fix in [CALCITE-5678](https://issues.apache.org/jira/browse/CALCITE-5678), which change to use ISO8601 to validate the timestamp string.
I think we can safely modify this test case to use `checkFails` with message `Invalid DATE value`. And more, we may need a flag to turn this modified test case off to pass latest avatica CI pipeline, after avatica-1.24 is released, we can trun the test case on.

herunkang2018:
@zabetak I agree with you. Since it is not a bug after the changes of [CALCITE-5678](https://issues.apache.org/jira/browse/CALCITE-5678), I have changed the broken tests's in this PR and turned off them. Now the CI with avatica main is passed, and after avatica-1.24 is released, these tests can be turned on.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1253,15 +1254,28 @@ void testCastStringToDateTime(CastType castType, SqlOperatorFixture f) {
...
+        f.checkFails("cast('1945-1-24 12:23:34.454' as TIMESTAMP)",
+            "Invalid DATE value, '1945-1-24 12:23:34.454'", true);
+      } else {
+        // test cases for 'SAFE_CAST' and 'TRY_CAST'
+        f.checkNull("cast('1945-2-2 12:2:5' as TIMESTAMP)");
+        f.checkNull("cast('1241241' as TIMESTAMP)");
+        f.checkNull("cast('1945-20-24 12:42:25.34' as TIMESTAMP)");
+        f.checkNull("cast('1945-01-24 25:42:25.34' as TIMESTAMP)");
+        f.checkNull("cast('1945-1-24 12:23:34.454' as TIMESTAMP)");
+      }
```
zabetak, line 1276: Would it be possible to avoid this if-else block by using `f.checkCastFails` or something similar that takes the `castType` as parameter?

zabetak (commented):
I made some small comments but we can't merge this PR yet till we see if the changes in CALCITE-5678 are going to be reverted or not.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
herunkang2018: Thanks for giving valuable comment, I will check it.

core/src/main/java/org/apache/calcite/util/Bug.java
```diff
@@ -208,6 +208,12 @@ public abstract class Bug {
    * MILLISECOND and MICROSECOND units in INTERVAL literal</a> is fixed. */
   public static final boolean CALCITE_5422_FIXED = false;
 
+  /** Whether
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5678">[CALCITE-5678]
+   * Calcite should reject date literals not satisfying Gregorian calendar,
+   * per SQL standard</a> is fixed. */
+  public static final boolean CALCITE_5678_FIXED = false;
```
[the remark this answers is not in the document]
herunkang2018: I will make it more concise.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
herunkang2018: I think `f.checkCastFails` cannot be used to refactor the code. The reason is that`f.checkCastFails` can only check exceptions, but 'SAFE_CAST' and 'TRY_CAST' will return `NULL` instead of throwing exception if the input is not a valid value when interpreted as the desired type.

core/src/main/java/org/apache/calcite/util/Bug.java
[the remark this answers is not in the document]
herunkang2018: Resolved.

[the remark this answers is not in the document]
zabetak: What do you think of https://github.com/apache/calcite/pull/3364/commits/f7caa3fb33fb33c1ef16293003ac4439dae9d571 ? Let's see if all tests run as expected.

build.gradle.kts
```diff
@@ -818,6 +818,7 @@ allprojects {
                 passProperty("user.language", "TR")
                 passProperty("user.country", "tr")
                 passProperty("user.timezone", "UTC")
+                passProperty("calcite.avatica.version", props.string("calcite.avatica.version"))
```
herunkang2018, line 821: Thanks for the code, it looks better and more reasonable.

PR #3365, opened 2023-08-13
 [CALCITE-5894] Add SortRemoveRedundantRule to remove redundant sort fields if they are functionally dependent on other sort fields

JingDas:
Sorting is an expensive operation, however. Therefore, it is imperative that sorting is optimized to avoid unnecessary sort field.

Add SortRemoveRedundantRule, it removes redundant sort fields if they are functionally dependent on other sort fields
RelMdFunctionalDependency is a minimal implementation that needs to be gradually improved in the future.


LakeShen:
If possible, add a SQL to the misc.iq file to test the correctness of the results with your case.

JingDas:
The code has been ready, I will be very grateful if someone can help me review the code

PR #3368, opened 2023-08-15
[CALCITE-5931] Allow round decimals like 1.00 in window ranges

arkanovicz:
Row counts literals in window expressions must be integral values.

This PR fixes the condition used to check that, to allow round decimal literals like "1.00" which have a strictly positive scale but are still integral values.


JiajunBernoulli:
@arkanovicz ,
We usually only use `[MINOR]` for document changes or code that users are not aware of.

Can you create a JIRA ticket for your PR.

arkanovicz:
The [PR](https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-5931) is open.

PR #3369, opened 2023-08-16
[CALCITE-5910] Add REGEXP_EXTRACT and REGEXP_SUBSTR functions (enabled in BigQuery library)

Anthrino:


babel/src/test/resources/sql/big-query.iq
```diff
@@ -3657,4 +3915,5 @@ FROM items;
 
 !ok
 
+
```
tanclary, line 3918: Did you mean to add this blank line back?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -355,11 +355,85 @@ public static boolean regexpContains(String value, String regex) {
       Pattern regexp = Pattern.compile(regex);
       return regexp.matcher(value).find();
     } catch (PatternSyntaxException ex) {
-      throw RESOURCE.invalidInputForRegexpContains(ex.getMessage().replace("\r\n", " ")
-          .replace("\n", " ").replace("\r", " ")).ex();
+      throw RESOURCE.invalidRegexInputForRegexpFunctions(ex.getMessage().replace("\r\n", " ")
+              .replace("\n", " ").replace("\r", " "),
```
tanclary, line 359: We should see if there's a way to avoid having to use all these `.replace()`

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -355,11 +355,85 @@ public static boolean regexpContains(String value, String regex) {
...
+          "REGEXP_CONTAINS").ex();
     }
   }
 
+  /** SQL {@code REGEXP_EXTRACT(value, regexp[, position[, occurrence]])} function.
+   *  Returns NULL if there is no match, or if position or occurrence are beyond range.
+   *  Returns an exception if regex, position or occurrence are invalid.*/
+  public static @Nullable String regexpExtract(String value, String regex, Integer... params) {
```
tanclary, line 367: is extract just a synonym? you can add it as an alias in `StandardConvertletTable` and then you can refactor/remove a lot this code. You shouldn't need a method for both functions here if they're the same. Let me know if you need more info.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -355,11 +355,85 @@ public static boolean regexpContains(String value, String regex) {
...
+    return processRegexpExtractOrSubstr("REGEXP_EXTRACT", value, regex, params);
+  }
+
+  /** SQL {@code REGEXP_SUBSTR(value, regexp[, position[, occurrence]])} function.
+   *  Returns NULL if there is no match, or if position or occurrence are beyond range.
+   *  Returns an exception if regex, position or occurrence are invalid.*/
```
tanclary, line 373: Add a space here

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -355,11 +355,85 @@ public static boolean regexpContains(String value, String regex) {
...
+  public static @Nullable String regexpSubstr(String value, String regex, Integer... params) {
+    return processRegexpExtractOrSubstr("REGEXP_SUBSTR", value, regex, params);
+  }
+
+  private static @Nullable String processRegexpExtractOrSubstr(String methodName, String value,
```
tanclary, line 378: once you add one as the alias of the other you can just change this to `regexpExtract` to match other functions

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -486,6 +483,25 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
+      SqlBasicFunction.create("REGEXP_EXTRACT", ReturnTypes.VARCHAR_NULLABLE,
+          OperandTypes.STRING_STRING_OPTIONAL_INTEGER_OPTIONAL_INTEGER,
+          SqlFunctionCategory.STRING);
+
+  @LibraryOperator(libraries = {MYSQL, ORACLE})
+  public static final SqlFunction REGEXP_REPLACE = new SqlRegexpReplaceFunction();
+
+  /** The "REGEXP_SUBSTR(value, regexp[, position[, occurrence]])" function.
+   * Returns the substring in value that matches the regexp. Returns NULL if there is no match. */
+  @LibraryOperator(libraries = {BIG_QUERY})
```
tanclary, line 499: You can just do ` = REGEXP_EXTRACT.withName('REGEXP_SUBSTR')`

site/_docs/reference.md
```diff
@@ -2779,7 +2779,9 @@ BigQuery's type system uses confusingly different names for types and functions:
 | h s | PARSE_URL(urlString, partToExtract [, keyToExtract] ) | Returns the specified *partToExtract* from the *urlString*. Valid values for *partToExtract* include HOST, PATH, QUERY, REF, PROTOCOL, AUTHORITY, FILE, and USERINFO. *keyToExtract* specifies which query to extract
 | b | POW(numeric1, numeric2)                        | Returns *numeric1* raised to the power *numeric2*
 | b | REGEXP_CONTAINS(string, regexp)                | Returns whether *string* is a partial match for the *regexp*
+| b | REGEXP_EXTRACT(string, regexp[, position[, occurrence]]) | Returns the substring in *string* that matches the regexp. Returns NULL if there is no match. Use *position* for the start index of search range and *occurrence* for the specific occurence of match in *string*
```
tanclary, line 2782: surround with *

site/_docs/reference.md
```diff
@@ -2779,7 +2779,9 @@ BigQuery's type system uses confusingly different names for types and functions:
...
 | m o | REGEXP_REPLACE(string, regexp, rep [, pos [, occurrence [, matchType]]]) | Replaces all substrings of *string* that match *regexp* with *rep* at the starting *pos* in expr (if omitted, the default is 1), *occurrence* means which occurrence of a match to search for (if omitted, the default is 1), *matchType* specifies how to perform matching
+| b | REGEXP_SUBSTR(string, regexp[, position[, occurrence]]) | Synonym for REGEXP_EXTRACT. Returns the substring in *string* that matches the regexp
```
tanclary, line 2784: You can drop this part

tanclary (commented):
left some initial comments

babel/src/test/resources/sql/big-query.iq
Anthrino: They popped up with the auto formatting, I'll take it out thanks for pointing out.
Do you know if the IS_NAN ones are valid (taking out the extra space) else I'll remove those too.

tanclary: Those are valid, it should remove blank space at the end of lines. I introduced those. Oops.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -355,19 +355,88 @@ public static String sha512(ByteString string)  {
...
+    if (position <= 0) {
+      throw RESOURCE.invalidIntegerInputForRegexpFunctions(Integer.toString(position),
+          "position", methodName).ex();
+    }
+    if (position > value.length()) {
+      return null;
+    }
+    matcher.region(position - 1, value.length());
+
+    if (occurrence <= 0) {
```
tanclary, line 413: Should we move the checks for `position` and `occurrence` to the top? That way we avoid trying to find the regex or matching it just to find out the arguments are invalid. What do you think?

tanclary: On top of that- maybe a good exercise for determining the ordering of the checks could be to run several queries against BQ with multiple errors (i.e. bad position and bad occurrence) and see which error it throws.

site/_docs/reference.md
```diff
@@ -2783,7 +2783,9 @@ BigQuery's type system uses confusingly different names for types and functions:
 | h s | PARSE_URL(urlString, partToExtract [, keyToExtract] ) | Returns the specified *partToExtract* from the *urlString*. Valid values for *partToExtract* include HOST, PATH, QUERY, REF, PROTOCOL, AUTHORITY, FILE, and USERINFO. *keyToExtract* specifies which query to extract
 | b | POW(numeric1, numeric2)                        | Returns *numeric1* raised to the power *numeric2*
 | b | REGEXP_CONTAINS(string, regexp)                | Returns whether *string* is a partial match for the *regexp*
+| b | REGEXP_EXTRACT(string, regexp[, position[, occurrence]]) | Returns the substring in *string* that matches the *regexp*. Returns `NULL` if there is no match. Use *position* for the start index of search range and *occurrence* for the specific occurence of match in *string*
```
tanclary, line 2786: Don't think NULL should be quoted based on other usages. 

Do you think you could rewrite to sort of match the `INSTR` description? That might be more simple and is a good example of a function with optional arguments. Let me know what you think.

site/_docs/reference.md
```diff
@@ -2783,7 +2783,9 @@ BigQuery's type system uses confusingly different names for types and functions:
...
 | m o | REGEXP_REPLACE(string, regexp, rep [, pos [, occurrence [, matchType]]]) | Replaces all substrings of *string* that match *regexp* with *rep* at the starting *pos* in expr (if omitted, the default is 1), *occurrence* means which occurrence of a match to search for (if omitted, the default is 1), *matchType* specifies how to perform matching
+| b | REGEXP_SUBSTR(string, regexp[, position[, occurrence]]) | Synonym for REGEXP_EXTRACT
```
tanclary, line 2788: add spaces between the parameter names and `[`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4568,6 +4568,25 @@ private static void checkIf(SqlOperatorFixture f) {
...
+    final SqlOperatorFixture f =
+        fixture().setFor(SqlLibraryOperators.REGEXP_EXTRACT).withLibrary(SqlLibrary.BIG_QUERY);
+
+    f.checkString("regexp_extract('abc def ghi', 'def')", "def", "VARCHAR NOT NULL");
+    f.checkString("regexp_extract('abcadcaecghi', 'a.c', 1, 3)", "aec", "VARCHAR NOT NULL");
+    f.checkString("regexp_extract('abcadcaecghi', 'abc(a.c)')", "adc", "VARCHAR NOT NULL");
+    f.checkString("regexp_extract('55as56664as422', '\\d{3}')", "566", "VARCHAR NOT NULL");
+
+    f.checkNull("regexp_extract('abc def ghi', 'asd')");
+    f.checkNull("regexp_extract('abc def ghi', cast(null as varchar))");
```
tanclary, line 4581: Can you add more tests that check the 3rd and 4th args (if they are out of range, null, etc)

tanclary (changes requested):
Left a few comments, the `SqlFunctions` refactor looks pretty good 👍

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
Anthrino: Added those as unit tests in the SqlFunctionsTest and big-query.iq file, i didnt add many here as they seemed redundant at every location should I include some here as well?

tanclary: Good question. I am a bit biased because I write most of my tests in `SqlOperatorTest` when adding a new operator. Might be best to get a second opinion.

site/_docs/reference.md
Anthrino: Thanks Tanner, but I also saw a couple of instances (MAP_CONCAT, ARRAY_CONCAT) where it was `NULL`, and another few places with returns null in lower case or NULL in upper, what should be the standard here a bit confused.

tanclary: oh interesting. in that case i guess either is fine but it should probably be standardized.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
Anthrino: Good point Tanner, technically the matcher is only initialized with the base string and regex at the top, and the pattern matching happens when we call the find() method on line 417, but I also think its better to catch the exceptions first and then initialize it. I had tested the order of errors, will reorganize them to do the exception handling first.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: I think it might be best to add some to `SqlOperatorTest` because that seems to serve as the main source of truth for how operators handle their edge cases, I don't necessarily think it's redundant. The operator tests check and showcase that the operator follows the behavior it should, while the function tests show that the actual implementation abides by that behavior as well. What do you think?

Anthrino: Sure Tanner, that makes sense we have test cases at each wrapper level so its best to validate at every stage, will add some here.

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
```diff
@@ -254,18 +254,88 @@ static <E> List<E> list() {
       final boolean b = f.regexpContains("abc def ghi", "[z-a]");
       fail("expected error, got " + b);
     } catch (RuntimeException e) {
-      assertThat(e.getMessage(),
-          is("Invalid regular expression for REGEXP_CONTAINS: 'Illegal "
-              + "character range near index" + " 3 [z-a]    ^'"));
+      assertThat(
+          e.getMessage(), is("Invalid regular expression for REGEXP_CONTAINS: 'Illegal "
```
tanclary, line 258: Is this formatting right?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -569,6 +570,8 @@ Builder populate() {
       defineReflective(PARSE_URL, BuiltInMethod.PARSE_URL2.method,
           BuiltInMethod.PARSE_URL3.method);
       defineReflective(REGEXP_CONTAINS, BuiltInMethod.REGEXP_CONTAINS.method);
+      defineReflective(REGEXP_EXTRACT, BuiltInMethod.REGEXP_EXTRACT2.method,
```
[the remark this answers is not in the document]
Anthrino: New for me too, followed the patterns Julian was using for the other functions.

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
Anthrino: It's not throwing any formatting errors, I just removed the extra str concat within the same line as it looked unnecessary

tanclary (approved):
Looks great, thanks for responding to all of the comments!

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
Anthrino: Added additional cases to cover the optional arguments, appreciate the feedback!

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -554,6 +555,7 @@ Builder populate() {
       defineMethod(SPLIT, "split", NullPolicy.STRICT);
       defineMethod(PARSE_URL, BuiltInMethod.PARSE_URL.method, NullPolicy.STRICT);
       defineMethod(REGEXP_CONTAINS, "regexpContains", NullPolicy.STRICT);
+      defineMethod(REGEXP_EXTRACT, "regexpExtract", NullPolicy.STRICT);
```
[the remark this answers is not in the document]
Anthrino: Rebased and aligned! :)

PR #3370, opened 2023-08-16
[CALCITE-5640] Add SAFE_ADD function (enabled in BigQuery library)

tanclary:
This PR continues working towards the goal of implementing all BigQuery safe arithmetic functions, captured in [CALCITE-5591](https://issues.apache.org/jira/browse/CALCITE-5591). SAFE_MULTIPLY was merged recently in [1b8dc6](https://github.com/apache/calcite/commit/1b8dc60b7673d83445f06ecaebd5f89dfd5781f6) and this PR follows a very similar structure, except of course using addition rather than multiplication. 

Any questions or suggestions would be appreciated as always!

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -1714,6 +1714,61 @@ public class SqlFunctions {
     throw notArithmetic("*", b0, b1);
   }
 
+  /** SQL <code>SAFE_ADD</code> function applied to long values. */
+  public static @Nullable Long safeAdd(long b0, long b1) {
+    try {
+      return Math.addExact(b0, b1);
+    } catch (ArithmeticException e) {
+      return null;
+    }
+  }
+
+  /** SQL <code>SAFE_ADD</code> function applied to long and BigDecimal values. */
+  public static @Nullable BigDecimal safeAdd(long b0, BigDecimal b1) {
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1714,6 +1714,64 @@ public static int multiply(int b0, int b1) {
...
+
+  /** SQL <code>SAFE_ADD</code> function applied to long and BigDecimal values. */
+  public static @Nullable BigDecimal safeAdd(long b0, BigDecimal b1) {
+    BigDecimal ans = BigDecimal.valueOf(b0).add(b1);
+    return safeDecimal(ans) ? ans : null;
+  }
+
+  /** SQL <code>SAFE_ADD</code> function applied to BigDecimal and long values. */
+  public static @Nullable BigDecimal safeAdd(BigDecimal b0, long b1) {
+    BigDecimal ans = b0.add(BigDecimal.valueOf(b1));
```
macroguo-ghy, line 1734: I believe using `return safeAdd(b1, b0)` would be better. 
Same comment for 
`Double safeAdd(long b0, double b1)`
`Double safeAdd(BigDecimal b0, double b1)`

tanclary: This is a great suggestion. I have updated the methods for safeAdd and safeMultiply as well. Let me know if you have any other thoughts, thanks!

macroguo-ghy (approved):
LGTM

PR #3375, opened 2023-08-18
[CALCITE-5583] JDBC adapter does not generate 'SELECT *' when duplicate field names

guestli:


core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -261,9 +263,24 @@ protected SqlCall as(SqlNode e, String alias, String... fieldNames) {
 
   /** Returns whether a list of expressions projects all fields, in order,
    * from the input, with the same names. */
-  public static boolean isStar(List<RexNode> exps, RelDataType inputRowType,
+  public static boolean isStar(List<RexNode> exps, RelNode input,
```
JiajunBernoulli, line 266: This is a public static method, it is best not to modify its parameters.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -261,9 +263,24 @@ protected SqlCall as(SqlNode e, String alias, String... fieldNames) {
...
       RelDataType projectRowType) {
     assert exps.size() == projectRowType.getFieldCount();
+    // Duplicate column names
+    // For example, EMP JOIN DEPT, both of which have columns named DEPTNO
+    if (input instanceof Join || input instanceof MultiJoin) {
```
JiajunBernoulli, line 271: Encapsulate this part of the code into an independent function, maybe it named `isJoinDuplicateName`.

Call function by `!isStar() && !isJoinDuplicateName()`.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -261,9 +263,24 @@ protected SqlCall as(SqlNode e, String alias, String... fieldNames) {
 
   /** Returns whether a list of expressions projects all fields, in order,
```
JiajunBernoulli, line 264: According to the code description, users may use it to determine Star.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
guestli: @JiajunBernoulli Thank you for your advice. I've considered adding a method before, but it might still cause an error when the isStar method is called from somewhere else. If the method name of 'isStar' is changed to something like 'projectAllField', `!projectAllField() && ! isJoinDuplicateName()` would that be better?
If you insist that the method name not be changed, I will also take your advice.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
guestli: According to the code description, I don't think it's enough to determine Star, so I think it's more appropriate to change the method name to 'projectAllField'.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
guestli: Oh, combined with the first comment, I get it, because it's a public static method, so the method name won't change. Then I'll take your advice.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
JiajunBernoulli: Maybe, but it's not necessary(The public static method needs to be stable).

Please remove 264-265 line, because it is same as 262-263 line.

guestli: Thanks for correction.

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -451,7 +453,8 @@ public Result visit(Project e) {
     }
     parseCorrelTable(e, x);
     final Builder builder = x.builder(e);
-    if (!isStar(e.getProjects(), e.getInput().getRowType(), e.getRowType())) {
+    if (isJoinDuplicateName(e.getInput()) || !isStar(e.getProjects(), e.getInput().getRowType(),
```
julianhyde, line 456: Break the line so that '||' is at start of line.

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -478,6 +481,26 @@ public Result visit(Project e) {
...
+   * Whether the relNode is a join and whether its inputs have duplicate field names.
+   * For example, EMP JOIN DEPT, both of which have columns named DEPTNO
+   */
+  private boolean isJoinDuplicateName(RelNode relNode) {
+    if (relNode instanceof Join || relNode instanceof MultiJoin) {
+      List<RelNode> inputs = relNode.getInputs();
+      RelNode firstRel = inputs.get(0);
+      List<String> firstRelFiledNames = Lists.newArrayList(firstRel.getRowType().getFieldNames());
+      for (int i = 1; i < inputs.size(); i++) {
+        RelNode input = inputs.get(i);
```
julianhyde, line 494: logic doesn't seem right if there are more than two inputs, and inputs 1 and 2 (0-based) have duplicate fields.

maybe write the field names into a list, and use `Util.isDistinct`

julianhyde: What if the target SQL dialect is case-insensitive? For such a dialect "select t0.x, t1.X from t0, t1" would be a problem.

Maybe add a `boolean caseSensitive` parameter to `Util.isDistinct`, default true, and if it's false create a set with `new TreeSet<String>(String.CASE_INSENSITIVE_ORDER)`

julianhyde: If it is a join with `USING` or `NATURAL JOIN` then the duplicate columns will have been filtered out. It's worth adding a test for `select * from emp join dept using (deptno)` and `select * from emp natural join dept` and make sure that `deptno` only occurs once in the output.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3529,21 +3553,39 @@ private SqlDialect nonOrdinalDialect() {
...
-    final String expected = "SELECT *\n"
+    final String expected = "SELECT employee.employee_id, employee.full_name, employee.first_name, "
+        + "employee.last_name, employee.position_id, employee.position_title, employee.store_id, "
+        + "employee.department_id, employee.birth_date, employee.hire_date, employee.end_date, "
+        + "employee.salary, employee.supervisor_id, employee.education_level, "
+        + "employee.marital_status, employee.gender, employee.management_role, "
+        + "department.department_id AS department_id0, department.department_description\n"
         + "FROM foodmart.employee AS employee\n"
-        + "INNER JOIN foodmart.department AS department "
-        + "ON employee.department_id = department.department_id";
```
julianhyde, line 3535: this diff (and a few others below) is spurious. you've just moved `ON` from one line to the next. please revert to the original formatting.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3750,17 +3792,35 @@ private SqlDialect nonOrdinalDialect() {
...
-        + "INNER JOIN \"foodmart\".\"customer\" "
-        + "ON \"sales_fact_1997\".\"customer_id\" = \"customer\".\"customer_id\""
-        + " OR \"sales_fact_1997\".\"customer_id\" IS NULL"
-        + " AND \"customer\".\"customer_id\" IS NULL"
-        + " OR \"customer\".\"occupation\" IS NULL\n"
-        + "INNER JOIN \"foodmart\".\"product\" "
-        + "ON \"sales_fact_1997\".\"product_id\" = \"product\".\"product_id\""
-        + " OR \"sales_fact_1997\".\"product_id\" IS NOT NULL"
-        + " OR \"product\".\"product_id\" IS NOT NULL";
+        + "INNER JOIN \"foodmart\".\"customer\" ON \"sales_fact_1997\".\"customer_id\" = "
```
julianhyde, line 3818: this diff is spurious. and the formatting is worse than before.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7013,6 +7157,38 @@ private void checkLiteral2(String expression, String expected) {
         .withOracle(11).throws_("Lower Oracle version(<12) doesn't support offset/fetch syntax!");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5583">[CALCITE-5583]
+   * Rel2Sql will get an error when select * and join is present</a>. */
+  @Test void testSelectStarWithJoin() {
+    final RelBuilder builder = relBuilder();
```
julianhyde, line 7164: modern tests use the `relFn()` method rather than `toSql(RelNode)`; don't call `relBuilder()` explicitly

guestli: julianhyde, thank you for your review and pointing out these errors. I will fix them.

core/src/main/java/org/apache/calcite/util/Util.java
```diff
@@ -2166,6 +2167,24 @@ public static <E> boolean isDistinct(List<E> list) {
...
+
+  /**
+   * Returns whether the elements of {@code list} are distinct according to a given
+   * case-sensitivity policy.
+   */
+  public static boolean isDistinct(List<String> list, boolean caseSensitive) {
+    if (caseSensitive) {
+      return isDistinct(list);
+    }
+    final Set<String> set = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
```
julianhyde, line 2179: if list has 0 or 1 elements you can immediately return true

guestli: @julianhyde Thank you for taking the time to review, and please review it again.

Lars-H:
Any updates on this issue?

PR #3378, opened 2023-08-18
[CALCITE-5770] Add SAFE_SUBTRACT function (enabled in BigQuery library)

tanclary:
This PR contains two commits implementing BigQuery's SAFE_DIVIDE and SAFE_SUBTRACT functions, working towards the goal of [CALCITE-5591]. These commits follow a very similar structure to those for [SAFE_ADD](https://github.com/apache/calcite/commit/a651ea6347f1ef87a5bf4de14b0c7e778ba393b5) and [SAFE_MULTIPLY](https://github.com/apache/calcite/commit/1b8dc60b7673d83445f06ecaebd5f89dfd5781f6). One notable difference is that there is logic to handle dividing by zero in the case of SAFE_DIVIDE. 

Any comments or suggestions are appreciated as always.


zinking:
looks good.

babel/src/test/resources/sql/big-query.iq
```diff
@@ -668,6 +668,77 @@ SELECT SAFE_ADD(CAST('NaN' AS DOUBLE), CAST(3 as BIGINT)) as NaN_result;
 (1 row)
 
 !ok
+
+#####################################################################
+# SAFE_DIVIDE
+#
+# SAFE_DIVIDE(value1, value2)
```
JiajunBernoulli, line 675: Can we have some tests with some expression parameters?

`SAFE_DIVIDE(1*1, 1-1)`
`SAFE_DIVIDE(1, 1/9223372036854775807)`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7401,6 +7470,87 @@ private static void checkIf(SqlOperatorFixture f) {
     f.checkNull("safe_multiply(cast(3 as double), cast(null as bigint))");
   }
 
+  @Test void testSafeSubtractFunc() {
```
JiajunBernoulli, line 7473: JIRA Doc is always popular (not required).

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1859,7 +1983,7 @@ private static boolean safeDouble(double d) {
     // negated MIN and MAX double values, overflow has not occurred. Otherwise,
     // overflow has occurred. Important to note that 'Double.MIN_VALUE' refers to
     // minimum positive value.
-    if (d < Double.MAX_VALUE && d > Double.MIN_VALUE) {
+    if (d < Double.MAX_VALUE && d > Double.MIN_VALUE || d == 0) {
```
zoudan, line 1986: Why we have to check `d == 0` here?

zoudan (changes requested):

babel/src/test/resources/sql/big-query.iq
tanclary: Will add these soon, good idea

babel/src/test/resources/sql/big-query.iq
```diff
@@ -668,6 +668,77 @@ SELECT SAFE_ADD(CAST('NaN' AS DOUBLE), CAST(3 as BIGINT)) as NaN_result;
 (1 row)
 
 !ok
+
+#####################################################################
+# SAFE_DIVIDE
```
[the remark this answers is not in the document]
tanclary: Yeah I figured there implementations were similar enough that it might be a bit tedious for reviewers to review two very similar PRs, but I'm happy to split it up if it's easier. What do you think?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: The way I check if a `Double` is "safe" or not is by checking whether it falls within the following ranges:

`Largest Negative Value` < `value` < `Smallest negative value` 

or

`Smallest positive value` < `value` < `Largest positive value`

The exception to this is zero, so I check for that as well.

Let me know if this makes sense or you think it should be changed.

babel/src/test/resources/sql/big-query.iq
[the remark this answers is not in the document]
zoudan: Two separate PRs look better to me, it would make things more clear.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
zoudan: Get it. Then how about adjusting it like this: 
```
return (d < Double.MAX_VALUE && d > Double.MIN_VALUE)
    || (d > -Double.MAX_VALUE && d < -Double.MIN_VALUE)
    || d == 0;
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: Done

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: DOne

tanclary:
@JiajunBernoulli @zoudan Thank you both for your comments, I have dropped the SAFE_DIVIDE commit from this PR to make it more clear. The expression parameters tests will be added in the SAFE_DIVIDE PR. Let me know if you have any other comments about SAFE_SUBTRACT, thanks!

zoudan:
LGTM

PR #3381, opened 2023-08-20
[CALCITE-5939] Support mv rewrite between join types

suibianwanwank:


core/src/main/java/org/apache/calcite/plan/expand/ExpandJoinOnLeftCalcToJoinUnifyRule.java
```diff
@@ -0,0 +1,140 @@
...
+import static org.apache.calcite.plan.expand.JoinRexUtils.isJoinRewritable;
+import static org.apache.calcite.plan.expand.JoinRexUtils.isSupportedPredict;
+import static org.apache.calcite.plan.expand.JoinRexUtils.tryCastRexInputRef;
+import static org.apache.calcite.plan.expand.SubstitutionRuleUtils.canPullUpFilterUnderJoin;
+import static org.apache.calcite.plan.expand.SubstitutionRuleUtils.fieldCnt;
+import static org.apache.calcite.plan.expand.SubstitutionRuleUtils.referenceByMapping;
+import static org.apache.calcite.plan.expand.SubstitutionRuleUtils.tryMergeParentCalcAndGenResult;
+
+/**
+ * Enhancements to joinOnLeftCalcToJoinUnifyRule.
```
LakeShen, line 49: Can you make the comments more detailed? It's better to include examples in the comments so that others can understand them easily,the other classes are the same

LakeShen:
Please check you PR workflow,there are many failures.

core/src/main/java/org/apache/calcite/plan/expand/ExpandJoinOnRightCalcToJoinUnifyRule.java
```diff
@@ -0,0 +1,147 @@
...
+import static org.apache.calcite.plan.expand.JoinRexUtils.tryCastRexInputRef;
+import static org.apache.calcite.plan.expand.SubstitutionRuleUtils.canPullUpFilterUnderJoin;
+import static org.apache.calcite.plan.expand.SubstitutionRuleUtils.fieldCnt;
+import static org.apache.calcite.plan.expand.SubstitutionRuleUtils.referenceByMapping;
+import static org.apache.calcite.plan.expand.SubstitutionRuleUtils.tryMergeParentCalcAndGenResult;
+
+/**
+ * Enhancements to joinOnRightCalcToJoinUnifyRule.
+ */
+public class ExpandJoinOnRightCalcToJoinUnifyRule
```
LakeShen, line 52: ExpandJoinOnRightCalcToJoinUnifyRule,ExpandJoinOnLeftCalcToJoinUnifyRule,ExpandJoinOnCalcToJoinUnifyRule,What is the difference between them？

suibianwanwank:
> Please check you PR workflow,there are many failures.

ok，I will fix it in these days

core/src/main/java/org/apache/calcite/plan/expand/ExpandJoinOnRightCalcToJoinUnifyRule.java
suibianwanwank: You can see that there are prototypes of these three rules in the rules that already exist in SubstitutionVisitor, because the unifyRule is an adaptation of the operator to the operator level, so different input structures of Join are treated differently, which you can also reflect in  tests
For example, the following two SQL statements correspond to JoinOnLeftCal and JoinOnRightCal, respectively
`    String query = ""
        + "select \"A\".\"empid\", \"A\".\"deptno\", \"depts\".\"deptno\" from\n"
        + " (select \"empid\", \"deptno\" from \"emps\" where \"deptno\" > 10) A"
        + " join \"depts\"\n"
        + "on \"A\".\"deptno\" = \"depts\".\"deptno\"" ;
`
`    String query = ""
        + "select \"A\".\"empid\", \"A\".\"deptno\", \"depts\".\"deptno\" from\n"
        + "  \"depts\"\n"
        + " join (select \"empid\", \"deptno\" from \"emps\" where \"deptno\" > 10) A"
        + "on \"depts\".\"deptno\" =  \"A\".\"deptno\" ;
`

suibianwanwank: Thank you for your attention to this issue, after some time to verify, I refactored this part of the code, optimized some processes and added more comments, you can reread this code if you're interested @LakeShen

PR #3383, opened 2023-08-20
[MINOR] Remove the unused imports in SqlAbstractParserImpl

Shoothzj:


macroguo-ghy:
I think using `[MINOR] Remove the unused imports in SqlAbstractParserImpl` would be better.

libenchao (approved):
+1, thanks for the PR!

I'm surprised that `./gradlew autostyleApply` does not throw for this class.

PR #3384, opened 2023-08-20
[CALCITE-5946] TimeString should allow fractional seconds ending in zero

Shoothzj:
This PR corrects the example usage of the withMillis, withNanos, and withFraction methods to match the TimeString constructor which only accepts a time string. And some other improvements

### DIFF
--- a/core/src/main/java/org/apache/calcite/util/TimeString.java
+++ b/core/src/main/java/org/apache/calcite/util/TimeString.java
@@ -20,13 +20,14 @@ import org.apache.calcite.avatica.util.DateTimeUtils;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.util.Calendar;
 import java.util.regex.Pattern;
+
+import static com.google.common.base.Preconditions.checkArgument;
 
 import static java.lang.Math.floorMod;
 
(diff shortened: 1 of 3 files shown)


mihaibudiu (approved):

core/src/main/java/org/apache/calcite/util/TimeString.java
```diff
@@ -76,35 +76,54 @@ private static String hms(int h, int m, int s) {
     return b.toString();
   }
 
-  /** Sets the fraction field of a {@code TimeString} to a given number
-   * of milliseconds. Nukes the value set via {@link #withNanos}.
+  /**
```
julianhyde, line 79: keep formatting (no newline after `/**`)

There are several arbitrary changes in this PR, each of which take the documentation further away from the parallel documentation in `TimestampString`.

core/src/main/java/org/apache/calcite/util/TimeString.java
```diff
@@ -76,35 +76,54 @@ private static String hms(int h, int m, int s) {
...
+   * Sets the fraction field of a {@code TimeString} to a given number of milliseconds.
+   * This will override any fraction value previously set via {@link #withNanos}.
```
julianhyde, line 81: use present tense, 'This overrides' rather than future tense 'This will override'.

core/src/main/java/org/apache/calcite/util/TimeString.java
```diff
@@ -76,35 +76,54 @@ private static String hms(int h, int m, int s) {
...
    *
-   * <p>For example,
-   * {@code new TimeString(1970, 1, 1, 2, 3, 4).withMillis(56)}
-   * yields {@code TIME '1970-01-01 02:03:04.056'}. */
+   * <p>For instance,
```
julianhyde, line 83: no good reason to change 'for example' to 'for instance'

core/src/main/java/org/apache/calcite/util/TimeString.java
```diff
@@ -76,35 +76,54 @@ private static String hms(int h, int m, int s) {
...
   }
 
-  /** Sets the fraction field of a {@code TimeString}.
-   * The precision is determined by the number of leading zeros.
-   * Trailing zeros are stripped.
+  /**
+   * Sets the fraction field of a {@code TimeString}.
+   * The precision is determined by the number of digits in the given fraction string.
+   * Trailing zeros are stripped to maintain precision.
+   *
```
julianhyde, line 119: I'm not sure what 'to maintain precision' means. In a sense, when you change millis from '100' to '110' you are adding one decimal digit of precision (because the time value now has one more significant digit), and when you change it back to '100' you are losing a digit of precision.

If we really need to explain the details - and I'm not sure we do - we are maintaining the canonical form, which is that there are no trailing zeros after the decimal point.

julianhyde:
Can you check whether CALCITE-5538 (`TimestampString`) also applies to `TimeString`? If so a similar fix may be necessary.

Shoothzj:
@julianhyde Thanks for your review. I will check it later.

Shoothzj:
@julianhyde Could you please take a look again? Thanks

julianhyde:
You force-pushed a single commit. It's difficult to re-review.

julianhyde:
Ah, I see, this is a port of the fix for https://issues.apache.org/jira/browse/CALCITE-5538 and also corrects the mistakes in the javadoc comments.

Makes sense to do it as a single commit, and looks good.

Shoothzj:
> You force-pushed a single commit. It's difficult to re-review.

Sorry, But I think it's a little big change for this PR. I prefer to restart it.

PR #3387, opened 2023-08-21
[CALCITE-5911] Add REGEXP_EXTRACT_ALL function (enabled in BigQuery library)

Anthrino:


core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -216,6 +216,8 @@
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.PARSE_URL;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.POW;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.REGEXP_CONTAINS;
```
tanclary, line 218: Is this PR for REGEXP_EXTRACT and REGEXP_EXTRACT_ALL? Can you update the title/commit message if so

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -347,19 +347,119 @@ public static String sha512(ByteString string)  {
...
+    return match;
+  }
+
+  /** SQL {@code REGEXP_EXTRACT_ALL(value, regexp)} function.
+   *  Returns an empty array if there is no match, returns an exception if regex is invalid.*/
+  public static List<String> regexpExtractAll(String value, String regex) {
+    // Uses java.util.regex as a standard for regex processing
+    // in Calcite instead of RE2 used by BigQuery/GoogleSQL
+    Pattern regexp;
+    String methodName = "REGEXP_EXTRACT_ALL";
```
tanclary, line 438: REGEXP_EXTRACT and REGEXP_EXTRACT_ALL are pretty similar. What do you think of anytime REGEXP_EXTRACT gets called, you can call REGEXP_EXTRACT_ALL method and then just retrieve the requested element (e.g. if the occurrence is 2 then retrieve the second element). Would this simplify things?

site/_docs/reference.md
```diff
@@ -2779,7 +2779,10 @@ BigQuery's type system uses confusingly different names for types and functions:
 | h s | PARSE_URL(urlString, partToExtract [, keyToExtract] ) | Returns the specified *partToExtract* from the *urlString*. Valid values for *partToExtract* include HOST, PATH, QUERY, REF, PROTOCOL, AUTHORITY, FILE, and USERINFO. *keyToExtract* specifies which query to extract
 | b | POW(numeric1, numeric2)                        | Returns *numeric1* raised to the power *numeric2*
 | b | REGEXP_CONTAINS(string, regexp)                | Returns whether *string* is a partial match for the *regexp*
+| b | REGEXP_EXTRACT(string, regexp[, position[, occurrence]]) | Returns the substring in *string* that matches the *regexp*. Returns `NULL` if there is no match. Use *position* for the start index of search range and *occurrence* for the specific occurence of match in *string*
+| b | REGEXP_EXTRACT_ALL(string, regexp)             | Returns an array of all substrings in *string* that matches the regexp. Returns an empty array if there is no match
```
tanclary, line 2783: Surround with *

tanclary (commented):
Left a few initial comments

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
Anthrino: That makes sense @tanclary, I considered that approach but REGEXP_EXTRACT_ALL does not have the position or occurrence params like REGEXP_EXTRACT, so merging the functionality as you suggested would mean we find all occurrences of the pattern for every call to EXTRACT and then picking the required index (also these are optional params so many a times we would be only looking for the default first occurrence). This can be avoided by passing the occurrence to EXTRACT_ALL and stopping the loop on reaching that count, but we'd have to include some logic to skip this when we need the whole list of matches (extract_all). I felt this might be added complexity with some hit on the performance, hence decoupled it even though they have similar functionality. What do you think is it better to consolidate them?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
Anthrino: I know that's a bit confusing, this PR includes changes from EXTRACT cherrypicked here due to dependencies. Should be resolved when we merge that PR and rebase this one.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: That makes sense. I just noticed some of the logic is repeated between the functions so maybe at the least we can extract that into a helper or something. We should try to avoid repeated checks/logic across these methods if we can.

Anthrino: Hi @tanclary, moved the redundant error checks into a couple of helper functions to reuse majority of the code, I've kept the matching logic separate due to the above reasons would be good if you can have a look and suggest any improvements if needed.

tanclary: That sounds good, once we get REGEXP_EXTRACT merged and can rebase this I'll come back for a second review 👍

Anthrino: @tanclary I've rebased the changes from previous PRs, also consolidated few of the redundant error checks into helper methods would be good if you can give it a review!

tanclary (commented):
I think this looks pretty good..no additional comments at first glance but I will wait a bit to see if anyone else has comments before I merge.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4675,6 +4675,29 @@ private static void checkIf(SqlOperatorFixture f) {
...
+        + "NOT NULL ARRAY NOT NULL");
+    f.checkScalar("regexp_extract_all('abcde', '.')", "[a, b, c, d, e]", "CHAR(5) NOT NULL ARRAY "
+        + "NOT NULL");
+    f.checkScalar("regexp_extract_all('abcadcabcaecghi', 'ac')", "[]", "CHAR(15) NOT NULL ARRAY "
+        + "NOT NULL");
+    f.checkScalar("regexp_extract_all('foo@bar.com, foo@gmail.com, foo@outlook.com', "
+        + "'@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+')", "[@bar.com, @gmail.com, @outlook.com]", "CHAR(43) "
+        + "NOT NULL ARRAY NOT NULL");
+
+    f.checkNull("regexp_extract_all('abc def ghi', cast(null as varchar))");
```
ILuffZhe, line 4692: In your tests, I noticed that either the value or the regex is null, the function returns null.
What if we use `ReturnTypes.ARG0_NULLABLE` for this func?

ILuffZhe (commented):
Overall looks good, just one minor comment.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
Anthrino: Good catch @ILuffZhe, thanks! 
Made the change in sql lib operator definition, I missed that assuming its always returning an array (empty for no matches).

tanclary (approved):
Looks good, Jerin. Thanks for your patience on this.

PR #3389, opened 2023-08-22
[CALCITE-5935] Add CODE_POINTS_TO_BYTES function (enabled in BigQuery…

macroguo-ghy:
… library)

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -1116,6 +1116,28 @@ public class SqlFunctions {
     return String.valueOf(Character.toChars(n));
   }
 
+  /**
+   * SQL CODE_POINTS_TO_BYTES function.
+   */
+  public static @Nullable ByteString codePointsToBytes(List codePoints) {
+    int length = codePoints.size();
+    byte[] bytes = new byte[length];
+    for (int i = 0; i < length; i++) {
+      Object codePoint = codePoints.get(i);
+      if (codePoint == null) {
+        return null;
+      }
+      assert codePoint instanceof Number;
+      long cp = ((Number) codePoint).longValue();
+      if (cp < 0 || cp > 255) {
+        throw RESOURCE.inputArgumentsOfCodePointsToBytesOutOfRange(cp).ex();
+      }
+      bytes[i] = (byte) cp;
+    }
+
+    return new ByteString(bytes);
+  }
+
   /** SQL OCTET_LENGTH(binary) function. */
   public static int octetLength(ByteString s) {
     return s.length();
--- a/core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
@@ -513,6 +513,38 @@ public abstract class OperandTypes {
         }
       };
 
+  public static final SqlSingleOperandTypeChecker ARRAY_OF_INTEGER =
+      new FamilyOperandTypeChecker(ImmutableList.of(SqlTypeFamily.ARRAY), i -> false) {
+        @Override public boolean checkSingleOperandType(SqlCallBinding callBinding, SqlNode operand,
+            int iFormalOperand, SqlTypeFamily family, boolean throwOnFailure) {
+          if (!super.checkSingleOperandType(
+              callBinding,
+              operand,
+              iFormalOperand,
+              family,
+              throwOnFailure)) {
+            return false;
+          }
+          RelDataType type = SqlTypeUtil.deriveType(callBinding, operand);
+          if (SqlTypeUtil.isNull(type)) {
+            return true;
+          }
+          RelDataType componentType =
+              requireNonNull(type.getComponentType(), "componentType");
+          if (SqlTypeUtil.isIntType(componentType) || SqlTypeUtil.isNull(componentType)) {
+            return true;
+          }
+
+          if (throwOnFailure) {
+            throw callBinding.newValidationSignatureError();
+          }
+          return false;
+        }
+
+        @Override public String getAllowedSignatures(SqlOperator op, String opName) {
+          return opName + "(<INTEGER ARRAY>)";
+        }
+      };
 
   /** Checks that returns whether a value is a multiset or an array.
    * Cf Java, where list and set are collections but a map is not. */
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2685,6 +2685,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | CEIL(value)                                    | Similar to standard `CEIL(value)` except if *value* is an integer type, the return type is a double
 | m s | CHAR(integer)                                | Returns the character whose ASCII code is *integer* % 256, or null if *integer* &lt; 0
 | b o p | CHR(integer)                               | Returns the character whose UTF-8 code is *integer*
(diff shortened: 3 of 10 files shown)


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -929,6 +929,29 @@ public static String charFromUtf8(int n) {
     return String.valueOf(Character.toChars(n));
   }
 
+  /**
+   * SQL CODE_POINTS_TO_BYTES function.
+   */
+  public static @Nullable ByteString codePointsToBytes(List codePoints) {
```
tanclary, line 935: Is `codePoints` always a list of integers? If so you could set the parameter type as List<Integer> and then some of the logic below becomes unnecessary. Let me know if you see problems with this.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -929,6 +929,29 @@ public static String charFromUtf8(int n) {
...
+    int length = codePoints.size();
+    byte[] bytes = new byte[length];
+    for (int i = 0; i < length; i++) {
+      Object codePoint = codePoints.get(i);
+      if (codePoint == null) {
+        return null;
+      }
+      if (codePoint instanceof Number) {
```
tanclary, line 943: This is what I was referring to in my above comment.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1804,6 +1804,35 @@ void testCastToBoolean(CastType castType, SqlOperatorFixture f) {
...
+            + "Supported form\\(s\\): CODE_POINTS_TO_BYTES\\(<INTEGER ARRAY>\\)",
+        false);
+
+    f.checkFails("code_points_to_bytes(array[-1])",
+        "Input arguments of CODE_POINTS_TO_BYTES out of range: -1", true);
+    f.checkFails("code_points_to_bytes(array[2147483648, 1])",
+        "Input arguments of CODE_POINTS_TO_BYTES out of range: 2147483648", true);
+
+    f.checkString("code_points_to_bytes(array[65, 66, 67, 68])", "41424344", "VARBINARY NOT NULL");
+    f.checkString("code_points_to_bytes(array[255, 254, 65, 64])", "fffe4140",
```
tanclary, line 1828: What happens if you pass in an empty array? Also can the array arguments be expressions that evaluate to integers like: `array[1+2, 3, 4]`?

tanclary (commented):
Left a couple of small comments!

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
macroguo-ghy: Thanks for you review:) @tanclary.
Actually, the elements of codePoints can be of type Long or Integer. Additionally, let's consider an input of `17179869185` (which is equal to `(1L << 34) + 1L`). Although this is an illegal argument, when it is cast to an integer (using the `intValue()` method), it will become 1! This will result in an incorrect output. So, I believe using this logic is reasonable. WDYT?

herunkang2018: I notice that operand type chceker is `ARRAY_OF_INTEGER`, so IMO in runtime, it always be array of integer as operand, and not necessary to add the check logic for long value.

macroguo-ghy: @herunkang2018 Thanks for your advice. Calcite often use `SqlTypeFamily.INTEGER`(which is collection of `INT_TYPES` in `SqlTypeName`) as a integer operand checker instead of `SqlTypeName.INTEGER`. This can easily be confusing. So `ARRAY_OF_INTEGER` means an array whose component could be `TINYINT, SMALLINT, INTEGER or BIGINT`.
Back to your comment, array component could be any of the `INT_TYPES`, including `BIGINT`(Correspond Long in java). The reason I allow the `BIGINT` component to pass this checker is because if `BIGINT` component is invaild input type, `code_points_to_byte(array[cast(1 as bigint)])` will fail during operand type checking, this does not meet expectations. And  if I execute `select code_points_to_bytes(array[2147483648])` in BigQuery, BigQuery will throw a exception :`Input arguments of CODE_POINTS_TO_BYTES out of range: 2147483648` rather than input type error(Actully BigQuery see all integers as int64).

herunkang2018: Thanks for clarification, I got it.

macroguo-ghy:
Hi~ @tanclary @herunkang2018. Could you please review this PR again?

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -510,6 +510,38 @@ public static SqlOperandTypeChecker variadic(
         }
       };
 
+  public static final SqlSingleOperandTypeChecker ARRAY_OF_INTEGER =
```
tanclary, line 513: Just to clarify, what is the benefit of adding this if you just end up checking that each array element is a number in the `SqlFunctions` method you added?

site/_docs/reference.md
```diff
@@ -2685,6 +2685,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | CEIL(value)                                    | Similar to standard `CEIL(value)` except if *value* is an integer type, the return type is a double
 | m s | CHAR(integer)                                | Returns the character whose ASCII code is *integer* % 256, or null if *integer* &lt; 0
 | b o p | CHR(integer)                               | Returns the character whose UTF-8 code is *integer*
+| b | CODE_POINTS_TO_BYTES(integers)                 | Converts *integers*, an array of integers between 0 and 255 inclusive, throws error if any element is out of range
```
tanclary, line 2688: To keep consistent with other function docs, should it be rephrased to something like "Converts integers to ..."?

tanclary (commented):
I think this looks pretty good. I just left a couple of comments.

site/_docs/reference.md
macroguo-ghy: fixed

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
tanclary: Sounds good, thanks for explaining!

tanclary (approved):
I think this looks good. I will give it a bit to see if anyone else has comments and if not, then I will merge. Thank you for your patience!

site/_docs/reference.md
herunkang2018: Minor comment, could we add the target result type in doc? Like "Converts *integers*, an array of integers between 0 and 255 inclusive, to a binary string, throws error if any element is out of range"

macroguo-ghy: Hi @herunkang2018, I have made this change, you refered to the outdated code.

macroguo-ghy:
> 

Hi, @tanclary would you like to merge this PR?

tanclary:
@macroguo-ghy yes, would you mind rebasing/squashing down to a final commit? Sorry for any inconvenience and delay in review I have been a bit busy.

herunkang2018:
@macroguo-ghy LGTM. Remind that there are some duplicated commits compare to main branch, would you like to rebase from the main branch to fix this?

macroguo-ghy:
I have rebased and squashed it. @tanclary

PR #3390, opened 2023-08-22
[CALCITE-5949] RexExecutable should return unchanged original expressions when it fails

arkanovicz:
While reducing, when encountering an invalid expression in the list of constant expressions, RexExecutor is meant to return all initial expressions unchanged.

It fails to do so, because already handled correct expressions have already been added to the returned list, which can be greater than the input list.

For instance, when given the list `{ LN(2), LN(-2) }`, the RexExecutor will output a list of length 3.

This PR corrects this problem, and adds a corresponding test case.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexExecutable.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexExecutable.java
@@ -88,7 +88,7 @@ public class RexExecutable {
         values = new Object[constExps.size()];
       } else {
         assert values.length == constExps.size();
-        final List<RexNode> successfullyReduced = new ArrayList<>();
+        final List<RexNode> successfullyReduced = new ArrayList<>(constExps.size());
         final List<@Nullable Object> valueList = Arrays.asList(values);
         for (Pair<RexNode, @Nullable Object> value : Pair.zip(constExps, valueList)) {
           successfullyReduced.add(
--- a/core/src/main/java/org/apache/calcite/rex/RexExecutable.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexExecutable.java
@@ -32,6 +32,7 @@ import java.io.Serializable;
 import java.io.StringReader;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
@@ -87,11 +88,14 @@ public class RexExecutable {
         values = new Object[constExps.size()];
       } else {
         assert values.length == constExps.size();
+        final List<RexNode> successfullyReduced = new ArrayList<>();
         final List<@Nullable Object> valueList = Arrays.asList(values);
         for (Pair<RexNode, @Nullable Object> value : Pair.zip(constExps, valueList)) {
-          reducedValues.add(
+          successfullyReduced.add(
               rexBuilder.makeLiteral(value.right, value.left.getType(), true));
         }
+        assert successfullyReduced.size() == constExps.size();
+        reducedValues.addAll(successfullyReduced);
       }
     } catch (RuntimeException e) {
       // One or more of the expressions failed.
--- a/core/src/test/java/org/apache/calcite/rex/RexExecutorTest.java
+++ b/core/src/test/java/org/apache/calcite/rex/RexExecutorTest.java
@@ -378,4 +378,28 @@ class RexExecutorTest {
   interface Action {
     void check(RexBuilder rexBuilder, RexExecutorImpl executor);
   }
+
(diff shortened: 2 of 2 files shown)


rubenada:
@arkanovicz could you please rebase with latest main and squash commits?

arkanovicz:
> @arkanovicz could you please rebase with latest main and squash commits?

@rubenada Done.

core/src/test/java/org/apache/calcite/rex/RexExecutorTest.java
```diff
@@ -378,4 +378,28 @@ public void run() {
   interface Action {
     void check(RexBuilder rexBuilder, RexExecutorImpl executor);
   }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5949">[CALCITE-5949]
+   * RexExecutable should properly handle invalid constant expressions in reduction</a>.
```
rubenada, line 384: @arkanovicz could you please align this comment with the latest title / commit message ("RexExecutable should return unchanged original expressions when it fails")?
Then please squash commits into a single one, and I'll proceed with the merge

core/src/main/java/org/apache/calcite/rex/RexExecutable.java
```diff
@@ -87,11 +88,14 @@ public void reduce(RexBuilder rexBuilder, List<RexNode> constExps,
...
       } else {
         assert values.length == constExps.size();
+        final List<RexNode> successfullyReduced = new ArrayList<>();
         final List<@Nullable Object> valueList = Arrays.asList(values);
         for (Pair<RexNode, @Nullable Object> value : Pair.zip(constExps, valueList)) {
-          reducedValues.add(
+          successfullyReduced.add(
               rexBuilder.makeLiteral(value.right, value.left.getType(), true));
         }
+        assert successfullyReduced.size() == constExps.size();
```
JiajunBernoulli, line 97: `values.length == constExps.size();` and `successfullyReduced.size() == values.length`.

Should we need `assert successfullyReduced.size() == constExps.size();`?

rubenada: @JiajunBernoulli  sorry I don't quite understand your comment?, wdym?

core/src/main/java/org/apache/calcite/rex/RexExecutable.java
```diff
@@ -87,11 +88,14 @@ public void reduce(RexBuilder rexBuilder, List<RexNode> constExps,
         values = new Object[constExps.size()];
       } else {
         assert values.length == constExps.size();
+        final List<RexNode> successfullyReduced = new ArrayList<>();
```
rubenada, line 91: minor optim: `... = new ArrayList<>(constExps.size());`

arkanovicz: Added.

core/src/main/java/org/apache/calcite/rex/RexExecutable.java
JiajunBernoulli: Maybe `assert successfullyReduced.size() == constExps.size()` is redundant.

core/src/test/java/org/apache/calcite/rex/RexExecutorTest.java
rubenada: @arkanovicz  kind reminder about this minor (old) remark regarding the ticket title on this javadoc.

core/src/main/java/org/apache/calcite/rex/RexExecutable.java
rubenada: Certainly I can't imagine a scenario where that equality won't hold (if the `Pair#zip` behaves as it should), but the assertion doesn't hurt (and assertions can be disabled anyway), so I'm not against having it.

core/src/test/java/org/apache/calcite/rex/RexExecutorTest.java
rubenada: @arkanovicz we are approaching the preparation for the next release, could you please carry out these last details ⬆️  in order to merge this PR? thanks

libenchao:
@arkanovicz Are you available to move this forward and get it in 1.36.0? It seems this PR is in a very good shape already.

rubenada:
@libenchao , considering that we are approaching the RC deadline (it would be sad to not include this on it); in case @arkanovicz cannot respond in 24h, I'll create a separate PR (tagging him as co-author) to finalize this fix.

rubenada:
Closing this in favor of https://github.com/apache/calcite/pull/3486

PR #3392, opened 2023-08-22
[CALCITE-5938] Update HSQLDB to Version 2.7.2 (using JDK8 JAR, default supports JDK11+)

tanclary:
Had to use special classifier syntax to use alternative JAR that supports JDK8
@shalaniw


julianhyde:
@tanclary Well done finding the special classifier syntax. I spent a long time searching and never found it.

Do you think it would work if you made the change only in `bom/build.gradle.kts`? That file is the one and only place where dependencies are declared; all other places are just references to that.

tanclary:
@julianhyde Tried the change you suggested, seems to be failing as it tries to call ".version" on the classifier which of course doesn't exist. Going to revert.

julianhyde:
Thanks for trying the BOM update.

+1 to your original PR, but please add comments for why the '::jdk8' suffix is necessary. The commit message would seem to be a good place.

tanclary:
@julianhyde Done. Thanks for your help.

PR #3393, opened 2023-08-23
[CALCITE-5950] Default column constraint is erroneously processed

zstan:


zstan:
patch is ready for review, did i neet to fix all CI / CheckerFramework issues ? seems they are false positive.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4769,21 +4779,47 @@ private void convertValuesImpl(
...
-              : tmpBb.root;
+              SqlValidatorUtil.alias(operand, i));
+        }
+      });
+
+      RelNode in = (null == tmpBb.root) ? LogicalValues.createOneRow(cluster) : tmpBb.root;
+
       relBuilder.push(in)
-          .project(exps.leftList(), exps.rightList());
+          .project(Pair.left(exps), Pair.right(exps));
```
julianhyde, line 4822: why did you revert?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4792,6 +4828,35 @@ private void convertValuesImpl(
...
+          mapping[pos] = i;
+        }
+      }
+    });
+
+    if (this.getClass().desiredAssertionStatus()) {
+      long items = Arrays.stream(mapping).count();
+      if (items != targetFields.size()) {
+        throw new AssertionError("Columns partially mapped, src=" + targetFields
+            + ", dest=" + tblFields.stream().map(f -> f.getName()).collect(Collectors.toList()));
```
julianhyde, line 4853: prefer `Util.transform(tblFields, RelField::getName)`

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4809,6 +4874,30 @@ private static class RegisterArgs {
     }
   }
 
+  /** Determine presence of {@link SqlKind.DEFAULT} type. */
+  private static class DefaultChecker extends SqlShuttle {
```
julianhyde, line 4878: move this into SqlUtil. you should be able to use a pattern very similar to SqlUtil.containsIn

testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
```diff
@@ -118,7 +118,6 @@ public abstract class QuidemTest {
 
   @SuppressWarnings({"BetaApi", "UnstableApiUsage"})
   protected static Collection<String> data(String first) {
-    // inUrl = "file:/home/fred/calcite/core/target/test-classes/sql/agg.iq"
```
julianhyde, line 121: why delete this comment?

server/src/test/java/org/apache/calcite/test/ServerParserTest.java
```diff
@@ -56,6 +56,11 @@ class ServerParserTest extends SqlParserTest {
         .ok("CREATE SCHEMA `X`");
   }
 
+  @Test void testProcessCreateTableWithDefaul() {
```
julianhyde, line 59: s/Defaul/Default/

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4792,6 +4828,35 @@ private void convertValuesImpl(
...
+      if (strategies.get(i) == ColumnStrategy.DEFAULT) {
+        int pos = targetFields.indexOf(fld.getName());
+        if (pos != -1) {
+          mapping[pos] = i;
+        }
+      }
+    });
+
+    if (this.getClass().desiredAssertionStatus()) {
+      long items = Arrays.stream(mapping).count();
```
julianhyde, line 4850: why not just `mapping.length`?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4792,6 +4828,35 @@ private void convertValuesImpl(
...
+    Ord.forEach(tblFields, (fld, i) -> {
+      if (strategies.get(i) == ColumnStrategy.DEFAULT) {
+        int pos = targetFields.indexOf(fld.getName());
+        if (pos != -1) {
+          mapping[pos] = i;
+        }
+      }
+    });
+
+    if (this.getClass().desiredAssertionStatus()) {
```
julianhyde, line 4849: I'd remove this `if`. This is not performance-critical code.  The assertion is worth checking even if assertions are disabled.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4769,21 +4779,47 @@ private void convertValuesImpl(
...
+    assert insertOp instanceof SqlInsert || !processDefaults : "operation: " + insertOp;
+
+    @Nullable InitializerExpressionFactory initializerFactory = processDefaults
+            ? getInitializerFactory(getNamespace(requireNonNull(insertOp, "insertOp")).getTable())
+        : null;
+    @Nullable RelOptTable targetTable = processDefaults
+        ? getTargetTable(requireNonNull(insertOp, "insertOp")) : null;
+    int[] mapping = processDefaults
+        ? processColumnsMapping(requireNonNull(targetRowType, "targetRowType"),
+        requireNonNull(targetTable, "targetTable")) : null;
```
julianhyde, line 4799: This code block is hard to read. All 3 variables have `processDefaults ?` in their expression, and your inconsistent formatting doesn't help. One `if (processDefaults)` block would be easier to read.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -281,6 +285,8 @@ public class SqlToRelConverter {
    */
   private final Deque<String> datasetStack = new ArrayDeque<>();
 
+  private final Deque<SqlCall> callStack = new ArrayDeque<>();
```
julianhyde, line 288: needs javadoc

zstan:
@tanclary can u make a review plz ?

LakeShen:
Hi @zstan ,could you rebase the latest  code of main branch?

Then the commits and file changed will be clean.

zstan:
@LakeShen thanks, done

core/src/main/java/org/apache/calcite/sql/SqlUtil.java
```diff
@@ -1238,6 +1238,17 @@ public static boolean containsIn(SqlNode node) {
     return containsCall(node, callPredicate);
   }
 
+  /**
+   * Determine presence at least one node with {@link SqlKind#DEFAULT} type.
+   *
+   * @param node AST tree
+   */
+  public static boolean containsDefault(SqlNode node) {
+    final Predicate<SqlCall> callPredicate = call ->
```
LakeShen, line 1247: I saw other containsXXX method used `call.getOperator() instanceof XXXOperator`.

Could we use `call -> call.getOperator() instanceof SqlDefaultOperator` here to keep the same behavior with others?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -281,6 +283,9 @@ public class SqlToRelConverter {
    */
   private final Deque<String> datasetStack = new ArrayDeque<>();
 
+  /** Stack of currently processed operators. */
```
LakeShen, line 286: This code comment could be more detailed,you could give a code comment as to why this stack is there,it is better to help others understand.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -3801,13 +3806,17 @@ private static boolean all(SqlCall call) {
...
+    callStack.push(call);
+
     final RelDataType targetRowType =
         validator().getValidatedNodeType(call);
     assert targetRowType != null;
     RelNode sourceRel =
         convertQueryRecursive(call.getSource(), true, targetRowType).project();
     RelNode massagedRel = convertColumnList(call, sourceRel);
 
+    callStack.pop();
```
LakeShen, line 3818: Why callStack only used in convertInsert,will the other convertXXX methods use callStack?

Such as convertMerge method.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4792,6 +4833,32 @@ private void convertValuesImpl(
         true);
   }
 
+  private int[] processColumnsMapping(RelDataType targetRowType, RelOptTable targetTable) {
```
LakeShen, line 4836: I think that it is better to give the code comments to explain what this method does

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4769,19 +4778,51 @@ private void convertValuesImpl(
       return;
     }
 
-    for (SqlNode rowConstructor1 : values.getOperandList()) {
-      SqlCall rowConstructor = (SqlCall) rowConstructor1;
+    SqlCall insertOp = callStack.peek();
+    boolean processDefaults = insertOp != null && containsDefault(values);
```
LakeShen, line 4782: I think this large block of code is the main core logic, but I think there are some code comments that could help others understand and read the code.

WDYT?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4769,19 +4778,51 @@ private void convertValuesImpl(
...
+      targetTable = getTargetTable(requireNonNull(insertOp, "insertOp"));
+      initializerFactory =
+          getInitializerFactory(getNamespace(requireNonNull(insertOp, "insertOp"))
+              .getTable());
+      mapping =
+          processColumnsMapping(requireNonNull(targetRowType, "targetRowType"), targetTable);
+    }
+
+    for (SqlNode rowConstructor : values.getOperandList()) {
+      SqlCall rowConstructor0 = (SqlCall) rowConstructor;
```
LakeShen, line 4805: Does 0 mean anything else？If not,maybe remove 0 could make code clean.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
LakeShen, line 4782: boolean processDefaults = (insertOp != null) && containsDefault(values) maybe better?

WDYT?

LakeShen:
Thank you for opening this PR, @zstan overall looks good,I left a couple of comments!

core/src/main/java/org/apache/calcite/sql/SqlUtil.java
zstan: "eq" comparator is faster than instanceof, so i prefer this form

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
zstan: Merge sql syntax will also transform into SqlInsert operand and since i use it only for insertion - no other places are covered by now, thus no need to append it.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
zstan: plz suggest - what form do you prefer ? 
```
SqlCall rowConst = (SqlCall) rowConstructor;
SqlCall newRowConst = (SqlCall) rowConstructor;
```
as for me - no one of them not looks good.
i think usage of "0" is clear, it suggest other engineers that this is local scoped param, no ?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4769,19 +4778,53 @@ private void convertValuesImpl(
       return;
     }
 
-    for (SqlNode rowConstructor1 : values.getOperandList()) {
-      SqlCall rowConstructor = (SqlCall) rowConstructor1;
+    SqlCall insertOp = callStack.peek();
```
JiajunBernoulli, line 4781: We like final variable.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4769,19 +4778,53 @@ private void convertValuesImpl(
...
+    int[] mapping = null;
+
+    if (processDefaults) {
+      targetTable = getTargetTable(requireNonNull(insertOp, "insertOp"));
+      initializerFactory =
+          getInitializerFactory(getNamespace(requireNonNull(insertOp, "insertOp"))
+              .getTable());
+      mapping =
+          processColumnsMapping(requireNonNull(targetRowType, "targetRowType"), targetTable);
+    }
```
JiajunBernoulli, line 4803: Can we use new a private method that is called by convertValuesImpl? 
Maybe line that from 4782 to 4803 can be extract.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4769,19 +4778,53 @@ private void convertValuesImpl(
...
+
+    for (SqlNode rowConstructor : values.getOperandList()) {
+      SqlCall rowConstructor0 = (SqlCall) rowConstructor;
       Blackboard tmpBb = createBlackboard(bb.scope, null, false);
-      replaceSubQueries(tmpBb, rowConstructor,
-          RelOptUtil.Logic.TRUE_FALSE_UNKNOWN);
+      replaceSubQueries(tmpBb, rowConstructor, RelOptUtil.Logic.TRUE_FALSE_UNKNOWN);
```
JiajunBernoulli, line 4808: Try to keep the original code as much as possible, as this will affect the Git history.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
julianhyde: Makes sense, but I’d feel more confident if the test also tested a MERGE statement.

core/src/main/java/org/apache/calcite/sql/SqlUtil.java
LakeShen: Get your point.

You think in terms of performance, I think in terms of code style,I don't know the performance difference between them,but it is ok for me.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
LakeShen: Hi @zstan ,for me, when I see a 0 or 1 in the code, I usually assume that the 0 or 1 has a special meaning, it may represent the element index information, or in some cases the first or second element.

If I were to suggest, I would suggest the second,but this is just my advice and thoughts,It has no impact on this PR.

So It's ok for you to follow your ideas:)

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
zstan: EnumerableTableModify#implement is not implemented for MERGE, throws exception (
```
    switch (getOperation()) {
    case INSERT:
      method = BuiltInMethod.INTO.method;
      break;
    case DELETE:
      method = BuiltInMethod.REMOVE_ALL.method;
      break;
    default:
      throw new AssertionError(getOperation());
```

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
zstan: done, thnks !

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
zstan: @JiajunBernoulli catch it, but try to simple comment lines you suggested (just to understand what variables are involved) and as i can see - the code will become harder to read. So i prefer to store as is.

zstan:
@LakeShen thanks for review, i fix more of comments, probably it ready for merge ?

LakeShen (approved):

LakeShen:
@zstan LGTM.

Hi @JiajunBernoulli @julianhyde ,do you have any other suggestions？

PR #3394, opened 2023-08-23
[CALCITE-5914] Cache compiled regular expressions in SQL function runtime

julianhyde:


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -352,54 +358,81 @@ public static String sha512(ByteString string)  {
...
-      throw RESOURCE.invalidInputForRegexpContains(ex.getMessage().replace("\r\n", " ")
-          .replace("\n", " ").replace("\r", " ")).ex();
+  /** State for {@code REGEXP_CONTAINS}, {@code REGEXP_REPLACE}, {@code RLIKE}.
+   *
+   * <p>Marked deterministic so that the code generator instantiates one once
+   * per query, not once per row. */
+  @Deterministic
+  public static class RegexFunction {
+    private final LoadingCache<Ord<String>, Pattern> cache =
+        CacheBuilder.newBuilder().build(
```
chucheng92, line 368: Can we add maximumSize and expire strategy (add a extra config to control it). I'm asking this because in the production environment, users like to use functions such as regular expressions very much, and there will be many such functions in one query or parallel queries, the cache may increase very quickly.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -352,54 +358,81 @@ public static String sha512(ByteString string)  {
...
+
+    /** SQL {@code REGEXP_CONTAINS(value, regexp)} function.
+     * Throws a runtime exception for invalid regular expressions.*/
+    public boolean regexpContains(String value, String regex) {
+      final Pattern pattern;
+      try {
+        // Uses java.util.regex as a standard for regex processing
+        // in Calcite instead of RE2 used by BigQuery/GoogleSQL
+        pattern = cache.getUnchecked(Ord.of(0, regex));
+      } catch (UncheckedExecutionException e) {
```
chucheng92, line 386: It seems that only `regexpContains` catch this UncheckedExecutionException. other regex or like functions without this behavior. they throw pre-defined error msg in the parameters' checking.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4058,67 +4059,40 @@ private static class TranslateImplementor extends AbstractRexCallImplementor {
...
    */
   private static class ReflectiveImplementor extends AbstractRexCallImplementor {
-    protected final Method method;
+    protected final ImmutableList<? extends Method> methods;
 
-    ReflectiveImplementor(Method method, @Nullable NullPolicy nullPolicy) {
-      super("reflective_" + method.getName(), nullPolicy, false);
-      this.method = requireNonNull(method, "method");
+    ReflectiveImplementor(List<? extends Method> methods) {
+      super("reflective_" + methods.get(0).getName(), NullPolicy.STRICT, false);
```
chucheng92, line 4074: shall we check input methods list? otherwise it may cause NPE

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3694,14 +3693,13 @@ String getVariableName() {
       return variableName;
     }
 
-    @Nullable NullPolicy getNullPolicy() {
+    public NullPolicy getNullPolicy() {
```
chucheng92, line 3696: why add this `public`? IMHO, It should be a similar qualifier to other methods in this static class such as `getCondition`, `getVariableName`. 

and if we remove @Nullable (we expect to pass specific nonnull enumeration values), shall we add a `requireNonNull` checking in constructor? otherwise the below `unboxIfNecessary`method
```
switch (nullPolicy)
```
may cause NPE.

chucheng92:
Hi, @julianhyde function cache is a really useful optimization, and note that `RexImpTable`
many of the previous string methods have been unified into `BuiltInMethod.method`, which seems to be a very good refactor, neat and consistent.

only left a few comments.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
julianhyde: In a later commit I removed `class RegexFunction`, but the comments are still applicable.

The caches are within a single function object and are not static. Therefore the cache only holds for the duration of a particular query, there is a separate cache for each call site if a function is used multiple times in the same query.

I agree that we need a maximum size, in case a query of 100 million rows generates a unique regex for every row. But I don't think we need a time-based expiration strategy.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
julianhyde: This is the only one where I needed to catch the exception, because this is the only one that had negative tests (i.e. do we get a good error message if the regex is invalid).

If the functions do not have sufficient negative tests, I do not intend to fix that in this commit.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
julianhyde: The NPE is a sufficient check. This constructor is only called by Calcite developers. If they get an NPE or IndexOutOfBoundsException they will know what to do.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
julianhyde: Adding the `public` was a mistake. I will revert. (I considered moving `getNullPolicy` up to the interface, but it didn't work.)

Method parameters are implicitly not-nullable, so a `requireNotNull` is not necessary. We sometimes use `requireNonNull` in public APIs (so that users get a nicer exception) but this is not a public API.

julianhyde:
@chucheng92 Thank you for the thoughtful review.

Glad you liked the refactoring that moving things into `BuiltInMethod`. I started getting runtime errors when I started moving methods around, and it reminded me why we have `BuiltInMethod` - so that any invalid method references are discovered when the program loads.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
julianhyde: I'm adding `CalciteSystemProperty.FUNCTION_LEVEL_CACHE_MAX_SIZE`.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3160,97 +3293,176 @@ public static String timeWithLocalTimeZoneToString(int v, TimeZone timeZone) {
...
+    final DateFormatFunction f = new DateFormatFunction();
+
+    private final LoadingCache<Key, DateFormat> cache =
+        CacheBuilder.newBuilder()
+            .maximumSize(FUNCTION_LEVEL_CACHE_MAX_SIZE.value())
+            .build(CacheLoader.from(key -> key.toDateFormat(f)));
+
+    private <T> T withParser(String fmt, String timeZone,
+        Function<DateFormat, T> action) {
+      final DateFormat dateFormat = cache.getUnchecked(new Key(fmt, timeZone));
```
thomasrebele, line 3369: I'm just wondering whether this method can be called by several threads, as DateFormat is not thread-safe.

thomasrebele (commented):
Thanks for the implementation!

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
julianhyde: It's an implicit assumption that a function is only used on one thread. I'll make it explicit.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -1195,6 +1220,10 @@ private static <E> boolean allSame(List<E> list) {
     return true;
   }
 
+  private static boolean isStatic(Method method) {
+    return (method.getModifiers() & Modifier.STATIC) != 0;
+  }
```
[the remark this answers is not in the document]
julianhyde: Using `isStatic` is a good idea. Bit operations are error-prone.

PR #3399, opened 2023-08-28
[CALCITE-5962] Support parse Spark-style syntax LEFT ANTI JOIN in Babel parser

jiefei30:
[CALCITE-5962](https://issues.apache.org/jira/browse/CALCITE-5962) Allow Spark-style syntax "LEFT ANTI JOIN" in Babel parser

### DIFF
--- a/babel/src/main/codegen/config.fmpp
+++ b/babel/src/main/codegen/config.fmpp
@@ -42,6 +42,7 @@ data: {
     # List of new keywords. Example: "DATABASES", "TABLES". If the keyword is
     # not a reserved keyword, add it to the 'nonReservedKeywords' section.
     keywords: [
+      "ANTI"
       "DISCARD"
       "IF"
       "PLANS"
@@ -56,6 +57,7 @@ data: {
     # items in this list become non-reserved
     nonReservedKeywordsToAdd: [
       # not in core, added in babel
+      "ANTI"
       "DISCARD"
       "IF"
       "PLANS"
@@ -545,6 +547,7 @@ data: {
     # List of additional join types. Each is a method with no arguments.
     # Example: "LeftSemiJoin".
     joinTypes: [
+      "LeftAntiJoin",
       "LeftSemiJoin"
     ]
 
--- a/babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
+++ b/babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
@@ -18,6 +18,7 @@ package org.apache.calcite.test;
 import org.apache.calcite.sql.SqlDialect;
 import org.apache.calcite.sql.dialect.MysqlSqlDialect;
 import org.apache.calcite.sql.dialect.PostgresqlSqlDialect;
+import org.apache.calcite.sql.dialect.SparkSqlDialect;
 import org.apache.calcite.sql.parser.SqlAbstractParserImpl;
 import org.apache.calcite.sql.parser.SqlParser;
 import org.apache.calcite.sql.parser.SqlParserFixture;
@@ -428,6 +429,39 @@ class BabelParserTest extends SqlParserTest {
     f.sql("DISCARD SEQUENCES").same();
     f.sql("DISCARD TEMPORARY").same();
     f.sql("DISCARD TEMP").same();
+  }
+
+  @Test void testSparkLeftAntiJoin() {
+    final SqlParserFixture f = fixture().withDialect(SparkSqlDialect.DEFAULT);
+    final String sql = "select a.cid, a.cname, count(1) as amount\n"
+        + "from geo.area1 as a\n"
+        + "left anti join (select distinct cid, cname\n"
+        + "from geo.area2\n"
+        + "where cname = 'cityA') as b on a.cid = b.cid\n"
+        + "group by a.cid, a.cname";
+    final String expected = "SELECT A.CID, A.CNAME, COUNT(1) AMOUNT\n"
+        + "FROM GEO.AREA1 A\n"
+        + "LEFT ANTI JOIN (SELECT DISTINCT CID, CNAME\n"
+        + "FROM GEO.AREA2\n"
+        + "WHERE (CNAME = 'cityA')) B ON (A.CID = B.CID)\n"
+        + "GROUP BY A.CID, A.CNAME";
+    f.sql(sql).ok(expected);
+  }
+
+  @Test void testLeftAntiJoin() {
+    final String sql = "select a.cid, a.cname, count(1) as amount\n"
+        + "from geo.area1 as a\n"
+        + "left anti join (select distinct cid, cname\n"
+        + "from geo.area2\n"
+        + "where cname = 'cityA') as b on a.cid = b.cid\n"
+        + "group by a.cid, a.cname";
+    final String expected = "SELECT `A`.`CID`, `A`.`CNAME`, COUNT(1) AS `AMOUNT`\n"
+        + "FROM `GEO`.`AREA1` AS `A`\n"
+        + "LEFT ANTI JOIN (SELECT DISTINCT `CID`, `CNAME`\n"
+        + "FROM `GEO`.`AREA2`\n"
+        + "WHERE (`CNAME` = 'cityA')) AS `B` ON (`A`.`CID` = `B`.`CID`)\n"
+        + "GROUP BY `A`.`CID`, `A`.`CNAME`";
+    sql(sql).ok(expected);
   }
 
   /** Similar to {@link #testHoist()} but using custom parser. */
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -1988,8 +1988,8 @@ SqlLiteral JoinType() :
 }
 {
     (
-    LOOKAHEAD(3) // required for "LEFT SEMI JOIN" in Babel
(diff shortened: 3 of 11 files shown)


LakeShen:
@jiefei30 Thank you for opening this PR,could you add the unit tests to test your logic?

LakeShen:
also make the commits same as title

linorosa:
@jiefei30 Happy you're working on this, as I was looking for that feature myself. 

Just a question, doesn't `RelToSqlConverter` also warrants an update on `visitAntiOrSemiJoin`?  It looks like it's currently parsing `ANTI` joins as `NOT EXISTS`, but with your addition it could become a `LEFT_ANTI_JOIN` instead?

jiefei30:
> @jiefei30 Happy you're working on this, as I was looking for that feature myself.
> 
> Just a question, doesn't `RelToSqlConverter` also warrants an update on `visitAntiOrSemiJoin`? It looks like it's currently parsing `ANTI` joins as `NOT EXISTS`, but with your addition it could become a `LEFT_ANTI_JOIN` instead?

@linorosa Yeah, but I think the `LEFT ANTI JOIN` needs to be identified during the SQL parsing, it's a real syntax of Spark, otherwise it cannot be converted to RelNode, this is a bad case. And parsing `ANTI` joins as `NOT EXISTS` should be a matter of the semantic analysis stage, such as during a validation

NobiGo:
> 

Same comment.

NobiGo:
@jiefei30 If we only support parse this syntax, Maybe we should use `Support parse Spark-style syntax LEFT ANTI JOIN in Babel parser` . Also need to add parse test in PR.

jiefei30:
> @jiefei30 If we only support parse this syntax, Maybe we should use `Support parse Spark-style syntax LEFT ANTI JOIN in Babel parser` . Also need to add parse test in PR.

@NobiGo ok, thanks for your reply. I've already added parse test in it . And squash them into a single commit.

macroguo-ghy:
Hi @jiefei30, I think you should consider this PR[https://github.com/apache/calcite/pull/3291], LEFT ANTI JOIN should not refer right-hand table.

jiefei30:
> Hi @jiefei30, I think you should consider this PR[#3291], LEFT ANTI JOIN should not refer right-hand table.

@macroguo-ghy ok, thanks for your suggestion, I'll check it

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1988,8 +1988,8 @@ SqlLiteral JoinType() :
 }
 {
     (
-    LOOKAHEAD(3) // required for "LEFT SEMI JOIN" in Babel
```
JiajunBernoulli, line 1991: Why move it?

jiefei30: In fact, I also have doubts about this `LOOKAHEAD(3)`, because `LEFT ANTI JOIN` also requires it, otherwise it will recognize errors, but `LOOKAHEAD(3)` is outside `<#list>`, so `joinType = ${method} ()` will generate  `LOOKAHEAD(2)` except for the first one `LEFT SEMI JOIN`. Give me some time to think about it and  reply to you as soon as possible

JiajunBernoulli: You can execute `javacc core/build/fmpp/fmppMain/javacc/Parser.jj`.

The result should be
```
$ javacc core/build/fmpp/fmppMain/javacc/Parser.jj
Java Compiler Compiler Version 7.0.9 (Parser Generator)
(type "javacc" with no arguments for help)
Reading from file core/build/fmpp/fmppMain/javacc/Parser.jj . . .
Note: UNICODE_INPUT option is specified. Please make sure you create the parser/lexer using a Reader with the correct character encoding.
File "TokenMgrError.java" does not exist.  Will create one.
File "ParseException.java" does not exist.  Will create one.
File "Token.java" does not exist.  Will create one.
File "SimpleCharStream.java" does not exist.  Will create one.
Parser generated successfully.
```
If you find warning, you should solve it.

jiefei30: ok, thanks for your suggestion. I'll check it

jiefei30: @JiajunBernoulli I just execute `javacc core/build/fmpp/fmppMain/javacc/Parser.jj`. 
The execution results are as expected : 
```shell
Java Compiler Compiler Version 7.0.10 (Parser Generator)
(type "javacc" with no arguments for help)
Reading from file core/build/fmpp/fmppMain/javacc/Parser.jj . . .
Note: UNICODE_INPUT option is specified. Please make sure you create the parser/lexer using a Reader with the correct character encoding.
File "TokenMgrError.java" does not exist.  Will create one.
File "ParseException.java" does not exist.  Will create one.
File "Token.java" does not exist.  Will create one.
File "SimpleCharStream.java" does not exist.  Will create one.
Parser generated successfully.
```
There is no warning

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -430,6 +431,23 @@ private void checkParseInfixCast(String sqlType) {
...
+        + "from geo.area2\n"
+        + "where cname = 'cityA') as b on a.cid = b.cid\n"
+        + "group by a.cid, a.cname";
+    final String expected = "SELECT A.CID, A.CNAME, COUNT(1) AMOUNT\n"
+        + "FROM GEO.AREA1 A\n"
+        + "LEFT ANTI JOIN (SELECT DISTINCT CID, CNAME\n"
+        + "FROM GEO.AREA2\n"
+        + "WHERE (CNAME = 'cityA')) B ON (A.CID = B.CID)\n"
+        + "GROUP BY A.CID, A.CNAME";
+    f.sql(sql).ok(expected);
```
JiajunBernoulli, line 448: How about other SqlDialect?

babel/src/main/codegen/config.fmpp
```diff
@@ -49,7 +49,8 @@ data: {
       "SEMI"
       "SEQUENCES"
       "TEMP"
-      "VOLATILE"
+      "VOLATILE",
+      "ANTI"
```
JiajunBernoulli, line 53: Please keep alphabetic order.

babel/src/main/codegen/config.fmpp
```diff
@@ -62,7 +63,8 @@ data: {
       "SEED"
       "SEMI"
       "SEQUENCES"
-      "TEMP"
+      "TEMP",
```
JiajunBernoulli, line 66: Please keep alphabetic order.

babel/src/main/codegen/config.fmpp
```diff
@@ -545,7 +547,8 @@ data: {
     # List of additional join types. Each is a method with no arguments.
     # Example: "LeftSemiJoin".
     joinTypes: [
-      "LeftSemiJoin"
+      "LeftSemiJoin",
+      "LeftAntiJoin"
```
JiajunBernoulli, line 551: Please keep alphabetic order.

babel/src/main/codegen/config.fmpp
jiefei30: already fixed

babel/src/main/codegen/config.fmpp
jiefei30: already fixed

babel/src/main/codegen/config.fmpp
jiefei30: already fixed

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
jiefei30: According to [antlr/grammer](https://github.com/antlr/grammars-v4/tree/master/sql/hive) and [apache/hive](https://github.com/apache/hive/blob/master/parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g), `LEFT ANTI JOIN`  is also supported in `hive4`, but the `hive2` and `hive3` are not supported. Should we think that hive also supports this syntax?

JiajunBernoulli: This is arbitrary, but unit test needs to demonstrate to demonstrate.

JiajunBernoulli: @jiefei30 , If we not use `withDialect(SparkSqlDialect.DEFAULT)`.

We also can unparse `LEFT ANTI JOIN`.

Just a unit test, I will merge it.

jiefei30: @JiajunBernoulli  Ok. Thank you for your review during this time. I will squash this changes to a single commit

LakeShen:
Hi @JiajunBernoulli ,this pr workflows is awaiting approval,could you approve this?

JiajunBernoulli:
@jiefei30 
1. Your message is duplicated.
image

2. Please add one unit test that is not SparkSqlDialect in BabelParserTest.

JiajunBernoulli (approved):

jiefei30:
> All checks have passed

jiefei30:
@JiajunBernoulli  Sorry, I just accidentally closed this pr ccause I pressed the wrong button. I opened it again. I added a unit test of `leftAntiJoin` with `SqlParserFixture.DEFAULT` and rewirte commit message. All checks have passed at last time.

PR #3403, opened 2023-08-29
[CALCITE-5967] UnsupportedOperationException while implementing a call that requires a special collator

rubenada:
Jira: CALCITE-5967

Fix regression introduced as a side effect of CALCITE-5914


thomasrebele:
LGTM

PR #3408, opened 2023-09-05
[CALCITE-5978] Add REGEXP_INSTR function (enabled in BigQuery library)

Anthrino:


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -404,16 +404,39 @@ private Pattern validateRegexPattern(String regex, String methodName) {
...
+      }
+      if (occurrence <= 0) {
+        throw RESOURCE.invalidIntegerInputForRegexpFunctions(Integer.toString(occurrence),
+            "occurrence", methodName).ex();
+      }
+      if (occurrencePosition < 0 || occurrencePosition > 1) {
+        throw RESOURCE.invalidIntegerInputForRegexpFunctions(Integer.toString(occurrencePosition),
+            "occurrence_position", methodName).ex();
+      }
+      if (position > value.length()) {
```
tanclary, line 432: can this be simplified to `return position <= value.length()`?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -404,16 +404,39 @@ private Pattern validateRegexPattern(String regex, String methodName) {
...
+        return true;
+      }
+      return false;
+    }
+
     /** SQL {@code REGEXP_CONTAINS(value, regexp)} function.
-     * Throws a runtime exception for invalid regular expressions.*/
+     *  Throws a runtime exception for invalid regular expressions. */
```
tanclary, line 439: Did you mean to add this?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -488,13 +502,74 @@ public List<String> regexpExtractAll(String value, String regex) {
...
           matches.add(match);
         }
       }
       return matches.build();
     }
 
+    /** SQL {@code REGEXP_INSTR(value, regexp)} function.
+     *  Returns 0 if there is no match or regex is empty. Returns an exception if regex is invalid.
+     *  Uses position=1, occurrence=1, occurrencePosition=0 as default values if not specified. */
+    public Integer regexpInstr(String value, String regex) {
```
tanclary, line 515: should we use `int` in favor of `Integer`? It seems that's what used in most other cases.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -488,13 +502,74 @@ public List<String> regexpExtractAll(String value, String regex) {
...
+    /** SQL {@code REGEXP_INSTR(value, regexp, position)} function.
+     *  Returns 0 if there is no match, regex is empty, or if position is beyond range.
+     *  Returns an exception if regex or position is invalid.
+     *  Uses occurrence=1, occurrencePosition=0 as default value when not specified. */
+    public Integer regexpInstr(String value, String regex, int position) {
+      return regexpInstr(value, regex, position, 1, 0);
+    }
+
+    /** SQL {@code REGEXP_INSTR(value, regexp, position, occurrence)} function.
+     *  Returns NULL if there is no match, regex is empty, or if position or occurrence
```
tanclary, line 528: Is it correct that this one should return null if there is no match while others return 0? Just checking

tanclary (commented):
Nice one 👍 left a few minor comments

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
Anthrino: My bad missed on correcting this instance, should be aligned with the others thanks for catching it!

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
Anthrino: kept it this way since thats the failing condition, can be inverted too is that to make it more intuitive with return values true/false or for some other reason?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
Anthrino: yes just to align this line with the ones above

tanclary (commented):
left a couple nits, otherwise LGTM

tanclary (approved):
LGTM, I have been looking at a lot of these REGEX PRs and they start to blend together a bit. Going to give it a day or so and if there are no other comments I will merge.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -404,56 +405,70 @@ private Pattern validateRegexPattern(String regex, String methodName) {
...
+    /** Helper for multiple capturing group regex check in REGEXP_* fns. */
     private void checkMultipleCapturingGroupsInRegex(Matcher matcher, String methodName) {
       if (matcher.groupCount() > 1) {
         throw RESOURCE.multipleCapturingGroupsForRegexpExtract(
             Integer.toString(matcher.groupCount()), methodName).ex();
       }
     }
 
+    /** Helper for checking values of position and occurrence arguments in REGEXP_* fns.
+     * Regex Fns not using occurrencePosition param pass a default value as 0.
```
olivrlee, line 417: lowercase `fns` to keep consistent

olivrlee: `default value of 0`

tanclary (approved):
Thanks for the PR 👍

PR #3411, opened 2023-09-08
[CALCITE-5986] The typeFamily property of SqlTypeName is used inconsistently

mihaibudiu:
I have discovered that the notion of typeFamily for a type was ambiguous. There are two kinds of type families: primary families, and secondary families, as described in the JavaDoc [`core/src/main/java/org/apache/calcite/sql/type/SqlTypeFamily.java`](https://github.com/apache/calcite/blob/3aee0b86aa23476cbdecc75ad5d43b936a6fff7b/core/src/main/java/org/apache/calcite/sql/type/SqlTypeFamily.java#L38). Both kinds are bundled in the same Enum. I have added a Boolean flag to the enum which will enable us in the future to check for misuses. The `SqlTypeName` class used them interchangeably. 

With this change each `SqlTypeName` has both a primary and a secondary family. For backwards compatibility, the `getFamily()` function keeps its original behavior. But each field can be accessed independently, and ideally the original function should be eventually deprecated.

This confusion led to at least one bug in RelToSqlConverter, which would generate DECIMAL literals instead of DOUBLE literals. That bug has been fixed in this PR as well. There is one test case for Piglet which has a changed output, so I haven't added a new test.

The confusion described may have consequences (i.e. bugs) elsewhere in the codebase, but this will require auditing for other uses of getTypeFamily(). (There aren't too many.)

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
@@ -612,7 +612,7 @@ public class RexToLixTranslator implements RexVisitor<RexToLixTranslator.Result>
     case INTERVAL_MINUTE_SECOND:
     case INTERVAL_SECOND:
       final SqlTypeFamily family =
-          requireNonNull(sourceType.getSqlTypeName().getFamily(),
+          requireNonNull(sourceType.getSqlTypeName().getPrimaryOrSecondaryFamily(),
               () -> "null SqlTypeFamily for " + sourceType + ", SqlTypeName "
                   + sourceType.getSqlTypeName());
       switch (family) {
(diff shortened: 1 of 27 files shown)


mihaibudiu:
@julianhyde This solution seems relatively painless.

mihaibudiu:
I have eliminated some uses of the SqlTypeName.getTypeFamily() method, but there are quite a few left.
Removing the remaining ones will require a deeper understanding of the code than I have. I could not deprecate the function for this reason.

mihaibudiu:
I realize that this is a more intrusive change, but I think it does make a small step in improving the use to type families.
This PR also fixes a bug introduced by the incorrect use of type families.
I will have submitted #3472 with just that fix, it illustrates why the type families are wrong.

mihaibudiu:
I plan to close this PR if people are not interested in this approach to cleanup the meaning of type families.
The bug related to this was fixed in a separate PR.

julianhyde:
Not uninterested. Just reviewer fatigue. I've made one pass and left detailed comments (in Jira), some of which may be mistaken (because I don't have as much time and energy to study the problem). If you think you're addressed the comments that are not mistaken, and you think this is a positive change, go ahead and merge.

PR #3414, opened 2023-09-10
[CALCITE-5991] Reduce the number of tasks related to AbstractConverter in TopDownRuleDriver

asdfgh19:
jira: [CALCITE-5991](https://issues.apache.org/jira/browse/CALCITE-5991)

This PR is based on the following facts.
1. AbstractConverter is not an instance of PhysicalNode. So we don't have to try to passThrough or derive Trait on it.
2. AbstractConverter must only exist in the root RelSet, and its TraitSet must be equal to the TraitSet of the root RelSubset group.


LakeShen:
Could you add the unit test for you logic?

PR #3415, opened 2023-09-10
[CALCITE-5992] Avoid comparing root Relsubset's TraitSet with itself

asdfgh19:
jira: [CALCITE-5992](https://issues.apache.org/jira/browse/CALCITE-5992)

Since VolcanoPlanner#root will definitely exist in the subsets collection of the Relset related to it, we can skip the comparison between the root Relsubset and itself.


core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
```diff
@@ -565,6 +565,9 @@ void ensureRootConverters() {
       }
     }
     for (RelSubset subset : root.set.subsets) {
+      if (subset == root) {
```
LakeShen, line 568: I'm not sure whether the logic of these lines of code is correct, could you add a unit test for this logic?

PR #3416, opened 2023-09-10
[CALCITE-4189] Simplify 'p OR (p IS NOT TRUE)' to 'TRUE'

macroguo-ghy:
Simplify 'p OR NOT p' to 'TRUE' (if p is not nullable), or to 'null AND p IS NOT NULL' (if p is nullable).
GeodeFilter support converting 'p IS NOT NULL' to 'p <> null'.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexSimplify.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexSimplify.java
@@ -2060,6 +2060,26 @@ public class RexSimplify {
           }
         }
         break;
+      case IS_NOT_TRUE:
+        RexNode arg = ((RexCall) term).getOperands().get(0);
+        if (isSafeExpression(arg) && terms.contains(arg)) {
+          return trueLiteral;
+        }
+        break;
+      case NOT:
+        RexNode x = ((RexCall) term).getOperands().get(0);
+        if (isSafeExpression(x) && terms.contains(x)) {
+          if (!x.getType().isNullable()) {
+            return trueLiteral;
+          }
+
+          final RexNode isNotNull =
+              rexBuilder.makeCall(SqlStdOperatorTable.IS_NOT_NULL, x);
+          terms.set(terms.indexOf(x), simplifyIs((RexCall) isNotNull, unknownAs));
+          terms.set(i, rexBuilder.makeNullLiteral(x.getType()));
+          i--;
+        }
+        break;
       default:
         break;
       }
--- a/core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
+++ b/core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
@@ -2318,7 +2318,7 @@ class RexProgramTest extends RexProgramTestBase {
             isTrue(vBool()), literal(1),
             isNotTrue(vBool()), literal(1),
             literal(2)),
-        "CASE(OR(?0.bool0, IS NOT TRUE(?0.bool0)), 1, 2)");
+        "1");
   }
 
   @Test void testSimplifyCaseBranchesCollapse2() {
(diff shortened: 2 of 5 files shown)


core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -1993,6 +1993,7 @@ private RexNode simplifyOrs(List<RexNode> terms, RexUnknownAs unknownAs) {
     // or 'X IS NOT NULL' if a simplification takes place (because another 'X<>B' is found)
     final Map<RexNode, RexNode> notEqualsComparisonMap = new HashMap<>();
     final RexLiteral trueLiteral = rexBuilder.makeLiteral(true);
+    boolean removeNull = false;
```
thomasrebele, line 1996: I would rather use a flag `hasNullTerm`, and also set it in the `case LITERAL` if the case does not remove it.

macroguo-ghy: I remove this logic because I found `simplifyOrs` method will call `flattenOr` eventually which use a `set` to eliminate duplicates. So we don't need to handle multiple `NULL` literal separately.

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -2060,6 +2060,25 @@ private RexNode simplifyOrs(List<RexNode> terms, RexUnknownAs unknownAs) {
           }
         }
         break;
+      case IS_NOT_TRUE:
+        if (terms.contains(((RexCall) term).getOperands().get(0))) {
+          return trueLiteral;
```
kgyrtkirk, line 2065: I'm not sure about this; I think safety should still be considered:
```
1/0 or x or (x is not true) 
```
I think even if its okay for `x` ; we can't replace it with `true` unless all other parts of the or passes `isSafeExpression` or something similar

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -2060,6 +2060,25 @@ private RexNode simplifyOrs(List<RexNode> terms, RexUnknownAs unknownAs) {
...
+        if (terms.contains(x)) {
+          if (!x.getType().isNullable()) {
+            return trueLiteral;
+          }
+
+          final RexNode isNotNull =
+              rexBuilder.makeCall(SqlStdOperatorTable.IS_NOT_NULL, x);
+          terms.set(terms.indexOf(x), simplifyIs((RexCall) isNotNull, unknownAs));
+          terms.set(i, rexBuilder.makeNullLiteral(x.getType()));
+          i--;
```
kgyrtkirk, line 2079: I don't see the reason for this `i--` why its needed?

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -2060,6 +2060,25 @@ private RexNode simplifyOrs(List<RexNode> terms, RexUnknownAs unknownAs) {
...
         }
         break;
+      case IS_NOT_TRUE:
+        if (terms.contains(((RexCall) term).getOperands().get(0))) {
+          return trueLiteral;
+        }
+        break;
+      case NOT:
+        RexNode x = ((RexCall) term).getOperands().get(0);
+        if (terms.contains(x)) {
```
kgyrtkirk, line 2070: I think to proceed in both cases `x` must be `isSafeExpression` to remove them

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -2318,7 +2318,7 @@ trueLiteral, literal(2),
             isTrue(vBool()), literal(1),
             isNotTrue(vBool()), literal(1),
             literal(2)),
-        "CASE(OR(?0.bool0, IS NOT TRUE(?0.bool0)), 1, 2)");
+        "1");
```
kgyrtkirk, line 2321: :cowboy_hat_face:

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
macroguo-ghy: @kgyrtkirk Thanks for your review! 
All operands in `OR` operator must be `boolean` type (base on the operand checker), `x` could be `1 / 0 <> 0` but could not be `1 / 0`. So I think wo don't need to care about if `x` safe or not. WDYT?

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
macroguo-ghy: Same viewpoint as my other comment.

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
kgyrtkirk: you've replied to that comment here..that's not great - because I think these are 2 separate issues - but they are somewhat similar.

`1/0` is not a boolean - my mistake; I just wanted to illustrate the problem quickly

Don't expect that `1 / 0` will appear as is ... it will be more likely `x / y > 17` which is a boolean; can't be decided at compile time that it will never or always throw exceptions - but still holds the possiblity of an exception because if `y` is `0` it will emit an exception.

I think the current patch will  remove and simplify unknown udfs away - we don't know for sure that  omitting those calls remove sideeffects or not.

I think
```
someUdf() or someUdf() is not true  ==> true
```
is wrong because it could be a possibly unsafe method - you may not make such transformation; or at least call that method - but I would rather stay on the safe side and not do that either...

and also
```
someUdf() or x or !x ==> true (in case x is not nullable)
```
almost the same as above - however in this case the unsafe expr is outside

In the above cases: what if `someUdf` throws an exception on every invocation? should the query be executed - because we were able to deduce that its return value is irrelevant?


Could you please add some illustrative testcases for the above cases?

kgyrtkirk (changes requested):

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
julianhyde: My position on safety is that only CASE has a defined evaluation order. Though we've never really implemented it.

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
macroguo-ghy: Thanks for your explaination, I got your points. 
For the first case `x or not(x)`, I add a new limitation where `x` must be safeExpression.
But for the second case `a or x or not(x) or b`, I think we can simplify `x or not(x)` no matter whether `a` and `b` safe or not.
Thinking about a simple case: `true or unSafeUdf()`,  calcite will simplify this expression to `true`. And I have tested `select * from t where 1/0 or true` expression in mysql and sparksql, both of them will simpilfy this expression to `true`. And third, in most of programing language (like Java, C++, Python), they always simplify `true or someConditions` to `true` in if clause (A little different, it is Short-circuit OR actully, but we check all items).
So I think we don't need to care about outside items whether safe or not.

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
macroguo-ghy: Thanks for your explaination, I got your points. 
For the first case `x or not(x)`, I add a new limitation where `x` must be safeExpression.
But for the second case `a or x or not(x) or b`, I think we can simplify x or not(x) no matter whether `a` and `b` safe or not.
Thinking about a simple case: `true or unSafeUdf()`,  calcite will simplify this expression to `true`. And I have tested `select * from t where 1/0 or true` expression in mysql and sparksql, both of them will simpilfy this expression to `true`. And third, in most of programing language(like Java, C++, Python), they always simplify `true or someConditions` to `true` in if clause (A little different, it is Short-circuit OR actully, but we check all items).
So I think we don't need to care about outside items whether safe or not.

macroguo-ghy:
Hi, @kgyrtkirk. Sorry for late of response, I was a little busy this week. 
I did some changes, would you like to review this PR again?

geode/src/test/java/org/apache/calcite/adapter/geode/rel/GeodeAllDataTypesTest.java
```diff
@@ -161,7 +161,7 @@ private CalciteAssert.AssertThat calciteAssert() {
             GeodeAssertions.query("SELECT stringValue AS stringValue "
                 + "FROM /allDataTypesRegion WHERE "
                 + "stringValue IN SET('abc', 'def') OR floatValue = 1.5678 "
-                + "OR booleanValue <> null"));
+                + "OR booleanValue = true OR booleanValue = false"));
```
kgyrtkirk: I wonder why this is not getting rewritten - 
why `booleanValue` isn't considered safe? (it seems like a column to me - isn't it)

kgyrtkirk (approved):

kgyrtkirk:
I don't want to hold you back - this is really usefull; I'm a bit sad to see that the simplification opportunity was missed in the geode adapter ... but the goal here was to handle `p || !p` - which looks good!

geode/src/test/java/org/apache/calcite/adapter/geode/rel/GeodeAllDataTypesTest.java
macroguo-ghy: In Geodo adapter, all types are represented as JavaType, so all refs will be wrapped with `CAST`.
However, `CAST` is not a safe expression as we cannot guarantee the success or failure of the cast in some cases.
For example:
```sql
cast('2023-03-23' as date) -- success
cast('aaaa' as date) -- fail
```
Both of these statements attempt to cast a varchar as a date, but the first one is legal while the second one will fail.

But I am not familiar with geodo adapters, so I am not very clear why all types are using Java types, I only discovered this during debugging.

macroguo-ghy:
Hi @kgyrtkirk,  could you please review this PR again or merge it?

PR #3417, opened 2023-09-10
[CALCITE-5993] Add CODE_POINTS_TO_STRING, TO_CODE_POINTS function (en…

macroguo-ghy:
…abled in BigQuery library)


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1130,14 +1130,68 @@ public static String charFromUtf8(int n) {
       assert codePoint instanceof Number;
       long cp = ((Number) codePoint).longValue();
       if (cp < 0 || cp > 255) {
-        throw RESOURCE.inputArgumentsOfCodePointsToBytesOutOfRange(cp).ex();
+        throw RESOURCE.inputArgumentsOfFunctionOutOfRange("CODE_POINTS_TO_BYTES", cp).ex();
```
tanclary, line 1133: Would you mind checking if there are other functions that have a similar error (input out of range?)? I think maybe some of the recently added regex functions have a similar error. It could be a good idea to reduce some redundancy if that's the case.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1130,14 +1130,68 @@ public static String charFromUtf8(int n) {
...
+    while (i < length) {
+      int cp = s.codePointAt(i);
+      builder.add(cp);
+      i += cp == s.charAt(i) ? 1 : 2;
+    }
+    return builder.build();
+  }
+
+  /**
+   * SQL TO_CODE_POINTS(binary) function.
```
tanclary, line 1182: nit: could you check if these javadocs are consistent with other functions that have two signatures (one for strings, one for binary strings)?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1793,6 +1793,22 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
           OperandTypes.ARRAY_OF_INTEGER,
           SqlFunctionCategory.STRING);
 
+  @LibraryOperator(libraries = {BIG_QUERY})
```
tanclary, line 1796: Nit: would you mind adding javadocs here? I know not all of the functions have them but given how many functions we've been adding to Calcite recently I think it would be a good practice to have them well-commented.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1793,6 +1793,22 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
+  public static final SqlFunction CODE_POINTS_TO_STRING =
+      SqlBasicFunction.create("CODE_POINTS_TO_STRING",
+          ReturnTypes.VARCHAR_NULLABLE,
+          OperandTypes.ARRAY_OF_INTEGER,
+          SqlFunctionCategory.STRING);
+
+  @LibraryOperator(libraries = {BIG_QUERY})
+  public static final SqlFunction TO_CODE_POINTS =
+      SqlBasicFunction.create("TO_CODE_POINTS",
+          ReturnTypes.INTEGER
```
tanclary, line 1806: Just a question not a comment: I think the answer is yes, but is there a reason you can't use INTEGER_NULLABLE and then transform it to array? is it due to the ordering?

tanclary (changes requested):
Left a few comments, thanks for opening this PR

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
macroguo-ghy: I replace it with `SQL TO_CODE_POINTS(string) function for binary string.`

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
macroguo-ghy: Add some comments for three `code_points` functions.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
macroguo-ghy: The order of transform will affect the return types. 
`INTEGER_NULLABLE.andThen(SqlTypeTransforms.TO_ARRAY)` means: an not null array, which element may be null.
`INTEGER.andThen(SqlTypeTransforms.TO_ARRAY).andThen(SqlTypeTransforms.TO_NULLABLE)` means: an array may be null, but its element is not null.

In `to_code_points` function, we can not get a result like`[1, NULL, 2, 3]`, the element type of array never be `NULL`, so i use `INTEGER` instead of `INTEGER_NULLABLE`.

macroguo-ghy:
Hi @tanclary, Thanks for your review, and sorry for late response, I was a litter busy this week. I do some change and add some comments, would you like to review this PR again?

macroguo-ghy (commented):
Hi @tanclary, Thanks for your review, and sorry for late response, I was a litter busy this week. I do some change and add some comments, would you like to review this PR again?

tanclary:
@macroguo-ghy No worries, thanks for responding to the comments. I don't have any other suggestions so if you squash your commits I will merge soon assuming no one else has any feedback.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1130,14 +1130,70 @@ public static String charFromUtf8(int n) {
...
+            "CODE_POINTS_TO_BYTES", cp, "[0, 255]").ex();
       }
       bytes[i] = (byte) cp;
     }
 
     return new ByteString(bytes);
   }
 
+  /**
+   * SQL CODE_POINTS_TO_STRING function.
```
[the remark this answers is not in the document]
macroguo-ghy: Fixed.

babel/src/test/resources/sql/big-query.iq
```diff
@@ -2029,6 +2029,160 @@ SELECT CODE_POINTS_TO_BYTES(array[2147483648, 1]);
...
++--------+
+(1 row)
+
+!ok
+
+SELECT CODE_POINTS_TO_STRING(array[1+2, 3]) as result;
++--------+
+| result |
++--------+
+|      |
```
tanclary, line 2062: Why does the spacing look off here, do you know?

tanclary (changes requested):

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
macroguo-ghy: Good suggestion! I add this `SqlTypeTransforms`  as you recommended. But a litter different from #3430, because in #3430, `TO_ARRAY_NULLABLE`  is defined as 
`ARG0`
`.andThen(SqlTypeTransforms.TO_ARRAY)`
`.andThen(SqlTypeTransforms.TO_NULLABLE)`, 
which can not be used in my PR.

macroguo-ghy:
Hi @tanclary, could you please review this PR again if you have time?

macroguo-ghy:
Hi @tanclary, could you please review this PR again when you have time? I believe we can merge it before the 1.36.0 release.

tanclary (approved):

PR #3421, opened 2023-09-11
[CALCITE-5982] Allow overloading the created enumerable in CalciteMetaImpl when calling getTables() or getColumns()

olivrlee:


core/src/main/java/org/apache/calcite/jdbc/CalciteMetaImpl.java
```diff
@@ -263,23 +288,30 @@ private static ImmutableMap.Builder<DatabaseProperty, Object> addProperty(
...
+        this.metaTableClass,
+        columnNames);
+  }
+
+  /** The provided subclass needs to overload getColumnNames() with the expected columns in the
+   * enumerable.
+   * */
+  private String[] getColumnNames(Class<?> clazz) {
+    try {
+      Method m = clazz.getMethod("getColumnNames");
```
wnob, line 306: No need for dynamic method lookup with the type changes.

olivrlee: I tried this modification: `getColumnNames(Class<? extends CalciteMetaTable> clazz){...}`  however it that `clazz` object still doesn't have direct access to the interface function getColumnNames()


Also, if this change worked, I would need 2 functions for this, one for `extends CalciteMetaTable` and one for `extends MetaColumn`
Thoughts?

core/src/main/java/org/apache/calcite/schema/Table.java
```diff
@@ -82,4 +85,20 @@ public interface Table {
...
+   */
+  default Map<String, Object> getTableMetadata() {
+    Map<String, Object> map = new HashMap<>();
+    return map;
+  }
+
+  /**
+   * Overload to provide additional column metadata if available.
+   */
+  default Map<String, Map<String, Object>> getColumnMetadata() {
```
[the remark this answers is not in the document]
olivrlee: Yeah I agree an untyped map is not great. 

Would you say this modeling is still required within Calcite, if I go ahead with the factory interface?
Calcite in this case would not need to be aware of metadata at all. The Calcite factory implementation would just create `CalciteMetaTable`s

When a user of Calcite implements the factory interface,  the implementation would then be responsible for how the metadata is accessed, enumerated, typed, etc. and maybe that modeling should sit there.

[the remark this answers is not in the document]
barrkel: I think, if a factory interface removes the need for the loosely typed map, then don't add the map at all.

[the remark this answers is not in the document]
olivrlee: Yeah I like the factory approach, thanks for the review and suggestion.

I pushed a commit with the rough idea but there are still some tests to update/ and I need to fix the plugin retrieval, so it's not ready for re-review yet.

olivrlee:
@barrkel  Thanks for the factory suggestion, I have cleaned up this PR to use factories and updated the tests. Please take a look when you have time!

core/src/main/java/org/apache/calcite/config/CalciteConnectionConfig.java
```diff
@@ -110,4 +110,16 @@ public interface CalciteConnectionConfig extends ConnectionConfig {
   boolean lenientOperatorLookup();
```
tanclary, line 110: nit: remove "/" from PR title?

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaColumnFactoryImpl.java
```diff
@@ -0,0 +1,98 @@
...
+      Integer columnSize,
+      Integer decimalDigits,
+      Integer numPrecRadix,
+      int nullable,
+      Integer charOctetLength,
+      int ordinalPosition,
+      String isNullable,
+      String isAutoincrement,
+      String isGeneratedcolumn) {
+    return new MetaColumn(
```
tanclary, line 73: Not sure if outdated but looks like there are some constructor errors ?

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaImpl.java
```diff
@@ -90,13 +90,34 @@
...
+      @Nullable CalciteMetaTableFactory metaTableFactory,
+      @Nullable CalciteMetaColumnFactory metaColumnFactory) {
     super(connection);
     this.connProps
         .setAutoCommit(false)
         .setReadOnly(false)
         .setTransactionIsolation(Connection.TRANSACTION_NONE);
     this.connProps.setDirty(false);
+
+    if (metaTableFactory != null) {
```
tanclary, line 111: Anytime I see this kind of `if` block I wish Java had the elvis operator

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaImpl.java
```diff
@@ -263,23 +289,16 @@ private static ImmutableMap.Builder<DatabaseProperty, Object> addProperty(
...
+    Enumerable<MetaTable> tables = schemas(catalog)
+        .where(schemaMatcher)
+        .selectMany(schema -> tables(schema, matcher(tableNamePattern)))
+        .where(typeFilter);
+    return createResultSet(tables,
+        metaTableFactory.getMetaTableClass(),
+        metaTableFactory.getColumnNames());
   }
 
+
```
tanclary, line 301: nit: remove blank line

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaImpl.java
```diff
@@ -378,6 +375,7 @@ Enumerable<MetaTable> tables(String catalog) {
             tables(schema, Functions.<String>truePredicate1()));
   }
 
+
```
tanclary, line 378: same as above

core/src/test/java/org/apache/calcite/test/JdbcFrontJdbcBackTest.java
```diff
@@ -73,6 +79,72 @@ class JdbcFrontJdbcBackTest {
...
+        .with(
+            CalciteConnectionProperty.META_TABLE_FACTORY.camelName(),
+            MetaExtraTableFactoryImpl.class.getName())
+        .doWithConnection(connection -> {
+          try {
+            ResultSet rset =
+                connection.getMetaData().getTables(
+                    null, null, null, null);
+            assertTrue(rset.next());
+            assertEquals(rset.getMetaData().getColumnCount(), 11);
```
tanclary, line 139: Clarification: where do these expected results come from ?

tanclary (commented):
I have limited knowledge on this but I think it looks great : )

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaColumnFactoryImpl.java
olivrlee: There's something I need to update, seems like Avatica main took out the last 2 parameters in the constructor. I haven't rebased my Avatica snapshot off of main, I'll do that

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaImpl.java
olivrlee: Indeed

tanclary:
I think some of the code smells on this might be helpful suggestions. Like marking some of the variables as final or protected. Of course maybe they're irrelevant.

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaColumnFactory.java
```diff
@@ -0,0 +1,48 @@
...
+      Integer columnSize,
+      @Nullable Integer decimalDigits,
+      Integer numPrecRadix,
+      int nullable,
+      Integer charOctetLength,
+      int ordinalPosition,
+      String isNullable);
+
+  /** Returns the list of expected column names. */
+  String[] getColumnNames();
```
julianhyde, line 44: Let's make this `List<String>`. Then you can return an `ImmutableList` and don't have to worry about making a defensive copy.

I'd give it a default implementation that returns the columns in the JDBC spec, and add the javadoc 'The default implementation returns the columns in the JDBC specification.'

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaColumnFactory.java
```diff
@@ -0,0 +1,48 @@
...
+
+  /** Returns the class of MetaColumn that is created. */
+  Class<?> getMetaColumnClass();
```
julianhyde, line 47: 'Returns the type of object created. Must be a sub-class of `MetaColumn`.'

Does it work if you change the return type from `Class<?>` to `Class<? extends MetaColumn>`?

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaImpl.java
```diff
@@ -90,13 +90,40 @@
...
+   * @deprecated Use
+   * {@link #CalciteMetaImpl(CalciteConnectionImpl, CalciteMetaTableFactory, CalciteMetaColumnFactory)}
+   * instead.
+   */
+  @Deprecated // to be removed before 2.0
   public CalciteMetaImpl(CalciteConnectionImpl connection) {
+    this(connection, new CalciteMetaTableFactoryImpl(), new CalciteMetaColumnFactoryImpl());
+  }
+
+  public CalciteMetaImpl(CalciteConnectionImpl connection,
```
julianhyde, line 107: Add javadoc.

I would remove the `@Nullable`. Caller can easily provide factories.

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaImpl.java
```diff
@@ -178,11 +205,16 @@ private <E> MetaResultSet createResultSet(Enumerable<E> enumerable,
...
         field = clazz.getField(fieldName);
       } catch (NoSuchFieldException e) {
-        throw new RuntimeException(e);
+        try {
+          // Check if subclass contains the desired field.
+          field = clazz.getDeclaredField(fieldName);
+        } catch (NoSuchFieldException e2) {
+          throw new RuntimeException(e2);
+        }
       }
```
julianhyde, line 218: I am surprised that `getDeclaredField` would ever succeed if `getField` has failed.

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaColumnFactory.java
```diff
@@ -0,0 +1,48 @@
...
+      Table table,
+      String tableCat,
+      String tableSchem,
+      String tableName,
+      String columnName,
+      int dataType,
+      String typeName,
+      Integer columnSize,
+      @Nullable Integer decimalDigits,
+      Integer numPrecRadix,
```
julianhyde, line 37: If `numPrecRadix` nullable? If not, make it `int`. If so, add `@Nullable`.

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaTableFactory.java
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.calcite.avatica.MetaImpl.MetaTable;
+import org.apache.calcite.schema.Table;
+
+/** Factory for creating MetaTables for getTables(). */
+public interface CalciteMetaTableFactory {
+  /** Instantiates a MetaTable. */
+  MetaTable newMetaTable(Table table, String tableCat, String tableSchem, String tableName);
+
+  /** Returns the list of expected column names. */
+  String[] getColumnNames();
```
julianhyde, line 28: Switch from `String[]` to `List<String>` so that you don't need defensive copy.

core/src/test/java/org/apache/calcite/test/JdbcFrontJdbcBackTest.java
```diff
@@ -73,6 +79,75 @@ class JdbcFrontJdbcBackTest {
         });
   }
 
+
+  /**
+   * Sample subclass used in {@link JdbcFrontJdbcBackTest#testTablesExtraColumn()}.
+   */
+  public static class MetaExtraTable extends CalciteMetaTable {
```
julianhyde, line 86: Does this class need to be public? Does its field need to be public?

If it can be private, that's good. If it needs to be public (because, say, reflection) then you should say that it needs to be public in the javadoc.

core/src/test/java/org/apache/calcite/test/JdbcFrontJdbcBackTest.java
```diff
@@ -73,6 +79,75 @@ class JdbcFrontJdbcBackTest {
...
+    MetaExtraTableFactoryImpl() {}
+
+    @Override public MetaTable newMetaTable(Table table, String tableCat, String tableSchem,
+        String tableName) {
+      return new MetaExtraTable(table, tableCat, tableSchem, tableName);
+    }
+
+    @Override public String[] getColumnNames() {
+      // 11 columns total.
+      return new String[] {
```
julianhyde, line 109: Would it be an error if the columns were not prefixed by the list of standard columns? (e.g. adding an extra column in the middle, or removing a column, or re-ordering columns)

Figure out what the necessary constraints are, and add a test that checks that those constraints are being enforced.

julianhyde (commented):
Nice work. The interface looks very clean.

My suggested changes are largely cosmetic, and I hope they're easy to apply.

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaImpl.java
olivrlee: I was testing on my subclasses and `getField()` returns the 10 fields of the default `MetaTable`  class attributes, and `getDeclaredField()` returns the additional properties on my subclass.

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaImpl.java
olivrlee: It's the Java 11 CI that is giving me issues when I remove the @nullable annotation. 

https://github.com/apache/calcite/actions/runs/6255263301/job/16984285491?pr=3421

julianhyde: I think CI is finding cases where you might accidentally supply null (e.g. when `Driver` calls `config.metaColumnFactory`). I'll fix.

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaImpl.java
julianhyde: Au contraire. The solution is to squash out the nulls.

The elvis operator makes it too easy to build horrible chains of optionals.

core/src/main/java/org/apache/calcite/jdbc/CalciteMetaColumnFactory.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.apache.calcite.avatica.MetaImpl.MetaColumn;
+import org.apache.calcite.schema.Table;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/** Factory for creating MetaColumns for getColumns(). */
+public interface CalciteMetaColumnFactory {
+  /** Instantiates a MetaColumn. */
+  MetaColumn newMetaColumn(
```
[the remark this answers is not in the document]
olivrlee: oh sorry I missed this last week somehow. Let me update right now

PR #3422, opened 2023-09-12
[CALCITE-5999] DECIMAL literals are sometimes unparsed looking as DOUBLE literals

mihaibudiu:


core/src/main/java/org/apache/calcite/sql/SqlNumericLiteral.java
```diff
@@ -90,7 +90,7 @@ public boolean isExact() {
   @Override public String toValue() {
     BigDecimal bd = getValueNonNull();
     if (isExact) {
-      return getValueNonNull().toString();
+      return getValueNonNull().toPlainString();
```
JiajunBernoulli, line 93: Why not use `bd.toPlainString()`?

And the `bd` should be `final`.

mihaibudiu: Good point, I pushed these changes. Since the PR is so small I have already squashed the two commits.

JiajunBernoulli (approved):

PR #3423, opened 2023-09-12
[CALCITE-5997] OFFSET operator is incorrectly unparsed

mihaibudiu:


mihaibudiu:
@tanclary I think you have added the "offset" field, so perhaps you want to review this PR. Thank you.

core/src/main/java/org/apache/calcite/sql/fun/SqlItemOperator.java
```diff
@@ -78,7 +78,13 @@ public SqlItemOperator(String name,
       SqlWriter writer, SqlCall call, int leftPrec, int rightPrec) {
     call.operand(0).unparse(writer, leftPrec, 0);
     final SqlWriter.Frame frame = writer.startList("[", "]");
-    call.operand(1).unparse(writer, 0, 0);
+    if (offset == 0) {
```
tanclary, line 81: do we need to add something similar for the other operators

tanclary: safe_offset, safe_ordinal etc

mihaibudiu: In fact, Calcite currently treats a simple array index as a SAFE_ORDINAL index, as the comment on SqlOperator ITEM says:

```
   * <p>The SQL standard calls the ARRAY variant a
   * &lt;array element reference&gt;. Index is 1-based. The standard says
   * to raise "data exception - array element error" but we currently return
   * null.
   public static final SqlOperator ITEM =
      new SqlItemOperator("ITEM", OperandTypes.ARRAY_OR_MAP, 1, true);
```

The last `true` argument is for the `safe` parameter.

If we take this approach, a raw index is unparsed as a SAFE_ORDINAL, reflecting the implementation semantics.
If you want to preserve the original syntax we need to carry additional information in the SqlItemOperator.
What is the right solution?

mihaibudiu: In other words, should the following test pass?
```java
@Test void testOffset() {
    sql("SELECT ARRAY[2,4,6][2]")
        .ok("SELECT (ARRAY[2, 4, 6])[SAFE_ORDINAL(2)]");
    sql("SELECT ARRAY[2,4,6][OFFSET(2)]")
        .ok("SELECT (ARRAY[2, 4, 6])[OFFSET(2)]");
    sql("SELECT ARRAY[2,4,6][SAFE_OFFSET(2)]")
        .ok("SELECT (ARRAY[2, 4, 6])[SAFE_OFFSET(2)]");
    sql("SELECT ARRAY[2,4,6][SAFE_ORDINAL(2)]")
        .ok("SELECT (ARRAY[2, 4, 6])[SAFE_ORDINAL(2)]");
  }
```

tanclary: I would imagine you do not want to unparse with SAFE_ORDINAL unless you are using BigQuery as it is dialect-specific syntax (to my knowledge). The comment you mentioned- it seems like a good time to address that as well if possible.

mihaibudiu: I don't see in the `unparse` function any context that can indicate which dialect is being unparsed.
The SqlItemOperator is not specific to BigQuery.
How would you even do that?

tanclary: Each dialect has a file like `BigQuerySqlDialect` and most have an "unparseCall" override with a switch statement that specifies certain `SqlKind` or `SqlOperator` that require special unparsing.

mihaibudiu: The BigQuerySqlDialect already handles this correctly.
The bug I am fixing was caused by SqlOperatorTest, where tests look like this:
```
final SqlOperatorFixture f0 = fixture();
    f0.setFor(SqlLibraryOperators.SAFE_ORDINAL);
    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.BIG_QUERY);
    f.checkScalar("ARRAY[2,4,6][SAFE_ORDINAL(3)]", "6", "INTEGER");    
```

Notice that this test only sets up the *library*, but not the dialect. In a new test fixture that I am writing, which unparses and reparses the query before executing, this test fails because the query is not unparsed correctly. In this case the unparsing does not go through the dialect.

I don't see a way to set the dialect in the SqlOperatorFixture. Do you have a suggestion?

mihaibudiu: Moreover, the SQL parser will happily parse the SAFE_* functions without the big query dialect.

mihaibudiu:
I hope I have found an adequate solution.
The BIG_QUERY dialect handled unparsing correctly, but since SAFE_ORDINAL and other related functions can be parsed with a standard SQL parser, there must be a way to unparse them even when the BIG_QUERY dialect is not being used.

core/src/main/java/org/apache/calcite/sql/fun/SqlItemOperator.java
tanclary: I know you can create the fixture with a specified parser config (see `withParserConfig()`) which I believe certain dialects like BigQuery have default parser configs. Perhaps you could pull on this thread a bit, the functionality may already be there.

mihaibudiu: True, but I still need to unparse such a query when the dialect is NOT specified.
The parser accepts SAFE_ORDINAL, so the unparser should too.

tanclary: Ah I see what you're saying. Yeah I believe when I originally added this cluster of functions that I essentially pair the OFFSET operator with ITEM because at the time they seemed identical. Admittedly I did not give enough thought to the unparsing so maybe treating them more independently would help also, given that that approach seems to work for most other BQ-specific operators.

core/src/main/java/org/apache/calcite/sql/fun/SqlItemOperator.java
```diff
@@ -50,13 +50,19 @@
...
+  /** Set to true if this is a simple index.
+   *  The semantics of the array index and SAFE_ORDINAL is identical;
+   *  this field tells them apart.  It is set to 'false' for SAFE_ORDINAL.
+   */
+  public final boolean simple;
 
   public SqlItemOperator(String name,
       SqlSingleOperandTypeChecker operandTypeChecker,
-      int offset, boolean safe) {
+      int offset, boolean safe, boolean simple) {
```
JiajunBernoulli, line 61: I think `String funcId` is more valuable than `boolean simple`.

core/src/main/java/org/apache/calcite/sql/fun/SqlItemOperator.java
```diff
@@ -78,7 +84,24 @@ public SqlItemOperator(String name,
       SqlWriter writer, SqlCall call, int leftPrec, int rightPrec) {
     call.operand(0).unparse(writer, leftPrec, 0);
     final SqlWriter.Frame frame = writer.startList("[", "]");
-    call.operand(1).unparse(writer, 0, 0);
+    if (offset == 0) {
+      // Technically these functions are extensions to standard SQL,
+      // however, the standard SQL parser accepts them,
+      // so there must be a way to unparse them.
+      String function = safe ? "SAFE_OFFSET" : "OFFSET";
```
JiajunBernoulli, line 91: If we have `String funcId`, we can write the code 
```
if (funcId != null) {
      final SqlWriter.Frame offsetFrame = writer.startFunCall(funcId);
      call.operand(1).unparse(writer, 0, 0);
      writer.endFunCall(offsetFrame);
}
else {
      call.operand(1).unparse(writer, 0, 0);
}
```

mihaibudiu: This is much better, thank you.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -725,6 +725,37 @@ protected static SortedSet<String> keywords(@Nullable String dialect) {
             + ".*");
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-5997">[CALCITE-5997]</a>
```
JiajunBernoulli, line 728: image

IMO, a tag should contain Jira summary.

JiajunBernoulli (approved):

JiajunBernoulli:
I think the current PR looks good.

@tanclary, Do you have any suggestions?

tanclary:
@JiajunBernoulli @mihaibudiu No, I think this looks good. I appreciate you opening the bug/fix. If you squash down to one commit I would be happy to merge.

tanclary (approved):

PR #3424, opened 2023-09-12
[CALCITE-5996] TRANSLATE operator is incorrectly unparsed

mihaibudiu:
@ILuffZhe this PR fixes a test which you recently wrote which I suspect was wrong.
Can you please take a look?

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
+++ b/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
@@ -5491,16 +5491,25 @@ public class SqlParserTest {
             + "FROM `T`");
 
     expr("convert('abc' using utf8)")
-        .ok("TRANSLATE('abc', `UTF8`)");
+        .ok("TRANSLATE('abc' USING `UTF8`)");
     sql("select convert(name using gbk) as newName from t")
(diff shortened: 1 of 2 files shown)


testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -782,6 +782,17 @@ protected static SortedSet<String> keywords(@Nullable String dialect) {
             + "FROM UNNEST((ARRAY[1, NULL, 3, 4])) AS x");
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-5996">[CALCITE-5996]</a>
+   * TRANSLATE operator is incorrectly unparsed. */
+  @Test void testTranslate() {
+    sql("select translate(col using utf8)\n"
```
ILuffZhe, line 788: I think it would be better if you move this test case to `testConvertAndTranslate()`.

ILuffZhe (commented):
Left one minor comment.

ILuffZhe (approved):
LGTM!

ILuffZhe:
Hi @mihaibudiu , can you squash your commits? I'll merge this PR soon if there are no more comments come.

PR #3425, opened 2023-09-12
[CALCITE-6001] Add dialect-specific character sets for encoding string literals

tanclary:


julianhyde:
Since literals also exist in source SQL, and intermediate SqlNode and RexNodes, you should change the summary to make clear that this relates to generating SQL. 'JDBC adapter should...' is one way.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1726,6 +1726,9 @@ private static String toSql(RelNode root, SqlDialect dialect,
     assertThat(toSql(root), isLinux(expectedSql));
   }
 
+
```
mihaibudiu, line 1729: maybe you want to take these empty lines back

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
```diff
@@ -1664,7 +1700,20 @@ private ContextImpl(DatabaseProduct databaseProduct,
...
-          quotedCasing, unquotedCasing, caseSensitive,
+          charset, quotedCasing, unquotedCasing, caseSensitive,
+          conformance, nullCollation, dataTypeSystem, jethroInfo);
+    }
+
+    @Override public String charset() {
+      return charset;
+    }
+
+    @Override public Context withCharset(String charset) {
```
mihaibudiu, line 1711: does this qualify as a breaking change which has to be documented in history.md?

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
julianhyde: I don't think so. ContextImpl is private.

PR #3426, opened 2023-09-12
[CALCITE-6003] JSON_ARRAY() with no arguments does not unparse correctly

mihaibudiu:


mihaibudiu:
@julianhyde I think you wrote this code originally

julianhyde (commented):
I preferred `json_array()` in upper-case in the commit message.

I would remove the 'Test case for ...' comment. The `<a>` is out of place since Java comments are not HTML, and the test case is self-explanatory.

+1 with those changes

mihaibudiu:
I hope I have addressed the comments, would be great if this can be approved and merged.
(I have already squashed the commits, since it's a very small PR).

mihaibudiu:
Tiny bugfix PR, hopefully someone can find a few minutes to review/approve/merge.
Thank you.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -8719,6 +8719,10 @@ private static Consumer<List<? extends Throwable>> checkWarnings(
   }
 
   @Test void testJsonArray() {
+    // https://issues.apache.org/jira/browse/CALCITE-6003
```
julianhyde, line 8722: please add the jira url as a javadoc comment, like the other test do

mihaibudiu: Done. Moved the new test to a separate function for that purpose, I have squashed the commits, since this is a small PR.

PR #3428, opened 2023-09-13
Incorrect test fixture used by SqlOperatorTest.testLeastFunc

mihaibudiu:
I understand I don't need to file an issue for such a small fix.
The test passes, but may fail in overloaded subclasses of SqlOperatorTest.


libenchao (approved):
+1, I'd always tag it as [MINOR] to explicitly tell the release manager that this is a trivial work which does not need to be included in the release note.

PR #3429, opened 2023-09-15
[CALCITE-6006] RelToSqlConverter loses charset information

mihaibudiu:
The problem addressed is that character literals with a charset specified lose charset information when the program is translated to Rel and back. This change adds the charset to SqlLiterals, but only if the charset is not the default one. (Prior to this fix the charset was always 'null'). We could always add the charset, but then many literals obtained from RelToSql would look like `"_ISO-8859-1'char'"`

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -16,6 +16,7 @@
  */
 package org.apache.calcite.rel.rel2sql;
 
+import org.apache.calcite.config.CalciteSystemProperty;
 import org.apache.calcite.linq4j.Ord;
 import org.apache.calcite.linq4j.tree.Expressions;
 import org.apache.calcite.plan.RelOptUtil;
@@ -90,6 +91,7 @@ import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.validate.SqlValidatorUtil;
 import org.apache.calcite.util.DateString;
 import org.apache.calcite.util.ImmutableBitSet;
+import org.apache.calcite.util.NlsString;
 import org.apache.calcite.util.Pair;
 import org.apache.calcite.util.RangeSets;
 import org.apache.calcite.util.Sarg;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1409,8 +1411,20 @@ public static SqlNode toSql(RexLiteral literal) {
...
     switch (family) {
-    case CHARACTER:
+    case CHARACTER: {
+      NlsString value = literal.getValueAs(NlsString.class);
+      if (value != null) {
+        String charsetName = value.getCharsetName();
+        if (charsetName != null
+            && charsetName.equals(CalciteSystemProperty.DEFAULT_CHARSET.value())) {
+          // Do not set the charset if it is the same as the default charset
+          charsetName = null;
```
JiajunBernoulli, line 1421: Why `null` mean default? Can we use DEFAULT_CHARSET rather than null?

mihaibudiu: Before this change it was always null, which is wrong.
A null charset seems to have the same meaning as the default charset.
But when you unparse them, they look different: the null charset shows just a literal, e.g., 'x', but the default charset (at least the default default charset) shows up as `_ISO-8859-1'`x'` 
So if I leave the charset always there many tests need to be changed.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1409,8 +1411,20 @@ public static SqlNode toSql(RexLiteral literal) {
             () -> "literal " + literal
                 + " has null SqlTypeFamily, and is SqlTypeName is " + typeName);
     switch (family) {
-    case CHARACTER:
+    case CHARACTER: {
+      NlsString value = literal.getValueAs(NlsString.class);
```
JiajunBernoulli, line 1415: Please use `final`

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -268,6 +268,14 @@ private static String toSql(RelNode root, SqlDialect dialect,
     sql(query).ok(expected);
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6006">[CALCITE-6006]</a>
+   * RelToSqlConverter loses charset information. */
```
JiajunBernoulli, line 272: For different character sets, we need more test cases.

mihaibudiu: I have added one test with UTF16, is that enough?

JiajunBernoulli (approved):

JiajunBernoulli:
@mihaibudiu , would you please squash commits?

JiajunBernoulli:
@mihaibudiu   Commit message should start with `[CALCITE-6006]`, please fix it.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1409,8 +1411,20 @@ public static SqlNode toSql(RexLiteral literal) {
             () -> "literal " + literal
                 + " has null SqlTypeFamily, and is SqlTypeName is " + typeName);
     switch (family) {
-    case CHARACTER:
+    case CHARACTER: {
+      final NlsString value = literal.getValueAs(NlsString.class);
```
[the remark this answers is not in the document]
mihaibudiu: I have applied this suggested change, and squashed the commits optimistically.

PR #3289, opened 2023-07-01
[CALCITE-5813] Type inference for sql functions REPEAT, SPACE, XML_TRANSFORM, and XML_EXTRACT is incorrect

mihaibudiu:
This bug is reproduced only when specific optimizations are run on the buggy (now corrected) function definitions.
I have added the test cases in RelOptRulesTest, because that's where it's easy to control the optimizations applied, but if there is a better file for these tests, please let me know. They depend on specific SQL dialects.

PR #3430, opened 2023-09-16
[CALCITE-6008] ARRAY_AGG should returns ARRAY NULL when there's no input rows

Jiabao-Sun:
ARRAY_AGG in BigQuery and Postgres returns NULL if there are zero input rows or expression evaluates to NULL for all rows.
However, it will return `ARRAY<arg0> NOT NULL` in the return type of the function definition.

https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#array_agg
https://www.postgresql.org/docs/8.4/functions-aggregate.html

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
@@ -671,6 +671,13 @@ public abstract class ReturnTypes {
       ARG0.andThen(SqlTypeTransforms.TO_ARRAY);
 
   /**
+   * Type-inference strategy whereby the result type of a call is nullable
+   * <code>ARRAY</code>.
+   */
+  public static final SqlReturnTypeInference TO_ARRAY_NULLABLE =
+      TO_ARRAY.andThen(SqlTypeTransforms.TO_NULLABLE);
+
+  /**
    * Returns a MAP type.
    *
    * <p>For example, given {@code Record(f0: INTEGER, f1: DATE)}, returns
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -11089,11 +11089,12 @@ public class SqlOperatorTest {
   }
 
   private static void checkArrayAggFunc(SqlOperatorFixture f) {
-    f.setFor(SqlLibraryOperators.ARRAY_CONCAT_AGG, VM_FENNEL, VM_JAVA);
+    f.setFor(SqlLibraryOperators.ARRAY_AGG, VM_FENNEL, VM_JAVA);
     final String[] values = {"'x'", "null", "'yz'"};
-    f.checkAgg("array_agg(x)", values, isSingle("[x, yz]"));
-    f.checkAgg("array_agg(x ignore nulls)", values, isSingle("[x, yz]"));
-    f.checkAgg("array_agg(x respect nulls)", values, isSingle("[x, yz]"));
+    f.checkAggType("array_agg(x)", "INTEGER NOT NULL ARRAY NOT NULL");
+    f.checkAgg("array_agg(x)", values, "CHAR(2) ARRAY", isSingle("[x, yz]"));
+    f.checkAgg("array_agg(x ignore nulls)", values, "CHAR(2) ARRAY", isSingle("[x, yz]"));
+    f.checkAgg("array_agg(x respect nulls)", values, "CHAR(2) ARRAY", isSingle("[x, yz]"));
     final String expectedError = "Invalid number of arguments "
         + "to function 'ARRAY_AGG'. Was expecting 1 arguments";
     f.checkAggFails("^array_agg(x,':')^", values, expectedError, false);
@@ -11104,7 +11105,7 @@ public class SqlOperatorTest {
   }
 
   private static void checkArrayAggFuncFails(SqlOperatorFixture t) {
-    t.setFor(SqlLibraryOperators.ARRAY_CONCAT_AGG, VM_FENNEL, VM_JAVA);
+    t.setFor(SqlLibraryOperators.ARRAY_AGG, VM_FENNEL, VM_JAVA);
     final String[] values = {"'x'", "'y'"};
     final String expectedError = "No match found for function signature "
         + "ARRAY_AGG\\(<CHARACTER>\\)";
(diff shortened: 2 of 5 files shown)


Jiabao-Sun:
Hi @julianhyde, could you help review it when you have time?
Thanks a lot.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -9314,11 +9314,12 @@ private static void checkGroupConcatFuncFails(SqlOperatorFixture t) {
   }
 
   private static void checkArrayAggFunc(SqlOperatorFixture f) {
-    f.setFor(SqlLibraryOperators.ARRAY_CONCAT_AGG, VM_FENNEL, VM_JAVA);
+    f.setFor(SqlLibraryOperators.ARRAY_AGG, VM_FENNEL, VM_JAVA);
```
mihaibudiu, line 9317: When fixing some similar bugs I have found that a RelOptRulesTest can often exhibit the bug by producing a wrong plan.
Here is an example: https://github.com/apache/calcite/pull/3289/files#diff-a430972be1169f680b7886b824720254fa1dabe37191db34bd2a63f10cf7a438
I don't know if it will work in your case as well.

Jiabao-Sun: Thanks @mihaibudiu.
The main issue with this question is incorrect function return type inference, and it seem's not be related to the planner rules. And the old test case of this line uses the wrong operator.

Jiabao-Sun: Sorry @mihaibudiu, I read the example again and it makes sense to me.
I can add some test in `RelOptRulesTest` as well.

Jiabao-Sun: But I can't find any proper rules to verify that `ARRAY_AGG($0)` should returns  `ARRAY NULL` type.

mihaibudiu: The optimization will replace the plan with a query returning constant null.

tanclary: Maybe I'm wrong but I'm not sure `RelOptRulesTest` is the best place to be testing appropriate return types. If you want to see the rel plan generated by certain sql, maybe `SqlToRelConverterTest` could be more what you are looking for?

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -645,6 +645,13 @@ public static SqlCall stripSeparator(SqlCall call) {
   public static final SqlReturnTypeInference TO_ARRAY =
       ARG0.andThen(SqlTypeTransforms.TO_ARRAY);
 
+  /**
+   * Same as {@link #TO_ARRAY} but returns with nullability if any of the
```
tanclary, line 649: Would you mind updating this javadoc to match the format of other `*_NULLABLE` return types? Here is `DATE` and `DATE_NULLABLE` as an example:

```
  /**
   * Type-inference strategy whereby the result type of a call is DATE.
   */
  public static final SqlReturnTypeInference DATE =
      explicit(SqlTypeName.DATE);

  /**
   * Type-inference strategy whereby the result type of a call is nullable
   * DATE.
   */
  public static final SqlReturnTypeInference DATE_NULLABLE =
      DATE.andThen(SqlTypeTransforms.TO_NULLABLE);
```

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -645,6 +645,13 @@ public static SqlCall stripSeparator(SqlCall call) {
...
+   * operands is nullable.
+   */
+  public static final SqlReturnTypeInference ARRAY_NULLABLE =
```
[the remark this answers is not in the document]
Jiabao-Sun: Thanks @tanclary for the review.
`ARRAY_NULLABLE`'s naming refers to `MULTISET` because they are all collection types.
I am not sure if this is appropriate. 
If you think it should be named `TO_ARRAY_NULLABLE `, I can correct it.

```java
  /**
   * Returns a MULTISET type.
   *
   * <p>For example, given <code>INTEGER</code>, returns
   * <code>INTEGER MULTISET</code>.
   */
  public static final SqlReturnTypeInference TO_MULTISET =
      ARG0.andThen(SqlTypeTransforms.TO_MULTISET);

  /**
   * Same as {@link #MULTISET} but returns with nullability if any of the
   * operands is nullable.
   */
  public static final SqlReturnTypeInference MULTISET_NULLABLE =
      MULTISET.andThen(SqlTypeTransforms.TO_NULLABLE);
```

Jiabao-Sun:
Many thanks to @tanclary, please help review it again when you have time.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: I have found that sometimes optimization rules produce results that differ from the runtime results for the same expression. In particular `CoreRules.PROJECT_REDUCE_EXPRESSIONS` will optimize constant expressions. Currently there is no easy way to check that the optimization produces the expected result. Does the SqlToRelConverterTest allow you to specify optimizations?

I am working towards https://issues.apache.org/jira/browse/CALCITE-5891, once that exists all the existing SqlOperatorTests will be automatically run in both ways. (But it may take awhile, it has uncovered at least 50 bugs in RelToSql and the optimizer.)

mihaibudiu: This was just a suggestion, I don't think it is required necessarily for this PR.

Jiabao-Sun:
Hi @tanclary, is there any other change opinions?

mihaibudiu (approved):

mihaibudiu:
@tanclary you asked for changes, are you fine with this PR?

Jiabao-Sun:
main branch rebased, thanks for the review.

mihaibudiu:
you should also squash the commits into a single one

mihaibudiu:
for the commit message please use the exact same string as the jira issue title and the PR title

Jiabao-Sun:
> for the commit message please use the exact same string as the jira issue title and the PR title

Thanks, done.

PR #3431, opened 2023-09-17
[CALCITE-5523] RelToSql converter generates GROUP BY clause with wind…

JiajunBernoulli:
…ow expression

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -1878,6 +1878,11 @@ public abstract class SqlImplementor {
             && (hasNestedAgg || hasNestedWindowedAgg)) {
           return true;
         }
+        if (!dialect.supportsNestedAggregations()
+            && agg.getInput() instanceof Project
+            && ((Project) agg.getInput()).containsOver()) {
+          return true;
+        }
 
         if (clauses.contains(Clause.GROUP_BY)) {
           // Avoid losing the distinct attribute of inner aggregate.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1878,6 +1878,11 @@ && hasSortByOrdinal()) {
             && (hasNestedAgg || hasNestedWindowedAgg)) {
           return true;
         }
+        if (!dialect.supportsNestedAggregations()
+            && agg.getInput() instanceof Project
```
LeonidChistov, line 1882: What if input is Window object?

LeonidChistov: And also `agg.getInput()` could be a `Calc` as well.

IMO something is deeply wrong with `needNewSubQuery` in general in terms of that it uses `instanceof Project` in multiple places, but does not really care about having `Calc` or `Window` on the same position, instead.

One of the possible solutions that could make whole code of RelToSqlConverter simpler, is to do some "normalization" of relational tree inside `SqlImplementor::visitRoot(...)` method in order to transform it to a form having less possible amount of node types.

But I guess that is something that needs to be discussed on the mailing list.

WDYT?

JiajunBernoulli: When will the input of Aggreagte be Window?

Window only appear when you use `ProjectToWindowRule` and I think the window will not be input of Aggregate.

LeonidChistov: @JiajunBernoulli 

It looks that you are right, at least I cannot reproduce an issue with `Window` under `Aggregate` that would also have a problem of aggregation by window expression.

But I guess that part of my comment about `Calc` is still relevant.

LakeShen:
Hi @JiajunBernoulli 

It looks like the title and commit's description is not clear.

PR #3433, opened 2023-09-18
[CALCITE-6014] Create a SqlOperatorFixture that parses, unparses, and…

mihaibudiu:
… then parses again before executing

The code contains disabled tests for 5 bugs; these bugs have been all reported, and 3 already have PRs fixing them.
These disabled tests should be removed once the fixes are merged.

The change to SqlOperatorFixtureImpl also fixes a small and obvious bug.

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/SqlOperatorUnparseTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlOperatorUnparseTest.java
@@ -1,0 +1,116 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.test;
+
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.apache.calcite.sql.parser.SqlParser;
+import org.apache.calcite.sql.test.SqlOperatorFixture;
+import org.apache.calcite.sql.test.SqlTestFactory;
+
+import org.junit.jupiter.api.Disabled;
+
+import java.util.function.Consumer;
+import java.util.function.UnaryOperator;
+
+/**
+ * Version of a SqlOperatorTest which first parses and unparses
+ * the test program before executing it. Although similar to
+ * {@link org.apache.calcite.sql.parser.SqlUnParserTest},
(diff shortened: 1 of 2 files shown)


mihaibudiu:
It would be great if someone could review this PR. It essentially adds a few hundred tests for the parser/unparser by reusing all the SqlOperatorTest tests. It caught at least 8 bugs already, most of which I have fixed. Going forward, it may make it much easier to write unparser tests.

core/src/test/java/org/apache/calcite/test/SqlOperatorUnparseTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+import org.junit.jupiter.api.Disabled;
+
+import java.util.function.Consumer;
+import java.util.function.UnaryOperator;
+
+/**
+ * Version of a SqlOperatorTest which first parses and unparses
+ * the test program before executing it.  Although similar to
+ * SqlUnparserTest, this test also validates the code after unparsing.
```
rubenada, line 33: Perhaps adding a link to `SqlUnparserTest`?

core/src/test/java/org/apache/calcite/test/SqlOperatorUnparseTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.calcite.sql.test.SqlTestFactory;
+
+import org.junit.jupiter.api.Disabled;
+
+import java.util.function.Consumer;
+import java.util.function.UnaryOperator;
+
+/**
+ * Version of a SqlOperatorTest which first parses and unparses
+ * the test program before executing it.  Although similar to
```
rubenada, line 32: Is it possible that there is an extra whitespace after the dot and "Although" ?

rubenada:
Thanks @mihaibudiu for moving this forward, I think this is a good idea to unveil bugs.
I think the PR is in a pretty good shape, I have just left some minor comments.

core/src/test/java/org/apache/calcite/test/SqlOperatorUnparseTest.java
mihaibudiu: As a longtime emacs users I have been trained to use two spaces after the real periods. I can remove this, although I personally find it more readable.

rubenada: Ok, no big deal, let's leave it like this. I just noticed that it is the same on the license comment at the beginning of the file.

core/src/test/java/org/apache/calcite/test/SqlOperatorUnparseTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+import org.junit.jupiter.api.Disabled;
+
+import java.util.function.Consumer;
+import java.util.function.UnaryOperator;
+
+/**
+ * Version of a SqlOperatorTest which first parses and unparses
+ * the test program before executing it. Although similar to
+ * {@link org.apache.calcite.sql.parser.SqlUnParserTest},
```
mihaibudiu, line 33: I have added a link, but at least intellij complains about that class not being accessible, so I added a SuppressWarnings below.

mihaibudiu:
I hope I have addressed the comments. I have optimistically squashed the commits to make this ready to merge.

core/src/test/java/org/apache/calcite/test/SqlOperatorUnparseTest.java
rubenada: Thanks @mihaibudiu , I hadn't noticed about the accessibility, SuppressWarnings is fine for me

PR #3434, opened 2023-09-18
[CALCITE-6013] RelBuilder should simplify plan by pruning unused measures

tanclary:
…truction

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
@@ -4691,6 +4691,23 @@ class SqlToRelConverterTest extends SqlToRelTestBase {
   }
 
   /** Test case for:
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6013">[CALCITE-6013]
+   * Unnecessary measures added as projects during rel construction</a>.
+   */
+  @Test void testAvoidUnnecessaryMeasureProject() {
+    final String sql = "select deptno\n"
+        + "from empm\n"
+        + "group by deptno";
+    fixture()
+        .withFactory(c ->
+            c.withOperatorTable(t ->
+                SqlValidatorTest.operatorTableFor(SqlLibrary.CALCITE)))
+        .withCatalogReader(MockCatalogReaderExtended::create)
+        .withSql(sql)
+        .ok();
+  }
+
+  /** Test case for:
    * <a href="https://issues.apache.org/jira/browse/CALCITE-3310">[CALCITE-3310]
    * Approximate and exact aggregate calls are recognized as the same
    * during sql-to-rel conversion</a>.
--- a/core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
@@ -524,6 +524,20 @@ LogicalProject(EXPR$0=[ITEM(ITEM($3, 1).DETAIL.SKILLS, +(2, 3)).DESC])
 ]]>
     </Resource>
   </TestCase>
+  <TestCase name="testAvoidUnnecessaryMeasureProject">
+    <Resource name="sql">
+      <![CDATA[select deptno
+from empm
+group by deptno]]>
+    </Resource>
+    <Resource name="plan">
+      <![CDATA[
+LogicalAggregate(group=[{0}])
+  LogicalProject(DEPTNO=[$7])
+    LogicalTableScan(table=[[CATALOG, SALES, EMPM]])
+]]>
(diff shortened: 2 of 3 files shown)


core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -4689,6 +4689,23 @@ void checkUserDefinedOrderByOver(NullCollation nullCollation) {
...
+  
+  /** Test case for:
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6013">[CALCITE-6013]
+   * Unnecessary measures added as projects during rel construction</a>.
+   */  
+    @Test void testAvoidUnnecessaryMeasureProject() {
+    final String sql = "select deptno\n"
+        + "from empm\n"
+        + "group by deptno";
+    fixture()
```
tanclary, line 4701: for context I added this so the mock table w/ measure columns gets used

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -4502,7 +4516,7 @@ group by deptno]]>
     <Resource name="plan">
       <![CDATA[
 LogicalAggregate(group=[{0}], C=[AGGREGATE($1)])
-  LogicalProject(DEPTNO=[$7], COUNT_PLUS_100=[$9], COUNT_TIMES_100=[$10])
+  LogicalProject(DEPTNO=[$7], COUNT_PLUS_100=[$9])
```
mihaibudiu, line 4519: why is there a count?

tanclary: I might be misunderstanding your question but there is a count in the original query (see new line 4512: `select deptno, aggregate(count_plus_100) as c`)

mihaibudiu: Thanks, I misread the diff file.

tanclary: No worries, appreciate you taking a look!

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -4690,6 +4690,23 @@ void checkUserDefinedOrderByOver(NullCollation nullCollation) {
...
   }
 
+  /** Test case for:
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6013">[CALCITE-6013]
+   * Unnecessary measures added as projects during rel construction</a>.
+   */
+  @Test void testAvoidUnnecessaryMeasureProject() {
+    final String sql = "select deptno\n"
+        + "from empm\n"
+        + "group by deptno";
```
tjbanghart, line 4700: The `GROUP BY` is needed here to trigger the bug correct? Otherwise `createAggImpl` is not called?

Are we certain this issue only impacted aggregate queries?

tanclary: Yeah that is my understanding. As a quick test I removed the `GROUP BY` and put the offending lines back in `SqlToRelConverter`. No error.

tjbanghart (approved):

julianhyde:
The jira summary (commit message) seems to describe the solution rather than the problem. Is 'GROUP BY query on a table with measures gives error xxx' a better description?

Once you have described the problem in the commit message, you easily then describe where the bug was (in this case I guess SqlToRelConverter).

Actually, reading the jira case now, it looks like there is no error, just a plan different than you expected. We should discuss in jira whether this is even a bug. Pruning is often better deferred.

tanclary:
@julianhyde I think the commit message does describe the problem which is that the projects are populated with all measures while only "relevant" grouped exprs are added. What do you think?

julianhyde:
I agree that 's the 'problem' at the code level, but it doesn't describe the symptoms that most people are likely to experience. The commit message should talk about the symptoms because that allows as many people as possible to find it.

Often, describing the features of a SQL query ('Query with LIMIT and no ORDER BY throws NullPointerException') is the way to go. In the first line of the commit message you can describe what's going on in the code that causes those symptoms.

tanclary:
Is this message better @julianhyde ? I am having some trouble describing it since in most instances it doesn't result in an actual error.

julianhyde:
I agree this one isn't a bug visible to the SQL user. How about 'RelBuilder should prune unused measures'.

And add a justification in the commit message or a code comment that it doesn't affect correctness, just simplifies planning process. The word 'should' conveys a lot of meaning.

+1 to merge the PR with whatever message you choose.

PR #3435, opened 2023-09-21
[CALCITE-6015] AssertionError during optimization of EXTRACT expression

mihaibudiu:
While attempting to fix the bug described in the issue I have discovered that many cases for EXTRACT are actually not implemented. So I implemented most of them, and I have improved the validation as well; the case indicated in the JIRA issue will now be rejected statically.

There is one case for EXTRACT which I haven't implemented: `EXTRACT(EPOCH FROM INTERVAL '...' YEAR TO MONTH)` because I wasn't sure how to properly use the convertlets to do it. Help would be appreciated.

This PR also makes some unexpected tiny changes in SqlValidatorImpl and SqlCall, so I would appreciate comments from people who are more knowledgeable.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlCall.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlCall.java
@@ -191,7 +191,7 @@ public abstract class SqlCall extends SqlNode {
    * Returns a string describing the actual argument types of a call, e.g.
    * "SUBSTR(VARCHAR(12), NUMBER(3,2), INTEGER)".
    */
-  protected String getCallSignature(
+  public String getCallSignature(
       SqlValidator validator,
       @Nullable SqlValidatorScope scope) {
     List<String> signatureList = new ArrayList<>();
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlExtractFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlExtractFunction.java
@@ -17,6 +17,7 @@
 package org.apache.calcite.sql.fun;
 
 import org.apache.calcite.avatica.util.TimeUnitRange;
+import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlFunctionCategory;
@@ -26,12 +27,16 @@ import org.apache.calcite.sql.SqlOperatorBinding;
 import org.apache.calcite.sql.SqlWriter;
 import org.apache.calcite.sql.type.OperandTypes;
 import org.apache.calcite.sql.type.ReturnTypes;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.validate.SqlMonotonicity;
 import org.apache.calcite.sql.validate.SqlValidator;
 import org.apache.calcite.sql.validate.SqlValidatorScope;
 import org.apache.calcite.util.Util;
 
+import com.google.common.collect.ImmutableSet;
+
 import static org.apache.calcite.sql.validate.SqlNonNullableAccessors.getOperandLiteralValueOrThrow;
+import static org.apache.calcite.util.Static.RESOURCE;
 
 /**
  * The SQL <code>EXTRACT</code> operator. Extracts a specified field value from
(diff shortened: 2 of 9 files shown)


mihaibudiu:
This PR implements several instances of the EXTRACT function which weren't supported. I would appreciate a review.

mihaibudiu:
Can anyone please review this PR?

core/src/main/java/org/apache/calcite/sql/fun/SqlExtractFunction.java
```diff
@@ -83,8 +88,102 @@ public SqlExtractFunction(String name) {
...
+          .add(SqlTypeName.TIMESTAMP_WITH_LOCAL_TIME_ZONE)
+          .addAll(SqlTypeName.YEAR_INTERVAL_TYPES)
+          .build()
+          .contains(typeName);
+      break;
+    case WEEK:
+    case DOW:
+    case ISODOW:
+    case DOY:
+      legal = new ImmutableList.Builder<SqlTypeName>()
```
snuyanzin, line 122: should we extract the result of the builder here and others into constants (may be sets instead of lists) to avoid building it each time we come here?

core/src/main/java/org/apache/calcite/sql/SqlCall.java
```diff
@@ -191,7 +191,7 @@ public int operandCount() {
    * Returns a string describing the actual argument types of a call, e.g.
    * "SUBSTR(VARCHAR(12), NUMBER(3,2), INTEGER)".
    */
-  protected String getCallSignature(
+  public String getCallSignature(
```
rubenada, line 194: This change on method visibility, although minor, should be considered a breaking change and IMO mentioned briefly on the RN: any downstream project that had overridden it as `protected` would have a compilation issue with the new version, and would need to adjust their code to override it as `public`.

mihaibudiu: what should I do about it?

rubenada: Usually in these cases, to make sure the Release Manager won't forget about it, what we do is editing the `history.md`: in that file, at the top, there is the "skeleton" with the sections of the next release (commented out), you could add a point in there in the "breaking changes" section mentioning this.

mihaibudiu: I have added this to history.md. Thank you for pointing this out.

core/src/main/java/org/apache/calcite/sql/fun/SqlExtractFunction.java
mihaibudiu: I am now using static fields for these constant lists.

rubenada: Why not using sets as @snuyanzin suggested?

site/_docs/history.md
```diff
@@ -57,6 +57,9 @@ other software versions as specified in gradle.properties.
 #### Bug-fixes, API changes and minor enhancements
 {: #fixes-1-37-0}
 
+* The visibility of the method SqlCall.getCallSignature has been converted
```
rubenada, line 60: Thanks for including this. I'd suggest to give a bit more info, e.g. something a long the lines:

In the context of  [<a href="https://issues.apache.org/jira/browse/CALCITE-6015">CALCITE-6015</a>] the visibility of the method `SqlCall.getCallSignature` has been converted from `protected` to `public`, so any subclass overriding it might need to be adjusted accordingly.

mihaibudiu: @rubenada changed List to Set and updated the history file.

rubenada:
Thanks @mihaibudiu , no further comment from my side. LGTM.

mihaibudiu:
Squashed the commits in preparation for merging.

site/_docs/history.md
```diff
@@ -57,6 +57,10 @@ other software versions as specified in gradle.properties.
 #### Bug-fixes, API changes and minor enhancements
 {: #fixes-1-37-0}
 
+* In the context of [CALCITE-6015] the visibility of the method
```
rubenada, line 60: @mihaibudiu I just noticed, this comment should better be placed a few lines up, right below  the 
```
#### Breaking Changes
{: #breaking-1-37-0}
```

mihaibudiu: sorry about that, I have fixed this, and also squashed the commits

rubenada:
Thanks @mihaibudiu !
@snuyanzin  do you have any further remark? Otherwise IMO this PR is in a good shape to be merged.

snuyanzin (approved):
Thanks for asking 
and thanks for addressing comments
it looks ok from my side

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10862,6 +10810,47 @@ private static void checkArrayConcatAggFuncFails(SqlOperatorFixture t) {
...
+    f.checkFails("^extract(quarter from time '12:34:56')^", fail, false);
+    f.checkFails("^extract(year from time '12:34:56')^", fail, false);
+    f.checkFails("^extract(isoyear from time '12:34:56')^", fail, false);
+    f.checkFails("^extract(doy from time '12:34:56')^", fail, false);
+    f.checkFails("^extract(dow from time '12:34:56')^", fail, false);
+    f.checkFails("^extract(week from time '12:34:56')^", fail, false);
+    f.checkFails("^extract(decade from time '12:34:56')^", fail, false);
+    f.checkFails("^extract(century from time '12:34:56')^", fail, false);
+    f.checkFails("^extract(century from time '12:34:56')^", fail, false);
+    f.checkFails("^extract(century from time '12:34:56')^", fail, false);
```
mihaibudiu, line 10849: noticed some duplicated tests, I will delete these before merging.

PR #3436, opened 2023-09-21
[CALCITE-5984] Allow disabling field trim via `SqlToRelConverter` config

WegdanGhazi:
[CALCITE-5984](https://issues.apache.org/jira/browse/CALCITE-5984) Allow disabling field trim via `SqlToRelConverter` config

Use the value of `isTrimUnusedFields` set to  `SqlToRelConverter.Config` to determine whether to trim unused fields, making it possible to completely disable trimming. 

As seen [here](https://github.com/apache/calcite/blob/454899a451d90726affdddfa75b1de24904d3072/core/src/main/java/org/apache/calcite/prepare/Prepare.java#L376), the Prepare#trim function creates a new instance of `SqlToRelConverter.Config`, overwriting existing `isTrimUnusedFields` settings.

This PR fixes this by checking the boolean `trimUnusedFields` before calling `#trim`.

Unit tests were added to ensure that trimming can be completely disabled by controlling:
1. Disabling `isTrimUnusedFields` on  `SqlToRelConverter.Config`.
2. Removing `TrimFieldsProgram` from the initialization of the optimization program.

### DIFF
--- a/core/src/main/java/org/apache/calcite/prepare/Prepare.java
+++ b/core/src/main/java/org/apache/calcite/prepare/Prepare.java
@@ -293,10 +293,12 @@ public abstract class Prepare {
       root = root.withRel(decorrelate(sqlToRelConverter, sqlQuery, root.rel));
     }
 
-    // Trim unused fields.
-    root = trimUnusedFields(root);
+    if (configHolder.get().isTrimUnusedFields()) {
+      // Trim unused fields.
+      root = trimUnusedFields(root);
 
-    Hook.TRIMMED.run(root.rel);
+      Hook.TRIMMED.run(root.rel);
+    }
 
     // Display physical plan after decorrelation.
     if (sqlExplain != null) {
--- a/core/src/main/java/org/apache/calcite/tools/Programs.java
+++ b/core/src/main/java/org/apache/calcite/tools/Programs.java
@@ -250,11 +250,12 @@ public class Programs {
 
   /** Returns the standard program used by Prepare. */
   public static Program standard() {
-    return standard(DefaultRelMetadataProvider.INSTANCE);
+    return standard(DefaultRelMetadataProvider.INSTANCE, true);
   }
 
   /** Returns the standard program with user metadata provider. */
-  public static Program standard(RelMetadataProvider metadataProvider) {
+  public static Program standard(RelMetadataProvider metadataProvider,
+      boolean enableFieldTrimming) {
     final Program program1 =
         (planner, rel, requiredOutputTraits, materializations, lattices) -> {
           for (RelOptMaterialization materialization : materializations) {
@@ -278,7 +279,8 @@ public class Programs {
           return rootRel3;
         };
 
-    return sequence(subQuery(metadataProvider),
+    List<Program> programs =
+        Lists.newArrayList(subQuery(metadataProvider),
         new DecorrelateProgram(),
         new TrimFieldsProgram(),
         program1,
@@ -286,6 +288,10 @@ public class Programs {
         // Second planner pass to do physical "tweaks". This the first time
         // that EnumerableCalcRel is introduced.
         calc(metadataProvider));
+
+    programs.removeIf(program -> !enableFieldTrimming && program instanceof TrimFieldsProgram);
+
+    return new SequenceProgram(ImmutableList.copyOf(programs));
   }
 
   /** Program backed by a {@link RuleSet}. */
--- a/core/src/main/java/org/apache/calcite/tools/Programs.java
+++ b/core/src/main/java/org/apache/calcite/tools/Programs.java
@@ -254,6 +254,11 @@ public class Programs {
   }
 
   /** Returns the standard program with user metadata provider. */
+  public static Program standard(RelMetadataProvider metadataProvider) {
+    return standard(metadataProvider, true);
+  }
+
+  /** Returns the standard program with user metadata provider and enableFieldTrimming config. */
   public static Program standard(RelMetadataProvider metadataProvider,
       boolean enableFieldTrimming) {
     final Program program1 =
--- a/core/src/test/java/org/apache/calcite/test/JdbcTest.java
+++ b/core/src/test/java/org/apache/calcite/test/JdbcTest.java
@@ -47,10 +47,13 @@ import org.apache.calcite.linq4j.Ord;
 import org.apache.calcite.linq4j.QueryProvider;
 import org.apache.calcite.linq4j.Queryable;
 import org.apache.calcite.linq4j.function.Function0;
+import org.apache.calcite.plan.RelOptLattice;
+import org.apache.calcite.plan.RelOptMaterialization;
 import org.apache.calcite.plan.RelOptPlanner;
 import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.prepare.CalcitePrepareImpl;
 import org.apache.calcite.prepare.Prepare;
+import org.apache.calcite.rel.RelNode;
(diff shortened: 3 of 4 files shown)


core/src/main/java/org/apache/calcite/prepare/Prepare.java
```diff
@@ -293,10 +293,12 @@ public PreparedResult prepareSql(
       root = root.withRel(decorrelate(sqlToRelConverter, sqlQuery, root.rel));
     }
 
-    // Trim unused fields.
-    root = trimUnusedFields(root);
+    if (configHolder.get().isTrimUnusedFields()) {
+      // Trim unused fields.
+      root = trimUnusedFields(root);
 
-    Hook.TRIMMED.run(root.rel);
```
LakeShen, line 299: I see that this is same code,why you changed this?

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -3119,6 +3125,96 @@ void testInnerJoinValues(String format) {
...
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5984">[CALCITE-5984]</a>
+   * Disabling trimming of unused fields via config and program. */
+  @Test void testJoinWithTrimmingEnabledByProgramAndConfig() {
+    CalciteAssert.hr().query("select \"d\".\"name\" from \"hr\".\"depts\" as \"d\" \n"
+                + "  join \"hr\".\"emps\" as \"e\" on \"d\".\"deptno\" = \"e\".\"deptno\" \n")
+        .withHook(Hook.SQL2REL_CONVERTER_CONFIG_BUILDER,
+            (Consumer<Holder<Config>>) configHolder ->
+                configHolder.set(configHolder.get().withTrimUnusedFields(true)))
```
LakeShen, line 3203: I see the sql is same as above,the withTrimUnusedFields method has different parameters, but the plan is the same.

Why did you use this SQL example？

WegdanGhazi: Allowing field trimming happens by

1. Specifying SQL2RelConverter Config  
**or**
2. Specifying a Program that contains the Field trimming program (the standard program)

I tried to showcase all the different ways config 1 & 2 will enable trimming, while trimming will be disabled if both 1. Converter config and 2. Program Config disable it

LakeShen: Hi @WegdanGhazi ,thanks for your reply.

I have saw the difference between `testJoinWithTrimmingEnabledByProgramAndConfig` and `testJoinWithTrimmingEnabledByProgram`  is the withTrimUnusedFields param value,one is false,another is true.

But the input sql and expected plan is same,so I'm a little confused about the purpose of these two unit tests.I thought a few code comments illustrating this case would make it easier for others to understand.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -8247,6 +8343,33 @@ private static String sums(int n, boolean c) {
     return b.toString();
   }
 
+  /** A program that omits {@link TrimFieldsProgram}. */
+  private static Program getProgram() {
```
LakeShen, line 8347: Maybe could use org.apache.calcite.tools.Programs#standard(org.apache.calcite.rel.metadata.RelMetadataProvider) method.

LakeShen:
@WegdanGhazi thank you for opening this PR, left a couple of comments!

core/src/test/java/org/apache/calcite/test/JdbcTest.java
WegdanGhazi: Got it! Will add comments, thanks!

core/src/test/java/org/apache/calcite/test/JdbcTest.java
WegdanGhazi: The standard program is initialized as follows:
```
    return sequence(subQuery(metadataProvider),
        new DecorrelateProgram(),
        new TrimFieldsProgram(),
        program1,

        // Second planner pass to do physical "tweaks". This the first time
        // that EnumerableCalcRel is introduced.
        calc(metadataProvider));
```
Which includes the `TrimFieldsProgram()`, so I had to manually initialize one to avoid adding that program.
What do you think? Do you suggest a better way to create the program without `TrimFieldsProgram()`?

LakeShen: Could we make the standard(RelMetadataProvider) method has parameters with RelMetadataProvider and `boolean` flag?

The `boolean` flag  indicates whether TrimFieldsProgram should be added,The standard() method default value is true.I finded Programs#standard(RelMetadataProvider) method only used in Programs.

WegdanGhazi: That sounds great! Will go for that!

LakeShen:
Hi @WegdanGhazi ,If you think we could add this feature in 1.36.0,please change this jira Fix Version to 1.36.0.

And also need the others review.

WegdanGhazi:
Hi @LakeShen, Thank you so much for the review! The current state is that we're waiting for more reviews, right?

LakeShen:
Yes,you are right.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -3119,6 +3122,100 @@ void testInnerJoinValues(String format) {
...
+                + "    LogicalTableScan(table=[[hr, depts]])\n"
+                + "    LogicalTableScan(table=[[hr, emps]])\n"
+                + "");
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5984">[CALCITE-5984]</a>
+   * Disabling trimming of unused fields via config and program.
+   * Case: trimmingByProgram: false ; trimmingByConfig: true. */
+  @Test void testJoinWithTrimmingEnabledByConfig() {
```
zstan, line 3151: you can use @ParameterizedTest from junit5 or make your custom input collection and unite 3 below tests into one, wdyt ?

Stephan202 (approved):
:rocket:

core/src/main/java/org/apache/calcite/tools/Programs.java
```diff
@@ -250,11 +250,12 @@ public static Program getProgram() {
 
   /** Returns the standard program used by Prepare. */
   public static Program standard() {
-    return standard(DefaultRelMetadataProvider.INSTANCE);
+    return standard(DefaultRelMetadataProvider.INSTANCE, true);
   }
 
   /** Returns the standard program with user metadata provider. */
-  public static Program standard(RelMetadataProvider metadataProvider) {
+  public static Program standard(RelMetadataProvider metadataProvider,
```
libenchao, line 257: I'm afraid that this is a public API that we need to consider compatibility.

testkit/src/main/java/org/apache/calcite/test/CalciteAssert.java
```diff
@@ -679,6 +679,7 @@ static void assertPrepare(
       Connection connection,
       String sql,
       boolean materializationsEnabled,
+      List<Pair<Hook, Consumer>> hooks,
```
libenchao, line 682: This is also a public API (testkit are expected to be used by third party projects)

WegdanGhazi: Added an overload function to maintain the publish API.

libenchao (approved):
+1

PR #3437, opened 2023-09-21
[CALCITE-6020] SqlToRelConverter should not replace windowed SUM with…

kgyrtkirk:
… equivalent SUM0

SqlToRelConverter is translating windowed SUM(...) OVER (...) aggregations into CASE WHEN COUNT(...) OVER (...) > 0 THEN SUM0(...) OVER (...) ELSE NULL END which is not necessary - as SUM should behave similarily


LakeShen:
Hi @kgyrtkirk ,I find some information in PR's title and commits are omitted, I don't know whether this is normal.

PR #3438, opened 2023-09-21
[CALCITE-5955] BigQuery PERCENTILE functions are unparsed incorrectly

tanclary:
The main solution I used for the problem I describe in the JIRA case is adding an override to `SqlBasicAggFunction` for `allowsFraming()`, this should be `FALSE` for BigQuery's percentile functions and was previously being defaulted to `TRUE`.


tanclary:
@julianhyde @olivrlee What do we think of this for handling the most basic cases of PERCENTILE? I think it is a good starting point while we don't have the functions fully implemented. If either of you could review that'd be great, thanks!

julianhyde:
I'd change `writer.print(" OVER")` to `writer.keyword("OVER")` but it looks like progress, so +1.

Please check that we are not breaking things on non-BigQuery dialects. If there's not a test for one such dialect, add one. Or merge your test methods into something that tests several dialects (BQ + others).

tanclary:
> I'd change `writer.print(" OVER")` to `writer.keyword("OVER")` but it looks like progress, so +1.
> 
> Please check that we are not breaking things on non-BigQuery dialects. If there's not a test for one such dialect, add one. Or merge your test methods into something that tests several dialects (BQ + others).

I will adjust the writer syntax, as for testing: https://issues.apache.org/jira/browse/CALCITE-6036 describes how `WITHIN GROUP() OVER()` is not supported yet for the standard `PERCENTILE` functions but no dialect that I can find allows framing anyways so this change should not break any current behavior. The `OVER` clause can still use `PARTITION BY` in this context so I will add a test that showcases that.

caicancai:
@tanclary Hello, should we add corresponding tests to SqlOperator

tanclary:
> @tanclary Hello, should we add corresponding tests to SqlOperatorTest

Could you clarify what you mean? I believe there are already tests for these functions in `SqlOperatorTest`. Also, these functions are not fully implemented so we're a bit restricted in what kind of tests we can write.

caicancai:
> > @tanclary Hello, should we add corresponding tests to SqlOperatorTest
> 
> @caicancai Could you clarify what you mean? I believe there are already tests for these functions in `SqlOperatorTest`. Also, these functions are not fully implemented so we're a bit restricted in what kind of tests we can write.

Thank you for your reply for answering my questions

PR #3444, opened 2023-09-27
[MINOR] Add example for MAP type in reference docs

chucheng92:
The background is that I was going to implement Spark's MAP function, and found that the MAP constructor embedded in calcite had no examples in docs. So try to add it here.

Background and context can be found at: https://issues.apache.org/jira/browse/CALCITE-5918

### DIFF
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -1199,8 +1199,8 @@ Note:
 |:-------- |:---------------------------|:---------------
 | ANY      | The union of all types |
 | UNKNOWN  | A value of an unknown type; used as a placeholder |
-| ROW      | Row with 1 or more columns | Example: Row(f0 int null, f1 varchar)
-| MAP      | Collection of keys mapped to values |
+| ROW      | Row with 1 or more columns | Example: row(f0 int null, f1 varchar)
+| MAP      | Collection of keys mapped to values | Example: (int, varchar) map
 | MULTISET | Unordered collection that may contain duplicates | Example: int multiset
 | ARRAY    | Ordered, contiguous collection that may contain duplicates | Example: varchar(10) array
 | CURSOR   | Cursor over the result of executing a query |


mihaibudiu (commented):
Is there a test which confirms this syntax?

chucheng92:
> Is there a test which confirms this syntax?

thanks for reviewing. Yes, we have.

https://github.com/apache/calcite/blob/aef9bdb5c091e4df84eeb47a9aa70770dd3d3fb8/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java#L6639

The background is that I was going to implement Spark's map function, and found that the MAP document embedded in calcite had no examples.

LakeShen:
LGTM

site/_docs/reference.md
```diff
@@ -1200,7 +1200,7 @@ Note:
 | ANY      | The union of all types |
 | UNKNOWN  | A value of an unknown type; used as a placeholder |
 | ROW      | Row with 1 or more columns | Example: Row(f0 int null, f1 varchar)
-| MAP      | Collection of keys mapped to values |
+| MAP      | Collection of keys mapped to values | Example: MAP\['washington', 1, 'obama', 44\]
```
JiajunBernoulli, line 1203: This example is different from other examples.

I think the example should be `map(varchar, int)`.

site/_docs/reference.md
```diff
@@ -1200,7 +1200,7 @@ Note:
 | ANY      | The union of all types |
 | UNKNOWN  | A value of an unknown type; used as a placeholder |
 | ROW      | Row with 1 or more columns | Example: Row(f0 int null, f1 varchar)
```
JiajunBernoulli, line 1202: `row` maybe better than `Row`

chucheng92: thanks. updated.

site/_docs/reference.md
chucheng92: thanks. I have updated to map[varchar, int] instead of map(varchar, int), because the previous one is calcite standard form, the later one is Spark map form. We haven't implemented this yet. WDYT?

JiajunBernoulli: @chucheng92  Do you know `generateTypeString`? Please take a look at them.
- ArraySqlType#generateTypeString
- MapSqlType#generateTypeString
- MultisetSqlType#generateTypeString

chucheng92: @JiajunBernoulli thanks. Jiajun, I get your point. I actually wanted to add an example of map function at the beginning, here is actually map type. 

I have updated the PR. thanks for reminding me.

JiajunBernoulli (approved):

site/_docs/reference.md
JiajunBernoulli: @chucheng92 
In fact, we have many map constructor examples.

https://github.com/apache/calcite/blob/6f79436c178beec639e559d9152c237bbf8ec3e8/core/src/test/resources/sql/misc.iq#L2346

https://github.com/apache/calcite/blob/6f79436c178beec639e559d9152c237bbf8ec3e8/core/src/test/resources/sql/misc.iq#L2595

chucheng92:
hi,@JiajunBernoulli I have squashed the commit, If u have time, could you help to merge this?

PR #3446, opened 2023-09-27
[CALCITE-6030] DATE_PART is not handled by the RexToLixTranslator

mihaibudiu:
Please note that running this test requires #3445 to be merged first.


mihaibudiu:
I have reworked this bugfix so that it can be merged right away, and we can debate #3445 separately.

libenchao (approved):
LGTM

PR #3451, opened 2023-10-03
[CALCITE-5387] Fix Type-mismatch on nullability in JOIN_PUSH_TRANSITIVE_PREDICATES

keatoooon:
Resolves an issue with the nullability of inputRefs in RexNodes returned by inferPredicates, which was causing JOIN_PUSH_TRANSITIVE_PREDICATES to throw errors.

The root issue was that the RexInputRefs in the left/rightInferredPredicates returned from the inferPredicates function were using the types of the join, instead of using the types from the left/right inputs respectively. When the type of the join node had a different nullability than the left/right inputs, the InputRefs would have the wrong nullability. In JOIN_PUSH_TRANSITIVE_PREDICATES, we use the left/right predicates in  to construct filters on the left/right inputs. The relBuilder would throw an error because in the RexInputRefs in filter condition, have a different nullability than the input to the filter node.

Background and context can be found at: https://issues.apache.org/jira/browse/CALCITE-5387

I have just applied for JIRA access, I will update the ticket when I can.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
@@ -56,6 +56,7 @@ import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.util.BitSets;
 import org.apache.calcite.util.Bug;
 import org.apache.calcite.util.ImmutableBitSet;
+import org.apache.calcite.util.Pair;
 import org.apache.calcite.util.Util;
 import org.apache.calcite.util.mapping.Mapping;
 import org.apache.calcite.util.mapping.MappingType;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -666,24 +667,22 @@ static class JoinConditionBasedPredicateInference {
       equivalence = BitSets.closure(equivalence);
     }
 
+
```
keatoooon, line 670: While I was debugging, I refactored `inferPredicates` into several helper functions to help me understand what was going on. I think that these changes are good, but it does complicates the git diff. I've left comments to highlight the real changes.

If needed, I can remove the refactoring from this PR and do it separately.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -710,16 +709,63 @@ public RelOptPredicateList inferPredicates(
...
               0, nSysFields + nFieldsLeft, nFieldsRight);
       final RexPermuteInputsShuttle rightPermute =
-          new RexPermuteInputsShuttle(rightMapping, joinRel);
+          new TypeChangingRexPermuteInputsShuttle(rightMapping, joinRel.getRight());
       Mappings.TargetMapping leftMapping =
           Mappings.createShiftMapping(nSysFields + nFieldsLeft, 0, nSysFields,
               nFieldsLeft);
       final RexPermuteInputsShuttle leftPermute =
-          new RexPermuteInputsShuttle(leftMapping, joinRel);
+          new TypeChangingRexPermuteInputsShuttle(leftMapping, joinRel.getLeft());
```
keatoooon, line 768: The changes to these two lines are the only real change.

keatoooon:
Going to split this into two PRs to simplify the git diff

PR #3452, opened 2023-10-03
[CALCITE-5387] Type-mismatch on nullability in JoinPushTransitivePredicatesRule RelRule

keatoooon:
Resolves an issue with the nullability of inputRefs in RexNodes returned by inferPredicates, which was causing JOIN_PUSH_TRANSITIVE_PREDICATES to throw errors.

The root issue was that the RexInputRefs in the left/rightInferredPredicates returned from the inferPredicates function were using the types of the join, instead of using the types from the left/right inputs respectively. When the type of the join node had a different nullability than the left/right inputs, the InputRefs would have the wrong nullability. In JOIN_PUSH_TRANSITIVE_PREDICATES, we use the left/right predicates in to construct filters on the left/right inputs. The relBuilder would throw an error because in the RexInputRefs in filter condition, have a different nullability than the input to the filter node.

Background and context can be found at: https://issues.apache.org/jira/browse/CALCITE-5387

I have just applied for JIRA access, I will update the ticket when I can.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
@@ -670,8 +670,8 @@ public class RelMdPredicates
      * As RexPermuteInputsShuttle, with one exception. When visiting an inputRef,
      * it will replace the type of the InputRef with the type found in the input fields,
      * instead of keeping the original type. This is used within
-     * generateLeftRightInferredPredicates, to avoid nullability mismatches between the types of
-     * the join and the types of the inputs.
+     * when generating the Left/RightInferredPredicates, to avoid nullability mismatches
+     * between the types of the join and the types of the inputs.
      */
     private class TypeChangingRexPermuteInputsShuttle
         extends org.apache.calcite.rex.RexPermuteInputsShuttle {
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
@@ -39,6 +39,7 @@ import org.apache.calcite.rel.core.Sort;
 import org.apache.calcite.rel.core.TableModify;
 import org.apache.calcite.rel.core.TableScan;
 import org.apache.calcite.rel.core.Union;
+import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexExecutor;
@@ -667,11 +668,11 @@ public class RelMdPredicates
     }
 
     /**
-     * As RexPermuteInputsShuttle, with one exception. When visiting an inputRef,
-     * it will replace the type of the InputRef with the type found in the input fields,
-     * instead of keeping the original type. This is used within
-     * when generating the Left/RightInferredPredicates, to avoid nullability mismatches
-     * between the types of the join and the types of the inputs.
+     * As RexPermuteInputsShuttle, with the modification that an RexInputRef is updated with
+     * the type found in the input fields (instead of keeping the original type).
+     * This is use when generating the Left/RightInferredPredicates, permitting different but
+     * compatible types for expressions which are moved around the tree (such as moving an
+     * expression to a location where an input type's nullability is different).
      */
     private class TypeChangingRexPermuteInputsShuttle
         extends org.apache.calcite.rex.RexPermuteInputsShuttle {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -666,6 +666,29 @@ static class JoinConditionBasedPredicateInference {
...
+      TypeChangingRexPermuteInputsShuttle(Mappings.TargetMapping mapping, RelNode... inputs) {
+        super(mapping, inputs);
+      }
+
+      @Override public RexNode visitInputRef(RexInputRef local) {
+        final int index = local.getIndex();
+        int target = mapping.getTarget(index);
+        return new RexInputRef(
+            target,
+            fields.get(target).getType());
```
ian-bertolacci, line 688: It would be nice if we were able to check that the original type and the input type are at least compatible, or only change w.r.t. nullability.
That being said, I don't think theres a lot of type checking that happens at the RexNode level.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -666,6 +666,29 @@ static class JoinConditionBasedPredicateInference {
       equivalence = BitSets.closure(equivalence);
     }
 
+    /**
```
ian-bertolacci, line 669: An optional suggestion for the comment:

> As RexPermuteInputsShuttle, with the modification that an RexInputRef is updated with the type the type found in the input fields (instead of keeping the original type).
> This is use when generating the Left/RightInferredPredicates, permitting different but compatible types for expressions which are moved around the tree (such as moving an expression to a location where a an input type's nullability is different).

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -666,6 +666,29 @@ static class JoinConditionBasedPredicateInference {
...
+     * As RexPermuteInputsShuttle, with one exception. When visiting an inputRef,
+     * it will replace the type of the InputRef with the type found in the input fields,
+     * instead of keeping the original type. This is used within
+     * when generating the Left/RightInferredPredicates, to avoid nullability mismatches
+     * between the types of the join and the types of the inputs.
+     */
+    private class TypeChangingRexPermuteInputsShuttle
+        extends org.apache.calcite.rex.RexPermuteInputsShuttle {
```
ian-bertolacci, line 677: Is there a situation where the behavior of `TypeChangingRexPermuteInputsShuttle.visitInputRef` is *not* correct for the regular `RexPermuteInputsShuttle`? 
If not, then maybe this should just be the behavior of `RexPermuteInputsShuttle`

LakeShen:
Hi @keatoooon ,a small suggestion,please keep the same title with your jira's title.

And I think that `Type-mismatch on nullability in JoinPushTransitivePredicatesRule RelRule` title maybe better.

Generally, a title should not start with Fix, but should describe the specific behavior of the problem.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
keatoooon: I tried doing something like this in `RexPermuteInputsShuttle` 

```
if (fields.isEmpty()) {
      return new RexInputRef(
          target,
          local.getType());
    } else {
      return new RexInputRef(
          target,
          fields.get(target).getType());
    }
```
(Note that the if statement is needed because it's valid to create a `RexPermuteInputsShuttle` without any backing fields.)

This caused several tests to fail with index out of bound errors. The reason wasn't obvious to me. I'll do a bit of poking around and see if I can figure out why.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
keatoooon: I think in order to check that the two types are equivalent except for nullability, I would need to have access to a typefactory, so I could do something like `typefatory.createTypeWithNullability(t1, t2.isNullable()).equals(t1)`.

For now, I added a sanity check to confirm that the two types are of the same type family. I'm not certain if this is a good check for compatibility, and am open to changing it.

LakeShen:
More details could see developing Calcite guide:https://calcite.apache.org/develop/

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
ian-bertolacci: Maybe there are situations where fields is not empty, but also not the same size as the number of input columns?
I could see something like this working:
```java
RelDataType dataType = target < fields.size() ? local.getType() : fields.get(target).getType();
```

mihaibudiu:
Can it be revived?

mihaibudiu:
Closing in favor of https://github.com/apache/calcite/pull/4350

PR #3457, opened 2023-10-05
[CALCITE-6021] Add CURRENT_DATETIME function (enabled in BigQuery library)

tanclary:


core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -995,7 +995,7 @@ private List<RexNode> toRexList(RelInput relInput, List operands) {
     operatorTable.lookupOperatorOverloads(
         new SqlIdentifier(name, new SqlParserPos(0, 0)),
         null,
-        sqlSyntax,
+        sqlSyntax.family,
```
tjbanghart, line 998: Why use `family` now? 

Could this change the expected output of JSON -> Rex translations?

tanclary: This change came from one of Julian's commits. I think it's to account for the discrepancy from his last comment on CALCITE-6021. pushing a commit momentarily to try and fix this

tanclary: See https://github.com/apache/calcite/pull/3457/commits/31fb1970ab558eabc308f7d4ab18f18dace36937 for my fix for this

babel/src/test/resources/sql/big-query.iq
```diff
@@ -148,23 +148,25 @@ SELECT current_date() AS the_date, t.current_date FROM t;
...
 !ok
 
-SELECT CURRENT_DATETIME as now;
-+----------------------------+
-| now                        |
-+----------------------------+
-| 2016-05-19T10:38:47.046465 |
-+----------------------------+
+SELECT CURRENT_DATETIME > DATETIME '2008-12-25 15:30:00' as now;
++------+
```
julianhyde, line 162: Would it be useful if Quidem had a way to mask out results? E.g.
```
!mask 6 "[0-9]" "x"
SELECT CURRENT_DATETIME as now;
+----------------------------+
| now                        |
+----------------------------+
| 2016-05-19T10:38:47.046465 |
+----------------------------+
```
would succeed because '2016-05-19T10:38:47.046465' and '2023-11-11T16:38:47.012345' would both be converted to 'xxxx-xx-xxTxx:xx:xx.xxxxxx'. The 6 means 'the next 6 lines'.

I logged https://github.com/julianhyde/quidem/issues/79. Please add comments there.

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
```diff
@@ -682,6 +682,9 @@ public enum BuiltInMethod {
   CURRENT_TIMESTAMP(SqlFunctions.class, "currentTimestamp", DataContext.class),
   CURRENT_TIME(SqlFunctions.class, "currentTime", DataContext.class),
   CURRENT_DATE(SqlFunctions.class, "currentDate", DataContext.class),
+  CURRENT_DATETIME(SqlFunctions.class, "currentDatetime", DataContext.class),
+  CURRENT_DATETIME2(SqlFunctions.class, "currentDatetime", DataContext.class,
+    String.class),
```
julianhyde, line 687: indent 4, not 2

core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
julianhyde: I agree with @tjbanghart that this change is mysterious. You should explain it in a commit comment.

Something like the fact that CURRENT_DATETIME has FUNCTION_ID syntax (to allow being called without parentheses) but is registered in the op table as a FUNCTION. I don't recall, but it's worth digging up the rationale.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -1605,25 +1605,23 @@ void testLikeAndSimilarFails() {
...
-  @Test void testCurrentDatetime() throws SqlParseException, ValidationException {
+  @Test void testCurrentDatetime() {
     final String currentDateTimeExpr = "select ^current_datetime^";
     SqlValidatorFixture shouldFail = sql(currentDateTimeExpr)
         .withConformance(SqlConformanceEnum.BIG_QUERY);
     final String expectedError = "query [select CURRENT_DATETIME]; exception "
         + "[Column 'CURRENT_DATETIME' not found in any table]; class "
         + "[class org.apache.calcite.sql.validate.SqlValidatorException]; pos [line 1 col 8 thru line 1 col 8]";
-    shouldFail.fails("Column 'CURRENT_DATETIME' not found in any table");
+//    shouldFail.fails("Column 'CURRENT_DATETIME' not found in any table");
```
julianhyde: remove the comment or the line

julianhyde (approved):
LGTM when you've fixed some cosmetic changes.

core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
tanclary: I removed this change in a later commit

babel/src/test/resources/sql/big-query.iq
tanclary: Left a comment on the issue, sounds like a good idea.

PR #3458, opened 2023-10-06
[CALCITE-6037] The function category of ARRAY/EXTRACT_VALUE/XML_TRANSFORM/EXTRACT_XML/EXISTSNODE is incorrect

chucheng92:
https://issues.apache.org/jira/browse/CALCITE-6037

the definition of function category in ANSI-SQL(not very sure this is the sql standard, but the implementation of calcite is similar to it): 
https://www.oreilly.com/library/view/sql-in-a/9780596155322/ch04s04.html.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -532,26 +532,30 @@ public abstract class SqlLibraryOperators {
   public static final SqlFunction EXTRACT_VALUE =
       SqlBasicFunction.create("EXTRACTVALUE",
           ReturnTypes.VARCHAR_2000.andThen(SqlTypeTransforms.FORCE_NULLABLE),
-          OperandTypes.STRING_STRING);
+          OperandTypes.STRING_STRING,
+          SqlFunctionCategory.STRING);
 
   @LibraryOperator(libraries = {ORACLE})
   public static final SqlFunction XML_TRANSFORM =
       SqlBasicFunction.create("XMLTRANSFORM",
           ReturnTypes.VARCHAR.andThen(SqlTypeTransforms.FORCE_NULLABLE),
-          OperandTypes.STRING_STRING);
+          OperandTypes.STRING_STRING,
+          SqlFunctionCategory.STRING);
 
   @LibraryOperator(libraries = {ORACLE})
   public static final SqlFunction EXTRACT_XML =
       SqlBasicFunction.create("EXTRACT",
           ReturnTypes.VARCHAR.andThen(SqlTypeTransforms.FORCE_NULLABLE),
-          OperandTypes.STRING_STRING_OPTIONAL_STRING);
+          OperandTypes.STRING_STRING_OPTIONAL_STRING,
+          SqlFunctionCategory.STRING);
 
   @LibraryOperator(libraries = {ORACLE})
   public static final SqlFunction EXISTS_NODE =
       SqlBasicFunction.create("EXISTSNODE",
           ReturnTypes.INTEGER_NULLABLE
               .andThen(SqlTypeTransforms.FORCE_NULLABLE),
-          OperandTypes.STRING_STRING_OPTIONAL_STRING);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -532,26 +532,30 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
   public static final SqlFunction EXTRACT_VALUE =
       SqlBasicFunction.create("EXTRACTVALUE",
           ReturnTypes.VARCHAR_2000.andThen(SqlTypeTransforms.FORCE_NULLABLE),
-          OperandTypes.STRING_STRING);
+          OperandTypes.STRING_STRING,
```
LakeShen, line 535: image

I find that EXTRACT_VALUE function would return the numeric value in some Scene.

I don't know whether it is appropriate to divide it into a String type function. I saw that it was divided into numeric function before.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1075,7 +1079,8 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
   public static final SqlFunction ARRAY =
       SqlBasicFunction.create("ARRAY",
           SqlLibraryOperators::arrayReturnType,
-          OperandTypes.SAME_VARIADIC);
+          OperandTypes.SAME_VARIADIC,
+          SqlFunctionCategory.SYSTEM);
```
LakeShen, line 1083: what is meaning of System function?And why array function belong to system function.

LakeShen:
Hi @chucheng92 ,Thank you for opening this PR, left a couple of comments!

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
chucheng92: @LakeShen I get your point. Usually we classify functions according to the operands (in some cases, they can also be determined according to the return value type), which is not clearly specified by calcite. So we need to take a comprehensive view. These functions actually process XML. In database systems such as MySql or Oracle, they are divided into XML functions. The closest type in calcite is String.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
chucheng92: current the function type of ARRAY is Numeric currently. Obviously not right. The closest type in calcite is System. it means built-in or system-in. we can see many other functions if not belongs to timedate/string/numeric will let it be System.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
mihaibudiu: I have a meta-question: what is this field used for?
It doesn't seem to affect any of the tests.
Clearly, the meaning of this field is ambiguous too - it's not clear what the right choice is in all cases.
Can we just remove it completely?

chucheng92: Hi, @mihaibudiu. thanks for reviewing. good question. 

Frankly speaking, `SqlFunctionCategory` is a very loose field now. In fact, there is no strict verification during lookup, so no test case covers it. However, it is currently widely used and cannot be deleted for subsequent expansion. We can refer to the comments of the current `lookupOperatorOverloads` and use it immediately when we need it.

https://github.com/apache/calcite/blob/597b53dbb075ab72700d40f66db81d2ac9c182b8/core/src/main/java/org/apache/calcite/sql/util/ReflectiveSqlOperatorTable.java#L94

JiajunBernoulli (approved):

JiajunBernoulli:
For me, this is an improvement.

If there is no objection, I will merge it.

chucheng92:
> For me, this is an improvement.
> 
> If there is no objection, I will merge it.

thanks for approve + 1.

PR #3459, opened 2023-10-07
[CALCITE-5918] Add MAP function (enabled in Spark library)

chucheng92:
https://issues.apache.org/jira/browse/CALCITE-5918

the spark map function has 2 different points with calcite map constructor:
1. use map(k, v, ...) rather than map[k, v, ...]
2. allows empty map such as map(); however calcite map constructor not support it.

### DIFF
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -4886,14 +4886,21 @@ SqlNode MapConstructor() :
 {
     <MAP> { s = span(); }
     (
-        LOOKAHEAD(1)
-        <LPAREN>
-        // by sub query "MAP (SELECT empno, deptno FROM emp)"
-        e = LeafQueryOrExpr(ExprContext.ACCEPT_QUERY)
-        <RPAREN>
+        (
+            // empty map function call: "map()"
+            LOOKAHEAD(2)
+            <LPAREN> <RPAREN> { args = SqlNodeList.EMPTY; }
+        |
+            args = ParenthesizedQueryOrCommaList(ExprContext.ACCEPT_ALL)
+        )
         {
-            return SqlStdOperatorTable.MAP_QUERY.createCall(
-                s.end(this), e);
+            if (args.size() == 1 && args.get(0).isA(SqlKind.QUERY)) {
+                // MAP query constructor e.g. "MAP (SELECT empno, deptno FROM emps)"
+                return SqlStdOperatorTable.MAP_QUERY.createCall(s.end(this), args.get(0));
+            } else {
+                // MAP function e.g. "MAP(1, 2)" equivalent to standard "MAP[1, 2]"
+                return SqlLibraryOperators.MAP.createCall(s.end(this), args.getList());
+            }
         }
     |
         // by enumeration "MAP[k0, v0, ..., kN, vN]"
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -205,6 +205,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_AND;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_OR;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LPAD;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_CONCAT;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_ENTRIES;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_FROM_ARRAYS;
@@ -880,6 +881,7 @@ public class RexImpTable {
       map.put(MAP_VALUE_CONSTRUCTOR, value);
       map.put(ARRAY_VALUE_CONSTRUCTOR, value);
       defineMethod(ARRAY, BuiltInMethod.ARRAYS_AS_LIST.method, NullPolicy.NONE);
+      defineMethod(MAP, BuiltInMethod.MAP.method, NullPolicy.NONE);
 
       // ITEM operator
       map.put(ITEM, new ItemImplementor());
--- a/core/src/main/java/org/apache/calcite/prepare/LixToRelTranslator.java
+++ b/core/src/main/java/org/apache/calcite/prepare/LixToRelTranslator.java
@@ -97,7 +97,7 @@ class LixToRelTranslator {
   public RelNode translate(Expression expression) {
     if (expression instanceof MethodCallExpression) {
       final MethodCallExpression call = (MethodCallExpression) expression;
-      BuiltInMethod method = BuiltInMethod.MAP.get(call.method);
+      BuiltInMethod method = BuiltInMethod.FUNCTIONS_MAPS.get(call.method);
       if (method == null) {
         throw new UnsupportedOperationException(
             "unknown method " + call.method);
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -5323,6 +5323,20 @@ public class SqlFunctions {
     return map;
   }
 
+  /** Support the MAP function.
+   *
+   * <p>odd-indexed elements are keys and even-indexed elements are values.
+   */
+  public static Map map(Object... args) {
+    final Map map = new LinkedHashMap<>();
+    for (int i = 0; i < args.length; i += 2) {
+      Object key = args[i];
+      Object value = args[i + 1];
+      map.put(key, value);
+    }
+    return map;
+  }
+
   /** Support the STR_TO_MAP function. */
   public static Map strToMap(String string, String stringDelimiter, String keyValueDelimiter) {
     final Map map = new LinkedHashMap();
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -46,7 +46,9 @@ import org.apache.calcite.sql.validate.SqlValidator;
 import org.apache.calcite.sql.validate.SqlValidatorUtil;
 import org.apache.calcite.util.Litmus;
 import org.apache.calcite.util.Optionality;
+import org.apache.calcite.util.Pair;
 import org.apache.calcite.util.Static;
+import org.apache.calcite.util.Util;
 
 import com.google.common.collect.ImmutableList;
 
@@ -1081,6 +1083,44 @@ public abstract class SqlLibraryOperators {
       SqlBasicFunction.create("ARRAY",
           SqlLibraryOperators::arrayReturnType,
           OperandTypes.SAME_VARIADIC);
+
+  private static RelDataType mapReturnType(SqlOperatorBinding opBinding) {
+    Pair<@Nullable RelDataType, @Nullable RelDataType> type =
+        getComponentTypes(
+            opBinding.getTypeFactory(), opBinding.collectOperandTypes());
+    return SqlTypeUtil.createMapType(
+        opBinding.getTypeFactory(),
+        requireNonNull(type.left, "inferred key type"),
+        requireNonNull(type.right, "inferred value type"),
+        false);
+  }
+
+  private static Pair<@Nullable RelDataType, @Nullable RelDataType> getComponentTypes(
+      RelDataTypeFactory typeFactory,
+      List<RelDataType> argTypes) {
+    // special case, allows empty map
+    if (argTypes.isEmpty()) {
+      return Pair.of(typeFactory.createUnknownType(), typeFactory.createUnknownType());
+    }
+    // Util.quotientList(argTypes, 2, 0):
+    // This extracts all elements at even indices from argTypes.
+    // It represents the types of keys in the map as they are placed at even positions
+    // e.g. 0, 2, 4, etc.
+    // Symmetrically, Util.quotientList(argTypes, 2, 1) represents odd-indexed elements.
+    // details please see Util.quotientList.
+    return Pair.of(
+        typeFactory.leastRestrictive(Util.quotientList(argTypes, 2, 0)),
+        typeFactory.leastRestrictive(Util.quotientList(argTypes, 2, 1)));
+  }
+
+  /** The "MAP(key, value, ...)" function (Spark);
+   * compare with the standard map value constructor, "MAP[key, value, ...]". */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction MAP =
+      SqlBasicFunction.create("MAP",
+          SqlLibraryOperators::mapReturnType,
+          OperandTypes.MAP_FUNCTION,
+          SqlFunctionCategory.SYSTEM);
 
   @SuppressWarnings("argument.type.incompatible")
   private static RelDataType arrayAppendPrependReturnType(SqlOperatorBinding opBinding) {
--- a/core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
@@ -29,6 +29,7 @@ import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.SqlUtil;
 import org.apache.calcite.sql.validate.SqlValidatorScope;
 import org.apache.calcite.util.ImmutableIntList;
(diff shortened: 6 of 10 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1077,6 +1079,38 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
...
+        requireNonNull(type.right, "inferred value type"),
+        false);
+  }
+
+  private static Pair<@Nullable RelDataType, @Nullable RelDataType> getComponentTypes(
+      RelDataTypeFactory typeFactory,
+      List<RelDataType> argTypes) {
+    // special case, allows empty map
+    if (argTypes.size() == 0) {
+      return Pair.of(typeFactory.createUnknownType(), typeFactory.createUnknownType());
```
macroguo-ghy, line 1098: There is a subtle difference between `unknown` and `any`, can you please explain why `unknown` is used?

site/_docs/reference.md
```diff
@@ -2769,6 +2769,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | TO_HEX(binary)                                 | Converts *binary* into a hexadecimal varchar
 | b | FROM_HEX(varchar)                              | Converts a hexadecimal-encoded *varchar* into bytes
 | b o | LTRIM(string)                                | Returns *string* with all blanks removed from the start
+| s | MAP(key, value [, key, value]*)                | Returns a map with the given key/value pairs
```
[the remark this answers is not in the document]
chucheng92: yes, good point. I will fix it later.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -4886,14 +4886,21 @@ SqlNode MapConstructor() :
...
+        )
         {
-            return SqlStdOperatorTable.MAP_QUERY.createCall(
-                s.end(this), e);
+            if (args.size() == 1 && args.get(0).isA(SqlKind.QUERY)) {
+                // MAP query constructor e.g. "MAP (SELECT empno, deptno FROM emps)"
+                return SqlStdOperatorTable.MAP_QUERY.createCall(s.end(this), args.get(0));
+            } else {
+                // MAP function e.g. "MAP(1, 2)" equivalent to standard "MAP[1, 2]"
+                return SqlLibraryOperators.MAP.createCall(s.end(this), args.getList());
```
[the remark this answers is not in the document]
chucheng92: I hope to retain the original parser logic here, because it will be detected later on OperandTypeChecker. And here is similar with Array handling. pls see: https://github.com/apache/calcite/pull/3141/files

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6604,11 +6605,44 @@ private static void checkIf(SqlOperatorFixture f) {
     // test operands not in same type family.
     f.checkFails("^map_concat(map[1, null], array[1])^",
         "Parameters must be of the same type", false);
+
+    // 2. check with map function, map(k, v ...)
```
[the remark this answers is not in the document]
chucheng92: yes, good point. I will fix it later.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
chucheng92: because Spark use UNKNOWN type when it's empty, it's similar with current spark ARRAY handling. pls see: https://github.com/apache/calcite/pull/3141/files

chucheng92:
@macroguo-ghy hi, thanks for reviewing. 
Because we have implemented Spark's ARRAY function, their processing logic on parser and empty array/map is similar, if compare with 2 PRs to review may be more clearer.

https://github.com/apache/calcite/pull/3141/files

site/_docs/reference.md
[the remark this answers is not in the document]
chucheng92: Fixed.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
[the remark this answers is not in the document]
chucheng92: updated, but currently calcite not support 'cast(map() as map<string, int>)', so I have added the case in IF block until we support it.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1082,6 +1084,38 @@ private static RelDataType arrayReturnType(SqlOperatorBinding opBinding) {
...
+        requireNonNull(type.left, "inferred key type"),
+        requireNonNull(type.right, "inferred value type"),
+        false);
+  }
+
+  private static Pair<@Nullable RelDataType, @Nullable RelDataType> getComponentTypes(
+      RelDataTypeFactory typeFactory,
+      List<RelDataType> argTypes) {
+    // special case, allows empty map
+    if (argTypes.size() == 0) {
```
tanclary, line 1102: should we use `isEmpty()` instead?

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -568,6 +569,9 @@ public static SqlOperandTypeChecker variadic(
   public static final SqlSingleOperandTypeChecker MAP_FROM_ENTRIES =
       new MapFromEntriesOperandTypeChecker();
 
+  public static final SqlSingleOperandTypeChecker MAP_FUNCTION =
```
tanclary, line 572: same naming comment as above

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -1221,6 +1225,53 @@ private static class MapFromEntriesOperandTypeChecker
...
   }
 
+  /**
+   * Operand type-checking strategy for a MAP function, it allows empty map.
+   */
+  private static class MapFunctionOperandTypeChecker
+      extends SameOperandTypeChecker {
+
+    MapFunctionOperandTypeChecker() {
+      super(-1);
```
tanclary, line 1235: where does the -1 come from?

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -1221,6 +1225,53 @@ private static class MapFromEntriesOperandTypeChecker
...
+    }
+
+    @Override public boolean checkOperandTypes(final SqlCallBinding callBinding,
+        final boolean throwOnFailure) {
+      final List<RelDataType> argTypes =
+          SqlTypeUtil.deriveType(callBinding, callBinding.operands());
+      // allows empty map
+      if (argTypes.size() == 0) {
```
tanclary, line 1243: `isEmpty()`

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -1221,6 +1225,53 @@ private static class MapFromEntriesOperandTypeChecker
...
+        return true;
+      }
+      // the size of map arg types must be even.
+      if (argTypes.size() % 2 > 0) {
```
tanclary, line 1247: is the condition more clear if we say `argTypes.size() % 2 != 0`

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -1221,6 +1225,53 @@ private static class MapFromEntriesOperandTypeChecker
...
+    }
+
+    /**
+     * Extract the key type and value type of arg types.
+     */
+    private static Pair<@Nullable RelDataType, @Nullable RelDataType> getComponentTypes(
+        RelDataTypeFactory typeFactory,
+        List<RelDataType> argTypes) {
+      return Pair.of(
+          typeFactory.leastRestrictive(Util.quotientList(argTypes, 2, 0)),
```
tanclary, line 1270: Where are these constants coming from and what do they signify?

site/_docs/reference.md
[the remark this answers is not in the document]
tanclary: Also when referencing the function arguments please surround them with * to remain consistent with other functions

site/_docs/reference.md
```diff
@@ -2769,6 +2769,8 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | TO_HEX(binary)                                 | Converts *binary* into a hexadecimal varchar
 | b | FROM_HEX(varchar)                              | Converts a hexadecimal-encoded *varchar* into bytes
 | b o | LTRIM(string)                                | Returns *string* with all blanks removed from the start
+| s | MAP()                                          | Returns a empty map
```
tanclary, line 2772: nit: an

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6703,11 +6704,53 @@ private static void checkIf(SqlOperatorFixture f) {
...
+
+    // test concat with empty map, in fact, spark will return MAP(CHAR, INTEGER), because
+    // it will add a cast 'cast(map() as map<string, int>)' to convert UNKNOWN type,
+    // but currently calcite not support cast to map type.
+    f1.checkScalar("map_concat(map('foo', 1), map())", "{foo=1}",
+        "(UNKNOWN NOT NULL, UNKNOWN NOT NULL) MAP NOT NULL");
+
+    // after calcite supports cast(null as map<string, int>), cast(map() as map<string, int>)
+    // it should add these tests.
+    if (TODO) {
```
tanclary, line 6732: instead of `if (TODO)` would it be better to add an enum to the `Bug` class? Here is an example of what I mean: https://github.com/apache/calcite/blob/main/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java#L624

tanclary (changes requested):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -5251,6 +5251,17 @@ public static Map mapFromArrays(List keysArray, List valuesArray) {
     return map;
   }
 
+  /** Support the MAP function. */
+  public static Map mapFunction(Object... args) {
+    final Map map = new LinkedHashMap<>();
+    for (int i = 0; i < args.length; i++) {
+      Object key = args[i++];
```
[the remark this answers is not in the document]
chucheng92: yes, updated.

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
chucheng92: Util.quotientList(argTypes, 2, 0):
This extracts all elements at even indices from argTypes.
It represents the types of keys in the map as they are placed at even positions
e.g. 0, 2, 4, etc. details please see Util.quotientList.

I've added some comments here for better readability.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
chucheng92: I get your point. but the form such as `cast(map() as map<string, int>)` may not necessarily be a bug, and there is currently no JIRA to track it, calcite is not ready to support it yet, and may not support it, so it is kept as a TODO for the time being.

site/_docs/reference.md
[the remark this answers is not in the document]
chucheng92: yes, fixed.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -873,6 +874,7 @@ Builder populate2() {
       map.put(MAP_VALUE_CONSTRUCTOR, value);
       map.put(ARRAY_VALUE_CONSTRUCTOR, value);
       defineMethod(ARRAY, BuiltInMethod.ARRAYS_AS_LIST.method, NullPolicy.NONE);
+      defineMethod(MAP, BuiltInMethod.MAP_FUNCTION.method, NullPolicy.NONE);
```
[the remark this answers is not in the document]
chucheng92: In fact, there is a MAP object in BuiltInMethod, occupy this name. but yes, if we rename that, we can use MAP. I have updated it. WDYT?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -5251,6 +5251,17 @@ public static Map mapFromArrays(List keysArray, List valuesArray) {
     return map;
   }
 
+  /** Support the MAP function. */
+  public static Map mapFunction(Object... args) {
```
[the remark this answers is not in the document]
chucheng92: fixed.

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
chucheng92: fixed.

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
chucheng92: fixed.

chucheng92:
hi, @tanclary I have addressed all your comments except 
`instead of if (TODO) would it be better to add an enum to the Bug class?` this one.

I have rebased the main accidentally, I hope it won’t cause any trouble to your review. 

If you have time, pls help me to review it ?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
chucheng92: yes, fixed

chucheng92:
I have rebased to the main to use [CALCITE-5570](https://issues.apache.org/jira/browse/CALCITE-5570), it supports the form like 'cast(null as map<int, int>)'

chucheng92:
hi, @tanclary Sorry to pin you, I have solved the comments. could you help to re-check it?

core/src/main/java/org/apache/calcite/prepare/LixToRelTranslator.java
```diff
@@ -97,7 +97,7 @@ public <T> RelNode translate(Queryable<T> queryable) {
   public RelNode translate(Expression expression) {
     if (expression instanceof MethodCallExpression) {
       final MethodCallExpression call = (MethodCallExpression) expression;
-      BuiltInMethod method = BuiltInMethod.MAP.get(call.method);
+      BuiltInMethod method = BuiltInMethod.FUNCTIONS_MAPS.get(call.method);
```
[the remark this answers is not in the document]
chucheng92: because this is a static field in BuiltInMethod to store all the functions. it use 'MAP' name before, so I can't continue to use this name as our spark map method name as you suggested. (I set MAP_FUNCTION previously).  

one disadvantage I can think of here is that it may cause compatibility issues? If there is an upper-layer application that reflects the use of this field. 

or should i reset spark MAP function to 'MAP_FUNCTION' name? what do you think?

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
JiajunBernoulli: Can we add UT for `typeFactory.leastRestrictive`?

How about?
`MAP('k1',  1, 'k2', 2.0, 'k3', '3')`

chucheng92: > Can we add UT for `typeFactory.leastRestrictive`?
> 
> How about? `MAP('k1', 1, 'k2', 2.0, 'k3', '3')`

yes, added it. thanks for reviewing.

tanclary (commented):
I think this looks good, you might find some of the sonarcloud reports helpful, it looks like they have some suggestions on refactoring the `for` loops and providing types for the generic `Map`.

chucheng92:
@tanclary 

Thanks for the reminder. I checked sonar report and problems are roughly divided into three categories:
1. Assignment of index i in for-loop
2. The MAP of SqlFunctions does not specify a generic type
3. Reuse string constants

I have modified 1, but 2 and 3 actually not applicable for us. 

2: The methods of SqlFunctions are all used by reflection. It can be written as generics or not. However, the current collection types Map and List are expressed using raw, and they tend to be consistent with them.

3: is something like "CHAR(10) NOT NULL", there are multiple checkScalar, sonar wants us to define a variable and reuse it. In fact, it is not necessary. The current way of writing is more intuitive.

If you agree with these, I will squash the commits. WDYT?

tanclary (approved):
LGTM, think there might be some checkstyle issues but feel free to squash and ping me after!

chucheng92:
thank you all for patient reviewing! all comments are resolved. ci passed.

@tanclary could you help to merge it? so we can add it in up-coming 1.36.

tanclary (approved):
LGTM, sorry for the delay! Congrats on becoming a committer!

chucheng92:
> LGTM, sorry for the delay! Congrats on becoming a committer!

thanks tanner. you helped me to a lot. :)

PR #3466, opened 2023-10-13
[CALCITE-6035] Unparse 'WITHIN GROUP' for BigQuery dialect to match BigQuery documentation

olivrlee:


core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
```diff
@@ -292,6 +295,17 @@ private static void unparseItem(SqlWriter writer, SqlCall call, int leftPrec) {
     writer.endList(frame);
   }
 
+  private static void unparseWithinGroup(SqlWriter writer, SqlCall call, int leftPrec, int rightPrec) {
+    assert call.operandCount() == 2;
+    call.operand(0).unparse(writer, 0, 0);
+    writer.keyword("OVER");
```
tanclary, line 301: so I think you're more familiar with these clauses than me, is it guaranteed that the WITHIN GROUP translates to BQ's OVER in every case? are they really just synonyms for each other, I mean

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1294,6 +1294,18 @@ private static String toSql(RelNode root, SqlDialect dialect,
         .withPostgresql().ok(expectedPostgresql);
   }
 
+  @Test void testPercentileContWithinGroupClauseBigQuery() {
+    final String query = "select percentile_cont(0.5) WITHIN GROUP (ORDER BY `product_class_id`)\n"
+        + "from `foodmart`.`product`";
+    final String expected = "SELECT PERCENTILE_CONT(0.5) OVER (ORDER BY product_class_id NULLS LAST)\n"
+        + "FROM foodmart.product";
+    final SqlParser.Config parserConfig =
```
tanclary, line 1302: May want to fix the style errors

tanclary (commented):
Looks good! Just left a couple questions

core/src/main/java/org/apache/calcite/sql/dialect/BigQuerySqlDialect.java
olivrlee: So I detailed in this [comment](https://issues.apache.org/jira/browse/CALCITE-5955?focusedCommentId=17772373&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17772373) how Calcite currently only supports 
`select percentile_cont(0.5) within group (order by X)`
so I want to say that for what is supported at this moment, it is sufficient

tanclary: Ohhh ok. So if Calcite ended up supporting the `partition by` case (your `input_2` from that comment), do you know if this would break?

olivrlee: Julian had a comment [here](https://issues.apache.org/jira/browse/CALCITE-6039) about how we should approach the followup tickets
I think it won't work out of the box, there are things to figure out 
We would still be parsing it as `within_group` in the parser, but want it to be handled differently in the validator.

 
questions to then consider: would we still use a SqlWithinGroupOperator? If yes, then that would be unparsed into "OVER" correctly. If not, this may have to change

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1294,6 +1294,19 @@ private static String toSql(RelNode root, SqlDialect dialect,
         .withPostgresql().ok(expectedPostgresql);
   }
 
+  @Test void testPercentileContWithinGroupClauseBigQuery() {
+    final String query = "select percentile_cont(0.5) WITHIN GROUP (ORDER BY `product_class_id`)\n"
+        + "from `foodmart`.`product`";
+    final String expected = "SELECT PERCENTILE_CONT(0.5) OVER "
```
tanclary, line 1300: I think the PERCENTILE_CONT function for BigQuery must have two arguments. I think this expected query would probably fail since the column isn't passed to the function directly as an argument

PR #3468, opened 2023-10-13
[CALCITE-5806] add a String data type (enabled in Apache Spark and BigQuery)

MasseGuillaume:


mihaibudiu (commented):
How is string different from varchar?
If string is the same as varchar, can it be implemented by just having
`rootSchema.add("STRING", factory -> factory.createSqlType(sqlTypeName.VARCHAR))`?

MasseGuillaume:
Ah, to unparse user defined types it's via:

`SqlDialect.getCastSpec` then you can use `SqlUserDefinedTypeNameSpec`

```scala
val pos = SqlParserPos.ZERO
new SqlDataTypeSpec(new SqlUserDefinedTypeNameSpec(new SqlIdentifier("STRING", pos), pos), pos)
```

While it's possible to add a type without modifying Calcite's codebase, it's a lot of gymnastics for a type that is fairly common in practice:

https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#string_type
https://spark.apache.org/docs/latest/sql-ref-datatypes.html

mihaibudiu:
Once you add a new type you have to maintain it forever. Moreover, technically you should write many more tests; everything that works with varchar should be tested with string if you want to be sure it works. If something can be done without modifying Calcite I think that's the right way to do it.

MasseGuillaume:
Okay, I will close this PR and write a how-to guide to defined custom types and use string as an example.

However, there is a bug for the Spark SQL dialect in RelToSqlConverter since it does not unparse varchar to string.

mihaibudiu:
I just expressed my personal opinion above; my experience with Calcite isn't that significant, so I have been wrong before.

Regarding SPARK, you should file an issue, and it looks like you can perhaps send a PR for the fix for it as well.
Does the SPARK parser parse STRING types already?

MasseGuillaume:
Yes `string` is a first class citizen.

```
scala> spark.sql("""select cast(NULL as string)""").show()
+--------------------+
|CAST(NULL AS STRING)|
+--------------------+
|                null|
+--------------------+
```

PR #3471, opened 2023-10-15
[CALCITE-5921] SqlOperatorFixture.checkFails and checkAggFails don't check runtime failure

mihaibudiu:
This supersedes #3412.

Like #3412, this temporarily disables a test for a regression described by https://issues.apache.org/jira/browse/CALCITE-5990.
I have a pending PR fixing 5990, but it needs this PR to be merged first.

As https://issues.apache.org/jira/browse/CALCITE-5921 has pointed out, none of the checkFails test that fail at runtime were ever executed. This PR (like #3412) remedies this situation, and it also corrects some error messages which were never validated. Some of these error messages leak internal compiler data structures (e.g., "was not expecting value for enumeration), but cleaning them should be part of a future PR.

While trying to fix 5990 I have discovered that the tests that were run by `fixture.checkFail` were rewritten in different ways from the tests that were executed by `fixture.checkScalar*`. So it was possible to have test T such that `f.checkScalar(T)` fails with an exception but `f.checkFails(T)` fails because no exception is thrown! The change to SqlRuntimeTester ensures that the same test is run, and thus the execution proceeds in both cases in the same way. This change is in addition to the changes in #3412.

As soon as this PR is merged I will submit a fix for CALCITE-5990 as well.


mihaibudiu:
Thanks to @herunkang2018 for the original PR on which this one is based.

rubenada:
LGTM

PR #3472, opened 2023-10-16
[CALCITE-6052] SqlImplementor writes REAL, FLOAT, and DOUBLE literals as DECIMAL literals

mihaibudiu:
This bug was also fixed in #3411, but perhaps we can merge this PR more easily since it is much smaller.
I didn't add a new test case since the pig test case covers it.

### DIFF
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -1157,9 +1157,10 @@ name will have been converted to upper case also.
 | INTEGER, INT | 4 byte signed integer    | Range is -2147483648 to 2147483647
 | BIGINT      | 8 byte signed integer     | Range is -9223372036854775808 to 9223372036854775807
 | DECIMAL(p, s) | Fixed point             | Example: 123.45 or DECIMAL '123.45' is a DECIMAL(5, 2) value.
-| NUMERIC     | Fixed point               |
-| REAL, FLOAT | 4 byte floating point     | 6 decimal digits precision
-| DOUBLE      | 8 byte floating point     | 15 decimal digits precision
+| NUMERIC(p, s) | Fixed point             | A synonym for DECIMAL
+| REAL        | 4 byte floating point     | 6 decimal digits precision; examples: CAST(1.2 AS REAL), CAST('Infinity' AS REAL)
+| DOUBLE      | 8 byte floating point     | 15 decimal digits precision; examples: 1.4E2, CAST('-Infinity' AS DOUBLE), CAST('NaN' AS DOUBLE)
+| FLOAT       | 8 byte floating point     | A synonym for DOUBLE
 | CHAR(n), CHARACTER(n) | Fixed-width character string | 'Hello', '' (empty string), _latin1'Hello', n'Hello', _UTF16'Hello', 'Hello' 'there' (literal split into multiple parts), e'Hello\nthere' (literal containing C-style escapes)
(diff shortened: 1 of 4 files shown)


mihaibudiu:
I have added more test cases.
I have also tried to also fix a bug in the reference manual where the line for `CONTAINS_SUBSTR` is not rendered, but I don't know how to test that).
I will rename the commit when I squash.

site/_docs/reference.md
```diff
@@ -1158,8 +1158,8 @@ name will have been converted to upper case also.
 | BIGINT      | 8 byte signed integer     | Range is -9223372036854775808 to 9223372036854775807
 | DECIMAL(p, s) | Fixed point             | Example: 123.45 or DECIMAL '123.45' is a DECIMAL(5, 2) value.
 | NUMERIC     | Fixed point               |
-| REAL, FLOAT | 4 byte floating point     | 6 decimal digits precision
-| DOUBLE      | 8 byte floating point     | 15 decimal digits precision
+| REAL, FLOAT | 4 byte floating point     | 6 decimal digits precision.  Examples: 1.2E2, CAST('Infinity' AS REAL), CAST('-Infinity' AS FLOAT), CAST('NaN' AS FLOAT)
```
julianhyde, line 1161: FLOAT is actually the same as DOUBLE. (It's complicated, but standard SQL allows variable-precision floating point, and but we only allow 8 bytes.) I think you should put FLOAT on a separate line saying 'Equivalent to DOUBLE'.

No '.' at end of line.

mihaibudiu:
I have applied the suggested changes and squashed the commits optimistically.
But an audit of the code has uncovered the fact that DOUBLE, REAL, and FLOAT are treated widely inconsistently in the code. I have filed https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6074 to handle that. I hope that the documentation I updated it in this PR is correct - it was wrong before.

site/_docs/reference.md
```diff
@@ -1157,9 +1157,10 @@ name will have been converted to upper case also.
 | INTEGER, INT | 4 byte signed integer    | Range is -2147483648 to 2147483647
 | BIGINT      | 8 byte signed integer     | Range is -9223372036854775808 to 9223372036854775807
 | DECIMAL(p, s) | Fixed point             | Example: 123.45 or DECIMAL '123.45' is a DECIMAL(5, 2) value.
-| NUMERIC     | Fixed point               |
-| REAL, FLOAT | 4 byte floating point     | 6 decimal digits precision
-| DOUBLE      | 8 byte floating point     | 15 decimal digits precision
+| NUMERIC(p, s) | Fixed point             | A synonym for DECIMAL
+| REAL        | 4 byte floating point     | 6 decimal digits precision; examples: CAST(1.2 AS REAL), CAST('Infinity' AS REAL)
+| DOUBLE      | 8 byte floating point     | 15 decimal digits precision; examples: 1.4E2, CAST('-Infinity' AS DOUBLE), CAST('NaN' AS DOUBLE)
+| FLOAT       | 8 byte floating point     | A synonym for DOUBLE
```
JiajunBernoulli, line 1163: Sorry, what I missed?

I don't know why FLOAT is a synonym for DOUBLE.

JiajunBernoulli: Oh, I get it:https://lists.apache.org/thread/81sk8q4z8j5swj2r3bf8x8jc80k6b4sb

PR #3474, opened 2023-10-17
[CALCITE-3933] Incorrect SQL Emitted for Unicode for Several Dialects

zhougit86:


core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
```diff
@@ -537,7 +537,19 @@ public interface SqlConformance {
    */
   boolean isValueAllowed();
 
-  /**
+	/**
```
mihaibudiu, line 540: indentation seems to be off here

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1486,6 +1486,9 @@ private static String toSql(RelNode root, SqlDialect dialect,
     assertThat(toSql(root), isLinux(expectedSql));
   }
 
+
```
mihaibudiu, line 1489: are these necessary?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7133,6 +7136,19 @@ private void checkLiteral2(String expression, String expected) {
     sql(query).ok(expected);
   }
 
+  @Test void testBigQueryUnicode() {
```
mihaibudiu, line 7139: would this need a JavaDoc comment indicating the solved issue?

mihaibudiu (commented):
Can you also take a look at https://issues.apache.org/jira/browse/CALCITE-6006?
Does this affect RelToSql?
Is it necessary to update the corresponding tests in #3429?

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -1792,13 +1792,12 @@ void checkPeriodPredicate(Checker checker) {
     expr("cast('foo' as bar)")
         .ok("CAST('foo' AS `BAR`)");
   }
-  
+
   @Test void testParsingNonIsoCharacter() {
   	String sql = "select 'ק' ";
-		sql(sql).ok("SELECT u&'\\05e7'");
```
julianhyde: why remove this test? The `u&` is valid standard SQL and is still valid Calcite SQL

zhougit86: the CALCITE-6001 mentioned 
'This means that when converting a query like:
select 'ק' as result;
you will get the following the error: Failed to encode 'ק' in character set 'ISO-8859-1'.'

I justed it according to this behavior

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
zhougit86: removed

core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
zhougit86: modified

tanclary:
i see you based this on a couple of my commits, there's some more updated work on https://github.com/apache/calcite/pull/3425 which might be a bit better. Let me know what you think and we can work together. Feel free to review that also^

zhougit86:
@tanclary I think rework covers more dialects, [#3425 ](https://github.com/apache/calcite/pull/3425) looks good to me

PR #3475, opened 2023-10-18
[CALCITE-6041] MAP sub-query gives NullPointerException

chucheng92:
Currently the map sub-query such as `map(select a, b)` can't work.
details pls see: https://issues.apache.org/jira/browse/CALCITE-6041

### DIFF
--- a/core/src/test/resources/sql/sub-query.iq
+++ b/core/src/test/resources/sql/sub-query.iq
@@ -3685,4 +3685,40 @@ FROM dept d1;
 
 !ok
 
+# [CALCITE-6041] MAP sub-query gives NullPointerException
+# map size > 1
+SELECT map(SELECT empno, deptno from emp where deptno < 20);
++-----------------------------+
+| EXPR$0                      |
++-----------------------------+
+| {7782=10, 7839=10, 7934=10} |
++-----------------------------+
+(1 row)
+
+!ok
+
+# [CALCITE-6041] MAP sub-query gives NullPointerException
+# map size = 1
+SELECT map(SELECT empno, deptno from emp where empno = 7369);
++-----------+
+| EXPR$0    |
++-----------+
+| {7369=20} |
++-----------+
+(1 row)
+
+!ok
+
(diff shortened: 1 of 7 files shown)


core/src/test/resources/sql/sub-query.iq
```diff
@@ -3685,4 +3685,15 @@ FROM dept d1;
 
 !ok
 
+# [CALCITE-6041] MAP sub-query gives NullPointerException
+SELECT map(SELECT empno, deptno from emp where deptno < 20);
```
julianhyde, line 3689: test may be nondeterministic. if Calcite allows an `order by` in the subquery I think you should add it

also add tests where the map contains 1 row and 0 rows

core/src/test/resources/sql/sub-query.iq
```diff
@@ -3685,4 +3685,26 @@ FROM dept d1;
...
+| EXPR$0                      |
++-----------------------------+
+| {7782=10, 7839=10, 7934=10} |
++-----------------------------+
+(1 row)
+
+!ok
+
+# [CALCITE-6041] MAP sub-query gives NullPointerException
+SELECT map(SELECT empno, deptno from emp where false);
```
chucheng92, line 3700: here return an empty map is similar to array query. If array query has no input rows will return a empty array.

core/src/test/resources/sql/sub-query.iq
julianhyde: Leave it in. I agree that it will be deterministic, knowing where the test data comes from.

I think we need tests with N, 1 and 0 rows.

chucheng92: ok, I have rebased the main and squashed the commits. If you have time, PTAL~ thanks.

chucheng92:
hi, @julianhyde Sorry to tag you. I have addressed the comments. could you help to merge this?
so we can add it in up-coming 1.36.

chucheng92:
If there are no other comments, I will try to merge it. So we can add it in upcoming 1.36.

PR #3477, opened 2023-10-19
[CALCITE-6060] Rename the named parameter OFFSET in window functions to avoid conflicts with reserved keywords

lincoln-lil:
This is a breaking change. Users should change the named parameter in window table function from ``OFFSET`` to `OFFSETSIZE`.
More details: https://issues.apache.org/jira/browse/CALCITE-6060

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlWindowTableFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlWindowTableFunction.java
@@ -56,7 +56,7 @@ public class SqlWindowTableFunction extends SqlFunction
   protected static final String PARAM_SIZE = "SIZE";
 
   /** The optional align offset for each window. */
-  protected static final String PARAM_OFFSET = "OFFSET";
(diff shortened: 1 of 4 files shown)


mihaibudiu:
At a superficial look the grammar does not really look ambiguous, so I expect that this change is not needed; rather, the parser should be changed. 

How does the conflict in the title of this issue manifest? Are programs parsed incorrectly? Or are legal programs rejected?

julianhyde:
We use techniques like this already. But my comment about a 'mush of special cases' stands. There are far too many functions already that have special treatment in the parser.

In fact, one of those functions is called `OFFSET` (required for BigQuery) and so allowing a named method parameter called `OFFSET` would make the grammar ambiguous.

lincoln-lil:
@mihaibudiu  @julianhyde thanks for looking at this issue and giving your input! 
Knowing that an `OFFSET` function already exists for BigQuery, I agree with julian that quoting `OFFSET` in window functions is acceptable.  So, I'll cancel this pr and close the issue.

PR #3480, opened 2023-10-23
[CALCITE-129] Support recursive WITH queries

HanumathRao:
This pull request introduces an end to end support for recursive queries, building upon the previous Jira changes (CALCITE-2812) that provided a functional solution for relnode trees with RepeatUnion. Here's an overview of the approach:
* **Parser Changes:** Parser rule for WITH is enhanced to support RECURSIVE keyword. When a RECURSIVE keyword is used in the query a SqlLiteral(true) is passed onto SqlWithItem node.
* **Validator Changes:** The primary focus of this PR is on validation. The existing mechanism of passing the enclosed node to each validation function is leveraged to pass the withItem node for recursive queries. In the case of "with recursive," the enclosed node is the current withItem node; otherwise, it is the existing with node(i.e with). This enables specific processing for SetOperators (specifically for Union). When registering the SetOperator and the enclosing node is a withItem for which recursive flag is true, then a new WithRecursiveScope is passed along, allowing for control to create a WithRecursiveNameSpace instead of WithNamespace during the creation of namespace. Doing so helps break the cyclic dependency during the from clause registration. A WITH_ITEM_TABLE_REF, a new internal SQL node type, is introduced to resolve the cyclic dependency and creating a transientScan at the appropriate location in the code.
* **SqlToRelConverter Changes:** These changes involve the creation of the RepeatUnion relnode and TransientScan relnode. TransientScan is created when a SqlKind.WITH_ITEM_TABLE_REF is encountered during conversion. The prerequisites for creating RepeatUnion are determined by the node type (a UNION in set operation traversal) and whether the enclosingNode is a WITH_ITEM node. The final trigger check (to create RepeatUnion) is to find the WITH_ITEM name in the right child of the union.
* **Added Various Test Cases:** New test cases have been added to the CoreTests. Some of the SQL queries used for the end to end tests are derived from the existing tests introduced in the JIRA (CALCITE-2812).

In summary, this pull request enhances support for recursive queries by making changes in the parser, validator, and SqlToRelConverter, along with the addition of relevant test cases.


core/src/test/java/org/apache/calcite/test/enumerable/EnumerableRepeatUnionTest.java
```diff
@@ -45,6 +45,17 @@
  */
 class EnumerableRepeatUnionTest {
 
+  @Test void testGenerateNumbersUsingSql() {
+    CalciteAssert.that()
+        .query("WITH RECURSIVE delta(n) AS (\n"
+            + "  VALUES (1)\n"
+            + " UNION ALL\n"
```
mihaibudiu, line 52: shouldn't UNION ALL actually never converge?
Why isn't the result collection a multiset that grows infinitely?

rubenada: @mihaibudiu the `WHERE n < 10` breaks the "infinite growth". We get the same behavior [in e.g. PostgreSQL](http://sqlfiddle.com/#!17/9eecb/111270)

mihaibudiu: The `WHERE` breaks the growth in the value of n, but UNION would break the growth in cardinality. 
UNION ALL should produce a multiset, so you should get the following values in `delta` in successive iterations: [1], [1, 2], [1, 2, 2, 3], [1, 2, 2, 2, 3, 3, 4] etc in your relation in successive iterations. I understand how this should work with `UNION`, but not with `UNION ALL`. Why is delta a set and not a multiset?

HanumathRao: @mihaibudiu  Thanks for reviewing the changes. I agree with @rubenada regarding the explanation as to why this use case doesn't run into infinite growth. I believe in general that semantics of UNION / UNION ALL are subtle in Recursive queries. 

Here is an example (again from postgresql) where usage of the union all (http://sqlfiddle.com/#!17/9eecb/111278) will run into infinite loop but usage of union works fine.
The infinite loop for the above example actually is because of the result set being a multi set when union all is used.

rubenada: More info about how RECURSIVE UNION [ALL] works internally can be found here:
https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-RECURSIVE
Calcite implementation was heavily inspired by this postgresql description.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -3774,6 +3791,48 @@ protected RelRoot convertQueryRecursive(SqlNode query, boolean top,
     }
   }
 
+  public static boolean hasATableScanWithName(RelNode root, final String recurTableName) {
```
rubenada, line 3794: There is a RelOptTableFinder class inside RelBuider which seems to have the same purpose. Currently it is private static, but perhaps it could be moved to another localion (RelOptUtil?) as a public element, and reuse the same code here and in RelBuilder.

core/src/test/java/org/apache/calcite/test/enumerable/EnumerableRepeatUnionTest.java
```diff
@@ -70,6 +81,47 @@ class EnumerableRepeatUnionTest {
...
+            + "   SELECT * FROM aux\n")
+        .returnsOrdered("I=0", "I=1", "I=2", "I=3", "I=4", "I=5", "I=6", "I=7", "I=8", "I=9");
+  }
+
+  @Test void testGenerateNumbers2UsingSqlCheckPlan() {
+    CalciteAssert.that()
+        .with(CalciteConnectionProperty.LEX, Lex.JAVA)
+        .with(CalciteConnectionProperty.FORCE_DECORRELATE, false)
+        .withSchema("s", new ReflectiveSchema(new HierarchySchema()))
+        .withHook(Hook.PLANNER, (Consumer<RelOptPlanner>) planner -> {
```
rubenada, line 100: Do we really need the Hook.PLANNER for this test? It adds/removes join-related rules, but the query does not have any join....

rubenada:
Nice work @HanumathRao !  I have left some (minor) comments.
I'm not the biggest expert on the parser and validator, so I'd prefer if someone else would take a look at that part.
Thanks for adding such a thorough amount of tests.

I think there is one part of the documentation that needs to be updated in the PR: in `algebra.md`, in the "Recursive Queries", this paragraph should be removed :)
```
Note that there is no support for recursive queries in the SQL layer yet
([CALCITE-129](https://issues.apache.org/jira/browse/CALCITE-129));
the `WITH RECURSIVE` query above is only for illustrative purposes.
```

HanumathRao:
> Nice work @HanumathRao ! I have left some (minor) comments. I'm not the biggest expert on the parser and validator, so I'd prefer if someone else would take a look at that part. Thanks for adding such a thorough amount of tests.
> 
> I think there is one part of the documentation that needs to be updated in the PR: in `algebra.md`, in the "Recursive Queries" section, this paragraph should be removed :)
> 
> ```
> Note that there is no support for recursive queries in the SQL layer yet
> ([CALCITE-129](https://issues.apache.org/jira/browse/CALCITE-129));
> the `WITH RECURSIVE` query above is only for illustrative purposes.
> ```

Thanks @rubenada for the review. I have addressed the review comments, please take a look when you get a chance.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -5232,6 +5232,39 @@ void testReturnsCorrectRowTypeOnCombinedJoin() {
...
+    sql("with RECURSIVE emp2 as (select * from emp union select * from emp2)\n"
+        + "select * from emp2")
+        .type(EMP_RECORD_TYPE);
+
+    // union all with recursive working case.
+    sql("with RECURSIVE emp2 as (select * from emp union all select * from emp2)\n"
+        + "select * from emp2")
+        .type(EMP_RECORD_TYPE);
+
+    // union all with recursive working case.
```
rubenada, line 5258: minor: this comment seems wrong ("working case" but the test actually checks for a failure)

core/src/main/resources/org/apache/calcite/runtime/CalciteResource.properties
```diff
@@ -63,6 +63,8 @@ IncompatibleValueType=Values passed to {0} operator must have compatible types
 IncompatibleTypesInList=Values in expression list must have compatible types
 IncompatibleCharset=Cannot apply {0} to the two different charsets {1} and {2}
 InvalidOrderByPos=ORDER BY is only allowed on top-level SELECT
+RecursiveWithMustHaveUnionSetOp=A recursive query only supports UNION [ALL] operator
```
rubenada, line 66: Would it be possible to have unit tests forcing to have these two error messages?

core/src/main/java/org/apache/calcite/sql/SqlWithItem.java
```diff
@@ -30,13 +30,16 @@
 public class SqlWithItem extends SqlCall {
   public SqlIdentifier name;
   public @Nullable SqlNodeList columnList; // may be null
+  public SqlLiteral recursive;
   public SqlNode query;
 
   public SqlWithItem(SqlParserPos pos, SqlIdentifier name,
-      @Nullable SqlNodeList columnList, SqlNode query) {
+      @Nullable SqlNodeList columnList, SqlNode query,
+      SqlLiteral recursive) {
```
zabetak, line 38: Do we want to keep previous constructor for backward compatibility purposes? Please check other SqlNode classes and check if we are using a deprecation pattern.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -7487,6 +7487,21 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
...
+        + ")\n"
+        + "SELECT * FROM aux";
+
+    sql(sql)
+        .withLateDecorrelate(true)
+        .withTrim(true)
+        .withRule() // empty program
+        .checkUnchanged();
+  }
+
```
zabetak, line 7504: We are not testing any rules here so the test seems misplaced. Consider removing it.

core/src/test/java/org/apache/calcite/test/enumerable/EnumerableRepeatUnionTest.java
```diff
@@ -45,6 +45,17 @@
  */
 class EnumerableRepeatUnionTest {
 
+  @Test void testGenerateNumbersUsingSql() {
```
zabetak, line 48: The new tests added here are basically testing recursive queries end-to-end. For end-to-end tests it is better to use the .iq files so consider moving the tests in `recursive_queries.iq` and remove them from here unless there is a specific reason to have them here.

site/_docs/algebra.md
```diff
@@ -309,10 +309,6 @@ LogicalRepeatUnion(all=[true])
         LogicalTableScan(table=[[aux]])
 {% endhighlight %}
 
-Note that there is no support for recursive queries in the SQL layer yet
-([CALCITE-129](https://issues.apache.org/jira/browse/CALCITE-129));
-the `WITH RECURSIVE` query above is only for illustrative purposes.
-
```
zabetak, line 315: I think the https://calcite.apache.org/docs/reference.html part of the documentation should be updated as well to reflect the new changes.

core/src/main/java/org/apache/calcite/sql/validate/WithItemRecursiveNameSpace.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.SqlWith;
+import org.apache.calcite.sql.SqlWithItem;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.Pair;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/** Very similar to {@link WithItemNamespace} but created only for RECURSIVE queries. */
+public class WithItemRecursiveNameSpace extends WithItemNamespace {
```
zabetak, line 34: Should the class be `public`?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
zabetak: This seems like a test that should be added in `SqlToRelConverterTest` class.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -3304,6 +3305,29 @@ public static RelNode createProject(RelNode child, Mappings.TargetMapping mappin
     return createProject(projectFactory, child, Mappings.asListNonNull(mapping.inverse()));
   }
 
+  /** Returns the relational table node for {@code tableName} if it occurs within a
+   * relational expression {@code root} otherwise an empty option is returned. */
+  public static Optional<RelOptTable> findTable(RelNode root, final String tableName) {
```
zabetak, line 3310: I don't think we use `Optional` much in the project. I don't have a strong opinion for using it here or not but it seems that all the other methods in this class are simply returning null so for keeping things more consistent I would not add the wrapper.

zabetak (commented):
I did a quick pass and overall LGTM. I left some small comments but nothing major.

core/src/main/resources/org/apache/calcite/runtime/CalciteResource.properties
HanumathRao: These errors are protected by "Object not found", currently it is very hard to hit these errors but I thought it is good to protect code path in regards to any future changes with scope etc.

Please let me know if you think these are redundant, then I can remove them.

HanumathRao:
Thanks @zabetak  and @rubenada  for the review, I addressed the review comments.

core/src/main/java/org/apache/calcite/sql/validate/WithItemRecursiveNameSpace.java
```diff
@@ -0,0 +1,75 @@
...
+      if (columnList == null || columnList.size() == 0) {
+        // This should never happen but added to protect against the NullPointerException.
+        return leftChildType;
+      }
+      final RelDataTypeFactory.Builder builder =
+          validator.getTypeFactory().builder();
+      Pair.forEach(SqlIdentifier.simpleNames(columnList),
+          leftChildType.getFieldList(),
+          (name, field) -> builder.add(name, field.getType()));
+      rowType = builder.build();
```
macroguo-ghy, line 67: We don't need to set the `rowtype`. Method `org.apache.calcite.sql.validate.AbstractNamespace#validate` will help us do this.

core/src/main/java/org/apache/calcite/sql/validate/WithItemRecursiveNameSpace.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.SqlWith;
+import org.apache.calcite.sql.SqlWithItem;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.Pair;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/** Very similar to {@link WithItemNamespace} but created only for RECURSIVE queries. */
+class WithItemRecursiveNameSpace extends WithItemNamespace {
```
macroguo-ghy, line 34: Use `Namespace`.

core/src/main/java/org/apache/calcite/sql/validate/WithItemRecursiveNameSpace.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Override public RelDataType getRowType() {
+    if (rowType == null) {
+      SqlBasicCall call;
+      if (this.withItem.query.getKind() == SqlKind.WITH) {
+        call = (SqlBasicCall) ((SqlWith) this.withItem.query).body;
+      } else {
+        call = (SqlBasicCall) this.withItem.query;
+      }
+      // As this is a recursive query we only should evaluate left child for getting the rowType.
```
macroguo-ghy, line 53: I am not sure if you check `UNION` operands type. Can you give me a test like
```sql
WITH RECURSIVE cte (n) AS
(
  SELECT 1, 2
  UNION ALL
  SELECT n + 1 FROM cte WHERE n < 5
)
SELECT * FROM cte;
```
This case is an illegal SQL, it should fail.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3126,17 +3141,21 @@ private void registerWith(
     SqlValidatorScope scope = parentScope;
     for (SqlNode withItem_ : with.withList) {
       final SqlWithItem withItem = (SqlWithItem) withItem_;
-      final WithScope withScope = new WithScope(scope, withItem);
+
+      final boolean isRecursiveWith = withItem.recursive.booleanValue();
+      final SqlValidatorScope withScope =
```
macroguo-ghy, line 3146: What are the benefits of using a nested scope compared to using `WithRecursiveScope` directly

core/src/main/java/org/apache/calcite/sql/SqlWithItem.java
```diff
@@ -30,13 +30,26 @@
 public class SqlWithItem extends SqlCall {
   public SqlIdentifier name;
   public @Nullable SqlNodeList columnList; // may be null
+  public SqlLiteral recursive;
```
macroguo-ghy, line 33: Do you have to use the `SqlLiteral` type? I think using `boolean` is simpler.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -3486,14 +3486,16 @@ SqlNodeList WithList() :
 {
     final Span s;
     final List<SqlWithItem> list = new ArrayList<SqlWithItem>();
+    boolean recursive = false;
 }
 {
-    <WITH> { s = span(); }
-    AddWithItem(list) ( <COMMA> AddWithItem(list) )*
+    <WITH> [ <RECURSIVE> { recursive = true; } ]{ s = span(); }
```
macroguo-ghy, line 3492: I think you'd better add some tests for `WITH RECURSIVE` in `SqlParserTest`.

macroguo-ghy (commented):
I left some comments on parse and validation, `WITH RECURSIVE` is an exciting and important syntax. Thank you for your contribution.

core/src/main/resources/org/apache/calcite/runtime/CalciteResource.properties
rubenada: ok, np

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3099,9 +3100,23 @@ private void registerSetop(
 
     // A setop is in the same scope as its parent.
     scopes.put(call, parentScope);
-    for (SqlNode operand : call.getOperandList()) {
+    @NonNull SqlValidatorScope recursiveScope = parentScope;
+    if (enclosingNode.getKind() == SqlKind.WITH_ITEM) {
+      if (node.getKind() != SqlKind.UNION) {
```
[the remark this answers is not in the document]
HanumathRao: The recursive flag is not explicitly checked here because for only recursive queries we pass WITH_ITEM as enclosed node(essentially it is kind of an implicit check). For non recursive queries we pass the original enclosed node which is either WITH or its parent. 

`And I think we should implement this validation logic in org.apache.calcite.sql.validate.WithItemNamespace#validateImpl or org.apache.calcite.sql.validate.SetopNamespace#validateImpl instead of registerSetop. WDYT?`

As regards to the validation logic implementation I did go that path but I think code is getting messy. I thought if we can create a WithRecursive* classes for these we could do custom logic for recursive queries alone.

mihaibudiu:
I will need more time to review this, but if you want to get it merged for 1.36 you can go ahead. I will file issues later if I find problems.

core/src/main/java/org/apache/calcite/sql/SqlWithItem.java
julianhyde: SqlLiteral contains a position, so you can use it in error messages if necessary.

core/src/main/java/org/apache/calcite/sql/validate/WithItemRecursiveNameSpace.java
julianhyde: Agreed. (@HanumathRao, note lower-case 's'.)

core/src/main/java/org/apache/calcite/sql/validate/SqlWithItemTableRef.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A <code>SqlWithItemTableRef</code> is a node created during validation for
+ * recursive queries which represents a recursive with table reference.
+ */
```
julianhyde, line 37: I'd change "a recursive with table reference" to "a table reference in a {@code WITH RECURSIVE} clause". The sentence is easier for humans to parse.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -2759,6 +2769,13 @@ protected void convertUnpivot(Blackboard bb, SqlUnpivot unpivot) {
     bb.setRoot(relBuilder.build(), true);
   }
 
+  private void convertTransientScan(Blackboard bb, SqlWithItem withItem) {
+    final SqlValidatorNamespace fromNamespace = getNamespace(withItem).resolve();
+    bb.setRoot(
+        relBuilder.transientScan(withItem.name.getSimple(),
+        fromNamespace.getRowType()).build(), true);
+  }
```
julianhyde, line 2776: need more indentation before `fromNamespace`

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -3304,6 +3304,29 @@ public static RelNode createProject(RelNode child, Mappings.TargetMapping mappin
...
+  public static @Nullable RelOptTable findTable(RelNode root, final String tableName) {
+    RelOptTable[] table = new RelOptTable[1];
+    try {
+      RelShuttle visitor = new RelHomogeneousShuttle() {
+        @Override public RelNode visit(TableScan scan) {
+          final RelOptTable scanTable = scan.getTable();
+          final List<String> qualifiedName = scanTable.getQualifiedName();
+          if (qualifiedName.get(qualifiedName.size() - 1).equals(tableName)) {
+            table[0] = scanTable;
+            throw Util.FoundOne.NULL;
```
julianhyde, line 3318: a bit simpler is `throw new Util.FoundOne(scanTable)`; then you don't need the `table[]` variable.

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -195,6 +195,9 @@ public enum SqlKind {
   /** Item in WITH clause. */
   WITH_ITEM,
 
+  /** Represents a recursive CTE as a table ref. **/
+  WITH_ITEM_TABLE_REF,
```
julianhyde, line 199: change `**/` to `*/`

core/src/main/java/org/apache/calcite/sql/SqlWithItem.java
julianhyde: @HanumathRao Thanks keeping the old constructor, as @zabetak requested. Please add '// to be removed before 2.0'.

And make it call the new constructor (via `this(...)`); it prevents the code diverging, and is a single place to put a debugger break-point.

core/src/test/resources/sql/recursive_queries.iq
```diff
@@ -0,0 +1,507 @@
...
++---------+
+| A       |
++---------+
+| 3628800 |
++---------+
+(1 row)
+
+!ok
+
+WITH RECURSIVE FibonacciCTE(n, a, b) AS (
```
julianhyde, line 38: Thanks for all these fabulous test queries.

I don't think I saw any mutually recursive queries. E.g.
```
WITH RECURSIVE f AS (SELECT ... FROM g),
   g AS (SELECT ... FROM f),
  z AS (SELECT ... FROM somethingElse)
SELECT ... FROM f, z
```

Are mutually recursive queries valid (per the SQL standard)? Does your implementation support them?

core/src/main/java/org/apache/calcite/sql/validate/WithItemRecursiveNameSpace.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.SqlWith;
+import org.apache.calcite.sql.SqlWithItem;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.Pair;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Objects;
```
julianhyde, line 33: In new code, I am doing `import static java.util.Objects.requireNonNull` rather than `import java.util.Objects`.

core/src/main/java/org/apache/calcite/sql/validate/WithItemRecursiveNameSpace.java
```diff
@@ -0,0 +1,78 @@
...
+
+/** Very similar to {@link WithItemNamespace} but created only for RECURSIVE queries. */
+class WithItemRecursiveNameSpace extends WithItemNamespace {
+  private final SqlWithItem withItem;
+  private final SqlWithItemTableRef withItemTableRef;
+  WithItemRecursiveNameSpace(SqlValidatorImpl validator,
```
julianhyde, line 39: and blank line and javadoc comment before constructor

core/src/main/java/org/apache/calcite/sql/validate/WithRecursiveScope.java
```diff
@@ -0,0 +1,76 @@
...
+ * <p>{@code t1} provides a scope that is used to validate {@code q2} and {@code q1 with union}
+ * (and therefore {@code q2} may reference {@code t1} and {@code t2}),
+ * and {@code t2} provides a scope that is used to validate {@code q3}
+ * (and therefore q3 may reference {@code t1} and {@code t2}).
+ */
+class WithRecursiveScope extends ListScope {
+  private final SqlWithItem withItem;
+
+  /** Creates a WithScope. */
+  WithRecursiveScope(SqlValidatorScope parent, SqlWithItem withItem) {
```
julianhyde, line 45: should be 'Creates a WithRecursiveScope'.

core/src/main/java/org/apache/calcite/sql/validate/WithRecursiveScope.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlWithItem;
+
+import com.google.common.collect.ImmutableList;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.List;
+
+/** Scope providing the objects that are being defined using the with clause to
```
julianhyde, line 29: change "with clause" to "{@code WITH} clause"

core/src/main/java/org/apache/calcite/sql/validate/WithScope.java
```diff
@@ -70,4 +73,8 @@ class WithScope extends ListScope {
     }
     super.resolveTable(names, nameMatcher, path, resolved);
   }
+
+  public @Nullable WithRecursiveScope getRecursiveScope() {
+    return recursiveScope;
+  }
 }
```
julianhyde, line 80: Since the class is not `public`, there's no need for a `public` method. I'd remove this method, and just make the `recursiveScope` field package-protected.

julianhyde (changes requested):
Looks great. I've suggested a few changes, but when they're done I'll be +1.

HanumathRao:
Thanks @julianhyde and @macroguo-ghy for the review. I have addressed all the review comments. Please take a look when you get a chance.

core/src/test/resources/sql/recursive_queries.iq
HanumathRao: Thanks for asking this question. Yes, mutually recursive queries are valid as per SQL standard. However, this implementation doesn't support mutually recursive queries. In the latest commit, I have added a test case which reports an error (in SqlValidatorTest).

I think it is better to track support for mutually recursive queries as a new functionality rather than clubbing it with the general recursive query support. WDYT?

Please find the sqlfiddle [link](http://sqlfiddle.com/#!17/9eecb/111736) for postgres which reports an error for mutually recursive queries.

core/src/main/java/org/apache/calcite/sql/validate/WithItemRecursiveNameSpace.java
julianhyde: I agree, @macroguo-ghy. And I think that's the only remaining issue.

@HanumathRao, I have created https://github.com/julianhyde/calcite/tree/129-with-recursive to address that one issue. Please take a look. If it's OK I will merge.

PR #3481, opened 2023-10-24
[CALCITE-5990] Explicit cast to numeric type doesn't check overflow

mihaibudiu:
It turns out that in SQL casts are checked: a cast that cannot represent the resulting value in the destination type should report a runtime error. This PR adds range checking for Number values. The tests for this PR require https://github.com/apache/calcite/pull/3471 to fix a bug in the test fixture.

### DIFF
--- a/linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
+++ b/linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
@@ -22,6 +22,8 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.lang.reflect.Type;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.AbstractList;
@@ -367,28 +369,68 @@ public enum Primitive {
   }
 
   /**
-   * Converts a number into a value of the type
-   * specified by this primitive.
+   * Check if a value after rounding falls within a specified range.
+   *
+   * @param value  Value to compare.
+   * @param min    Minimum value allowed.
+   * @param max    Maximum value allowed.
+   */
+  static void checkRoundedRange(Number value, double min, double max) {
+    double dbl = value.doubleValue();
+    // The equivalent of DOWN rounding for BigDecimal
+    dbl = dbl > 0 ? Math.floor(dbl) : Math.ceil(dbl);
+    if (dbl < min || dbl > max) {
+      throw new ArithmeticException("Value " + value + " out of range");
+    }
+  }
+
+  /**
+   * Converts a number into a value of the type specified by this primitive
+   * using the SQL CAST rules.  If the value conversion causes loss of significant digits,
+   * an exception is thrown.
    *
    * @param value  Value to convert.
-   * @return       The converted value, or null if the
-   *               conversion cannot be performed.
+   * @return       The converted value, or null if the type of the result is not a number.
    */
   public @Nullable Object numberValue(Number value) {
     switch (this) {
     case BYTE:
+      checkRoundedRange(value, Byte.MIN_VALUE, Byte.MAX_VALUE);
       return value.byteValue();
     case CHAR:
+      // No overflow checks for char values.
+      // For example, Postgres has this behavior.
       return (char) value.intValue();
     case SHORT:
+      checkRoundedRange(value, Short.MIN_VALUE, Short.MAX_VALUE);
       return value.shortValue();
     case INT:
+      checkRoundedRange(value, Integer.MIN_VALUE, Integer.MAX_VALUE);
       return value.intValue();
     case LONG:
-      return value.longValue();
+      if (value instanceof Byte
+          || value instanceof Short
+          || value instanceof Integer
+          || value instanceof Long) {
+        return value.longValue();
+      }
+      if (value instanceof Float
+          || value instanceof Double) {
+        // The value Long.MAX_VALUE cannot be represented exactly as a double,
+        // so we cannot use checkRoundedRange.
+        BigDecimal decimal = BigDecimal.valueOf(value.doubleValue())
+            // Round to an integer
+            .setScale(0, RoundingMode.DOWN);
+        // longValueExact will throw ArithmeticException if out of range
+        return decimal.longValueExact();
+      }
+      throw new AssertionError("Unexpected Number type "
+          + value.getClass().getSimpleName());
     case FLOAT:
+      // out of range values will be represented as infinities
       return value.floatValue();
     case DOUBLE:
+      // out of range values will be represented as infinities
       return value.doubleValue();
     default:
       return null;
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -1470,7 +1470,7 @@ Algorithms for implicit conversion are subject to change across Calcite releases
 
 | Operator syntax                         | Description
 |:----------------------------------------| :----------
-| CAST(value AS type)                     | Converts a value to a given type
+| CAST(value AS type)                     | Converts a value to a given type. Casts between integer types truncate towards 0
 | CONVERT(string, charSet1, charSet2)     | Converts *string* from *charSet1* to *charSet2*
 | CONVERT(value USING transcodingName)    | Alter *value* from one base character set to *transcodingName*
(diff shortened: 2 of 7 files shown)


mihaibudiu:
The issue https://issues.apache.org/jira/browse/CALCITE-5990 fixed by this PR is marked as a blocker for the 1.36.0 release. This is a relatively small but subtle fix, so I would appreciate a code review.

rubenada:
@mihaibudiu  do we need to update this PR to re-enable the test that was disabled on CALCITE-5921?

mihaibudiu:
I have updated the PR and reenabled the test that was disabled.
I have also discovered a lot of related broken tests in SqlOperatorTest, and I re-enabled many of them (but not all).
I have filed another issue to look at the remaining broken tests further: https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6073
I realize that there are a lot of things going in this PR, but hopefully it can make it for 1.36.

rubenada:
Thanks for moving this forward @mihaibudiu . PR looks already in a good shape, I'm sure it will make it into 1.36.
I just realized, there's another ticket which might be potentially related to broken (i.e. skipped) tests in `SqlOperatorTest`: CALCITE-2539.
Could you please check if any of the tests inside `SqlOperatorTest#testCastExactNumericLimits` or `testCastInvalid` which are currently skipped via `if (Bug.CALCITE_2539_FIXED) { ... }` are actually fixed with the current patch (and could be re-enabled)?

mihaibudiu:
In this PR I have handled casts to values that are represented as Number in Java, but not DECIMAL.
It looks like there are some bugs remaining there, including Bug.CALCITE_2539. I will assign that to myself.
But I may have solved some cases covered by that bug in this PR. I will re-enable those tests. Expect a new commit today.

mihaibudiu:
That bug also covers casts from VARCHAR to various numeric types, which I also didn't address in this PR.

mihaibudiu:
I have re-enabled a few tests that now pass, but not all, though a new commit.
I won't assign CALCITE-2539 to myself, since it covers many other operations besides casts.
I will file another issue to track the remaining cast cases and assign that to myself.

mihaibudiu:
I believe that this PR is ready for 1.36.

mihaibudiu:
Actually, I think I can cover a few more cases. Give me a bit more time to see if I can handle the string literal casts.

mihaibudiu:
I pushed a second commit re-enabling a few more tests. Now the PR is ready.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -624,7 +624,7 @@ void testCastExactNumericLimits(CastType castType, SqlOperatorFixture f) {
         f.checkCastFails(numeric.minOverflowNumericString,
             type, LITERAL_OUT_OF_RANGE_MESSAGE, false, castType);
       } else {
-        if (Bug.CALCITE_2539_FIXED) {
+        if (numeric != Numeric.DECIMAL5_2 || Bug.CALCITE_2539_FIXED) {
           f.checkCastFails(numeric.maxOverflowNumericString,
               type, OUT_OF_RANGE_MESSAGE, true, castType);
```
rubenada, line 629: It seems that `OUT_OF_RANGE_MESSAGE` is defined as a "generic string"; in the tests that you have re-enabled, could we replace it for a more meaningful value with the actual expected error message?

rubenada:
Thanks for the effort improving the tests and going beyond the original scope @mihaibudiu .
I have just left one minor comment; other than that lgtm

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: Yes, sorry about that; unfortunately the error message produced by ParseException is not very good.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
```diff
@@ -367,28 +369,67 @@ public static List<Double> asList(double[] elements) {
...
+          || value instanceof Byte
+          || value instanceof Long) {
+        return value.longValue();
+      }
+      if (value instanceof Float
+          || value instanceof Double) {
+        BigDecimal decimal = BigDecimal.valueOf(value.doubleValue())
+            // Round to an integer
+            .setScale(0, RoundingMode.DOWN);
+        // longValueExact will throw if out of range
```
herunkang2018, line 423: minor comment, longValueExact will throw `ArithmeticException` if out of range

site/_docs/reference.md
```diff
@@ -1469,7 +1469,7 @@ Algorithms for implicit conversion are subject to change across Calcite releases
 
 | Operator syntax                         | Description
 |:----------------------------------------| :----------
-| CAST(value AS type)                     | Converts a value to a given type
+| CAST(value AS type)                     | Converts a value to a given type.  Casts between integer types truncate towards 0
```
herunkang2018, line 1472: It would be better to remove the extra space.

herunkang2018:
Thanks for the PR, I left some minor comments, other looks good.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
```diff
@@ -367,28 +369,67 @@ public static List<Double> asList(double[] elements) {
...
-   *               conversion cannot be performed.
+   * @return       The converted value, or null if the type of the result is not a number.
    */
   public @Nullable Object numberValue(Number value) {
     switch (this) {
     case BYTE:
+      checkRoundedRange(value, Byte.MIN_VALUE, Byte.MAX_VALUE);
       return value.byteValue();
     case CHAR:
+      // No overflow checks for char values
```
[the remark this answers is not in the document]
mihaibudiu: This is a good question. You see, there is actually no written specification of what should happen in this case, that's why I filed https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6073.
I checked Postgres and it doesn't actually report an error when casting to CHAR, it look like it only truncates the value.
But in the absence of a spec for what CASTs are supposed to do there is no good answer.

site/_docs/reference.md
mihaibudiu: I think markdown will actually ignore the extra spaces, but I will remove it.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
rubenada: Thanks for the improvements @mihaibudiu . One last question, shouldn't we adjust the expected error message in these lines for the actual expected error message (instead of `OUT_OF_RANGE_MESSAGE`)?
```
 if (numeric != Numeric.DECIMAL5_2 || Bug.CALCITE_2539_FIXED) {
  f.checkCastFails(numeric.maxOverflowNumericString,
    type, OUT_OF_RANGE_MESSAGE, true, castType);  // <-- *****
  f.checkCastFails(numeric.minOverflowNumericString,
    type, OUT_OF_RANGE_MESSAGE, true, castType);  // <-- *****
}
```

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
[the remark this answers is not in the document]
rubenada: I agree with @mihaibudiu , in the absence of clear specs, following the lead of Postgresql seems a good idea. I see the comment was improved to clarify this fact 👍

herunkang2018:
LGTM.

herunkang2018 (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
rubenada: I think the test does not fail because `OUT_OF_RANGE_MESSAGE` is currently defined as `(?s).*`, so any error message would be fine (and I think we could be a bit more specific and test an actually expected error thanks to your improvements).

mihaibudiu: This depended on a BUG_* which I couldn't locate anymore, so I improved the message and removed the reference to the BUG altogether.

rubenada:
LGTM

@mihaibudiu , thanks for your work, no further remarks on my side, I think you can proceed to squash + merge.

PR #3482, opened 2023-10-25
[CALCITE-5921] SqlOperatorFixture.checkFails and checkAggFails don't check runtime failure (follow-up)

rubenada:
Follow-up of [CALCITE-5921](https://issues.apache.org/jira/browse/CALCITE-5921) (https://github.com/apache/calcite/pull/3471) to fix some broken tests

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/SqlOperatorUnparseTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlOperatorUnparseTest.java
@@ -99,6 +99,11 @@ public class SqlOperatorUnparseTest extends CalciteSqlOperatorTest {
     }
   }
 
+  @Override @Disabled("Runtime error message differs after parsing and unparsing")
+  void testBitAndFuncRuntimeFails() {
+    super.testContainsSubstrFunc();
+  }
+
   // Every test that is Disabled below corresponds to a bug.
   // These tests should just be deleted when the corresponding bugs are fixed.
 
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -919,8 +919,8 @@ public class SqlOperatorTest {
   void testCastWithRoundingToScalar(CastType castType, SqlOperatorFixture f) {
     f.setFor(SqlStdOperatorTable.CAST, VmName.EXPAND);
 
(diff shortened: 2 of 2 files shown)


libenchao (approved):
+1, thanks @rubenada for the quick fix!

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -919,8 +919,8 @@ void testCastIntervalToInterval(CastType castType, SqlOperatorFixture f) {
   void testCastWithRoundingToScalar(CastType castType, SqlOperatorFixture f) {
     f.setFor(SqlStdOperatorTable.CAST, VmName.EXPAND);
 
-    f.checkFails("cast(1.25 as int)", "INTEGER", true);
-    f.checkFails("cast(1.25E0 as int)", "INTEGER", true);
+    f.checkScalar("cast(1.25 as int)", 1, "INTEGER NOT NULL");
```
mihaibudiu, line 922: The new version of this test looks correct to me. I wonder why the original test was expected to fail.

core/src/test/java/org/apache/calcite/test/SqlOperatorUnparseTest.java
```diff
@@ -99,6 +99,11 @@ String rewrite(String sql) throws SqlParseException {
     }
   }
 
+  @Override @Disabled("Runtime error message differs after parsing and unparsing")
```
mihaibudiu, line 102: Something is weird here, because the super function called doesn't have the same name. 
But since the test is disabled, it doesn't really matter.
This fixture was created to find bugs in unparsing, and it's possible that it does find such a bug. I will investigate, and if that's the case I will file an issue for it.

rubenada: @mihaibudiu  ooops, my bad, this looks like a copy-paste error (which does not have any impact due to the test being disabled), I meant to write:
```
@Override @Disabled("Runtime error message differs after parsing and unparsing")
void testBitAndFuncRuntimeFails() {
  super.testBitAndFuncRuntimeFails();
}
```
Feel free to correct this in the PR of [CALCITE-5990](https://issues.apache.org/jira/browse/CALCITE-5990).

The reason I disabled it was that the runtime error message contains the SQL query, and it seems that after the parse+unparse, the SQL query was not exactly the same as the original one (e.g. newlines were added), so the actual error was not matching the expected one.

PR #3483, opened 2023-10-26
[CALCITE-6024] ListSqlOperatorTable is inefficient

julianhyde:
ListSqlOperatorTable is inefficient if it contains a large number of operators. It currently examines the operators one by one. ReflectiveSqlOperatorTable (and its subclass, SqlStdOperatorTable) builds a map of operators by name (actually two maps, one case-sensitive and one case-insensitive).

ListSqlOperatorTable should do the same, at least in its immutable form.


core/src/main/java/org/apache/calcite/sql/fun/OracleSqlOperatorTable.java
```diff
@@ -65,15 +69,6 @@ public class OracleSqlOperatorTable extends ReflectiveSqlOperatorTable {
    * Returns the Oracle operator table, creating it if necessary.
    */
   public static synchronized OracleSqlOperatorTable instance() {
```
libenchao, line 71: `synchronized` is no longer needed.

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -1364,6 +1368,7 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
   /**
    * The <code>UNNEST WITH ORDINALITY</code> operator.
    */
+  @LibraryOperator(libraries = {}) // do not include in index
```
libenchao, line 1371: Why do we need to remove it from the standard operators, is it a breaking change?

julianhyde: The operators for `UNNEST` and `UNNEST WTH ORDINALITY` have the same name and syntax, and therefore clash when added to a map. The latter was never accessed by name anyway.

core/src/main/java/org/apache/calcite/sql/fun/OracleSqlOperatorTable.java
julianhyde: good catch; will fix

PR #3485, opened 2023-10-26
 [CALCITE-6075]  The git link to download the source code is incorrect

caicancai:
git clone git://github.com/apache/calcite.git does not have this writing method. It should be https://github.com/apache/calcite.git or git@github.com:apache/calcite.git

### DIFF
--- a/site/develop/index.md
+++ b/site/develop/index.md
@@ -54,7 +54,7 @@ Create a local copy of the Git repository, `cd` to its root directory,
 then build using Gradle:
 
 {% highlight bash %}
-$ git clone git://github.com/apache/calcite.git
+$ git clone https://github.com/apache/calcite.git
 $ cd calcite
 $ ./gradlew build
 {% endhighlight %}


caicancai:
cc @chucheng92 , I just changed it directly in the makerdown file on github, but it was difficult to change it to the commit name. I'm sorry for this.

chucheng92:
> cc @chucheng92 , I just changed it directly in the makerdown file on github, but it was difficult to change it to the commit name. I'm sorry for this.

you can use 'git commit --amend' to modify the local commit name. But you have reopened a new PR, maybe you can try this next time

caicancai:
@chucheng92 Hello, there seems to be some errors in the cli at present, but I have not changed the code in it.

zabetak (approved):
LGTM, will merge soon!

zabetak (commented):
Actually the `git://` URL is not invalid it just points to the use of the git protocol and I believe this used to work in the past. Any idea on why it is not working at the moment?

Why do we need to switch to https?

zabetak:
> Actually the git:// URL is not invalid it just points to the use of the git protocol and I believe this used to work in the past. Any idea on why it is not working at the moment?
> 
> Why do we need to switch to https?

OK, I just noticed that git protocol was deprecated for GitHub sometime ago.

https://github.blog/changelog/2022-03-15-removed-unencrypted-git-protocol-and-certain-ssh-keys/

chucheng92:
> Actually the `git://` URL is not invalid it just points to the use of the git protocol and I believe this used to work in the past. Any idea on why it is not working at the moment?
> 
> Why do we need to switch to https?

i think we can change to "git@github.com:apache/calcite.git", it works. rather than 'https'

PR #3486, opened 2023-10-26
[CALCITE-5949] RexExecutable should return unchanged original expressions when it fails

rubenada:
[CALCITE-5949] RexExecutable should return unchanged original expressions when it fails


rubenada:
This PR supersedes https://github.com/apache/calcite/pull/3390

FYI @libenchao @arkanovicz

mihaibudiu (approved):

libenchao:
+1

PR #3488, opened 2023-10-27
[CALCITE-6075] Site: Cloning source code from GitHub using git protocol fails

caicancai:
fix [#6075](https://issues.apache.org/jira/browse/CALCITE-6075)

### DIFF
--- a/site/_docs/howto.md
+++ b/site/_docs/howto.md
@@ -58,7 +58,7 @@ Create a local copy of the GitHub repository,
 then build using the included Gradle wrapper:
 
 {% highlight bash %}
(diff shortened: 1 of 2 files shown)


caicancai:
cc @zabetak

chucheng92:
maybe use `git clone https://github.com/apache/calcite.git` is better. I have checked some other apache projects such as apache flink, they use `https` such as 

```
git clone https://github.com/apache/flink.git

```
within the readme.

I guess `https` is more convenient, and `ssh` requires some configuration, although the former may be less secure.

PR #3489, opened 2023-10-28
[CALCITE-6066] Add HYPOT function (enabled in Spark library)

herunkang2018:


core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -664,6 +665,7 @@ Builder populate() {
       defineMethod(SINH, BuiltInMethod.SINH.method, NullPolicy.STRICT);
       defineMethod(TAN, BuiltInMethod.TAN.method, NullPolicy.STRICT);
       defineMethod(TANH, BuiltInMethod.TANH.method, NullPolicy.STRICT);
+      defineMethod(HYPOT, BuiltInMethod.HYPOT.method, NullPolicy.ANY);
```
julianhyde, line 668: NullPolicy should be STRICT. ANY means 'returns null only if all arguments are null'.

In my opinion, HYPOT(3, NULL) and HYPOT(NULL, 4) should return null, which is consistent with STRICT.

Also test that the type is `DOUBLE` (not `DOUBLE NOT NULL`) when 1 argument is nullable. (Maybe `checkNull` could do that checking?)

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -664,6 +665,7 @@ Builder populate() {
...
       defineMethod(TRUNC, BuiltInMethod.STRUNCATE.method, NullPolicy.STRICT);
```
julianhyde, line 669: Move HYPOT up so that it is after DEGREES and before IS_INF. The list is approximately ordered, so let's keep it that way.

julianhyde (changes requested):
Nearly there. Need the strictness fixed before I can merge.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
herunkang2018: In the doc, ANY means 'If any of the arguments are null, return null', and ALL means 'returns null only if all arguments are null'.

NullPolicy's class comment says STRICT and ANY are similar. And I checked that in codegen phase, STRICT and ANY is the same when generates the null condition, see more at `AbstractRexCallImplementor.getCondition`.

BTW I find that NullPolicy.STRICT's comment is not very clear, it says 'Returns null if and only if one of the arguments are null', which makes me confused that more arguments are null will do not return null.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
herunkang2018: Resolved.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
herunkang2018: `checkNull` onlys check result, so need to check type individually using `checkType`. Added test case for this.

julianhyde: I agree that the javadoc for STRICT is a little ambiguous. It's not clear whether it should return NULL if two arguments are NULL.

Can you please change the javadoc for STRICT from "Returns null if and only if one of the arguments are null." to "Returns null if and only if at least one of the arguments is null."

I have logged https://issues.apache.org/jira/browse/CALCITE-6079 for `checkNull`.

PR #3492, opened 2023-10-31
[CALCITE-6074] The size of REAL, DOUBLE, and FLOAT is not consistent

mihaibudiu:
This is based on a code audit.
The assumption is that FLOAT=DOUBLE=64 bit and REAL=32 bit
I also deleted some deprecated functions instead of fixing them.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
@@ -491,24 +491,6 @@ public abstract class SqlTypeUtil {
     return typeName == SqlTypeName.DECIMAL;
   }
 
-  /** Returns whether a type is DOUBLE. */
-  public static boolean isDouble(RelDataType type) {
-    SqlTypeName typeName = type.getSqlTypeName();
-    if (typeName == null) {
-      return false;
-    }
-    return typeName == SqlTypeName.DOUBLE;
-  }
-
-  /** Returns whether a type is BIGINT. */
-  public static boolean isBigint(RelDataType type) {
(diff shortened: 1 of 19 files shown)


mihaibudiu:
Does anyone want to review this PR?
It relies on the core assumption that in Calcite FLOAT=DOUBLE.

This PR clearly fixes multiple bugs caused by misunderstandings about the floating point types.
In our compiler backend we have disabled the FLOAT type altogether (using a shuttle), because I think it's too dangerous for users to use correctly. My original proposal was to remove the FLOAT type from the Calcite IR and generate DOUBLE in the parser. Then such mistakes would become impossible in the future. This PR is less radical.

dssysolyatin:
@mihaibudiu PR looks good for me. I faced with the same problem when tried to convert sql from postgresql dialect to spark dialect. I was really surprise that spark float is 4 byte and postgres float is double precision by default.

dssysolyatin (approved):

core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
```diff
@@ -704,82 +686,6 @@ public static long getMaxValue(RelDataType type) {
     }
   }
 
-  /** Returns whether a type has a representation as a Java primitive (ignoring
-   * nullability). */
-  @Deprecated // to be removed before 2.0
```
mihaibudiu, line 709: Apparently I can't delete deprecated functions, so I will put these back.

mihaibudiu:
I am planning to merge this PR soon. If you have comments now is the time.
It touches a lot of different parts.

PR #3493, opened 2023-10-31
[CALCITE-6081] Remove bouncycastle dependency

macroguo-ghy:


zabetak (changes requested):
The bouncycastle dependency may be coming transitively from the Spark module.

By removing the entries from apiv I am not sure if you really remove the dependency. I am afraid that the proposed changes will just lead to an older version of bouncycastle to be fetched and used.

macroguo-ghy:
Hi @zabetak. Thanks for your review. You are right! bouncycastle is still present in dependency tree(In fact, `org.apache.pig` introduced it). So I have maed a exclusion in configations. I have checked by executing `./gradlew -q dependencies` and confirmed that bouncycastle is no longer a dependency.

zabetak (approved):
The exclusion may break parts of the module at runtime but given that we have no test coverage, neither claims, for cryptographic support in Calcite I think it is safe to add the exclusion. LGTM!

PR #3495, opened 2023-11-01
[CALCITE-6044] RelMetadataQuery should regard single-row relational expressions as unique

PaulJackson123:
A single-row relation can result from a LIMIT 1 or an aggregation without GROUP BY. Every column in one of these relations should be unique by virtue of having a max row count of 1.

When joining with a single-row relation on a key field, the join result should no longer require that key field for uniqueness. For example, suppose the emp table had a composite key (empno,hiredate). If we join on hiredate=max(hiredate) then empno alone should be a unique column.

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/RelOptPredicateList.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptPredicateList.java
@@ -18,12 +18,9 @@ package org.apache.calcite.plan;
 
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexCall;
-import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexNode;
-import org.apache.calcite.rex.RexSlot;
 import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlKind;
-import org.apache.calcite.util.ImmutableBitSet;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -247,38 +244,5 @@ public class RelOptPredicateList {
       return true;
     }
     return false;
-  }
-
-  /** Return the set of columns that are set to a constant or scalar. */
-  public ImmutableBitSet getInvariantColumnSet() {
-    ImmutableBitSet.Builder builder = ImmutableBitSet.builder();
-    constantMap.keySet()
-        .stream()
-        .filter(RexInputRef.class::isInstance)
-        .map(RexInputRef.class::cast)
-        .map(RexSlot::getIndex)
-        .forEach(builder::set);
-
-    pulledUpPredicates.forEach(rex -> {
-      if (rex.getKind() == SqlKind.EQUALS
-          || rex.getKind() == SqlKind.IS_NOT_DISTINCT_FROM) {
-        List<RexNode> ops = ((RexCall) rex).getOperands();
-        RexNode op0 = ops.get(0);
-        RexNode op1 = ops.get(1);
-        addInputRefIfOtherInvariant(builder, op0, op1);
-        addInputRefIfOtherInvariant(builder, op1, op0);
-      }
-    });
-
-    return builder.build();
-  }
-
-  private void addInputRefIfOtherInvariant(ImmutableBitSet.Builder builder, RexNode inputRef,
-      RexNode invariant) {
-    if (inputRef instanceof RexInputRef
-        && (invariant.getKind() == SqlKind.LITERAL
-        || invariant.getKind() == SqlKind.SCALAR_QUERY)) {
-      builder.set(((RexInputRef) inputRef).getIndex());
-    }
   }
 }
--- a/core/src/main/java/org/apache/calcite/plan/RelOptPredicateList.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptPredicateList.java
@@ -18,9 +18,12 @@ package org.apache.calcite.plan;
 
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexSlot;
 import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.util.ImmutableBitSet;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -244,5 +247,37 @@ public class RelOptPredicateList {
       return true;
     }
     return false;
+  }
+
+  /** Return the set of columns that are set to a constant or scalar. */
+  public ImmutableBitSet getInvariantColumnSet() {
+    ImmutableBitSet.Builder builder = ImmutableBitSet.builder();
+    constantMap.keySet()
+        .stream()
+        .filter(RexInputRef.class::isInstance)
+        .map(RexInputRef.class::cast)
+        .map(RexSlot::getIndex)
+        .forEach(builder::set);
+
+    pulledUpPredicates.forEach(rex -> {
+      if (rex.getKind() == SqlKind.EQUALS
+          || rex.getKind() == SqlKind.IS_NOT_DISTINCT_FROM) {
+        List<RexNode> ops = ((RexCall) rex).getOperands();
+        RexNode op0 = ops.get(0);
+        RexNode op1 = ops.get(1);
+        if (op0 instanceof RexInputRef
+            && (op1.getKind() == SqlKind.LITERAL
+            || op1.getKind() == SqlKind.SCALAR_QUERY)) {
+          builder.set(((RexInputRef) op0).getIndex());
+        }
+        if (op1 instanceof RexInputRef
+            && (op0.getKind() == SqlKind.LITERAL
+            || op0.getKind() == SqlKind.SCALAR_QUERY)) {
+          builder.set(((RexInputRef) op1).getIndex());
+        }
+      }
+    });
+
+    return builder.build();
   }
 }
--- a/core/src/main/java/org/apache/calcite/plan/RelOptPredicateList.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptPredicateList.java
@@ -265,19 +265,20 @@ public class RelOptPredicateList {
         List<RexNode> ops = ((RexCall) rex).getOperands();
         RexNode op0 = ops.get(0);
         RexNode op1 = ops.get(1);
-        if (op0 instanceof RexInputRef
-            && (op1.getKind() == SqlKind.LITERAL
-            || op1.getKind() == SqlKind.SCALAR_QUERY)) {
-          builder.set(((RexInputRef) op0).getIndex());
-        }
-        if (op1 instanceof RexInputRef
-            && (op0.getKind() == SqlKind.LITERAL
-            || op0.getKind() == SqlKind.SCALAR_QUERY)) {
-          builder.set(((RexInputRef) op1).getIndex());
-        }
+        addInputRefIfOtherInvariant(builder, op0, op1);
+        addInputRefIfOtherInvariant(builder, op1, op0);
       }
     });
 
     return builder.build();
+  }
+
+  private void addInputRefIfOtherInvariant(ImmutableBitSet.Builder builder, RexNode inputRef,
+      RexNode invariant) {
+    if (inputRef instanceof RexInputRef
+        && (invariant.getKind() == SqlKind.LITERAL
+        || invariant.getKind() == SqlKind.SCALAR_QUERY)) {
+      builder.set(((RexInputRef) inputRef).getIndex());
+    }
   }
 }
--- a/core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
@@ -83,13 +83,18 @@ public abstract class BuiltInMetadata {
      * represented as an {@link org.apache.calcite.util.ImmutableBitSet}, where
      * each bit position represents a 0-based output column ordinal.
      *
+     * <p>Note that a key plus other columns is still unique. Therefore a table
+     * with all columns unique has a unique key consisting of the empty set.
+     * This is the case with all single-row tables.
+     *
      * <p>Nulls can be ignored if the relational expression has filtered out
      * null values.
      *
      * @param ignoreNulls if true, ignore null values when determining
      *                    whether the keys are unique
      * @return set of keys, or null if this information cannot be determined
-     * (whereas empty set indicates definitely no keys at all)
+     * (whereas empty set indicates definitely no keys at all, and a set
+     * containing the empty set implies every column is unique)
      */
     @Nullable Set<ImmutableBitSet> getUniqueKeys(boolean ignoreNulls);
 
--- a/core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
@@ -84,8 +84,11 @@ public abstract class BuiltInMetadata {
      * each bit position represents a 0-based output column ordinal.
      *
      * <p>Note that a unique key plus other columns is still unique.
-     * Conversely, all columns are unique in a table with a unique key
-     * consisting of the empty set, as is the case for all single-row tables.
+     * Therefore, all columns are unique in a table with a unique key
+     * consisting of the empty set, as is the case for zero-row and
+     * single-row tables. The converse is not true: a table with all
+     * columns unique does necessary have the empty set as a key -
+     * that is never true with multi-row tables.
      *
      * <p>Nulls can be ignored if the relational expression has filtered out
      * null values.
--- a/core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
@@ -83,9 +83,9 @@ public abstract class BuiltInMetadata {
      * represented as an {@link org.apache.calcite.util.ImmutableBitSet}, where
      * each bit position represents a 0-based output column ordinal.
      *
-     * <p>Note that a key plus other columns is still unique. Therefore a table
-     * with all columns unique has a unique key consisting of the empty set.
-     * This is the case with all single-row tables.
+     * <p>Note that a unique key plus other columns is still unique.
+     * Conversely, all columns are unique in a table with a unique key
+     * consisting of the empty set, as is the case for all single-row tables.
      *
      * <p>Nulls can be ignored if the relational expression has filtered out
      * null values.
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
@@ -382,21 +382,9 @@ public class RelMdColumnUniqueness
       }
 
       if (Aggregate.isSimple(rel)) {
-        // If an input's unique column value is returned (passed through) by an aggregation
-        // function, then the result of the function is also unique.
         Set<ImmutableBitSet> inputUniqueKeys = mq.getUniqueKeys(rel.getInput(), ignoreNulls);
-        if (inputUniqueKeys != null) {
-          List<AggregateCall> aggCallList = rel.getAggCallList();
-          for (int aggIndex = 0; aggIndex < aggCallList.size(); aggIndex++) {
-            AggregateCall call = aggCallList.get(aggIndex);
-            if (PASSTHROUGH_AGGREGATIONS.contains(call.getAggregation().getKind())) {
-              Integer inputIndex = call.getArgList().get(0);
-              if (inputUniqueKeys.contains(ImmutableBitSet.of(inputIndex))
-                  && columns.contains(ImmutableBitSet.of(aggIndex + rel.getGroupCount()))) {
-                return true;
-              }
-            }
-          }
+        if (inputUniqueKeys != null && columnsContainUniqueAgg(columns, rel, inputUniqueKeys)) {
+          return true;
         }
       }
 
@@ -412,6 +400,30 @@ public class RelMdColumnUniqueness
       return mq.areColumnsUnique(rel.getInput(), targetColumns.build(), ignoreNulls);
     }
     return null;
+  }
+
+  /**
+   * If an input's unique column value is returned (passed through) by an aggregation function, then
+   * the result of the function is also unique.
+   *
+   * @param columns         the columns to test for uniqueness
+   * @param rel             the Aggregate {@link RelNode}
+   * @param inputUniqueKeys the set of unique key sets of the Aggregate input (child RelNode)
+   *
+   * @return whether the input columns are unique by containing a unique aggregated column
+   */
+  private boolean columnsContainUniqueAgg(ImmutableBitSet columns, Aggregate rel,
+      Set<ImmutableBitSet> inputUniqueKeys) {
+    List<AggregateCall> aggCallList = rel.getAggCallList();
+    for (int aggIndex = 0; aggIndex < aggCallList.size(); aggIndex++) {
+      AggregateCall call = aggCallList.get(aggIndex);
+      if (PASSTHROUGH_AGGREGATIONS.contains(call.getAggregation().getKind())
+          && inputUniqueKeys.contains(ImmutableBitSet.of(call.getArgList().get(0)))
+          && columns.contains(ImmutableBitSet.of(aggIndex + rel.getGroupCount()))) {
+        return true;
+      }
+    }
+    return false;
   }
 
   public Boolean areColumnsUnique(Values rel, RelMetadataQuery mq,
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
@@ -19,10 +19,10 @@ package org.apache.calcite.rel.metadata;
 import org.apache.calcite.plan.RelOptPredicateList;
 import org.apache.calcite.plan.volcano.RelSubset;
 import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.RelShuttleImpl;
 import org.apache.calcite.rel.SingleRel;
 import org.apache.calcite.rel.convert.Converter;
 import org.apache.calcite.rel.core.Aggregate;
-import org.apache.calcite.rel.core.AggregateCall;
 import org.apache.calcite.rel.core.Calc;
 import org.apache.calcite.rel.core.Correlate;
 import org.apache.calcite.rel.core.Exchange;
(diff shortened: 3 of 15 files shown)


core/src/main/java/org/apache/calcite/plan/RelOptPredicateList.java
```diff
@@ -245,4 +248,36 @@ public boolean isEffectivelyNotNull(RexNode e) {
     }
     return false;
   }
+
+  /** Return the set of columns that are set to a constant or scalar. */
+  public ImmutableBitSet getInvariantColumnSet() {
```
PaulJackson123, line 253: I am not committed to the word 'invariant'. I was looking for a word that describes something known to be constant that isn't necessarily a constant literal.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -67,6 +70,9 @@ public class RelMdColumnUniqueness
   public static final RelMetadataProvider SOURCE =
       ReflectiveRelMetadataProvider.reflectiveSource(
           new RelMdColumnUniqueness(), BuiltInMetadata.ColumnUniqueness.Handler.class);
+  static final Set<SqlKind> PASSTHROUGH_AGGREGATIONS =
```
PaulJackson123, line 73: If a row has a unique column, then an aggregation that returns a value from that column is also to be unique. These are the aggregations that do that,

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -266,21 +276,13 @@ public Boolean areColumnsUnique(Intersect rel, RelMetadataQuery mq,
       }
     }
 
-    // If no columns can affect uniqueness, then return unknown
-    if (childColumns.cardinality() == 0) {
-      return null;
-    }
-
```
PaulJackson123, line 273: An empty set can be unique. It is when all columns in the relation are unique.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -266,21 +276,13 @@ public Boolean areColumnsUnique(Intersect rel, RelMetadataQuery mq,
...
     return mq.areColumnsUnique(rel.getInput(), childColumns.build(),
         ignoreNulls);
   }
 
   public @Nullable Boolean areColumnsUnique(Join rel, RelMetadataQuery mq,
       ImmutableBitSet columns, boolean ignoreNulls) {
     columns = decorateWithConstantColumnsFromPredicates(columns, rel, mq);
-    if (columns.cardinality() == 0) {
-      return false;
-    }
```
PaulJackson123, line 283: An empty set can be unique.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -348,11 +361,15 @@ public Boolean areColumnsUnique(Intersect rel, RelMetadataQuery mq,
       return leftJoinColsUnique && rightUnique;
     }
 
-    throw new AssertionError();
```
PaulJackson123, line 351: The constraint that `columns.cardinality() == 0` is removed at the beginning of this method.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -1302,16 +1461,15 @@ private static ImmutableBitSet bitSetOf(int... bits) {
...
   }
 
   @Test void calcMultipleColumnsAreUniqueCalc3() {
     sql("select a1.empno, a2.empno, a2.empno\n"
         + " from emp a1 join emp a2\n"
         + " on (a1.empno=a2.empno)")
         .convertingProjectAsCalc()
-        .assertThatUniqueKeysAre(bitSetOf(0), bitSetOf(0, 1), bitSetOf(0, 1, 2),
-            bitSetOf(0, 2), bitSetOf(1), bitSetOf(1, 2), bitSetOf(2));
+        .assertThatUniqueKeysAre(bitSetOf(0), bitSetOf(1), bitSetOf(1, 2), bitSetOf(2));
```
PaulJackson123, line 1472: Ideally this would be:
```
        .assertThatUniqueKeysAre(bitSetOf(0), bitSetOf(1), bitSetOf(2));
```
I thought digging into this was scope creep.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -4571,7 +4571,7 @@ private void checkEmptyJoin(RelOptFixture f) {
       final RelDataType type = rexBuilder.getTypeFactory().createSqlType(SqlTypeName.BIGINT);
 
       RelNode left = b
-          .values(new String[]{"x", "y"}, 1, 2).build();
+          .values(new String[]{"x", "y"}, 1, 2, 2, 1).build();
```
PaulJackson123, line 4574: I added a second row to prevent `RexSimplify` from seeing the columns as constants and reducing this to `true`.

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -3813,11 +3813,10 @@ LogicalProject(DEPTNO=[$9], SAL=[$7])
     LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
     LogicalJoin(condition=[true], joinType=[left])
       LogicalTableScan(table=[[CATALOG, SALES, EMP]])
-      LogicalAggregate(group=[{}], agg#0=[SINGLE_VALUE($0)])
-        LogicalProject(EXPR$0=[>($0, 0)])
-          LogicalAggregate(group=[{}], agg#0=[AVG($0)])
-            LogicalProject(SAL=[$5])
-              LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+      LogicalProject(EXPR$0=[>($0, 0)])
+        LogicalAggregate(group=[{}], agg#0=[AVG($0)])
```
PaulJackson123, line 3817: There's only one row so $0 is unique, and so `Blackboard.reRegister` is satisfied:
```
          relBuilder.aggregate(relBuilder.groupKey(),
              relBuilder.aggregateCall(SqlStdOperatorTable.SINGLE_VALUE,
                  relBuilder.field(0)));
```
doesn't need an additional aggregation.

core/src/test/resources/sql/sub-query.iq
```diff
@@ -1044,9 +1044,8 @@ EnumerableCalc(expr#0..2=[{inputs}], expr#3=[IS NOT NULL($t2)], SAL=[$t1], EXPR$
   EnumerableNestedLoopJoin(condition=[true], joinType=[left])
     EnumerableCalc(expr#0..7=[{inputs}], EMPNO=[$t0], SAL=[$t5])
       EnumerableTableScan(table=[[scott, EMP]])
-    EnumerableAggregate(group=[{0}])
-      EnumerableCalc(expr#0..2=[{inputs}], expr#3=[true], expr#4=[10], expr#5=[=($t4, $t0)], cs=[$t3], $condition=[$t5])
-        EnumerableTableScan(table=[[scott, DEPT]])
+    EnumerableCalc(expr#0..2=[{inputs}], expr#3=[true], expr#4=[10], expr#5=[=($t4, $t0)], cs=[$t3], $condition=[$t5])
```
PaulJackson123, line 1047: `Dept.deptno` is unique, and it is set equal to `10`, so result is a single row.

core/src/test/resources/sql/sub-query.iq
```diff
@@ -1904,12 +1901,12 @@ select sal from "scott".emp e
...
 !ok
-EnumerableCalc(expr#0..4=[{inputs}], expr#5=[NOT($t4)], expr#6=[IS NOT NULL($t4)], expr#7=[OR($t5, $t6)], expr#8=[IS NOT TRUE($t7)], SAL=[$t1], $condition=[$t8])
-  EnumerableMergeJoin(condition=[=($2, $3)], joinType=[left])
+EnumerableCalc(expr#0..4=[{inputs}], expr#5=[NOT($t3)], expr#6=[IS NOT NULL($t3)], expr#7=[OR($t5, $t6)], expr#8=[IS NOT TRUE($t7)], SAL=[$t1], $condition=[$t8])
+  EnumerableMergeJoin(condition=[=($2, $4)], joinType=[left])
     EnumerableSort(sort0=[$2], dir0=[ASC])
       EnumerableCalc(expr#0..7=[{inputs}], EMPNO=[$t0], SAL=[$t5], DEPTNO=[$t7])
         EnumerableTableScan(table=[[scott, EMP]])
-    EnumerableCalc(expr#0..2=[{inputs}], expr#3=[true], expr#4=[10], expr#5=[=($t4, $t0)], DEPTNO1=[$t0], $f1=[$t3], $condition=[$t5])
+    EnumerableCalc(expr#0..2=[{inputs}], expr#3=[true], expr#4=[10], expr#5=[=($t4, $t0)], cs=[$t3], DEPTNO1=[$t0], $condition=[$t5])
```
PaulJackson123, line 1909: Would appreciate if this is reviewed carefully. I don't have experience with Calc plan nodes.

core/src/main/java/org/apache/calcite/plan/RelOptPredicateList.java
```diff
@@ -245,4 +248,36 @@ public boolean isEffectivelyNotNull(RexNode e) {
...
+        .map(RexSlot::getIndex)
+        .forEach(builder::set);
+
+    pulledUpPredicates.forEach(rex -> {
+      if (rex.getKind() == SqlKind.EQUALS
+          || rex.getKind() == SqlKind.IS_NOT_DISTINCT_FROM) {
+        List<RexNode> ops = ((RexCall) rex).getOperands();
+        RexNode op0 = ops.get(0);
+        RexNode op1 = ops.get(1);
+        if (op0 instanceof RexInputRef
```
tanclary, line 268: I see two very similar conditional blocks here, would a small helper method help? It would reduce duplicate code and remove one level of nesting. Let me know what you think.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -364,6 +381,25 @@ public Boolean areColumnsUnique(Intersect rel, RelMetadataQuery mq,
         return false;
       }
 
+      if (Aggregate.isSimple(rel)) {
```
tanclary, line 384: 5 levels of nesting might be a bit much, I know this is kind of a vague comment but could you check if there's any opportunities for early returns or just a way of splitting this up?

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -1141,14 +1141,88 @@ private void checkColumnUniquenessForFilterWithConstantColumns(String sql) {
         .assertThatAreColumnsUnique(bitSetOf(0, 1), is(false));
   }
 
+  @Test void testColumnUniquenessForLimit1() {
+    final String sql = ""
+                       + "select *\n"
+                       + "from emp\n"
+                       + "limit 1";
```
tanclary, line 1148: Could we format this similar to other tests (e.g. remove empty string at beginning)? I know there are some other tests in this class like this, but from my experience it is not how most tests are formatted In Calcite

tanclary (changes requested):
Thanks for the PR! I left a couple of comments, my main feedback is that some of the code could use some reduced nesting if possible (this is my personal opinion, feel free to push back).

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
PaulJackson123: I see that it is not consistent. There are quite a few like this in both this class and `RelOptRulesTest`. I did make the change everywhere and it was mostly fine with one exception - `testPullUpPredicatesFromUnion2`. In that case the SQL extends beyond the end of line and requires formatting that, IMO, makes it less pleasing.

I'd like to ask how strongly you feel about this one. One argument for leaving it as is is it makes all the lines of the SQL start on the same column, which I think is nice.

I will fix the issue with too much leading space on lines 1146-1148 and elsewhere. I must not have had my Code Style set properly when I wrote these tests.

PaulJackson123: I can make the change but knowing where to stop is a bit tricky. Do I change just mine, all the ones in this file, all the ones in the files I changed, or all in all test files? I found 550 occurrences in Test, 372 in core:
image

tanclary: Feel free to leave it as is!

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -378,12 +402,36 @@ public Boolean areColumnsUnique(Intersect rel, RelMetadataQuery mq,
...
+   * @param columns         the columns to test for uniqueness
+   * @param rel             the Aggregate {@link RelNode}
+   * @param inputUniqueKeys the set of unique key sets of the Aggregate input (child RelNode)
+   *
+   * @return whether the input columns are unique by containing a unique aggregated column
+   */
+  private boolean columnsContainUniqueAgg(ImmutableBitSet columns, Aggregate rel,
+      Set<ImmutableBitSet> inputUniqueKeys) {
+    List<AggregateCall> aggCallList = rel.getAggCallList();
+    for (int aggIndex = 0; aggIndex < aggCallList.size(); aggIndex++) {
```
julianhyde, line 418: This `for` loop seems backwards to me. It would make more sense to iterate over `columns`, and ask whether each is a pass-through aggregate function. Whereas you loop over all aggregate functions.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -533,6 +535,36 @@ public RelOptPredicateList getPredicates(Exchange exchange,
...
+        }
+      }
+      RexBuilder rexBuilder = values.getCluster().getRexBuilder();
+      List<RexNode> predicates = new ArrayList<>();
+      for (int i = 0; i < constants.size(); i++) {
+        RexLiteral literal = constants.get(i);
+        if (literal != null) {
+          predicates.add(
+              rexBuilder.makeCall(SqlStdOperatorTable.EQUALS,
+              rexBuilder.makeInputRef(literal.getType(), i),
```
julianhyde, line 559: fix indentation

core/src/main/java/org/apache/calcite/rel/metadata/RelMdUniqueKeys.java
```diff
@@ -74,11 +84,30 @@ private RelMdUniqueKeys() {}
...
+      return null;
+    }
+    // Remove constant columns from each key
+    RelOptPredicateList predicates = mq.getPulledUpPredicates(rel);
+    if (RelOptPredicateList.isEmpty(predicates)) {
+      return uniqueKeys;
+    } else {
+      ImmutableBitSet invariantColumns = predicates.getInvariantColumnSet();
+      return uniqueKeys.stream()
+          .map(key -> key.rebuild()
```
julianhyde, line 98: Use `ImmutableBitSet.except`

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
julianhyde: Also consider making this an interface, similar to `SqlStaticAggFunction`, `SqlSingletonAggFunction`. That way, user-defined aggregate functions can have this property.

julianhyde (changes requested):
Good start. With some revisions, we can definitely get this done before 1.37.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
julianhyde: Definitely don't go changing unrelated code. When a reviewer is asking for consistency, they are never asking for that.

There often inconsistencies when people have applied judgment. If there you can find reasons for those inconsistencies - even if they are reasons you don't personally agree with - it's not constructive to harp on those inconsistencies, because in so doing, you are criticizing people for applying judgment.

I would tend to write
```
String sql1 = "select x\n"
  + "from t";
String sql2 = ""
  + "select (select * from aVeryComplicatedSubqueryThatWouldMakeTheLineTooLong)\n"
  + from t"; 
```
There is a consistent rule there, if you look for it.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
julianhyde: When you have pushed a new commit, please make a comment on the Jira case and I will re-review. This PR seems close to completion.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -526,12 +525,78 @@ private static ImmutableBitSet decorateWithConstantColumnsFromPredicates(
...
+          filter.getCondition().accept(new RexShuttle() {
+            @Override public RexNode visitFieldAccess(final RexFieldAccess fieldAccess) {
+              if (fieldAccess.getReferenceExpr().getKind() == SqlKind.CORREL_VARIABLE) {
+                hasCorrelatedVars.setTrue();
+              }
+              return super.visitFieldAccess(fieldAccess);
+            }
+          });
+          return super.visit(filter);
+        }
```
PaulJackson123, line 596: This change was made to account to the fact that scalar subqueries are not constant when they have correlating variables. Not sure this is the best way to detect correlating variables. Also, only checking filters here - not sure what to do about Calc or others.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdUniqueKeys.java
```diff
@@ -334,30 +334,23 @@ public Set<ImmutableBitSet> getUniqueKeys(Aggregate rel, RelMetadataQuery mq,
...
       if (inputUniqueKeys != null) {
-        List<AggregateCall> aggCallList = rel.getAggCallList();
-        for (int aggIndex = 0; aggIndex < aggCallList.size(); aggIndex++) {
-          AggregateCall call = aggCallList.get(aggIndex);
-          if (PASSTHROUGH_AGGREGATIONS.contains(call.getAggregation().getKind())) {
-            Integer inputIndex = call.getArgList().get(0);
-            if (inputUniqueKeys.contains(ImmutableBitSet.of(inputIndex))) {
-              keysBuilder.add(ImmutableBitSet.of(aggIndex + rel.getGroupCount()));
-            }
-          }
```
PaulJackson123: This was half-baked - it wasn't accounting for multiple mappings from input to output or multi-column keys.

core/src/main/java/org/apache/calcite/sql/SqlOverOperator.java
```diff
@@ -115,6 +115,7 @@ public SqlOverOperator() {
 
     // Copied from validateOperands
     validator.setValidatedNodeType(call, ret);
+    // TODO: Should the type of the agg change just because it is an OVER operand?
```
PaulJackson123, line 118: I removed the following line to work around [CALCITE-6126], but that broke other tests in ways I did not understand, so I left this comment. I'm happy to remove it if this is not best practice.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -1211,6 +1211,24 @@ private void checkColumnUniquenessForFilterWithConstantColumns(String sql) {
...
+        + "from emp A\n"
+        + "where empno = (\n"
+        + "  select max(empno)\n"
+        + "  from emp B\n"
+        + "  where A.deptno = B.deptno\n"
+        + ")";
+    sql(sql)
+        .assertThatAreColumnsUnique(bitSetOf(0), is(true))
+        // This requires drilling into the subquery
+//        .assertThatAreColumnsUnique(bitSetOf(1), is(true))
```
PaulJackson123, line 1226: This test was added to test whether we do not see a scalar subquery as constant if it has correlating variables, which it does not, as designed, however, for more complex reasons that were not addressed in this PR, `deptno` is a unique column.

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
```diff
@@ -83,13 +83,18 @@ public interface UniqueKeys extends Metadata {
      * represented as an {@link org.apache.calcite.util.ImmutableBitSet}, where
      * each bit position represents a 0-based output column ordinal.
      *
+     * <p>Note that a key plus other columns is still unique. Therefore a table
+     * with all columns unique has a unique key consisting of the empty set.
+     * This is the case with all single-row tables.
```
julianhyde, line 88: Not true. The tuple with zero columns has only one possible value. Therefore it is unique only in tables with 0 or 1 rows. (In such tables every combination of columns is unique.)

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -68,6 +79,15 @@ public class RelMdColumnUniqueness
       ReflectiveRelMetadataProvider.reflectiveSource(
           new RelMdColumnUniqueness(), BuiltInMetadata.ColumnUniqueness.Handler.class);
 
+  /**
+   * If a row has a unique column, then an aggregation that returns a value from that column is
+   * also unique. These are the aggregations that do that. Note that this quality is not
+   * guaranteed in the presence of an OVER clause. NOTE: if a multi-argument function is added,
+   * methods that use this Set must be enhanced to select the appropriate column to pass through.
+   */
+  static final Set<SqlKind> PASSTHROUGH_AGGREGATIONS =
```
julianhyde, line 88: The first sentence should be a description of the field. "The set of aggregate functions A such that if x is unique then A(x) will also be unique. An aggregate function with this property is called 'passthrough'."

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -533,6 +535,36 @@ public RelOptPredicateList getPredicates(Exchange exchange,
     return mq.getPulledUpPredicates(input);
   }
 
+  /**
+   * Infers predicates for a Values.
+   */
```
julianhyde, line 540: A comment
```
The predicates on T (x, y, z) with rows (1, 2, null), (1, 2, null), (5, 2, null) are 'y = 2' and 'z is null'
```
would be useful.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -478,17 +525,78 @@ private static ImmutableBitSet decorateWithConstantColumnsFromPredicates(
...
       if (!constantIndexes.isEmpty()) {
         return checkingColumns.union(ImmutableBitSet.of(constantIndexes));
       }
     }
     // If no constant columns deduced, return the original "checkingColumns".
     return checkingColumns;
   }
+
+  /**
+   * Return the set of columns that are set to a constant literal or a scalar query (as
```
julianhyde, line 538: s/Return/Returns/

core/src/main/java/org/apache/calcite/sql/SqlOverOperator.java
julianhyde: Yes, we don't want to add TODOs. They never get done.

If you have a question, ask it in jira.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -478,17 +525,78 @@ private static ImmutableBitSet decorateWithConstantColumnsFromPredicates(
...
+   * in a correlated subquery). Examples of constants are {@code x} in the following:
+   * <pre>SELECT x FROM table WHERE x = 5</pre>
+   * and
+   * <pre>SELECT x, y FROM table WHERE x = (SELECT MAX(x) FROM table)</pre>
+   * <p/>
```
julianhyde, line 543: Remove `</p>`. We don't use end-of-paragraph markers.

Add `<p>` at start of line, after blank line.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
julianhyde: Could you use `RexUtil.containsCorrelation(filter.getCondition())`?

julianhyde (changes requested):

core/src/main/java/org/apache/calcite/rel/metadata/BuiltInMetadata.java
```diff
@@ -83,9 +83,9 @@ public interface UniqueKeys extends Metadata {
      * represented as an {@link org.apache.calcite.util.ImmutableBitSet}, where
      * each bit position represents a 0-based output column ordinal.
      *
-     * <p>Note that a key plus other columns is still unique. Therefore a table
-     * with all columns unique has a unique key consisting of the empty set.
-     * This is the case with all single-row tables.
+     * <p>Note that a unique key plus other columns is still unique.
+     * Conversely, all columns are unique in a table with a unique key
+     * consisting of the empty set, as is the case for all single-row tables.
```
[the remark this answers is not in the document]
julianhyde: Re-reading, I guess your comment is true, but it is extremely misleading. Only 0 and 1 row tables have `()` as a key. From that it follows that all rows are unique, all columns are unique, and all combinations of columns are unique. But we knew that, because the table had 0 or 1 rows.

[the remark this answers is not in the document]
PaulJackson123: I changed the wording to accommodate the the zero-column case that you mentioned - all zero-or-one-row tables have `()` as a key, but not all tables that have all columns unique do.

However, I think it is true that all (even multi-row) tables with one or more columns that have all columns unique have `()` as a unique key. So I might be missing another point.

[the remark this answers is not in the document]
julianhyde: x is a key if, for two rows, their values of x will be different if are not the same row.

(x, y) is a key if, for two rows, their values of x and y will be the different if they are not the same row.

() is a key if, for two rows the value of () will be different if they are not the same row. But () is a constant. So the table cannot have more than one row.

Sorry to keep hammering on this. But it's like trying to discuss logic if we disagree on the meaning of true and false. (The above is closely related to the finding in logic that "false implies anything".)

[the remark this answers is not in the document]
PaulJackson123: No, this is good. I see your point now. In retrospect, your zero-column argument should have been sufficient. My confusion went like this:
1. If (A) is a unique key, (A,B) is also a unique key
2. A 0-or-1-row table has () as a unique key
3. All columns in a table with () as a key are unique
4. A table with all columns unique has () as a key

I didn't catch the leap between 3 and 4. I've been quietly unsettled about empty keys and this clarification helps with that.

julianhyde:
I think "RelMetadataQuery should regard single-row relational expressions as unique" would be a better title than "Make single-row rels unique".

They don't need to be made unique; they are already unique, they just need to be recognized as such.

PR #3501, opened 2023-11-03
[CALCITE-6091] Char that in array or map is truncated if CASE WHEN statement contains null

JiajunBernoulli:


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10497,6 +10497,27 @@ private static void checkArrayConcatAggFuncFails(SqlOperatorFixture t) {
         "[1         , 2         , Hi        , null]", "CHAR(10) ARRAY NOT NULL");
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6091">[CALCITE-6091]
+   * Char that in array or map is truncated in CASE WHEN statement</a>.
+   */
+  @Test void testTruncatedCharInCaseWhen() {
```
mihaibudiu, line 10505: I don't understand how this test relates to the change in the code.
The code is about nullability, but this test does not have any null values.
But the issue is about the type inference across cases in a case statement.
This would suggest that the bug is actually in leastRestrictive, which doesn't handle properly nullable types?

JiajunBernoulli: @mihaibudiu , Thanks for your review.

I change the JIRA summary because https://issues.apache.org/jira/browse/CALCITE-4603 has been resolved.
I just add the case if `types` has null.

mihaibudiu: Where do these new tests use null values?
Will these tests pass if you remove your fixes? If yes, then these tests are not sufficient.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10497,6 +10497,29 @@ private static void checkArrayConcatAggFuncFails(SqlOperatorFixture t) {
...
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6091">[CALCITE-6091]
+   * Char that in array or map is truncated if CASE WHEN statement contains null</a>.
+   */
+  @Test void testTruncatedCharInCaseWhen() {
+    final SqlOperatorFixture fixture = fixture().withLibrary(SqlLibrary.SPARK);
+    fixture.check("select case "
+            + "when true then array('abc') "
+            + "when false then array('d') "
+            + "else null "
```
JiajunBernoulli, line 10510: @mihaibudiu 
Here is null value.
> When we omit `else`, it is also null.

If I remove my fix, the test will failed.

PR #1733, opened 2020-01-08
[CALCITE-3679] Allow lambda expressions in SQL queries

ritesh-kapoor:

PR #3502, opened 2023-11-04
[CALCITE-3679] Allow lambda expressions in SQL queries

macroguo-ghy:
This PR is about three things:

1. Implementing parsing, validating, and converting SqlNode to RelNode, and RelNode to SqlNode for lambda expressions.
2. Implementing the conversion of lambda expression RexNode to Linq4j Expression to execute higher order functions.
3. Adding the EXISTS function, which is enabled in the Spark library.

## Parse
Lambda expression is only accepted as a parameter of function call.
```sql
select func(1, (x, y) -> x + y); // accepted
select (x,y) -> x + y; // illegal for now
```

## Validate
I add a new type `LAMBDA_EXPRESSION` to describe type of lambda expression.
For example:
```sql
  public static final SqlFunction EXISTS =
      SqlBasicFunction.create("EXISTS",
          ReturnTypes.BOOLEAN_NULLABLE,
          OperandTypes.sequence("EXISTS(<ARRAY>, <FUNCTION(ANY)->BOOLEAN>)",
              OperandTypes.ARRAY, OperandTypes.lambda(SqlTypeFamily.BOOLEAN, SqlTypeFamily.ANY)));
```
The type of lambda in `exists` function is a function type where the input parameter is of type `ANY` and the return type is of type `BOOLEAN`.
The lambda expression will be validated twice. For the first time, calcite will validate the lambda expression itself, which means all the types of parameters of lambda expression are considered as `ANY`. For example, `(x, y) -> x + y`, both `x` and `y` are `ANY` type, we can deduce that `x + y` is also`ANY`, so the type of `(x, y) -> x + y` is `FUNCTION(ANY, ANY) -> ANY`. And then we need to know if the type of lambda is legal in a function, this is what `LambdaExpressionOperandTypeChecker` do. It will reset the type of parameters, and infer the type of lambda expression, and check if the derived type satisfy this checker. After the checker, we can infer the real type of lamda expression.

## RelNode
In this part, I create a class `RexLambdaRef` to reference the parameter of lambda expression. From another perspective, we can view lambda expressions as "constants" rather than tables, so we cannot directly use 'RexInputRef' and 'RexLocalRef'`.

----
I referred to #1733, so I listed Ritesh as a co-author.

### DIFF
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -73,6 +73,7 @@ import org.apache.calcite.sql.SqlJsonQueryWrapperBehavior;
 import org.apache.calcite.sql.SqlJsonValueEmptyOrErrorBehavior;
 import org.apache.calcite.sql.SqlJsonValueReturning;
 import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlLambda;
 import org.apache.calcite.sql.SqlLiteral;
 import org.apache.calcite.sql.SqlMatchRecognize;
 import org.apache.calcite.sql.SqlMerge;
@@ -1034,6 +1035,9 @@ void AddArg0(List<SqlNode> list, ExprContext exprContext) :
     (
         e = Default()
     |
+        LOOKAHEAD((SimpleIdentifierOrList() | <LPAREN> <RPAREN>) <LAMBDA>)
+        e = LambdaExpression()
+    |
         LOOKAHEAD(3)
         e = TableParam()
     |
@@ -1060,6 +1064,9 @@ void AddArg(List<SqlNode> list, ExprContext exprContext) :
     )
     (
         e = Default()
+    |
+        LOOKAHEAD((SimpleIdentifierOrList() | <LPAREN> <RPAREN>) <LAMBDA>)
+        e = LambdaExpression()
     |
         e = Expression(exprContext)
     |
@@ -3910,6 +3917,43 @@ SqlNode Expression3(ExprContext exprContext) :
                 (List<SqlNode>) list1);
         }
     }
+}
+
+/**
+ * Parses a lambda expression.
+ */
+SqlNode LambdaExpression() :
+{
+    final SqlNodeList parameters;
+    final SqlNode expression;
+    final Span s;
+}
+{
+    parameters = SimpleIdentifierOrListOrEmpty()
+    <LAMBDA> { s = span(); }
+    expression = Expression(ExprContext.ACCEPT_NON_QUERY)
+    {
+        return new SqlLambda(s.end(this), parameters, expression);
+    }
+}
+
+/**
+ * List of simple identifiers in parentheses or empty parentheses or one simple identifier.
+ * <ul>Examples:
+ * <li>{@code ()}
+ * <li>{@code DEPTNO}
+ * <li>{@code (EMPNO, DEPTNO)}
+ * </ul>
+ */
+SqlNodeList SimpleIdentifierOrListOrEmpty() :
+{
+    SqlNodeList list;
+}
+{
+    LOOKAHEAD(2)
+    <LPAREN> <RPAREN> { return SqlNodeList.EMPTY; }
+|
+    list = SimpleIdentifierOrList() { return list; }
 }
 
 SqlOperator periodOperator() :
@@ -8787,6 +8831,7 @@ void NonReservedKeyWord2of3() :
 |   < NE2: "!=" >
 |   < PLUS: "+" >
 |   < MINUS: "-" >
+|   < LAMBDA: "->" >
 |   < STAR: "*" >
 |   < SLASH: "/" >
 |   < PERCENT_REMAINDER: "%" >
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -44,6 +44,8 @@ import org.apache.calcite.rex.RexDynamicParam;
 import org.apache.calcite.rex.RexFieldAccess;
 import org.apache.calcite.rex.RexFieldCollation;
 import org.apache.calcite.rex.RexInputRef;
+import org.apache.calcite.rex.RexLambda;
+import org.apache.calcite.rex.RexLambdaRef;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexLocalRef;
 import org.apache.calcite.rex.RexNode;
@@ -66,6 +68,7 @@ import org.apache.calcite.sql.SqlDynamicParam;
 import org.apache.calcite.sql.SqlIdentifier;
 import org.apache.calcite.sql.SqlJoin;
 import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlLambda;
 import org.apache.calcite.sql.SqlLiteral;
 import org.apache.calcite.sql.SqlMatchRecognize;
 import org.apache.calcite.sql.SqlNode;
@@ -778,6 +781,19 @@ public abstract class SqlImplementor {
         } else {
           return SqlStdOperatorTable.NOT.createCall(POS, node);
         }
+
+      case LAMBDA:
+        final RexLambda lambda = (RexLambda) rex;
+        final SqlNodeList parameters = new SqlNodeList(POS);
+        for (RexLambdaRef parameter : lambda.getParameters()) {
+          parameters.add(toSql(program, parameter));
+        }
+        final SqlNode expression = toSql(program, lambda.getExpression());
+        return new SqlLambda(POS, parameters, expression);
+
+      case LAMBDA_REF:
+        final RexLambdaRef lambdaRef = (RexLambdaRef) rex;
+        return new SqlIdentifier(lambdaRef.getName(), POS);
 
       default:
         if (rex instanceof RexOver) {
--- a/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
+++ b/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
@@ -135,6 +135,17 @@ public interface RelDataTypeFactory {
       RelDataType valueType);
 
   /**
+   * Creates a function type.
+   *
+   * @param parameterType type of parameters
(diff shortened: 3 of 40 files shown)


mihaibudiu:
Is there a design document for this feature?
This looks extremely useful, but before I review it I would like to have a big picture understanding of how it's supposed to work.
For example, is there type inference for lambdas?

Moreover, if EXISTS is independent on the lambdas perhaps it should be in a separate PR.

mihaibudiu:
Leaving aside the design document, I don't see any changes to the documentation files either.

mihaibudiu:
The real tests of whether the design works properly is to allow a lambda expression wherever a function call is allowed.
So you should be allowed to say `SELECT (x -> x + 1)(age) FROM Person` or
`SELECT  ":" || ((x -> x || x)(SUBSTRING(name, 3, 3))) FROM Person WHERE (x -> x > 5)(age)`.
I think there is limited value to having a lambda expression that only works in some very restricted contexts.

macroguo-ghy:
Hi @mihaibudiu, thanks for your review. I have made some commits to address your feedback:
- Add docs about lambda expressions and higher-order functions. 
- Add some negative tests in `SqlParserTest`.
- Log a new jira case [CALCITE-6116](https://issues.apache.org/jira/browse/CALCITE-6116) to track the EXISTS function.

site/_docs/reference.md
```diff
@@ -1205,6 +1205,7 @@ Note:
 | MULTISET | Unordered collection that may contain duplicates | Example: int multiset
 | ARRAY    | Ordered, contiguous collection that may contain duplicates | Example: varchar(10) array
 | CURSOR   | Cursor over the result of executing a query |
+| LAMBDA_EXPRESSION| A function definition that is not bound to an identifier.| Example FUNCTION(INTEGER, VARCHAR(30)) -> INTEGER
```
julianhyde, line 1208: Let's just call this a FUNCTION type. Add spaces around `|`. Remove '.'

You should note that it is not fully supported e.g. in CAST or DDL

site/_docs/reference.md
```diff
@@ -2614,6 +2615,22 @@ Note:
...
+lambdaExpression:
+      parameters '->' expression
+
+parameters:
+      '(' [ identifier [, identifier ] ] ')'
+  |   identifier
+{% endhighlight %}
+
+Higher-order functions are not included in the SQL standard, so all the functions will be listed in the next chapter.
+
```
julianhyde, line 2633: Is it true that higher-order functions must be built-in?

I think it would be helpful to list the higher-order functions here.

testkit/src/main/java/org/apache/calcite/test/MockSqlOperatorTable.java
```diff
@@ -623,4 +625,24 @@ public CompositeFunction() {
       return typeFactory.createSqlType(SqlTypeName.BIGINT);
     }
   }
+
+  private static final SqlFunction HIGHER_ORDER_FUNCTION =
+      new SqlFunction("HIGHER_ORDER_FUNCTION",
```
julianhyde, line 630: Use `SqlBasicFunction.create`.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1033,6 +1034,9 @@ void AddArg0(List<SqlNode> list, ExprContext exprContext) :
     )
     (
         e = Default()
+    |
+        LOOKAHEAD((SimpleIdentifierOrList() | <LPAREN> <RPAREN>) <LAMBDA_OPERATOR>)
+        e = LambdaExpression()
```
julianhyde, line 1039: How expensive is this LOOKAHEAD?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1060,6 +1064,9 @@ void AddArg(List<SqlNode> list, ExprContext exprContext) :
...
```
julianhyde, line 1069: How expensive is this LOOKAHEAD?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -3912,6 +3919,29 @@ SqlNode Expression3(ExprContext exprContext) :
...
+    final SqlNodeList parameters;
+    final SqlNode expression;
+    final Span s;
+}
+{
+    (
+        LOOKAHEAD(2)
+        <LPAREN> <RPAREN> { parameters = SqlNodeList.EMPTY; }
+    |
+        parameters = SimpleIdentifierOrList()
```
julianhyde, line 3936: should there be a rule that allows `()`, `x` and `(x [, y]*)`?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -8787,6 +8817,7 @@ void NonReservedKeyWord2of3() :
 |   < NE2: "!=" >
 |   < PLUS: "+" >
 |   < MINUS: "-" >
+|   < LAMBDA_OPERATOR: "->" >
```
julianhyde, line 8820: just `LAMBDA`

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -779,6 +782,17 @@ public SqlNode toSql(@Nullable RexProgram program, RexNode rex) {
...
 
+      case LAMBDA:
+        final RexLambdaExpression lambda = (RexLambdaExpression) rex;
+        final SqlNodeList parameters = new SqlNodeList(POS);
+        for (RexLambdaRef parameter : lambda.getParameters()) {
+          parameters.add(toSql(program, parameter));
+        }
+        final SqlNode expression = toSql(program, lambda.getExpression());
+        return new SqlLambdaExpression(POS, parameters, expression);
+      case LAMBDA_REF:
```
julianhyde, line 793: add blank line before

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -6683,6 +6683,31 @@ void testGroupExpressionEquivalenceParams() {
...
+        .type("RecordType(INTEGER NOT NULL EXPR$0) NOT NULL");
+
+    s.withSql("select ^HIGHER_ORDER_FUNCTION(1, null)^")
+        .fails("Cannot apply '(?s).*HIGHER_ORDER_FUNCTION' to arguments of type "
+            + "'HIGHER_ORDER_FUNCTION\\(<INTEGER>, <NULL>\\)'.*");
+    s.withSql("select ^HIGHER_ORDER_FUNCTION(1, (x, y, z) -> x + 1)^")
+        .fails("Cannot apply '(?s).*HIGHER_ORDER_FUNCTION' to arguments of type "
+            + "'HIGHER_ORDER_FUNCTION\\(<INTEGER>, <FUNCTION\\(ANY, ANY, ANY\\) -> ANY>\\)'.*");
+  }
+
```
julianhyde, line 6710: need to see more tests for EXISTS. including negative tests. wrong type of arguments, wrong number of arguments, arguments of wrong name, arguments of wrong case.

check type derivation (if you didn't do it in SqlOperatorTest)

julianhyde (changes requested):
This looks very good.

I've suggested quite a few changes to names and scope (e.g. removing `public`) to make this easier to maintain.

Can you add a quidem test? I would like to see this working end-to-end.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
```diff
@@ -134,6 +134,18 @@ RelDataType createMapType(
       RelDataType keyType,
       RelDataType valueType);
 
+  /**
+   * Create a lambda expression type. Lambda expressions are functions that
```
julianhyde, line 138: s/Create/Creates/

core/src/main/java/org/apache/calcite/rex/RexLambdaExpression.java
```diff
@@ -0,0 +1,101 @@
...
+    RexLambdaExpression rexLambdaExpression = (RexLambdaExpression) o;
+    return Objects.equals(expression, rexLambdaExpression.expression)
+        && Objects.equals(parameters, rexLambdaExpression.parameters);
+  }
+
+  @Override public int hashCode() {
+    return Objects.hash(expression, parameters);
+  }
+
+  @Override public String toString() {
```
julianhyde, line 88: why override `toString` rather than `computeDigest`?

I think it would be clearer and more concise if you use a `for` loop rather than functional style.

julianhyde: Best thing is to do what `RexMRAggCall` does - make the class final and initialize `digest` in the constructor

core/src/main/java/org/apache/calcite/sql/type/LambdaExpressionSqlType.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.calcite.linq4j.Ord;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFamily;
+import org.apache.calcite.rel.type.RelDataTypeField;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * SQL lambda expression type.
+ */
```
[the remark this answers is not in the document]
macroguo-ghy: I have renamed this class to `FunctionSqlType`. But I think all the `RelDataType` classes are public, `FunctionSqlType` should be consistent with them.

testkit/src/main/java/org/apache/calcite/test/MockSqlOperatorTable.java
macroguo-ghy: Fixed

macroguo-ghy (commented):
Thanks for your review and encouragement, @julianhyde. I have made some changes based on your comments. And, I will be adding an end-to-end test later.

mihaibudiu:
Do the lambdas allow capturing values?
`SELECT EXISTS(T.a,  x -> x < T.VALUE) FROM T`

macroguo-ghy:
Thank you for your review. I apologize for the delayed response; I have been quite busy over the past few weeks.

Regarding this PR, I have made modifications to almost all of the reviews. And then, I will write more tests. Once that is done, I will split this PR into two parts: one for lambda and another for the exists function

PR #3506, opened 2023-11-07
[CALCITE-6071] RexCall should carry source position information for runtime error reporting

mihaibudiu:
This PR adds source position (`SqlParserPos`) information to two kinds of nodes:
- RexCall
- AggregateCall

Not all RexCall nodes and AggregateCall operations need to have meaningful source position information. The most useful nodes are the ones that could lead to runtime errors. So operations like addition (which can overflow) or casts (which can fail when the target type is too narrow) need to carry such information. Nodes where the operation cannot fail dynamically (e.g., MAX, comparisons, or IS_NULL) may have the source position set to ZERO.

In this PR I have carefully audited the entire code base and tried to thread the flow of the position information everywhere, but there are a few places where I couldn't figure out where it should come from. There are, for example, a few casts inserted by the rewrite rules which have no direct correspondence to source constructs. Such casts have now a TODO comment wondering about their runtime safety.

In the process I have also removed a few `@Deprecated` functions which would have required a signature change. I figured that since they are deprecated there is no point to update them, since no one can call the variant with the new signature.

For methods where there are many calls where the position information may legitimately be ZERO I have left variants of the original methods without source position information (e.g., see RexBuilder.makeCall)

Currently there are no uses of this information, but I have inspected the IR with a debugger after optimizations to check that the information is present. A future commit (part of this PR or another one) should add an example runtime error that displays this information. This is not trivial to do for the Calcite Linq4j-based executor, so I haven't done it yet.

None of this should affect the behavior of any of the compiled programs, but I noticed that it's easy to mess up something in practice a few times. Currently no tests need to change.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/core/AggregateCall.java
+++ b/core/src/main/java/org/apache/calcite/rel/core/AggregateCall.java
@@ -26,6 +26,7 @@ import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.type.SqlTypeUtil;
 import org.apache.calcite.util.ImmutableBitSet;
 import org.apache.calcite.util.Optionality;
@@ -50,6 +51,14 @@ import static java.util.Objects.requireNonNull;
 public class AggregateCall {
   //~ Instance fields --------------------------------------------------------
 
+  /**
+   * Some aggregate calls may produce runtime errors.  For these
+   * we need to keep around the original source position information
+   * so that the runtime can produce error messages pointing to
+   * the offending source operation.  For "safe" aggregations
+   * this field may be ZERO.
+   */
+  private final SqlParserPos pos;
   private final SqlAggFunction aggFunction;
 
   private final boolean distinct;
@@ -84,7 +93,7 @@ public class AggregateCall {
       List<Integer> argList,
       RelDataType type,
       String name) {
-    this(aggFunction, distinct, false, false,
+    this(SqlParserPos.ZERO, aggFunction, distinct, false, false,
         ImmutableList.of(), argList, -1, null,
         RelCollations.EMPTY, type, name);
   }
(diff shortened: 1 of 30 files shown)


core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableMergeUnionRule.java
```diff
@@ -125,6 +125,7 @@ public EnumerableMergeUnionRule(Config config) {
           RexNode node = fields.get(i);
           if (fieldsRequiringCast.get(i)) {
             final RelDataType targetType = unionFieldList.get(i).getType();
+            // TODO: can this cast fail?
```
julianhyde, line 128: Don't add a TODO. Either do it or log a bug.

core/src/main/java/org/apache/calcite/rel/core/AggregateCall.java
```diff
@@ -129,58 +111,8 @@ private AggregateCall(SqlAggFunction aggFunction, boolean distinct,
...
-      boolean distinct, boolean approximate, boolean ignoreNulls,
-      List<Integer> argList, int filterArg,
-      @Nullable ImmutableBitSet distinctKeys, RelCollation collation,
-      int groupCount,
-      RelNode input, @Nullable RelDataType type, @Nullable String name) {
-    return create(aggFunction, distinct, approximate, ignoreNulls,
-        ImmutableList.of(), argList, filterArg,
-        distinctKeys, collation, groupCount, input, type, name);
-  }
-
```
julianhyde, line 181: Sorry, we can't remove deprecated stuff.

mihaibudiu: I have put all the deprecated functions back

mihaibudiu:
I really would like to merge this PR, but no one wants to review it.
Since it doesn't change any of the functionality of Calcite (only adds a new orthogonal feature) the risk is low.
But it is a bit tedious, I agree.

asolimando (approved):
LGTM, I have verified the changes, I have tried to look a bit for missed cases and I couldn't find any, but it's way harder to be sure on misses

mihaibudiu:
During development I verified that there are no missing cases by deleting the old functions without source position and re-enabling them only after I checked that the existing ones cover all cases.
I will rebase. The tests should pass.

mihaibudiu:
I have resolved the conflicts, rebased and optimistically squashed the commits too

asolimando:
> I have resolved the conflicts, rebased and optimistically squashed the commits too

As soon as the last tests finish successfully you can go ahead and merge. Thanks for your great work @mihaibudiu!

PR #3511, opened 2023-11-09
[CALCITE-6094] Linq4j.ConstantExpression.write crashes on special FP values

mihaibudiu:


linq4j/src/main/java/org/apache/calcite/linq4j/tree/ConstantExpression.java
```diff
@@ -118,7 +126,15 @@ private static ExpressionWriter write(ExpressionWriter writer,
         if (value instanceof BigDecimal) {
           bigDecimal = (BigDecimal) value;
         } else {
-          bigDecimal = BigDecimal.valueOf((Double) value);
+          double d = (Double) value;
+          if (d == Double.POSITIVE_INFINITY) {
```
tanclary, line 130: I see that you're checking for `Double` and `Float`. I'd like to avoid adding another level of nesting + adding two very similar logic blocks if possible. Do you think a helper could help fix this?

tanclary (commented):
Left one question

linq4j/src/main/java/org/apache/calcite/linq4j/tree/ConstantExpression.java
tanclary: It's fine to leave it I am just weary of adding similar nested `if` blocks. If you don't think there is a way to improve it that is ok.

mihaibudiu:
I will wait a couple more days for reviews, but otherwise I am planning to merge this PR.

PR #3513, opened 2023-11-09
[CALCITE-5607] Error when deserializing TIMESTAMP_DIFF in RelJson#toJson

olivrlee:
https://issues.apache.org/jira/browse/CALCITE-5607

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
+++ b/core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
@@ -622,6 +622,7 @@ public class RelJson {
         }
         map.put("operands", list);
         switch (node.getKind()) {
+        case MINUS:
         case CAST:
           map.put("type", toJson(node.getType()));
           break;
--- a/core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
+++ b/core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
@@ -786,7 +786,8 @@ class RelWriterTest {
       throw TestUtil.rethrow(e);
     }
     final RelJson relJson = RelJson.create()
-        .withInputTranslator(RelWriterTest::translateInput);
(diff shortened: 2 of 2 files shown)


core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
```diff
@@ -1052,6 +1053,41 @@ void testAggregateWithAlias(SqlExplainFormat format) {
     assertThat(result, isLinux(expected));
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5607">[CALCITE-5607]</a>
+   *
+   * <p>Before the fix, RelJson.toRex would throw an ArrayIndexOutOfBounds error
+   * when deserialization.
+   * */
+  @Test void testDeserializeMinusDateOperator() {
```
tjbanghart, line 1062: The commit message and test case don't seem to align. This checks that `MINUS` can be deserialized but I don't see anything to suggest we've changed how the return `type` is being serialized.

core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -622,6 +622,7 @@ public Object toJson(RexNode node) {
         }
         map.put("operands", list);
         switch (node.getKind()) {
+        case MINUS:
```
tjbanghart, line 625: We should change the Jira issue to reflect that we are only changing the behavior of `MINUS`.

core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
olivrlee: Yeah let me update the JIRA summary and the ticket title for this.
We're changing it to only affect MINUS

tjbanghart (approved):
LGTM Thanks!

PR #3517, opened 2023-11-11
Site: Troubleshooting/Website publishing improvements in release guide

libenchao:


site/_docs/howto.md
```diff
@@ -806,7 +806,7 @@ git clean -xn
 ./gradlew prepareVote -Prc=0
 
 # Push release candidate to ASF servers
-./gradlew prepareVote -Prc=0 -Pasf -Pasf.git.pushRepositoryProvider=GITBOX
```
libenchao, line 809: Throws following exception with `GITBOX`:
```
> Task :pushRcTag FAILED

Build calcite FAILURE reason:
    Execution failed for task ':pushRcTag':
        Caused by: org.eclipse.jgit.api.errors.TransportException: https://gitbox.apache.org/repos/asf/calcite.git: not authorized
            at org.eclipse.jgit.api.PushCommand.call(PushCommand.java:147)
            at com.github.vlsi.gradle.release.jgit.dsl.GitExtensionsKt.push(GitExtensions.kt:132)
            at com.github.vlsi.gradle.release.GitPushTask$pushTag$1.invoke(GitPushTask.kt:54)
            at com.github.vlsi.gradle.release.GitPushTask$pushTag$1.invoke(GitPushTask.kt:30)
            at com.github.vlsi.gradle.release.DefaultGitTask.jgit(DefaultGitTask.kt:45)
            at com.github.vlsi.gradle.release.GitPushTask.pushTag(GitPushTask.kt:50)
```

site/_docs/howto.md
```diff
@@ -927,12 +927,6 @@ thread to discuss.
 Julian
 {% endhighlight %}
 
-Use the [Apache URL shortener](https://s.apache.org) to generate
-shortened URLs for the vote proposal and result emails. Examples:
-[s.apache.org/calcite-1.2-vote](https://s.apache.org/calcite-1.2-vote) and
-[s.apache.org/calcite-1.2-result](https://s.apache.org/calcite-1.2-result).
```
libenchao, line 933: It seems there are no steps need to use this, I propose to remove it to make the document clean.

site/_docs/howto.md
F21: For avatica, I also have `-Pasf.git.pushRepositoryProvider=GITBOX` in the release script: https://github.com/apache/calcite-avatica/blob/main/docker.sh#L298

I haven't used this recently, so not sure if it still works.

site/_docs/howto.md
F21: +1, Since the release script generates the emails, I think most RMs will just copy that verbatim, so the shortening of links seem redundant.

site/_docs/howto.md
libenchao: CC @NobiGo @zabetak since you are the latest release managers.

site/_docs/howto.md
zabetak: My understanding is that these steps are for creating mnemonic URLs that people can easily find and refer to it in future. I don't think was ever meant to be used for the voting process it self but for later reference to it if needed. I don't have a strong opinion on keeping or removing them but they do serve a purpose. For instance, typing the following just works and brings me what I need:
* https://s.apache.org/calcite-1.34.0-vote
* https://s.apache.org/calcite-1.34.0-result

site/_docs/howto.md
zabetak: The `asf.git.pushRepositoryProvider` should be inline with `asfGitSourceUsername` and `asfGitSourcePassword`. If you are using `GITBOX` then you should use the ASF credentials; if you use `GITHUB` you should use the GitHub credentials. I assume both providers should work fine if the correct credentials are set.

I think there were some misconception that GITHUB was not working in the past (https://issues.apache.org/jira/browse/CALCITE-4856) but I suspect that the underlying reason was the same (mixing ASF vs. GitHub credentials).

If we want to improve something it makes sense to clarify somewhere this story about the different credentials and not just change again the default values.

libenchao: I was following this line to setup the environment: https://github.com/apache/calcite/blob/816edd1fa1a1d6af7e72416d791eb01d8c66b6ea/site/_docs/howto.md?plain=1#L686-L687

I thought this is the suggested way to interact with Git. (Actually I'm not sure how to setup Gitbox credentials, I guess `asfGitSourcePassword` is the apache password which is the same with `asfNexusPassword` and
`asfSvnPassword`?)

site/_docs/howto.md
libenchao: I also agree that it served for some purpose. The problem now is that it somethings confuses new RMs (this has happened during 1.35.0 releasing: https://lists.apache.org/thread/nv2xclvgrw84866o0q2ngdx4c2dg6pqq).

site/_docs/howto.md
zabetak: Yes, the Gitbox credentials is the usual username and password combination that is used to access most (if not all) ASF services.

site/_docs/howto.md
zabetak: If the explanation that I gave above clarifies a bit the situation then we could enrich the doc accordingly. If you feel that there might be still confusion then feel free to remove the section altogether.

site/_docs/howto.md
libenchao: I'll leave the default as 'GITBOX', and add more comments about it.

libenchao:
@zabetak @F21 I've addressed the comments above (kept the removing the shortening URL section since I think it's not a required step in current release process, removing it would make the document clean and concise), if there is no more remarks, I'll merge it later.

F21 (approved):

PR #3520, opened 2023-11-14
[CALCITE-6117] Converting SAFE_CAST from RexCall to SqlCall fails to add the type as an argument

wnob:
With the fix to `SqlImplementor.java`, the added test passes.

Without the fix, it triggers [this assertion](https://github.com/apache/calcite/blob/3843ede3d28783235780c0f81f725dc7e64a7828/core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java#L254) because the type of the rex function is not converted to an argument of the sql function (which thus only has 1 argument).

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -2251,6 +2251,15 @@ class RelToSqlConverterTest {
         .ok(expectedBqFormatDatetime);
   }
 
+  @Test void testBigQuerySafeCast() {
+    final String query = "select safe_cast(\"product_name\" as date) "
+        + "from \"foodmart\".\"product\"";
+    final String expected = "SELECT SAFE_CAST(\"product_name\" AS DATE)\n"
+        + "FROM \"foodmart\".\"product\"";
+
+    sql(query).withLibrary(SqlLibrary.BIG_QUERY).ok(expected);
+  }
+
   /** Test case for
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):

mihaibudiu:
I think that people prefer that the ticket name (and the issue name) state the problem rather than "fix".

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -2251,6 +2251,15 @@ private SqlDialect nonOrdinalDialect() {
         .ok(expectedBqFormatDatetime);
   }
 
+  @Test void testBigQuerySafeCast() {
```
mihaibudiu, line 2254: you should also add here a JavaDoc similar to the existing ones for other tests.

tanclary (commented):
I think this is mergeable after adding the JavaDoc @mihaibudiu mentioned, wdyt @mihaibudiu ?

JiajunBernoulli:
Druid Tests failed.
It should not be related to the change.

olivrlee:
Needs a reword to commit message but looks good, thanks Will

olivrlee (approved):

olivrlee:
> What kind of reword are you thinking of? I'm assuming the commits will be squashed on merge with an opportunity to reword, right? Or is the pull request title too long?

The title is good, just reminder to amend commit message to that when merging
I recently talked to Julian about commit messages in general and he suggests that the commit message describes the problem, so yeah the title currently satisfies that

PR #3521, opened 2023-11-14
[CALCITE-5508] Allow no-op versions of DATETIME and TIMESTAMP

wnob:
This is a bit of unfinished work from CALCITE-5508. BigQuery supports calling the `TIMESTAMP()` function with a `TIMESTAMP`-typed argument, and the `DATETIME()` function with a `DATETIME`-typed argument. Both functions should simply return their argument in these cases, which is functionality we get for free from the existing implementations.

The problem is that, without proper argument type checking, Calcite will try to coerce the argument types, which in the case of `DATETIME()` causes it to try casting as `DATE` and eliminating the time component.

### DIFF
--- a/babel/src/test/resources/sql/big-query.iq
+++ b/babel/src/test/resources/sql/big-query.iq
@@ -2708,6 +2708,18 @@ SELECT TIMESTAMP(DATE "2008-12-25") AS timestamp_date;
 
 !ok
 
+# This tests the no-op TIMESTAMP and DATETIME functions.
+select TIMESTAMP(TIMESTAMP "2008-01-01 01:03:05") as ts,
+       DATETIME(DATETIME "2008-01-01 01:03:05") as dt;
++---------------------+---------------------+
+| ts                  | dt                  |
++---------------------+---------------------+
+| 2008-01-01 01:03:05 | 2008-01-01 01:03:05 |
++---------------------+---------------------+
+(1 row)
+
+!ok
+
 # All these timestamps should be equal.
 # This tests the BQ timestamp literal string formatter
 # (optional 'T', optional leading zeros, optional offset with conversion).
--- a/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
@@ -5055,4 +5055,24 @@ class SqlToRelConverterTest extends SqlToRelTestBase {
     String sql = "SELECT CAST(CAST(? AS INTEGER) AS CHAR)";
     sql(sql).ok();
   }
+
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5508">DATETIME and TIMESTAMP constructor
+   * functions</a> that take single arguments with the same type as the function's return type.
+   *
+   * <p>The function counts as a no-op in this case.
+   */
+  @Test void testNoOpTimeDateFunctions() {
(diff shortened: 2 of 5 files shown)


core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -5055,4 +5055,24 @@ void checkUserDefinedOrderByOver(NullCollation nullCollation) {
     String sql = "SELECT CAST(CAST(? AS INTEGER) AS CHAR)";
     sql(sql).ok();
   }
+
```
mihaibudiu, line 5058: there is no documentation added in this PR for these behaviors.
Also, it would be nice to have some negative tests, where type inference rejects the query, and with NULL arguments.

wnob: I'll try to come up with a negative test shortly. In the meantime, I've added documentation for the new overloads, and made some minor corrections to the existing documentation so they more accurately reflect how BigQuery actually handles these functions.

The no-op versions are undocumented by BigQuery, but easily verifiable as working with a simple example query: `SELECT DATETIME(DATETIME '2022-02-02 12:12:12')`

babel/src/test/resources/sql/big-query.iq
```diff
@@ -2708,6 +2708,18 @@ SELECT TIMESTAMP(DATE "2008-12-25") AS timestamp_date;
 
 !ok
 
+# This tests the no-op TIMESTAMP and DATETIME functions.
+select TIMESTAMP(TIMESTAMP "2008-01-01 01:03:05") as ts,
```
tanclary, line 2712: Maybe using `!type` would be helpful here since the aim is to show they return the same type as they are given.

site/_docs/reference.md
```diff
@@ -2833,12 +2835,13 @@ BigQuery's type system uses confusingly different names for types and functions:
 | b | TIME(timestamp)                                | Extracts the TIME from *timestamp* (a local time; BigQuery's DATETIME type)
 | b | TIME(instant)                                  | Extracts the TIME from *timestampLtz* (an instant; BigQuery's TIMESTAMP type), assuming UTC
 | b | TIME(instant, timeZone)                        | Extracts the time from *timestampLtz* (an instant; BigQuery's TIMESTAMP type), in *timeZone*
+| b | TIMESTAMP(timestampLtz)                        | Returns its argument unchanged (no-op)
```
tanclary, line 2838: would `returns *timestampLtz* unchanged (no-op)` fit better with other descriptions?

site/_docs/reference.md
```diff
@@ -2711,9 +2711,11 @@ BigQuery's type system uses confusingly different names for types and functions:
 | p q | DATEADD(timeUnit, integer, datetime)         | Equivalent to `TIMESTAMPADD(timeUnit, integer, datetime)`
 | p q | DATEDIFF(timeUnit, datetime, datetime2)      | Equivalent to `TIMESTAMPDIFF(timeUnit, datetime, datetime2)`
 | q | DATEPART(timeUnit, datetime)                   | Equivalent to `EXTRACT(timeUnit FROM  datetime)`
+| b | DATETIME(timestamp)                            | Returns its argument unchanged (no-op)
```
tanclary, line 2714: same comment as below

tanclary (changes requested):
Left a couple of suggestions

babel/src/test/resources/sql/big-query.iq
tanclary: or use `!ok` and `!type` in conjunction somehow

PR #3522, opened 2023-11-14
[CALCITE-6111] Explicit cast from expression to numeric type doesn't check overflow

mihaibudiu:
This also fixes a few more tests which were disabled in SqlOperatorTests.
I still have to handle casts to decimal values, but hopefully this fixes the behavior of casts to integer types.
This behavior also has to be documented, so a future PR will describe in more details how Calcite evaluates casts.

### DIFF
--- a/linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
+++ b/linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
@@ -384,6 +384,10 @@ public enum Primitive {
     }
   }
 
+  public static @Nullable Object integerCast(Primitive primitive, final Object value) {
+    return requireNonNull(primitive, "primitive").numberValue((Number) value);
+  }
+
   /**
    * Converts a number into a value of the type specified by this primitive
    * using the SQL CAST rules.  If the value conversion causes loss of significant digits,
@@ -419,6 +423,13 @@ public enum Primitive {
         // The value Long.MAX_VALUE cannot be represented exactly as a double,
         // so we cannot use checkRoundedRange.
         BigDecimal decimal = BigDecimal.valueOf(value.doubleValue())
+            // Round to an integer
+            .setScale(0, RoundingMode.DOWN);
+        // longValueExact will throw ArithmeticException if out of range
+        return decimal.longValueExact();
+      }
+      if (value instanceof BigDecimal) {
+        BigDecimal decimal = ((BigDecimal) value)
             // Round to an integer
             .setScale(0, RoundingMode.DOWN);
         // longValueExact will throw ArithmeticException if out of range
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -133,6 +133,7 @@ import static org.apache.calcite.sql.test.SqlOperatorFixture.INVALID_EXTRACT_UNI
 import static org.apache.calcite.sql.test.SqlOperatorFixture.INVALID_EXTRACT_UNIT_VALIDATION_ERROR;
 import static org.apache.calcite.sql.test.SqlOperatorFixture.LITERAL_OUT_OF_RANGE_MESSAGE;
 import static org.apache.calcite.sql.test.SqlOperatorFixture.OUT_OF_RANGE_MESSAGE;
+import static org.apache.calcite.sql.test.SqlOperatorFixture.WRONG_FORMAT_MESSAGE;
 import static org.apache.calcite.util.DateTimeStringUtils.getDateFormatter;
 
 import static org.hamcrest.CoreMatchers.equalTo;
@@ -623,13 +624,16 @@ public class SqlOperatorTest {
 
       // Overflow test
       if (numeric == Numeric.BIGINT) {
-        // Literal of range
+        // Calcite cannot even represent a literal so large, so
+        // for this query even the safe casts fail at compile-time
+        // (runtime == false).
         f.checkCastFails(numeric.maxOverflowNumericString,
             type, LITERAL_OUT_OF_RANGE_MESSAGE, false, castType);
         f.checkCastFails(numeric.minOverflowNumericString,
             type, LITERAL_OUT_OF_RANGE_MESSAGE, false, castType);
       } else {
-        if (numeric != Numeric.DECIMAL5_2 || Bug.CALCITE_2539_FIXED) {
+        if (numeric != Numeric.DECIMAL5_2) {
+          // This condition is for bug [CALCITE-6078], not yet fixed
           f.checkCastFails(numeric.maxOverflowNumericString,
               type, OUT_OF_RANGE_MESSAGE, true, castType);
           f.checkCastFails(numeric.minOverflowNumericString,
@@ -643,11 +647,12 @@ public class SqlOperatorTest {
       f.checkCastToScalarOkay("'" + numeric.minNumericString + "'",
           type, numeric.minNumericString, castType);
 
-      if (Bug.CALCITE_2539_FIXED) {
+      if (numeric != Numeric.DECIMAL5_2) {
+        // The above condition is for bug CALCITE-6078
         f.checkCastFails("'" + numeric.maxOverflowNumericString + "'",
-            type, OUT_OF_RANGE_MESSAGE, true, castType);
+            type, WRONG_FORMAT_MESSAGE, true, castType);
         f.checkCastFails("'" + numeric.minOverflowNumericString + "'",
-            type, OUT_OF_RANGE_MESSAGE, true, castType);
+            type, WRONG_FORMAT_MESSAGE, true, castType);
       }
 
       // Convert from type to string
@@ -657,10 +662,8 @@ public class SqlOperatorTest {
       f.checkCastToString(numeric.minNumericString, null, null, castType);
       f.checkCastToString(numeric.minNumericString, type, null, castType);
 
-      if (Bug.CALCITE_2539_FIXED) {
-        f.checkCastFails("'notnumeric'", type, INVALID_CHAR_MESSAGE, true,
-            castType);
-      }
+      f.checkCastFails("'notnumeric'", type, INVALID_CHAR_MESSAGE, true,
+          castType);
     });
   }
 
@@ -1128,14 +1131,14 @@ public class SqlOperatorTest {
     // ExceptionInInitializerError.
     f.checkScalarExact("cast('15' as integer)", "INTEGER NOT NULL", "15");
     if (castType == CastType.CAST) { // Safe casts should not fail
-      f.checkFails("cast('15.4' as integer)", "Number has wrong format.*", true);
-      f.checkFails("cast('15.6' as integer)", "Number has wrong format.*", true);
+      f.checkFails("cast('15.4' as integer)", WRONG_FORMAT_MESSAGE, true);
+      f.checkFails("cast('15.6' as integer)", WRONG_FORMAT_MESSAGE, true);
       f.checkFails("cast('ue' as boolean)", "Invalid character for cast.*", true);
       f.checkFails("cast('' as boolean)", "Invalid character for cast.*", true);
-      f.checkFails("cast('' as integer)", "Number has wrong format.*", true);
-      f.checkFails("cast('' as real)", "Number has wrong format.*", true);
-      f.checkFails("cast('' as double)", "Number has wrong format.*", true);
-      f.checkFails("cast('' as smallint)", "Number has wrong format.*", true);
+      f.checkFails("cast('' as integer)", WRONG_FORMAT_MESSAGE, true);
+      f.checkFails("cast('' as real)", WRONG_FORMAT_MESSAGE, true);
+      f.checkFails("cast('' as double)", WRONG_FORMAT_MESSAGE, true);
+      f.checkFails("cast('' as smallint)", WRONG_FORMAT_MESSAGE, true);
     } else {
       f.checkNull("cast('15.4' as integer)");
       f.checkNull("cast('15.6' as integer)");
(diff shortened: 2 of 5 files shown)


mihaibudiu:
The CI failure is from the Druid tests, so I guess this PR is ready for review.

zstan:
I already realize the custom logic for such a cases, you can take only tests, check:
https://github.com/apache/ignite-3/pull/2786/files#diff-4ddeb0625d1cc7604c3e33e3d57ca209307a8f3e5b6aea01bee459ce06963870R675
https://github.com/apache/ignite-3/pull/2786/files#diff-ef8a24b20a4922cd2b5e92e466e3a12263444ec6887d0d87d8c885359033cf76R426

mihaibudiu:
These tests are very useful. They cover many more cases than this PR handles. Here we only do casts.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13412,6 +13414,34 @@ private static void checkLogicalOrFunc(SqlOperatorFixture f) {
...
+    f.checkFails("SELECT cast(10000000000+30 as int)",
+        ".* out of range", true);
+
+    f.checkFails("SELECT cast(1e60+30 as tinyint)",
+        ".* out of range", true);
+    f.checkFails("SELECT cast(1e60+30 as smallint)",
+        ".* out of range", true);
+    f.checkFails("SELECT cast(1e60+30 as int)",
+        ".* out of range", true);
+    f.checkFails("SELECT cast(1e60+30 as bigint)",
```
zstan, line 13436: what about: SELECT CAST(2147483648 as BIGINT) ?

mihaibudiu: These cases should be covered by the test `testCastExactNumericLimits`, where I re-enabled some of the tests that were disabled.

mihaibudiu: Well, this cast in particular should not fail, so I guess it's not covered by the NumericLimits test.
I have checked that this test would pass. 
Are you suggesting adding tests where I use the numeric limits for a type (INT) when casting to a different type (BIGINT)?

mihaibudiu:
This is a bugfix PR, can someone please review it?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13412,6 +13414,34 @@ private static void checkLogicalOrFunc(SqlOperatorFixture f) {
     }
   }
 
+  /**
+   * Test cases for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6111">[CALCITE-6111]
+   * Explicit cast from expression to numeric type doesn't check overflow</a>. */
+  @Test public void testOverflow() {
+    final SqlOperatorFixture f = fixture();
+    f.checkFails("SELECT cast(100+30 as tinyint)",
```
zstan, line 13423: as for me - it`s hard to parse such a tests.
```suggestion
    f.checkFails(String.format("SELECT cast(%d+30 as tinyint)", Byte.MAX_VALUE),
```
and the same for all near

zstan: what about implicit casts, is it all ok there ? or it out of scope ? 
SELECT 9223372036854775807 + 1;
SELECT 2147483647 + 1;

mihaibudiu: This PR is only about casts. I plan a follow up PR about arithmetic, which has been broken for many years. There are many disabled tests because of this. That PR will handle the expressions you describe.

Regarding implicit casts: in a normal compiler, after validation the compiler should convert all implicit casts into explicit casts. This leaves many fewer cases to worry about for the subsequent stages. Unfortunately it doesn't look like Calcite works this way. I don't know enough about validation to promise I will fix this part.

However, implicit casts usually convert values from a narrow type to a wider type, so they should not be a cause of runtime failures.

But only writing lots of tests will ensure that all cases are properly handled.

mihaibudiu:
@zstan I have rewritten the tests you suggested to make the intent clearer.
Thank you for the review, please let me know if you have more suggestions.
Note that this PR does not intend to handle casts to decimal types, that will be a separate PR.
It does not handle arithmetic overflow either, that will be yet another PR.

mihaibudiu:
@zstan, I would like to merge this PR, since the fixes I plan to address depend on these being merged. 
Please let me know if you have more comments. I will interpret silence as an approval if you don't mind.

zstan:
@mihaibudiu honestly, overall looks good but as for me this issue need additional eyes here, can you found one more reviewer on dev list ?

asolimando (commented):
LGTM, few comments but nothing major

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13695,6 +13697,43 @@ private static void checkLogicalOrFunc(SqlOperatorFixture f) {
...
   }
 
+  /**
+   * Test cases for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6111">[CALCITE-6111]
+   * Explicit cast from expression to numeric type doesn't check overflow</a>. */
+  @Test public void testOverflow() {
+    final SqlOperatorFixture f = fixture();
+    f.checkFails(String.format(Locale.US, "SELECT cast(%d+30 as tinyint)", Byte.MAX_VALUE),
+        ".* out of range", true);
```
snuyanzin, line 13707: I tried to see what happens if I locally change expectedError to something else...
Surprisingly the tests continue passing
it seems that the reason is that this method is called https://github.com/apache/calcite/blob/0bec957071468a2e54a22519290ac101a752fcad/testkit/src/main/java/org/apache/calcite/test/SqlOperatorFixtureImpl.java#L157-L171

which means it checks for failure only if it is not runtime...
so it could be any expected message and any number under `cast` and it continues passing...

Or did I do something wrong?

asolimando: @snuyanzin, I think it should fail as you say, there seems to be something wrong.

I also tried changing `runtime` to `false` just to see, in this way it always fails (with a cast with overflow, and cast without), I guess there is something broken in the test auxiliary methods we rely on here.

The source of the problem lies outside the present PR, but it's nonetheless a blocker for merging (if confirmed), as we can't rely on the unit tests.

Maybe @mihaibudiu has more familiarity with this part of the code, and can shed some light.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
```diff
@@ -384,6 +384,10 @@ static void checkRoundedRange(Number value, double min, double max) {
     }
   }
 
+  public static @Nullable Object integerCast(Primitive primitive, final Object value) {
```
asolimando, line 387: Q: Is this used via reflection? My IDE suggests that the method is unused.

asolimando (changes requested):
After @snuyanzin's comment I feel we need to get to the bottom of the test issue before merging, therefore marking the PR as "request changes" until clarified

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
mihaibudiu: Not quite reflection. The way Calcite evaluates constant expressions is to generate a Java class with a Java method which is then compiled, loaded and executed at runtime.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -657,10 +661,8 @@ void testCastExactNumericLimits(CastType castType, SqlOperatorFixture f) {
       f.checkCastToString(numeric.minNumericString, null, null, castType);
       f.checkCastToString(numeric.minNumericString, type, null, castType);
 
-      if (Bug.CALCITE_2539_FIXED) {
-        f.checkCastFails("'notnumeric'", type, INVALID_CHAR_MESSAGE, true,
-            castType);
-      }
+      f.checkCastFails("'notnumeric'", type, INVALID_CHAR_MESSAGE, true,
```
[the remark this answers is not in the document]
mihaibudiu: Yes, CALCITE-2539 seems to cover multiple issues, and I have fixed only one of them here. There are other disabled tests guarded by this Bug.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: If you call the test from SqlOperatorTest it doesn't involve the evaluation.
You have to call the test from CalciteSqlOperatorTest. That class extends SqlOperatorTest but uses a test fixture which also evaluates the expressions. In the IDE you can do this easiest if you paste the following in CalciteSqlOperatorTest:

```
  @Test
  public void testOverflow() {
    super.testOverflow();
  }
```

mihaibudiu:
> After @snuyanzin's comment I feel we need to get to the bottom of the test issue before merging, therefore marking the PR as "request changes" until clarified

I have made the fixes you suggested. The tests should be run from the class CalciteSqlOperatorTest to exercise the evaluator.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
asolimando: Thanks for the refresher Mihai, I always forget about this behaviour in some of our test classes.

I have verified the tests, they correctly fail if a non-overflowing value/expression is passed, so they do their job.

asolimando:
> > After @snuyanzin's comment I feel we need to get to the bottom of the test issue before merging, therefore marking the PR as "request changes" until clarified
> 
> I have made the fixes you suggested. The tests should be run from the class CalciteSqlOperatorTest to exercise the evaluator.

Thanks Mihai, the changes and the patch LGTM, can you please address the Sonar violation regarding the repeated occurrence of the string `".* out of range"`? (just define it once at the beginning of the test for better maintainability in case the error messages changes).

PS: please don't force-push while the review process is still on-going, it makes reviewers' life more difficult for not much gain, thanks!

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
snuyanzin: Was able to test, thanks for clarification

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13695,6 +13698,43 @@ private static void checkLogicalOrFunc(SqlOperatorFixture f) {
...
+    f.checkFails(String.format(Locale.US, "SELECT cast(%d+30 as smallint)", Short.MAX_VALUE),
+        ".* out of range", true);
+    // We use a long value because otherwise calcite interprets the literal as an integer.
+    f.checkFails(String.format(Locale.US, "SELECT cast(%d as int)", Long.MAX_VALUE),
+        ".* out of range", true);
+
+    // Casting a floating point value larger than the maximum allowed value.
+    // 1e60 is larger than the largest BIGINT value allowed.
+    f.checkFails("SELECT cast(1e60+30 as tinyint)",
+        ".* out of range", true);
```
snuyanzin, line 13718: nit:
since there is already existing `org.apache.calcite.sql.test.SqlOperatorFixture#OUT_OF_RANGE_MESSAGE` would it make sense to reuse it here?

mihaibudiu: My last commit does exactly that. It removes one sonarcloud complaint, as pointed out by @asolimando. 

The other sonarcloud complaint is about a complex method which I didn't really write, but I slightly modified, so I didn't try to reduce the complexity. This is after all just a test method.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -643,11 +647,12 @@ void testCastExactNumericLimits(CastType castType, SqlOperatorFixture f) {
       f.checkCastToScalarOkay("'" + numeric.minNumericString + "'",
           type, numeric.minNumericString, castType);
 
-      if (Bug.CALCITE_2539_FIXED) {
+      if (numeric != Numeric.DECIMAL5_2) {
+        // The above condition is for bug CALCITE-6078
         f.checkCastFails("'" + numeric.maxOverflowNumericString + "'",
-            type, OUT_OF_RANGE_MESSAGE, true, castType);
+            type, WRONG_FORMAT_MESSAGE, true, castType);
```
asolimando, line 653: From the variable name, it seems that `OUT_OF_RANGE_MESSAGE` was the correct error we expected (similarly to the checks in the `if` statement starting at line 635)?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
asolimando: Yes, Sonar has false positives sometimes, it's fine to address only what's reasonable (I consider fixing the complexity of an existing test method as as nice-to-have and totally optional), but most of the time it provides good hints for keeping the code quality high.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: Actually this variant of the test invokes something like `Integer.parseInt(string)`. So this is in fact a cast from a string to integer -- notice the single quotes added around the literal -- which gives a different error than a cast from a larger integer to an integer. The former produces a WRONG_FORMAT, the latter gives OUT_OF_RANGE. We could in principle trap the exception and rewrite the message, but it *is* a different exception in Java in the simplification code.

mihaibudiu: The integer-to-integer cast is tested a few lines above.

asolimando: You are right, I missed the extra quote, sorry about that!

asolimando (approved):
LGTM, thank you, Mihai!

PR #3523, opened 2023-11-14
Jitender.yadav/rsfb 2016 handling of dateadd with current timestamp

jitender-datametica:
Handling the CURRENT_TIMESTAMP() in DATEADD


jitender-datametica:
wrong place

PR #3524, opened 2023-11-15
[CALCITE-6118] Missing empty ARRAY function usage in reference doc

chucheng92:
https://issues.apache.org/jira/browse/CALCITE-6118

### DIFF
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2654,7 +2654,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | p | expr :: type                                   | Casts *expr* to *type*
 | m | expr1 <=> expr2                                | Whether two values are equal, treating null values as the same, and it's similar to `IS NOT DISTINCT FROM`
 | * | ACOSH(numeric)                                 | Returns the inverse hyperbolic cosine of *numeric*
-| s | ARRAY(expr [, expr ]*)                         | Construct an array in Apache Spark
+| s | ARRAY(expr [, expr ]*)                         | Construct an array in Apache Spark. The function allows users to use `ARRAY()` to create an empty array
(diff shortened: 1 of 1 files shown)


site/_docs/reference.md
```diff
@@ -2654,6 +2654,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | p | expr :: type                                   | Casts *expr* to *type*
 | m | expr1 <=> expr2                                | Whether two values are equal, treating null values as the same, and it's similar to `IS NOT DISTINCT FROM`
 | * | ACOSH(numeric)                                 | Returns the inverse hyperbolic cosine of *numeric*
+| s | ARRAY()                                        | Construct an empty array in Apache Spark
```
mihaibudiu, line 2657: Couldn't this and the next line be unified?

mihaibudiu (approved):

site/_docs/reference.md
chucheng92: thanks for reviewing. IMHO, The empty array is special, and it will be highlighted here to make it clearer. Of course, we can merge it in the next line, but I may not want to modify the parameter form of `ARRAY(expr [, expr ]*)` and only add a description of an empty array. what do you think?

mihaibudiu: I don't see any other case where the same function has two rows.

JiajunBernoulli:
@mihaibudiu , Why approve it ?

It seems that the two rows are not very suitable.

mihaibudiu:
> @mihaibudiu , Why approve it ?
>
> It seems that the two rows are not very suitable.

The spirit of the changes is right, but the form can be improved. Once the author has improved the presentation, if they have authority, hopefully they can merge without re-requesting another review from me. There is a certain amount of trust implied. The goal is to reduce the number of messages exchanged about this small PR.

site/_docs/reference.md
chucheng92: > I don't see any other case where the same function has two rows.

make sense, just a small improvement. merged.

chucheng92:
thanks for reviewing. merging...

PR #3527, opened 2023-11-15
[CALCITE-6095] Arithmetic expression with VARBINARY value causes AssertionFailure

mihaibudiu:


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1465,6 +1465,26 @@ void testCastToBoolean(CastType castType, SqlOperatorFixture f) {
     f.checkNull("cast(null as row(f0 varchar, f1 varchar))");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6095">
+   * [CALCITE-6095] Arithmetic expression with VARBINARY value causes AssertionFailure</a>.
+   */
+  @Test public void testVarbitArithmetic() {
+    SqlOperatorFixture f = fixture();
```
tanclary, line 1473: Would this belong better in `TypeCoercionTest`? I get that `CAST` is an operator but maybe that class is more tailored to the purpose of this ticket. Let me know what you think.

mihaibudiu: I actually see that TypeCoercionTest has negative tests for `binary` in the method `testImplicitCasts`. So these tests are not sufficient. 
The problem is in the type inference for the binary operator. The only negative tests in this file are using `shouldNotCast`, which does not seem to exercise the inference for binary operators.

There is an additional benefit having operator tests: we are also checking the error message. But that is less important.

JiajunBernoulli:
LGTM.

Clear error message is an improvement compared to AssertionFailure.

mihaibudiu:
What is the conclusion with respect to the Druid tests?
Can I merge this PR even though the CI has failed for these?

PR #3531, opened 2023-11-17
[draft][CALCITE-5087] Support MySQL bitwise operators in Babel Parser

macroguo-ghy:


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2173,6 +2175,80 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
+          38,
+          OperandTypes.ANY_ANY);
+
+  /** The Bitwise inversion "~" OPERATOR used by MySQL, for example
+   * {@code ~1}. */
+  @LibraryOperator(libraries = {MYSQL})
+  public static final SqlPrefixOperator BITWISE_INVERSION =
+      new SqlPrefixOperator(
+          "~",
+          SqlKind.BITWISE_RIGHT_SHIFT,
```
mihaibudiu, line 2234: is this a typo?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2173,6 +2175,80 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
+          80,
+          ReturnTypes.BIGINT_NULLABLE,
```
mihaibudiu, line 2236: why BIGINT and not the type of the operand?

gradle.properties
```diff
@@ -89,14 +89,14 @@ cassandra-all.version=4.0.1
 cassandra-java-driver-core.version=4.13.0
 cassandra-unit.version=4.3.1.0
 chinook-data-hsqldb.version=0.2
-commons-codec.version=1.16.0
-commons-dbcp2.version=2.11.0
-commons-io.version=2.15.0
-commons-lang3.version=3.13.0
+commons-codec.version=1.13
```
mihaibudiu, line 92: These changes are unrelated to this PR, I don't think they should be here.
Are you missing a rebase?

site/_docs/reference.md
```diff
@@ -1248,18 +1248,23 @@ The operator precedence and associativity, highest to lowest.
...
 | BETWEEN, IN, LIKE, SIMILAR, OVERLAPS, CONTAINS etc. | -
 | < > = <= >= <> != <=>                             | left
 | IS NULL, IS FALSE, IS NOT TRUE etc.               | -
 | NOT                                               | right
 | AND                                               | left
 | OR                                                | left
 
-Note that `::`,`<=>` is dialect-specific, but is shown in this table for
-completeness.
+Note that `::`,`!`,`~`,`^`,`<<`,`>>`,`&`,`|`,`<=>` is dialect-specific,
```
mihaibudiu, line 1266: are dialect specific.

core/src/main/java/org/apache/calcite/sql/SqlBitwiseBinaryOperator.java
```diff
@@ -0,0 +1,77 @@
...
+      String name,
+      SqlKind kind,
+      int prec,
+      @Nullable SqlOperandTypeChecker operandTypeChecker) {
+    super(
+        name,
+        kind,
+        prec,
+        true,
+        ReturnTypes.BIGINT_NULLABLE,
```
mihaibudiu, line 47: why bigint?
I think it should be the type of the operands - which should be coerced to have the same type.
Please note that some dialects such as MySQL allow some of these operations on BINARY types as well.
It's fine to start with just integers for now, but the solution should extend to BINARY eventually.

babel/src/test/java/org/apache/calcite/test/BabelTest.java
```diff
@@ -239,6 +239,47 @@ private void checkInfixCast(Statement statement, String typeName, int sqlType)
...
+
+    f.withSql("select 1 | 1")
+        .ok()
+        .type("RecordType(BIGINT NOT NULL EXPR$0) NOT NULL");
+
+    f.withSql("select x'65' & x'77'")
+        .ok()
+        .type("RecordType(BIGINT NOT NULL EXPR$0) NOT NULL");
+
+    f.withSql("select x'65' ^^ 1")
```
mihaibudiu, line 258: This seems wrong, the output type should be BINARY or VARBINARY.
Or perhaps shifts of BINARY values should be forbidden.

babel/src/test/java/org/apache/calcite/test/BabelTest.java
```diff
@@ -239,6 +239,47 @@ private void checkInfixCast(Statement statement, String typeName, int sqlType)
...
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-5087">[CALCITE-5087]
+   * Support bitwise operators</a>. */
+  @Test void testBitwiseOperator() {
+    final SqlValidatorFixture f = Fixtures.forValidator()
+        .withParserConfig(c -> c.withParserFactory(SqlBabelParserImpl.FACTORY))
+        .withOperatorTable(operatorTableFor(SqlLibrary.MYSQL));
+
+    f.withSql("select 1 | 1")
```
mihaibudiu, line 250: As commented below, BIGINT seems like the wrong type.
I would also add some tests that require implicit casts, like operations between int and tinyint values, to make sure that the type coercion rules work correctly.

core/src/main/java/org/apache/calcite/sql/SqlBitwiseBinaryOperator.java
mihaibudiu: BTW: Calcite already has aggregation functions BIT_AND and so on, and these return the same type as the input type. So the typing for these operators should be consistent with the aggregation functions.

mihaibudiu (changes requested):
This is great, this feature has been proposed a long time ago. I think there are quite a few issues related.

gradle.properties
JiajunBernoulli: Because he want test `Revert "Bump 6 dependencies"`.

Now, we know druid test failed is not about it.

So you can drop  the revert commit.

macroguo-ghy:
Thanks for your review, @mihaibudiu and @JiajunBernoulli, but this PR is not finished.
Many SQL engines support bitwise operators, but they have different usage and operand limits. In this PR, I only considered bitwise operators in MySQL, which I believe is insufficient. Therefore, I will close this PR.

PR #3536, opened 2023-11-19
[MINOR] Update javadoc for RexChecker with Litmus argument

yeahthisisrob:
Update the javadoc for `RexChecker` to reflect changes from https://issues.apache.org/jira/browse/CALCITE-955

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexChecker.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexChecker.java
@@ -31,28 +31,29 @@ import static java.util.Objects.requireNonNull;
 /**
  * Visitor which checks the validity of a {@link RexNode} expression.
  *
- * <p>There are two modes of operation:
- *
+ * <p>This visitor can operate in two modes, controlled by the {@link Litmus} argument:
  * <ul>
- * <li>Use<code>fail=true</code> to throw an {@link AssertionError} as soon as
- * an invalid node is detected:
- *
- * <blockquote><code>RexNode node;<br>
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/calcite/rex/RexChecker.java
```diff
@@ -31,31 +31,24 @@
...
+ * <li>{@link Litmus#IGNORE} tests validity without throwing an error.</li>
  * </ul>
  *
+ * <p>Example usage:
+ * <blockquote><pre>
+ * RexNode node;
+ * RelDataType rowType;
+ * RexChecker checker = new RexChecker(rowType, Litmus.IGNORE);
+ * node.accept(checker);
+ * if (!checker.isValid()) {
```
mihaibudiu, line 47: The original javadoc had two examples, whereas here there is only one.
I do find that the assert-based example is useful.

julianhyde:
Remove 'Fix:' from the commit message. That's not a prefix we use in this project.

core/src/main/java/org/apache/calcite/rex/RexChecker.java
yeahthisisrob: Thanks, I completely missed the first example. I recently used `IGNORE` and was focused there. Updated!

mihaibudiu (approved):

JiajunBernoulli:
@yeahthisisrob , Would you please add `[MINOR]` to commit message？

PR #2452, opened 2021-07-05
[CALCITE-4677] Fix different physical conventions conversion on new physical node producing

chenzl25:
…roducing

PR #3541, opened 2023-11-23
[CALCITE-6130] Transformation optimization of different physical conventions on the production of new physical nodes

caicancai:
[CALCITE-6130] Transformation optimization of different physical conventions on the production of new physical nodes


JiajunBernoulli:
@caicancai 
May we have a unit tests? Even if it's not SQL, RelNode is ok.

caicancai:
> @caicancai May we have a unit tests? Even if it's not SQL, RelNode is ok.

ok, thank you for your reply. Let me think about the best way to display test

caicancai:
> @caicancai May we have a unit tests? Even if it's not SQL, RelNode is ok.

ok, thank you for your reply. Let me think about the best way to display test

caicancai:
hi, I found that there is a pr that is similar to this pr, do I need to close this pr

caicancai:
https://github.com/apache/calcite/pull/2452

PR #3545, opened 2023-11-25
[CALCITE-6132] The exception thrown in the RelBuilderExample is not accurate

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6132

### DIFF
--- a/core/src/test/java/org/apache/calcite/examples/RelBuilderExample.java
+++ b/core/src/test/java/org/apache/calcite/examples/RelBuilderExample.java
@@ -66,7 +66,7 @@ public class RelBuilderExample {
     case 4:
       return example4(builder);
     default:
-      throw new AssertionError("unknown example " + i);
+      throw new AssertionError("unknown example" + i);
     }
   }
 


core/src/test/java/org/apache/calcite/examples/RelBuilderExample.java
```diff
@@ -66,7 +66,7 @@ private RelBuilder doExample(RelBuilder builder, int i) {
     case 4:
       return example4(builder);
     default:
-      throw new AssertionError("unknown example " + i);
+      throw new AssertionError("unknown example" + i);
```
mihaibudiu, line 69: why is this more accurate?

caicancai: Because it's the same way it was written, like i is 5 for example5, but test is all example4, and example5 has an extra space, so I've removed the space and turned it into example5

caicancai:
I will create a case without jira
https://issues.apache.org/jira/browse/CALCITE-6132

PR #3546, opened 2023-11-26
[MINOR] update message for RelBuilderExample

caicancai:


caicancai:
Because it's the same way it was written, like i is 5 for example 5, but test is all example4, and example 5 has an extra space, so I've removed the space and turned it into example5

PR #3548, opened 2023-11-26
[CALCITE-6133] Improve testBox tests for FilteratorTest

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6133

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/FilteratorTest.java
+++ b/core/src/test/java/org/apache/calcite/test/FilteratorTest.java
@@ -28,6 +28,8 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


julianhyde:
Doesn't seem like an improvement to me. The preceding code has built `result` but you no longer check it.

Since you call two different `filter` functions it would be possible for one of them to no-op and for the test to still pass.

Also, new test code should use `assertThat` rather than `assertEquals` or `assertNull`.

caicancai:
> Doesn't seem like an improvement to me. The preceding code has built `result` but you no longer check it.
> 
> Since you call two different `filter` functions it would be possible for one of them to no-op and for the test to still pass.
> 
> Also, new test code should use `assertThat` rather than `assertEquals` or `assertNull`.

If it  just check the result without using Filterator to check, I think this part can be improved

PR #3550, opened 2023-11-26
[MINOR] update results are accurate

caicancai:


caicancai:
I have a question about whether we should add a test to throw an exception for the 10/0 error test

PR #3556, opened 2023-12-01
[CALCITE-6149] Unparse for CAST Nullable with ClickHouseSqlDialect

Chen768959:
In ClickHouse, there is an issue with RelDataType of Nullable type when casting to a non-Nullable type. For example, `SELECT CAST(NULL, 'Int32')` will throw an error. In such cases, we should use `SELECT CAST(NULL, 'Nullable(Int32)')` instead.

```
CREATE TABLE default.test
(
    `id` Int64,
    `a` Nullable(Int64)
)
ENGINE = MergeTree
ORDER BY id
```
`insert into test(id,a) values(1,1);`
`insert into test(id,a) values(2,null);`
`SELECT CAST(a, 'String') FROM test;`

DB::Exception: Cannot convert NULL value to non-Nullable type
image

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -7120,6 +7120,21 @@ class RelToSqlConverterTest {
         .withBigQuery().ok(expectedBiqquery);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6149">[CALCITE-6149]
+   * Unparse for CAST Nullable with ClickHouseSqlDialect</a>. */
+  @Test void testCastToNullableInClickhouse() {
+    final String query = ""
+        + "SELECT CASE WHEN \"product_id\" IS NULL "
+        + "THEN CAST(\"product_id\" AS TINYINT) END, CAST(\"product_id\" AS TINYINT)\n"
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7453,6 +7453,18 @@ private void checkLiteral2(String expression, String expected) {
     sql(query).withSpark().withLibrary(SqlLibrary.SPARK).ok(expectedSql);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6149">[CALCITE-6149]
+   * unparse for CAST Nullable with ClickHouseSqlDialect</a>. */
+  @Test void testClickhouseNullableFieldCast() {
+    final String query = "SELECT CASE WHEN \"product_name\" IS NULL THEN \"product_name\"+ 1 END\n"
+        + "FROM \"foodmart\".\"product\"";
+    final String expectedSql = "SELECT CAST(NULL AS `Nullable(Int32)`)\n"
```
JiajunBernoulli, line 7462: Do you have not-null case?

mihaibudiu:
If this is ready for merging can you please rebase and resolve the conflicts?
Thank you.

Chen768959:
> If this is ready for merging can you please rebase and resolve the conflicts?
> Thank you.

Sure

PR #3557, opened 2023-12-01
[CALCITE-6149] Unparse for certain EXTRACT Unit with ClickHouseSqlDialect

Chen768959:
Clickhouse does not support the following EXTRACT Unit syntax, and ClickHouseSqlDialect may need some adaptation.

`SELECT EXTRACT(DOW FROM Date '2023-12-01')`
image
`SELECT EXTRACT(DOY FROM Date '2023-12-01')`
image
`SELECT EXTRACT(WEEK FROM Date '2023-12-01')`
image

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -7453,6 +7453,23 @@ class RelToSqlConverterTest {
     sql(query).withSpark().withLibrary(SqlLibrary.SPARK).ok(expectedSql);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6150">[CALCITE-6150]
+   * Clickhouse does not support certain EXTRACT Unit syntax.</a>. */
+  @Test void testClickhouseSpecialExtract() {
+    final String sql1 = "SELECT EXTRACT(DOW FROM Date '2023-12-01')";
+    final String expected1 = "SELECT DAYOFWEEK(toDate('2023-12-01'))";
+    sql(sql1).withClickHouse().ok(expected1);
+
+    final String sql2 = "SELECT EXTRACT(DOY FROM Date '2023-12-01')";
+    final String expected2 = "SELECT DAYOFYEAR(toDate('2023-12-01'))";
+    sql(sql2).withClickHouse().ok(expected2);
+
+    final String sql3 = "SELECT EXTRACT(WEEK FROM Date '2023-12-01')";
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7453,6 +7453,23 @@ private void checkLiteral2(String expression, String expected) {
     sql(query).withSpark().withLibrary(SqlLibrary.SPARK).ok(expectedSql);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6150">[CALCITE-6150]
```
julianhyde, line 7457: Convert this to `testExtract`, move it to the right place in the file (at the end will collide with other changes), and test at least one other dialect.

julianhyde: This is good. Can you break up the strings to make the lines shorter (just one call per line). Put `+` at the start of the line (per our style checker).

Move the method after `testBigQueryDatetimeFormatFunctions`, which is similar. (Anywhere but the end of the file.)

Chen768959:
> @mihaibudiu, By 'no tests for EXTRACT previously' I believe @Chen768959 is referring specifically to `RelToSqlConverterTest`.

Yes

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
Chen768959: > This is good. Can you break up the strings to make the lines shorter (just one call per line). Put `+` at the start of the line (per our style checker).
> 
> Move the method after `testBigQueryDatetimeFormatFunctions`, which is similar. (Anywhere but the end of the file.)

I have fixed the code style and moved `testExtract` after `testBigQuery`. I look forward to your further review. Thank you.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -2264,6 +2264,74 @@ private SqlDialect nonOrdinalDialect() {
     sql(query).withLibrary(SqlLibrary.BIG_QUERY).ok(expected);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6150">[CALCITE-6150]
+   * Generate dialect-specific SQL for EXTRACT operator.</a>. */
```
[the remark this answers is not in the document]
Chen768959: > Maybe you can remove `.` before a tag.

Okay, it seems that other tags also don't have a . at the end, I will remove it.

PR #3558, opened 2023-12-02
[CALCITE-6146] Target charset should be used when comparing two strings through CONVERT/TRANSLATE function during validation

ILuffZhe:


core/src/main/java/org/apache/calcite/sql/SqlBinaryOperator.java
```diff
@@ -171,6 +177,22 @@ private RelDataType convertType(SqlValidator validator, SqlCall call, RelDataTyp
     return type;
   }
 
+  @Nullable Charset getCharsetAfterConvert(SqlBasicCall call, @Nullable Charset typeCharset) {
```
mihaibudiu, line 180: Are there other functions which should preserve charsets?
For example, if you do string concatenation, or cast from CHAR to VARCHAR?

ILuffZhe: As far as I know, only CONVERT and TRANSLATE function explicitly use charset.
In MySQL, CAST function can be used like `cast(name as CHAR CHARACTER SET utf8)`, but we don't support such form currently.

mihaibudiu: I was asking about something like `CAST(TRANSLATE("name" using BIG5) AS CHAR(5))`.
Is the charset of the result of this expression correct?

ILuffZhe: The `char(5)` has the charset ISO-8859-1 preserved in SqlCollation, so the charset of `CAST(TRANSLATE("name" using BIG5) AS CHAR(5))` should also be ISO-8859-1.
I add a test for this, please take a look.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -6827,6 +6827,35 @@ private void checkGetTimestamp(Connection con) throws SQLException {
...
+        + "where _BIG5'Eric'=translate(\"name\" using BIG5)")
+        .returns("name=Eric; empid=200\n");
+    with.query("select \"name\", \"empid\" from \"hr\".\"emps\"\n"
+        + "where convert(convert(\"name\" using GBK) using BIG5)=_BIG5'Eric'")
+        .returns("name=Eric; empid=200\n");
+    // the charset of char(5) is ISO-8859-1 preserved in Collate
+    with.query("select \"name\", \"empid\" from \"hr\".\"emps\"\n"
+        + "where cast(convert(\"name\" using GBK) as char(5))=_BIG5'Eric'")
+        .throws_(
+            "Cannot apply = to the two different charsets ISO-8859-1 and Big5");
```
mihaibudiu, line 6849: I know you didn't write this error message, but I think it can be improved.
I would write "Cannot apply operation '=' to strings with different charsets 'ISO-8859-1' and 'Big5'`

mihaibudiu: Have you tried a similar test as a SqlOperatorTest as well, just to make sure that it triggers there as well?
I don't think this change is JDBC specific.

core/src/main/java/org/apache/calcite/sql/SqlBinaryOperator.java
```diff
@@ -140,7 +140,13 @@ private RelDataType convertType(SqlValidator validator, SqlCall call, RelDataTyp
     if (SqlTypeUtil.inCharFamily(operandType0)
         && SqlTypeUtil.inCharFamily(operandType1)) {
       Charset cs0 = operandType0.getCharset();
+      if (call.operand(0) instanceof SqlBasicCall) {
```
mihaibudiu, line 143: Is this the right place for this check?
Why does the operandType0 have the wrong charset?
Maybe the bug is in the place where the charset for operand0 type was inferred.

ILuffZhe:
@mihaibudiu I mark the PR to draft, and I'll let you know if anything updated in the next few days.

core/src/main/java/org/apache/calcite/sql/SqlBinaryOperator.java
ILuffZhe: > I was asking about something like `CAST(TRANSLATE("name" using BIG5) AS CHAR(5))`. Is the charset of the result of this expression correct?

The result charset of `TRANSLATE("name" using BIG5)` is `BIG5`, so the cast to `CHAR(5)` with `ISO-8859-1` is not allowed.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
ILuffZhe: > I know you didn't write this error message, but I think it can be improved. I would write "Cannot apply operation '=' to strings with different charsets 'ISO-8859-1' and 'Big5'`

Agreed, and I've changed the message.

ILuffZhe: > Have you tried a similar test as a SqlOperatorTest as well, just to make sure that it triggers there as well? I don't think this change is JDBC specific.

I've added some tests in SqlOperatorTest, however the tests `testStringComparisonWithConvertFunc` in JdbcTest still fail.
When creating a new JavaType with different charset(by `createTypeWithCharsetAndCollation`) during validation, the wrapped charset in this JavaType is overwritten through canonizing process due to the `DATATYPE_CACHE`.
I'm not sure this is a bug or specific designed, and I think considering the charset for JavaType's(java.lang.String) digest computing might be a way to go.

core/src/main/java/org/apache/calcite/sql/SqlBinaryOperator.java
ILuffZhe: You are right. I've added inferReturnType for both CONVERT and TRANSLATE function.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -6998,7 +6998,38 @@ private void checkGetTimestamp(Connection con) throws SQLException {
...
-            "Cannot apply = to the two different charsets ISO-8859-1 and UTF-16LE");
+            "Cannot apply operation '=' to strings with different charsets 'ISO-8859-1' and 'UTF-16LE'");
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6146">[CALCITE-6146]
+   * Target charset should be used when comparing two strings through
+   * CONVERT/TRANSLATE function during validation</a>. */
+  @Test void testStringComparisonWithConvertFunc() {
+    if (false) {
```
mihaibudiu, line 7009: this doesn't look ready yet, so I will wait until you request a review

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -6998,7 +6998,51 @@ private void checkGetTimestamp(Connection con) throws SQLException {
...
+    with.query("select \"name\", \"empid\" from \"hr\".\"emps\"\n"
+          + "where _BIG5'Sebastian'=convert(\"name\", UTF8, BIG5)")
+        .returns("name=Sebastian; empid=150\n");
+
+    // check cast
+    with.query("select \"name\", \"empid\" from \"hr\".\"emps\"\n"
+          + "where cast(convert(\"name\" using LATIN1) as char(5))='Eric'")
+        .returns("name=Eric; empid=200\n");
+    with.query("select \"name\", \"empid\" from \"hr\".\"emps\"\n"
+          + "where cast(convert(\"name\" using GBK) as char(5))='Eric'")
```
mihaibudiu, line 7042: why is this the expected result?

core/src/main/java/org/apache/calcite/sql/fun/SqlTranslateFunction.java
```diff
@@ -78,6 +85,40 @@ protected SqlTranslateFunction(String name) {
...
+      descCharsetName = ((RexCallBinding) opBinding).getStringLiteralOperand(1);
+    }
+    assert descCharsetName != null;
+    Charset descCharset = SqlUtil.getCharset(descCharsetName);
+    return typeFactory.createTypeWithCharsetAndCollation(ret, descCharset, getCollation(ret));
+  }
+
+  @Override public RelDataType deriveType(SqlValidator validator,
+      SqlValidatorScope scope, SqlCall call) {
+    // special case for TRANSLATE: don't need to derive type for Charsets
```
mihaibudiu, line 108: it's not obvious what this comment refers to. Special compared to what?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4479,6 +4479,78 @@ static void checkRlikeFails(SqlOperatorFixture f) {
...
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6146">[CALCITE-6146]
+   * Target charset should be used when comparing two strings through
+   * CONVERT/TRANSLATE function during validation</a>. */
+  @Test void testStringComparisonWithConvertFunc() {
+    final SqlOperatorFixture f = fixture();
+    f.setFor(SqlStdOperatorTable.CONVERT, VM_JAVA);
+    f.check("select 'a' as col\n"
```
mihaibudiu, line 4489: this example is confusing, because the column name is col and the literal is also col.
Can you change one of them?

core/src/test/java/org/apache/calcite/test/JdbcTest.java
ILuffZhe: > why is this the expected result?

The result of `convert(\"name\" using GBK)` has `GBK` charset while `CHAR(5)` has `ISO-8859-1` charset, which is not allowed in SqlCastFunction.

core/src/main/java/org/apache/calcite/sql/fun/SqlTranslateFunction.java
ILuffZhe: Got your point, I'll change it to `don't need to derive type for Charsets` in both CONVERT and TRANSLATE. WDYT?

core/src/test/java/org/apache/calcite/test/JdbcTest.java
mihaibudiu: That is not obvious. I would add that as a comment

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
ILuffZhe: > this example is confusing, because the column name is col and the literal is also col. Can you change one of them?

Sure, I've changed the column alia.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
ILuffZhe: OK, I've added comment in both `JdbcTest` and `SqlOperatorTest`

mihaibudiu (approved):
I have never used these functions myself, but, assuming the semantics implemented is the correct one, this PR looks fine.

Can some of these results be validated on another DB?

ILuffZhe:
> I have never used these functions myself, but, assuming the semantics implemented is the correct one, this PR looks fine.
> 
> Can some of these results be validated on another DB?

Thank you for the review! @mihaibudiu 

In earlier's PR, we've tested CONVERT(mysql) and TRANSLATE(bigquery) in function.iq , and you can take a look if it's satisfied.

BTW, I'll try to squash the commits if no more new comments come.

PR #3562, opened 2023-12-04
[CALCITE-4188] Implement BatchNestedLoopJoin for JDBC

kramerul:
See [CALCITE-4188](https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-4188)

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContext.java
+++ b/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContext.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.adapter.jdbc;
+
+import org.apache.calcite.DataContext;
+import org.apache.calcite.adapter.java.JavaTypeFactory;
+import org.apache.calcite.linq4j.QueryProvider;
+import org.apache.calcite.schema.SchemaPlus;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static java.lang.Integer.parseInt;
+
+/**
+ * A special DataContext which handles correlation variable for batch nested loop joins.
+ */
+public class JdbcCorrelationDataContext implements DataContext {
+  public static final int OFFSET = Integer.MAX_VALUE - 10000;
+
+  private final DataContext delegate;
+  private final Object[] parameters;
+
+  public JdbcCorrelationDataContext(DataContext delegate, Object[] parameters) {
+    this.delegate = delegate;
+    this.parameters = parameters;
+  }
+
+  @Override public @Nullable SchemaPlus getRootSchema() {
+    return delegate.getRootSchema();
+  }
+
+  @Override public JavaTypeFactory getTypeFactory() {
+    return delegate.getTypeFactory();
+  }
+
+  @Override public QueryProvider getQueryProvider() {
+    return delegate.getQueryProvider();
+  }
+
+  @Override public @Nullable Object get(String name) {
+    if (name.startsWith("?")) {
+      int index = parseInt(name.substring(1));
+      if (index >= OFFSET && index < OFFSET + parameters.length) {
+        return parameters[index - OFFSET];
+      }
+    }
+    return delegate.get(name);
+  }
+}
--- a/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContextBuilder.java
+++ b/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContextBuilder.java
@@ -1,0 +1,32 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.adapter.jdbc;
+
+import org.apache.calcite.rel.core.CorrelationId;
+
+import java.lang.reflect.Type;
+
+/**
+ * An interface to collect all correlation variables
+ * required to create a JdbcCorrelationDataContext.
+ */
+interface JdbcCorrelationDataContextBuilder {
+  /**
+   * Collect a correlation variable.
+   */
+  int add(CorrelationId id, int ordinal, Type type);
+}
--- a/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcImplementor.java
+++ b/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcImplementor.java
@@ -18,20 +18,66 @@ package org.apache.calcite.adapter.jdbc;
 
 import org.apache.calcite.adapter.java.JavaTypeFactory;
 import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.CorrelationId;
 import org.apache.calcite.rel.rel2sql.RelToSqlConverter;
+import org.apache.calcite.rel.rel2sql.SqlImplementor;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rex.RexCorrelVariable;
 import org.apache.calcite.sql.SqlDialect;
-import org.apache.calcite.util.Util;
+import org.apache.calcite.sql.SqlDynamicParam;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+import java.lang.reflect.Type;
+import java.util.List;
 
 /**
  * State for generating a SQL statement.
  */
 public class JdbcImplementor extends RelToSqlConverter {
-  public JdbcImplementor(SqlDialect dialect, JavaTypeFactory typeFactory) {
+
+  private final JdbcCorrelationDataContextBuilder dataContextBuilder;
+  private final JavaTypeFactory typeFactory;
+
+  JdbcImplementor(SqlDialect dialect, JavaTypeFactory typeFactory,
+      JdbcCorrelationDataContextBuilder dataContextBuilder) {
     super(dialect);
-    Util.discard(typeFactory);
+    this.typeFactory = typeFactory;
+    this.dataContextBuilder = dataContextBuilder;
+  }
+
+  public JdbcImplementor(SqlDialect dialect, JavaTypeFactory typeFactory) {
+    this(dialect, typeFactory, new JdbcCorrelationDataContextBuilder() {
+      private int counter = 1;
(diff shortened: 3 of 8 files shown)


core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContext.java
```diff
@@ -0,0 +1,60 @@
...
+  public static final int OFFSET = 10000;
+
+  private final DataContext delegate;
+  private final Object[] parameters;
+
+  public JdbcCorrelationDataContext(DataContext delegate, Object[] parameters) {
+    this.delegate = delegate;
+    this.parameters = parameters;
+  }
+  @Override public @Nullable SchemaPlus getRootSchema() {
```
tanclary, line 39: Needs newline

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcImplementor.java
```diff
@@ -18,20 +18,64 @@
...
   public Result implement(RelNode node) {
     return dispatch(node);
   }
+
+  @Override protected Context getAliasContext(RexCorrelVariable variable) {
+    Context context = correlTableMap.get(variable.id);
+    if (context != null) {
+      return context;
+    }
+    List<RelDataTypeField>  fieldList = variable.getType().getFieldList();
```
tanclary, line 67: Maybe I'm just not very familiar with JDBC but could we potentially add a comment or two explaining what is going on here. (If it's obvious to everyone but me then maybe not 🤷

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -1147,6 +1153,52 @@ private LockWrapper exclusiveCleanDb(Connection c) throws SQLException {
     });
   }
 
+  @Test void testBatchNestedLoopJoinPlan() {
```
tanclary, line 1156: You could add comment with link to JIRA

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -1147,6 +1153,52 @@ private LockWrapper exclusiveCleanDb(Connection c) throws SQLException {
...
+    final String sql = "SELECT *\n"
+        + "FROM \"s\".\"emps\" A\n"
+        + "LEFT OUTER JOIN \"foodmart\".\"store\" B ON A.\"empid\" = B.\"store_id\"";
+    final String explain = "JdbcFilter(condition=[OR(=($cor0.empid0, $0), =($cor1.empid0, $0)";
+    final String jdbcSql = "SELECT *\n"
+        + "FROM \"foodmart\".\"store\"\n"
```
tanclary, line 1162: Are there any other edge cases or situations we should test? Asked another way: Is one test enough to ensure that this works the way we intend?

tanclary (changes requested):
Thanks for the PR!

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContext.java
kramerul: OK

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcImplementor.java
kramerul: I added a comment

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
kramerul: Sure

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
kramerul: I think one test should be enough, as only one additional execution path is added to the code.

```java
    Context context = correlTableMap.get(variable.id);
    if (context != null) {
      return context;
    }
```

kramerul:
We ares still interested to get this PR merged.

mihaibudiu:
The CI fails due to some formatting problems, which should be easy to fix.
Can you please make this pass CI and then ask for a review when it's ready?

kramerul:
We are still interested to get this PR merged.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContextBuilder.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+package org.apache.calcite.adapter.jdbc;
+
+import org.apache.calcite.rel.core.CorrelationId;
+
+import java.lang.reflect.Type;
+
+/**
+ * A class to build an object of type JdbcCorrelationDataContext.
```
mihaibudiu, line 24: Technically this is an interface.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContextBuilder.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public interface JdbcCorrelationDataContextBuilder {
+  int add(CorrelationId id, int ordinal, Type type);
```
mihaibudiu, line 27: I think you should document the methods in an interface.
I am also not sure how "add" builds an object.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcImplementor.java
```diff
@@ -18,20 +18,66 @@
...
-  public JdbcImplementor(SqlDialect dialect, JavaTypeFactory typeFactory) {
+
+  private final JdbcCorrelationDataContextBuilder dataContextBuilder;
+  private final JavaTypeFactory typeFactory;
+
+  public JdbcImplementor(SqlDialect dialect, JavaTypeFactory typeFactory,
+      JdbcCorrelationDataContextBuilder dataContextBuilder) {
     super(dialect);
-    Util.discard(typeFactory);
+    this. typeFactory = typeFactory;
```
mihaibudiu, line 45: extra space

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcImplementor.java
```diff
@@ -18,20 +18,66 @@
...
 /**
  * State for generating a SQL statement.
  */
 public class JdbcImplementor extends RelToSqlConverter {
-  public JdbcImplementor(SqlDialect dialect, JavaTypeFactory typeFactory) {
+
+  private final JdbcCorrelationDataContextBuilder dataContextBuilder;
+  private final JavaTypeFactory typeFactory;
+
+  public JdbcImplementor(SqlDialect dialect, JavaTypeFactory typeFactory,
```
mihaibudiu, line 42: I don't think you can change public methods in public classes, you need to provide a new constructor and deprecate this one.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -1424,6 +1430,57 @@ private LockWrapper exclusiveCleanDb(Connection c) throws SQLException {
...
+   * Support EnumerableBatchNestedLoopJoin for JDBC</a>. */
+  @Test void testBatchNestedLoopJoinPlan() {
+    final String sql = "SELECT *\n"
+        + "FROM \"s\".\"emps\" A\n"
+        + "LEFT OUTER JOIN \"foodmart\".\"store\" B ON A.\"empid\" = B.\"store_id\"";
+    final String explain = "JdbcFilter(condition=[OR(=($cor0.empid0, $0), =($cor1.empid0, $0)";
+    final String jdbcSql = "SELECT *\n"
+        + "FROM \"foodmart\".\"store\"\n"
+        + "WHERE ? = \"store_id\" OR (? = \"store_id\" OR ? = \"store_id\") OR (? = \"store_id\" OR"
+        + " (? = \"store_id\" OR ? = \"store_id\")) OR (? = \"store_id\" OR (? = \"store_id\" OR ? "
```
mihaibudiu, line 1446: this plan looks terrible.
do you know why?

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContext.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static java.lang.Integer.parseInt;
+
+/**
+ * A special DataContext which handles correlation variable for batch nested loop joins.
+ */
+public class JdbcCorrelationDataContext implements DataContext {
+  public static final int OFFSET = 10000;
```
mihaibudiu, line 32: constant offsets scare me.
can there be some safety check that the number of "standard" parameters does not exceed OFFSET?

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -1424,6 +1430,57 @@ private LockWrapper exclusiveCleanDb(Connection c) throws SQLException {
...
+          planner.addRule(EnumerableRules.ENUMERABLE_BATCH_NESTED_LOOP_JOIN_RULE);
+        })
+        .query(sql)
+        .explainContains(explain)
+        .runs()
+        .enable(CalciteAssert.DB == CalciteAssert.DatabaseInstance.HSQLDB
+            || CalciteAssert.DB == DatabaseInstance.POSTGRESQL)
+        .planHasSql(jdbcSql)
+        .returnsCount(4);
+  }
```
mihaibudiu, line 1482: how about a test with 2 nested correlated subqueries, i.e., nested loops?

mihaibudiu (commented):
I see that @tanclary requested changes, so it would be good for him to review again.

zabetak:
Since CALCITE-5409 is close as a duplicate, I edited the title and description to point to the existing open ticket (CALCITE-4188).

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcImplementor.java
kramerul: I changed the new constructor to package private.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContextBuilder.java
kramerul: I changed the documentation.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContextBuilder.java
kramerul: I changed the documentation accordingly.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcImplementor.java
kramerul: I removed the extra space.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
kramerul: This is caused by the configured batch size (100) of the batch nested loop join.
The batch nested loop join reads 100 entries of the left side and than tries to read the matching 100 entries from the right side. If there are less then 100 entries on the right side, the conditions in this query are filled with the last existing one from the right side.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
kramerul: This will not work. If the 2 nested correlated subqueries are both located in a JDBC datasource, the planner will choose a normal SQL join because this will produce less costs.
If one of the 2 correlated subqueries is not located in a JDBC datasource, the test will not produce additional test coverage for this feature.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCorrelationDataContext.java
kramerul: I increase the OFFSET and added a check.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
mihaibudiu: is there any way to test the fact that the fixed offset does not produce incorrect results? I am worried about the case when there are two contexts that choose the same offsets. 10K or MAX_INT-10K is the same potential problem.

kramerul: There is exactly one `JdbcCorrelationDataContextBuilderImpl` instance used to translate a relation into an SQL statement. All dynamic parameters are local to this statement. Therefore, it's not possible that the same dynamic parameter index is used twice in an SQL statement.

mihaibudiu (approved):

mihaibudiu:
@tanclary you requested changes, so I will wait for your approval before I tag this good-to-merge

tanclary:
Feel free to merge

mihaibudiu:

kramerul:
Sure

PR #3565, opened 2023-12-05
[CALCITE-6156] Add ENDSWITH, STARTSWITH functions (enabled in Postgres, Snowflake libraries)

tanclary:


tanclary:
Going to fix the formatting issues soon.

core/src/main/java/org/apache/calcite/sql/dialect/SnowflakeSqlDialect.java
```diff
@@ -36,6 +39,30 @@ public SnowflakeSqlDialect(Context context) {
...
+      super.unparseCall(writer, call, leftPrec, rightPrec);
+    }
+  }
+
+	/** 
+	 * Remove underscore for {@code STARTS_WITH} and {@code ENDS_WITH} operators to 
+	 * comply with Snowflake syntax.
+	 */
+  private static void unparseEndsStartsWith(final SqlWriter writer, final SqlCall call) {
+    final String name = call.getKind() == SqlKind.ENDS_WITH ? "ENDSWITH" : "STARTSWITH";
```
olivrlee, line 59: Is it convention here to not have underscore?

tanclary: That's just how Snowflake names those functions see: https://docs.snowflake.com/en/sql-reference/functions/startswith

core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
```diff
@@ -73,7 +73,9 @@ public enum SqlLibrary {
   POSTGRESQL("p", "postgresql"),
   /** A collection of operators that are in Apache Spark but not in standard
    * SQL. */
-  SPARK("s", "spark");
+  SPARK("s", "spark"),
+  /** A collection of operators that are in Snowflake but not in standard SQL. */
+  SNOWFLAKE("s", "snowflake");
```
Anthrino, line 78: I could be wrong, but do these single chars point to the symbol we use in `reference.md` to indicate the supported libraries? 
If so should they be distinct symbols for each or its okay to use the same, I see `s` is used for Spark, Snowflake and Spatial but 'MSSQL' = `q` and 'MYSQL' = `m`

Anthrino: Just saw `reference.md` defines its own symbols for each library, just wanted to confirm what these chars are intended for.

julianhyde: List should be alphabetical.

tanclary: Changed to `f` and re-ordered list.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -6557,6 +6557,34 @@ private void checkLiteral2(String expression, String expected) {
...
+   * Add ENDSWITH, STARTSWITH functions (enabled in Postgres, Snowflake libraries)</a>. */
+  @Test void testSnowflakeStartsWith() {
+    final String bigQuerySql = 	"select starts_with(\"brand_name\", 'a')\n"
+        + "from \"product\"";
+    final String snowflakeSql = "select startswith(\"brand_name\", 'a')\n"
+        + "from \"product\"";
+    final String expected = "SELECT STARTSWITH(\"brand_name\" 'a')\n"
+        + "FROM \"foodmart\".\"product\"";
+    sql(bigQuerySql).withLibrary(SqlLibrary.BIG_QUERY).withSnowflake().ok(expected);
+    sql(snowflakeSql).withLibrary(SqlLibrary.SNOWFLAKE).withSnowflake().ok(expected);
```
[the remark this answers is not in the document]
tanclary: Adjusted

core/src/main/java/org/apache/calcite/sql/dialect/SnowflakeSqlDialect.java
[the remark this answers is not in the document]
tanclary: I rewrote this bit in my latest commit.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8878,51 +8878,106 @@ private void testCurrentDateFunc(Pair<String, Hook.Closeable> pair) {
...
-    f.checkNull("starts_with('12345', null)");
-    f.checkNull("starts_with(null, '123')");
-    f.checkBoolean("starts_with('', '123')", false);
-    f.checkBoolean("starts_with('', '')", true);
-    f.checkNull("starts_with(x'aa', null)");
-    f.checkNull("starts_with(null, x'aa')");
-    f.checkBoolean("starts_with(x'1234', x'')", true);
-    f.checkBoolean("starts_with(x'', x'123456')", false);
-    f.checkBoolean("starts_with(x'', x'')", true);
+    final SqlOperatorFixture f0 = fixture();
```
julianhyde, line 8881: I wish there was a way to create tests for functions that are aliases (e.g. starts_with and startswith, length and char_length) that did not use copy-paste. Any ideas?

julianhyde (approved):
Looks good.

If you have ideas about how to combine tests for alias functions please log a jira case.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: Yeah I was thinking this too. 

I filed https://issues.apache.org/jira/browse/CALCITE-6172 in the meantime while I think about it.

PR #3567, opened 2023-12-06
[CALCITE-6048] ServerTest#testTruncateTable fails intermittently due …

libenchao:
…to method not found exception

### DIFF
--- a/core/src/main/java/org/apache/calcite/util/ReflectUtil.java
+++ b/core/src/main/java/org/apache/calcite/util/ReflectUtil.java
@@ -436,7 +436,7 @@ public abstract class ReflectUtil {
             Collections.emptyList());
       }
 
-      @Override public @Nullable Method lookupVisitMethod(
+      @Override public synchronized @Nullable Method lookupVisitMethod(
           Class<? extends R> visitorClass,
           Class<? extends E> visiteeClass,
           String visitMethodName,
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/calcite/util/ReflectUtil.java
```diff
@@ -450,7 +450,8 @@ E extends Object> ReflectiveVisitDispatcher<R, E> createDispatcher(
         Method method = map.get(key);
         if (method == null) {
           if (map.containsKey(key)) {
-            // We already looked for the method and found nothing.
+            // We should get again because it may be putted an object by another thread.
```
mihaibudiu, line 453: Does this solve the race or just reduce the window of vulnerability?
The right way to do this is with some locking mechanism.

libenchao: You are right, adding synchronizing was in my first version. However, after looking at all the failure history, I only see the wrong result, not see the concurrent accessing exceptions, so I just proposed to add a minimal improvement without adding too much synchronizing burden to it.

I don't have a strong opinion on it, if you think adding locking mechanism is a more thorough solution which we should do, I can change back to it.

mihaibudiu: I don't really understand where the threads come from, so I have no idea how to do this.
If all it takes is to make this method synchronized (or some other one), it looks more principled.
If you think this works always, I say it's fine.

libenchao: The multi-threading comes from tests concurrent running, and they share a singleton `ServerDdlExecutor`, which will result concurrent accessing here.

I agree that adding `synchronized` is a better choice, since we already have some similar usages such as in `JaninoRelMetadataProvider`.

mihaibudiu (approved):

JiajunBernoulli:
It seems that the CI running time has not changed.

LGTM.

JiajunBernoulli (approved):

PR #3569, opened 2023-12-07
[CALCITE-6138] Parser does not accept TIMESTAMP WITH TIME ZONE as a data type

mihaibudiu:
This is a large pull request which adds preliminary support for the types `TIME WITH TIME ZONE` and `TIMESTAMP WITH TIME ZONE`. This pull requests contains mostly front-end work, for parsing and properly representing such literals. Follow-up work will need to implement many of the primitive operations on such values, such as casts, field extraction. I plan to contribute some of these in subsequent PRs. 

So this is the first part required to implement CALCITE-208 https://issues.apache.org/jira/browse/CALCITE-208 which was opened 9 years ago.


mihaibudiu:
I have simplified the naming of the types, as suggested by @julianhyde in a JIRA discussion for the uber-issue 208 that this PR is a part of: https://issues.apache.org/jira/browse/CALCITE-208
This is ready for review. If you were waiting for time with timezones, this is your chance to get it!

cassandra/src/main/java/org/apache/calcite/adapter/cassandra/CassandraFilter.java
```diff
@@ -192,6 +193,9 @@ private String translateMatch(RexNode condition) {
     private static Object literalValue(RexLiteral literal) {
       Comparable<?> value = RexLiteral.value(literal);
       switch (literal.getTypeName()) {
+      case TIMESTAMP_TZ:
```
tanclary, line 196: I think we should follow existing naming convention, if local time zone is spelled out then maybe this should be `TIMESTAMP_WITH_TIME_ZONE`

tanclary: If you make this change you can move this `case` below the one beneath to preserve alphabetical order.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystemImpl.java
```diff
@@ -126,8 +128,10 @@ public abstract class RelDataTypeSystemImpl implements RelDataTypeSystem {
       return 65536;
     case TIME:
     case TIME_WITH_LOCAL_TIME_ZONE:
+    case TIME_TZ:
```
tanclary, line 131: same naming comment as above for `TIME`

core/src/main/java/org/apache/calcite/sql/SqlBasicTypeNameSpec.java
```diff
@@ -245,16 +249,31 @@ private static boolean isWithLocalTimeZoneDef(SqlTypeName typeName) {
     }
   }
 
+  private static boolean isWithTimeZoneDef(SqlTypeName typeName) {
```
tanclary, line 252: Alternatively you could add a set/list to `SqlTypeName` (see `DATETIME_TYPES` for example) and then just use `.contains()` or something similar.

core/src/main/java/org/apache/calcite/sql/parser/SqlParserUtil.java
```diff
@@ -351,6 +356,31 @@ public static SqlTimeLiteral parseTimeLiteral(String s, SqlParserPos pos) {
     return SqlLiteral.createTime(t, pt.getPrecision(), pos);
   }
 
+  public static SqlTimeTzLiteral parseTimeTzLiteral(
+      String s, SqlParserPos pos) {
+    // We expect the string to end in a timezone.
+    int lastSpace = s.lastIndexOf(" ");
```
tanclary, line 362: Should this be marked `final`?

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -473,6 +473,9 @@ public static SqlOperandTypeChecker variadic(
   public static final SqlSingleOperandTypeChecker TIMESTAMP_LTZ =
       new TypeNameChecker(SqlTypeName.TIMESTAMP_WITH_LOCAL_TIME_ZONE);
 
+  public static final SqlSingleOperandTypeChecker TIMESTAMP_TZ =
```
tanclary, line 476: Needs comment

core/src/main/java/org/apache/calcite/sql/type/SqlTypeAssignmentRule.java
```diff
@@ -175,13 +175,21 @@ private SqlTypeAssignmentRule(
     rules.add(SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE,
         EnumSet.of(SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE));
 
+    // TIME WITH TIME ZONE is assignable from...
+    rules.add(SqlTypeName.TIME_TZ,
```
tanclary, line 179: This can probably be one line right

core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
```diff
@@ -65,12 +65,16 @@ public enum SqlTypeName {
   DATE(PrecScale.NO_NO, false, Types.DATE, SqlTypeFamily.DATE),
   TIME(PrecScale.NO_NO | PrecScale.YES_NO, false, Types.TIME,
       SqlTypeFamily.TIME),
-  TIME_WITH_LOCAL_TIME_ZONE(PrecScale.NO_NO | PrecScale.YES_NO, false, Types.OTHER,
+  TIME_WITH_LOCAL_TIME_ZONE(PrecScale.NO_NO | PrecScale.YES_NO, false, Types.TIME,
+      SqlTypeFamily.TIME),
+  TIME_TZ(PrecScale.NO_NO | PrecScale.YES_NO, false, Types.TIME,
```
tanclary, line 70: Going through this PR I'm also not opposed to changing the local time zones to something like `TIMESTAMP_LTZ`, I think we actually use that in some parts of the code. Either way should probably be consistent. What do you think?

core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
```diff
@@ -309,7 +314,10 @@ public enum SqlTypeName {
    * matches the given name, or throws {@link IllegalArgumentException}; never
    * returns null. */
   public static SqlTypeName lookup(String tag) {
-    String tag2 = tag.replace(' ', '_');
+    // Special handling for TIME WITH TIME ZONE and
+    // TIMESTAMP WITH TIME ZONE, whose names are TIME_TZ and TIMESTAMP_TZ
+    final String tag1 = tag.replace("WITH TIME ZONE", "TZ");
```
tanclary, line 319: This is kind of personal preference but you could simplify this with a single regex like 
`  return re.sub(r"(?i)WITH TIME ZONE|\s", "TZ", tag)`

tanclary (changes requested):
Looks great 😄

cassandra/src/main/java/org/apache/calcite/adapter/cassandra/CassandraFilter.java
mihaibudiu: The first commit in this PR uses this type name, but in a comment on Jira @julianhyde suggested the new name so the third commit changes the name to this one. The truth is that the long name is very hard to read and is very easy to mistype.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
mihaibudiu: That would be nice, but this may break any third party software that uses Calcite.
Since we can't remove deprecated functions I assume we can't rename public enum fields either.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
mihaibudiu: Since this is not performance-critical I think I prefer the version I wrote, because it is obvious what it does. To understand your version you have to be an expert in regular expressions, and most people aren't. Your version seems to be case-insensitive, but I think that this string will always be uppercase, since it is inserted here by the parser. I will document this behavior.

cassandra/src/main/java/org/apache/calcite/adapter/cassandra/CassandraFilter.java
tanclary: I understand, I am not a fan of the naming style between TIMESTAMP-related types being different though, truth be told.

mihaibudiu:
Thank you @tanclary for the review, I have tried to address your suggestions in the last commit.

mihaibudiu:
This PR is ready for a review. It adds a feature to Calcite which has been requested 9 years ago.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
julianhyde: +1 mihai's perspective

julianhyde:
+1

Thanks for your work, @mihaibudiu!

mihaibudiu:
I am interpreting the comments as an approval, so I plan to merge this soon.
Let me know if you have objections.

mihaibudiu:
No objections from me

snuyanzin (approved):

PR #3570, opened 2023-12-07
[CALCITE-5811] Error messages produced for constant out-of-bounds arg…

mihaibudiu:
…uments are confusing

This turned out to be a bug in the type inference for the arguments of the `substring` function.
The type inference allowed BIGINT arguments.
The `SqlFunctions.substring` method only supports integers as arguments.


mihaibudiu:
I will submit an alternate fix, this one does not seem to be good enough.

PR #3571, opened 2023-12-07
Clarify semantics of SAFE* functions

mihaibudiu:
I expected that I could use `SAFE_ADD(INTEGER, INTEGER)` to check for overflow. It turns out that doesn't work, because `SAFE_ADD` will convert the arguments to `BIGINT` first. I think this deserves to be documented. At least this is the BIGQUERY behavior: https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#safe_add

### DIFF
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -1351,11 +1351,11 @@ comp:
 | DEGREES(numeric)          | Converts *numeric* from radians to degrees
 | PI()                      | Returns a value that is closer than any other value to *pi*
 | RADIANS(numeric)          | Converts *numeric* from degrees to radians
-| ROUND(numeric1 [, numeric2]) | Rounds *numeric1* to optionally *numeric2* (if not specified 0) places right to the decimal point
+| ROUND(numeric1 [, integer2]) | Rounds *numeric1* to optionally *integer2* (if not specified 0) places right to the decimal point
 | SIGN(numeric)             | Returns the signum of *numeric*
 | SIN(numeric)              | Returns the sine of *numeric*
 | TAN(numeric)              | Returns the tangent of *numeric*
-| TRUNCATE(numeric1 [, numeric2]) | Truncates *numeric1* to optionally *numeric2* (if not specified 0) places right to the decimal point
+| TRUNCATE(numeric1 [, integer2]) | Truncates *numeric1* to optionally *integer2* (if not specified 0) places right to the decimal point
 
 ### Character string operators and functions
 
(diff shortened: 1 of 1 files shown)


site/_docs/reference.md
```diff
@@ -2806,14 +2806,14 @@ BigQuery's type system uses confusingly different names for types and functions:
 | h s | string1 NOT RLIKE string2                    | Whether *string1* does not match regex pattern *string2* (similar to `NOT LIKE`, but uses Java regex)
 | b o | RPAD(string, length[, pattern ])             | Returns a string or bytes value that consists of *string* appended to *length* with *pattern*
 | b o | RTRIM(string)                                | Returns *string* with all blanks removed from the end
-| b | SAFE_ADD(numeric1, numeric2)                   | Returns *numeric1* + *numeric2*, or NULL on overflow
+| b | SAFE_ADD(numeric1, numeric2)                   | Returns *numeric1* + *numeric2*, or NULL on overflow.  Only supports BIGINT, DOUBLE, or DECIMAL arguments
```
JiajunBernoulli, line 2809: In fact, you want to emphasize that It will not throw an error when it exceeds integer.

But for me, the comment is not helpful.

mihaibudiu: The fact that it does not throw was already in the message.
I want to emphasize that it does not work as you may expect for integers.

chucheng92: I think there will be ambiguity. The newly added comment will make users think that this function can only use BIGINT, DOUBLE, or DECIMAL. The internal transformation of integer is only internal processing.

mihaibudiu: But this is exactly how all SQL functions work. For example, trigonometric functions are only defined for DOUBLE values. When you call them the compiler inserts an implicit cast for the argument to convert it to double. This kind of "internal processing" happens almost always, it has nothing to do with these functions. It is really part of the definition of SQL. These functions indeed can only use the types you enumerate, if you use a value with another type as argument it will be cast to one of these three types. That is exactly what I am documenting.

What is important in this case is to understand that one cannot use SAFE_* to perform safe computations on integers. If you do SAFE_ADD(int, int) the result will be a BIGINT. If you try to assign this to an integer column and it overflows you will get an exception, which is completely unexpected since this is what the SAFE function was designed to avoid. You also need a SAFE_CAST if you want to avoid the error.

mihaibudiu:
I have changed the descriptions of the SAFE_* functions to say: `Arguments are implicitly cast to one of the types BIGINT, DOUBLE, or DECIMAL`. I hope this is more precise.

mihaibudiu:
I think I have addressed the comments, so I will merge this PR, because no one stepped to provide additional reviews. It only touches the documentation, and I think it's a clear improvement.

PR #3572, opened 2023-12-08
[CALCITE-5811] Error messages produced for constant out-of-bounds arguments are confusing

mihaibudiu:
The bug was a combination of type inference and lack of support for versions of `substring` with long arguments.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlSubstringFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlSubstringFunction.java
@@ -22,7 +22,6 @@ import org.apache.calcite.sql.SqlCallBinding;
 import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlFunctionCategory;
 import org.apache.calcite.sql.SqlKind;
-import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlOperandCountRange;
 import org.apache.calcite.sql.SqlOperatorBinding;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlSubstringFunction.java
```diff
@@ -109,17 +108,16 @@ public class SqlSubstringFunction extends SqlFunction {
...
-      // Reset the operands because they may be coerced during
-      // implicit type coercion.
-      final List<SqlNode> operands = callBinding.getCall().getOperandList();
       final RelDataType t1 = callBinding.getOperandType(1);
       final RelDataType t2 = callBinding.getOperandType(2);
-      if (SqlTypeUtil.inCharFamily(t1)) {
-        if (!SqlTypeUtil.isCharTypeComparable(callBinding, operands,
-            throwOnFailure)) {
-          return false;
+      if (!(SqlTypeUtil.isIntType(t1) || SqlTypeUtil.isNull(t1))) {
```
JiajunBernoulli, line 118: Should we allow null? 
What is the result of its execution?

mihaibudiu: Without null you get a type error when you call the function with a NULL argument. The is a test for this case

JiajunBernoulli (approved):

JiajunBernoulli:
Oh, @mihaibudiu  
We don't have squash commits.

mihaibudiu:
Sorry, will try to fix it tomorrow

PR #3574, opened 2023-12-08
[CALCITE-6161] The equalsDeep of sqlCall should compare sqlOperator's sqlKind

jiefei30:
fix for [CALCITE-6161](https://issues.apache.org/jira/browse/CALCITE-6161)

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlCall.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlCall.java
@@ -181,6 +181,10 @@ public abstract class SqlCall extends SqlNode {
     if (!this.getOperator().getName().equalsIgnoreCase(that.getOperator().getName())) {
       return litmus.fail("{} != {}", this, node);
     }
+    // For same name but different kind
+    if (this.getOperator().getKind() != that.getOperator().getKind()) {
+      return litmus.fail("{} != {}", this, node);
+    }
     if (!equalDeep(this.getFunctionQuantifier(), that.getFunctionQuantifier(), litmus)) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/SqlCall.java
```diff
@@ -181,6 +181,10 @@ public int operandCount() {
     if (!this.getOperator().getName().equalsIgnoreCase(that.getOperator().getName())) {
       return litmus.fail("{} != {}", this, node);
     }
+    // For same name but different kind
+    if (this.getOperator().getKind() != that.getOperator().getKind()) {
```
mihaibudiu, line 185: This looks wrong to me, shouldn't this be `return false`?
Doesn't this throw?

jiefei30: @mihaibudiu You mean the test case i write should `return false`? It will `return true` for now. cause `UNARY_MINUS` and `MINUS` , their names are same, `-`. But It should probably check the operator.  I'll think about this change again

JiajunBernoulli: @jiefei30 
If you could provide an additional method, 
it might be better than modifying the public method.

JiajunBernoulli:
It seems that the CI running time has not changed.

LGTM.

core/src/main/java/org/apache/calcite/sql/SqlCall.java
jiefei30: @JiajunBernoulli There are more test cases i need to add.I'll add them these days.

PR #3578, opened 2023-12-10
[MINOR]  Better message when no corresponding example

caicancai:
example1, no example 1

### DIFF
--- a/core/src/test/java/org/apache/calcite/examples/RelBuilderExample.java
+++ b/core/src/test/java/org/apache/calcite/examples/RelBuilderExample.java
@@ -66,7 +66,7 @@ public class RelBuilderExample {
     case 4:
       return example4(builder);
     default:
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):
This isn't really a typo, but the fix is fine

JiajunBernoulli:
1. The first letter should be capitalized.
2. Message is not accurate enough: We should use `Better message when ....`.

caicancai:
> 1. The first letter should be capitalized.
> 2. Message is not accurate enough: We should use `Better message when ....`.

So do I need to modify the commit message?

JiajunBernoulli:
> > 1. The first letter should be capitalized.
> > 2. Message is not accurate enough: We should use `Better message when ....`.
> 
> So do I need to modify the commit message?

Yes, remember to rebase main.

caicancai:
I finally learned git rebase, thank you all

PR #3584, opened 2023-12-14
[MINOR] Delete useless comments

caicancai:
According to the comments there should be no errors other than unset

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
+++ b/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
@@ -7140,7 +7140,6 @@ public class SqlParserTest {
 
   @Test void testParensInFrom() {
     // UNNEST may not occur within parentheses.
-    // FIXME should fail at "unnest"
     sql("select *from (^unnest(x)^)")
         .fails("Expected query or join");
 


testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -7140,7 +7140,7 @@ private static Consumer<List<? extends Throwable>> checkWarnings(
   @Test void testParensInFrom() {
     // UNNEST may not occur within parentheses.
     // FIXME should fail at "unnest"
-    sql("select *from (^unnest(x)^)")
+    sql("select * from (^unnest(x)^)")
```
mihaibudiu, line 7143: I wonder whether the comment should be removed too. It looks wrong.

caicancai: I don't seem to see where there is a problem. I wrote in the comments that there is a bug in this place and it needs to be fixed in the future.

mihaibudiu: There is a `// FIXME` comment which seems wrong.

caicancai:
```java
sql("select * from (^unnest(x)^)")
        .fails("Expected query or join");
```
At present, the error message of fail is wrong and no information about unset is returned.

mihaibudiu: If the error message returned by the `sql()` function is wrong, then this whole test should fail. But the test passes, so the `fails()` function succeeds, and thus the error message seems correct: the message is `Expected query or join`. 

I don't know what you mean by "no information about unset is returned". Returned by whom? BTW: it is not "unset", but "unnest".

I think that the comment with the FIXME should be deleted. If you are cleaning up the code here you may as well delete the incorrect comment. The comment was inserted 10 years ago in this revision of the file: https://github.com/apache/calcite/commit/45b58ab07223cbb94c7cbbb9a1434a2b01a9497f#diff-53800ccbf3ab769886658b596c055f0e088cc393e1630b774269966154d088ffR5102

The error message has been fixed in the meantime, but the comment has not been deleted at that time. 

Moreover, there are many more instances where `*` is not separated by a space in this file. Why only fix this one?

caicancai: @mihaibudiu Perhaps it expects the error message not to be an Expected query or join, but to display an error message related to the unnset

JiajunBernoulli:
We can use ' to emphasize it.

For example:
Modify `*from` to `* from`.

caicancai:
> We can use ' to emphasize it.
> 
> For example: Modify `*from` to `* from`.



> We can use ' to emphasize it.
> 
> For example: Modify `*from` to `* from`.

Sorry, I don't understand what you mean. Is this a modified comment or a commit message?

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
caicancai: thank you

mihaibudiu (approved):

PR #3585, opened 2023-12-16
[CALCITE-3329] Implement osquery for OS adapter

XuQianJin-Stars:
Fix ISSUE [CALCITE-3329](https://issues.apache.org/jira/browse/CALCITE-3329), Achieve similar features of FaceBook's osquery. E.g:
```
select * from os_version;
select * from system_info;
select * from mounts;
select * from interface_addresses
select * from memory_info;
select * from interface_addresses;
select * from cpu_info;
select * from cpu_time;
select * from java_info;
```


plus/src/main/java/org/apache/calcite/adapter/utils/OsQueryTableUtil.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+    return list;
+  }
+
+  public static List<Object[]> getMountsInfo() {
+    final List<Object[]> list = new ArrayList<>();
+    FileSystem fileSystem = OshiUtil.getOs().getFileSystem();
+    List<OSFileStore> fileStores = fileSystem.getFileStores();
+    for (OSFileStore fileStore : fileStores) {
+      Object[] objects = new Object[8];
```
julianhyde, line 204: I prefer
```
Object[] objects = {
   fileStore.getName(),
   fileStore.getName(),
   getNetFileSizeDescription(fileStore.getTotalSpace()),
    ...
};
```
Without explicit length and offset it is more maintainable.

plus/src/main/java/org/apache/calcite/adapter/os/SqlShell.java
```diff
@@ -64,6 +64,10 @@ public class SqlShell {
     this.err = Objects.requireNonNull(err, "err");
   }
 
+  public static void setModel(String newModel) {
```
julianhyde, line 67: Does this really need to be mutable?

Could you achieve what you need with a `withModel(String model)` method?

julianhyde (changes requested):

PR #3587, opened 2023-12-19
[CALCITE-6116] Add EXISTS function (enabled in Spark library)

macroguo-ghy:


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7382,6 +7382,25 @@ private void checkLiteral2(String expression, String expected) {
     sql(sql6).ok(expected6);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6116">[CALCITE-6116]
+   * Add EXISTS function (enabled in Spark library)</a>. */
+  @Test void testExistsFunctionInSpark() {
```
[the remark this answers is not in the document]
mihaibudiu: also, a test with an empty array

core/src/test/resources/sql/higher-order-functions.iq
```diff
@@ -0,0 +1,90 @@
+# higher-order-functions.iq - Queries involving higher-order functions
```
julianhyde, line 1: Can you rename this file to `lambda.iq`?

It's debatable whether SQL lambdas are true higher-order functions.

site/_docs/reference.md
```diff
@@ -3132,6 +3133,10 @@ parameters:
   |   identifier
 {% endhighlight %}
 
+| C | Operator syntax                                | Description
+|:- |:-----------------------------------------------|:-----------
+| s | EXISTS(array, predicate)                       | Returns whether a *predicate* holds for one or more elements in the *array*
+
```
julianhyde, line 3139: I don't think you need this table twice in the same file. I think you should replace this table with a paragraph 'Examples of functions with a lambda argument are EXISTS, ...'.

site/_docs/reference.md
```diff
@@ -2730,6 +2730,7 @@ BigQuery's type system uses confusingly different names for types and functions:
 | p | DIFFERENCE(string, string)                     | Returns a measure of the similarity of two strings, namely the number of character positions that their `SOUNDEX` values have in common: 4 if the `SOUNDEX` values are same and 0 if the `SOUNDEX` values are totally different
 | b | ENDS_WITH(string1, string2)                    | Returns whether *string2* is a suffix of *string1*
 | o | EXTRACT(xml, xpath, [, namespaces ])           | Returns the XML fragment of the element or elements matched by the XPath expression. The optional namespace value that specifies a default mapping or namespace mapping for prefixes, which is used when evaluating the XPath expression
+| s | EXISTS(array, predicate)                       | Returns whether a *predicate* holds for one or more elements in the *array*
```
julianhyde, line 2733: you should describe what the `predicate` argument means. Not in this row, but before the table, where other arguments are described.

julianhyde (changes requested):

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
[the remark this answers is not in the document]
macroguo-ghy: Thanks for your review! You are correct that using the `any` type is inaccurate, and it would be better to use the element type of the array for type validation. I will fix it later.

PR #3588, opened 2023-12-19
[CALCITE-6168] RexExecutor can throw during compilation

mihaibudiu:


mihaibudiu:
Maybe I should refine the catch block to catch fewer exceptions.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1418,6 +1418,14 @@ protected static Calendar getCalendarNotTooNear(int timeUnit) {
     }
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6168">[CALCITE-6168]
```
tanclary, line 1422: I don't think this test belongs here, it's not a test case for a specific operator. I think there is a test class called `RexExecutorTest` that might be better.

tanclary (changes requested):

mihaibudiu:
Thank you @tanclary for the review, I have pushed one more commit which addresses your suggestion.

mihaibudiu:
Can someone please review this PR which fixes a bug?

testkit/src/main/java/org/apache/calcite/sql/test/SqlTests.java
```diff
@@ -265,6 +265,20 @@ public static void checkEx(@Nullable Throwable ex,
...
+    // Search for an IllegalStateException somewhere in the stack.
+    // These are thrown by the enumerable implementors when evaluating
+    // an expression that produces an error.
+    IllegalStateException ise = null;
+    for (Throwable x = ex; x != null; x = x.getCause()) {
+      if (x instanceof IllegalStateException) {
+        ise = (IllegalStateException) x;
+        break;
+      }
+      if (x.getCause() == x) {
```
tanclary, line 277: Is this just to prevent SO?

tanclary (approved):
LGTM

testkit/src/main/java/org/apache/calcite/sql/test/SqlTests.java
mihaibudiu: I have seen this pattern used elsewhere in this function, so I have followed it. I am not sure whether a Throwable can have itself as a cause. It looks like the Java APIs make that possible, since you can create a Throwable and assign a cause to it later using initCause.

PR #3589, opened 2023-12-21
[CALCITE-3522] SqlValidator.validateLiteral rejects literals with a DECIMAL type that require more than 64 bits

mihaibudiu:
The comment below is no longer accurate after the PR was split into two separate PRs.

--------------

I initially started with the goal of fixing just [CALCITE-3522], https://issues.apache.org/jira/browse/CALCITE-3522, which is limiting the size of decimal literals to 64 bits. The fix is to actually check the precision of the type system when rejecting a decimal literal. That's easy. I am not checking the scale, under the assumption that rounding is OK for literals but losing precision is not.

However, fixing this problem unmasked a bunch of other bugs in handling large values, so in order to make the tests pass I had to also fix [CALCITE-6169], which is about incorrect casts between numeric values. There is still one case which is not handled, which is casts between decimal values that change the scale of the result, but I plan to do that in a separate PR (because I also have to think about it more.)

You will notice that this fix re-enables some tests that were disabled a long time ago. I think that disabling failing tests is not a good practice, and should be avoided. There are still some disabled tests in these categories, but strictly fewer than before.

I couldn't make this into 2 separate PRs that both keep all tests running. 

I am not 100% happy with the code, in particular with the implementation of the new Expression `ConvertChecked`. The implementation (which is in the unparse method...) in fact just calls the existing code in `Primitive` which handles numeric casts. I think that's better than duplicating the code. There is already a very large number of places where these conversions are handled, which is shown by the many files that I had to patch.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlNumericLiteral.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlNumericLiteral.java
@@ -103,7 +103,7 @@ public class SqlNumericLiteral extends SqlLiteral {
           BigDecimal bd = getValueNonNull();
           SqlTypeName result;
           // Will throw if the number cannot be represented as a long.
-          long l = bd.longValue();
+          long l = bd.longValueExact();
           if ((l >= Integer.MIN_VALUE) && (l <= Integer.MAX_VALUE)) {
             result = SqlTypeName.INTEGER;
           } else {
@@ -115,7 +115,6 @@ public class SqlNumericLiteral extends SqlLiteral {
           // Fallback to DECIMAL.
         }
       }
-
       // else we have a decimal
       return typeFactory.createSqlType(
           SqlTypeName.DECIMAL,
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
@@ -133,7 +133,6 @@ import org.checkerframework.dataflow.qual.Pure;
 import org.slf4j.Logger;
 
 import java.math.BigDecimal;
-import java.math.BigInteger;
 import java.util.AbstractList;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -3417,20 +3416,21 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
   @Override public void validateLiteral(SqlLiteral literal) {
     switch (literal.getTypeName()) {
     case DECIMAL:
-      // Decimal and long have the same precision (as 64-bit integers), so
-      // the unscaled value of a decimal must fit into a long.
+      // Accept any decimal value that does not exceed the max
+      // precision and scale of the type system.
+      final RelDataTypeSystem typeSystem = getTypeFactory().getTypeSystem();
+      final BigDecimal bd = literal.getValueAs(BigDecimal.class);
+      final BigDecimal noTrailingZeros = bd.stripTrailingZeros();
+      // If we don't strip trailing zeros we may reject values such as 1.000....0.
 
-      // REVIEW jvs 4-Aug-2004:  This should probably be calling over to
-      // the available calculator implementations to see what they
-      // support.  For now use ESP instead.
-      //
-      // jhyde 2006/12/21: I think the limits should be baked into the
-      // type system, not dependent on the calculator implementation.
-      BigDecimal bd = literal.getValueAs(BigDecimal.class);
-      BigInteger unscaled = bd.unscaledValue();
-      long longValue = unscaled.longValue();
-      if (!BigInteger.valueOf(longValue).equals(unscaled)) {
-        // overflow
+      final int maxPrecision = typeSystem.getMaxNumericPrecision();
+      if (noTrailingZeros.precision() > maxPrecision) {
+        throw newValidationError(literal,
+            RESOURCE.numberLiteralOutOfRange(bd.toString()));
+      }
+
+      final int maxScale = typeSystem.getMaxNumericScale();
+      if (noTrailingZeros.scale() > maxScale) {
         throw newValidationError(literal,
             RESOURCE.numberLiteralOutOfRange(bd.toString()));
       }
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -430,6 +430,43 @@ public class SqlOperatorTest {
         false);
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-3522">
+   * Sql validator limits decimal literals to 64 bits</a>. */
+  @Test void testLargeLiterals() {
+    // Some of these literals were too large to be accepted previously, but
+    // now are legal as decimal literals.
+    SqlOperatorFixture f = fixture();
+    f.checkCastFails("9223372036854775808", "INTEGER",
+        OUT_OF_RANGE_MESSAGE, true, SqlOperatorFixture.CastType.CAST);
+    f.checkCastFails("9223372036854775808.1", "INTEGER",
+        "Numeric literal.*out of range", false, SqlOperatorFixture.CastType.CAST);
+    f.checkCastFails("223372036854775808", "INTEGER",
+        OUT_OF_RANGE_MESSAGE, true, SqlOperatorFixture.CastType.CAST);
+    f.checkCastFails("9223372036854775808", "BIGINT",
+        "Overflow", true, SqlOperatorFixture.CastType.CAST);
+    f.checkCastFails("'" + Numeric.TINYINT.maxOverflowNumericString + "'",
+        "TINYINT", OUT_OF_RANGE_MESSAGE, true, SqlOperatorFixture.CastType.CAST);
+    String largePrecision = "1234567891011.0";
+    String largeScale = "1.01234567891011";
+    f.checkScalarExact(largePrecision, "DECIMAL(14, 1) NOT NULL", largePrecision);
+    f.checkScalarExact(largeScale, "DECIMAL(15, 14) NOT NULL", largeScale);
+
+    // Check that the type system can reject large decimal literals
+    SqlOperatorFixture f0 = f.withFactory(tf ->
+            tf.withTypeSystem(typeSystem ->
+                new DelegatingTypeSystem(typeSystem) {
+                  @Override public int getMaxNumericPrecision() {
+                    return 10;
+                  }
+
+                  @Override public int getMaxNumericScale() {
+                    return 10;
+                  }
+                }));
+    f0.checkFails("^" + largePrecision + "^", OUT_OF_RANGE_MESSAGE, false);
+    f0.checkFails("^" + largeScale + "^", OUT_OF_RANGE_MESSAGE, false);
+  }
+
   @Test void testNotBetween() {
     final SqlOperatorFixture f = fixture();
     f.setFor(SqlStdOperatorTable.NOT_BETWEEN, VM_EXPAND);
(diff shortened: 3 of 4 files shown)


core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
```diff
@@ -406,11 +407,28 @@ public static Expression convert(Expression operand, Type fromType,
     if (toPrimitive != null) {
       if (fromPrimitive != null) {
         // E.g. from "float" to "double"
-        return Expressions.convert_(
+        if (toPrimitive == Primitive.BOOLEAN) {
```
tanclary, line 410: nit: third level of nested `if` might be a bit much, no worries if refactoring isn't simple/easy.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
```diff
@@ -406,11 +407,28 @@ public static Expression convert(Expression operand, Type fromType,
...
+          return Expressions.convert_(
+              operand, toPrimitive.getPrimitiveClass());
+        }
+        return Expressions.convertChecked(
             operand, toPrimitive.getPrimitiveClass());
       }
-      if (fromNumber || fromBox == Primitive.CHAR) {
-        // Generate "x.shortValue()".
+      if (fromType == BigDecimal.class
+          && (toPrimitive == Primitive.INT
```
tanclary, line 418: could add a method to the `Primitive` class similar to `isNumeric()` in case another check like this is needed in the future, also reduces `if` complexity.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3334,20 +3333,11 @@ private void registerOperandSubQueries(
   @Override public void validateLiteral(SqlLiteral literal) {
     switch (literal.getTypeName()) {
     case DECIMAL:
-      // Decimal and long have the same precision (as 64-bit integers), so
-      // the unscaled value of a decimal must fit into a long.
-
```
tanclary, line 3339: is there a better comment to leave in its place or no?>

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3334,20 +3333,11 @@ private void registerOperandSubQueries(
...
-      // jhyde 2006/12/21: I think the limits should be baked into the
-      // type system, not dependent on the calculator implementation.
+      final RelDataTypeSystem typeSystem = getTypeFactory().getTypeSystem();
+      final int precision = typeSystem.getMaxNumericPrecision();
       BigDecimal bd = literal.getValueAs(BigDecimal.class);
-      BigInteger unscaled = bd.unscaledValue();
-      long longValue = unscaled.longValue();
-      if (!BigInteger.valueOf(longValue).equals(unscaled)) {
-        // overflow
+      BigDecimal simpler = bd.stripTrailingZeros();
```
tanclary, line 3339: Some of the naming here confuses me, what does `simpler` mean? (in this context, haha).

linq4j/src/main/java/org/apache/calcite/linq4j/tree/UnaryExpression.java
```diff
@@ -51,6 +51,24 @@ public class UnaryExpression extends Expression {
...
+    case ConvertChecked:
+      // This is ugly, but Java does not seem to have any facilities
+      // to perform checked cast between scalar types!
+      // So we use the existing linq4j Primitive.numberValue method
+      // which does overflow checking.
+      if (!writer.requireParentheses(this, lprec, rprec)) {
+        // Generate e.g.,
+        // ((Number)org.apache.calcite.linq4j.tree.Primitive.of(int.class)
+        //     .numberValue(literal_value)).intValue();
+        writer.append("((Number)")
```
tanclary, line 63: Unfortunate you had to write this yourself.

mihaibudiu:
Thank you @tanclary for the review, I pushed a new commit which hopefully addresses your comments.

mihaibudiu:
This bugfix is also ready for review.

tanclary (commented):
This is a little over my head, might want to get someone else to review.

mihaibudiu:
If anyone wants to do a pair review we can arrange a short meeting so I can explain what's going on.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
mihaibudiu: this comment is now obsolete, we are no longer tied to the precision of long

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
```diff
@@ -406,11 +407,24 @@ public static Expression convert(Expression operand, Type fromType,
     if (toPrimitive != null) {
       if (fromPrimitive != null) {
         // E.g. from "float" to "double"
-        return Expressions.convert_(
+        if (toPrimitive == Primitive.BOOLEAN) {
+          return Expressions.convert_(
+              operand, toPrimitive.getPrimitiveClass());
+        }
+        return Expressions.convertChecked(
```
mihaibudiu, line 414: convertChecked is implemented in this PR.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
```diff
@@ -406,11 +407,24 @@ public static Expression convert(Expression operand, Type fromType,
...
+      if (fromType == BigDecimal.class && toPrimitive.isFixedNumeric()) {
+        // Conversion from decimal to an exact type
+        ConstantExpression zero = Expressions.constant(0);
+        // Elsewhere Calcite uses this rounding mode implicitly, so we have to be consistent.
+        // E.g., this is the rounding mode used by BigDecimal.longValue().
+        Expression rounding = Expressions.constant(RoundingMode.DOWN);
+        // rounded = operand.setScale(0, RoundingMode.DOWN);
+        Expression rounded = Expressions.call(operand, "setScale", zero, rounding);
+        // return rounded.to*ValueExact()
+        return Expressions.unboxExact(rounded, toPrimitive);
```
mihaibudiu, line 426: conversion from decimal to some integer type needs rounding, here we use the DOWN rounding, which is implicitly used in Calcite wherever BigDecimal.longValue is used. If a different rounding mode is desired we should file an issue.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -699,22 +699,23 @@ boolean canRemoveCastFromLiteral(RelDataType toType, @Nullable Comparable value,
 
     if (SqlTypeName.INT_TYPES.contains(sqlType)) {
       final BigDecimal decimalValue = (BigDecimal) value;
-      final int s = decimalValue.scale();
-      if (s != 0) {
+      try {
+        // Will throw if the value does not fit in a long
+        long l = decimalValue.longValueExact();
```
mihaibudiu, line 704: check if the decimal value will fit in the target type

core/src/main/java/org/apache/calcite/sql/SqlNumericLiteral.java
```diff
@@ -99,17 +99,23 @@ public boolean isExact() {
...
       if (0 == scaleValue) {
-        BigDecimal bd = getValueNonNull();
-        SqlTypeName result;
-        long l = bd.longValue();
-        if ((l >= Integer.MIN_VALUE) && (l <= Integer.MAX_VALUE)) {
-          result = SqlTypeName.INTEGER;
-        } else {
-          result = SqlTypeName.BIGINT;
+        try {
+          BigDecimal bd = getValueNonNull();
```
mihaibudiu, line 103: numeric literals are always represented as Decimal by the parser.
Here we check whether the value can be represented by a narrower type without information loss.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3334,20 +3333,14 @@ private void registerOperandSubQueries(
...
-      if (!BigInteger.valueOf(longValue).equals(unscaled)) {
-        // overflow
+      // Accept any decimal value that does not exceed the max
+      // precision of the type system.
+      final RelDataTypeSystem typeSystem = getTypeFactory().getTypeSystem();
+      final int maxPrecision = typeSystem.getMaxNumericPrecision();
+      final BigDecimal bd = literal.getValueAs(BigDecimal.class);
+      final BigDecimal noTrailingZeros = bd.stripTrailingZeros();
+      // If we don't strip trailing zeros we may reject values such as 1.000....0.
+      if (noTrailingZeros.precision() > maxPrecision) {
```
mihaibudiu, line 3343: this code will accept any literal which fits in the specified precision.
I am not sure whether we have to check the scale too.

mihaibudiu:
@libenchao I have added some review comments to this PR, if you can help with a review we can hopefully merge this. Thank you.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
jduo: Should these checks factor in unsigned types?

mihaibudiu: There are no unsigned types in SQL.

mihaibudiu:
I have submitted https://github.com/apache/calcite/pull/3764, which provides just a fix for [CALCITE-6169].
Once that is merged, I will resubmit a much smaller PR here to fix separately [CALCITE-3522]

mihaibudiu:
I have reworked this PR since it was split into two separate PRs, one of which has just been merged #3764. Now it's much simpler.
This will hopefully allow us to close a 5-year old issue in Calcite.
Incidentally it also seems to fix [CALCITE-2539] which I plan to resolve if this is merged.

core/src/main/java/org/apache/calcite/sql/SqlNumericLiteral.java
```diff
@@ -103,7 +103,7 @@ public boolean isExact() {
           BigDecimal bd = getValueNonNull();
           SqlTypeName result;
           // Will throw if the number cannot be represented as a long.
-          long l = bd.longValue();
+          long l = bd.longValueExact();
```
asolimando, line 106: As per the [documentation](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#longValueExact--), this will throw an `ArithmeticException` in case of overflow, should we catch it and fallback do decimal?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -626,16 +644,15 @@ void testCastExactNumericLimits(CastType castType, SqlOperatorFixture f) {
...
-            LITERAL_OUT_OF_RANGE_MESSAGE, false);
+        // Overflow for casting decimals produces a different error
+        f.checkCastFails(numeric.maxOverflowNumericString,
+            type, "Overflow", true, castType);
+        f.checkCastFails(numeric.minOverflowNumericString,
+            type, "Overflow", true, castType);
       } else {
         if (numeric != Numeric.DECIMAL5_2) {
           // This condition is for bug [CALCITE-6078], not yet fixed
+          // FIXME: decimal casts are still incorrect.
```
asolimando, line 655: Nit: Is this comment adding anything more than the existing one?

core/src/main/java/org/apache/calcite/sql/SqlNumericLiteral.java
mihaibudiu: yes, that is exactly what is going on here. This is within a try-catch block

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: this is actually fixed as part of another PR which I have had in the queue for a long time. I will remove this comment.

mihaibudiu:
@asolimando I hope I have answered your questions

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
asolimando: I think we should, I can't think of any reason why we should not, do you have anything in mind?

To this extend, we could use
[BigDecimal::scale()](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#scale--) to check compatibility with what the type system supports.

WDYT?

asolimando (commented):
Apart from the discussion around handling the scale, I am happy with the PR and I don't have further comments, let's clarify that point so we can finalize it

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
mihaibudiu: I added this check, no tests required changes.

asolimando: Thanks @mihaibudiu, can we add a negative test for it so we make sure it works as intended and it will prevent regressions? After that for me it's good to be merged.

mihaibudiu: I should have done this from the beginning, thank you for reminding me. I added a tiny commit with negative tests.

asolimando (approved):
LGTM, thanks @mihaibudiu for fixing this long-standing issue!

PR #3590, opened 2023-12-21
[CALCITE-6172] PoC for parameterized tests

snuyanzin:


mihaibudiu (commented):
There is another way to do this, which requires fewer manual string updates, but perhaps more work: since we already have a SQL compiler, we can just parse the program/expression and use a shuttle to replace the call with a different function. The shuttle can do even things like swapping arguments.

snuyanzin:
I didn't get how shuttle could help with different `SQLLibraries` per alias...

mihaibudiu:
You leave the queries unchanged, and then you use a shuttle to replace the use of `starts_with` with `startswith`.

mihaibudiu:
It's not obvious that it's simpler than your solution, but it is optimized towards not changing the tested queries. Fewer chances to introduce errors during the manual rewrite.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13686,6 +13673,31 @@ void testCastTruncates(CastType castType, SqlOperatorFixture f) {
     }
   }
 
+  static class AliasInfo {
+    private final SqlFunction function;
+    private final String alias;
+    private final SqlLibrary[] libraries;
```
julianhyde, line 13679: Constructor parameter should be varargs, not array.

`libraries` field should be immutable list, not array.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8897,56 +8898,42 @@ private void testCurrentDateFunc(Pair<String, Hook.Closeable> pair) {
     f0.forEachLibrary(list(SqlLibrary.BIG_QUERY, SqlLibrary.ORACLE), consumer);
   }
 
-  @Test void testStartsWithFunction() {
```
julianhyde, line 8900: Need a comment saying that it tests STARTS_WITH and STARTSWITH.

julianhyde:
@mihaibudiu The danger with your approach is that, by working on the AST, you automatically fix errors. E.g. you might accidentally change STARTSWITH to STARTS_WITH and not notice that STARTSWITH is a reserved word. Or the unparser might convert things to upper-case.

For this, I favor a brute force approach that generates strings.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13686,6 +13673,31 @@ void testCastTruncates(CastType castType, SqlOperatorFixture f) {
     }
   }
 
+  static class AliasInfo {
+    private final SqlFunction function;
+    private final String alias;
+    private final SqlLibrary[] libraries;
+
+    private  AliasInfo(SqlFunction function, String alias, SqlLibrary[] libraries) {
```
julianhyde, line 13681: Use a better name than `AliasInfo`. Info means data, after all. Maybe `FunctionAlias`. Also, move it to the end of the file.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8897,56 +8898,44 @@ private void testCurrentDateFunc(Pair<String, Hook.Closeable> pair) {
...
     final Consumer<SqlOperatorFixture> consumer = f -> {
-      f.checkBoolean("starts_with('12345', '123')", true);
-      f.checkBoolean("starts_with('12345', '1243')", false);
-      f.checkBoolean("starts_with(x'11', x'11')", true);
-      f.checkBoolean("starts_with(x'112211', x'33')", false);
-      f.checkFails("^starts_with('aabbcc', x'aa')^",
-          "Cannot apply 'STARTS_WITH' to arguments of type "
-              + "'STARTS_WITH\\(<CHAR\\(6\\)>, <BINARY\\(1\\)>\\)'\\. Supported "
-              + "form\\(s\\): 'STARTS_WITH\\(<STRING>, <STRING>\\)'",
+      f.checkBoolean(alias + "('12345', '123')", true);
```
mihaibudiu, line 8910: Another possible suggestion is to have the alias implement a "rewrite" method which rewrites a specific template. Then you can write tests like `f.checkBoolean(alias.rewrite("%function%('12345', '1234')"))`. But I don't know if the readability of the test improves much.

snuyanzin: Do we need then `"%function%`?
May be just name this method `for` or `applyFor` and use it like
```java
f.checkBoolean(alias.applyFor("('12345', '1234')"));
```
?

mihaibudiu: Your proposal works for standard function syntax, but sql is a weird language. But your proposal may be good enough for most use cases.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8897,56 +8898,44 @@ private void testCurrentDateFunc(Pair<String, Hook.Closeable> pair) {
...
-    f.checkNull("startswith(null, x'aa')");
-    f.checkBoolean("startswith(x'1234', x'')", true);
-    f.checkBoolean("startswith(x'', x'123456')", false);
-    f.checkBoolean("startswith(x'', x'')", true);
+  static Stream<FunctionAlias> startsWithAliases() {
+    return Stream.of(
+        FunctionAlias.of(
+            SqlLibraryOperators.STARTS_WITH,
+            "starts_with", SqlLibrary.BIG_QUERY, SqlLibrary.POSTGRESQL),
+        FunctionAlias.of(SqlLibraryOperators.STARTSWITH, "startswith", SqlLibrary.SNOWFLAKE));
```
tanclary, line 8938: Do we need to pass the String names in or could we just use the `.name()` of the operator?

snuyanzin: yes, makes sense, changed it, also added similar approach to retrieve libraries

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
snuyanzin: i put it in a separate commit here to see how it could look like
https://github.com/apache/calcite/pull/3590/commits/88fc66326964d1019de37b9b924445ac5648ad6e

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8897,56 +8899,43 @@ private void testCurrentDateFunc(Pair<String, Hook.Closeable> pair) {
     f0.forEachLibrary(list(SqlLibrary.BIG_QUERY, SqlLibrary.ORACLE), consumer);
   }
 
-  @Test void testStartsWithFunction() {
+  /** Tests the {@code STARTS_WITH} and {@code STARTSWITH} operator. */
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("startsWithAliases")
+  void testStartsWithFunction2(FunctionAlias functionAlias) {
```
tanclary, line 8905: I think this is looking good, if we merge it you can remove the 2

tanclary (approved):
I'm happy with how this looks. If Julian and Mihai agree we can merge.

mihaibudiu:
It's still marked as draft for some reason.

mihaibudiu (approved):

tanclary:
> It's still marked as draft for some reason.

I converted to ready for review, hope that's alright.

julianhyde:
Is this PR just a proof of concept? If so, let’s not merge the PoC, let’s merge a commit that does the actual task - reduces duplication of testing for aliased functions. By all means give Sergey co-author credit in that commit if appropriate. JulianOn Dec 27, 2023, at 14:07, Tanner Clary ***@***.***> wrote:﻿

It's still marked as draft for some reason.

I converted to ready for review, hope that's alright.

mihaibudiu:
This has developed a conflict, can you please fix it and I will merge.

tanclary:
> This has developed a conflict, can you please fix it and I will merge.

This is just a PoC it can be closed, I merged the actual impl

snuyanzin:
closing since the PR was merged

PR #3591, opened 2023-12-22
[CALCITE-6174] Upgrade gradle from 7.6.1 to 8.7

snuyanzin:
The PR is to bump Gradle from 7.6.1 to 8.7 as a required step before starting activity to support java 20+


gradle.properties
```diff
@@ -174,3 +174,5 @@ xercesImpl.version=2.9.1
 systemProp.sonar.organization=apache
 systemProp.sonar.projectKey=apache_calcite
 systemProp.sonar.host.url=https://sonarcloud.io
+
+kotlin.jvm.target.validation.mode = IGNORE
```
snuyanzin, line 178: This is a result of a weird Kotlin bug https://youtrack.jetbrains.com/issue/KT-48745
which is unlikely to be fixed

Jenkinsfile
```diff
@@ -44,9 +44,9 @@ node('ubuntu') {
       withEnv(["Path+JDK=$JAVA_JDK_17/bin","JAVA_HOME=$JAVA_JDK_17"]) {
         withCredentials([string(credentialsId: 'SONARCLOUD_TOKEN', variable: 'SONAR_TOKEN')]) {
           if ( env.BRANCH_NAME.startsWith("PR-") ) {
-            sh './gradlew --no-parallel --no-daemon jacocoAggregateTestReport sonar -PenableJacoco -Dsonar.pullrequest.branch=${CHANGE_BRANCH} -Dsonar.pullrequest.base=${CHANGE_TARGET} -Dsonar.pullrequest.key=${CHANGE_ID} -Dsonar.login=${SONAR_TOKEN}'
+            sh './gradlew --no-parallel --no-daemon jacocoAggregateTestReport sonar -PenableJacoco -Porg.sonarqube.version=4.4.1.3373 -Dsonar.pullrequest.branch=${CHANGE_BRANCH} -Dsonar.pullrequest.base=${CHANGE_TARGET} -Dsonar.pullrequest.key=${CHANGE_ID} -Dsonar.login=${SONAR_TOKEN}'
```
snuyanzin, line 47: From one side 4.x is required to work with gradle 8 when it is invoked for `jacocoAggregateTestReport`
from another side gradle can not load it while building with jdk8 since 4.x can not be used with jdk8

build.gradle.kts
```diff
@@ -66,15 +67,6 @@ repositories {
     mavenCentral()
 }
 
-tasks.wrapper {
```
snuyanzin, line 69: now this task is embedded into gradle, so no need for a custom one

caicancai (approved):
Local test successful

Jenkinsfile
zabetak: I am not sure I understood why we are not updating the version directly to the gradle.properties file. I have the impression that running sonar with jdk8 is not supported/recommended.

build.gradle.kts
zabetak: Maybe worth adding this clarification in the final commit message in case people in the future wonder why it was removed.

build.gradle.kts
```diff
@@ -965,9 +982,11 @@ allprojects {
                             asNode()
                         }
                         name.set(
-                            (project.findProperty("artifact.name") as? String) ?: "Calcite ${project.name.capitalize()}"
+                            (project.findProperty("artifact.name") as? String) ?: "Calcite ${project.name.replaceFirstChar {
```
zabetak, line 985: For our context I think that `String.replaceFirstChar { it.uppercaseChar() }` should be sufficient. We don't have exotic characters in the project name.

Also maybe worth mentioning in the commit message that we replace `capitalize` cause it is deprecated.

core/build.gradle.kts
```diff
@@ -272,15 +273,18 @@ val integTestAll by tasks.registering() {
 }
 
 for (db in listOf("h2", "mysql", "oracle", "postgresql")) {
-    val task = tasks.register("integTest" + db.capitalize(), Test::class) {
+    val task = tasks.register("integTest" + db.replaceFirstChar { if (it.isLowerCase()) it.titlecase(
```
zabetak, line 276: Since we use the capitalize logic twice in here maybe worth refactoring the logic to a separate line.

zabetak (approved):
Overall LGTM, left some minor comments/questions.

Jenkinsfile
snuyanzin: The problem with `gradle.properties` is that gradle tries to resolve all dependencies from there and during resolving of sonarqube plugin 4.x for jdk8 it fails(since it requires a newer jdk) as 
```
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not resolve org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:4.4.1.3373.
     Required by:
         project : > org.sonarqube:org.sonarqube.gradle.plugin:4.4.1.3373
      > No matching variant of org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:4.4.1.3373 was found. The consumer was configured to find a library for use during runtime, compatible with Java 8, packaged as a jar, and its dependencies declared externally, as well as attribute 'org.gradle.plugin.api-version' with value '8.7' but:
          - Variant 'apiElements' declares a library, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares a component for use during compile-time, compatible with Java 11 and the consumer needed a component for use during runtime, compatible with Java 8
              - Other compatible attribute:
                  - Doesn't say anything about org.gradle.plugin.api-version (required '8.7')
          - Variant 'javadocElements' declares a component for use during runtime, and its dependencies declared externally:
              - Incompatible because this component declares documentation and the consumer needed a library
              - Other compatible attributes:
                  - Doesn't say anything about its elements (required them packaged as a jar)
                  - Doesn't say anything about its target Java version (required compatibility with Java 8)
                  - Doesn't say anything about org.gradle.plugin.api-version (required '8.7')
          - Variant 'runtimeElements' declares a library for use during runtime, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares a component, compatible with Java 11 and the consumer needed a component, compatible with Java 8
              - Other compatible attribute:
                  - Doesn't say anything about org.gradle.plugin.api-version (required '8.7')
          - Variant 'sourcesElements' declares a component for use during runtime, and its dependencies declared externally:
              - Incompatible because this component declares documentation and the consumer needed a library
              - Other compatible attributes:
                  - Doesn't say anything about its elements (required them packaged as a jar)
                  - Doesn't say anything about its target Java version (required compatibility with Java 8)
                  - Doesn't say anything about org.gradle.plugin.api-version (required '8.7')

```

build.gradle.kts
snuyanzin: yep, sure, I also updated howto page with steps to upgrade gradle

core/build.gradle.kts
snuyanzin: done

gradle.properties
snuyanzin: finally was able to make it working without this property

build.gradle.kts
zabetak: Awesome, thanks!

Jenkinsfile
zabetak: If we cannot easily exclude the plugin for JDK8 then the changes here are good with me.

PR #3594, opened 2023-12-22
[CALCITE-6172] Allow aliased operators to re-use existing tests (PoC)

tanclary:
slightly different approach from Sergey's with a helper that makes it easy to retrieve the libraries that support an operator without having to iterate through all of them.

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -45,6 +45,7 @@ import org.apache.calcite.sql.SqlOperandCountRange;
 import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.SqlSyntax;
 import org.apache.calcite.sql.dialect.AnsiSqlDialect;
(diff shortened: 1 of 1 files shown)


mihaibudiu:
There is another PR for this issue #3590

tanclary:
> There is another PR for this issue #3590

@mihaibudiu Yes I know, this is just a different PoC showing a different approach.

mihaibudiu (commented):
This looks fine. I would do a quick code inspection to see what other cases this can apply to to make sure this pattern is fine.

The main downside is that the for loop is written manually compared with the parameterized tests approach. But it's not much code.

snuyanzin:
I like the idea of retrieving sql libraries, also applied it for parameterized tests

agree about downside of the loop especially it could be annoyng while debugging a specific case

tanclary:
@serge

> I like the idea of retrieving sql libraries, also applied it for parameterized tests
> 
> agree about downside of the loop especially it could be annoyng while debugging a specific case

Yeah I agree too. That was probably the main contribution so I'm going to close this.

PR #3596, opened 2023-12-24
[MINOR] Add checkNull Test on SqlOperatorTest

caicancai:


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -12376,6 +12376,9 @@ void testTimestampDiff(boolean coercionEnabled) {
     f.checkScalar("FORMAT_DATE('The date is: %x', DATE '2008-12-25')",
         "The date is: 12/25/08",
         "VARCHAR(2000) NOT NULL");
+    f.checkNull("FORMAT_DATE('%x', CAST(NULL AS DATE))");
```
tanclary, line 12379: If there aren't tests where the format itself is null we should add some of those as well while we're at it/

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -12439,6 +12442,9 @@ void testTimestampDiff(boolean coercionEnabled) {
     f.checkScalar("PARSE_DATE('%F', '2022-06-01')",
         "2022-06-01",
         "DATE NOT NULL");
+    f.checkNull("PARSE_DATE('%F', CAST(NULL AS DATE))");
```
tanclary, line 12445: same comment as above

tanclary (changes requested):

tanclary (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: Hello, thank you for your review
```java
f.checkNull("FORMAT_DATE(NULL, DATE '2008-12-25'");
```
This way of writing shows parser error. Do I need to add comments?

tanclary: It depends what the anticipated behavior is. If a parser error is expected then that makes sense. If we expect a `null` result then something is wrong. Do you know what the expected behavior should be if a null format is passed?

tanclary: I ran a simple test against my BQ database (`SELECT FORMAT_DATE(null, CURRENT_DATE())`) and it returned `null` so I would expect this to return`null` as well unless other dialects behave differently.

JiajunBernoulli:
LGTM

Thanks for your PR @caicancai 
Thanks for your review @tanclary

caicancai:
@JiajunBernoulli Thank you for your review. Regarding the issue I discussed with Clary above, should I open a Jira ticket to discuss it? I can try to fix it (if it is a bug)

JiajunBernoulli:
@caicancai 
I don't think it's necessary.

No matter what dialect, return null is a reasonable behavior for me.

PR #3597, opened 2023-12-25
[CALCITE-6162] Add rule(s) to remove joins with constant single tuple…

HanumathRao:
… relations

This PR introduces four rules for simplifying joins with a single value row on either side of the input. Please find the information about this optimizations in the related [JIRA](https://issues.apache.org/jira/browse/CALCITE-6162).
**Testing**: Apart from the new tests added in RelOptRulesTest, these rules are enabled as part of the abstract rules and so they get triggered during the core quidem test suite.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/SingleValuesOptimizationRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SingleValuesOptimizationRules.java
@@ -1,0 +1,397 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRule;
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Join;
+import org.apache.calcite.rel.core.JoinRelType;
+import org.apache.calcite.rel.core.Project;
+import org.apache.calcite.rel.core.Values;
+import org.apache.calcite.rel.logical.LogicalValues;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexInputRef;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexShuttle;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.tools.RelBuilder;
+import org.apache.calcite.util.ImmutableBitSet;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.immutables.value.Value;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.BiFunction;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
+ * Collection of rules which simplify joins which have one of their input as constant relations
+ * {@link Values} that produce a single row.
+ *
+ * <p>Conventionally, the way to represent a single row constant relational expression is
+ * with a {@link Values} that has one tuple.
+ *
+ * @see LogicalValues#createOneRow
+ */
+public abstract class SingleValuesOptimizationRules {
+
+  public static final RelOptRule JOIN_LEFT_INSTANCE =
+      SingleValuesOptimizationRules.JoinLeftSingleRuleConfig.DEFAULT.toRule();
+
+  public static final RelOptRule JOIN_RIGHT_INSTANCE =
+      SingleValuesOptimizationRules.JoinRightSingleRuleConfig.DEFAULT.toRule();
+
+  public static final RelOptRule JOIN_LEFT_PROJECT_INSTANCE =
+      SingleValuesOptimizationRules.JoinLeftSingleValueRuleWithExprConfig.DEFAULT.toRule();
+
+  public static final RelOptRule JOIN_RIGHT_PROJECT_INSTANCE =
+      SingleValuesOptimizationRules.JoinRightSingleValueRuleWithExprConfig.DEFAULT.toRule();
+
+  /**
+   * Transformer class to transform a Join rel node tree with single constant row {@link Values}
+   * on either side of the Join to a simplified tree without a Join rel node.
+   */
+  private static class SingleValuesRelTransformer {
+
+    private final Join join;
+    private final RelNode relNode;
+    private final Predicate<Join> transformable;
+    private final BiFunction<RexNode, List<RexNode>, List<RexNode>> litTransformer;
+    private final boolean valuesAsLeftChild;
+    private final List<RexNode> literals;
+
+    /**
+     * A transformer object which transforms a Join rel node tree with constant relation
+     * node as one of its input to a rel node tree without a Join.
+     *
+     * @param join Join which is eligible for removal.
+     * @param rexNodes List of the expressions that are part of Project
+     * @param otherNode RelNode which is other side of the Join (apart from Values node)
+     * @param transformable Predicate to check if the given Join is transformable or not.
+     * @param isValuesLeftChild TRUE if Values is left child of join, FALSE otherwise.
+     * @param litTransformer A transformer function supplied by the caller.
+     *                       This function is specific to Join Type.
+     *                       LEFT/ RIGHT => has logic to produce null values for unmatched rows.
+     *                       INNER => produce the rexLiterals specified in the Values node.
+     */
+    protected SingleValuesRelTransformer(
+        Join join, List<RexNode> rexNodes, RelNode otherNode,
+        Predicate<Join> transformable, boolean isValuesLeftChild,
+        BiFunction<RexNode, List<RexNode>, List<RexNode>> litTransformer) {
+      this.relNode = otherNode;
+      this.join = join;
+      this.transformable = transformable;
+      this.litTransformer = litTransformer;
+      this.valuesAsLeftChild = isValuesLeftChild;
+      this.literals = rexNodes;
+    }
+
+    /**
+     * A transform function which simplifies the join tree when eligibility criteria is met.
+     * Refer to {@link PruneSingleValueRule#onMatch} function for details about the transformation.
+
+     * Another very subtle point is that this works even if the VALUES contain NULL entries.
+     * This is because this generates code with comparisons to NULL, and comparing a value to NULL
+     * using = always returns false, so the result is an empty collection, as expected.
+     *
+     * @param relBuilder Relation Builder supplied by the planner framework.
+     * @return Simplified relNode tree by replacing Join,
+     *          Otherwise returns null when criteria is not met.
+     */
+    public @Nullable RelNode transform(RelBuilder relBuilder) {
+      if (!transformable.test(join)) {
+        return null;
+      }
+      int end = valuesAsLeftChild
+          ? join.getLeft().getRowType().getFieldCount()
+          : join.getRowType().getFieldCount();
+
+      int start = valuesAsLeftChild
+          ? 0
+          : join.getLeft().getRowType().getFieldCount();
+      ImmutableBitSet bitSet = ImmutableBitSet.range(start, end);
+      RexNode trueNode = relBuilder.getRexBuilder().makeLiteral(true);
+      final RexNode filterCondition =
+          new RexNodeReplacer(bitSet,
+              literals,
+              (valuesAsLeftChild ? 0 : -1) * join.getLeft().getRowType().getFieldCount())
+              .go(join.getCondition());
+
+      RexNode fixedCondition =
+          valuesAsLeftChild
+              ? RexUtil.shift(filterCondition,
+              -1 * join.getLeft().getRowType().getFieldCount())
+              : filterCondition;
+
+      List<RexNode> rexLiterals = litTransformer.apply(fixedCondition, literals);
+      relBuilder.push(relNode)
+          .filter(join.getJoinType().isOuterJoin() ? trueNode : fixedCondition);
+
+      List<RexNode> rexNodes = relNode
+          .getRowType()
+          .getFieldList()
+          .stream()
+          .map(fld -> relBuilder.field(fld.getIndex()))
+          .collect(Collectors.toList());
+
+      List<RexNode> projects = new ArrayList<>();
+      projects.addAll(valuesAsLeftChild ? rexLiterals : rexNodes);
+      projects.addAll(valuesAsLeftChild ? rexNodes : rexLiterals);
+      return relBuilder.project(projects).build();
+    }
+  }
+
+  /**
+   * A rex shuttle to replace field refs with supplied rexNodes from a {@link Values} row.
+   */
+  private static class RexNodeReplacer extends RexShuttle {
+
+    private final ImmutableBitSet bitSet;
+    private final List<RexNode> fieldValues;
+    private final int offset;
+
+    /**
+     * A RexNode replacer which replaces an inputRef with a corresponding
+     * RexNode value supplied in values parameter of the constructor.
+     *
+     * @param bitSet A bitmap to track indices of the inputRef that get replaced.
+     * @param values RexNodes list that are used to replace inputRefs.
+     * @param offset offset to be applied for the inputRef index to get the corresponding
+     *              RexNode value.
+     */
+    RexNodeReplacer(ImmutableBitSet bitSet, List<RexNode> values, int offset) {
+      this.bitSet = bitSet;
+      this.fieldValues = values;
+      this.offset = offset;
+    }
+    @Override public RexNode visitInputRef(RexInputRef inputRef) {
+      if (bitSet.get(inputRef.getIndex())) {
+        return this.fieldValues.get(inputRef.getIndex() + offset);
+      }
+      return super.visitInputRef(inputRef);
+    }
+
+    public RexNode go(RexNode expression) {
+      return expression.accept(this);
+    }
+  }
+
+  /**
+   * Abstract class for all the SingleValuesOptimizationRules. All the specific
+   * SingleValuesOptimizationRules call the onMatch function to replace a Join
+   * with equivalent tree when eligible.
+   */
+  protected abstract static class PruneSingleValueRule
+      extends RelRule<PruneSingleValueRule.Config>
+      implements SubstitutionRule {
+    protected PruneSingleValueRule(PruneSingleValueRule.Config config) {
+      super(config);
+    }
+
+    protected BiFunction<RexNode, List<RexNode>, List<RexNode>>
+        getRexTransformer(RexBuilder rexBuilder,
+        JoinRelType joinRelType) {
+      switch (joinRelType) {
+      case LEFT:
+      case RIGHT:
+        return (condition, rexLiterals) -> rexLiterals.stream().map(lit ->
+            rexBuilder.makeCall(SqlStdOperatorTable.CASE, condition,
+                lit, rexBuilder.makeNullLiteral(lit.getType()))).collect(Collectors.toList());
+      default:
+        return (condition, rexLiterals) -> rexLiterals;
+      }
+    }
+
+    /**
+     * onMatch function contains common optimization logic for all the
+     * SingleValueOptimization rules. It simplifies the rel node tree by
+     * removing a Join node and creating a required filter as applicable.
+     * In case of the LEFT/RIGHT joins, a case expression which produce NULL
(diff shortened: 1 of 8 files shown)


core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -4177,6 +4178,109 @@ RelOptFixture checkDynamicFunctions(boolean treatDynamicCallsAsConstant) {
         .check();
   }
 
+  @Test void testJoinWithConstantRowOnRight() {
```
mihaibudiu, line 4181: How about a test where some of the `VALUE`s are NULL?

HanumathRao: These tests are designed to validate the generated plans for a specified query. I've incorporated output oriented tests into the join.iq file, and couple of them generate null values in the context of a left outer join. If you think any test cases are missing in join.iq, please let me know, and I'll be sure to add them. Thank you.

mihaibudiu: equi-joining with a tuple that has a null value for one of the ON fields should produce an empty set. Is there a test about that?

HanumathRao: I have added two test cases with null values as input. Please take a look and let me know. Thank you.

mihaibudiu: I meant a test like this one: `select e.empno, e.ename, c.* from emp e cross join (select 5, NULL) c`
This test seems to produce an incorrect plan because you just substitute the NULL for the corresponding column:

```
LogicalProject(EMPNO=[$0], ENAME=[$1], EXPR$0=[$9], EXPR$1=[$10])
  LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], $f9=[5], $f10=[null:NULL])
```

I believe that the result should be the empty set.

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,361 @@
...
+          }
+        }
+      };
+    }
+  }
+
+  /** Configuration for a rule that simplifies join node with a project on a constant row
+   *  on its left input. */
+  @Value.Immutable
+  interface JoinLeftSingleValueRuleWithExprConfig extends PruneSingleValueRule.Config {
```
JiajunBernoulli, line 275: `JoinLeftSingleValueRuleWithExprConfig` and `JoinRightSingleValueRuleWithExprConfig` has many same code.

Can we extract them?

macroguo-ghy: Agree.

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,361 @@
...
+                b1 -> b1.operand(RelNode.class).anyInputs(),
+                b2 -> b2.operand(Project.class).inputs(
+                    b21 -> b21.operand(Values.class).predicate(Values::isSingleValue).noInputs())))
+        .withDescription("PruneJoinSingleValueRuleWithExpr(right)");
+
+    @Override default SingleValueOptimizationRules.PruneSingleValueRule toRule() {
+      return new SingleValueOptimizationRules.PruneSingleValueRule(this) {
+        @Override public void onMatch(RelOptRuleCall call) {
+          Predicate<Join> predicate = elibilityPredicate(false);
+          Join join = call.rel(0);
```
macroguo-ghy, line 333: Use `final` modifier

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
HanumathRao: @mihaibudiu  Thanks for the review, the above query doesn't result in an empty subset as it is a cross join. I have added the test case and made sure that the results are same with and without the optimization (by removing the rules  and executing the test case).

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
HanumathRao: Thanks @JiajunBernoulli  and @macroguo-ghy  for the suggestion. I moved the redundant code to onMatch base class function.

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,361 @@
...
+    private Predicate<Join> cannotTransform;
+    private BiFunction<RexNode, List<RexNode>, List<RexNode>> litTransformer;
+    private boolean valuesAsLeftChild;
+    private List<RexNode> literals;
+
+    protected SingleValueRelTransformer(
+        Join join, List<RexNode> rexNodes, RelNode otherNode,
+        Predicate<Join> nonTransformable, boolean isValuesLeftChild,
+        BiFunction<RexNode, List<RexNode>, List<RexNode>> litTransformer) {
+      this.relNode = otherNode;
```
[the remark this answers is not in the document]
HanumathRao: @macroguo-ghy  sorry didn't quite get this. If the parameters don't match then it will result in compilation failure right?

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,361 @@
...
+      case RIGHT:
+        return (condition, rexLiterals) -> rexLiterals.stream().map(lit ->
+            rexBuilder.makeCall(SqlStdOperatorTable.CASE, condition,
+                lit, rexBuilder.makeNullLiteral(lit.getType()))).collect(Collectors.toList());
+      default:
+        return (condition, rexLiterals) -> rexLiterals.stream().collect(Collectors.toList());
+      }
+    }
+
+    static final Predicate<Join> elibilityPredicate(boolean isLeft) {
```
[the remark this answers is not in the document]
HanumathRao: I thought about it but left it for now as I plan to add explore for semi join and anti join in a different JIRA.

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,329 @@
...
+import java.util.List;
+import java.util.function.BiFunction;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
+ * Collection of rules which simplify sections of query plan which are known to
+ * produce single row.
+ *
+ * <p>Conventionally, the way to represent a single row relational expression is
```
mihaibudiu, line 50: Values can represent only constant relations. So this comment is not entirely accurate.
I think a more precise JavaDoc comment would be that these rules handle joins where one of the relations is a constant relation (VALUES) with a single row. The existing comment implies that you are handling all cases that can generate a single row.

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,329 @@
...
+  public static final RelOptRule JOIN_LEFT_INSTANCE =
+      SingleValueOptimizationRules.JoinLeftSingleRuleConfig.DEFAULT.toRule();
+
+  public static final RelOptRule JOIN_RIGHT_INSTANCE =
+      SingleValueOptimizationRules.JoinRightSingleRuleConfig.DEFAULT.toRule();
+
+  public static final RelOptRule JOIN_LEFT_PROJECT_INSTANCE =
+      SingleValueOptimizationRules.JoinLeftSingleValueRuleWithExprConfig.DEFAULT.toRule();
+
+  public static final RelOptRule JOIN_RIGHT_PROJECT_INSTANCE =
```
mihaibudiu, line 66: Can you add a Javadoc comment on these describing how they differ from each other?

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,329 @@
...
+      SingleValueOptimizationRules.JoinRightSingleValueRuleWithExprConfig.DEFAULT.toRule();
+
+  /**
+   * Transformer class to transform a single value nodes on either side of the join.
```
mihaibudiu, line 70: you should probably capitalize VALUES, since it's a SQL construct.
Does "single" here mean "with a single tuple"?

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,329 @@
...
+      RelNode transformed =
+          new SingleValueRelTransformer(join, rexNodes, other,
+              predicate, isLeft, getRexTransformer(relBuilder.getRexBuilder(),
+              join.getJoinType())).transform(relBuilder);
+      if (transformed != null) {
+        call.transformTo(transformed);
+      }
+    }
+
+    static Predicate<Join> elibilityPredicate(boolean isLeft) {
```
mihaibudiu, line 214: the word is "eligibility".

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,329 @@
...
+        BiFunction<RexNode, List<RexNode>, List<RexNode>> litTransformer) {
+      this.relNode = otherNode;
+      this.join = join;
+      this.cannotTransform = nonTransformable;
+      this.litTransformer = litTransformer;
+      this.valuesAsLeftChild = isValuesLeftChild;
+      this.literals = rexNodes;
+    }
+
+    public @Nullable RelNode transform(RelBuilder relBuilder) {
```
mihaibudiu, line 94: I would give here a more elaborate comment. What happens is that a JOIN is converted into a FILTER where the filter predicate uses the join condition. I would in fact give a small example of a plan before and after.

Another very subtle point is that this works even if the VALUES contain NULL entries. This is because this generates code with comparisons to NULL, and comparing a value to NULL using `=` always returns `false`, so the result is an empty collection, as expected.

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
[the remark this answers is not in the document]
mihaibudiu: More importantly, I think you should add JavaDoc for the parameters describing what they mean.

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,329 @@
...
+
+  /**
+   * A rex shuttle to replace field refs with constants from a {@link Values} row.
+   */
+  private static class ReplaceExprWithConstValue extends RexShuttle {
+
+    private final ImmutableBitSet bitSet;
+    private final List<RexNode> fieldValues;
+    private final int offset;
+    ReplaceExprWithConstValue(ImmutableBitSet bitSet, List<RexNode> values, int offset) {
```
mihaibudiu, line 145: documenting parameters would help read the code.

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,329 @@
...
+      case RIGHT:
+        return (condition, rexLiterals) -> rexLiterals.stream().map(lit ->
+            rexBuilder.makeCall(SqlStdOperatorTable.CASE, condition,
+                lit, rexBuilder.makeNullLiteral(lit.getType()))).collect(Collectors.toList());
+      default:
+        return (condition, rexLiterals) -> new ArrayList<>(rexLiterals);
+      }
+    }
+
+    protected void onMatch(RelOptRuleCall call, Values values,
```
mihaibudiu, line 186: please document these parameters as well. It's not clear what "project" for example is.

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
```diff
@@ -0,0 +1,329 @@
...
+          final Join join = call.rel(0);
+          final Values values = call.rel(1);
+          final RelNode other = call.rel(2);
+          onMatch(call, values, null, join, other, true);
+        }
+      };
+    }
+  }
+
+  /** Configuration for a rule that simplifies join node with a project on a constant row
```
mihaibudiu, line 280: what is a project on a constant row? Is it a projection applied to a VALUES node?
Isn't that case handled by a separate optimization step?

mihaibudiu: This should be a comment in the code.

HanumathRao:
@mihaibudiu  Thanks for the suggestion. I have added the javadoc comments wherever I think are needed for clarity. Regarding the comments for the rules, I have already added comments for the rule configs so I think it might be redundant to add them again for the rules. Please take a look and let me know if there is anything not clear or missing.

core/src/main/java/org/apache/calcite/rel/rules/SingleValuesOptimizationRules.java
```diff
@@ -0,0 +1,374 @@
...
+        RexShuttle shuttle =
+            new ReplaceExprWithConstValue(bitSet,
+                    new ArrayList<>(values.getTuples().get(0)),
+                0);
+
+        rexNodes = project.getProjects().stream()
+            .map(shuttle::apply)
+            .collect(Collectors.toList());
+      } else {
+        rexNodes = new ArrayList<>(values.tuples.get(0));
```
mihaibudiu, line 245: Maybe you can factor out this as a helper function.
You previously used getTuples(), here you just used tuples.
This could be called getSingle(Values) and could also assert that values has a single tuple.

core/src/main/java/org/apache/calcite/rel/rules/SingleValuesOptimizationRules.java
```diff
@@ -0,0 +1,374 @@
...
+      }
+      RelBuilder relBuilder = call.builder();
+      RelNode transformed =
```
mihaibudiu, line 248: I find that it's more convenient for debugging to have one statement per line.
So I would make a separate statement with the new and one with 'transform'.

core/src/main/java/org/apache/calcite/rel/rules/SingleValuesOptimizationRules.java
```diff
@@ -0,0 +1,374 @@
...
+              predicate, isLeft, getRexTransformer(relBuilder.getRexBuilder(),
+              join.getJoinType())).transform(relBuilder);
+      if (transformed != null) {
+        call.transformTo(transformed);
+      }
+    }
+
+    static Predicate<Join> eligibilityPredicate(boolean isLeft) {
+      if (isLeft) {
+        return jn -> jn.getJoinType() == JoinRelType.LEFT
```
mihaibudiu, line 259: This may be clearer if you factor out the common case outside the 'if'.
One other thing that is confusing is that this is actually the "nonEligibilityPredicate".
Perhaps it would be clearer if you actually complement all definitions and uses of this predicate.
Then you can call the previous field "canTransform".

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
mihaibudiu: Can you please add a comment in the code containing the second paragraph in my comment? I had to run in the debugger to understand why this happens.

core/src/main/java/org/apache/calcite/rel/rules/SingleValueOptimizationRules.java
mihaibudiu: please add this as a comment in the code too.

mihaibudiu (commented):
The code looks pretty good, my comments are about readability mostly.
I am not familiar with this part of the codebase; perhaps a more experienced reviewer would understand faster.
But in general I really prefer the code to be as simple and clear as possible. It helps everyone: reviewers, people who debug it, people who edit it in the future. Some of my suggestions may be not strictly necessary, like reverting the negated variables "cannotTransform", but I think they do really help.
Some of the comments you made in the PR should probably be in the code as well; the PR will be lost, but the code stays.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -2973,27 +2973,29 @@ void testInnerJoinValues(String format) {
...
+          + "...\" -> \"EnumerableAggregate\\n"
+          + "group = {0, 3}\\n"
+          + "\" [label=\"0\"]\n"
+          + "\"EnumerableTableScan\\n"
+          + "table = [SALES, EMPS\\n]\\n"
+          + "\" -> \"EnumerableCalc\\n"
+          + "expr#0..1 = {inputs}\\n"
+          + "expr#2 = 10\\n"
+          + "expr#3 = 'SameName'\\n"
+          + "expr#4 = CAST($t1):I\\nNTEGER NOT NULL\\n"
```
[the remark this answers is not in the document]
HanumathRao: I believe it is due to formatting (when dot format is used). There are other cases in the same file where \\n was in middle of classname or fieldname 
e.g. ""EnumerableTableScan\ntable" and "$t1\n0"

core/src/main/java/org/apache/calcite/rel/rules/SingleValuesOptimizationRules.java
HanumathRao: I changed the code to use tuples in both the cases. I haven't created a helper function here because it might be an overkill.

core/src/main/java/org/apache/calcite/rel/rules/SingleValuesOptimizationRules.java
```diff
@@ -0,0 +1,374 @@
...
+        getRexTransformer(RexBuilder rexBuilder,
+        JoinRelType joinRelType) {
+      switch (joinRelType) {
+      case LEFT:
+      case RIGHT:
+        return (condition, rexLiterals) -> rexLiterals.stream().map(lit ->
+            rexBuilder.makeCall(SqlStdOperatorTable.CASE, condition,
+                lit, rexBuilder.makeNullLiteral(lit.getType()))).collect(Collectors.toList());
+      default:
+        return (condition, rexLiterals) -> new ArrayList<>(rexLiterals);
```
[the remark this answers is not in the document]
HanumathRao: No need to copy.

HanumathRao:
Not sure why the tests are showing failed in the PR but all of them are successful on the branch commit.
https://github.com/HanumathRao/calcite/commits/CALCITE-6162/, build run https://github.com/HanumathRao/calcite/actions/runs/7521737012/job/20472927213

HanumathRao:
> The code looks pretty good, my comments are about readability mostly. I am not familiar with this part of the codebase; perhaps a more experienced reviewer would understand faster. But in general I really prefer the code to be as simple and clear as possible. It helps everyone: reviewers, people who debug it, people who edit it in the future. Some of my suggestions may be not strictly necessary, like reverting the negated variables "cannotTransform", but I think they do really help. Some of the comments you made in the PR should probably be in the code as well; the PR will be lost, but the code stays.

Thank you for your review comments. I have addressed the comments and provided inline replies where necessary.

core/src/main/java/org/apache/calcite/rel/rules/SingleValuesOptimizationRules.java
```diff
@@ -95,24 +95,26 @@ private static class SingleValuesRelTransformer {
      */
     protected SingleValuesRelTransformer(
         Join join, List<RexNode> rexNodes, RelNode otherNode,
-        Predicate<Join> nonTransformable, boolean isValuesLeftChild,
+        Predicate<Join> tranfsormable, boolean isValuesLeftChild,
```
mihaibudiu, line 98: parameter name is mistyped "sf"

core/src/main/java/org/apache/calcite/rel/rules/SingleValuesOptimizationRules.java
```diff
@@ -245,24 +256,29 @@ protected void onMatch(RelOptRuleCall call, Values values,
...
       RelBuilder relBuilder = call.builder();
+
+      BiFunction<RexNode, List<RexNode>, List<RexNode>> transformer =
+          getRexTransformer(relBuilder.getRexBuilder(), join.getJoinType());
+
       RelNode transformed =
           new SingleValuesRelTransformer(join, rexNodes, other,
-              predicate, isLeft, getRexTransformer(relBuilder.getRexBuilder(),
-              join.getJoinType())).transform(relBuilder);
+              transformableCheck, isOnLefSide, transformer).transform(relBuilder);
```
mihaibudiu, line 269: At least with IntelliJ I have trouble stepping in "transform" here.
So I would even allocate a temp variable for the transformer and invoke `transform` in a separate statement.

mihaibudiu (approved):

mihaibudiu:
I have approved, but I still made a few small suggestions.
It looks like the CI is unhappy.

HanumathRao:
> I have approved, but I still made a few small suggestions. It looks like the CI is unhappy.

Thanks. I am not sure why only PR runs are showing this issue. The tests executed on the commit and also same commit on another branch are always successful. I am guessing that once I squash all the commits and rebase this problem might go away.

mihaibudiu:
I have no further comments. Please squash the commits so we can merge this PR. Thank you for the improvements!

HanumathRao:
> I have no further comments. Please squash the commits so we can merge this PR. Thank you for the improvements!

Done. Thanks for review.

PR #3598, opened 2023-12-25
[CALCITE-6178] WITH RECURSIVE query when cloned using sqlshuttle loos…

HanumathRao:
…es RECURSIVE property

Fix the getters and setters of SqlWithItem to pass recursive flag as one of the operand.

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
+++ b/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
@@ -43,6 +43,7 @@ import org.apache.calcite.test.IntervalTest;
 import org.apache.calcite.tools.Hoist;
 import org.apache.calcite.util.Bug;
 import org.apache.calcite.util.ConversionUtil;
+import org.apache.calcite.util.Litmus;
 import org.apache.calcite.util.Pair;
 import org.apache.calcite.util.TestUtil;
 import org.apache.calcite.util.Util;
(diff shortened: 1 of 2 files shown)


testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -1233,6 +1235,17 @@ protected static SortedSet<String> keywords(@Nullable String dialect) {
         .fails("Bang equal '!=' is not allowed under the current SQL conformance level");
   }
 
+  @Test void testRecursiveQueryCloned() throws Exception {
```
mihaibudiu, line 1238: May be useful to add a Javadoc comment here with the issue that is being solved.

mihaibudiu (approved):

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
HanumathRao: Thanks @mihaibudiu for the review, added a Javadoc comment to the test case.

mihaibudiu:
If no one else has comments I think we should merge this PR

mihaibudiu:
@HanumathRao if you can squash the commits we can merge this PR. Thank you.

HanumathRao:
> @HanumathRao if you can squash the commits we can merge this PR. Thank you.

Done. Thank you.

PR #3599, opened 2023-12-26
[MINOR] Add FormatElementEnumTests

caicancai:
I found that there are some format functions that can be supported. If you want to consider supporting them, I can add this pr, e.g: WEEK_OF_MONTH

### DIFF
--- a/core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
+++ b/core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
@@ -33,4 +33,148 @@ class FormatElementEnumTest {
     FormatElementEnum.DAY.format(ts, Date.from(Instant.parse("2014-09-30T10:00:00Z")));
     assertThat(ts, hasToString("Tuesday"));
   }
+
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
```diff
@@ -33,4 +33,148 @@ class FormatElementEnumTest {
...
+  @Test void TestMON(){
+    StringBuilder ts = new StringBuilder();
+    FormatElementEnum.MON.format(ts, Date.from(Instant.parse("2014-09-30T10:00:00Z")));
+    assertThat(ts, hasToString("Sep"));
+  }
+
+  @Test void TestMONTH(){
+    StringBuilder ts = new StringBuilder();
+    FormatElementEnum.MONTH.format(ts, Date.from(Instant.parse("2014-06-30T10:00:00Z")));
+    assertThat(ts, hasToString("Sep"));
```
caicancai, line 136: This seems to be a bug, the full month name is not displayed

PR #3600, opened 2023-12-26
[CALCITE-6179] Support weekofmonth function format and add test

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6179

### DIFF
--- a/core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
+++ b/core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
@@ -201,6 +201,14 @@ public enum FormatElementEnum implements FormatElement {
       // TODO: how to support timezones?
     }
   },
+  W("W", "The week number of the month (Sunday as the first day of the week) as a decimal "
+      + "number (1-5)") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      sb.append(String.format(Locale.ROOT, "%d", calendar.get(Calendar.WEEK_OF_MONTH)));
+    }
+  },
   WW("w", "The week number of the year (Sunday as the first day of the week) as a decimal "
       + "number (00-53)") {
     @Override public void format(StringBuilder sb, Date date) {
--- a/core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
+++ b/core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
@@ -29,8 +29,149 @@ import static org.hamcrest.Matchers.hasToString;
  */
 class FormatElementEnumTest {
   @Test void testDay() {
+    final String date = "2014-09-30T10:00:00Z";
     StringBuilder ts = new StringBuilder();
-    FormatElementEnum.DAY.format(ts, Date.from(Instant.parse("2014-09-30T10:00:00Z")));
+    FormatElementEnum.DAY.format(ts, Date.from(Instant.parse(date)));
     assertThat(ts, hasToString("Tuesday"));
+  }
+
+  @Test void testD() {
+    final String date = "2014-09-30T10:00:00Z";
+    StringBuilder ts = new StringBuilder();
+    FormatElementEnum.D.format(ts, Date.from(Instant.parse(date)));
+    assertThat(ts, hasToString("3"));
+  }
+
+  @Test void testDD() {
+    final String date = "2014-09-30T10:00:00Z";
+    StringBuilder ts = new StringBuilder();
+    FormatElementEnum.DD.format(ts, Date.from(Instant.parse(date)));
+    assertThat(ts, hasToString("30"));
+  }
+
+  @Test void testDDD() {
+    final String date = "2014-09-30T10:00:00Z";
+    StringBuilder ts = new StringBuilder();
+    FormatElementEnum.DDD.format(ts, Date.from(Instant.parse(date)));
+    assertThat(ts, hasToString("273"));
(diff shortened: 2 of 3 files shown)


caicancai:
Is the result of format affected by the jdk version?

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
```diff
@@ -33,4 +33,130 @@ class FormatElementEnumTest {
     FormatElementEnum.DAY.format(ts, Date.from(Instant.parse("2014-09-30T10:00:00Z")));
     assertThat(ts, hasToString("Tuesday"));
   }
+
+  @Test void testD() {
+    StringBuilder ts = new StringBuilder();
+    FormatElementEnum.D.format(ts, Date.from(Instant.parse("2014-09-30T10:00:00Z")));
```
mihaibudiu, line 39: Since the date is the same for all tests maybe you can factor it out in a separate helper function.
This would also enable extending this testing code for other dates.

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
```diff
@@ -33,4 +33,130 @@ class FormatElementEnumTest {
...
+    assertThat(ts, hasToString("Tue"));
+  }
+
+  @Test void testFF1() {
+    StringBuilder ts = new StringBuilder();
+    FormatElementEnum.FF1.format(ts, Date.from(Instant.parse("2014-09-30T10:00:00Z")));
+    assertThat(ts, hasToString("0"));
+  }
+
+  @Test void testFF2() {
```
mihaibudiu, line 67: if these are microseconds perhaps you can use a datetime with microseconds?
Does Java support microseconds in Date?

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
caicancai: I'm trying, and I seem to have found some problems. If there are too many problems, I will open a Jira and try to fix them in this PR.

macroguo-ghy:
You need to log a JIRA case, because you have added new features.

caicancai:
> You need to log a JIRA case, because you have added new features.

Thank you for your reply

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
caicancai: I didn't find the corresponding API

mihaibudiu: What API didn't you find?
It looks like Instant.parse supports microseconds and date supports milliseconds.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -201,6 +201,14 @@ public enum FormatElementEnum implements FormatElement {
       // TODO: how to support timezones?
     }
   },
+  W("w", "The week number of the month (Sunday as the first day of the week) as a decimal "
```
macroguo-ghy, line 204: I think the javaFormat of `week of month` should be `W` instead of `w`.

macroguo-ghy (changes requested):
Utilities for FormatModel in `org.apache.calcite.util.format.FormatModels` should be taken into consideration.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
caicancai: done

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
caicancai: Sorry, I misunderstood the meaning before, but I have corrected it now

caicancai:
@macroguo-ghy If you have time, can you help review my code? Thank you.

macroguo-ghy:
@caicancai Did you miss my previous comment? I think you should add `week of month` to Format model for PostgreSQL. Please refer org.apache.calcite.util.format.FormatModels#POSTGRESQL and https://www.postgresql.org/docs/current/functions-formatting.html.

caicancai:
> @caicancai Did you miss my previous comment? I think you should add `week of month` to Format model for PostgreSQL. Please refer org.apache.calcite.util.format.FormatModels#POSTGRESQL and https://www.postgresql.org/docs/current/functions-formatting.html.

@macroguo-ghy I didn't seem to understand it before, I'm sorry about that, I understand it now, thank you for your reivew

macroguo-ghy (approved):
+1

caicancai:
@macroguo-ghy @mihaibudiu Good to merge？

caicancai:
Thank you for the review

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
```diff
@@ -29,8 +29,149 @@
...
+    assertThat(ts, hasToString("000123"));
+  }
+
+  @Test void testIW() {
+    final String date = "2014-09-30T10:00:00Z";
+    StringBuilder ts = new StringBuilder();
+    FormatElementEnum.IW.format(ts, Date.from(Instant.parse(date)));
+    assertThat(ts, hasToString("40"));
+  }
+
```
XuQianJin-Stars, line 114: There are too many duplicate codes in the test, which can be optimized and merged.

PR #3601, opened 2023-12-27
CALCITE-6180 Append possibility to escape backslash in LIKE with ESCAPE operator

zstan:
fix possibility to execute:
`select 'Dev\ops' like 'Dev#\ops' escape '#';`


mihaibudiu (approved):

zstan:
@mihaibudiu thank you for review ! can you also update issue status ? 
seems it more correct if you change it ?

mihaibudiu:
Usually the issue status is updated automatically when you merge the commit.
However, I notice that your commit message does not match the issue message, perhaps that's why it wasn't automatically closed. I will change the issue message to match this one.

PR #3604, opened 2023-12-27
[CALCITE-6182] Add LENGTH/LEN functions (enabled in Snowflake library)

tanclary:
Enabling LENGTH for Snowflake and adding the LEN alias

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -6696,6 +6696,23 @@ class RelToSqlConverterTest {
     sql(query).withLibrary(SqlLibrary.SNOWFLAKE).withSnowflake().ok(expectedSnowflake);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6182">[CALCITE-6182]
+   * Add LENGTH/LEN functions (enabled in Snowflake library)</a>. */
+  @Test void testSnowflakeLength() {
+    final String query = "select CHAR_LENGTH(\"brand_name\")\n"
+        + "from \"product\"";
+    final String expectedBigQuery = "SELECT CHAR_LENGTH(brand_name)\n"
+        + "FROM foodmart.product";
+    // Snowflake would accept either LEN or LENGTH, but we currently unparse into "LENGTH"
+    // since it seems to be used across more dialects.
+    final String expectedSnowflake = "SELECT LENGTH(\"brand_name\")\n"
+        + "FROM \"foodmart\".\"product\"";
+    Sql sql = sql(query).withLibrary(SqlLibrary.BIG_QUERY);
+    sql.withBigQuery().ok(expectedBigQuery);
+    sql.withSnowflake().ok(expectedSnowflake);
+  }
+
   @Test void testSubstringInSpark() {
     final String query = "select substring(\"brand_name\" from 2) "
         + "from \"product\"\n";
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -4108,6 +4108,23 @@ public class SqlOperatorTest {
     f.checkNull("CHARACTER_LENGTH(cast(null as varchar(1)))");
   }
 
+  /** Tests {@code LEN} function from Snowflake. {@code LEN} is a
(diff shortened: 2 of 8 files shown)


mihaibudiu:
Should this wait for #3590?

tanclary:
> Should this wait for #3590?

Yes going to rebase once thats merged.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4108,6 +4108,19 @@ static void checkRlikeFails(SqlOperatorFixture f) {
     f.checkNull("CHARACTER_LENGTH(cast(null as varchar(1)))");
   }
 
+  @Test void testLenFunc() {
```
julianhyde, line 4111: Could use a comment - that LEN is a snowflake-specific alias for CHAR_LENGTH, and that snowflake also supports LENGTH

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4108,6 +4108,19 @@ static void checkRlikeFails(SqlOperatorFixture f) {
...
+    final SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.LENGTH);
+    f0.checkFails("^length('hello')^",
+        "No match found for function signature LENGTH\\(<CHARACTER>\\)",
+        false);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SNOWFLAKE);
+    f.checkScalar("length('hello')", "5", "INTEGER NOT NULL");
+    f.checkScalar("length('')", "0", "INTEGER NOT NULL");
+    f.checkScalar("length(CAST('x' as CHAR(3)))", "3", "INTEGER NOT NULL");
+    f.checkScalar("length(CAST('x' as VARCHAR(4)))", "1", "INTEGER NOT NULL");
+    f.checkNull("length(CAST(NULL as CHAR(5)))");
```
julianhyde, line 4121: None of these tests has a call to 'len(...)'. Maybe LEN is getting tested via some aliasing mechanism. It's difficult to tell. 

So, let's have at least one test for LEN. So we can still trust the test even if the clever stuff fails.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -6696,6 +6696,20 @@ private void checkLiteral2(String expression, String expected) {
...
+   * Add LENGTH/LEN functions (enabled in Snowflake library)</a>. */
+  @Test void testSnowflakeLength() {
+    final String query = "select CHAR_LENGTH(\"brand_name\")\n"
+        + "from \"product\"";
+    final String expectedBigQuery = "SELECT CHAR_LENGTH(brand_name)\n"
+        + "FROM foodmart.product";
+    final String expectedSnowflake = "SELECT LENGTH(\"brand_name\")\n"
+        + "FROM \"foodmart\".\"product\"";
+    sql(query).withLibrary(SqlLibrary.BIG_QUERY).withBigQuery().ok(expectedBigQuery);
+    sql(query).withLibrary(SqlLibrary.BIG_QUERY).withSnowflake().ok(expectedSnowflake);
```
julianhyde, line 6710: can you combine those lines? you only need to parse the query once. then you generate SQL for two dialects.

the structure of the test should make it clear that people can other target dialects too.

I don't see any mention of LEN in the test. Not that there should be. It's ok to generate LENGTH for Snowflake. But it needs explanation, given that the jira case mentions LEN.

julianhyde (changes requested):
I suggested a few clarifying comments.

Also I would like clarity whether this will go in before or after the fix for CALCITE-6172. Before is probably better. But if before, then 6172 should add tests for LEN and LENGTH as aliases of CHAR_LENGTH.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: Added

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: I just forgot to change them, fixed now

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
tanclary: Added a comment and combined lines

PR #3608, opened 2024-01-02
[CALCITE-6184] Add checkNullTest on SqlOperatorTest

caicancai:
I noticed that some examples of SqlOperatorTest lack checknull tests. I am adding them one by one, so this PR can not be merged yet.
https://issues.apache.org/jira/browse/CALCITE-6184

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -440,6 +440,10 @@ public class SqlOperatorTest {
         false);
     f.checkBoolean("x'0A00015A' not between x'0A0001A0' and x'0A0001B0'",
         true);
+    f.checkNull("cast(null as integer) not between -1 and 2");
+    f.checkNull("1 not between -1 and cast(null as integer)");
+    f.checkNull("1 not between cast(null as integer) and cast(null as integer)");
+    f.checkNull("1 not between cast(null as integer) and 1");
   }
 
   /** Generates parameters to test both regular and safe cast. */
@@ -1882,6 +1886,7 @@ public class SqlOperatorTest {
         "Cannot apply 'CHAR' to arguments of type 'CHAR\\(<DECIMAL\\(3, 1\\)>\\)'\\. "
             + "Supported form\\(s\\): 'CHAR\\(<INTEGER>\\)'",
         false);
+    f.checkNull("char(null)");
   }
 
   @Test void testChr() {
@@ -12247,6 +12252,7 @@ public class SqlOperatorTest {
     f.checkScalar("time_trunc(time '12:34:56', hour)",
         "12:00:00", "TIME(0) NOT NULL");
     f.checkNull("time_trunc(cast(null as time), second)");
+    f.checkNull("time_trunc(cast(null as time), minute)");
   }
 
   @Test void testTimestampTrunc() {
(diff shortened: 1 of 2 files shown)


XuQianJin-Stars (approved):

XuQianJin-Stars:
hi @caicancai When you have finished adding it, please ping me.

caicancai:
@XuQianJin-Stars finish

JiajunBernoulli:
Should we use `[MINOR]`?

Here is a discussion: https://lists.apache.org/thread/39mfnbz8sqborty1lqs1j2kpx656g5o4

caicancai:
> Should we use `[MINOR]`?
> 
> Here is a discussion: https://lists.apache.org/thread/39mfnbz8sqborty1lqs1j2kpx656g5o4

Thank you. It seems that it is necessary to open a jira for my PR. I will sort out which tests should add checknull and which tests should not add checknull. At present, it seems that adding checknull is necessary for some.

caicancai:
Please wait until my jira is completed before merging it. Thank you.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -12230,6 +12235,8 @@ void testTimestampDiff(boolean coercionEnabled) {
     f.checkScalar("time_trunc(time '12:34:56', hour)",
         "12:00:00", "TIME(0) NOT NULL");
     f.checkNull("time_trunc(cast(null as time), second)");
+    f.checkNull("time_trunc(cast(null as time), minute)");
```
tanclary, line 12238: I don't really see how these tests help us. We already have a `checkNull` test for `second`, the value of the 2nd operand isn't going to influence the behavior (as long as it is the same type).

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -12292,6 +12299,18 @@ void testTimestampDiff(boolean coercionEnabled) {
             + "'TIMESTAMP_TRUNC\\(<TIME\\(0\\)>, <INTERVAL HOUR>\\)'\\. "
             + "Supported form\\(s\\): 'TIMESTAMP_TRUNC\\(<TIMESTAMP>, <DATETIME_INTERVAL>\\)'",
         false);
+    f.checkNull("timestamp_trunc(CAST(NULL AS TIMESTAMP), second)");
```
tanclary, line 12302: Same comment as above. Maybe one or two of these is good but an exhaustive list doesn't add much value other than making the test methods longer than necessary.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -12348,6 +12367,18 @@ void testTimestampDiff(boolean coercionEnabled) {
             + "'DATETIME_TRUNC\\(<TIME\\(0\\)>, <INTERVAL HOUR>\\)'\\. "
             + "Supported form\\(s\\): 'DATETIME_TRUNC\\(<TIMESTAMP>, <DATETIME_INTERVAL>\\)'",
         false);
+    f.checkNull("datetime_trunc(CAST(NULL AS TIMESTAMP), second)");
```
tanclary, line 12370: Are there tests for when the second operand is null? If not, that would be a worthwhile test to add. Same comment for other methods.

tanclary (changes requested):
I think if we are going to add tests we should make sure they are testing different things (i.e. not redundant). Adding 2 tests that test 2 different things is more valuable than adding 10 tests that test 1 thing (in my own opinion).

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: I just tested and it seems that the second parameter is null is not allowed, which will fail to parse

caicancai: thanks

caicancai:
My fork repository seems to have exploded, I will mention a new PR

PR #3609, opened 2024-01-03
[CALCITE-6189] Improve FormatElementEnumTest

XuQianJin-Stars:
Improve FormatElementEnumTest make code clean.
https://issues.apache.org/jira/browse/CALCITE-6189

PR #3611, opened 2024-01-04
[CALCITE-6185] Support more date format

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6185


core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -230,6 +249,13 @@ public enum FormatElementEnum implements FormatElement {
       calendar.setTime(date);
       sb.append(String.format(Locale.ROOT, "%d", calendar.get(Calendar.YEAR)));
     }
+  },
+  CC("cc", "") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      sb.append(String.format(Locale.ROOT, "%2d", calendar.get(Calendar.YEAR) / 100 + 1));
+    }
```
julianhyde, line 258: alphabetical?

tanclary (commented):
If you rebase on main it might be easier to tell what exactly is being added in this PR (at least for someone like me who reviews based on "Files changed").

caicancai:
I initially wanted to support timezone, but according to my tests, there seems to be no unified standard for timezone.

XuQianJin-Stars:
Please merge after this pr https://github.com/apache/calcite/pull/3609.

core/src/main/java/org/apache/calcite/util/format/FormatModels.java
```diff
@@ -55,6 +56,8 @@
 import static org.apache.calcite.util.format.FormatElementEnum.WW;
 import static org.apache.calcite.util.format.FormatElementEnum.YY;
 import static org.apache.calcite.util.format.FormatElementEnum.YYYY;
+import static org.apache.calcite.util.format.FormatElementEnum.day;
+import static org.apache.calcite.util.format.FormatElementEnum.dy;
```
XuQianJin-Stars, line 60: The imports here should be in alphabetical order.

tanclary (changes requested):

core/src/main/java/org/apache/calcite/util/format/FormatModels.java
caicancai: I tried it, but it will report an error. It seems that there is a restriction on the upper and lower case order.
import static org.apache.calcite.util.format.FormatElementEnum.*; ?

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -54,6 +61,12 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(work.eeeeFormat.format(date));
     }
   },
+  day("eeee", "The full lower case day name") {
+    @Override public void format(StringBuilder sb, Date date) {
```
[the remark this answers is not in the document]
caicancai: I didn't think of a good solution, or delete this part of the code

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -41,6 +41,13 @@
  * @see FormatModels#DEFAULT
  */
 public enum FormatElementEnum implements FormatElement {
+  CC("cc", "century (2 digits) (the twenty-first century starts on 2001-01-01)") {
```
tanclary, line 44: I think the commit message should probably be more descriptive, I wasn't sure what `Support date format` meant until I reviewed the PR.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -54,6 +61,12 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(work.eeeeFormat.format(date));
     }
   },
+  day("eeee", "The full lower case day name") {
```
tanclary, line 64: Enum constants should be all uppercase I believe. 
https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
```diff
@@ -35,6 +35,13 @@ class FormatElementEnumTest {
     assertThat(ts, hasToString("Tuesday"));
   }
 
+  @Test void testday() {
```
tanclary, line 38: Test name should be camelcase

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
```diff
@@ -63,6 +70,13 @@ class FormatElementEnumTest {
     assertThat(ts, hasToString("Tue"));
   }
 
+  @Test void testdy() {
```
tanclary, line 73: same comment as above

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
```diff
@@ -174,4 +188,11 @@ class FormatElementEnumTest {
     FormatElementEnum.YYYY.format(ts, Date.from(Instant.parse(date)));
     assertThat(ts, hasToString("2014"));
   }
+
+  @Test void testCC() {
+    final String date = "2014-09-30T10:00:00Z";
```
tanclary, line 193: Instead of adding to the end of the file why don't we make these alphabetical

tanclary (changes requested):

core/src/main/java/org/apache/calcite/util/format/FormatModels.java
mihaibudiu: Wouldn't checkstyle complain if the imports are in the wrong order?

julianhyde: Possibly checkstyle is using a different collation (e.g. different treatment of upper/lower case, spaces, underscores). Do whatever makes checkstyle happy.

julianhyde: Easiest thing is to make the enum constants upper-case.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
julianhyde: +1

I won't even review a PR until the jira case has a good description of its goals. Sounds pedantic but it's very pragmatic.

This jira case also needs to say which functions these date formats are supported in. Changes to FormatElementEnum are not useful unless they are surfaced in functions.

caicancai:
Thanks everyone

caicancai:
I will open this pr after clarifying the overall architecture and code structure of calcite. If I feel that this pr has a pr at that time,

PR #3612, opened 2024-01-04
[CALCITE-6172] Allow aliased operators to re-use existing tests

tanclary:
I used @snuyanzin 's PoC PR and @julianhyde 's improvements to make this PR and have credited them in the latest commit in this chain.

I did not modify MOD and % because the syntax seemed different enough that I preferred keeping them separate.

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -45,6 +45,7 @@ import org.apache.calcite.sql.SqlOperandCountRange;
 import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.SqlSyntax;
 import org.apache.calcite.sql.dialect.AnsiSqlDialect;
+import org.apache.calcite.sql.fun.LibraryOperator;
 import org.apache.calcite.sql.fun.SqlLibrary;
 import org.apache.calcite.sql.fun.SqlLibraryOperators;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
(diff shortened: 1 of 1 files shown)


snuyanzin:
Thanks for driving this
Seems you need to change one of your commit messages since there is a test checking that every commit message starts with a capital letter 
https://github.com/apache/calcite/actions/runs/7415936383/job/20179990922?pr=3612#step:4:642

tanclary:
> Thanks for driving this Seems you need to change one of your commit messages since there is a test checking that every commit message starts with a capital letter https://github.com/apache/calcite/actions/runs/7415936383/job/20179990922?pr=3612#step:4:642

Yeah after y'all approved I was going to squash down to the most recent commit that has a valid message if that's alright.

snuyanzin:
IMHO the issue is that this test (check for capital letter at the beginning) makes most or even all the tests are failing here...  It's better to see that tests are still green before approval...

tanclary:
> IMHO the issue is that this test (check for capital letter at the beginning) makes most or even all the tests are failing here... It's better to see that tests are still green before approval...

Ahh gotcha, I've fixed this. Thanks for pointing it out.

snuyanzin:
thanks

I noticed that currently there is `FunctionAlias` for parameterized tests however there is no junit parameterized tests using it, is it intentional?

tanclary:
> thanks
> 
> I noticed that currently there is `FunctionAlias` for parameterized tests however there is no junit parameterized tests using it, is it intentional?

@snuyanzin Julian suggested still giving each operator its own test method and then having aliases use a common method for general tests. I think this makes it a bit more readable and makes it easier to add tests for a specific alias should the need arise. What do you think?

snuyanzin:
ok, thanks for clarification
then I think we don't need a javadoc
```java
/** Contains alias data for parameterized tests. */
```
or at least need to change it since it's talking about non existing parameterized tests

tanclary:
@snuyanzin addressed your comment in latest commit

snuyanzin (commented):
Thanks for addressing comments

snuyanzin (approved):

PR #3613, opened 2024-01-05
Initial work on adding operator map

tanclary:


julianhyde:
There should be an auxiliary call 'RexNode transform(RexCall call, SqlDialect targetDialect)'. For a given operator, it should look for an equivalent operator that is avaialble in the target dialect, and convert the call into a call to the new operator.

Some operators - e.g. DATEDIFF - would permute the operands.

The BiMap is a nice idea, but I think it might run out of steam. I would have one map, and a 'registerBijection' method that would put two entries into the map, one for each target dialect.

PR #3616, opened 2024-01-08
[CALCITE-6193] SubstitutionVisitor stop trying incorrect subtree beca…

l4wei:
…use of wrong parent relationship when replace query

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
+++ b/core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
@@ -623,7 +623,7 @@ public class SubstitutionVisitor {
                   // stopTrying be wrong when exploratory replacements still exist.
                   final List<Replacement> realReplacements =
                       undoAndRemoveExploratoryReplacements(
-                          attempted, targetDescendants, lastIsFinalReplacement);
+                          attempted, targetDescendants, replacement, lastIsFinalReplacement);
                   substitutions.add(ImmutableList.copyOf(realReplacements));
                   realReplacements.clear();
                   attempted.clear();
@@ -667,7 +667,7 @@ public class SubstitutionVisitor {
    */
   private List<Replacement> undoAndRemoveExploratoryReplacements(
       List<Replacement> attempted, List<MutableRel> targetDescendants,
-      boolean lastIsFinalReplacement) {
+      MutableRel replacement, boolean lastIsFinalReplacement) {
     final List<Replacement> realReplacements = new ArrayList<>();
     final List<Replacement> exploratoryReplacements =
         new ArrayList<>(lastIsFinalReplacement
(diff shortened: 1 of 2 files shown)


JiajunBernoulli:
Can you consider reusing existing code as much as possible?

You just provided a workaround to solve it.

l4wei:
> Can you consider reusing existing code as much as possible?
> 
> You just provided a workaround to solve it.

I don't think so, I think the real problem is that exploratory replacements have not been undo, so I fix it, this is not a workaround.

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -606,14 +607,25 @@ assert rowTypesAreEquivalent(
...
-                  substitutions.add(ImmutableList.copyOf(attempted));
+                  // Let attempted only contains replacements of real substitution and the rest
+                  // must undo. Otherwise, the rest (exploratory replacements) will affect the next
+                  // attempt, because all the replacements will let these child nodes share the same
+                  // targetDescendant, which they also share the same parent node, it will make
+                  // stopTrying be wrong when exploratory replacements still exist.
+                  final List<Replacement> realReplacements =
+                      undoAndRemoveExploratoryReplacements(
+                          attempted, targetDescendants, replacement, lastIsFinalReplacement);
+                  substitutions.add(ImmutableList.copyOf(realReplacements));
```
jduo, line 627: Nit: Why not just add realReplacements directly to substitutions instead of copying it then clearing it?

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -649,6 +661,96 @@ assert rowTypesAreEquivalent(
...
+    Collections.reverse(realReplacements);
+    exploratoryReplacements.removeAll(realReplacements);
+    undoReplacement(exploratoryReplacements);
+    if (lastIsFinalReplacement) {
+      realReplacements.add(attempted.get(attempted.size() - 1));
+    }
+    return realReplacements;
+  }
+
+  private void findOutRealReplacements(
```
jduo, line 691: Rename findOutRealReplacements -> findRealReplacements() (or maybe identifyRealReplacements())

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -649,6 +661,96 @@ assert rowTypesAreEquivalent(
...
+  private boolean isFirstReplacement(
+      Replacement r,
+      List<MutableRel> targetDescendants,
+      MutableRel replacement) {
+    final MutableRel before = r.before;
+    for (MutableRel cur : MutableRels.descendants(before)) {
+      if (replacement.equals(cur)) {
+        return false;
+      }
+      for (MutableRel targetDescendant : targetDescendants) {
```
jduo, line 745: Does this need to descend again to MutableRels.descendants(targetDescendant)?

PR #3623, opened 2024-01-11
[CALCITE-6196] Accept BigQuery PERCENTILE functions without OVER clause

tanclary:


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -6235,7 +6235,6 @@ private SqlNode navigationInDefine(SqlNode node, String alpha) {
         break;
       case 2:
         assert op.allowsNullTreatment();
-        assert op.requiresOver();
```
mihaibudiu, line 6238: does the comment above this if block have to be updated?

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
tanclary: Good catch, updated

core/src/main/java/org/apache/calcite/rel/core/AggregateCall.java
```diff
@@ -467,9 +467,8 @@ public Aggregate.AggCallBinding createBinding(
     final RelDataTypeFactory typeFactory =
         aggregateRelBase.getCluster().getTypeFactory();
 
-    if (aggFunction.getKind() == SqlKind.PERCENTILE_DISC
-        || aggFunction.getKind() == SqlKind.PERCENTILE_CONT) {
-      assert collation.getKeys().size() == 1;
```
JiajunBernoulli, line 472: Throw assert error when size >1.

But now we will 
```
return new Aggregate.AggCallBinding(typeFactory, aggFunction,
        RexUtil.types(rexList), SqlTypeUtil.projectTypes(rowType, argList),
        aggregateRelBase.getGroupCount(), hasFilter());
```

Can the change be declared in JIRA?

tanclary: Where would this assertion go? BQ % functions have 2 arguments but should not fail.

tanclary: BQ PERCENT functions do not use `WITHIN GROUP` clause to declare the column (ex. `WITHIN GROUP(ORDER BY users.age)`) so the `collation` should always be empty.

tanclary: To my understanding thats why there's a specific `PercentileDiscAggCall`.

julianhyde: Did we agree that BQ-specific percentile functions would persist into the RelNode/RexNode layer?

That was and is still yuck.

They have different syntax, but after parsing, they're the same old percentile functions. Algebraically they are the same. Someone could choose to generate Oracle SQL from them (with Oracle syntax of course).

PR #3624, opened 2024-01-13
[CALCITE-6193] If a query has more than one subexpression that matche…

JiajunBernoulli:
…s a materialized view, only the first is substituted

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
+++ b/core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
@@ -559,7 +559,7 @@ public class SubstitutionVisitor {
 
     final List<Replacement> attempted = new ArrayList<>();
     List<List<Replacement>> substitutions = new ArrayList<>();
-
+    int globalInputIndex = 0;
     for (;;) {
       int count = 0;
       MutableRel queryDescendant = query;
@@ -575,9 +575,11 @@ public class SubstitutionVisitor {
           }
         }
         final MutableRel next = MutableRels.preOrderTraverseNext(queryDescendant);
+        final int currentInputIndex = queryDescendant.getInputs().size() > globalInputIndex
+            ? globalInputIndex : 0;
         final MutableRel childOrNext =
             queryDescendant.getInputs().isEmpty()
-                ? next : queryDescendant.getInputs().get(0);
+                ? next : queryDescendant.getInputs().get(currentInputIndex);
         for (MutableRel targetDescendant : targetDescendants) {
           for (UnifyRule rule
               : applicableRules(queryDescendant, targetDescendant)) {
@@ -616,6 +618,7 @@ public class SubstitutionVisitor {
                   substitutions.add(ImmutableList.copyOf(attempted));
                   attempted.clear();
                   queryDescendant = next;
+                  globalInputIndex++;
                   continue outer;
                 }
                 // We will try walking the query tree all over again to see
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/test/MaterializedViewSubstitutionVisitorTest.java
```diff
@@ -1849,6 +1849,43 @@ protected final MaterializedViewFixture sql(String materialize,
...
+        + matchQuery
+        + ") as t1 inner join (\n"
+        + matchQuery
+        + ") as t2 on t1.\"deptno\" = t2.\"deptno\"\n";
+    sql(mv, query)
+        .checkingThatResultContains(""
+            + "LogicalCalc(expr#0..1=[{inputs}], deptno=[$t0])\n"
+            + "  LogicalJoin(condition=[=($0, $1)], joinType=[inner])\n"
+            + "    LogicalUnion(all=[true])\n"
+            + "      LogicalCalc(expr#0..1=[{inputs}], deptno=[$t1])\n"
```
l4wei, line 1879: The 1879 and 1880 lines are still be replaced by mv's target descendant, it's parent may be shared with a wrong node, so I think this solution is not very is not very correct.

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -575,9 +575,11 @@ assert rowTypesAreEquivalent(
           }
         }
         final MutableRel next = MutableRels.preOrderTraverseNext(queryDescendant);
+        final int currentInputIndex = queryDescendant.getInputs().size() > globalInputIndex
+            ? globalInputIndex : 0;
         final MutableRel childOrNext =
             queryDescendant.getInputs().isEmpty()
-                ? next : queryDescendant.getInputs().get(0);
+                ? next : queryDescendant.getInputs().get(currentInputIndex);
```
l4wei, line 582: I think getInputs.get(currentInputIndex) is just a walk-around to force traverse the missed nodes. Actually, preOrderTraverseNext method can traverse all the nodes of this plan tree, There is no need to add a new status to record the current node's position.

JiajunBernoulli: It is a demo.

I think that always `get(0)`  is error. 
We should find a better way to traverse the missing nodes.

l4wei: I think get(0) is correct, get(0) is just a starting point when traverse child-nodes, preOrderTraverseNext method can traverse the missing nodes.

core/src/test/java/org/apache/calcite/test/MaterializedViewSubstitutionVisitorTest.java
JiajunBernoulli: If I don't commit, the RelNode is
```
LogicalCalc(expr#0..1=[{inputs}], deptno=[$t0])
  LogicalJoin(condition=[=($0, $1)], joinType=[inner])
    LogicalUnion(all=[true])
      LogicalCalc(expr#0..1=[{inputs}], deptno=[$t1]) -- same as 1879
        LogicalCalc(expr#0..4=[{inputs}], proj#0..1=[{exprs}]) -- same as 1880
          LogicalTableScan(table=[[hr, emps]])
      LogicalAggregate(group=[{1}])
        EnumerableTableScan(table=[[hr, MV0]])
    LogicalAggregate(group=[{0}])
      LogicalCalc(expr#0..4=[{inputs}], deptno=[$t1])
        LogicalTableScan(table=[[hr, emps]])
```
There are two `LogicalCalc`, do you mean this is Error?

l4wei: Yes, because these two LogicalCalc operators both are replaced by mv's target descendant, I think these temporary replacements should be undo after the real replacement has been done. This is the correct solution for this problem.

PR #3627, opened 2024-01-15
[CALCITE-6202] `sqlsh` does not print error message when query fails

libenchao:


mihaibudiu (approved):

libenchao:
@mihaibudiu Thanks for the review, I have slightly changed the idea about how to fix it, see more in the Jira comment of CALCITE-6202. And it seems not a best practice to close the `System.out` according to https://stackoverflow.com/questions/8941298/system-out-closed-can-i-reopen-it

plus/src/main/java/org/apache/calcite/adapter/os/SqlShell.java
```diff
@@ -118,17 +118,18 @@ private static String model() {
...
-                 new OutputStreamWriter(System.err, StandardCharsets.UTF_8));
-         InputStreamReader in =
-             new InputStreamReader(System.in, StandardCharsets.UTF_8);
-         PrintWriter out =
-             new PrintWriter(
-                 new OutputStreamWriter(System.out, StandardCharsets.UTF_8))) {
+    try {
+      final PrintWriter err =
+          new PrintWriter(
+              new OutputStreamWriter(System.err, StandardCharsets.UTF_8));
```
JiajunBernoulli, line 124: Previously, `try-with-resource` syntax will close stream, 
but now we haven't closed them.

Should we close them?

mihaibudiu (approved):

plus/src/main/java/org/apache/calcite/adapter/os/SqlShell.java
libenchao: I think it's ok to not close them, we only need to ensure they flush all the data into enclosed stream (for now, only `out` is used, and it's flushed in `finally` block).

PR #3629, opened 2024-01-16
[CALCITE-6207] Improve JDBC connection resource in JDBCUtils

xuzifu666:
Improve JDBC connection resource in JDBCUtils: not need multiple judge connection whether close


mihaibudiu (approved):

mihaibudiu:
I have merged, then cancelled the associated commit, since the change is still under debate in JIRA.

mihaibudiu:
The discussion in the JIRA is about the behavior in the presence of exceptions. It seems that this improvement may cause a different exception to be thrown.

mihaibudiu:
Apparently there is no way to reopen a merged PR, so please open a new one if there is an alternate fix.

xuzifu666:
> Apparently there is no way to reopen a merged PR, so please open a new one if there is an alternate fix.

OK，I would open a new one

PR #3630, opened 2024-01-16
[CALCITE-6205] Add BITAND_AGG, BITOR_AGG functions (enabled in Snowfl…

tanclary:
…ake library)

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -143,6 +143,8 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.ARRAY_TO_STRING;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.ARRAY_UNION;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.ASINH;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.ATANH;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.BITAND_AGG;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.BITOR_AGG;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.BIT_GET;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.BIT_LENGTH;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.BOOL_AND;
@@ -1041,6 +1043,8 @@ public class RexImpTable {
       aggMap.put(LOGICAL_OR, minMax);
       final Supplier<BitOpImplementor> bitop =
           constructorSupplier(BitOpImplementor.class);
+      aggMap.put(BITAND_AGG, bitop);
+      aggMap.put(BITOR_AGG, bitop);
       aggMap.put(BIT_AND, bitop);
       aggMap.put(BIT_OR, bitop);
       aggMap.put(BIT_XOR, bitop);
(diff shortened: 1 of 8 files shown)


mihaibudiu (approved):
I find it a bit ironic that we have aggregation functions with bitwise operations but we have no corresponding scalar functions that do the same thing.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -6662,6 +6662,28 @@ private void checkLiteral2(String expression, String expected) {
...
+    sql(query).withLibrary(SqlLibrary.SNOWFLAKE).withSnowflake().ok(expectedSnowflake);
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6205">[CALCITE-6205]
+   * Add BITAND_AGG, BITOR_AGG functions (enabled in Snowflake library)</a>. */
+  @Test void testBitOrAgg() {
+    final String query = "select bit_or(\"product_id\")\n"
+        + "from \"product\"";
+    final String expectedSnowflake = "SELECT BITOR_AGG(\"product_id\")\n"
```
JiajunBernoulli, line 6682: Should we replace **BITOR_AGG** to **BITOR** when `unparse` BigQuery sql?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -1793,7 +1797,7 @@ static class BitOpImplementor extends StrictAggImplementor {
       if (SqlTypeUtil.isBinary(info.returnRelType())) {
         start = Expressions.field(null, ByteString.class, "EMPTY");
       } else {
-        Object initValue = info.aggregation() == BIT_AND ? -1L : 0;
+        Object initValue = info.aggregation().kind == SqlKind.BIT_AND ? -1L : 0;
```
olivrlee, line 1800: This change is just to be consistent with the other functions right?

tanclary: This change is to make sure "BITAND_AGG" and "BIT_AND" are caught, previously it was just checking for a specific operator so "BITAND_AGG" was not going down the correct codepath

core/src/main/java/org/apache/calcite/sql/fun/SqlBitOpAggFunction.java
```diff
@@ -39,7 +39,7 @@ public class SqlBitOpAggFunction extends SqlAggFunction {
 
   //~ Constructors -----------------------------------------------------------
 
-  /** Creates a SqlBitOpAggFunction. */
+  /** Creates a SqlBitOpAggFunction from a SqlKind. */
```
olivrlee, line 42: nit: What do you think about changing the previous call sites of this to call the new function with 
`SqlBitOpAggFunction(call.getKind().getName(), call.getKind())`?

tanclary: I'm indifferent, the standard operators can derive the name from the `SqlKind`, the library-specific ones cannot so they need the name `specified`. Whether it's refactored so they can use the same constructor doesn't matter too much to me

tanclary: Actually re-reading this I think I would prefer keeping the `SqlBitOpAggFunction(SqlKind kind)` constructor, I think it makes sense to only pass in one parameter if possible.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
tanclary: My JIRA case was actually wrong, `BIT_OR` and `BIT_AND` are both standard operators not specific to BigQuery, I updated the case.

tanclary:
Are there any other comments @JiajunBernoulli @olivrlee @mihaibudiu ? If not, I will probably merge this within the next day or so.

mihaibudiu (approved):

JiajunBernoulli (approved):

PR #3631, opened 2024-01-17
[CALCITE-6198] Add TO_TIMESTAMP_LTZ function (enabled in Snowflake li…

tanclary:
…brary)


JiajunBernoulli:
Does CI stability failed?
https://github.com/apache/calcite/actions/runs/7549299627/attempts/1?pr=3631

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1610,6 +1610,36 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
           OperandTypes.STRING_STRING,
           SqlFunctionCategory.TIMEDATE);
 
+  /** The "TO_TIMESTAMP_LTZ" function returns a Calcite
```
julianhyde, line 1613: It's worth spelling out that a `TIMESTAMP WITH LOCAL TIME ZONE` is stored as millis since UTC epoch.

Describe which time zone is used for conversion from `DATE` and `TIMESTAMP`.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1610,6 +1610,36 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
+   * {@code TIMESTAMP WITH LOCAL TIME ZONE}.
+   * It has the following overloads (Snowflake also
+   * supports a variant and quoted integer overload but
+   * those are not yet supported):
+   *
+   * <ul>
+   *   <li>{@code TO_TIMESTAMP_LTZ(numeric[, scale)]}
+   *   <li>{@code TO_TIMESTAMP_LTZ(date)}
+   *   <li>{@code TO_TIMESTAMP_LTZ(datetime)}
```
julianhyde, line 1622: datetime is not a Calcite type

site/_docs/reference.md
```diff
@@ -2746,7 +2746,7 @@ In the following:
 | h s | FORMAT_NUMBER(value, decimalVal)             | Formats the number *value* like '#,###,###.##', rounded to decimal places *decimalVal*. If *decimalVal* is 0, the result has no decimal point or fractional part
 | h s | FORMAT_NUMBER(value, format)                 | Formats the number *value* to MySQL's FORMAT *format*, like '#,###,###.##0.00'
 | b | FORMAT_TIME(string, time)                      | Formats *time* according to the specified format *string*
-| b | FORMAT_TIMESTAMP(string timestamp)             | Formats *timestamp* according to the specified format *string*
+| b | FORMAT_TIMESTAMP(string, timestamp)             | Formats *timestamp* according to the specified format *string*
```
julianhyde, line 2749: remove space before bar

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -457,6 +457,9 @@ public enum SqlKind {
   /** {@code TIME_SUB} function (BigQuery). */
   TIME_SUB,
 
+  /** {@code TIMESTAMP} function (BigQuery). */
+  TIMESTAMP,
```
julianhyde, line 461: not a good name, since TIMESTAMP is a type

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4461,6 +4461,51 @@ public static int time(long timestampMillis, String timeZone) {
...
+    return (long) (timestampSeconds * multiplier);
+  }
+
+  /** SQL {@code TO_TIMESTAMP_LTZ(timestampSeconds, scale)}
+  * function for a String expression of a timestamp. */
+  public static long toTimestampLtz(String timestamp) {
+    return timestamp(timestamp);
+  }
+
+  /** SQL {@code TO_TIMESTAMP_LTZ(timestamp)} function
```
julianhyde, line 4503: be explicit what the argument means. millis since local epoch? millis since UTC epoch?

using 'timestamp' in the function name would by default mean a Calcite timestamp, i.e. ltz. But then the function would be a no-op.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4461,6 +4461,51 @@ public static int time(long timestampMillis, String timeZone) {
         / (1000L * 1000L)); // milli > micro > nano
   }
 
+  /** SQL {@code TO_TIMESTAMP_LTZ(date)} function
+  * for a date values. */
+  public static long toTimestampLtzDate(int days) {
+    return timestamp(days);
```
julianhyde, line 4467: having overloaded `timestamp(int)` and `timestamp(long)` functions with different semantics is error-prone

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -2480,6 +2484,32 @@ private static class RegexpReplaceImplementor extends AbstractRexCallImplementor
     }
   }
 
+  /** Implementor for the {@code TO_TIMESTAMP_LTZ} function. */
```
julianhyde, line 2487: is an implementor necessary? it would be better if calls to TO_TIMESTAMP_LTZ were translated to existing functions. Then we would not need new methods in `SqlFunctions`

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4461,6 +4461,51 @@ public static int time(long timestampMillis, String timeZone) {
...
+  }
+
+  /** SQL {@code TO_TIMESTAMP_LTZ(timestampSeconds, scale)}
+  * function for BigDecimal values with a specified scale. */
+  public static long toTimestampLtz(BigDecimal timestampSeconds, int scale) {
+    return toTimestampLtz(timestampSeconds.longValue(), scale);
+  }
+
+  /** SQL {@code TO_TIMESTAMP_LTZ(timestampSeconds, scale)}
+  * function for long values with a specified scale. */
```
julianhyde, line 4489: Can you add examples? Does `toTimestampLtz(86_400L, 0)` mean `TIMESTAMP '1970-01-02 00:00:00'`?

What would `toTimestampLtz(86_400L, 3)` mean?

julianhyde (changes requested):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: I was under the impression a Calcite `TIMESTAMP` did not have a LTZ but `TIMESTAMP WITH LOCAL TIME ZONE` did. So if you provide the function a Calcite `TIMESTAMP` it would return a `TIMESTAMP WITH LOCAL TIME ZONE` (not a no-op).

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: Yeah I just was not sure how else to handle the fact that you can provide the function a `DATE` which gets converted to an `int` or you can provide the function with a numeric value (which could be a long).

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
tanclary: The main reason I added the implementor was to avoid the "error proneness" you mentioned in your above comment. It checks the type of the first argument (i.e. `DATE`, `TIMESTAMP`, etc.) to ensure it does not call the wrong method via reflection.

Before I added the implementor, when I was trying to use `TO_TIMESTAMP_LTZ(numeric)`, it would end up using the method for `DATE` arguments since dates are represented as integers. This way, that confusion is avoided.

site/_docs/reference.md
tanclary: Done

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
tanclary: Added a longer comment that explains the `scale` parameter better. Added an example too. Let me know what you think.

core/src/main/java/org/apache/calcite/sql/SqlKind.java
tanclary: Done, didn't need to add that anyways I don't think.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
tanclary: Done

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
tanclary: Done

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8747,6 +8747,55 @@ private void testCurrentDateFunc(Pair<String, Hook.Closeable> pair) {
...
+    f.checkScalar("to_timestamp_ltz(123456, 3)",
+        "1970-01-01 00:02:03",
+        "TIMESTAMP_WITH_LOCAL_TIME_ZONE(0) NOT NULL");
+    f.checkScalar("to_timestamp_ltz(123456891234343434, 9)",
+        "1973-11-29 21:34:51",
+        "TIMESTAMP_WITH_LOCAL_TIME_ZONE(0) NOT NULL");
+    f.checkScalar("to_timestamp_ltz(12345689.123, 2)",
+        "1970-01-02 10:17:36",
+        "TIMESTAMP_WITH_LOCAL_TIME_ZONE(0) NOT NULL");
+    f.checkNull("to_timestamp_ltz(null, 2)");
```
caicancai, line 8779: Hello，to_timestamp_ltz(null, null) will report an error?

xuzifu666:
This pr had been inactivity for a long time，could assgine to me to finish it? @julianhyde  @caicancai @mihaibudiu

PR #3632, opened 2024-01-17
[CALCITE-5647] RelMdPopulationSize should use mq.getRowCount(rel) instead of rel.estimateRowCount(mq)

jduo:
(From: #3151)
This PR corrects a row count call to use the RelMetadataQuery instead of going directly to rel.estimateRowCount.

Using rel.estimateRowCount directly will behave incorrectly when using a cardinality estimator that is implemented entirely in a MetadataHandler class, as it will evade the metadata handler's implementation of getRowCount and fall through to the underlying default method for Values.

This corrects the problem by using mq.getRowCount instead of rel.estimateRowCount. In the default case this results in the same value but in the custom cardinality model case the expected handler is called.

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 package org.apache.calcite.test;
+
 import org.apache.calcite.adapter.enumerable.EnumerableConvention;
 import org.apache.calcite.adapter.enumerable.EnumerableLimit;
 import org.apache.calcite.adapter.enumerable.EnumerableMergeJoin;
@@ -76,6 +77,7 @@ import org.apache.calcite.rel.metadata.ReflectiveRelMetadataProvider;
 import org.apache.calcite.rel.metadata.RelColumnOrigin;
 import org.apache.calcite.rel.metadata.RelMdCollation;
 import org.apache.calcite.rel.metadata.RelMdColumnUniqueness;
+import org.apache.calcite.rel.metadata.RelMdPopulationSize;
 import org.apache.calcite.rel.metadata.RelMdUtil;
 import org.apache.calcite.rel.metadata.RelMetadataProvider;
 import org.apache.calcite.rel.metadata.RelMetadataQuery;
@@ -3637,6 +3639,43 @@ public class RelMetadataTest {
         is(mq.getPopulationSize(rel, bitSetOf(0))));
     assertThat(mq.getPopulationSize(rel, bitSetOf(0, 1, 2)),
         is(mq.getPopulationSize(rel, bitSetOf(0))));
+  }
+
+  /**
+   * Test that RelMdPopulationSize is calculated based on the RelMetadataQuery#getRowCount().
+   *
+   * @see <a href="https://issues.apache.org/jira/browse/CALCITE-5647">[CALCITE-5647]</a>
+   */
+  @Test public void testPopulationSizeFromValues() {
+    final String sql = "values(1,2,3),(1,2,3),(1,2,3),(1,2,3)";
+    final RelNode rel = sql(sql).toRel();
+    assertThat(rel, instanceOf(Values.class));
+
+    RelMetadataProvider provider = RelMdPopulationSize.SOURCE;
+
+    List<MetadataHandler<?>> handlers =
+        provider.handlers(BuiltInMetadata.PopulationSize.Handler.class);
+
+    // The population size is calculated to be half the row count. (The assumption is that half
+    // the rows are duplicated.) With the default handler it should evaluate to 2 since there
+    // are 4 rows.
+    RelMdPopulationSize populationSize = (RelMdPopulationSize) handlers.get(0);
+    Double popSize =
+        populationSize.getPopulationSize((Values) rel, rel.getCluster().getMetadataQuery(),
+            bitSetOf(0, 1, 2));
+    assertEquals(2.0, popSize);
+
+    // If we use a custom RelMetadataQuery and override the row count, the population size
+    // should be half the reported row count. In this case we will have the RelMetadataQuery say
+    // the row count is 12 for testing purposes, so we should expect a population size of 6.
+    RelMetadataQuery customQuery = new RelMetadataQuery() {
+      @Override public Double getRowCount(RelNode rel) {
+        return 12.0;
+      }
+    };
+
+    popSize = populationSize.getPopulationSize((Values) rel, customQuery, bitSetOf(0, 1, 2));
+    assertEquals(6.0, popSize);
   }
 
   private static final SqlOperator NONDETERMINISTIC_OP =
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdPopulationSize.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdPopulationSize.java
@@ -113,7 +113,7 @@ public class RelMdPopulationSize
   public Double getPopulationSize(Values rel, RelMetadataQuery mq,
       ImmutableBitSet groupKey) {
     // assume half the rows are duplicates
-    return rel.estimateRowCount(mq) / 2;
+    return mq.getRowCount(rel) / 2;
   }
 
   public @Nullable Double getPopulationSize(Project rel, RelMetadataQuery mq,


jduo:
@libenchao @chunweilei

mihaibudiu:
There are some style violations

jduo:
> There are some style violations

Thanks. Should be correct now.

JiajunBernoulli:
@jduo Thanks for your PR.

Just minor suggestion:
1. Commit message should start with `[CALCITE-5647]`.
2. We like adding Java Document and JIRA link before unit test.

core/src/test/java/org/apache/calcite/rel/metadata/RelMdPopulationSizeTest.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests for {@link RelMdPopulationSize}.
+ */
+public class RelMdPopulationSizeTest {
+
+  @Test public void testPopulationSizeFromValues() {
```
JiajunBernoulli, line 48: Can we add it in `RelMetadataTest`?

jduo:
Thanks @JiajunBernoulli.

Do you mean I should squash my commits and give it the [CALCITE-5647] prefix?

JiajunBernoulli:
@jduo Yes, you can `squash` after any committer approve.

But now we maybe need some new commits to make PR better.

core/src/test/java/org/apache/calcite/rel/metadata/RelMdPopulationSizeTest.java
jduo: Done.

jduo:
> @jduo Thanks for your PR.
> 
> Just minor suggestion:
> 
> 1. Commit message should start with `[CALCITE-5647]`.
> 2. We like adding Java Document and JIRA link before unit test.

Done.

jduo:
@JiajunBernoulli @mihaibudiu , were there any other concerns to address in this PR?

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -3639,6 +3642,37 @@ public void checkAllPredicatesAndTableSetOp(String sql) {
...
+    final String sql = "values(1,2,3)";
+    final RelNode rel = sql(sql).toRel();
+    assertThat(rel, instanceOf(Values.class));
+
+    RelMetadataProvider provider = RelMdPopulationSize.SOURCE;
+
+    List<MetadataHandler<?>> handlers =
+        provider.handlers(BuiltInMetadata.PopulationSize.Handler.class);
+
+    // Act
```
mihaibudiu, line 3661: I am not sure that these 3 comments are useful.
In fact, this one confused me.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -3639,6 +3642,37 @@ public void checkAllPredicatesAndTableSetOp(String sql) {
         is(mq.getPopulationSize(rel, bitSetOf(0))));
   }
 
+  /**
+   * Test that RelMdPopulationSize is calculated based on the RelMetadataQuery#getRowCount().
+   *
+   * @see <a href="https://issues.apache.org/jira/browse/CALCITE-5647">[CALCITE-5647]</a>
+   */
+  @Test public void testPopulationSizeFromValues() {
```
mihaibudiu, line 3650: This isn't really displaying a bug that is being fixed.
From the JIRA comments I understand that it's difficult to create a small reproduction.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
jduo: This is a pretty common way of documenting tests:
https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/

- Arrange would be setting up the test (that isn't global setup like the @Setup function)
- Act would be executing the test behavior
- Assert would be validating the result.

I can remove these comments but don't really see the harm in them.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
jduo: The bug is that the population size was being calculated without utilizing RelMetadataQuery#getRowCount(). The idea was that this test demonstrates the getRowCount() call is being used.

Does that clarify the bug and purpose of the test? This sort of test you'd probably want to spy() instead but I didn't want to introduce Mockito into the core sub-module just for this.

jduo:
Hi @mihaibudiu and @JiajunBernoulli , just wondering what the next steps are with this PR.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
mihaibudiu: Comments should provide information that is not clear from the code. "Assert" for example does not. And "act" could mean many thing. I think it would be better without them.

But here the test is not showing that you fixed a bug, it is showing just that a specific method is called. The bug should manifest itself in producing an incorrect or suboptimal result for some code transformation. That would be a convincing test.

jduo:
@mihaibudiu I've changed the test. It's still a bit artificial because it makes a RelMetadataQuery which intentionally reports an incorrect statistic but hopefully demonstrates how the bug is fixed. Hope the comments are clear enough.

jduo:
Forgot to add @adamkennedy to this PR.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
mihaibudiu: Yes, the test is still artificial.

mihaibudiu (approved):
I am approving, but I am still unsure about exactly what this improvement fixes.

jduo:
> I am approving, but I am still unsure about exactly what this improvement fixes.

The intent with RelMetadataQuery is that the implementation can provide custom behavior for each of the statistics you can requet with it.

For the case of the population size of Values clauses, it should be based on the row count / 2. However RelMetadataQuery was not using invoking its (potentially custom) method to get the row count, it was just inlining the row count estimate from the node.

mihaibudiu:
You are describing how something is implemented. A bug description should be something like "Query X produces result Y instead of the expected Z", or "Optimizer produces the plan X instead of the expected Y because it uses the wrong statistics".

jduo:
> You are describing how something is implemented. A bug description should be something like "Query X produces result Y instead of the expected Z", or "Optimizer produces the plan X instead of the expected Y because it uses the wrong statistics".

I guess in this case, the original implementation happens to return the correct value, but is not behaving as a valid extension point. So I'm not sure what would be a good concrete test for this type of problem.

mihaibudiu:
Supposedly you have changed this behavior because it causes some problems in some application someone cares about. The ideal test case would be a small reproduction of the bug which caused this quest in the first place. The JIRA cases suggests that a small reproduction is difficult.

rubenada:
LGTM.
@jduo , Just to avoid confusion, would it be possible to align commit message with PR title and Jira title?

jduo:
@rubenada , I've made the commit and PR have the same title. Should the JIRA have the same title now? It states the problem whereas the commit/PR describe what changed.

rubenada:
@jduo, usually the commit message is the same as Jira subject (more info here: https://calcite.apache.org/develop/#contributing).

What if we use (for both Jira and commit) something that describes in-between the problem and the change, i.e.  `[CALCITE-5647] RelMdPopulationSize should use mq.getRowCount(rel) instead of rel.estimateRowCount(mq)`   ?

jduo:
> @jduo, usually the commit message is the same as Jira subject (more info here: https://calcite.apache.org/develop/#contributing).
> 
> What if we use (for both Jira and commit) something that describes in-between the problem and the change, i.e. `[CALCITE-5647] RelMdPopulationSize should use mq.getRowCount(rel) instead of rel.estimateRowCount(mq)` ?

Sounds good. I've updated the name of the PR and the commit message. However I cannot seem to edit the JIRA title (perhaps because I'm not the original reporter of the issue).

rubenada:
Thanks @jduo , I have updated the Jira title myself.

@mihaibudiu , @JiajunBernoulli  I think the PR is in a good shape, do you have further remarks? Otherwise I'll merge it in the next 24-48h

mihaibudiu:
I don't have any comments.

PR #3633, opened 2024-01-17
[CALCITE-6208] update JSON_VALUE return type inference to make explicit array return types be nullable with nullable elements

clintropolis:
From [CALCITE-6208](https://issues.apache.org/jira/browse/CALCITE-6208)

This PR modifies the `SqlJsonValueFunction` return type inference to special handle explicit array syntax so that the arrays are returned with nullable elements.

The added tests fail with:
```
Expected: is "INTEGER ARRAY"
     but: was "INTEGER NOT NULL ARRAY"
```
and
```
Expected: is "INTEGER ARRAY ARRAY"
     but: was "INTEGER NOT NULL ARRAY NOT NULL ARRAY"
```

without the changes in this PR.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlJsonValueFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlJsonValueFunction.java
@@ -30,11 +30,10 @@ import org.apache.calcite.sql.SqlOperandCountRange;
 import org.apache.calcite.sql.SqlOperatorBinding;
 import org.apache.calcite.sql.SqlWriter;
 import org.apache.calcite.sql.type.OperandTypes;
-import org.apache.calcite.sql.type.ReturnTypes;
 import org.apache.calcite.sql.type.SqlOperandCountRanges;
 import org.apache.calcite.sql.type.SqlTypeFamily;
 import org.apache.calcite.sql.type.SqlTypeName;
-import org.apache.calcite.sql.type.SqlTypeTransforms;
+import org.apache.calcite.sql.type.SqlTypeUtil;
 
 import com.google.common.collect.ImmutableList;
 
@@ -54,9 +53,9 @@ public class SqlJsonValueFunction extends SqlFunction {
 
   public SqlJsonValueFunction(String name) {
     super(name, SqlKind.OTHER_FUNCTION,
-        ReturnTypes.cascade(
-            opBinding -> explicitTypeSpec(opBinding).orElse(getDefaultType(opBinding)),
-            SqlTypeTransforms.FORCE_NULLABLE),
+        opBinding -> explicitTypeSpec(opBinding)
+            .map(relDataType -> deriveExplicitType(opBinding, relDataType))
+            .orElseGet(() -> getDefaultType(opBinding)),
         null,
         OperandTypes.family(
             ImmutableList.of(SqlTypeFamily.ANY, SqlTypeFamily.CHARACTER),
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlJsonValueFunction.java
```diff
@@ -54,20 +53,33 @@ public class SqlJsonValueFunction extends SqlFunction {
...
+          true);
+    }
+    return opBinding.getTypeFactory().createTypeWithNullability(type, true);
+  }
+
   /** Returns VARCHAR(2000) as default. */
   private static RelDataType getDefaultType(SqlOperatorBinding opBinding) {
     final RelDataTypeFactory typeFactory = opBinding.getTypeFactory();
-    return typeFactory.createSqlType(SqlTypeName.VARCHAR, 2000);
+    final RelDataType baseType = typeFactory.createSqlType(SqlTypeName.VARCHAR, 2000);
```
mihaibudiu, line 81: why 2000? This will truncate results to 2000 characters. why not just VARCHAR?

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -11413,6 +11413,14 @@ private void checkCustomColumnResolving(String table) {
     expr("json_value('{\"foo\":100}', 'lax $.foo' returning boolean"
         + " default 100 on empty)")
         .columnType("BOOLEAN");
+
+    expr("json_value('{\"foo\":[100, null, 200]}', 'lax $.foo'"
+        + "returning integer array)")
+        .columnType("INTEGER ARRAY");
+
+    expr("json_value('{\"foo\":[[100, null, 200]]}', 'lax $.foo'"
```
mihaibudiu, line 11421: I am not familiar with this function. Can it be used when the array element types are not the same? JSON certainly allows that.

core/src/main/java/org/apache/calcite/sql/fun/SqlJsonValueFunction.java
clintropolis: i don't have the context for why this was originally chosen, i was just preserving existing behavior

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
clintropolis: I am unsure exactly of the standard; I can only speak for the behavior in Apache Druid where we have a dedicated json type rather than operating on strings and we coerce array elements to a common 'least restrictive type'.

core/src/main/java/org/apache/calcite/sql/fun/SqlJsonValueFunction.java
mihaibudiu: I have fixed a couple of bugs in the past related to that: https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-5813. This very much looks like a bug to me. It means that if you get a result longer than 2000 characters it will be truncated. And that's not what you want. Whether you want to fix it in this PR or another one, that's a different issue.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
mihaibudiu: This has been a desired feature in Calcite for quite some time: https://issues.apache.org/jira/browse/CALCITE-4918
Does anyone plan to contribute it (assuming you are using Calcite for this purpose)?
I was thinking to do it myself, but if you already have the code and the tests, it would be great.

core/src/main/java/org/apache/calcite/sql/fun/SqlJsonValueFunction.java
clintropolis: yeah, happy to change it in this PR, it seems like a strange default to me

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
clintropolis: Yeah, I agree it would be nice to have, but don't really have any pre-existing code for the calcite side of things for this. Druid could probably make use of it if such a thing exists as we store this type of varying json data in their original types in our segments, but right now at query time we best effort convert them to a common type for query processing.

core/src/main/java/org/apache/calcite/sql/fun/SqlJsonValueFunction.java
clintropolis: looking closer, it seems like a lot of the JSON functions have varchar 2000 as the default. I'm kind of thinking should split that change into a separate PR to change them all at once instead of just fixing one (or all of them) as part of this PR

mihaibudiu: sounds reasonable

XuQianJin-Stars:
hi @clintropolis resubmit this pr to trigger ci.

JiajunBernoulli (changes requested):
Commit message should start with `[CALCITE-6208]`.

If you are ready, we can merge it.

XuQianJin-Stars:
image

clintropolis:
updated the commit message, not sure why CI is still failing

PR #3154, opened 2023-04-15
CALCITE-5649 Produce row count statistics from ReflectiveSchema for array based tables

adamkennedy:
This PR adds row count statistics for simple reflective schemas with array field tables.

This small enhancement makes it a little easier to write tests for optimizer rules, particularly rules that might behave specially when the row count is 1 or 0.

PR #3634, opened 2024-01-17
[CALCITE-5649] Get row count statistics from ReflectiveSchema

jduo:
(from #3154)
This PR adds row count statistics for simple reflective schemas with array and Collection field tables.

This small enhancement makes it a little easier to write tests for optimizer rules, particularly rules that might behave specially when the row count is 1 or 0.

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/ReflectiveSchemaTest.java
+++ b/core/src/test/java/org/apache/calcite/test/ReflectiveSchemaTest.java
@@ -29,11 +29,14 @@ import org.apache.calcite.linq4j.tree.Expressions;
 import org.apache.calcite.linq4j.tree.ParameterExpression;
 import org.apache.calcite.linq4j.tree.Types;
 import org.apache.calcite.schema.SchemaPlus;
+import org.apache.calcite.schema.Statistic;
+import org.apache.calcite.schema.Table;
 import org.apache.calcite.schema.impl.AbstractSchema;
 import org.apache.calcite.schema.impl.TableMacroImpl;
 import org.apache.calcite.schema.impl.ViewTable;
 import org.apache.calcite.test.schemata.catchall.CatchallSchema;
 import org.apache.calcite.test.schemata.catchall.CatchallSchema.EveryType;
+import org.apache.calcite.test.schemata.hr.Dependent;
 import org.apache.calcite.test.schemata.hr.Employee;
 import org.apache.calcite.test.schemata.hr.HrSchema;
 import org.apache.calcite.util.Smalls;
@@ -65,6 +68,7 @@ import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasToString;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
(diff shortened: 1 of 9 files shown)


jduo:
@rubenada @zoudan @libenchao @adamkennedy

core/src/test/java/org/apache/calcite/test/ReflectiveSchemaTest.java
```diff
@@ -919,4 +922,30 @@ public static class DateColumnSchema {
...
             "EXPR$0=null");
   }
+
+  /**
+   * Test that the row count statistic can be retrieved from a
+   * ReflectiveSchema.
+   *
+   * @see <a href="https://issues.apache.org/jira/browse/CALCITE-5649">[CALCITE-5649]</a>
+   *
+   * @throws Exception on error
```
mihaibudiu, line 932: well, a test does either throw or doesn't
I think this comment is not useful.

jduo:
It looks like many unrelated tests are failing, and it's probably because reporting row counts has the side effect of altering join orderings. I'm thinking to deal with this we conditionally enable row count statistics on tables.

jduo:
@mihaibudiu , were there any other concerns to address with this PR?

core/src/test/java/org/apache/calcite/examples/foodmart/java/JdbcExample.java
```diff
@@ -43,8 +43,8 @@ public void run() throws ClassNotFoundException, SQLException {
     CalciteConnection calciteConnection =
         connection.unwrap(CalciteConnection.class);
     SchemaPlus rootSchema = calciteConnection.getRootSchema();
-    rootSchema.add("hr", new ReflectiveSchema(new Hr()));
-    rootSchema.add("foodmart", new ReflectiveSchema(new Foodmart()));
+    rootSchema.add("hr", new ReflectiveSchemaWithoutRowCount(new Hr()));
```
mihaibudiu, line 46: what happens here if you don't make the change?

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -709,7 +708,7 @@ private void checkTableFunctionInModel(Class<?> clazz) {
       CalciteConnection connection = (CalciteConnection)
           driver.connect("jdbc:calcite:", new Properties());
       SchemaPlus rootSchema = connection.getRootSchema();
-      rootSchema.add("hr", new ReflectiveSchema(new HrSchema()));
+      rootSchema.add("hr", new ReflectiveSchemaWithoutRowCount(new HrSchema()));
```
mihaibudiu, line 711: same question: what happens if you don't change the schema

mihaibudiu: for many of these tests I would expect they would continue to work.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -5002,7 +5001,7 @@ public boolean moveNext() {
 
     CalciteAssert.that()
         .withSchema("s",
-            new ReflectiveSchema(
+            new ReflectiveSchemaWithoutRowCount(
```
mihaibudiu, line 5004: this test even seems to be disabled, why do you need to change it?

jduo:
@mihaibudiu , I've reduced the number of tests changed now.

mihaibudiu (approved):

jduo:
Hi @mihaibudiu ,
Just wondering if there are any other steps needed to get this one merged. Thanks.

mihaibudiu:
If there are no other comments we can merge this. You would have to squash the commits into a single one with the right title.

mihaibudiu:
I will merge once the CI passes.

PR #3635, opened 2024-01-18
[CALCITE-6210] Cast to VARBINARY causes an assertion failure

mihaibudiu:
I have also changed the way strings with non-ASCII characters are unparsed since that would lead to the failure of the tests that parse/unparse/reparse the query. This suggests that the previous way to unparse was not preserving the semantics of the original query, so it was incorrect.

The documentation never said how strings are converted to BINARY values, so I documented that as well. The semantics I chose matches Postgres and MySQL.


mihaibudiu:
This bugfix PR is ready for review.

core/src/main/java/org/apache/calcite/rex/RexUtil.java
```diff
@@ -1683,7 +1683,7 @@ public static boolean isLosslessCast(RexNode node) {
    *
    * @param source source type
    * @param target target type
-   * @return true iff the conversion is a loss-less cast
+   * @return true if the conversion is a loss-less cast
```
JiajunBernoulli, line 1686: `iff` mean `if and only if`

mihaibudiu: I know, but this function does not "true" if and only if, it returns it only "if", because it is an aproximation

gortiz:
I have a couple of questions about this PR.

## does it actually fix the issue?

The issue was the assertion error when casting STRING to VARBINARY. I'm not an expert on Apache Calcite, so I may be wrong, but it looks to me that this is only fixing the issue in the Enumerable convention.

What about other cases? For example in Apache Pinot we found this issue when executing `SqlToRelConverted.trimUnusedFields`. Specifically, when the expression is a literal like `CAST('sometext' as VARBINARY)`.

As said, I'm a noob here, but I don't see a call to `RexToLixTranslator.translateCast` in our stacktrace.

## how can we change the implementation used?

Specifically, this PR defines an arbitrary way to transform from String to Binary, but other systems may want to apply different logic to do so. In Pinot case we apply this cast by interpreting the String as hex codified. I guess we could override the implementation of `stringToBinary`, but I'm not sure.

mihaibudiu:
These are very good questions. Indeed, this PR the "Enumerable" implementation. But that implementation is special, because it is the one used when Calcite optimizations simplify a query plan. In particular, the optimization rule PROJECT_REDUCE_EXPRESSIONS uses this code to simplify constant expressions.

As far as I understand, the behavior of built-in SQL functions and casts in Calcite is *not* configurable to depend on the dialect being compiled. 

If you want a different behavior of these casts in different dialects the only solution may be to never let Calcite simplify the code (assuming that is possible - I don't understand all the paths on which code is simplified, some simplifications are used even when you do not use PROJECT_REDUCE_EXPRESSIONS). If you can avoid Calcite optimizing these cases, then the cast implementation is purely a question of what your runtime chooses to do.

Another alternative would be to indeed "override" the function I wrote stringToBinary, but I don't see how you can do this in a dialect-dependent version.

Perhaps someone more knowledgeable (@julianhyde?) about the Calcite architecture can answer this question: "can the behavior of a cast made to be dialect-dependent?" If yes, what is the right way to do it?

julianhyde:
Can we have design discussions in jira?

gortiz:
I've tried, but for some reason I my credentials in Apache Jira are not valid anymore. Maybe something got messed up when I became Pinot Contributor or maybe I'm just unable to remember my password. I'll try to solve the problem and continue the discussion there

julianhyde:
@gortiz If necessary, you could create a new jira account. We lean rather heavily on Jira in Calcite because Jira provides a permanent record of design discussion that can be searched and cross-referenced.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -709,6 +709,24 @@ void testCastToExactNumeric(CastType castType, SqlOperatorFixture f) {
         "CHAR(1) NOT NULL", "3");
   }
 
+  /**
+   * Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6210">
+   * Cast to VARBINARY causes an assertion failure</a>. */
+  @Test public void testVarbinaryCast() {
```
jduo, line 715: Check a null string as well (eg CAST(CAST(NULL AS VARCHAR) AS VARINARY)

mihaibudiu:
This PR implements a semantics for casting strings to BINARY values.
I think that the discussion related to the semantics is relatively settled, but there are no approvals. 
I think it's a clear improvement (converts a crash into a well-defined behavior), so it would be nice to have for 1.37.

gortiz:
+1. I think it would be great to be able to change the casting semantics, but that is a more complex problem and the correct solution would be to solve the error right now and discuss about the casting extensibility in the future.

PR #3636, opened 2024-01-18
[CALCITE-6211] SUBSTRING with Integer.MIN_VALUE as a second parameter raise unexpected exception

zstan:


mihaibudiu (approved):
I am not sure whether the first added test should have a comment indicating the issue resolved. Usually that's done in a JavaDoc in a separate test. I personally would not mind a line comment there.

JiajunBernoulli:
Here is same issue:
https://github.com/apache/calcite/blob/3d66b346ff2dcf23b58a9fcb144f40600cce7b2c/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java#L1041

Can you fix it in the PR?

zstan:
@mihaibudiu @JiajunBernoulli done, thanks !

mihaibudiu (approved):

PR #3640, opened 2024-01-19
[CALCITE-6032] Multilevel correlated query is failing in RelDecorrela…

HanumathRao:
…tor code path

This pull request includes the following modifications to address the Null Pointer Exception (NPE) when decorrelating a multi-level correlated subquery:

- Eliminated the TrimUnusedFields in the Prepare phase, as the optimize call (in the same function) handles the trimming of unused fields as part of the standard program. Additionally, this code path is brought in line with the Optimize code path in the PlannerImpl, where the TrimUnusedFields flag is set to False. This ensures consistency between both code paths.

- Implemented a call to the getCorrelationUse function in certain code paths. This call rectifies the rel node tree to standardize the correlated variables.

- The fix introduced in response to [CALCITE-5683](https://issues.apache.org/jira/browse/CALCITE-5683) also addresses the identified issue in this JIRA.

All of the aforementioned code changes are necessary to resolve the reported problem.

result from sqlfiddle postgresql: https://sqlfiddle.com/postgresql/online-compiler?id=98100522-acf6-46b3-9b19-eb14b7015078

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
@@ -271,7 +271,7 @@ public abstract class RelOptUtil {
   }
 
   /**
-   * Returns a set of variables used by a relational expression or its
+   * Returns the set of variables used by a relational expression or its
    * descendants.
    *
    * <p>The set may contain "duplicates" (variables with different ids that,
@@ -284,6 +284,24 @@ public abstract class RelOptUtil {
     CorrelationCollector visitor = new CorrelationCollector();
     rel.accept(visitor);
     return visitor.vuv.variables;
+  }
+
+  /**
+   * Returns the set of variables used by the given list of sub-queries and its descendants.
+   *
+   * @param subQueries The sub-queries containing correlation variables
+   * @return A list of correlation identifiers found within the sub-queries.
+   *          The type of the [CorrelationId] parameter corresponds to
+   *          {@link org.apache.calcite.rex.RexCorrelVariable#id}.
+   */
+  public static Set<CorrelationId> getVariablesUsed(List<RexSubQuery> subQueries) {
+    // Internally this function calls getVariablesUsed on a RelNode to get all the
+    // correlated variables in that RelNode
+    Set<CorrelationId> correlationIds = new HashSet<>();
+    for (RexSubQuery subQ : subQueries) {
+      correlationIds.addAll(getVariablesUsed(subQ.rel));
+    }
+    return correlationIds;
   }
 
   /** Finds which columns of a correlation variable are used within a
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -5491,18 +5491,25 @@ public class SqlToRelConverter {
               builder.add(convertExpression(node));
             }
             final ImmutableList<RexNode> list = builder.build();
+            RelNode rel = root.rel;
+            // Fix the correlation namespaces and de-duplicate the correlation variables.
+            CorrelationUse correlationUse = getCorrelationUse(this, root.rel);
+            if (correlationUse != null) {
+              rel = correlationUse.r;
+            }
+
             switch (kind) {
             case IN:
-              return RexSubQuery.in(root.rel, list);
+              return RexSubQuery.in(rel, list);
             case NOT_IN:
               return rexBuilder.makeCall(SqlStdOperatorTable.NOT,
-                  RexSubQuery.in(root.rel, list));
+                  RexSubQuery.in(rel, list));
             case SOME:
-              return RexSubQuery.some(root.rel, list,
+              return RexSubQuery.some(rel, list,
                   (SqlQuantifyOperator) call.getOperator());
             case ALL:
               return rexBuilder.makeCall(SqlStdOperatorTable.NOT,
-                  RexSubQuery.some(root.rel, list,
+                  RexSubQuery.some(rel, list,
                       negate((SqlQuantifyOperator) call.getOperator())));
             default:
               throw new AssertionError(kind);
@@ -5515,6 +5522,12 @@ public class SqlToRelConverter {
           query = Iterables.getOnlyElement(call.getOperandList());
           root = convertQueryRecursive(query, false, null);
           RelNode rel = root.rel;
+          // Fix the correlation namespaces and de-duplicate the correlation variables.
+          CorrelationUse correlationUse = getCorrelationUse(this, root.rel);
+          if (correlationUse != null) {
+            rel = correlationUse.r;
+          }
+
           while (rel instanceof Project
               || rel instanceof Sort
               && ((Sort) rel).fetch == null
@@ -5533,7 +5546,13 @@ public class SqlToRelConverter {
           call = (SqlCall) expr;
           query = Iterables.getOnlyElement(call.getOperandList());
           root = convertQueryRecursive(query, false, null);
-          return RexSubQuery.scalar(root.rel);
+          rel = root.rel;
+          // Fix the correlation namespaces and de-duplicate the correlation variables.
+          correlationUse = getCorrelationUse(this, root.rel);
+          if (correlationUse != null) {
+            rel = correlationUse.r;
+          }
+          return RexSubQuery.scalar(rel);
 
         case ARRAY_QUERY_CONSTRUCTOR:
           call = (SqlCall) expr;
--- a/core/src/test/java/org/apache/calcite/test/JdbcTest.java
+++ b/core/src/test/java/org/apache/calcite/test/JdbcTest.java
@@ -8271,6 +8271,75 @@ public class JdbcTest {
   }
 
   /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6032">[CALCITE-6032]
+   * NullPointerException in Reldecorrelator for a Multi level correlated subquery</a>. */
+  @Test void testMultiLevelDecorrelation() throws Exception {
+    String hsqldbMemUrl = "jdbc:hsqldb:mem:.";
+    Connection baseConnection = DriverManager.getConnection(hsqldbMemUrl);
+    Statement baseStmt = baseConnection.createStatement();
+    baseStmt.execute("create table invoice (inv_id integer, col1\n"
+        + "integer, inv_amt integer)");
+    baseStmt.execute("create table item(item_id integer, item_amt\n"
+        + "integer, item_col1 integer, item_col2 integer, item_col3\n"
+        + "integer,item_col4 integer )");
+    baseStmt.execute("INSERT INTO invoice VALUES (1, 1, 1)");
+    baseStmt.execute("INSERT INTO invoice VALUES (2, 2, 2)");
+    baseStmt.execute("INSERT INTO invoice VALUES (3, 3, 3)");
+    baseStmt.execute("INSERT INTO item values (1, 1, 1, 1, 1, 1)");
+    baseStmt.execute("INSERT INTO item values (2, 2, 2, 2, 2, 2)");
+    baseStmt.close();
+    baseConnection.commit();
+
+    Properties info = new Properties();
+    info.put("model",
+        "inline:"
+            + "{\n"
+            + "  version: '1.0',\n"
+            + "  defaultSchema: 'BASEJDBC',\n"
+            + "  schemas: [\n"
+            + "     {\n"
+            + "       type: 'jdbc',\n"
+            + "       name: 'BASEJDBC',\n"
+            + "       jdbcDriver: '" + jdbcDriver.class.getName() + "',\n"
+            + "       jdbcUrl: '" + hsqldbMemUrl + "',\n"
+            + "       jdbcCatalog: null,\n"
+            + "       jdbcSchema: null\n"
+            + "     }\n"
+            + "  ]\n"
+            + "}");
+
+    Connection calciteConnection =
+        DriverManager.getConnection("jdbc:calcite:", info);
+
+    String statement = "SELECT Sum(invoice.inv_amt * (\n"
+        + "              SELECT max(mainrate.item_id + mainrate.item_amt)\n"
+        + "              FROM   item AS mainrate\n"
+        + "              WHERE  mainrate.item_col1 is not null\n"
+        + "              AND    mainrate.item_col2 is not null\n"
+        + "              AND    mainrate.item_col3 = invoice.col1\n"
+        + "              AND    mainrate.item_col4 = (\n"
+        + "                        SELECT max(cr.item_col4)\n"
+        + "                        FROM   item AS cr\n"
+        + "                        WHERE  cr.item_col3 = mainrate.item_col3\n"
+        + "                                AND    cr.item_col1 =\n"
+        + "mainrate.item_col1\n"
+        + "                                AND    cr.item_col2 =\n"
+        + "mainrate.item_col2 \n"
+        + "                                AND    cr.item_col4 <=\n"
+        + "invoice.inv_id))) AS invamount,\n"
+        + "count(*)       AS invcount\n"
+        + "FROM    invoice\n"
+        + "WHERE  invoice.inv_amt < 10 AND  invoice.inv_amt > 0";
+    ResultSet rs = calciteConnection.prepareStatement(statement).executeQuery();
+    assert rs.next();
+    assertEquals(rs.getInt(1), 10);
+    assertEquals(rs.getInt(2), 3);
+    assert !rs.next();
+    rs.close();
+    calciteConnection.close();
+  }
+
+  /** Test case for
    * <a href="https://issues.apache.org/jira/browse/CALCITE-5414">[CALCITE-5414]</a>
    * Convert between standard Gregorian and proleptic Gregorian calendars for
    * literal dates in local time zone. */
--- a/core/src/main/java/org/apache/calcite/rex/RexUtil.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexUtil.java
@@ -2854,6 +2854,29 @@ public class RexUtil {
     }
   }
 
+  /** Visitor that collects all the top level SubQueries {@link RexSubQuery}
+   *  in a projection list of a given {@link Project}.*/
+  public static class SubQueryCollector extends RexVisitorImpl<Void> {
+    private List<RexSubQuery> subQueries;
+    private SubQueryCollector() {
+      super(true);
+      this.subQueries = new ArrayList<>();
+    }
+
+    @Override public Void visitSubQuery(RexSubQuery subQuery) {
+      subQueries.add(subQuery);
+      return null;
+    }
+
+    public static List<RexSubQuery> collect(Project project) {
+      SubQueryCollector subQueryCollector = new SubQueryCollector();
+      for (RexNode node : project.getProjects()) {
+        node.accept(subQueryCollector);
+      }
+      return subQueryCollector.subQueries;
+    }
+  }
+
   /** Visitor that throws {@link org.apache.calcite.util.Util.FoundOne} if
    * applied to an expression that contains a {@link RexSubQuery}. */
   public static class SubQueryFinder extends RexVisitorImpl<Void> {
--- a/core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
@@ -55,6 +55,7 @@ import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexPermuteInputsShuttle;
 import org.apache.calcite.rex.RexProgram;
+import org.apache.calcite.rex.RexSubQuery;
 import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.rex.RexVisitor;
 import org.apache.calcite.sql.SqlExplainFormat;
@@ -489,7 +490,24 @@ public class RelFieldTrimmer implements ReflectiveVisitor {
         ord.e.accept(inputFinder);
       }
     }
-    ImmutableBitSet inputFieldsUsed = inputFinder.build();
+
+    // Collect all the SubQueries in the projection list.
+    List<RexSubQuery> subQueries = RexUtil.SubQueryCollector.collect(project);
+    // Get all the correlationIds present in the SubQueries
+    Set<CorrelationId> correlationIds = RelOptUtil.getVariablesUsed(subQueries);
+    ImmutableBitSet requiredColumns = ImmutableBitSet.of();
+    if (correlationIds.size() > 0) {
+      assert correlationIds.size() == 1;
+      // Correlation columns are also needed by SubQueries, so add them to inputFieldsUsed.
+      requiredColumns = RelOptUtil.correlationColumns(correlationIds.iterator().next(), project);
+    }
+
+    ImmutableBitSet finderFields = inputFinder.build();
+
+    ImmutableBitSet inputFieldsUsed = ImmutableBitSet.builder()
+        .addAll(requiredColumns)
+        .addAll(finderFields)
+        .build();
 
     // Create input with trimmed columns.
     TrimResult trimResult =


core/src/main/java/org/apache/calcite/prepare/Prepare.java
```diff
@@ -293,13 +293,6 @@ public PreparedResult prepareSql(
       root = root.withRel(decorrelate(sqlToRelConverter, sqlQuery, root.rel));
     }
 
-    if (configHolder.get().isTrimUnusedFields()) {
```
mihaibudiu, line 296: This config feature seems to only be used here.
So perhaps it can be removed completely?
My question is whether the bug is in the trimUnusedFields method, or is it in invoking it here.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -5491,18 +5491,24 @@ ImmutableList<RelNode> retrieveCursors() {
               builder.add(convertExpression(node));
             }
             final ImmutableList<RexNode> list = builder.build();
+            RelNode rel = root.rel;
+            CorrelationUse correlationUse = getCorrelationUse(this, root.rel);
```
mihaibudiu, line 5495: Could you add a comment explaining what happens here?

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -8217,6 +8217,72 @@ private void checkGetTimestamp(Connection con) throws SQLException {
         .returns("EXPR$0=[1, 1.1]\n");
   }
 
+  @Test void testMultiLevelDecorrelation() throws Exception {
```
mihaibudiu, line 8220: Should this function have a JavaDoc describing the issue fixed?

core/src/test/java/org/apache/calcite/test/enumerable/EnumerableCorrelateTest.java
```diff
@@ -143,13 +143,13 @@ class EnumerableCorrelateTest {
         .query(
             "select empid, name from emps e where exists (select 1 from depts d where d.deptno=e.deptno)")
         .explainContains(""
-            + "EnumerableCalc(expr#0..3=[{inputs}], empid=[$t0], name=[$t2])\n"
-            + "  EnumerableCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{1}])\n"
-            + "    EnumerableCalc(expr#0..4=[{inputs}], proj#0..2=[{exprs}])\n"
-            + "      EnumerableTableScan(table=[[s, emps]])\n"
+            + "EnumerableCalc(expr#0..5=[{inputs}], empid=[$t0], name=[$t2])\n  "
```
mihaibudiu, line 146: Please move the spaces at the end of the previous line here.
Same in the subsequent lines.

mihaibudiu: Is this case even changed?

core/src/test/resources/sql/sub-query.iq
```diff
@@ -2225,7 +2225,7 @@ where exists
 # following plan (two scans of EMP table instead of three).
 EnumerableCalc(expr#0..2=[{inputs}], ENAME=[$t1])
   EnumerableHashJoin(condition=[=($2, $3)], joinType=[semi])
-    EnumerableCalc(expr#0..7=[{inputs}], expr#8=[IS NOT NULL($t3)], expr#9=[CAST($t3):INTEGER], expr#10=[0], expr#11=[CASE($t8, $t9, $t10)], proj#0..1=[{exprs}], $f3=[$t11])
+    EnumerableCalc(expr#0..7=[{inputs}], expr#8=[IS NOT NULL($t3)], expr#9=[CAST($t3):INTEGER], expr#10=[0], expr#11=[CASE($t8, $t9, $t10)], proj#0..1=[{exprs}], $f8=[$t11])
```
mihaibudiu, line 2228: This seems to be a test for a prior bug.
Are you saying that the result was incorrect and is correct now?

mihaibudiu (commented):
There are lots of issues open related to correlated subqueries, but this fix does not seem to address them.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
HanumathRao: Sure, I can add a comment here, but I believe this function/pattern is used elsewhere so thought adding only here might not be  relevant.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
HanumathRao: Yeah sure, I can add with the reference to the CALCITE JIRA.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
mihaibudiu: I am not familiar with this part of the code, so I need help to understand why this works.
Maybe other reviewers won't need this help.

core/src/test/resources/sql/sub-query.iq
HanumathRao: There is no change in the result, only plan field number is changed (isn't it ?)

core/src/main/java/org/apache/calcite/prepare/Prepare.java
HanumathRao: As explained it in the PR description, calling trimUnusedFields is different in different paths (In the Prepare path we call the trimUnusedFields first before optimization but seems like the preferred way as implemented in  PlannerImpl is to skip the trimUnusedFields first and call later.

HanumathRao:
> There are lots of issues open related to correlated subqueries, but this fix does not seem to address them.

This fix is to address specific case discussed in this JIRA. This PR is not for handling all the correlated subquery issues which might not be relevant to the issue reported in this JIRA.

core/src/main/java/org/apache/calcite/prepare/Prepare.java
mihaibudiu: You are fixing a bug. If I understand correctly the bug happens here, because trimUnusedFields returns an incorrect plan. Is the bug in trimUnusedFields? Or does trimUnusedFields need some invariants being called, and the invariants are not established in this point? If the bug is in trimUnusedFields, this change only hides the bug but does not really fix it. That is my first question.

I am not sure what you mean by "preferred". Preferences are subjective, but whether this is a bug or not is not a preference.

HanumathRao: The TrimUnusedFields function currently lacks comprehensive handling for cases involving correlation variables. It appears that the trimUnusedFields in the Prepare call was retained for legacy use cases, where a config object with trimUnusedFields can be set to True. However, upon inspecting the code at [link](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/prepare/PlannerImpl.java#L265), trimUnusedFields is set to False, and a series of optimizations are applied to the RelNodeTree. Subsequently, the TrimUnusedFields Program is executed on the RelNodeTree in a later part of the code path. Therefore, the proposed change aims to synchronize these two code paths.

It's worth noting that addressing this issue is complex due to the interplay of seemingly disparate changes that contribute to encountering various problems. While I appreciate your questions and engagement, I believe consulting an expert in this domain would be beneficial to ensure the correctness of my proposal and explore any potential alternative solutions to resolve this issue. Thank you for your understanding.

HanumathRao:
@mihaibudiu Appreciate the prompt review. I've addressed the testcase by including a Javadoc comment explaining the failure. Additionally, I've added comments in the code where relevant.

@libenchao and @rubenada, your review of the proposed fix would be greatly appreciated. Given your previous examination of the code fix for JIRA [CALCITE-5683](https://issues.apache.org/jira/browse/CALCITE-5683), your insights will be valuable in ensuring the quality of the solution.

HanumathRao:
@mihaibudiu, Please let me know if you have any further questions or clarifications on these changes.

@libenchao and @rubenada, a friendly reminder to review the PR changes.

HanumathRao:
@mihaibudiu, Please let me know if you have any further questions or clarifications on these changes.

@libenchao and @rubenada, a friendly reminder to review the PR changes.

rubenada:
@HanumathRao  I'll try to take a look as soon as I have a bit of time.
In the meanwhile, could you please check if your patch also fixes the cases mentioned in CALCITE-5390 ? (also NPE related to decorrelation). If it doesn't, no problem, I'd like just to confirm if CALCITE-6032 and CALCITE-5390 are / aren't due to same root cause.

HanumathRao:
> @HanumathRao I'll try to take a look as soon as I have a bit of time. In the meanwhile, could you please check if your patch also fixes the cases mentioned in [CALCITE-5390](https://issues.apache.org/jira/browse/CALCITE-5390) ? (also NPE related to decorrelation). If it doesn't, no problem, I'd like just to confirm if [CALCITE-6032](https://issues.apache.org/jira/browse/CALCITE-6032) and [CALCITE-5390](https://issues.apache.org/jira/browse/CALCITE-5390) are / aren't due to same root cause.

Thanks @rubenada for the reply. I tried out the query in CALCITE-5390 (https://issues.apache.org/jira/browse/CALCITE-5390) and they are not the same root cause. I can take a look at it later (post this JIRA merge), if no one is working on it. thanks

rubenada:
@HanumathRao  could you please elaborate why we need in this PR the changes in `CassandraSchema.java` and `Prepare.java`? Is is related to the decorrelation problem, or could they be part of a separate issue?

HanumathRao:
> @HanumathRao could you please elaborate why we need in this PR the changes in `CassandraSchema.java` and `Prepare.java`? Is is related to the decorrelation problem, or could they be part of a separate issue?

1. Changes to Prepare.java file, 
Yes in a subtle way. The trimUnusedField ,in the prepareSql function call path, is called conditionally based on the configHolder's isTrimUnusedFields flag and due to this it results in trimUnusedFields getting called before decorrelateQuery is called. However, the same functionality in PlannerImpl.rel code path (most calcite users seem to use it), trimUnusedFields is set to false and this makes sure the decorrelateQuery is called before the trimUnusedFields.

In both the cases the RelFieldTrimmer program is called in the optimize phase of the query, so essentially fields will get trimmed, but in the case of Prepare (which I believe is less used by users) doing so will result in a tree which will cause other issues due to lack of comprehensive handling of correlate subqueries.

Essentially I am trying to make both the paths to behave same.

2. The changes in CassandraSchema.java are done because earlier CassandraSchema Test code loaded the materializations using TRIMMED Hook. As this code path is no more available, I changed it to load on CONVERTED Hook. This way the materializations are available for the optimizations, otherwise the materializations are not loaded and MaterializedView optimization is skipped. 

Hope the above description made it clear. Please let me know if you have any further questions/suggestions.

rubenada:
> The trimUnusedField ,in the prepareSql function call path, is called conditionally based on the configHolder's isTrimUnusedFields flag and due to this it results in **trimUnusedFields getting called before decorrelateQuery** is called.

Maybe I'm missing something, but isn't `decorrelateQuery` called before `trimUnusedFields` in `prepareSql` ? (so either we trim after decorrelate, or we don't trim at all, depending on the configHolder)
```
    if (this.context.config().forceDecorrelate()) {
      // Sub-query decorrelation.
      root = root.withRel(decorrelate(sqlToRelConverter, sqlQuery, root.rel));
    }

    if (configHolder.get().isTrimUnusedFields()) {
      // Trim unused fields.
      root = trimUnusedFields(root);

      Hook.TRIMMED.run(root.rel);
    }
```

HanumathRao:
> > The trimUnusedField ,in the prepareSql function call path, is called conditionally based on the configHolder's isTrimUnusedFields flag and due to this it results in **trimUnusedFields getting called before decorrelateQuery** is called.
> 
> Maybe I'm missing something, but isn't `decorrelateQuery` called before `trimUnusedFields` in `prepareSql` ? (so either we trim after decorrelate, or we don't trim at all, depending on the configHolder)
> 
> ```
>     if (this.context.config().forceDecorrelate()) {
>       // Sub-query decorrelation.
>       root = root.withRel(decorrelate(sqlToRelConverter, sqlQuery, root.rel));
>     }
> 
>     if (configHolder.get().isTrimUnusedFields()) {
>       // Trim unused fields.
>       root = trimUnusedFields(root);
> 
>       Hook.TRIMMED.run(root.rel);
>     }
> ```

Yeah, your observation is right, and there's a subtle issue that requires attention in this case. 

Before the decorrelate function is invoked, the plan (for the added test in the PR) appears as follows:
```
LogicalAggregate(group=[{}], INVAMOUNT=[SUM($0)], INVCOUNT=[COUNT()])
  LogicalProject($f0=[*($2, $SCALAR_QUERY({
LogicalAggregate(group=[{}], EXPR$0=[MAX($0)])
  LogicalProject($f0=[+($0, $1)])
    LogicalFilter(condition=[AND(IS NOT NULL($2), IS NOT NULL($3), =($4, $cor0.C5633_503), =($5, $SCALAR_QUERY({
LogicalAggregate(group=[{}], EXPR$0=[MAX($0)])
  LogicalProject(C5633_2186=[$5])
    LogicalFilter(condition=[AND(=($4, $cor1.C5633_2187), =($2, $cor1.C5633_2184), =($3, $cor1.C5633_2185), <=($5, $cor0.C5633_485))])
      JdbcTableScan(table=[[BASEJDBC, T1704]])
})))], variablesSet=[[$cor1]])
      JdbcTableScan(table=[[BASEJDBC, T1704]])
}))])
    LogicalFilter(condition=[AND(<($2, 10), >($2, 0))])
      JdbcTableScan(table=[[BASEJDBC, INVOICE]])
 ```
 
The decorrelate function before the trimFields operation is essentially a no-op. This is because decorrelation only occurs when a Correlate node is created, which is determined by a triggering check.
 
 ```
     if (!corelMap.hasCorrelation()) {
      return rootRel;
    }
```
The hasCorrelation function returns true only if mapCorToCorRel is nonEmpty. However, in this scenario, mapCorToCorRel is empty because no Correlate node exists. Therefore, the decorrelate function essentially does nothing.

However, after the trimUnusedFields operation is executed on the tree, it prunes certain fields that are essential for correlated variables. Consequently, a new node, LogicalProject(C5633_505=[$2]), is introduced.

```
LogicalAggregate(group=[{}], INVAMOUNT=[SUM($0)], INVCOUNT=[COUNT()])
  LogicalProject($f0=[*($0, $SCALAR_QUERY({
LogicalAggregate(group=[{}], EXPR$0=[MAX($0)])
  LogicalProject($f0=[+($0, $1)])
    LogicalFilter(condition=[AND(IS NOT NULL($2), IS NOT NULL($3), =($4, $cor0.C5633_503), =($5, $SCALAR_QUERY({
LogicalAggregate(group=[{}], EXPR$0=[MAX($0)])
  LogicalProject(C5633_2186=[$5])
    LogicalFilter(condition=[AND(=($4, $cor1.C5633_2187), =($2, $cor1.C5633_2184), =($3, $cor1.C5633_2185), <=($5, $cor0.C5633_485))])
      JdbcTableScan(table=[[BASEJDBC, T1704]])
})))], variablesSet=[[$cor1]])
      JdbcTableScan(table=[[BASEJDBC, T1704]])
}))])
    LogicalFilter(condition=[SEARCH($0, Sarg[(0..10)])])
      LogicalProject(C5633_505=[$2]). --- a new node introduced by trimUnusedFields
        JdbcTableScan(table=[[BASEJDBC, INVOICE]])
```

The subsequent optimization of this tree leads to an exception because the tree is no longer correct.

This issue is specific to the prepareSql path and is not observed during the plannerImpl rel path. [link](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/prepare/PlannerImpl.java#L265).

Two potential solutions have been identified:

1. Fix the RelTrimmer to avoid generating the problematic node.
2. Ensure consistency between both code paths by refraining from calling the RelTrimmer in the prepareSql path.
I chose the second option, as it seems logical to call the RelTrimmer only after the SubQueryRemovalRule has been applied.

Please let me know if it is clear and thank you for reviewing the changes.

rubenada:
Thanks for your patience and the thorough explanation @HanumathRao , I really appreciate it.

I understand a bit more the situation now... and I beg to differ.
IMHO fix 2 is just "sweeping the problem under the rug", we could avoid the problem in `Prepare.java`, but the underlying issue still remains. Plus, before removing this [Chesterton's fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence), we would need to understand the reasons why field trimming was included in there in the first place. Also, removing the trimming can have undesired consequences (e.g. some projects might have fields which are expensive to retrieve/compute, so they should only be part of the final plan if they are really used, otherwise they should be trimmed; so this change would lead to a performance regression).

I think fix 1 is the right way to go (although I understand it will be harder). Field trimming is quite a powerful tool (not without some bugs), which can be used by downstream projects (either via SqlToRelConverter#trimUnusedFields or directly via RelFieldTrimmer, since both are public); so if we find a bug on it, I think it would be better to try to fix it.

HanumathRao:
> Thanks for your patience and the thorough explanation @HanumathRao , I really appreciate it.
> 
> I understand a bit more the situation now... and I beg to differ. IMHO fix 2 is just "sweeping the problem under the rug", we could avoid the problem in `Prepare.java`, but the underlying issue still remains. Plus, before removing this [Chesterton's fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence), we would need to understand the reasons why field trimming was included in there in the first place. Also, removing the trimming can have undesired consequences (e.g. some projects might have fields which are expensive to retrieve/compute, so they should only be part of the final plan if they are really used, otherwise they should be trimmed; so this change would lead to a performance regression).
> 
> I think fix 1 is the right way to go (although I understand it will be harder). Field trimming is quite a powerful tool (not without some bugs), which can be used by downstream projects (either via SqlToRelConverter#trimUnusedFields or directly via RelFieldTrimmer, since both are public); so if we find a bug on it, I think it would be better to try to fix it.



> Thanks for your patience and the thorough explanation @HanumathRao , I really appreciate it.
> 
> I understand a bit more the situation now... and I beg to differ. IMHO fix 2 is just "sweeping the problem under the rug", we could avoid the problem in `Prepare.java`, but the underlying issue still remains. Plus, before removing this [Chesterton's fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence), we would need to understand the reasons why field trimming was included in there in the first place. Also, removing the trimming can have undesired consequences (e.g. some projects might have fields which are expensive to retrieve/compute, so they should only be part of the final plan if they are really used, otherwise they should be trimmed; so this change would lead to a performance regression).
> 
> I think fix 1 is the right way to go (although I understand it will be harder). Field trimming is quite a powerful tool (not without some bugs), which can be used by downstream projects (either via SqlToRelConverter#trimUnusedFields or directly via RelFieldTrimmer, since both are public); so if we find a bug on it, I think it would be better to try to fix it.


Thank you, @rubenada, for your comments. I have addressed the issue in RelFieldTrimmer to ensure that correlated variables in the subqueries are not trimmed. Nevertheless, I maintain my stance that executing RelFieldTrimmer in this context might be unnecessary. Since RelFieldTrimmer is run as a separate program after the SubQueryRemoval Rule ([link](https://github.com/apache/calcite/blob/8fd9518887302af43e7e74cdb155474dcebfc9f5/core/src/main/java/org/apache/calcite/tools/Programs.java#L290)), calling it in Prepare.java may not yield any benefits.

Kindly review the changes at your earliest convenience.

core/src/test/java/org/apache/calcite/test/enumerable/EnumerableCorrelateTest.java
```diff
@@ -142,10 +142,10 @@ class EnumerableCorrelateTest {
     tester(false, new HrSchema())
         .query(
             "select empid, name from emps e where exists (select 1 from depts d where d.deptno=e.deptno)")
-        .explainContains(""
-            + "EnumerableCalc(expr#0..5=[{inputs}], empid=[$t0], name=[$t2])\n"
-            + "  EnumerableCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{1}])\n"
-            + "    EnumerableTableScan(table=[[s, emps]])\n"
+        .explainContains("EnumerableCalc(expr#0..3=[{inputs}], empid=[$t0], name=[$t2])\n "
```
mihaibudiu, line 145: was the previous plan incorrect?
If not, can you provide a test case for which the previous plan was incorrect, but it is now correct?

HanumathRao: There is no change in the plan, it was just an indentation fix. I have reverted it in the latest commit just to avoid any confusion. Thanks

mihaibudiu: The second half of my question stands.
If the RelFieldTrimmer was wrong, is there a test case where you can show the difference?

HanumathRao: Sorry, I assumed that you have looked at the new test case added in JdbcTest (testMultiLevelDecorrelation). If there is no RelFieldTrimmer fix then this new testcase fails with same exception as reported in the JIRA.

mihaibudiu: Great, thank you, I was only looking at the last commit.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -286,6 +286,22 @@ public static Set<CorrelationId> getVariablesUsed(RelNode rel) {
     return visitor.vuv.variables;
   }
 
+  /**
+   * Returns a set of variables used by the given list of sub-queries and its descendants.
```
mihaibudiu, line 290: "a" set is imprecise. I would say "the set"

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -286,6 +286,22 @@ public static Set<CorrelationId> getVariablesUsed(RelNode rel) {
...
+   *
+   * <p> Internally it uses the {@link org.apache.calcite.plan.RelOptUtil#getVariablesUsed}
+   * to get all the variables used in the relational expression of a given sub-query.
+   *
+   * <p> The item type is same as {@link org.apache.calcite.rex.RexCorrelVariable#id}.
```
mihaibudiu, line 295: I really don't know what item you are referring to here.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -286,6 +286,22 @@ public static Set<CorrelationId> getVariablesUsed(RelNode rel) {
     return visitor.vuv.variables;
   }
 
+  /**
+   * Returns a set of variables used by the given list of sub-queries and its descendants.
+   *
+   * <p> Internally it uses the {@link org.apache.calcite.plan.RelOptUtil#getVariablesUsed}
```
mihaibudiu, line 292: In general I would refrain from describing *implementation* in JavaDoc, unless it is very relevant. JavaDoc is about what the function does, not how. This comment can be a regular comment inside the function, although the function is pretty clear as is.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -8217,6 +8217,75 @@ private void checkGetTimestamp(Connection con) throws SQLException {
...
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6032">[CALCITE-6032]
+   * NullPointerException in Reldecorrelator for a Multi level correlated subquery</a>. */
+  @Test void testMultiLevelDecorrelation() throws Exception {
+    String hsqldbMemUrl = "jdbc:hsqldb:mem:.";
+    Connection baseConnection = DriverManager.getConnection(hsqldbMemUrl);
+    Statement baseStmt = baseConnection.createStatement();
+    baseStmt.execute("create table invoice (c5633_485 integer, c5633_503\n"
```
mihaibudiu, line 8227: this code looks like it's machine generated.
This makes it hard to read by a human.

mihaibudiu (approved):
This looks good, I left a few comments. I am approving, hoping that you will address them.
If anyone else wants to review this, it's great, otherwise we should plan to merge it.
Perhaps you can wait a couple more days before squashing the commits.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
HanumathRao: I used the same comment format as that of getVariablesUsed(RelNode rel) as both the functions return same output, In this context item type is a type parameter for Set<>

HanumathRao:
> This looks good, I left a few comments. I am approving, hoping that you will address them. If anyone else wants to review this, it's great, otherwise we should plan to merge it. Perhaps you can wait a couple more days before squashing the commits.

Thanks for the review. I fixed the review comments.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
mihaibudiu: Javadoc can document type parameters, use that format: https://www.baeldung.com/java-javadoc-generic-type-parameters

mihaibudiu (approved):
I still have one more improvement suggestion.
But you can also probably also squash the commits in preparation for merging.

HanumathRao:
> I still have one more improvement suggestion. But you can also probably also squash the commits in preparation for merging.

Thanks. Fixed the javadoc comment and squashed and re-based to latest.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -8270,6 +8270,75 @@ private void checkGetTimestamp(Connection con) throws SQLException {
...
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6032">[CALCITE-6032]
+   * NullPointerException in Reldecorrelator for a Multi level correlated subquery</a>. */
+  @Test void testMultiLevelDecorrelation() throws Exception {
+    String hsqldbMemUrl = "jdbc:hsqldb:mem:.";
+    Connection baseConnection = DriverManager.getConnection(hsqldbMemUrl);
+    Statement baseStmt = baseConnection.createStatement();
+    baseStmt.execute("create table invoice (inv_id integer, col1\n"
+        + "integer, inv_amt integer)");
+    baseStmt.execute("create table item(item_id integer, item_amt\n"
```
jinfengni, line 8282: Will it be better to use some existing table/schemas to reproduce this problem? For instance, the `hr` or `foodmart` schema?

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -8270,6 +8270,75 @@ private void checkGetTimestamp(Connection con) throws SQLException {
...
+            + "       jdbcCatalog: null,\n"
+            + "       jdbcSchema: null\n"
+            + "     }\n"
+            + "  ]\n"
+            + "}");
+
+    Connection calciteConnection =
+        DriverManager.getConnection("jdbc:calcite:", info);
+
+    String statement = "SELECT Sum(invoice.inv_amt * (\n"
```
jinfengni, line 8314: Can we simply the query to reproduce the issue?  Remove the unnecessary part that is not related to de-correlation logic.  

I tried the following query, using `foodmart`, and seems it hit the  similar stack trace.

```
SELECT SUM(p."min_scale" * (
        select max(e."salary" + e."salary")
        from FOODMART."employee" e
        where e."position_id" = p."position_id"
        AND     e."supervisor_id" = (select MAX(e2."supervisor_id")
                                 from FOODMART."employee" e2
                                 WHERE e."position_id" <= p."position_id"
                                 AND   e."last_name" = e2."last_name")
)), count(*) as cnt
from FOODMART."position" p;
```

```
Error: Error while executing SQL "SELECT SUM(p."min_scale" * (
        select max(e."salary" + e."salary")
        from FOODMART."employee" e
        where e."position_id" = p."position_id"
        AND     e."supervisor_id" = (select MAX(e2."supervisor_id")
                                 from FOODMART."employee" e2
                                 WHERE e."position_id" <= p."position_id"
                                 AND   e."last_name" = e2."last_name")
)), count(*) as cnt
from FOODMART."position" p": cm.mapCorToCorRel.get($cor2) (state=,code=0)
```

HanumathRao: Thank you, @jinfengni, for your review. While it's possible to reproduce the issue with the current schema, I opted to employ the same test case to ensure its resolution. Please inform me if you'd like to include this test case alongside the one I've added (although it may be redundant).

jinfengni: My view is it would be better to use a "minimal" query to reproduce the issue, and verify the fix, in particular in unit testcases.  Adding more components in the query will make it hard to understand what kind of features this query uses that actually hit the issue. 

For instance,  the following simplified query would hit the same stack trace:

```
SELECT SUM(p."min_scale" * (
        select max(e."salary")
        from FOODMART."employee" e
        where e."position_id" = p."position_id"
        AND     e."supervisor_id" = (select MAX(e2."supervisor_id")
                                 from FOODMART."employee" e2
                                 WHERE e."position_id" <= p."position_id")
))
from FOODMART."position" p;

Caused by: java.lang.NullPointerException: cm.mapCorToCorRel.get($cor2)
	at java.base/java.util.Objects.requireNonNull(Objects.java:336)
	at org.apache.calcite.sql2rel.RelDecorrelator.getCorRel(RelDecorrelator.java:929)
	at org.apache.calcite.sql2rel.RelDecorrelator.createValueGenerator(RelDecorrelator.java:818)
	at org.apache.calcite.sql2rel.RelDecorrelator.decorrelateInputWithValueGenerator(RelDecorrelator.java:1029)
	at org.apache.calcite.sql2rel.RelDecorrelator.maybeAddValueGenerator(RelDecorrelator.java:948)
	at org.apache.calcite.sql2rel.RelDecorrelator.decorrelateRel(RelDecorrelator.java:1154)
	at org.apache.calcite.sql2rel.RelDecorrelator.decorrelateRel(RelDecorrelator.java:1120)
```

So,  the issue would raise when mult-level nested subquery is referring to a column in outer query block. This may explain why the issue may be different from CALCITE-5390, where it uses a one-level nested subquery.

HanumathRao: Thanks @jinfengni I will look into CALCITE-5390 to understand a little bit more on the difference.

PR #3641, opened 2024-01-21
[CALCITE-6214] Remove DISTINCT in aggregate function if field is unique

JiajunBernoulli:
Here is JIRA ticket: https://issues.apache.org/jira/browse/CALCITE-6214

### DIFF
--- a/core/src/main/java/org/apache/calcite/tools/RelBuilder.java
+++ b/core/src/main/java/org/apache/calcite/tools/RelBuilder.java
@@ -2430,7 +2430,11 @@ public class RelBuilder {
     RelNode r = frame.rel;
     final List<AggregateCall> aggregateCalls = new ArrayList<>();
     for (AggCallPlus aggCall : aggCalls) {
-      aggregateCalls.add(aggCall.aggregateCall(registrar, groupSet, r));
+      AggregateCall aggregateCall = aggCall.aggregateCall(registrar, groupSet, r);
+      if (groupSets.size() <= 1) {
+        aggregateCall = removeRedundantAggregateDistinct(aggregateCall, groupSet, r);
+      }
+      aggregateCalls.add(aggregateCall);
     }
 
     assert ImmutableBitSet.ORDERING.isStrictlyOrdered(groupSets) : groupSets;
@@ -2523,6 +2527,30 @@ public class RelBuilder {
     aggregate_(groupSet, groupSets, r, distinctAggregateCalls,
         registrar.extraNodes, inFields);
     return project(projects.transform((i, name) -> aliasMaybe(field(i), name)));
+  }
+
+  /**
+   * Removed redundant distinct if an input is already unique.
+   */
+  private AggregateCall removeRedundantAggregateDistinct(
+      AggregateCall aggregateCall,
+      ImmutableBitSet groupSet,
+      RelNode relNode) {
+    if (aggregateCall.isDistinct() && config.removeRedundantDistinct()) {
+      final RelMetadataQuery mq = relNode.getCluster().getMetadataQuery();
+      final List<Integer> argList = aggregateCall.getArgList();
+      final ImmutableBitSet distinctArg = ImmutableBitSet.builder()
+          .addAll(argList)
+          .build();
+      final ImmutableBitSet columns = groupSet.union(distinctArg);
+      final Boolean alreadyUnique =
+          mq.areColumnsUnique(relNode, columns);
+      if (alreadyUnique != null && alreadyUnique) {
+        // columns have been distinct or columns are primary keys
+        return aggregateCall.withDistinct(false);
+      }
+    }
+    return aggregateCall;
   }
 
   /** Returns whether an input is already unique, and therefore a Project
@@ -4902,6 +4930,18 @@ public class RelBuilder {
 
     /** Sets {@link #convertCorrelateToJoin()}. */
     Config withConvertCorrelateToJoin(boolean convertCorrelateToJoin);
+
+    /** Whether to remove the distinct that in aggregate if we know that the input is
+     * already unique; default false. */
+    @Value.Default
+    default boolean removeRedundantDistinct() {
+      return false;
+    }
+
+    /**
+     * Sets {@link #removeRedundantDistinct()}.
+     */
+    Config withRemoveRedundantDistinct(boolean removeRedundantDistinct);
   }
 
 }
--- a/core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
@@ -6346,6 +6346,138 @@ LogicalAggregate(group=[{0}], CNT=[COUNT()])
 ]]>
     </Resource>
   </TestCase>
+  <TestCase name="testRemoveDistinctIfUnique1">
+    <Resource name="sql">
+      <![CDATA[SELECT
+    deptno,
+    COUNT(DISTINCT sal) as cds,
+    COUNT(sal) as cs,
+    SUM(DISTINCT sal) AS sds,
+    SUM(sal) AS ss
+FROM (
+    SELECT DISTINCT deptno, sal
+    FROM emp)
+GROUP BY deptno]]>
+    </Resource>
+    <Resource name="plan">
+      <![CDATA[
+LogicalProject(DEPTNO=[$0], CDS=[$1], CS=[$2], SDS=[$3], SS=[$3])
+  LogicalAggregate(group=[{0}], CDS=[COUNT($1)], CS=[COUNT()], SDS=[SUM($1)])
+    LogicalAggregate(group=[{0, 1}])
+      LogicalProject(DEPTNO=[$7], SAL=[$5])
+        LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testRemoveDistinctIfUnique2">
+    <Resource name="sql">
+      <![CDATA[SELECT
+    COUNT(DISTINCT sal) as cds,
+    COUNT(sal) as cs,
+    SUM(DISTINCT sal) AS sds,
+    SUM(sal) AS ss
+FROM (
+    SELECT deptno, 1 as sal
+    FROM emp    GROUP BY deptno)GROUP BY deptno
+]]>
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/apache/calcite/rel/rules/AggregateRemoveDistinctRule.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.util.ImmutableBitSet;
+
+import org.immutables.value.Value;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Planner rule that removes a distinct in count for {@link Aggregate}.
```
mihaibudiu, line 32: @julianhyde asked in JIRA to handle other aggregates too. Why just count?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6562,6 +6562,67 @@ private HepProgram getTransitiveProgram() {
...
+        + ") t ";
+    sql(sql)
+        .withRule(CoreRules.AGGREGATE_REMOVE_DISTINCT)
+        .check();
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6214">[CALCITE-6214]
+   * Remove `DISTINCT` in `COUNT` if field is unique</a>. */
+  @Test void testAggregateDistinctRemove3() {
```
mihaibudiu, line 6597: If you handle other aggregates you should probably add tests like `SELECT COUNT(x), SUM(x) FROM SELECT DISTINCT ...`

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6562,6 +6562,67 @@ private HepProgram getTransitiveProgram() {
         .check();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6214">[CALCITE-6214]
+   * Remove `DISTINCT` in `COUNT` if field is unique</a>. */
```
julianhyde, line 6567: Don't use backticks in jira summary or javadoc. They will be rendered as backticks.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6562,6 +6562,67 @@ private HepProgram getTransitiveProgram() {
...
+    sql(sql)
+        .withRule(CoreRules.AGGREGATE_REMOVE_DISTINCT)
+        .check();
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6214">[CALCITE-6214]
+   * Remove `DISTINCT` in `COUNT` if field is unique</a>. */
+  @Test void testAggregateDistinctRemove2() {
+    final String sql = ""
```
julianhyde, line 6583: Can you add at least one test where the outer query has a `GROUP BY`? The following query should benefit from the simplification:
```
SELECT deptno, COUNT(DISTINCT sal), 
FROM (
   SELECT DISTINCT deptno, sal
   FROM emp)
GROUP BY deptno
```
Note that `sal` is not distinct but it is distinct for each `deptno` (because `(deptno, sal)` is a key).

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
JiajunBernoulli: Yes, now we can handle other aggregate functions after new commits.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
JiajunBernoulli: Removed them.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
JiajunBernoulli: Added it.

core/src/main/java/org/apache/calcite/rel/rules/AggregateRemoveDistinctRule.java
JiajunBernoulli: I overlooked it before.

Now we can handle other aggregates.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -2525,6 +2529,29 @@ private RelBuilder aggregate_(GroupKeyImpl groupKey,
     return project(projects.transform((i, name) -> aliasMaybe(field(i), name)));
   }
 
+  /**
+   * Removed redundant distinct if an input is already unique.
```
mihaibudiu, line 2533: I would document that this is specifically for aggregates.
And perhaps a better function name would be removeRedundantAggregateDistinct.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -4902,6 +4929,18 @@ public interface Config {
 
     /** Sets {@link #convertCorrelateToJoin()}. */
     Config withConvertCorrelateToJoin(boolean convertCorrelateToJoin);
+
+    /** Whether to save the distinct if we know that the input is
+     * already unique; default true. */
+    @Value.Default
+    default boolean redundantDistinct() {
```
mihaibudiu, line 4936: This flag is a little unintuitive, since it *inhibits* the optimization rather than enabling it.
All the other similar flags are in the opposite way.

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -6343,6 +6343,161 @@ LogicalAggregate(group=[{0}], CNT=[COUNT()])
     LogicalFilter(condition=[>(ITEM($0, 'N_NATIONKEY'), 5)])
       LogicalProject(**=[$0])
         LogicalTableScan(table=[[CATALOG, SALES, NATION]])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testRemoveDistinctIfUnique">
```
mihaibudiu, line 6349: Where is the corresponding test?

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -6343,6 +6343,161 @@ LogicalAggregate(group=[{0}], CNT=[COUNT()])
...
+  </TestCase>
+  <TestCase name="testRemoveDistinctIfUnique5">
+    <Resource name="sql">
+      <![CDATA[SELECT COUNT(DISTINCT empno)
+FROM emp
+]]>
+    </Resource>
+    <Resource name="plan">
+      <![CDATA[
+LogicalAggregate(group=[{}], EXPR$0=[COUNT($0)])
```
mihaibudiu, line 6470: Why is the DISTINCT removed from this case?

mihaibudiu (commented):
Another question I have is whether this should be as you wrote it in RelBuilder, or it should be a rewrite rule in the optimizer. It is a matter of design choice rather than of correctness.

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
JiajunBernoulli: We can know `empno` is primary key by using `RelMetadataQuery#areColumnsUnique`.

Here is metadata:
https://github.com/apache/calcite/blob/f837ffa877c5b7f795fd175532f6ec9db4440533/testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReaderSimple.java#L80

mihaibudiu: There are so many empno references in the codebase that I couldn't figure out that this is a primary key. 
There are also multiple definitions of this column in multiple files, and I didn't know which one is being used here.
Can you please add a comment explaining this in the code?

JiajunBernoulli: Added

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
JiajunBernoulli: Forgot to delete.

Removed it.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
JiajunBernoulli: Good catch.

I changed default value, now we can set `true` to optimize. (Default is `false`)

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
JiajunBernoulli: Renamed to `removeRedundantAggregateDistinct`.

JiajunBernoulli:
> Another question I have is whether this should be as you wrote it in RelBuilder, or it should be a rewrite rule in the optimizer. It is a matter of design choice rather than of correctness.

Thank you for your review.

1. RelBuilder optimization can reuse subexpressions.
- Using RelBuilder
```
LogicalProject(DEPTNO=[$0], CDS=[$1], CS=[$2], SDS=[$3], SS=[$3]) -- SDS is same as SS
  LogicalAggregate(group=[{0}], CDS=[COUNT($1)], CS=[COUNT()], SDS=[SUM($1)])
    LogicalAggregate(group=[{0, 1}])
      LogicalProject(DEPTNO=[$7], SAL=[$5])
```
- Using Rule
```
LogicalProject(DEPTNO=[$0], CDS=[$1], CS=[$2], SDS=[$3], SS=[$4]) -- SDS is same as SS
  LogicalAggregate(group=[{0}], CDS=[COUNT($1)], CS=[COUNT()], SDS=[SUM($1)], SS=[SUM($1)])
    LogicalAggregate(group=[{0, 1}])
      LogicalProject(DEPTNO=[$7], SAL=[$5])
```
We need other rules to remove same function.

2. RelBuilder is easier to use than Rule.
- RelBuilder: `withRedundantDistinct(flag)` to enable or disable.
- Rule: Add or remove rule in programs.

PR #3643, opened 2024-01-22
[CALCITE-5976] Use explicit casting if inserted element type in ArrayPrepend/ArrayAppend/ArrayInsert does not equal derived component type

caicancai:
https://issues.apache.org/jira/browse/CALCITE-5976

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -5204,7 +5204,11 @@ public class SqlFunctions {
   public static List arrayAppend(List list, Object element) {
     final List result = new ArrayList(list.size() + 1);
     result.addAll(list);
-    result.add(element);
+    if (element instanceof Byte) {
+      result.add(((Byte) element).intValue());
+    } else {
+      result.add(element);
+    }
     return result;
   }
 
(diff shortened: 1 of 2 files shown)


caicancai:
@chucheng92 Hello, I'm not done yet, I want to make sure the way I'm fixing it is acceptable

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -5204,7 +5204,11 @@ public static List compact(List list) {
   public static List arrayAppend(List list, Object element) {
     final List result = new ArrayList(list.size() + 1);
     result.addAll(list);
-    result.add(element);
+    if (element instanceof Byte) {
+      result.add(((Byte) element).intValue());
```
mihaibudiu, line 5208: This doesn't seem right. Is the array always containing integers?
What happens for an array of BIGINT?
What happens for Short?

caicancai: Thanks I will try other methods

chucheng92: +1 can't do this. we must make sure when you call `arrayAppend` the input `List` already is a isomorphic result list (a result list has same operand's type). so it means we need an explicit casting before you call SqlFunctions.

PR #1763, opened 2020-01-15
[CALCITE-3738] Missing order by logical plan for insert statement

pengzhiwei2018:

PR #3644, opened 2024-01-24
[CALCITE-6063] If ARRAY subquery has ORDER BY (without LIMIT), rows are not sorted

chucheng92:
issue: 
https://issues.apache.org/jira/browse/CALCITE-6063

context:
The sql-standard-2003 allows array constrcutor by query:
```
<array value constructor by query> ::=
   ARRAY <left paren>  <query expression>  [ <order by clause>  ] <right paren> 
```

and calcite does support it.

however the orderby was removed by the issue https://issues.apache.org/jira/browse/CALCITE-2978.

The solution of this PR is same as #1763 to allow sort in subquery.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -5538,7 +5538,9 @@ public class SqlToRelConverter {
         case ARRAY_QUERY_CONSTRUCTOR:
           call = (SqlCall) expr;
           query = Iterables.getOnlyElement(call.getOperandList());
-          root = convertQueryRecursive(query, false, null);
+          // let top=true to make the query be top-level query,
+          // then ORDER BY will be reserved.
+          root = convertQueryRecursive(query, true, null);
           return RexSubQuery.array(root.rel);
 
         case MAP_QUERY_CONSTRUCTOR:
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -1643,6 +1643,14 @@ Implicit type coercion of following cases are ignored:
 | ARRAY '[' value [, value ]* ']' | Creates an array from a list of values.
 | MAP '[' key, value [, key, value ]* ']' | Creates a map from a list of key-value pairs.
 
+### Value constructors by query
+
+| Operator syntax                    | Description |
+|:-----------------------------------|:------------|
+| ARRAY (sub-query)                  | Creates an array from the result of a sub-query. Example: `ARRAY(SELECT empno FROM emp ORDER BY empno)` |
+| MAP (sub-query)                    | Creates a map from the result of a key-value pair sub-query. Example: `MAP(SELECT empno, deptno FROM emp)` |
+| MULTISET (sub-query)               | Creates a multiset from the result of a sub-query. Example: `MULTISET(SELECT empno FROM emp)` |
+
 ### Collection functions
 
 | Operator syntax | Description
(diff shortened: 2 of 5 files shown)


XuQianJin-Stars (approved):

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -5538,7 +5538,7 @@ ImmutableList<RelNode> retrieveCursors() {
         case ARRAY_QUERY_CONSTRUCTOR:
           call = (SqlCall) expr;
           query = Iterables.getOnlyElement(call.getOperandList());
-          root = convertQueryRecursive(query, false, null);
+          root = convertQueryRecursive(query, true, null);
```
mihaibudiu, line 5541: I think this deserves a comment. I think you are pretending this to be the toplevel query, because this will prevent ORDER BY from being eliminated.

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
julianhyde: Agree. Perhaps rename `boolean top` to `boolean preserveOrderBy` (assuming that `top` doesn't control any other behavior). Then the documentation can be on the `convertQueryRecursive` method.

site/_docs/reference.md
```diff
@@ -1643,6 +1643,14 @@ Implicit type coercion of following cases are ignored:
 | ARRAY '[' value [, value ]* ']' | Creates an array from a list of values.
 | MAP '[' key, value [, key, value ]* ']' | Creates a map from a list of key-value pairs.
 
+### Value constructors by query
+
+| Operator syntax                    | Description |
+|:-----------------------------------|:------------|
+| ARRAY (sub-query)                  | Creates an array from the result of a sub-query. Example: `ARRAY(SELECT empno FROM emp)` |
```
julianhyde, line 1650: maybe add `ORDER BY` to example, to illustrate that it is possible

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
chucheng92: thanks, fixed. however I still used top, The current codepath is passed in from top all the way. top is also easy to understand.

site/_docs/reference.md
chucheng92: Added it.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
julianhyde: You force-pushed?!

chucheng92:
I have updated the PR to fix:
1. Rename jira/commit/PR name (it's more clear)
2. Add example with OrderBy
3. Add test cases for multiset & map (including parser & operator test)

It should be noted that there are some differences between ARRAY subquery and MULTISET/MAP subquery. Please refer to the comments of the [CALCITE-6063](https://issues.apache.org/jira/browse/CALCITE-6063) for details. We may decide whether to make some improvements in follow-up issues or PRs.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
chucheng92: > You force-pushed?!

Sorry, I know the rules for review, but I accidentally force-push it ..

chucheng92:
Hi guys. Sorry, I accidentally forced-push PR instead of a new commit, but there are no new lines affecting the logic, only the test cases for map/multiset.

If there are no other comments, I will merge it later..

chucheng92:
merging...

PR #3645, opened 2024-01-24
[CALCITE-6222] Mysql does not have to_char function

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6222


caicancai:
mysql8.0 test

2024-01-25 00-03-05屏幕截图

PR #3648, opened 2024-01-25
[CALCITE-6224] Add LOG2 function (enabled in Mysql, Spark library)

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6224

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -208,6 +208,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.JSON_TYPE;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LEFT;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LEVENSHTEIN;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG2;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_AND;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_OR;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LPAD;
@@ -639,6 +640,7 @@ public class RexImpTable {
 
       map.put(LN, new LogImplementor());
       map.put(LOG, new LogImplementor());
+      map.put(LOG2, new LogImplementor());
       map.put(LOG10, new LogImplementor());
 
       defineReflective(RAND, BuiltInMethod.RAND.method,
@@ -4108,7 +4110,7 @@ public class RexImpTable {
     }
   }
 
-  /** Implementor for the {@code LN}, {@code LOG}, and {@code LOG10} operators.
+  /** Implementor for the {@code LN}, {@code LOG}, {@code LOG2}, and {@code LOG10} operators.
    *
    * <p>Handles all logarithm functions using log rules to determine the
    * appropriate base (i.e. base e for LN).
@@ -4137,6 +4139,8 @@ public class RexImpTable {
         return list.append(Expressions.constant(Math.exp(1)));
       case "LOG10":
         return list.append(Expressions.constant(BigDecimal.TEN));
+      case "LOG2":
+        return list.append(Expressions.constant(2));
       default:
         throw new AssertionError("Operator not found: " + call.getOperator());
       }
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -2764,7 +2764,7 @@ public class SqlFunctions {
   }
 
 
-  // LN, LOG, LOG10
+  // LN, LOG, LOG2, LOG10
 
   /** SQL {@code LOG(number, number2)} function applied to double values. */
   public static double log(double d0, double d1) {
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -2142,6 +2142,18 @@ public abstract class SqlLibraryOperators {
           OperandTypes.NUMERIC_OPTIONAL_NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
+  /** The "LOG2(value)" function.
+   *
+   * @see SqlStdOperatorTable#LN
+   * @see SqlStdOperatorTable#LOG10
+   */
+  @LibraryOperator(libraries = {MYSQL, SPARK})
+  public static final SqlFunction LOG2 =
+      SqlBasicFunction.create("LOG2",
+          ReturnTypes.DOUBLE_NULLABLE,
+          OperandTypes.NUMERIC,
+          SqlFunctionCategory.NUMERIC);
+
   @LibraryOperator(libraries = {BIG_QUERY})
   public static final SqlFunction POW =
       SqlStdOperatorTable.POWER.withName("POW");
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2783,6 +2783,7 @@ In the following:
 | b f | LENGTH(string)                               | Equivalent to `CHAR_LENGTH(string)`
 | h s | LEVENSHTEIN(string1, string2)                | Returns the Levenshtein distance between *string1* and *string2*
 | b | LOG(numeric1 [, numeric2 ])                    | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present
+| m s | LOG2(numeric)                                | Returns the base 2 logarithm of *numeric*
 | b o | LPAD(string, length [, pattern ])            | Returns a string or bytes value that consists of *string* prepended to *length* with *pattern*
 | b | TO_BASE32(string)                              | Converts the *string* to base-32 encoded form and returns an encoded string
 | b | FROM_BASE32(string)                            | Returns the decoded result of a base-32 *string* as a string
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -6236,7 +6236,9 @@ public class SqlOperatorTest {
    * <a href="https://issues.apache.org/jira/browse/CALCITE-6224">[CALCITE-6224]
    * Add LOG@ function (enabled in MYSQL, Spark library)</a>. */
   @Test void testLog2Func() {
-    final SqlOperatorFixture f0 = fixture();
+    final SqlOperatorFixture f0 = Fixtures.forOperators(true);
+    f0.checkFails("^log2(4)^",
+        "No match found for function signature LOG2\\(<NUMERIC>\\)", false);
     final Consumer<SqlOperatorFixture> consumer = f -> {
       f.setFor(SqlLibraryOperators.LOG2);
       f.checkScalarApprox("log2(2)", "DOUBLE NOT NULL",
@@ -6253,8 +6255,6 @@ public class SqlOperatorTest {
           isWithin(-1.0, 0.000001));
       f.checkScalarApprox("log2(-2)", "DOUBLE NOT NULL",
           "NaN");
-      f.checkScalarApprox("log2(0)", "DOUBLE NOT NULL",
-          "-Infinity");
       f.checkNull("log2(cast(null as real))");
     };
     f0.forEachLibrary(list(SqlLibrary.MYSQL, SqlLibrary.SPARK), consumer);
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -6186,6 +6186,12 @@ public class SqlOperatorTest {
         isWithin(3.0, 0.000001));
     f.checkScalarApprox("log10(cast(10e-3 as real))", "DOUBLE NOT NULL",
         isWithin(-2.0, 0.000001));
+    f.checkScalarApprox("log10(2.0/3)", "DOUBLE NOT NULL",
+        isWithin(-0.17609125905568118, 0.000001));
+    f.checkScalarApprox("log10(0.5)", "DOUBLE NOT NULL",
+        isWithin(-0.30102999566398114, 0.000001));
+    f.checkScalarApprox("log10(100.0/3)", "DOUBLE NOT NULL",
+        isWithin(1.5228787452803372, 0.000001));
     f.checkNull("log10(cast(null as real))");
   }
 
@@ -6211,6 +6217,12 @@ public class SqlOperatorTest {
         isWithin(9.0, 0.000001));
     f.checkScalarApprox("log(cast(10e-3 as real), 10)", "DOUBLE NOT NULL",
         isWithin(-2.0, 0.000001));
+    f.checkScalarApprox("log(2.0/3, 100)", "DOUBLE NOT NULL",
+        isWithin(-0.08804562952784059, 0.000001));
+    f.checkScalarApprox("log(0.5, 2)", "DOUBLE NOT NULL",
+        isWithin(-1.0, 0.000001));
+    f.checkScalarApprox("log(0, 100)", "DOUBLE NOT NULL",
+        "-Infinity");
     f.checkNull("log(cast(null as real), 10)");
     f.checkNull("log(10, cast(null as real))");
   }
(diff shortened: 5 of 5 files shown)


caicancai:
@tanclary Hello. If you have time, can you help me take a look? You wrote and modified the log function before, thank you.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -703,15 +705,15 @@ Builder populate2() {
       map.put(FLOOR,
           new FloorImplementor(BuiltInMethod.FLOOR.method,
               BuiltInMethod.UNIX_TIMESTAMP_FLOOR.method,
-            BuiltInMethod.UNIX_DATE_FLOOR.method,
-            BuiltInMethod.CUSTOM_TIMESTAMP_FLOOR.method,
-            BuiltInMethod.CUSTOM_DATE_FLOOR.method));
+              BuiltInMethod.UNIX_DATE_FLOOR.method,
```
caicancai, line 708: I will fix it later

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6162,6 +6162,18 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
   }
 
+  @Test void testLog2Func() {
+    final SqlOperatorFixture f = Fixtures.forOperators(true);
+    f.setFor(SqlStdOperatorTable.LOG2, VmName.EXPAND);
+    f.checkScalarApprox("log2(2)", "DOUBLE NOT NULL",
+        isWithin(1.0, 0.000001));
+    f.checkScalarApprox("log2(4)", "DOUBLE NOT NULL",
+        isWithin(2.0, 0.000001));
+    f.checkScalarApprox("log2(65536)", "DOUBLE NOT NULL",
```
tanclary, line 6172: Can we add more tests:
with decimal arguments, negative arguments, maybe an expression (2 / 3)

tanclary (changes requested):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2764,7 +2764,7 @@ public static double power(BigDecimal b0, BigDecimal b1) {
   }
 
 
-  // LN, LOG, LOG10
+  // LN, LOG, LOG10, LOG2
```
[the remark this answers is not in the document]
caicancai: done

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: done, thank you

site/_docs/reference.md
```diff
@@ -2775,6 +2775,7 @@ In the following:
 | b f | LENGTH(string)                               | Equivalent to `CHAR_LENGTH(string)`
 | h s | LEVENSHTEIN(string1, string2)                | Returns the Levenshtein distance between *string1* and *string2*
 | b | LOG(numeric1 [, numeric2 ])                    | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present
+| m s | | LOG2(numeric)                              | Returns the base 2 logarithm of *numeric*
```
tanclary, line 2778: Remove the extra ` |`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6188,6 +6188,29 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.MYSQL);
+    f.checkScalarApprox("log2(2)", "DOUBLE NOT NULL",
+        isWithin(1.0, 0.000001));
+    f.checkScalarApprox("log2(4)", "DOUBLE NOT NULL",
+        isWithin(2.0, 0.000001));
+    f.checkScalarApprox("log2(65536)", "DOUBLE NOT NULL",
+        isWithin(16.0, 0.000001));
+    f.checkScalarApprox("log2(-2)", "DOUBLE NOT NULL",
+        "NaN");
+    f.checkScalarApprox("log2(2/3)", "DOUBLE NOT NULL",
```
tanclary, line 6205: wait what, why is log2(2/3) `-Infinity`?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6188,6 +6188,29 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
     f.checkNull("log(10, cast(null as real))");
   }
 
+  @Test void testLog2Func() {
+    final SqlOperatorFixture f0 = Fixtures.forOperators(true);
+    f0.setFor(SqlLibraryOperators.LOG2, VmName.EXPAND);
+    f0.checkFails("^log2(4)^",
+        "No match found for function signature LOG2\\(<NUMERIC>\\)", false);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.MYSQL);
```
tanclary, line 6196: Even if the behavior does not vary across libraries, it is easy to test against all libraries using the `consumer` pattern. See any test that uses `forEachLibrary()` for examples.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: Using desmos, this should equal ~-0.5849625007

tanclary (changes requested):
Looks nearly good, just a couple of changes

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: It seems that there is a little bug in parsing SqlNode. It seems that log10 also has the same problem. I will open a jira to record the problem.

caicancai: I put the test into log10 and had the same problem

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6188,6 +6188,30 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
     f.checkNull("log(10, cast(null as real))");
   }
 
+  @Test void testLog2Func() {
```
JiajunBernoulli, line 6191: It will be better if there are JIRA document.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6188,6 +6188,30 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+    final SqlOperatorFixture f0 = Fixtures.forOperators(true);
+    final Consumer<SqlOperatorFixture> consumer = f -> {
+      f.setFor(SqlLibraryOperators.LOG2);
```
JiajunBernoulli, line 6194: What error message for  `log2(0)`?

caicancai: It seems to be a bug. I recorded it with log(2/3) in a jira because log10 also has the same problem.

caicancai: I'm trying to solve it

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6232

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2142,6 +2142,18 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
           OperandTypes.NUMERIC_OPTIONAL_NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
+  /** The "LOG2(value)" function.
+   *
+   * @see SqlStdOperatorTable#LN
+   * @see SqlStdOperatorTable#LOG10
+   */
+  @LibraryOperator(libraries = {MYSQL, SPARK})
```
XuQianJin-Stars, line 2150: I noticed that PostgreSQL, SQL Server, and Oracle all support the processing of the LOG2 function. You can check if this link can help improve this PR: https://www.w3resource.com/oracle/oracle-numeric-functions/oracle-log-function.php

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6188,6 +6188,33 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+      f.checkScalarApprox("log2(2)", "DOUBLE NOT NULL",
+          isWithin(1.0, 0.000001));
+      f.checkScalarApprox("log2(4)", "DOUBLE NOT NULL",
+          isWithin(2.0, 0.000001));
+      f.checkScalarApprox("log2(65536)", "DOUBLE NOT NULL",
+          isWithin(16.0, 0.000001));
+      f.checkScalarApprox("log2(-2)", "DOUBLE NOT NULL",
+          "NaN");
+      f.checkScalarApprox("log2(2/3)", "DOUBLE NOT NULL",
+          "-Infinity");
```
XuQianJin-Stars, line 6207: `Double.NEGATIVE_INFINITY` is used to represent negative infinity. Precision preservation needs to be given here. For example: mysql is 10 digits.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: https://www.postgresql.org/docs/current/functions-math.html#FUNCTIONS-MATH-FUNC-TABLE
The official document does not show it, I think there is no need to add it

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: Yes, but there is also the problem of node type conversion, otherwise it seems to be taking integer bits, which is unreasonable

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6188,6 +6188,35 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+      f.checkScalarApprox("log2(2.2)", "DOUBLE NOT NULL",
+          "1.1375035237499351");
+      f.checkScalarApprox("log2(0.5)", "DOUBLE NOT NULL",
+          "-1.0");
+      f.checkScalarApprox("log2(3)", "DOUBLE NOT NULL",
```
caicancai, line 6212: You can see that there are retained decimal places. 
The first one is obviously log2(0) and log10(0). We may need to process this. 
Second, when converting fractions like 2/3 into sqlnode nodes, an integer value is taken, which also needs further modification.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6188,6 +6188,35 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+          isWithin(1.5849625007211563, 0.000001));
```
caicancai, line 6213: You can see that there are retained decimal places.
The first one is obviously log2(0) and log10(0). We may need to process this.
Second, when converting fractions like 2/3 into sqlnode nodes, an integer value is taken, which also needs further modification.

caicancai: I will fix it in next PR

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: https://issues.apache.org/jira/browse/CALCITE-6232, Julian seems to think this is not a bug

tanclary: I mean it's definitely a bug but you may want to rewrite your case such that the issue is more clearly stated. I would say stick with a single test case, give the expected results and the actual results. Feel free to tag me on any cases or PRs you file.

I should have found this during my previous LOG work but I didn't. So let me know how I can help.

caicancai: @tanclary I specially added a test for this bug, I believe that users and developers can understand

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6188,6 +6188,65 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+    final SqlOperatorFixture f0 = Fixtures.forOperators(true);
+    f0.setFor(SqlLibraryOperators.LOG, VmName.EXPAND);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.BIG_QUERY);
+    f.checkScalarApprox("log(2/3, 2)", "DOUBLE NOT NULL",
+          "-Infinity");
+    f.checkScalarApprox("log(0,2)", "DOUBLE NOT NULL",
+          "-Infinity");
+    f.checkScalarApprox("log(1,2)", "DOUBLE NOT NULL",
+        "0.0");
+    f.checkScalarApprox("log(4/3,2)", "DOUBLE NOT NULL",
```
tanclary, line 6233: What happens when you wrap the arguments in casts?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6188,6 +6188,65 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+        "0.0");
+    f.checkScalarApprox("log(5/3,2)", "DOUBLE NOT NULL",
+        "0.0");
+    f.checkScalarApprox("log(2/3, 10)", "DOUBLE NOT NULL",
+          "-Infinity");
+    f.checkScalarApprox("log(0,10)", "DOUBLE NOT NULL",
+          "-Infinity");
+    f.checkScalarApprox("log(1,10)", "DOUBLE NOT NULL",
+        "0.0");
+    f.checkScalarApprox("log(4/3,10)", "DOUBLE NOT NULL",
```
tanclary, line 6243: These are good tests but I don't know that we should merge this until the pretty glaring bug is fixed

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: I add test

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: I'm trying to fix it, but I'm sure it won't last long. I think we have to be honest and tell the user about this bug. Now maybe our difficulty is how to tell the user about this bug, should I add a TODO tag

caicancai:
@mihaibudiu Maybe I need your opinion if you have time, thank you

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: Why don't we fix the bug? Do you have an updated JIRA case for it?

caicancai: ok, this may take a while for me because i am learning this

caicancai: > Why don't we fix the bug? Do you have an updated JIRA case for it?

I don't know if this is true. https://issues.apache.org/jira/browse/CALCITE-6232

tanclary: What happens if you use the decimal form like `.5` instead of `1/2`?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6188,6 +6188,67 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6232">[CALCITE-6232]
+   * Using fractions in LOG function does not return correct results</a>. */
+  @Test void testLogFuncByConvert() {
+    // The fractional conversion of the Log function is reserved only for integer bits
+    final SqlOperatorFixture f0 = Fixtures.forOperators(true);
+    f0.setFor(SqlLibraryOperators.LOG, VmName.EXPAND);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.BIG_QUERY);
+    f.checkScalarApprox("log(2/3, 2)", "DOUBLE NOT NULL",
```
mihaibudiu, line 6227: you have to write tests with illegal and NULL values for all combinations of arguments

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: The result looks correct, it's only different when it comes to fractions

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: done. The result looks correct, it's only different when it comes to fractions

caicancai:
@mihaibudiu @tanclary https://issues.apache.org/jira/browse/CALCITE-6237

caicancai:
I will fix it

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6215,6 +6215,71 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6232">[CALCITE-6232]
+   * Using fractions in LOG function does not return correct results</a>. */
+  @Test void testLogFuncByConvert() {
+    final SqlOperatorFixture f0 = fixture();
+    f0.setFor(SqlLibraryOperators.LOG, VmName.EXPAND);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.BIG_QUERY);
+    f.checkScalarApprox("log(2.0/3, 2)", "DOUBLE NOT NULL",
```
caicancai, line 6253: Hello, I seem to have forgotten the simplest truth, that is, there are no fractions in computers, only floating point numbers. I changed 2/3 in the test to 2.0/3, and it turns out that the result is correct. If 2/3 must output the correct result, then this bug does not only exist in the log function, but the pow and abs functions also have similar problems. Regarding log(0) I don't think it's a bug.

Thank you very much for your guidance on these two PRs of my adaptation function. I am sorry for causing you a lot of trouble. If there is anything missing in this PR, please point it out, thank you.

caicancai: I will remove this part of the test later

mihaibudiu: I would recommend studying the SQL language, otherwise you will have surprises.
It's very difficult to contribute to a compiler if you don't understand the language you are compiling.
In SQL floating point is generally not used. SQL has a family of DECIMAL (or NUMERIC -- same thing) data types, which are very different from floating point. 2.0 is a DECIMAL literal. DECIMAL is not a type, it's a family of types with different scales and precisions. Unfortunately, different dialects of SQL treat DECIMAL in incompatible ways. Floating point is an IEEE standard, but DECIMAL isn't.

tanclary: I am still of the belief that even though LOG(0) returning `-Infinity` makes sense mathematically, we should do our best to parity the dialects that support these functions. I am a fan of extending a function's behavior, but not augmenting it.

mihaibudiu: If a function is part of a dialect, it should definitely have the same behavior as the dialect requires. This should also be documented. We may be dealing with multiple LOG functions, with slightly different behaviors. That's why it is very important to test all corner cases, including writing negative tests.

tanclary: Totally agree, if necessary we can do something similar to the `SUBSTR` functions which I think is implemented per-dialect.

caicancai: I have a question. For example, the log2(0) function has different expressions in the results of spark and mysql. Do we need to adapt them separately?
Regarding negative tests, I would like to ask which ones are negative tests. At first I thought it was just log2(0).

caicancai: I have tested in both spark and mysql. The output of log2(0) in spark and mysql is null. Do I need to process the parameter 0 in sqlfunctions? Or simply adding test is enough.

caicancai: Hello, thank you very much for your suggestion. Regarding negative test, I think of the following case

- log10(-1): When the parameter in log10 is a negative number
- 
- log10(null): When the parameters in log10 are empty
- 
- log10(-`Infinity`): When the parameter in log10 is -`Infinity`
- 
- log10(0): When the parameter in log10 is 0

I tested based on mysql. Only when log(0) and log10(-1), calcite cannot match the results in mysql. In mysql, both of them return NULL. When log10(0), calcite returns -`Infinity. `log10(-1) returns NaN

test2

caicancai: My current idea is to add a judgment to the log function in sqlFunctions. If it is less than or equal to 0, it will return 0. But there is a problem, that is, log10 and log functions are adapted to the current calcite database. So what should we do? To solve this problem, we cannot guarantee that the results returned by the database will be the same.
In other words, I should write Sqlfunctions specifically for log2

caicancai: @mihaibudiu Or should we deal with this situation?

caicancai: Double.NaN == null？

caicancai: https://github.com/apache/calcite/pull/3669, It seems that there should be the same situation

tanclary: Yes, you need to adjust/add behavior s.t. when this function is used for these dialects, it follows the proper behavior.

tanclary: The examples you gave above are **not** negative tests (aside from the `null` one). Please read my previous comment. What do you expect of the user? It is a numeric function that is designed to support numeric arguments. Anything outside of that is unexpected.

caicancai: I probably understand what you mean. What you mean is that if you input types such as string and boolean that are not numbers, an error should be reported.

caicancai: Since my pr has a conflict, I will close it first, and then I will raise a pr after solving the problem of log2(0) returning null.
Thanks

PR #3649, opened 2024-01-25
[CALCITE-6225] Unparse LOGICAL_AND/OR as MIN/MAX for Snowflake

tanclary:


olivrlee (approved):
Following [BQ documentation](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#logical_and)  and [Snowflake documentation ](https://docs.snowflake.com/en/sql-reference/functions/min )

it seems like NULLs are handled the same for both cases.

That's my first thought when reading the jira ticket, but since it matches up maybe we don't need explicit tests for that?

LGTM

core/src/main/java/org/apache/calcite/sql/dialect/SnowflakeSqlDialect.java
```diff
@@ -42,35 +44,34 @@ public SnowflakeSqlDialect(Context context) {
...
+      break;
+    case MAX:
+      op = SqlStdOperatorTable.MAX;
+      break;
+    case MIN:
+      op = SqlStdOperatorTable.MIN;
       break;
     default:
-      super.unparseCall(writer, call, leftPrec, rightPrec);
+      op = call.getOperator();
```
julianhyde, line 71: If the call is fine, let's not make a new one.

tanclary: Yeah I didn't like doing that either.. just added a different approach in the second commit

core/src/main/java/org/apache/calcite/sql/dialect/SnowflakeSqlDialect.java
```diff
@@ -42,34 +47,20 @@ public SnowflakeSqlDialect(Context context) {
...
+    map.put(SqlKind.BIT_AND, SqlLibraryOperators.BITAND_AGG);
+    map.put(SqlKind.BIT_OR, SqlLibraryOperators.BITOR_AGG);
+    map.put(SqlKind.CHAR_LENGTH, SqlLibraryOperators.LENGTH);
+    map.put(SqlKind.ENDS_WITH, SqlLibraryOperators.ENDSWITH);
+    map.put(SqlKind.STARTS_WITH, SqlLibraryOperators.STARTSWITH);
+    map.put(SqlKind.MAX, SqlStdOperatorTable.MAX);
+    map.put(SqlKind.MIN, SqlStdOperatorTable.MIN);
+    SqlOperator op = map.get(call.getKind());
+    if (op != null) {
+        SqlCall newCall = op.createCall(SqlParserPos.ZERO, call.getOperandList());
```
JiajunBernoulli, line 60: 1. Check-style failed.
2. We can use `final`
3. Do you like static field?  It is free for you. 
    - We will build map when every unparse. If we have static map, this is only one.
    - The static field will also continue to occupy memory.

JiajunBernoulli (approved):

PR #3650, opened 2024-01-25
[CALCITE-6228] ELEMENT function infers incorrect return type

mihaibudiu:


core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -615,10 +615,11 @@ public static SqlCall stripSeparator(SqlCall call) {
       ARG0.andThen(SqlTypeTransforms.TO_MULTISET);
 
   /**
-   * Returns the element type of a MULTISET.
+   * Returns the element type of a MULTISET, with nullability enforced.
    */
   public static final SqlReturnTypeInference MULTISET_ELEMENT_NULLABLE =
```
JiajunBernoulli, line 620: Is the name appropriate?

There are two field:
- NULLABLE -> TO_NULLABLE
```
  public static final SqlReturnTypeInference TO_COLLECTION_ELEMENT_NULLABLE =
      TO_COLLECTION_ELEMENT.andThen(SqlTypeTransforms.TO_NULLABLE);
```
FORCE_NULLABLE -> FORCE_NULLABLE
```
  public static final SqlReturnTypeInference TO_COLLECTION_ELEMENT_FORCE_NULLABLE =
      TO_COLLECTION_ELEMENT.andThen(SqlTypeTransforms.FORCE_NULLABLE);
```

JiajunBernoulli: Yes, we should use old name.

I like using a new field to fix it. 

Please squash commits and I will merge it soon.

mihaibudiu: @JiajunBernoulli done rebasing/squashing. Thank you.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -620,6 +620,13 @@ public static SqlCall stripSeparator(SqlCall call) {
...
```
YiwenWu, line 620: Since only the ELEMENT operator is used, I think it can be removed

YiwenWu (approved):

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
mihaibudiu: Since it is public I was assuming it may be used in other third-party libraries. What is the policy about removing public fields in Calcite?

YiwenWu: This is a minor suggestion that can be ignored.
I'm not sure about the removal policy either.  Not deleting can avoid version upgrade issues, but maintenance may be lengthy.

rubenada: My two cents: in this particular case, I'd vote for removing the (now unused) `MULTISET_ELEMENT_NULLABLE`.
It is true that it might be used by a 3rd party library, but since the name was misleading, I think that those potential 3rd party libraries could be suffering from the same issue as CALCITE-6228, so by having a compilation problem after the next Calcite upgrade, they'd be force to analyze their usage of `MULTISET_ELEMENT_NULLABLE` and maybe switch to the new `MULTISET_ELEMENT_FORCE_NULLABLE` (or define their own `SqlReturnTypeInference` if they choose so).

mihaibudiu: I love removing code, I will do that right away.
I will squash the commits optimistically, assuming that it can be merged after that.

PR #3652, opened 2024-01-26
[CALCITE-6227] ELEMENT(NULL) causes an assertion failure

mihaibudiu:


core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -574,13 +574,13 @@ public static SqlCall stripSeparator(SqlCall call) {
...
+                if (type == null) {
+                  return opBinding.getTypeFactory().createSqlType(SqlTypeName.NULL);
+                }
                 return type;
               }
 
-              @Override public int size() {
-                return opBinding.getOperandCount();
-              }
+              @Override public int size() { return opBinding.getOperandCount(); }
```
mihaibudiu, line 583: I reindented this line because checkStyle was upset about it.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -1934,6 +1934,10 @@ void testLikeAndSimilarFails() {
         .columnType("DOUBLE NOT NULL");
     expr("element(multiset[multiset[cast(null as tinyint)]])")
         .columnType("TINYINT MULTISET NOT NULL");
+    // Test case for <a href="https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6227">
```
JiajunBernoulli, line 1937: Is the `a tag` valid here?

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -574,13 +574,13 @@ public static SqlCall stripSeparator(SqlCall call) {
                 RelDataType type =
                     opBinding.getOperandType(index)
                         .getComponentType();
-                assert type != null;
+                if (type == null) {
```
JiajunBernoulli, line 577: LGTM

PR #3653, opened 2024-01-26
[CALCITE-6226] Wrong ISOWEEK and no ISOYEAR on BigQuery FORMAT_DATE

rorueda:
https://issues.apache.org/jira/browse/CALCITE-6226

### DIFF
--- a/core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
+++ b/core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
@@ -52,7 +52,10 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%2d", calendar.get(Calendar.YEAR) / 100 + 1));
     }
   },
-  D("F", "The weekday (Monday as the first day of the week) as a decimal number (1-7)") {
+  D("", "The weekday (Sunday as the first day of the week) as a decimal number (1-7)") {
+    @Override public void toPattern(StringBuilder sb) throws UnsupportedOperationException {
+      throwToPatternNotImplemented();
+    }
     @Override public void format(StringBuilder sb, Date date) {
       final Calendar calendar = Work.get().calendar;
       calendar.setTime(date);
@@ -210,15 +213,42 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%02d", calendar.get(Calendar.HOUR_OF_DAY)));
     }
   },
-  // TODO: Ensure ISO 8601 for parsing
-  IW("w", "The ISO 8601 week number of the year (Monday as the first day of the week) "
-      + "as a decimal number (01-53)") {
+  ID("u", "The weekday (Monday as the first day of the week) as a decimal number (1-7)") {
     @Override public void format(StringBuilder sb, Date date) {
-      // TODO: ensure this is isoweek
       final Calendar calendar = Work.get().calendar;
       calendar.setTime(date);
-      calendar.setFirstDayOfWeek(Calendar.MONDAY);
+      int weekDay = calendar.get(Calendar.DAY_OF_WEEK);
+      // Converting Sun(1)...Sat(7) to Mon(1)...Sun(7)
+      sb.append(weekDay == 1 ? 7 : weekDay - 1);
+    }
+  },
+  IW("", "The ISO 8601 week number of the year (Monday as the first day of the week) "
+      + "as a decimal number (01-53). If the week containing January 1 has four or more days "
+      + "in the new year, then it is week 1; otherwise it is week 53 of the previous year, "
+      + "and the next week is week 1.") {
+    @Override public void toPattern(StringBuilder sb) throws UnsupportedOperationException {
+      throwToPatternNotImplemented();
+    }
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().iso8601Calendar;
+      calendar.setTime(date);
       sb.append(String.format(Locale.ROOT, "%02d", calendar.get(Calendar.WEEK_OF_YEAR)));
+    }
+  },
+  IYY("YY", "The ISO 8601 year without century as a decimal number. Each ISO year begins on "
+      + "the Monday before the first Thursday of the Gregorian calendar year.") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().iso8601Calendar;
+      calendar.setTime(date);
+      sb.append(String.format(Locale.ROOT, "%02d", calendar.getWeekYear() % 100));
+    }
+  },
+  IYYYY("YYYY", "The ISO 8601 year with century as a decimal number. Each ISO year begins on "
+      + "the Monday before the first Thursday of the Gregorian calendar year.") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().iso8601Calendar;
+      calendar.setTime(date);
+      sb.append(calendar.getWeekYear());
     }
   },
   MI("m", "The minute as a decimal number (00-59)") {
@@ -281,9 +311,8 @@ public enum FormatElementEnum implements FormatElement {
     }
   },
   Q("", "The quarter as a decimal number (1-4)") {
-    // TODO: Allow parsing of quarters.
     @Override public void toPattern(StringBuilder sb) throws UnsupportedOperationException {
-      throw new UnsupportedOperationException("Cannot convert 'Q' FormatElement to Java pattern");
+      throwToPatternNotImplemented();
     }
     @Override public void format(StringBuilder sb, Date date) {
       final Calendar calendar = Work.get().calendar;
@@ -369,7 +398,6 @@ public enum FormatElementEnum implements FormatElement {
     @Override public void format(StringBuilder sb, Date date) {
       final Calendar calendar = Work.get().calendar;
       calendar.setTime(date);
-      calendar.setFirstDayOfWeek(Calendar.SUNDAY);
       sb.append(String.format(Locale.ROOT, "%02d", calendar.get(Calendar.WEEK_OF_YEAR)));
     }
   },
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -132,17 +133,41 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%02d", calendar.get(Calendar.HOUR_OF_DAY)));
     }
   },
+  ID("u", "The weekday (Monday as the first day of the week) as a decimal number (1-7)") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      int weekDay = calendar.get(Calendar.DAY_OF_WEEK);
+      sb.append(String.format(Locale.ROOT, "%d", weekDay == 1 ? 7 : weekDay - 1));
```
tanclary, line 141: Does this just convert from Sunday being 1 to Monday being 1?

rorueda: Yes. Maybe change the comparison to `weekDay == Calendar.SUNDAY` to make it clearer? Do you see another way?

tanclary: That could work, a simple comment would help also. It is no big deal either way :)

tanclary: In my mind Calcite is complex enough that when we have opportunities to explain little pieces of logic to reduce mental load for the developer, we should

mihaibudiu: The DOW function implementation assumes that Sunday = 1.
In Postgres DOW assumes that Sunday = 0.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -132,17 +133,41 @@ public enum FormatElementEnum implements FormatElement {
...
-      // TODO: ensure this is isoweek
-      final Calendar calendar = Work.get().calendar;
+      final Calendar calendar = Work.get().iso8601Calendar;
       calendar.setTime(date);
-      calendar.setFirstDayOfWeek(Calendar.MONDAY);
       sb.append(String.format(Locale.ROOT, "%02d", calendar.get(Calendar.WEEK_OF_YEAR)));
     }
   },
+  IYY("YY", "The ISO 8601 year without century as a decimal number. Each ISO year begins on "
+      + "the Monday before the first Thursday of the Gregorian calendar year.") {
```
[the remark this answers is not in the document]
rorueda: Sure. Completely missed this test class.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -48,7 +48,8 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%2d", calendar.get(Calendar.YEAR) / 100 + 1));
     }
   },
-  D("F", "The weekday (Monday as the first day of the week) as a decimal number (1-7)") {
+  // TODO: Parsing of weekdays with sunday as first day of the week
```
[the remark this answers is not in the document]
julianhyde: Please remove the TODO, log a jira case.

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
```diff
@@ -77,8 +80,26 @@ class FormatElementEnumTest {
     assertFormatElement(FormatElementEnum.FF6, "2014-09-30T10:00:00.123456Z", "000123");
   }
 
-  @Test void testIW() {
-    assertFormatElement(FormatElementEnum.IW, "2014-09-30T10:00:00Z", "40");
+  @Test void testID() {
+    assertFormatElement(FormatElementEnum.ID, "2014-09-30T10:00:00Z", "2");
+  }
```
julianhyde, line 85: I don't understand why some of the tests in this class are parameterized and others are not.

Frankly I don't think the framework is helping make the tests more understandable. I would move to code, and add comments if a test is testing something subtle.

It seems to me that this test class should be very simple.

julianhyde (changes requested):

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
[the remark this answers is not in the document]
rorueda: Created CALCITE-6233 to handle the parsing of ISOWEEK.

It doesn't seem that parsing the weekday with Sunday as first day (1) is possible with java's DateFormat.

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
rorueda: As there is one method for each format element, I didn't want to add new methods to test the edge cases of  ISOWEEK and ISOYEAR, and I thought a good solution was to make those tests parameterized.

But I agree that the way it was done didn't help in making the added test cases understandable. I removed the parameters and made more assertions inside the method (with a comment).

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
rorueda: Added a comment - don't think my suggestion would make it clearer.

Looking at Postgres documentation, the `EXTRACT(DOW...` assumes Sunday = 0, but the format functions `to_char, to_date,...` also assume Sunday = 1 for day of week.

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
```diff
@@ -77,8 +77,32 @@ class FormatElementEnumTest {
...
 
-  @Test void testIW() {
+  @Test void testID() {
+    assertFormatElement(FormatElementEnum.ID, "2014-09-30T10:00:00Z", "2");
+  }
+
+  @Test
+  void testIW() {
     assertFormatElement(FormatElementEnum.IW, "2014-09-30T10:00:00Z", "40");
+    // edge case where ISO WEEK != WEEK
```
mihaibudiu, line 87: I have discovered that the Java Calendar class doesn't handle correctly some operations with dates before the Gregorian calendar introduction. Can you please add some tests for dates in that range?
https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6252
Can you also confirm that the values have been validated against BigQuery?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -12472,6 +12472,21 @@ void testTimestampDiff(boolean coercionEnabled) {
     f.checkScalar("FORMAT_DATE('The date is: %x', DATE '2008-12-25')",
         "The date is: 12/25/08",
         "VARCHAR(2000) NOT NULL");
+    f.checkScalar("FORMAT_DATE('%G-%V', DATE '2023-01-01')",
```
mihaibudiu, line 12475: same comment about dates before the Gregorian calendar

core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java
rorueda: I added additional test cases, all manually validated against BigQuery.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
rorueda: Done. I also had to fix one test case for PostgreSQL ISOWEEK - also manually validated using dbfiddle.

rorueda:
Since I had to rebase it onto master to address the last change request, I squashed the commits.

julianhyde:
It always struck me as strange that this change is using Calendar. Day of the week is simple if you convert to Julian date. JulianOn Mar 9, 2024, at 07:59, rorueda ***@***.***> wrote:﻿
@rorueda commented on this pull request.



In core/src/test/java/org/apache/calcite/util/format/FormatElementEnumTest.java:
>      assertFormatElement(FormatElementEnum.IW, "2014-09-30T10:00:00Z", "40");
+    // edge case where ISO WEEK != WEEK

I added additional test cases, all manually validated against BigQuery.

mihaibudiu (approved):
I trust that the results were all compared against the reference databases, I didn't check myself.

jduo (approved):

rorueda:
This is a bit old, so there were conflicts with the current master. I did a rebase and fixed the conflicts.

rubenada:
LGTM

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -436,8 +469,15 @@ static Work get() {
       return castNonNull(THREAD_WORK.get());
     }
 
-    final Calendar calendar =
-        Calendar.getInstance(DateTimeUtils.DEFAULT_ZONE, Locale.ROOT);
+    final Calendar calendar = new Calendar.Builder()
```
mihaibudiu, line 472: According to this issue https://issues.apache.org/jira/browse/CALCITE-6580
setLocale should not be used.

julianhyde: That bug refers to `Locale.setDefault`, which is a global change (affects the whole JVM).

This change seems harmless, because the behavior is deterministic, and does not depend on the locale of the user. 

According to <a href="https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Calendar.Builder.html#setLocale(java.util.Locale)">setLocale<a> the locale will override the locale of the format element, which is a slight change in behavior, and if it causes a bug we can revisit.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4828,7 +4828,7 @@ void testBitGetFunc(SqlOperatorFixture f, String functionName) {
           "2",
           "VARCHAR NOT NULL");
       f.checkString("to_char(timestamp '2022-06-03 13:15:48.678', 'IW')",
-          "23",
+          "22",
```
mihaibudiu, line 4831: this was a bug?

mihaibudiu (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
rorueda: Yes, same ISOWEEK issue, but for PostgreSQL:  https://github.com/apache/calcite/pull/3653#discussion_r1518608626

PR #3656, opened 2024-01-28
[CALCITE-6218] RelToSqlConverter fails to convert correlated lateral joins

YiwenWu:
fix the problem of  lateral field not exists。

https://issues.apache.org/jira/browse/CALCITE-6218


mihaibudiu (approved):

JiajunBernoulli:
I noticed that we are still discussing in Jira, so marked it.

YiwenWu:
> fix the problem of lateral field not exists。

@JiajunBernoulli Happy New Year，
I think this PR only deals with the problem of fields not existing. the optimization of lateral joins is another issue.

mihaibudiu (approved):

PR #3657, opened 2024-01-28
[CALCITE-6175] Upgrade opencsv version from 2.3 to 5.9

ymlai87416:
Jira link: https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6175

Upgrade opencsv to latest version 5.9


file/src/main/java/org/apache/calcite/adapter/file/CsvEnumerator.java
```diff
@@ -102,7 +101,7 @@ public CsvEnumerator(Source source, AtomicBoolean cancelFlag, boolean stream,
       } else {
         this.reader = openCsv(source);
       }
-      this.reader.readNext(); // skip header row
+      this.reader.readNextSilently(); // skip header row
```
macroguo-ghy, line 104: why `readNextSilently`?

PR #3658, opened 2024-01-29
[CALCITE-6221] JDBC adapter generates invalid query when the same table is joined multiple times

kramerul:


core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -257,6 +257,47 @@ public Result visit(Join e) {
...
+    }
+    Builder builder = rightResult.builder(e);
+    List<SqlNode> oldSelectList;
+    if (builder.select.getSelectList() == SqlNodeList.SINGLETON_STAR) {
+      oldSelectList = new ArrayList<>();
+      for (int i = 0; i < rightFieldNames.size(); i++) {
+        oldSelectList.add(new SqlIdentifier(rightFieldNames.get(i), POS));
+      }
+    } else {
+      oldSelectList = ImmutableList.<SqlNode>copyOf(builder.select.getSelectList().getList());
```
JiajunBernoulli, line 285: Please fix CI

kramerul: I fixed the CheckerFramework finding.

kramerul:
This PR could cause some problems. We had trouble with a statement like

```
SELECT 
FROM "TA" AS "A"  
INNER JOIN "TB" AS "B"
	ON "A"."APPLICATION_ID" =  "B"."CHILD_APPLICATION_ID"
INNER JOIN "TC" AS "C"
	ON "C"."APPLICATION_ID" = "B"."PARENT_APPLICATION_ID"
```

`"B"."CHILD_APPLICATION_ID"` was hidden from the additional SELECT.

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -257,6 +257,48 @@ public Result visit(Join e) {
...
+      for (SqlNode node : builder.select.getSelectList().getList()) {
+        oldSelectList.add(requireNonNull(node, "node"));
+      }
+    }
+    List<SqlNode> selectList = new ArrayList<>();
+    for (int i = 0; i < rightFieldNames.size(); i++) {
+      SqlNode column = oldSelectList.get(i);
+      if (!rightFieldNames.get(i).equals(fieldNames.get(offset + i))) {
+        column =
+            SqlStdOperatorTable.AS.createCall(POS, SqlUtil.stripAs(column),
```
mihaibudiu, line 293: What guarantees that this name is unique?
This may be another column name that appears on the other side.
I think you need to create a Set with all "forbidden" names and make sure that this name is not part of that set.

mihaibudiu (changes requested):

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -257,6 +257,48 @@ public Result visit(Join e) {
     return result(join, leftResult, rightResult);
   }
 
+  private Result maybeFixRenamedFields(Result rightResult, Join e) {
```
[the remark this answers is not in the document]
kramerul: The class `RelToSqlConverter` is only used in `JdbcImplementor` and `PigRelToSqlConverter`. Both classes are related to JDBC. Therefore, I would like to keep the title of the issue. Additionally Julian Hyde added the JDBC relation to the title

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -225,7 +225,7 @@ public Result visit(Join e) {
       break;
     }
     final Result leftResult = visitInput(e, 0).resetAlias();
-    final Result rightResult = visitInput(e, 1).resetAlias();
+    Result rightResult = maybeFixRenamedFields(visitInput(e, 1).resetAlias(), e);
```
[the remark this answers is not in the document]
kramerul: I changed this.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -809,8 +809,8 @@ private static String toSql(RelNode root, SqlDialect dialect,
     // RelFieldTrimmer maybe build the RelNode.
     relFn(fn).ok("SELECT \"t\".\"V\" AS \"l_v\"\n"
         + "FROM (VALUES (1, 2)) AS \"t\" (\"K\", \"V\")\n"
-        + "INNER JOIN "
-        + "(VALUES (1)) AS \"t0\" (\"K\") ON \"t\".\"K\" = \"t0\".\"K\"");
+        + "INNER JOIN (SELECT \"K\" AS \"K0\"\n"
```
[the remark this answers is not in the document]
kramerul: Yes it's quite hard to review all changes. I also had a look at most of the changes inside the tests.

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
kramerul: The uniqueness is currently guaranteed by the function `SqlValidatorUtil.deriveJoinRowType` which creates the unique field names for the `Join` relation. But this is not really visible here. Therefore, I will add a check.

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -257,6 +258,60 @@ public Result visit(Join e) {
...
+    // which was already done in SqlValidatorUtil.deriveJoinRowType
+    if (stack.stream().noneMatch(it -> it.r instanceof Filter)) {
+      return rightResult;
+    }
+    List<String> rightFieldNames = e.getRight().getRowType().getFieldNames();
+    List<String> fieldNames = e.getRowType().getFieldNames();
+    int offset = e.getLeft().getRowType().getFieldCount();
+    boolean hasFieldNameCollision = false;
+    for (int i = 0; i < rightFieldNames.size(); i++) {
+      if (!rightFieldNames.get(i).equals(fieldNames.get(offset + i))) {
```
mihaibudiu, line 280: why is it sufficient to compare with the field with the same relative index rather than compare all n^2 pairs?

kramerul:
I will rework the complete PR according to the last comment of Julian Hyde in https://issues.apache.org/jira/browse/CALCITE-6221

kramerul:
Closing this PR in favor of a new one.

PR #3661, opened 2024-01-31
[CALCITE-6236] EnumerableBatchNestedLoopJoin uses wrong row count for cost calculation

rubenada:
Store the join that originates the EnumerableBatchNestedLoopJoin, so that we can get the right rowCount from it.

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableBatchNestedLoopJoin.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableBatchNestedLoopJoin.java
@@ -55,6 +55,7 @@ import java.util.Set;
 public class EnumerableBatchNestedLoopJoin extends Join implements EnumerableRel {
 
   private final ImmutableBitSet requiredColumns;
+  private final Join originalJoin;
   protected EnumerableBatchNestedLoopJoin(
       RelOptCluster cluster,
       RelTraitSet traits,
@@ -63,9 +64,11 @@ public class EnumerableBatchNestedLoopJoin extends Join implements EnumerableRel
       RexNode condition,
       Set<CorrelationId> variablesSet,
       ImmutableBitSet requiredColumns,
-      JoinRelType joinType) {
+      JoinRelType joinType,
+      Join originalJoin) {
     super(cluster, traits, ImmutableList.of(), left, right, condition, variablesSet, joinType);
     this.requiredColumns = requiredColumns;
+    this.originalJoin = originalJoin;
   }
 
   public static EnumerableBatchNestedLoopJoin create(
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableBatchNestedLoopJoin.java
```diff
@@ -63,9 +64,11 @@ protected EnumerableBatchNestedLoopJoin(
...
       ImmutableBitSet requiredColumns,
-      JoinRelType joinType) {
+      JoinRelType joinType,
+      Join originalJoin) {
     super(cluster, traits, ImmutableList.of(), left, right, condition, variablesSet, joinType);
     this.requiredColumns = requiredColumns;
+    this.originalJoin = originalJoin;
   }
 
   public static EnumerableBatchNestedLoopJoin create(
```
JiajunBernoulli, line 74: This is public function.

Can we have an old function that be marked as Deprecated?

rubenada: @JiajunBernoulli  we could do that, but in that case if the deprecated function is used, we will have an EnumerableBatchNestedLoopJoin without its originalJoin (so originalJoin would be `null`), what shall we do in that scenario for the RelMdRowCount computation of the EBNLJ?

The alternative would be, we do this as a breaking change to ensure we apply the correction in all cases (we mention on the next release note that "EnumerableBatchNestedLoopJoin now requires on its construction a new parameter: the Join that originated it", tbd before merging the PR). And when downstream projects upgrade, they'd need to adjust their code (if they create EBNLJ on their own).

Both alternatives have pros and cons. Personally I prefer the second one, but I'm open to discuss it.

JiajunBernoulli: Second is ok.

Let's emphasize this change in JIRA.

rubenada: Ok @JiajunBernoulli .
I have committed the comment on the next RN about the breaking change.
When the PR gets merged, I will mention it also in the Jira.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableBatchNestedLoopJoin.java
```diff
@@ -55,6 +55,7 @@
 public class EnumerableBatchNestedLoopJoin extends Join implements EnumerableRel {
 
   private final ImmutableBitSet requiredColumns;
+  private final Join originalJoin;
```
zabetak, line 58: Keeping a RelNode in another RelNode is a strange pattern that we shall better avoid. Other than that adding attributes may necessitate touching the digest which may further complicate the situation.

zabetak (commented):
Left some comments about an alternative fix under the JIRA.

PR #3663, opened 2024-01-31
[CALCITE-2067] RexLiteral cannot represent accurately floating point values, including NaN, Infinity

mihaibudiu:
This is another attempt to fix [CALCITE-2067]. 
PR #572 attempted to do this, but wasn't merged about 8 years ago.
This PR includes very similar changes.

The current implementation of RexLiteral uses a BigDecimal value to store DOUBLE values. This is a problem for 2 reasons:
- it loses precision, since some FP values cannot be stored precisely in decimal notation
- it cannot represent special FP values, such as -0.0, Infinity, or NaN

With this PR these issues are solved. As a side-effect, this enables the simplification engine to simplify some expressions that it couldn't before, such as expressions that produce special FP values.

Objections to the original PR which can be seen here: https://issues.apache.org/jira/browse/CALCITE-2067 include the fact that this "changes the SQL language accepted by Calcite." I want to preempt this objection: this change does not affect the parser or SqlNode, only RexNode. So it is only useful for *evaluation* and *optimization*. Moreover, if we will add support to the parser for special values as suggested in https://issues.apache.org/jira/browse/CALCITE-6058, we will certainly need a way to represent them in the RelNode IR. I claim that the changes in this PR are necessary for that goal.

The most unpleasant effect of this PR is to slightly change the semantics of some programs that used FP values before. These programs were in fact buggy, but third parties may have taken dependencies on these bugs. This kind of behavior was discovered by comparing results produced by Postgres (or Java) with Calcite on FP functions; the results were slightly off. The BigDecimal-induced rounding was the problem.

I also think that the constructor for RexLiteral that creates a literal for a DOUBLE type from a BigDecimal should be deprecated. I didn't do it in this PR.

### DIFF
--- a/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
+++ b/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
@@ -228,6 +228,8 @@ class ArrowTranslator {
       }
     } else if (String.class.equals(literal.getClass())) {
       return "string";
+    } else if (literal instanceof Double) {
+      return "float";
     }
     throw new UnsupportedOperationException("Unsupported literal " + literal);
   }
--- a/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
+++ b/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
@@ -373,7 +373,7 @@ class ArrowAdapterTest {
     String sql = "select * from arrowdata\n"
         + " where \"floatField\"=15.0";
     String plan = "PLAN=ArrowToEnumerableConverter\n"
-        + "  ArrowFilter(condition=[=(CAST($2):DOUBLE, 15.0)])\n"
+        + "  ArrowFilter(condition=[=(CAST($2):DOUBLE, 15.0E0)])\n"
         + "    ArrowTableScan(table=[[ARROW, ARROWDATA]], fields=[[0, 1, 2, 3]])\n\n";
     String result = "intField=15; stringField=15; floatField=15.0; longField=15\n";
 
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -1498,7 +1498,7 @@ public abstract class SqlImplementor {
     case EXACT_NUMERIC: {
       if (SqlTypeName.APPROX_TYPES.contains(typeName)) {
         return SqlLiteral.createApproxNumeric(
-            castNonNull(literal.getValueAs(BigDecimal.class)).toPlainString(), POS);
+            castNonNull(literal.getValueAs(Double.class)).toString(), POS);
       } else {
         return SqlLiteral.createExactNumeric(
             castNonNull(literal.getValueAs(BigDecimal.class)).toPlainString(), POS);
--- a/core/src/main/java/org/apache/calcite/rex/RexBuilder.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexBuilder.java
@@ -1437,7 +1437,21 @@ public class RexBuilder {
   public RexLiteral makeApproxLiteral(@Nullable BigDecimal bd, RelDataType type) {
     assert SqlTypeFamily.APPROXIMATE_NUMERIC.getTypeNames().contains(
         type.getSqlTypeName());
-    return makeLiteral(bd, type, SqlTypeName.DOUBLE);
+    return makeLiteral(bd != null ? bd.doubleValue() : null, type, SqlTypeName.DOUBLE);
+  }
+
+  /**
+   * Creates an approximate numeric literal (double or float)
+   * from a Double value.
+   *
+   * @param val  literal value
+   * @param type approximate numeric type
+   * @return new literal
+   */
+  public RexLiteral makeApproxLiteral(Double val, RelDataType type) {
+    assert SqlTypeFamily.APPROXIMATE_NUMERIC.getTypeNames().contains(
+        type.getSqlTypeName());
+    return makeLiteral(val, type, SqlTypeName.DOUBLE);
   }
 
   /**
@@ -2013,7 +2027,10 @@ public class RexBuilder {
     case FLOAT:
     case REAL:
     case DOUBLE:
-      return makeApproxLiteral((BigDecimal) value, type);
+      if (value instanceof Double) {
+        return makeApproxLiteral((Double) value, type);
+      }
+      return makeApproxLiteral(((BigDecimal) value).doubleValue(), type);
     case BOOLEAN:
       return (Boolean) value ? booleanTrue : booleanFalse;
     case TIME:
@@ -2159,13 +2176,24 @@ public class RexBuilder {
               type.getSqlTypeName());
       return new BigDecimal(((Number) o).longValue());
     case REAL:
-    case FLOAT:
-    case DOUBLE:
       if (o instanceof BigDecimal) {
         return o;
       }
-      return new BigDecimal(((Number) o).doubleValue(), MathContext.DECIMAL64)
+      // Float values are stored as Doubles
+      if (o instanceof Float) {
+        return ((Float) o).doubleValue();
+      }
+      if (o instanceof Double) {
+        return o;
+      }
+      return new BigDecimal(((Number) o).doubleValue(), MathContext.DECIMAL32)
           .stripTrailingZeros();
+    case FLOAT:
+    case DOUBLE:
+      if (o instanceof Double) {
+        return o;
+      }
+      return ((Number) o).doubleValue();
     case CHAR:
     case VARCHAR:
       if (o instanceof NlsString) {
--- a/core/src/main/java/org/apache/calcite/rex/RexLiteral.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexLiteral.java
@@ -55,6 +55,7 @@ import org.locationtech.jts.geom.Geometry;
 
 import java.io.PrintWriter;
 import java.math.BigDecimal;
+import java.math.MathContext;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.text.SimpleDateFormat;
@@ -112,9 +113,11 @@ import static java.util.Objects.requireNonNull;
  * <td>{@link BigDecimal}</td>
  * </tr>
  * <tr>
- * <td>{@link SqlTypeName#DOUBLE}</td>
+ * <td>{@link SqlTypeName#DOUBLE},
+ *     {@link SqlTypeName#REAL},
+ *     {@link SqlTypeName#FLOAT}</td>
  * <td>Approximate number, for example <code>6.023E-23</code>.</td>
- * <td>{@link BigDecimal}</td>
+ * <td>{@link Double}.</td>
  * </tr>
  * <tr>
  * <td>{@link SqlTypeName#DATE}</td>
@@ -193,7 +196,7 @@ public class RexLiteral extends RexNode {
   /**
    * The value of this literal. Must be consistent with its type, as per
    * {@link #valueMatchesType}. For example, you can't store an
-   * {@link Integer} value here just because you feel like it -- all numbers are
+   * {@link Integer} value here just because you feel like it -- all exact numbers are
    * represented by a {@link BigDecimal}. But since this field is private, it
    * doesn't really matter how the values are stored.
    */
@@ -204,12 +207,9 @@ public class RexLiteral extends RexNode {
    */
   private final RelDataType type;
 
-  // TODO jvs 26-May-2006:  Use SqlTypeFamily instead; it exists
-  // for exactly this purpose (to avoid the confusion which results
-  // from overloading SqlTypeName).
   /**
    * An indication of the broad type of this literal -- even if its type isn't
-   * a SQL type. Sometimes this will be different than the SQL type; for
+   * a SQL type. Sometimes this will be different from the SQL type; for
    * example, all exact numbers, including integers have typeName
    * {@link SqlTypeName#DECIMAL}. See {@link #valueMatchesType} for the
    * definitive story.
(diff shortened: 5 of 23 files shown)


mihaibudiu:
@julianhyde you had most objections to the previous PR addressing this issue. I hope that you will find my arguments adequate.

core/src/main/java/org/apache/calcite/util/Util.java
```diff
@@ -544,6 +544,21 @@ public static void println(
     pw.println();
   }
 
+  /**
+   * Formats a double value to a String ensuring that the output
+   * is in scientific notation if the value is not "special".
+   * (Special values include infinities and NaN.)
+   */
+  public static String toScientificNotation(Double d) {
```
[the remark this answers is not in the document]
julianhyde: Should this method guarantee that there is a unique string for every possible `double` value? (Except for the many representations of `NaN` etc.) And that it has an inverse method that will convert the string to the same `double` value.

I believe that we are relying on this bijection, because there is no direct way to specify a binary floating point value in SQL.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -3027,6 +3027,31 @@ private void checkPushJoinThroughUnionOnRightDoesNotMatchSemiOrAntiJoin(JoinRelT
         .check();
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-2067">
```
julianhyde, line 3030: Conventional format includes `[`.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1110,6 +1110,14 @@ public RexLiteral makeApproxLiteral(BigDecimal bd) {
     return makeApproxLiteral(bd, typeFactory.createSqlType(SqlTypeName.DOUBLE));
   }
 
+  /**
```
julianhyde, line 1113: I think this method is too easy to invoke by accident.

julianhyde (changes requested):
The summary "RexBuilder can't handle NaN,Infinity double constants" doesn't describe the full scope of the work. This change allows literals to have scaled binary values.

core/src/main/java/org/apache/calcite/rex/RexLiteral.java
```diff
@@ -325,19 +324,18 @@ public static boolean valueMatchesType(
...
       // fall through
     case DECIMAL:
+    case BIGINT:
+      return value instanceof BigDecimal;
     case DOUBLE:
     case FLOAT:
     case REAL:
-    case BIGINT:
-      return value instanceof BigDecimal;
+      return value instanceof BigDecimal || value instanceof Double;
```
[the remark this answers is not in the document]
mihaibudiu: Indeed, the decimal option should be removed, but I kept it for backward compatibility with third-party users of Calcite.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
mihaibudiu: Yes, this should be deprecated. Can I do that?

core/src/main/java/org/apache/calcite/util/Util.java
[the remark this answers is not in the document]
mihaibudiu: Unfortunately there is no bijection. This conversion is lossy in both directions, from String to Double and from Double to String. This method is not 100% adequate for generating SQL, but it is adequate for e.g., displaying plans. 

However, the lossy conversion problem already exists in Calcite and is not specific to DOUBLE values: see https://issues.apache.org/jira/browse/CALCITE-5987. I have discussed in a message on the dev list about how this separate issue can be tackled.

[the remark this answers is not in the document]
mihaibudiu: I realize that my previous message is wrong, indeed, there is no way to specify a binary floating point value in other SQL dialects. I don't think we can solve this problem, but we didn't introduce this problem in this PR. The problem exists already if one allows simplifications that can produce FP results.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
mihaibudiu: I have deleted the method

core/src/main/java/org/apache/calcite/rex/RexLiteral.java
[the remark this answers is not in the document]
mihaibudiu: I have removed the choice here too.

mihaibudiu:
@julianhyde I hope to have addressed your comments.
This is a breaking change in some respect; one can see how the Druid Adapter code had to be modified to accommodate this change. But I think it fixes a genuine bug. It doesn't solve the problem of FP literals completely, but it enables more programs to produce correct results than before.

rubenada:
bq. This is a breaking change in some respect [...]
Just a minor reminder: before the final merge, @mihaibudiu , I think it'd be necessary to include the corresponding comment about this on `history.md` in the "breaking changes" section for the next release 1.37 (so that we don't forget to mention it when preparing the next release).

mihaibudiu:
@rubenada I have added a commented-out breaking change for the next release. Thank you for the reminder.

mihaibudiu:
@julianhyde I think you have requested changes, I hope I have addressed your comments.

core/src/main/java/org/apache/calcite/rex/RexLiteral.java
```diff
@@ -1060,22 +1065,46 @@ public boolean isNull() {
...
+        if (clazz == Long.class) {
+          return clazz.cast(d.longValue());
+        } else if (clazz == Integer.class) {
+          return clazz.cast(d.intValue());
+        } else if (clazz == Short.class) {
+          return clazz.cast(d.shortValue());
+        } else if (clazz == Byte.class) {
+          return clazz.cast(d.byteValue());
+        } else if (clazz == Double.class) {
+          return clazz.cast(d);
```
jduo, line 1099: Isn't this just d?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -388,7 +388,7 @@ private static String toSql(RelNode root, SqlDialect dialect,
         + "where \"net_weight\" <> 10 or \"net_weight\" is null";
     final String expected = "SELECT \"product_id\", \"shelf_width\"\n"
         + "FROM \"foodmart\".\"product\"\n"
-        + "WHERE \"net_weight\" <> 10 OR \"net_weight\" IS NULL";
+        + "WHERE \"net_weight\" <> CAST(10 AS DOUBLE) OR \"net_weight\" IS NULL";
```
jduo, line 391: Would users need to make changes to queries such as these? Is 10 no longer implicitly converted to double when comparing to double columns?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -4065,7 +4065,7 @@ private SqlDialect nonOrdinalDialect() {
         + "  select \"product_id\", 0 as \"net_weight\"\n"
         + "  from \"sales_fact_1997\") t0";
     final String expected = "SELECT SUM(CASE WHEN \"product_id\" = 0"
-        + " THEN \"net_weight\" ELSE 0 END) AS \"NET_WEIGHT\"\n"
+        + " THEN \"net_weight\" ELSE 0E0 END) AS \"NET_WEIGHT\"\n"
```
jduo, line 4068: I think we'd be hard-pressed to have users use this literal FYI

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
mihaibudiu: no, this is just the canonical representation of the program created by Calcite which proves that the 10 is implicitly cast to a double

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
mihaibudiu: same as above, they don't have to.

core/src/main/java/org/apache/calcite/rex/RexLiteral.java
mihaibudiu: Yes, it is, but the return type must be T, so the cast is necessary anyway. The Java compiler isn't smart enough to understand that.

mihaibudiu:
This PR is also fixing a 9-year feature request in Calcite.
From some remarks made by @julianhyde (which unfortunately I cannot find online) I deduce that he doesn't object to this change anymore.

mihaibudiu:
By reviewing this PR you can help close a 9-year issue for Calcite!
The changes look much bigger than they are because tests are affected, but the logic is actually very simple.

caicancai:
@mihaibudiu I will review this PR, before version 1.38.0, which is expected to be completed tomorrow

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
```diff
@@ -228,6 +228,8 @@ private static String getLiteralType(Object literal) {
       }
     } else if (String.class.equals(literal.getClass())) {
       return "string";
+    } else if (literal instanceof Double) {
+      return "float";
```
caicancai, line 232: why `float`?

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1498,7 +1498,7 @@ public static SqlNode toSql(RexLiteral literal) {
     case EXACT_NUMERIC: {
       if (SqlTypeName.APPROX_TYPES.contains(typeName)) {
         return SqlLiteral.createApproxNumeric(
-            castNonNull(literal.getValueAs(BigDecimal.class)).toPlainString(), POS);
+            castNonNull(literal.getValueAs(Double.class)).toString(), POS);
```
caicancai, line 1501: Maybe we should add Javadoc to prevent someone from changing it back to BigDecimal later.

caicancai:
Overall LGTM

core/src/test/java/org/apache/calcite/rex/RexExecutorTest.java
```diff
@@ -389,11 +389,14 @@ interface Action {
...
+      // Division by zero causes an exception during evaluation
       final RexCall second =
-          (RexCall) rexBuilder.makeCall(SqlStdOperatorTable.LN,
-          rexBuilder.makeLiteral(-2, integer, true));
+          (RexCall) rexBuilder.makeCall(SqlStdOperatorTable.DIVIDE_INTEGER,
+              rexBuilder.makeLiteral(-2, integer, true),
+              rexBuilder.makeLiteral(0, integer, true));
       executor.reduce(rexBuilder, ImmutableList.of(first, second),
           reducedValues);
+      System.out.println(reducedValues);
```
julianhyde, line 399: Let's remove this `println`

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
```diff
@@ -373,7 +373,7 @@ static void initializeArrowState(@TempDir Path sharedTempDir)
     String sql = "select * from arrowdata\n"
         + " where \"floatField\"=15.0";
     String plan = "PLAN=ArrowToEnumerableConverter\n"
-        + "  ArrowFilter(condition=[=(CAST($2):DOUBLE, 15.0)])\n"
+        + "  ArrowFilter(condition=[=(CAST($2):DOUBLE, 15.0E0)])\n"
```
julianhyde, line 376: I see a lot of `E0` added. Is it a goal of this change to make decimal literals print differently from double literals?

Maybe I missed it, but somewhere in the javadoc (e.g. for RexLiteral) should state that goal.

mihaibudiu: I haven't changed how floating point values are serialized as text. The difference comes because now more literals contain a double/float.

julianhyde: Makes sense.

+1 when you fix the issues in my other 2 comments

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
mihaibudiu: this is the same string that is used 3 lines above

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
mihaibudiu: If they change this to BigDecimal the program will crash with a type cast error.

julianhyde: It might crash today, but it's possible that someone will enhance `RexLiteral.getValueAs` to accept `BigDecimal.class` as an argument. That's not an unreasonable change.

I had different concerns about this code, namely that assumes that how Java formats Double values is the same as how SQL dialects represent FLOAT literals.

I am satisfied that code is OK only because there are already quite extensive unit tests for SQL generation.

mihaibudiu: BigDecimal.class should still be a legal argument for this function. 
There are a few places in the code where this happens.

This code assumes only that the createApproxNumeric will correctly handle Java string representations of double values.

mihaibudiu:
I have pushed a new commit which hopefully addresses the review comments.

caicancai (approved):
LGTM

mihaibudiu:
This PR has been approved in JIRA, so I will merge it.

PR #3664, opened 2024-02-01
[CALCITE-6221] JDBC adapter generates invalid query when the same table is joined multiple times

kramerul:


core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -437,11 +437,43 @@ public Result visit(Filter e) {
...
+  private static boolean fieldRequiresAlias(SqlNode field, String fieldName) {
+    if (!(field instanceof SqlIdentifier)) {
+      return true;
+    }
+    SqlIdentifier identifier = (SqlIdentifier) field;
+    if (identifier.isStar()) {
+      return true;
+    }
+    List<String> names = identifier.names;
+    return !names.get(names.size() - 1).equals(fieldName);
```
mihaibudiu, line 474: This is the inverse of what I expected.
Maybe you need to document this function if this is correct.

kramerul: I renamed this function to `fieldWasRenamedByJoinDueToNameClash`

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -437,11 +437,43 @@ public Result visit(Filter e) {
...
+          builder.setSelect(new SqlNodeList(selectList.build(), POS));
+        }
+      }
       builder.setWhere(builder.context.toSql(null, e.getCondition()));
       return builder.result();
     }
   }
 
+  private static boolean fieldWasRenamedByJoinDueToNameClash(SqlNode field, String fieldName) {
+    if (!(field instanceof SqlIdentifier)) {
```
YiwenWu, line 466: context(AliasContext) field is only allowed to return SqlIdentifier.
and Is there any other simpler way to determine whether rename is needed?
can we identify the Duplicate name by Result.qualifiedContext().fieldList()

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -437,11 +437,43 @@ public Result visit(Filter e) {
...
+      return true;
+    }
+    SqlIdentifier identifier = (SqlIdentifier) field;
+    if (identifier.isStar()) {
+      return true;
+    }
+    List<String> names = identifier.names;
+    return !names.get(names.size() - 1).equals(fieldName);
```
YiwenWu, line 474: can be simplified to  `last(names)equals(fieldName)`

YiwenWu (approved):

YiwenWu (commented):
Is this bug only handled by the Filter operator? Are there any other case need to be addressed?

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
kramerul: I changed it.

kramerul:
I'm not so deep into Calcite that I would be able to say if there are other cases. As far as I know, this is the only case because
* columns are only renamed in the `Join` class
* the generation of selected columns in SQL is only missing for `Filter`

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
kramerul: I also thought about simplifying this on base of `Result.qualifiedContext().fieldList()`. But I had no idea how this could be achieved.

YiwenWu: Whether this method can satisfy？
```
private static boolean fieldRequiresAlias(Result x) {
  List<String> uniqueList = new ArrayList();
  for(SqlNode node : x.qualifiedContext().fieldList()) {
    if(node instance SqlIdentifier) { 
      SqlIdentifier field = (SqlIdentifier) node;
      if (uniqueList.contains(last(field.names))) {
          return true;
      }
      uniqueList.add(last(field.names));
    }
  }
  return false;
}
```

kramerul: I changed it accordingly.

YiwenWu (approved):

YiwenWu (approved):

kramerul:
We are still interested in this PR

mihaibudiu:
This has an approval, any reason not to merge?

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -438,11 +440,41 @@ public Result visit(Filter e) {
...
+          builder.setSelect(new SqlNodeList(selectList.build(), POS));
+        }
+      }
       builder.setWhere(builder.context.toSql(null, e.getCondition()));
       return builder.result();
     }
   }
 
+  private static boolean selectListRequired(Context context) {
+    Set<String> uniqueFieldNames = new HashSet<>();
```
mihaibudiu, line 465: this could have benefited from a little comment: a select list is required if any identifier shows up twice (from the left and right inputs of a join?)

mihaibudiu (approved):

PR #3665, opened 2024-02-02
[CALCITE-6236] EnumerableBatchNestedLoopJoin::estimateRowCount returns wrong value

kramerul:
This is another approach to fix [CALCITE-6236] using the rowCount of the original relation.

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/enumerable/EnumerableBatchNestedLoopJoinTest.java
+++ b/core/src/test/java/org/apache/calcite/test/enumerable/EnumerableBatchNestedLoopJoinTest.java
@@ -226,6 +226,7 @@ class EnumerableBatchNestedLoopJoinTest {
             + "join locations l on e.empid <> l.empid and d.deptno = l.empid")
         .withHook(Hook.PLANNER, (Consumer<RelOptPlanner>) planner -> {
           planner.removeRule(EnumerableRules.ENUMERABLE_CORRELATE_RULE);
+          planner.removeRule(EnumerableRules.ENUMERABLE_JOIN_RULE);
           // Use a small batch size, otherwise we will run into Janino's
           // "InternalCompilerException: Code of method grows beyond 64 KB".
           planner.addRule(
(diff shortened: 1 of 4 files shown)


kramerul:
We are still interested in getting this PR merged.

core/src/test/java/org/apache/calcite/test/enumerable/EnumerableBatchNestedLoopJoinTest.java
```diff
@@ -226,6 +226,7 @@ class EnumerableBatchNestedLoopJoinTest {
             + "join locations l on e.empid <> l.empid and d.deptno = l.empid")
         .withHook(Hook.PLANNER, (Consumer<RelOptPlanner>) planner -> {
           planner.removeRule(EnumerableRules.ENUMERABLE_CORRELATE_RULE);
+          planner.removeRule(EnumerableRules.ENUMERABLE_JOIN_RULE);
```
mihaibudiu, line 229: is this related to the problem addressed?

mihaibudiu (changes requested):
There are no tests...
There was a healthy discussion in the JIRA, hopefully some of the people who commented can also review.

core/src/test/java/org/apache/calcite/test/enumerable/EnumerableBatchNestedLoopJoinTest.java
kramerul: It's related to the test. When the rule is not removed, the planner creates a `EnumerableHashJoin` instead of a `EnumerableBatchNestedLoopJoin` because the cost function is affected by this change.

rubenada:
Sorry but the proposed approach (storing "originalRowCount" on the BNLJ) seems to go against some of the concepts discussed [on the Jira](https://issues.apache.org/jira/browse/CALCITE-6236?focusedCommentId=17813606&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17813606):
```
pre-storing selectivity or other kind of metadata information in the operator also feels like we are going to the wrong direction
```

This is certainly a tricky one. IMO if we can't reach consensus about this topic, it'd not be a critical problem since each downstream project can always define their own `MyRelMdRowCount` handler and put in there whatever logic the require to refine BNLJ rowCount estimation.

kramerul:
I will close this PR. We will use a fix inside our application.

PR #3666, opened 2024-02-02
[CALCITE-2040] Create adapter for Apache Arrow

macroguo-ghy:


asolimando (commented):
@macroguo-ghy since the PR is already big enough for Arrow part, what do you say if all the Kafka changes are taken out and addressed in another PR/Jira ticket?

asolimando (commented):
From a cursory look, a good number of the 21 Sonar issues could be easily fixed and they would improve the quality of the PR.

Could you do a pass to review and fix them where it makes sense?

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowEnumerator.java
asolimando, line 1: @macroguo-ghy looking at [`ArrowTable`](https://github.com/apache/calcite/blob/5bb4d6c63b08fc27d02e6a417a010a84c6d0106d/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java#L97) it seems that my understanding that a query can only have projection or filter but not both is confirmed (of course that's not from a logical standpoint as it wouldn't make sense, it's necessarily an implementation choice, as the tests clearly describe).

Based on this consideration, I am proposing to replace the `ArrowEnumerator` class with an abstract class and two sub-classes, one for filter, one for projection, you can find the associated code here if you want to take a look: https://github.com/asolimando/calcite/tree/arrow-fix

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
```diff
@@ -0,0 +1,503 @@
...
+        + "      ArrowTableScan(table=[[ARROW, ARROWDATA]], fields=[[0, 1, 2, 3]])\n\n";
+    String result = "intField=25\n";
+    CalciteAssert.that()
+        .with(arrow)
+        .query(sql)
+        .returns(result)
+        .explainContains(plan);
+  }
+
+  // TODO: test a table whose field names contain spaces,
```
asolimando, line 239: From a cursory look, it seems that we should add at least all the tests associated with these TODOs, and possibly others.

asolimando (changes requested):
@macroguo-ghy apart from `ArrowEnumerable` and `ArrowEnumerator` I haven't looked at other files from that folder, and very quickly checked the tests to confirm some hypotheses, so the review is only partial at this stage, I should be able to finish it by the end of next week, ideally.

I feel there is already enough feedback for you to check, if you feel like it, otherwise you can wait for my review to be over, as you prefer.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
asolimando: There seems to be consensus in the [Jira ticket](https://issues.apache.org/jira/browse/CALCITE-2040) to get the consolidated version of what we currently have in, and file tickets for the current bugs and limitations. 

So we can either implement those tests and disable them if not working (and of course filing an associated ticket for "family" of failing tests, like "add support for queries over multiple tables" etc.), or directly file a ticket and add what's in the TODO as part of the description, so we don't lose the context those comments provide, before deleting them.

Let's see what will be better when we reach a more mature state.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowRules.java
```diff
@@ -0,0 +1,175 @@
...
+  public static final ArrowFilterRule FILTER_SCAN =
+      ArrowFilterRule.Config.DEFAULT.toRule();
+
+  public static final ArrowToEnumerableConverterRule TO_ENUMERABLE =
+      ArrowToEnumerableConverterRule.DEFAULT_CONFIG
+          .toRule(ArrowToEnumerableConverterRule.class);
+
+  public static final List<RelOptRule> RULES =
+      ImmutableList.of(PROJECT_SCAN,
+          FILTER_SCAN);
```
asolimando, line 59: I am sorry to insist with this kind of changes, but IMO all these un-needed newlines make the code less readable, the amount of lines that can fit into the screen at the same time helps, if we waste many lines in this way, we end up scrolling a lot for not much

asolimando (changes requested):
I am missing the `ArrowData` and `ArrowAdapterTest` test classes, I will look at them by the end of the week).

I have left more comments and you can find two additional commits in the usual branch: https://github.com/asolimando/calcite/tree/arrow-fix

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
asolimando: I have done an improvement pass on the tests (see [a90ddb8](https://github.com/asolimando/calcite/commit/a90ddb8bac43ac1716f46716d47de5e9a431c902)), for the remaining tests I would do the same (try to implement them, and disabled the test if not currently working)

asolimando (changes requested):
@macroguo-ghy: I am done reviewing, for the biggest changes I have provided commits in https://github.com/asolimando/calcite/tree/arrow-fix, for the rest the comments should be enough to move forward.

If you need help implementing the changes (either clarifications or for implementing them), let me know.

macroguo-ghy:
Thank you for your review! @asolimando 

I have addressed most of the comments you raised and cherry-picked your commmits.

I will log tickets for unimplemented features

macroguo-ghy (commented):
I have logged some sub-task tickets under the CALCITE-2040. 
Please feel free to ping me when you finish test refactor, I will cherry-pick it.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
asolimando: @macroguo-ghy could you file an additional ticket for improving test coverage and adding the different tests marked here as TODO in the description, and remove them from the file?

Because we don't really use TODOs in Calcite, but the suggested tests are good, so it would be a pity if we lost track of them.

asolimando (changes requested):
@macroguo-ghy great work, thanks for taking care of the many comments I had and for filing the tickets. I have left one more comment about removing the TODOs (move them into a ticket) and I have also commented that we should create a separate umbrella ticket for the tickets you have created.

Marking "Request changes" so we don't forget, but I have no further comment, once this is addressed for me we are good.

PR #3668, opened 2024-02-02
[CALCITE-6239, CALCITE-6263] Add PostGIS Dialect to Apache Calcite

bchapuis:
This PR introduces a new PostgisSqlDialect to Apache Calcite, providing initial support for PostGIS functions. It primarily addresses [CALCITE-6239](https://issues.apache.org/jira/browse/CALCITE-6239) and follows the architectural approach proposed by @julianhyde in [CALCITE-6263](https://issues.apache.org/jira/browse/CALCITE-6263).

**Summary of the changes:**
- New Dialect: PostgisSqlDialect class extends PostgresqlSqlDialect, specializing SQL generation for PostGIS functions.
- ST_ Function Support: Adds SQL translation for selected ST_ prefixed spatial functions (e.g., ST_Contains, ST_Within, etc.), enabling Calcite to generate syntactically correct PostGIS SQL from relational algebra expressions using these functions.
- Function Mapping: Introduced mapping from Calcite’s standard operator representations to PostGIS-compatible SQL functions.
- Test Coverage: Added unit tests to validate the translation logic for supported spatial functions under the new dialect.

**Coupling with Java Topology Suite (JTS):**
The main concern regarding this PR was about avoiding excessive coupling between Calcite’s core and third-party spatial libraries like JTS as described in [CALCITE-6263](https://issues.apache.org/jira/browse/CALCITE-6263). We now have a Built-in GEOMETRY Type independant from JTS. The coupling with JTS occurs in `JavaTypeFactoryImpl` which is better than in `RelDataTypeFactoryImpl`.


core/src/main/java/org/apache/calcite/sql/dialect/PostgisSqlDialect.java
```diff
@@ -0,0 +1,51 @@
...
+      .withDataTypeSystem(POSTGRESQL_TYPE_SYSTEM);
+
+  public static final SqlDialect DEFAULT = new PostgisSqlDialect(DEFAULT_CONTEXT);
+
+  /** Creates a PostgresqlSqlDialect. */
+  public PostgisSqlDialect(Context context) {
+    super(context);
+  }
+
+  @Override public void unparseCall(SqlWriter writer, SqlCall call, int leftPrec, int rightPrec) {
```
YiwenWu, line 43: Since PostGIS is an extension of PostgreSQL, Can these changes be processed directly in PostgresqlSqlDialect?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7876,6 +7878,26 @@ private void checkLiteral2(String expression, String expected) {
         .withSpark().ok(sparkExpected);
   }
 
+  /**
+   * Test case for ST_SRID function.
+   * All the spatial functions where the arguments are the same type should behave similarly.
+   */
+  @Test void testPostgisStSrid() {
+    String query = "select st_srid(\"point\") FROM \"points\"";
+    String expectedPostgis = "SELECT \"ST_SRID\"(\"point\")\nFROM \"GEO\".\"points\"";
```
YiwenWu, line 7887: Some UT tests failed

core/src/main/java/org/apache/calcite/sql/dialect/PostgisSqlDialect.java
bchapuis: Yes they could, I also had mixed feeling about extending the postgresql dialect. My current reasoning comes from the fact that postgis is an extension that must be installed on the server and then activated in the database with `CREATE EXTENSION postgis;`. Having a separate dialect makes things more explicit in calcite and isolates postgis specific customizations in a dedicated `unparseCall` method.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
bchapuis: I should probably have created a draft PR.

core/src/main/java/org/apache/calcite/sql/SqlDialectFactoryImpl.java
```diff
@@ -222,6 +223,8 @@ public class SqlDialectFactoryImpl implements SqlDialectFactory {
       return PhoenixSqlDialect.DEFAULT;
     case POSTGRESQL:
       return PostgresqlSqlDialect.DEFAULT;
+    case POSTGIS:
```
JiajunBernoulli, line 226: Is `POSTGIS` before `POSTGRESQL`?
> I is before R

core/src/main/java/org/apache/calcite/sql/dialect/PostgisSqlDialect.java
YiwenWu: It would be clearer to have an independent Dialect, but I am worried that it will introduce a maintenance burden. 
At present, PostGIS does not have many unique functions.

macroguo-ghy:
> @YiwenWu @JiajunBernoulli What would be the best way to test these changes against a postgis database? Ideally, I'd like to add an integration test that executes queries against postgis (e.g. with testcontainers). But I havn't been able to find such tests in calcite.

I remember that testContainer has been used in redis apapter test.

core/src/main/java/org/apache/calcite/sql/dialect/PostgisSqlDialect.java
bchapuis: ST_UNION is the main case I had in mind that required to introduce a new function name (ST_UNARYUNION) in calcite. However, some other functions, such as ST_DUMP, may require a similar treatment. I guess that the list will grow as the dialect gets more systematically tested against postgis.

JiajunBernoulli:
> @YiwenWu @JiajunBernoulli What would be the best way to test these changes against a postgis database? Ideally, I'd like to add an integration test that executes queries against postgis (e.g. with testcontainers). But I havn't been able to find such tests in calcite.

There are some integration tests for Druid: https://github.com/zabetak/calcite-druid-dataset
Here is CI config in calcite: https://github.com/apache/calcite/blob/2aabf210dc1918c6ca20e63b39661ff445535eb8/.github/workflows/main.yml#L440

bchapuis:
Thanks a lot for the pointer, I will adopt the same approach and implement some integration tests in a third-party repository for now.

core/src/main/java/org/apache/calcite/sql/dialect/PostgisSqlDialect.java
julianhyde: Simplest thing seems to be to make Postgis dialect a subclass of Postgres dialect. There are many similarities, but also many differences.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
```diff
@@ -433,6 +433,12 @@ private static RelDataType sqlType(RelDataTypeFactory typeFactory, int dataType,
                 typeFactory.createSqlType(SqlTypeName.ANY), true);
       }
       return typeFactory.createArrayType(component, -1);
+    case OTHER:
```
julianhyde, line 436: On most databases the type string will be upper-case, "GEOMETRY".

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcToEnumerableConverter.java
```diff
@@ -283,6 +284,15 @@ private static void generateGet(EnumerableRelImplementor implementor,
               java.sql.Array.class);
       source = Expressions.call(BuiltInMethod.JDBC_ARRAY_TO_LIST.method, x);
       break;
+    case GEOMETRY:
+      source =
+          Expressions.call(
```
julianhyde, line 289: Add methods to BuiltInMethod. We don't want reflection errors at runtime.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1488,6 +1505,7 @@ public static SqlNode toSql(RexLiteral literal) {
       default:
         break;
       }
+
```
julianhyde, line 1508: remove blank line

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
```diff
@@ -39,9 +39,11 @@
 import com.google.common.collect.Interners;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.locationtech.jts.geom.Geometry;
```
julianhyde, line 42: Do we really want this dependency here?

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
```diff
@@ -1382,6 +1384,7 @@ public enum DatabaseProduct {
     LUCIDDB("LucidDB", "\"", NullCollation.HIGH),
     INTERBASE("Interbase", null, NullCollation.HIGH),
     PHOENIX("Phoenix", "\"", NullCollation.HIGH),
+    POSTGIS("Postgis", "\"", NullCollation.HIGH),
```
julianhyde, line 1387: Usual capitalization is PostGIS, I believe

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -1140,27 +1140,447 @@ public enum SqlKind {
...
+  ST_ZMAX,
+
+  /** The {@code ST_ZMin} function. */
+  ST_ZMIN,
 
   /** The {@code Hilbert} function that converts (x, y) to a position on a
    * Hilbert space-filling curve. */
   HILBERT,
 
+  ST_MAKE_LINE,
```
julianhyde, line 1581: Are these obsolete or not?

core/src/test/java/org/apache/calcite/sql/test/SqlTypeNameTest.java
```diff
@@ -172,7 +173,7 @@ class SqlTypeNameTest {
   @Test void testOther() {
     SqlTypeName tn =
         SqlTypeName.getNameForJdbcType(Types.OTHER);
-    assertEquals(null, tn, "OTHER did not map to null");
+    assertEquals(OTHER, tn, "OTHER did not map to null");
```
julianhyde, line 176: huh?

core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
```diff
@@ -264,6 +264,8 @@ public enum SqlTypeName {
           .put(Types.DISTINCT, DISTINCT)
           .put(Types.STRUCT, STRUCTURED)
           .put(Types.ARRAY, ARRAY)
+          .put(Types.OTHER, OTHER)
+
```
julianhyde, line 268: remove blank line

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -1140,27 +1140,447 @@ public enum SqlKind {
...
+  ST_TRANSFORM,
+
+  /** The {@code ST_Translate} function. */
+  ST_TRANSLATE,
+
+  /** The {@code ST_UnaryUnion} function. */
+  ST_UNARYUNION,
+
+  /** The {@code ST_UnaryUnion} function. */
+  ST_UNION,
```
julianhyde, line 1545: javadoc should be ST_Union

core/src/main/java/org/apache/calcite/sql/dialect/PostgisSqlDialect.java
```diff
@@ -0,0 +1,53 @@
...
+    if (call.getOperator().getName().equals("ST_UNARYUNION")) {
+      RelToSqlConverterUtil.specialOperatorByName("ST_UNION")
```
julianhyde, line 47: use SqlKind rather than name?

core/src/main/java/org/apache/calcite/sql/dialect/PostgisSqlDialect.java
```diff
@@ -0,0 +1,53 @@
...
+  public static final SqlDialect.Context DEFAULT_CONTEXT = SqlDialect.EMPTY_CONTEXT
+      .withDatabaseProduct(DatabaseProduct.POSTGIS)
+      .withConformance(SqlConformanceEnum.LENIENT)
+      .withIdentifierQuoteString("\"")
+      .withUnquotedCasing(Casing.TO_LOWER)
+      .withDataTypeSystem(POSTGRESQL_TYPE_SYSTEM);
+
+  public static final SqlDialect DEFAULT = new PostgisSqlDialect(DEFAULT_CONTEXT);
+
+  /** Creates a PostgresqlSqlDialect. */
```
julianhyde, line 40: a few places you wrote 'Postgresql' where you meant 'PostGIS' or 'Postgis'.

core/src/main/java/org/apache/calcite/sql/dialect/PostgisSqlDialect.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.calcite.avatica.util.Casing;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.validate.SqlConformanceEnum;
+import org.apache.calcite.util.RelToSqlConverterUtil;
+
+/**
+ * A <code>SqlDialect</code> implementation for the PostgreSQL database.
+ */
```
julianhyde, line 28: explain why it is a subclass of PostgresSQLDialect

core/src/main/java/org/apache/calcite/sql/dialect/PostgisGeometryDecoder.java
```diff
@@ -0,0 +1,191 @@
...
+import org.locationtech.jts.geom.Point;
+import org.locationtech.jts.geom.Polygon;
+
+import java.lang.reflect.Array;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.Arrays;
+
+/**
+ * A decoder for the PostGIS geometry format.
```
julianhyde, line 40: This class seems like a utility class, not appropriate for the org.apache.calcite.sql.dialect package. I didn't see any unit tests, but it could use a few.

julianhyde (commented):
Can you change the summary to "In JDBC adapter, add a PostGIS dialect so that ST_ functions can be pushed down". It will make the release notes easier to understand.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -312,7 +312,9 @@ private Expression getConvertExpression(
       case CHAR:
       case VARCHAR:
         return Expressions.call(BuiltInMethod.ST_GEOM_FROM_EWKT.method, operand);
-
+      case BINARY:
+      case VARBINARY:
+        return Expressions.call(BuiltInMethod.ST_GEOM_FROM_EWKB.method, operand);
```
jnh5y, line 317: As a future consideration, there are other binary formats like Tiny WKB.  https://github.com/TWKB/Specification/blob/master/twkb.md

I suppose that may be handled by a specific cast.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
bchapuis: Yes, the use of OTHER with a "geometry" type name seems specific to the postgresql jdbc driver and postgis. I guess they did this to distinguish postgis geometry types from the regular geometry types implemented in postgresql. https://www.postgresql.org/docs/current/datatype-geometric.html

I will clarify this with a comment and try to investigate regular geometry data types a bit more.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
bchapuis: I'm trying to wrap my head around this. From what I understand, calling a method of RelDataTypeFactoryImpl without JTS would trigger a class not found exception. As the use of JTS is currently circumscribed to a few spatial type classes, calcite works just fine without adding a dependency to JTS. Is that correct?

Given the popularity of JTS in the geospatial world, I wouldn't try to abstract it completely in JTS. An option could be to use reflection to dynamically load the Geometry class and provide some sort of fallback if the class is not in the class path.

```java
public static class Geometry { }

static Class<?> loadGeometryClass() {
    try {
        return Class.forName("org.locationtech.jts.geom.Geometry");
    } catch (ClassNotFoundException e) {
        return Geometry.class;
    }
}

private static final Map<Class, RelDataTypeFamily> CLASS_FAMILIES =
  ImmutableMap.<Class, RelDataTypeFamily>builder()
      .put(String.class, SqlTypeFamily.CHARACTER)
      // ...
      .put(loadGeometryClass(), SqlTypeFamily.GEO)
      .build();
```

What do you think?

core/src/main/java/org/apache/calcite/sql/dialect/PostgisGeometryDecoder.java
bchapuis: ok, I will place it in org.apache.calcite.util and add some unit tests.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
bchapuis: Thanks a lot for the pointer, it would definitely make sense to implement these functions:
https://postgis.net/docs/ST_AsTWKB.html

core/src/main/java/org/apache/calcite/sql/SqlKind.java
bchapuis: I believe we can remove them. ST_MAKE_LINE is ST_MAKELINE and all results returned by search engines for ST_POINT3 are related to ST_POINT.

core/src/test/java/org/apache/calcite/sql/test/SqlTypeNameTest.java
bchapuis: I remember modifying this line and found the test cases unclear (and I forgot to adapt the error message). The javadoc of getNameForJdbcType states: "corresponding SqlTypeName, or null if the type is not known". As other is kind of a known type, does it makes sense to adapt the test accordingly? I will commit this fix for now.

```java
  @Test void testOther() {
    SqlTypeName tn =
        SqlTypeName.getNameForJdbcType(Types.OTHER);
    assertEquals(OTHER, tn, "OTHER did not map to OTHER");
  }
```

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
julianhyde: Can we discuss in jira? Increasing the coupling to JTS is an architectural change.

bchapuis: Yes, here is the jira issue I just created:
https://issues.apache.org/jira/browse/CALCITE-6263

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
bchapuis: I logged a jira issue for the introduction of this functions:
https://issues.apache.org/jira/browse/CALCITE-6264

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
```diff
@@ -433,6 +433,17 @@ private static RelDataType sqlType(RelDataTypeFactory typeFactory, int dataType,
                 typeFactory.createSqlType(SqlTypeName.ANY), true);
       }
       return typeFactory.createArrayType(component, -1);
+    case GEOMETRY:
```
YiwenWu, line 436: According to the elements in `SqlTypeName#JDBC_TYPE_TO_NAME` , whether there are any cases that can be matched `GEOMETRY`?

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1445,6 +1447,21 @@ public static SqlNode toSql(RexLiteral literal) {
       // Create a string without specifying a charset
       return SqlLiteral.createCharString((String) castNonNull(literal.getValue2()), POS);
     }
+    case GEO: {
+      Geometry geom = castNonNull(literal.getValueAs(Geometry.class));
+      switch (typeName) {
```
YiwenWu, line 1452: I am not sure that typeName `CHAR` or `VARCHAR` can be obtained under `GEO`  family.

mihaibudiu:
Will this be ready for 1.37?

bchapuis:
No, I hadn't enought time to work on this and the decoupling of the GEOMETRY type from the JTS Geometry class will require additional efforts.

https://issues.apache.org/jira/browse/CALCITE-6239

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
bchapuis: I believe this is necessary when geometries are represented as string in the query with WKT.

core/src/main/java/org/apache/calcite/sql/dialect/PostgisSqlDialect.java
bchapuis: Thanks, I fixed it.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
bchapuis: I'm not sure about this. There is no type name for geometry in java.sql.Types, but there is an `ExtraSqlTypes.GEOMETRY = 2015` defined in Calcite. I have now registered it in `SqlTypeName#JDBC_TYPE_TO_NAME`.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1511,6 +1513,21 @@ public static SqlNode toSql(RexLiteral literal) {
...
+      case CHAR:
+      case VARCHAR:
+        return SqlLiteral.createCharString(
+            SpatialTypeUtils.asEwkt(geom), POS);
+      case BINARY:
+      case VARBINARY:
+        return SqlLiteral.createBinaryString(
+            SpatialTypeUtils.asWkbArray(geom), POS);
+      default:
+        return SqlLiteral.createNull(POS);
```
mihaibudiu: why is this null?

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
```diff
@@ -65,6 +65,15 @@ public interface RelDataTypeFactory {
    */
   RelDataType createJavaType(Class clazz);
 
+  /**
+   * Creates a type that corresponds to a Java class, with a given family.
+   *
+   * @param clazz the Java class used to define the type
+   * @param family The family of the type, or null to infer
```
mihaibudiu, line 74: type families are a very confusing concept in Calcite, I think they should be deprecated.

mihaibudiu: but perhaps this PR is not the right place to have this conversation.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeAssignmentRule.java
```diff
@@ -193,6 +193,8 @@ private SqlTypeAssignmentRule(
     rule.add(SqlTypeName.GEOMETRY);
     rule.add(SqlTypeName.CHAR);
     rule.add(SqlTypeName.VARCHAR);
+    rule.add(SqlTypeName.BINARY);
```
mihaibudiu, line 248: these (and a few other similar changes) look like bug fixes that would be nice to have even before we can merge the full PR.

core/src/test/java/org/apache/calcite/util/PostgisGeometryDecoderTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+  @Test void decodeMultiPolygon() {
+    Geometry geometry = PostgisGeometryDecoder.decode(MULTIPOLYGON);
+    assertTrue(geometry instanceof org.locationtech.jts.geom.MultiPolygon);
+    assertEquals(
+        "MULTIPOLYGON (((0 0, 1 0, 1 1, 0 1, 0 0)), ((2 2, 3 2, 3 3, 2 3, 2 2)))",
+        new WKTWriter().write(geometry));
+  }
+
+  @Test void decodeGeometryCollection() {
```
mihaibudiu, line 167: how about a few negative tests, where the decoding cannot happen?

mihaibudiu (approved):
This looks pretty nice to me; the PR is relatively small given how much it accomplishes.
But I am not familiar with this subject.
One thing I wonder about: if other dialects want to provide geography support, what do they need to do? We don't really want only Postgis to do that.

core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java
```diff
@@ -440,4 +442,17 @@ private static class RecordFieldImpl implements Types.RecordField {
...
+  }
+
+  @Override public RelDataType createJavaType(
+      Class clazz,
+      @Nullable RelDataTypeFamily family) {
+    if (Geometry.class.isAssignableFrom(clazz)) {
+      return canonize(new JavaType(clazz, SqlTypeFamily.GEOMETRY));
+    }
+    return super.createJavaType(clazz, family);
+  }
```
zabetak, line 469: Why should we do this change here and not in `RelDataTypeFactoryImpl`?

core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
```diff
@@ -837,7 +837,7 @@ private static ColumnMetaData.AvaticaType avaticaType(JavaTypeFactory typeFactor
         }
         return ColumnMetaData.struct(columns);
       case ExtraSqlTypes.GEOMETRY:
-        typeOrdinal = Types.VARCHAR;
+        typeOrdinal = Types.OTHER;
```
zabetak, line 840: breaking change?

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
```diff
@@ -65,6 +65,15 @@ public interface RelDataTypeFactory {
...
+   * @return canonical Java type descriptor
+   */
+  RelDataType createJavaType(Class clazz, @Nullable RelDataTypeFamily family);
```
zabetak, line 77: This is a public API so adding a new method in the interface can break many consumers. Since family is nullable I guess we can add a default implementation that delegates to `createJavaType(Class)` to avoid that.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
```diff
@@ -644,15 +658,27 @@ public class JavaType extends RelDataTypeImpl {
     private final boolean nullable;
     private final @Nullable SqlCollation collation;
     private final @Nullable Charset charset;
+    private final @Nullable RelDataTypeFamily family;
```
zabetak, line 657: Should this really be nullable? I get the impression that we are always gonna set a family either explicit or implicit.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
```diff
@@ -680,6 +717,9 @@ public Class getJavaClass() {
     }
 
     @Override public RelDataTypeFamily getFamily() {
+      if (this.family != null) {
+        return this.family;
+      }
       RelDataTypeFamily family = CLASS_FAMILIES.get(clazz);
       return family != null ? family : this;
```
zabetak, line 720: It seems that after the changes in this PR we always gonna set a family during construction so maybe the method could be simplified to just `return this.family`.

core/src/main/java/org/apache/calcite/runtime/SpatialTypeFunctions.java
```diff
@@ -22,6 +22,7 @@
 import org.apache.calcite.linq4j.function.Deterministic;
 import org.apache.calcite.linq4j.function.Experimental;
 import org.apache.calcite.linq4j.function.Hints;
+import org.apache.calcite.linq4j.function.Parameter;
```
zabetak, line 25: Why do we need this annotation now?

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -1242,22 +1242,439 @@ public enum SqlKind {
...
+  //  Method[] methods = SpatialTypeFunctions .class.getDeclaredMethods();
+  //  Arrays.stream(methods)
+  //    .map(java.lang.reflect.Method::getName)
+  //    .filter(method -> method.startsWith("ST_"))
+  //    .distinct().sorted()
+  //    .forEach(method -> {
+  //      System.out.println("/** The {@code " + method + "} function. */");
+  //      System.out.print(method.toUpperCase());
+  //      System.out.print(",\n\n");
+  //    });
```
zabetak: Not sure why need to put this codeblock here. If we want to ensure that SqlKind entries are generated in a certain way we should better put this code in a unit test. 

Other than that, I am not sure why we opted to add all these use-case specific functions in this enum. Where do we use them? Where are they tested?

core/src/main/java/org/apache/calcite/sql/dialect/PostgisSqlDialect.java
```diff
@@ -0,0 +1,58 @@
...
+
+  /** Creates a PostgisSqlDialect. */
+  public PostgisSqlDialect(Context context) {
+    super(context);
+  }
+
+  @Override public void unparseCall(SqlWriter writer, SqlCall call, int leftPrec, int rightPrec) {
+    if (call.getOperator().getName().equals(SqlKind.ST_UNARYUNION.name())) {
+      RelToSqlConverterUtil.specialOperatorByName(SqlKind.ST_UNION.name())
+          .unparse(writer, call, 0, 0);
```
zabetak, line 53: Is this unparse specificity Postgres specific or operator specific? I am just trying to understand why we didn't opt to override the unparse method of the respective operator and we opted to do it at the dialect level.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeAssignmentRule.java
zabetak: I kind of agree with Mihai. Some changes in this PR such as those around binary types seem independent to to PostGIS. Would it be possible to isolate them in separate JIRA/PRs?

core/src/main/java/org/apache/calcite/sql/type/SqlTypeFamily.java
```diff
@@ -76,7 +76,7 @@ public enum SqlTypeFamily implements RelDataTypeFamily {
   ANY,
   CURSOR,
   COLUMN_LIST,
-  GEO,
+  GEOMETRY,
```
zabetak: breaking change?

core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
zabetak: I still see the blank lines here.

core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
```diff
@@ -905,13 +905,20 @@ boolean canImplicitTypeCast(List<RelDataType> types, List<SqlTypeFamily> familie
...
+    // BINARY -> GEOMETRY
+    if (SqlTypeUtil.isBinary(in) && expected == SqlTypeFamily.GEOMETRY) {
+      return expected.getDefaultConcreteType(factory);
+    }
+
     if ((SqlTypeUtil.isCharacter(in) || SqlTypeUtil.isBinary(in))
         && expected == SqlTypeFamily.UUID) {
       return expected.getDefaultConcreteType(factory);
     }
+
```
zabetak: (nit) Remove blank line. The intention is to minimize diff and keep history clean by avoiding unecessary changes.

linq4j/src/main/java/org/apache/calcite/linq4j/function/Parameter.java
```diff
@@ -81,4 +82,10 @@
    * is NULL, and therefore optional parameters must be nullable.
    */
   boolean optional() default false;
+
+  /** Returns the SQL type code.
+   *
+   * <p>Values are typically from {@link java.sql.Types}, for example
+   * {@link java.sql.Types#INTEGER}. */
+  int sqlType() default Types.OTHER;
```
zabetak, line 90: This is another improvement that could potentially stand on its own independent of PostGIS. Also I would be curious to understand why we need this now and we didn't before?

testkit/src/main/java/org/apache/calcite/test/CalciteAssert.java
```diff
@@ -888,6 +889,7 @@ static SchemaPlus addSchema_(SchemaPlus rootSchema, SchemaSpec schema) {
     case CLONE_FOODMART:
       foodmart = addSchemaIfNotExists(rootSchema, SchemaSpec.JDBC_FOODMART);
       return rootSchema.add("foodmart2", new CloneSchema(foodmart));
+
```
zabetak: nit: Irrelevant diff

testkit/src/main/java/org/apache/calcite/test/CalciteAssert.java
```diff
@@ -917,7 +920,6 @@ static SchemaPlus addSchema_(SchemaPlus rootSchema, SchemaSpec schema) {
           ViewTable.viewMacro(rootSchema, sql2,
               ImmutableList.of("GEO"), emptyPath, false);
       s.add("states", viewMacro2);
-
```
zabetak, line 937: nit: Irrelevant diff

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -116,6 +117,7 @@
 
 import org.checkerframework.checker.initialization.qual.UnknownInitialization;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.locationtech.jts.geom.Geometry;
```
zabetak, line 123: It seems that now the RelToSql conversion becomes coupled with `org.locationtech` library. I am wondering if there is a way to avoid it.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
zabetak: Indeed it is not clear why we return null here. It doesn't feel right.

core/src/test/java/org/apache/calcite/util/PostgisGeometryDecoderTest.java
```diff
@@ -0,0 +1,124 @@
...
+      + "0084000000000000008400000000000000040000000000000084000000000000000400"
+      + "000000000000040";
+
+  private static final String GEOMETRYCOLLECTION = "0107000020E6100000020000000"
+      + "1010000000000000000000000000000000000000001020000000200000000000000000"
+      + "0F03F000000000000F03F00000000000000400000000000000040";
+
+  @Test void decodeNull() {
+    Geometry geometry = PostgisGeometryDecoder.decode((String) null);
+    assertEquals(null, geometry);
```
zabetak: Use `assertNull` ?

core/src/test/java/org/apache/calcite/util/PostgisGeometryDecoderTest.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  @Test void decodePoint() {
+    Geometry geometry = PostgisGeometryDecoder.decode(POINT);
+    assertTrue(geometry instanceof Point);
+    assertEquals("POINT (1 2)", new WKTWriter().write(geometry));
```
zabetak: It seems that we don't need both assertions since if equality succeeds it implies that its a point. SImilar for the other test cases.

core/src/test/java/org/apache/calcite/util/PostgisGeometryDecoderTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.locationtech.jts.io.WKTWriter;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests for the {@link PostgisGeometryDecoder} class. These tests are based
+ * on the values retrieved from PostGIS using the JDBC driver for PostgreSQL.
+ */
+class PostgisGeometryDecoderTest {
```
zabetak, line 38: It seems that this class is not only about the `PostgisGeometryDecoder` but depends also on `WKTWriter`. I am wondering if it makes sense to test the decoder independently by explicitly creating the expected Point, Linestring, etc., object and rely on equality comparison between objects.

core/src/main/java/org/apache/calcite/config/Lex.java
```diff
@@ -63,6 +63,13 @@ public enum Lex {
   MYSQL_ANSI(Quoting.DOUBLE_QUOTE, Casing.UNCHANGED, Casing.UNCHANGED, false,
       CharLiteralStyle.STANDARD),
 
+  /** Lexical policy similar to PostgreSQL. Quoted identifiers are preserved;
+   * Unquoted identifiers are converted to lower-case; after which, identifiers
+   * are matched case-sensitively.
+   */
+  POSTGRESQL(Quoting.DOUBLE_QUOTE, Casing.TO_LOWER, Casing.UNCHANGED, false,
```
zabetak, line 70: Where do we use this? How is it tested?

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcToEnumerableConverter.java
```diff
@@ -323,6 +323,14 @@ private static void generateGet(EnumerableRelImplementor implementor,
               java.sql.Array.class);
       source = Expressions.call(BuiltInMethod.JDBC_ARRAY_TO_LIST.method, x);
       break;
+    case GEOMETRY:
```
zabetak, line 326: Are there tests that cover this path?

zabetak (changes requested):
Thanks for advancing this work @bchapuis !

My main concerns are around breaking changes that could potentially be avoided mostly around the type factories and the additional coupling with `org.locationtech` that appears in some places such as the `SqlImplementor`.

Most of the remaining comments/questions are rather minor.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
zabetak: Additionally, it seems that the new code in this class is not covered by existing tests.

core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java
bchapuis: @zabetak This relates to my interpretation of the architectural comments made in [CALCITE-6263](https://issues.apache.org/jira/browse/CALCITE-6263): `RelDataTypeFactoryImpl` is a core class, and the coupling with JTS should preferably occur in a child class. `JavaTypeFactoryImpl` looked to me like a good place to check if a class is a JTS Geometry.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
bchapuis: This modification was suggested in the following commit and seemed necessary decouple `RelDataTypeFactoryImpl` from JTS:

https://github.com/apache/calcite/pull/3668/commits/da8494213c3810584a75064d5fadf52ac1f95d12

> A related change is to add a 'family' parameter to
the `TypeFactory.createJavaType` method. It is nullable, and
if null, the type becomes its own family. In JavaType, we
should add a `family` field. 

> When this change is complete we should be able to remove JTS
Geometry from RelDataTypeFactoryImpl.CLASS_FAMILIES.

core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
bchapuis: Yes, this is probably a breaking change, but I don't think VARCHAR is appropriate here.

I think GEOMETRY is clearly database specific as indicated here:
https://docs.oracle.com/en/java/javase/11/docs/api/java.sql/java/sql/Types.html#OTHER

This choice also correspond to what is implemented in the postgresql jdbc driver.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcToEnumerableConverter.java
bchapuis: I will look into this. I believe it is covered by the integration tests I implemented with testcontainers in a separate repository.

core/src/main/java/org/apache/calcite/config/Lex.java
bchapuis: I will look into this. I believe it is covered by the integration tests I implemented with testcontainers in a separate repository. I'm not sure how to test it without starting a postgis container.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
bchapuis: sorry, I wasn't sure what to return in that case. Would the following be better?

```java
SqlLiteral.createUnknown(POS);
```

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
bchapuis: I'm not sure how to do it as well. We can probably try to abstract JTS, but I'm not sure the effort is worth it as JTS almost became a defacto standard in Java. Additionaly,  I think this coupling is internal as the WKT strings and WKB binaries outputed here can be read by other libraries if needed.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
bchapuis: The JavaType(Class clazz) constructor is called in many places, and I was worried to break things by specifying a RelDataTypeFamily everywhere. Is this something I should do?

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
bchapuis: Same question as before, isn't it a risky breaking change.

core/src/main/java/org/apache/calcite/sql/dialect/PostgisSqlDialect.java
bchapuis: I think this is somewhat related to https://issues.apache.org/jira/browse/CALCITE-5280

core/src/main/java/org/apache/calcite/sql/type/SqlTypeFamily.java
bchapuis: Maybe, I can revert it.

linq4j/src/main/java/org/apache/calcite/linq4j/function/Parameter.java
bchapuis: This was introduced in the following commit:
https://github.com/apache/calcite/pull/3668/commits/da8494213c3810584a75064d5fadf52ac1f95d12

> Add annotation Parameter.sqlType
> The annotation allows you to define the SQL type of a
parameter, for cases where it cannot be deduced automatically
from the Java type. This is especially useful for GEOMETRY
values, for which Calcite has no built-in Java type. (JTS
Geometry is widely used in Calcite, but it is an
implementation, and we do not want it to become the
specification.)

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
zabetak: Still you can add a default implementation to avoid the breakage.

core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java
zabetak: I don't see much benefit between doing the change here or in `RelDataTypeFactoryImpl` since both are widely used. My understanding from CALCITE-6263 is that we would have our own `org.apache.calcite.xxx.Geometry` interface and this is what would appear in `RelDataTypeFactoryImpl`. I haven't followed every discussion so maybe my understanding is not correct. 

Another side concern with this implementation is that the caller is supposed to pass explicitly the family to use but it seems that for the `Geometry` instances it is not possible. This behavior seems to be inconsistent with the API specification.

core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
zabetak: Sounds good but in general breaking changes should be mentioned in history.md unless they are clearly bug fixes.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
zabetak: From my point of view this also increases the coupling of Calcite with JTS and my understanding from CALCITE-6263 is that we were trying to avoid that. `SqlImplementor` is not as important as the `RelDataTypeFactory` but still it is a class that is used in many projects and so far it didn't have any non `org.apache.calcite` import.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcToEnumerableConverter.java
zabetak: We have usages of testcontainers in Calcite. Is there something that  prevents the integration tests to be committed as part of this PR?

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
zabetak: From modifying the constructors I got the impression that we were always passing a family. It seems that it is not true so feel free to ignore this comment.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
zabetak: You are right, let's avoid that especially in the content of this PR.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
bchapuis: I used createUnknown for now.

core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java
bchapuis: I abandoned the idea of introducing the wrappers described in CALCITE-6263 after the addition of the @SqlType annotation by @jhyde in [7124c3a](https://github.com/apache/calcite/commit/7124c3ae3e5001b85dc6617554c170be49476f75), as it allows Java types to explicitly declare their corresponding SQL type.

I’m still trying to wrap my head around the best solution 😅, but since spatial support is still evolving in Calcite (and JTS is almost a de facto standard for Java) moving this to `JavaTypeFactoryImpl` felt like a practical fit for now.

Let me know if you’d like to discuss broader architectural directions. I’m happy to revisit it (ideally in another PR), but I hope this strikes a reasonable balance in the meantime.

core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
bchapuis: I checked the history.md file. Is it modified by the release manager? I'm tempted to say that this change is closer to a bugfix and has a relatively low impact as it matches what is done in other databases and drivers.

core/src/main/java/org/apache/calcite/config/Lex.java
bchapuis: I added a test cases for this in `LexCaseSensitiveTest` and in `LexEscapeTest`.

My integration tests are located here: https://github.com/bchapuis/calcite-postgis-tests

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcToEnumerableConverter.java
bchapuis: My integration tests are located here: https://github.com/bchapuis/calcite-postgis-tests

I looked into how Calcite uses Testcontainers, but it seems the support is currently limited to a few Redis tests. The JDBC tests don’t appear to leverage Testcontainers yet, which is why I set up external integration tests. Did I overlook anything?

core/src/main/java/org/apache/calcite/sql/SqlKind.java
bchapuis: I used this as a convenience to generate the list of function but have now removed it.

core/src/main/java/org/apache/calcite/runtime/SpatialTypeFunctions.java
bchapuis: This is related to the approach introduced here:
https://github.com/apache/calcite/commit/7124c3ae3e5001b85dc6617554c170be49476f75

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
bchapuis: Would this imply adding a wrapper in Calcite to reduce coupling to JTS? I’m still not convinced that the added complexity is worth it. JTS is almost a standard for spatial functions and operators. GEOS, which is used in Postgis, is actually a C++ port of JTS, and their communities overlap significantly. To put it another way,  I think JTS is almost to geometry what Guava is to immutable maps and lists ;)

bchapuis:
@zabetak Thanks a lot for your review. I think most of the comments have been addressed. Please let me know if you want to further discuss de decoupling of JTS.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeAssignmentRule.java
bchapuis: I think some of these bug fixes have been addressed separately by Lucas Brenner in CALCITE-7073. I had a few merge conflicts when rebasing the current branch.

I'm not completely sure what else could be extracted into smaller PR/bug fixes and would prefer to squash merge this PR to move on ;)

At this point I'm not completely sure what else could be extracted into smaller PR/bug fixes. To keep things moving, I’d prefer to squash-merge, unless anyone has other suggestions.

rubenada:
@zabetak do you think you'd have a bit of time to take a look at the latest changes after your remarks please?

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
rubenada: @bchapuis I cannot see the default implementation suggested by @zabetak to avoid breaking consumers.
IIANM the proposal to maintain compatibility was:
```
RelDataType createJavaType(Class clazz);

default RelDataType createJavaType(Class clazz, @Nullable RelDataTypeFamily family) {
  return createJavaType(clazz);
}
```

core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java
rubenada: @bchapuis usually in cases like this the PR modifies the commented out part at the top of history.md corresponding to the next release, adding a line on the "Breaking Changes" section so that, when preparing the release, the RM will not forget about mentioning this breaking change.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
bchapuis: Maybe I missundersood the originial comment. However, since my PR introduced a new createJavaType method with two arguments, we wanted to keep the existing single-argument createJavaType method to avoid breaking consumers.

If I recall correctly, my fix (on 30.07.2025) involved adding a default implementation for the single-argument version, as shown below:

```java
  default RelDataType createJavaType(Class clazz) {
    return createJavaType(clazz, null);
  }

  RelDataType createJavaType(Class clazz, @Nullable RelDataTypeFamily family);
  ```

File: core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java (hunk 24 above)
Comment (bchapuis) [reply core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java:469]: @rubenada @zabetak Let me revive this thread, as I believe it’s important to decide whether to merge the PR or give it another round of modifications. Please note that I don’t want to block the current release cycle, but I’d really appreciate some hints on how to implement this change.

File: core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java (hunk 25 above)
Comment (bchapuis) [reply core/src/main/java/org/apache/calcite/prepare/CalcitePrepareImpl.java:840]: Thanks for the clarification, I added a note in history.md.

File: core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java (hunk 26 above)
Comment (rubenada) [reply core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java:77]: My understanding of @zabetak 's remark was the other way around.
IIANM your proposal would mean that all consumers implementing this (widely used) interface will break upon upgrade, since they'd need to implement the new method `createJavaType(Class clazz, @Nullable RelDataTypeFamily family)` ; whereas having a default implementation for it delegating to `createJavaType(Class clazz)` (as I, and I think @zabetak, proposed) would mean that these consumers shall continue working as before upon upgrade.

File: core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java (hunk 24 above)
Comment (rubenada) [reply core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java:469]: I appreciate your work @bchapuis .

Regarding `JavaTypeFactoryImpl` vs `RelDataTypeFactoryImpl` discussion, a small note that may be worth considering is that  `JavaTypeFactoryImpl` is marked as "experimental and subject to change/removal without notice", so it might seem a better fit for experimental things that we may need to revisits in the future.

I admit that I haven't followed all the discussions involving this issue, but I can see there are still some valid open points, like RelToSql becoming coupled with `org.locationtech`, test coverage for the new conversion paths, the default implementation of the new method in `RelDataTypeFactory`, and where to include our actual GEO implementation (`JavaTypeFactoryImpl` vs `RelDataTypeFactoryImpl`).
Also there's this comment by @zabetak : 
> Another side concern with this implementation is that the caller is supposed to pass explicitly the family to use but it seems that for the Geometry instances it is not possible. This behavior seems to be inconsistent with the API specification.

I think we can try our best to not let this PR stale and finalize it asap, but probably the timing would be better for 1.42.

File: core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java (hunk 26 above)
Comment (bchapuis) [reply core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java:77]: This change comes from a suggestion made by @julianhyde in https://github.com/apache/calcite/commit/da8494213c3810584a75064d5fadf52ac1f95d12

> Add annotation Parameter.sqlType
The annotation allows you to define the SQL type of a
parameter, for cases where it cannot be deduced automatically
from the Java type. This is especially useful for GEOMETRY
values, for which Calcite has no built-in Java type. (JTS
Geometry is widely used in Calcite, but it is an
implementation, and we do not want it to become the
specification.)
>
> This commit is incomplete, but illustrates the general idea.
> 
> A related change is to add a 'family' parameter to
the `TypeFactory.createJavaType` method. It is nullable, and
if null, the type becomes its own family. In JavaType, we
should add a `family` field.
> 
> When this change is complete we should be able to remove JTS
Geometry from RelDataTypeFactoryImpl.CLASS_FAMILIES.

In my understanding this proposition was an API change (possibly major) that would allow to define the SQL type of a parameter more explicitly. As the family parameter is nullable, it effectively allows the cosumers to continue working as before. I acknoledge that I probably don't get all the possible ramifications of this 😅

File: core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java (hunk 24 above)
Comment (bchapuis) [reply core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java:469]: I understand the concerns regarding the coupling to JTS. There is no problem in posponing this for 1.42. I suggested to include it in this release to unstall the discussion and try to move things forward. It would be great to improve geospatial support in calcite.

Comment (zabetak) [reply core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java:469]: I will leave out the discussion about the general coupling with JTS and will try to focus on more pragmatic stuff. 

There is a new API in RelDataTypeFactory and this interface has various implementations both in Calcite code base and elsewhere. The change here implies that only the `JavaTypeFactoryImpl` is able to handle properly GEOMETRY types. If we accept that Geometry types are now part of the core type system shouldn't the other implementations handle them as well? 

If we leave the change only here does it mean that all support for PostGIS dialect is possible only if we are using the `JavaTypeFactoryImpl`?

File: core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java (hunk 26 above)
Comment (zabetak) [reply core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java:77]: My suggestion to avoid breakage is the following:
```java
RelDataType createJavaType(Class clazz);

default RelDataType createJavaType(Class clazz, @Nullable RelDataTypeFamily family) {
  return createJavaType(clazz);
}
```
Moreover, I don't fully understand what the new API is doing so I would suggest to add unit tests for the most common implementations (`SqlTypeFactoryImpl` and `JavaTypeFactoryImpl`) especially covering cases where we use `Geometry` types that is the focus of the PR.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
zabetak: Still relevant

core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
```diff
@@ -287,6 +287,9 @@ public enum SqlTypeName {
           .put(Types.DISTINCT, DISTINCT)
           .put(Types.STRUCT, STRUCTURED)
           .put(Types.ARRAY, ARRAY)
+          .put(Types.OTHER, OTHER)
```
zabetak, line 290: Why does OTHER come into the picture now?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -198,6 +199,7 @@ private static Map<SqlDialect, DatabaseProduct> dialects() {
         .put(DatabaseProduct.MYSQL.getDialect(), DatabaseProduct.MYSQL)
         .put(mySqlDialect(NullCollation.HIGH), DatabaseProduct.MYSQL)
         .put(DatabaseProduct.ORACLE.getDialect(), DatabaseProduct.ORACLE)
+        .put(DatabaseProduct.POSTGIS.getDialect(), DatabaseProduct.POSTGIS)
```
zabetak, line 202: How about testing the rest of POSTGIS functions here? Can't we have a single test that builds different queries by iterating through all supported functions?

core/src/test/java/org/apache/calcite/sql/test/SqlTypeNameTest.java
```diff
@@ -202,7 +203,7 @@ class SqlTypeNameTest {
   @Test void testOther() {
     SqlTypeName tn =
         SqlTypeName.getNameForJdbcType(Types.OTHER);
-    assertThat("OTHER did not map to null", tn, nullValue());
+    assertThat("OTHER did not map to OTHER", tn, is(OTHER));
```
zabetak, line 206: Why there was a test explicitly testing for null?

core/src/test/java/org/apache/calcite/sql/validate/LexCaseSensitiveTest.java
```diff
@@ -123,6 +123,18 @@ private static void runProjectQueryWithLex(Lex lex, String sql)
...
+        + "  select empid from emps order by EMPS.DEPTNO)";
+    runProjectQueryWithLex(Lex.POSTGRESQL, sql);
+  }
+
+  @Test void testCalciteCasePostgresqlNoException() throws Exception {
+    String sql = "select EMPID, empid from\n"
+        + " (select empid from emps order by emps.deptno)";
+    runProjectQueryWithLex(Lex.POSTGRESQL, sql);
+  }
+
```
zabetak, line 137: Tests do not seem relevant to PostGIS dialect support.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcToEnumerableConverter.java
zabetak: If the only dependency is Testcontainers (that we already use in Calcite) then you could bring in the tests in the calcite code base (assuming that there are no big data files involved). They don't necessarily need to be under the `JdbcToEnumerableConverter` class; you can create a completely new one especially for `PostGIS`.

zabetak (commented):
Did another quick pass and left some comments/clarifications. Currently, I am pretty busy with other stuff so it's kinda hard to be very reactive on this PR. Feel free to move this forward if others are OK with it and I don't respond reasonably fast.

asuriy:
Hi there, 
I am hoping to be able to integrate the postgis functions added in this PR in one of my projects in future, so I was wondering if this PR is planned to be a part of the next release? Thanks for your hard work!

PR #3671, opened 2024-02-05
Add .gitignore and .ratignore for jenv

macroguo-ghy:
When using jenv to manage the local java environment, it will generate a `.java-version` file, we should ignore it.

See more details about jenv in https://github.com/jenv/jenv.


JiajunBernoulli (approved):
LGTM

PR #3672, opened 2024-02-05
[CALCITE-6238] Exception while evaluating ROUND/TRUNCATE functions

mihaibudiu:
I had to modify a test in testJdbcFun which no longer passes in the original form.
I don't understand why it was passing beforehand. 
To the reviewers: please double check this.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
@@ -1774,11 +1774,11 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
           OperandTypes.NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
-  /** The {@code ROUND(numeric [, numeric])} function. */
+  /** The {@code ROUND(numeric [, integer])} function. */
   public static final SqlFunction ROUND =
       SqlBasicFunction.create("ROUND",
           ReturnTypes.ARG0_NULLABLE,
-          OperandTypes.NUMERIC_OPTIONAL_INTEGER,
+          OperandTypes.NUMERIC.or(OperandTypes.NUMERIC_INT32),
           SqlFunctionCategory.NUMERIC);
 
   /** The {@code SIGN(numeric)} function. */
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -393,6 +393,12 @@ public static SqlOperandTypeChecker variadic(
           // Second operand optional (operand index 0, 1)
           number -> number == 1);
 
+  public static final SqlOperandTypeChecker NUMERIC_INT32 =
+      sequence("ROUND(<NUMERIC>, <INTEGER>)",
+          family(SqlTypeFamily.NUMERIC),
+          // Only 32-bit integer allowed for the second argument
+          new TypeNameChecker(SqlTypeName.INTEGER));
```
macroguo-ghy, line 401: It's okay for me to use `TypeNameChecker`, but if Calcite implicitly casts other integer types to 32-bit integer, it would be better for this case.

mihaibudiu: These rules explain how ROUND is typechecked.
These rules do not say anything about implicit casts, only about legal arguments accepted by this function.
I believe that type coercion is the one that inserts the implicit casts if they are necessary.

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -1802,11 +1802,11 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
           OperandTypes.NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
-  /** The {@code TRUNCATE(numeric [, numeric])} function. */
+  /** The {@code TRUNCATE(numeric [, integer])} function. */
   public static final SqlBasicFunction TRUNCATE =
       SqlBasicFunction.create("TRUNCATE",
           ReturnTypes.ARG0_NULLABLE,
-          OperandTypes.NUMERIC_OPTIONAL_INTEGER,
+          OperandTypes.NUMERIC.or(OperandTypes.NUMERIC_INT32.apply("TRUNCATE")),
```
rubenada, line 1809: I understand that TRUNCATE suffers from the same issue as ROUND. Perhaps we should enlarge the Jira title (and therefore the commit message) to reflect this, e.g.  `[CALCITE-6238] Exception while evaluating ROUND and TRUNCATE functions when the second parameter is not an integer`.
Also we should add a new unit test with TRUNCATE, analogous to the new `round(42, CAST(2 as BIGINT))` test in `SqlOperatorTest.java`.
Wdyt?

mihaibudiu: I have renamed the JIRA case and the PR, I will fix the commit message when I squash.
I have added the suggested new unit test.

rubenada:
LGTM

mihaibudiu:
I have squashed and rebased, in case anyone wants to click the "approve" button. Thank you for the reviews.

PR #3675, opened 2024-02-07
[CALCITE-6251] innerEnumerator in EnumerableDefaults::correlateBatchJoin is not closed

kramerul:


rubenada:
@kramerul  CI is failing due to commit message:
```
<[invalid git log message '[CALCITE-6251] innerEnumerator in EnumerableDefaults::correlateBatchJoin is not closed'; Message must start with upper-case letter]>
```

I think the simplest would be to amend the commit message and start with uppercase: `[CALCITE-6251] InnerEnumerator in ...`

kramerul:
I had to fix also another issue with `@Nullable`

rubenada:
LGTM

linq4j/src/main/java/org/apache/calcite/linq4j/EnumerableDefaults.java
```diff
@@ -1799,11 +1801,16 @@ private boolean innerHasNext() {
             i = -1;
           }
 
-          @Override public void close() {
-            outerEnumerator.close();
+          private void closeInner() {
```
JiajunBernoulli, line 1804: Should we use `synchronized`?

rubenada: Interesting point. 
Maybe I'm wrong, but my understanding is that our enumerators are not thread-safe, so if a 3rd party uses them on a multi-threaded environment, they would need to handle synchronization on their end.

JiajunBernoulli: Get it

JiajunBernoulli (approved):

PR #3676, opened 2024-02-07
[CALCITE-6250] Limitations of MongoDB adapter are not documented

TheOezzi:
Add hints with restrictions for documents/filters to howto site

### DIFF
--- a/site/_docs/howto.md
+++ b/site/_docs/howto.md
@@ -329,7 +329,10 @@ See the [tutorial]({{ site.baseurl }}/docs/tutorial.html).
 First, download and install Calcite,
 and <a href="https://www.mongodb.org/downloads">install MongoDB</a>.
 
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):

mihaibudiu:
The rule is for the PR name, the JIRA issue, and the commit message to be the same.

mihaibudiu:
The titles match, but they do not match the guidelines, which require the title to describe a *problem*, and not the solution to the problem. I personally don't have a problem with the title, but I think that it's good to follow the rules set up by the project. A problem-like title would be "Limitations of MongoDB adapter are not documented"

JiajunBernoulli:
FYI: https://calcite.apache.org/develop/

TheOezzi:
I've adapted the commit and title according to your suggestion.

PR #3677, opened 2024-02-08
[CALCITE-2980] Implement the FORMAT clause of the CAST operator

Anthrino:


core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -684,7 +686,13 @@ protected RexNode convertCast(
     final boolean safe = kind == SqlKind.SAFE_CAST;
     final SqlNode left = call.operand(0);
     final SqlNode right = call.operand(1);
+    final SqlNode format = call.getOperandList().size() > 2 ?
+        call.operand(2) : null;
+
     final RexBuilder rexBuilder = cx.getRexBuilder();
+    final RexNode arg = cx.convertExpression(left);
+    final RexNode formatArg = isNull(format) ? null : cx.convertExpression(format);
```
julianhyde, line 694: Why use the `isNull` method, here and elsewhere? `isNull` detects the `NULL` literal, but I don't think that `CAST(d AS DATE FORMAT NULL)` is even allowed (correct me if I'm wrong).

The only null is a java null, where you pass a RexLiteral whose value is null if there was no `FORMAT` clause.

I am wondering whether you intend to support an expression for the format clause, e.g. `CAST(d AS DATE FORMAT 'yyyy-' || 'mm')` but as there are no tests I can't answer that question.

Anthrino: Thanks for checking this Julian, I intended to use a Null-valued RexLiteral in cases where the format clause is not specified with the CAST expression. I've updated the flow to populate/check the constant null literal where required.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -318,141 +320,19 @@ private Expression getConvertExpression(
...
-        return RexImpTable.optimize2(operand,
-            Expressions.call(
-                BuiltInMethod.TIMESTAMP_WITH_LOCAL_TIME_ZONE_TO_TIMESTAMP.method,
-                operand,
-                Expressions.call(BuiltInMethod.TIME_ZONE.method, root)));
-
-      default:
-        return defaultExpression.get();
-      }
+      return translateCastToTimeStamp(sourceType, operand, format, defaultExpression);
```
tanclary, line 332: should Timestamp be camelcase?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -647,58 +541,225 @@ private static Expression checkExpressionPadTruncate(
...
+      Expression operand, ConstantExpression format,
+      Supplier<Expression> defaultExpression) {
+
+    switch (sourceType.getSqlTypeName()) {
+    case CHAR:
+    case VARCHAR:
+      // If format string is supplied, parse formatted string into date
+      return Expressions.isConstantNull(format)
+          ? Expressions.call(BuiltInMethod.STRING_TO_DATE.method, operand)
+          : Expressions.call(Expressions.new_(BuiltInMethod.PARSE_DATE.method.getDeclaringClass()),
```
tanclary, line 554: The declaring class is always SqlFunctions right? Or no? If it's always the same we can just pass the class directly instead of calling this getter? Same comment for above

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -647,58 +541,225 @@ private static Expression checkExpressionPadTruncate(
...
+  }
+
+  private Expression translateCastToTimeStamp(RelDataType sourceType,
+      Expression operand, ConstantExpression format, Supplier<Expression> defaultExpression) {
+
+    switch (sourceType.getSqlTypeName()) {
+    case CHAR:
+    case VARCHAR:
+      // If format string is supplied, parse formatted string into timestamp
+      return Expressions.isConstantNull(format)
```
tanclary, line 665: I'm noticing a pattern:

We check if format is null, and then we call the right method accordingly. Is there a general function we can write that takes in 3 parameters (format, and then the two methods). It would shorten a lot of these switch statements. let me know why this is or isn't possible

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -832,6 +858,25 @@ public RexNode makeAbstractCast(RelDataType type, RexNode exp, boolean safe) {
...
+   * Creates a call to CAST or SAFE_CAST operator with a FORMAT clause.
+   *
+   * @param type Type to cast to
+   * @param exp  Expression being cast
+   * @param safe Whether to return NULL if cast fails
+   * @param format Conversion format for target type
+   * @return Call to CAST operator
+   */
+  public RexNode makeAbstractCast(RelDataType type, RexNode exp, boolean safe, RexLiteral format) {
+    SqlOperator operator =
```
tanclary, line 871: operator can be final?

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -2267,16 +2267,21 @@ && sameTypeOrNarrowsNullability(e.getType(), intExpr.getType())) {
...
+          : rexBuilder.makeCast(e.getType(), operand, safe, safe);
       executor.reduce(rexBuilder, ImmutableList.of(simplifiedExpr), reducedValues);
       return requireNonNull(
           Iterables.getOnlyElement(reducedValues));
     default:
       if (operand == e.getOperands().get(0)) {
         return e;
       } else {
-        return rexBuilder.makeCast(e.getType(), operand, safe, safe);
+        return e.operands.size() > 1
```
tanclary, line 2281: Same comment as above about `== 2`

tanclary: Or can this be 2 or 3? I've confused myself

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1345,6 +1345,169 @@ void testCastStringToDateTime(CastType castType, SqlOperatorFixture f) {
...
+
+    if (Bug.CALCITE_6270_FIXED) {
+      f.checkString("cast(-12.23 as varchar FORMAT '999.999')",
+          "-12.230",
+          "VARCHAR NOT NULL");
+      f.checkString("cast(1234.56 as varchar FORMAT '$999,999.999')",
+          "$1,234.560",
+          "VARCHAR NOT NULL");
+      f.checkString("cast(123456 as varchar FORMAT '9.999EEEE')",
+          "1.235E+05",
```
tanclary, line 1506: Does adding a test for `SAFE_CAST` give us any value here?

tanclary (changes requested):

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3262,11 +3262,21 @@ private static class CastImplementor extends AbstractRexCallImplementor {
...
     @Override Expression implementSafe(final RexToLixTranslator translator,
         final RexCall call, final List<Expression> argValueList) {
-      assert call.getOperands().size() == 1;
+      assert call.getOperands().size() <= 2;
       final RelDataType sourceType = call.getOperands().get(0).getType();
 
-      // Short-circuit if no cast is required
       RexNode arg = call.getOperands().get(0);
+      ConstantExpression formatExpr;
+      if (call.getOperands().size() > 1) {
```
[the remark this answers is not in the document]
Anthrino: Makes sense, yes the options should only be 1 or 2 for now, if we have a `AT <Timezone>` clause added in later it would become 3, but irrespective we should make it a deterministic equality check.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
Anthrino: The declaring functions are from `SqlFunctions` but might belong to subclasses, e.g.
```
FORMAT_DATE => SqlFunctions.DateFormatFunction.class
PARSE_TIME => SqlFunctions.DateParseFunction.class
```
my assumption is we need to pass the inner class name as the `targetExpression`, I am not completely aware of the use of this [`targetExpression`](https://github.com/apache/calcite/blob/7ca6fc4bdd3e1274a62f00f4c8d5830162a74879/linq4j/src/main/java/org/apache/calcite/linq4j/tree/MethodCallExpression.java#L33) attribute in the  `MethodCallExpression` class, if anyone else can confirm?

tanclary: Yeah that makes sense. I think explicitly passing the class is more readable, it tells the user exactly where the method is defined. I also think it conforms better with the rest of Calcite. Up to you though

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
tanclary: SGTM 👍

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
Anthrino: Yes that should work, will make the change to the existing `makeAbstractCast` overload too.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
Anthrino: Essentially both operators should take the same flow, I will add some tests to see it there's any deviation in functionality.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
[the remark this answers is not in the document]
julianhyde: `SqlCall` has `<S extends SqlNode> S operand(int)` and `int operandCount()`. I think we should follow that model.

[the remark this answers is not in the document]
tanclary: Yeah either or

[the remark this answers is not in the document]
julianhyde: In situations like this, I use `> 1`. It is more future-proof, and asks for only what it needs.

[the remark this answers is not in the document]
tanclary: In the future if there was another argument we may not want it to follow the same logic, that's why I think equivalency is a better check.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: Ok no need to make this PR any bigger if it doesn't give us anything so it's up to you

Anthrino: Changed all the tests to use `safeParameters()` to test all CAST functions.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
julianhyde: A few other SQL functions are implemented by non-static methods (generally to allow caching of expensive state). Do whatever those functions do.

Anthrino: Thanks for the pointers, I couldn't find an explicit usage of a non-static built-in method in `Expressions.call()`, but saw `getDeclaringClass()` used to build the target expression so using that pattern; I think it's easier to trace to the class of the method call even though its a slight hit on readability.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -3262,11 +3262,21 @@ private static class CastImplementor extends AbstractRexCallImplementor {
...
     @Override Expression implementSafe(final RexToLixTranslator translator,
         final RexCall call, final List<Expression> argValueList) {
-      assert call.getOperands().size() == 1;
+      assert call.operandCount() <= 2;
       final RelDataType sourceType = call.getOperands().get(0).getType();
 
-      // Short-circuit if no cast is required
       RexNode arg = call.getOperands().get(0);
+      ConstantExpression formatExpr;
+      if (call.operandCount() == 2) {
```
tanclary, line 3270: comment here could be good

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -826,12 +852,31 @@ public RexNode makeAbstractCast(RelDataType type, RexNode exp) {
...
+  /**
+   * Creates a call to CAST or SAFE_CAST operator with a FORMAT clause.
+   *
+   * @param type Type to cast to
+   * @param exp  Expression being cast
+   * @param safe Whether to return NULL if cast fails
+   * @param format Conversion format for target type
+   * @return Call to CAST operator
+   */
+  public RexNode makeAbstractCast(RelDataType type, RexNode exp, boolean safe, RexLiteral format) {
```
tanclary, line 870: does this need to be nullable

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
```diff
@@ -239,6 +245,12 @@ private static RelDataType createTypeWithNullabilityFromExpr(RelDataTypeFactory
       }
       return false;
     }
+
+    // Validate format argument is string type if included
```
tanclary, line 249: can this be `return <insert conditions>`

core/src/main/java/org/apache/calcite/util/format/FormatModels.java
```diff
@@ -142,9 +142,7 @@ MI, literalElement(":"), SS, literalElement(" "),
     map.put("%Y", YYYY);
     map.put("%y", YY);
     map.put("%Z", TZR);
-    BIG_QUERY = create(map);
 
-    map.clear();
```
tanclary, line 147: why get rid of this

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -6218,6 +6218,22 @@ private static Matcher<SqlNode> isCharLiteral(String s) {
         .ok("CAST(`A` AS MAP< VARCHAR MULTISET, MAP< INTEGER, INTEGER > >)");
   }
 
+  @Test void testFormatClauseInCast() {
```
tanclary, line 6221: while adding comments may as well add one here to case

tanclary (changes requested):

core/src/main/java/org/apache/calcite/util/format/FormatModels.java
Anthrino: That was clearing off all the previous records added to the `FormatElements` map after creating the `BigQuery` FormatModel (line 145), to add new values and create the `POSTGRESQL` model (line 177).

All the below format elements used by `POSTGRESQL` are also supposed to be available in `BigQuery`, so instead of clearing it we can add all those elements and create both models at the end.

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
Anthrino: Good catch, simplified into single return statement!

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
Anthrino: If you're referring to the format argument being null, we're creating a null valued RexLiteral object if we receive an empty format string from the parser, so it doesnt need to be nullable.

core/src/main/java/org/apache/calcite/util/format/FormatModels.java
Anthrino: There would be some restructuring / additions to this set of FormatElements based on the changes required in [CALCITE-6269](https://issues.apache.org/jira/browse/CALCITE-6269)

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
tanclary: oh duh

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -469,34 +349,48 @@ private Expression getConvertExpression(
...
+            : Expressions.call(
+                Expressions.new_(
+                    BuiltInMethod.FORMAT_DATE.method.getDeclaringClass()),
+                BuiltInMethod.FORMAT_DATE.method, format, operand));
 
       case TIME:
-        return RexImpTable.optimize2(operand,
-            Expressions.call(BuiltInMethod.UNIX_TIME_TO_STRING.method,
-                operand));
+        return RexImpTable.optimize2(operand, Expressions.isConstantNull(format)
```
tanclary, line 362: can you evaluate this before the `switch` ? then you can just check the variable in each case

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -647,58 +541,225 @@ private static Expression checkExpressionPadTruncate(
...
+                      operand)),
+              Expressions.call(BuiltInMethod.TIME_ZONE.method, root));
 
     case TIMESTAMP:
-      return Expressions.convert_(
-          Expressions.call(BuiltInMethod.FLOOR_DIV.method,
-              operand, Expressions.constant(DateTimeUtils.MILLIS_PER_DAY)),
-          int.class);
+      return
+          Expressions.call(BuiltInMethod.TIMESTAMP_STRING_TO_TIMESTAMP_WITH_LOCAL_TIME_ZONE.method,
```
tanclary, line 638: is this aligned right? displaying weird on github

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -647,58 +541,225 @@ private static Expression checkExpressionPadTruncate(
...
+  private Expression translateCastToTimestampWithLocalTimeZone(RelDataType sourceType,
+      Expression operand, Supplier<Expression> defaultExpression) {
+
+    switch (sourceType.getSqlTypeName()) {
+    case CHAR:
+    case VARCHAR:
+      return
+          Expressions.call(BuiltInMethod.STRING_TO_TIMESTAMP_WITH_LOCAL_TIME_ZONE.method,
+              operand);
+
```
tanclary, line 718: do we want newline here

core/src/main/java/org/apache/calcite/rex/RexCall.java
```diff
@@ -251,6 +251,10 @@ public List<RexNode> getOperands() {
     return operands;
   }
 
+  public int operandCount() {
```
tanclary, line 254: i might have been the one to suggest this, but is `getOperandCount` better since it follows typical getter naming convention

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
```diff
@@ -191,12 +194,12 @@ private static RelDataType createTypeWithNullabilityFromExpr(RelDataTypeFactory
...
   @Override public String getSignatureTemplate(final int operandsCount) {
-    assert operandsCount == 2;
-    return "{0}({1} AS {2})";
+    assert operandsCount <= 3;
+    return "{0}({1} AS {2} [FORMAT {3}])";
   }
 
   @Override public SqlOperandCountRange getOperandCountRange() {
-    return SqlOperandCountRanges.of(2);
+    return SqlOperandCountRanges.between(2, 3);
```
tanclary, line 202: does this override still need to exist ? maybe try deleting it and see what happens. the super looks like it should do what you need

core/src/test/resources/sql/cast-with-format.iq
```diff
@@ -517,173 +530,182 @@ select cast('2010 14 january' as date FORMAT
...
 
-select cast(date'2010-08-19' as string FORMAT
+select cast(date'2010-08-19' as varchar FORMAT
     'FMMONTH FMMonth FMmonth');
 EXPR$0
 AUGUST August august
 !ok
 
-select cast(date'2010-09-19' as string FORMAT
+select cast(date'2010-09-19' as varchar FORMAT
```
tanclary, line 694: thanks for going through these im sure that took a bit

core/src/main/java/org/apache/calcite/rex/RexCall.java
Anthrino: That's right @tanclary, we considered that and I think Julian suggested we go with this method name to follow the `SqlCall` pattern:

`SqlCall` has `<S extends SqlNode> S operand(int)` and `int operandCount()`

core/src/main/java/org/apache/calcite/sql/fun/SqlCastFunction.java
Anthrino: I think that override is required in every function implementation to specify the number of operands being passed, it was already existing with a count = 2 just updated that to be 2 or 3. 
Removing that does fail a number of tests and breaks the build, it throws the `Util.needToImplement` exception from the super as `operandTypeChecker` is not defined and we need an override.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
Anthrino: Not exactly, we could actually only move the `RexImpTable.optimize2()` call outside the switch, but the specific method called to evaluate the CAST depends on the from and to cast data types, and with this change also if there is a `FORMAT` present. So the number of lines would still look the same could move out one level of nesting with that optimize call and return statements outside. Does that seem like a better pattern?

tanclary (approved):
I think this looks good thanks for addressing all the comments, maybe we can see if @julianhyde has any suggestions otherwise merge soon?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
Anthrino: Looks like it, didnt get any exceptions with the style check maybe github did a weird line wrap

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
Anthrino: I don't think its an added newline, just the line separators between each case statement like it was there before?

core/src/test/resources/sql/cast-with-format.iq
Anthrino: Good to have the working tests enabled, and the disabled ones gives a status on whats yet to be supported and need to be addressed in followup JIRAs.

PR #3678, opened 2024-02-08
[CALCITE-6247] BigQuery FORMAT_DATE function handles incorrectly the %e format specifier

mihaibudiu:


core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -81,6 +81,13 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(work.eeeFormat.format(date));
     }
   },
+  E("d", "The day of the month as a decimal number (01-31) left-padded with space") {
```
Anthrino, line 84: Hi @mihaibudiu, all changes looks good to me.

We can replace `(01-31)` in the comment to `(1-31)`, and maybe mention "single digits left-padded with space" to avoid confusion and align with BQ docs.

caicancai (commented):
It doesn't seem to affect the functionality of the original code, and it looks good to me.

Anthrino (approved):

PR #3679, opened 2024-02-08
[CALCITE-6256] Incorrect rendering of HTML on InnoDB adapter page

strongduanmu:
The new rendered HTML is as follows(IDEA Local preview without CSS style).

image

### DIFF
--- a/site/_docs/innodb_adapter.md
+++ b/site/_docs/innodb_adapter.md
@@ -95,9 +95,8 @@ a MySQL "scott" database:
 }
 {% endhighlight %}
 
-`sqlFilePath` is a list of DDL files, you can generate table
-definitions by executing `mysqldump -d -u<username> -p<password> -h
-<hostname> <dbname>` in command-line.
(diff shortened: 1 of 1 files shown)


julianhyde:
Seems to fix the problem.

I think that the "`mysqldump -d -u<username> -p<password> -h <hostname> <dbname>`" line would be better as its own code box, rather than inline; can you change it to use the highlight markup?

We don't allow commit messages with the word 'fix'. Can you change the commit message to match the jira case summary.

strongduanmu:
@julianhyde Thank you for your advice, I will modify this pr.

julianhyde:
You changed 'Fix' to 'Correct' in the commit message. Instead, can you change the commit message to the jira case summary? (If a commit message describes a bug, it is implied that the commit fixes the bug.)

strongduanmu:
Thank you very much for your guidance. I will pay attention to these norms of the Calcite community so that my next contribution can be more perfect.

julianhyde (approved):

PR #3680, opened 2024-02-09
[CALCITE-6252] BigQuery FORMAT_DATE uses the wrong calendar for Julian dates

mihaibudiu:
This was much trickier to fix than I expected.
I think that it's because of some bugs in the Java Calendar/SimpleDateFormatter implementation. 
Using the LocalDate class provides correct results.

I have also changed the return type of FORMAT_DATE from VARCHAR(2000) to VARCHAR, because it can really return results longer than 2000 characters, if the format string is long.


core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -57,8 +59,17 @@ public enum FormatElementEnum implements FormatElement {
...
   DAY("EEEE", "The full weekday name") {
     @Override public void format(StringBuilder sb, Date date) {
-      final Work work = Work.get();
-      sb.append(work.eeeeFormat.format(date));
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      // The Calendar and SimpleDateFormatter do not seem to give correct results
+      // for the day of the week prior to the Julian to Gregorian date change.
+      // So we resort to using a LocalDate representation.
+      LocalDate ld =
```
YiwenWu, line 67: Can use java.sql.Date directly?
`((java.sql.Date) date).toLocalDate()`

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -281,9 +301,9 @@ static Work get() {
 
     /** Uses Locale.US instead of Locale.ROOT to fix formatting in Java 11 */
     final DateFormat eeeeFormat = new SimpleDateFormat(DAY.javaFmt, Locale.US);
-    final DateFormat eeeFormat = new SimpleDateFormat(DY.javaFmt, Locale.ROOT);
-    final DateFormat mmmFormat = new SimpleDateFormat(MON.javaFmt, Locale.ROOT);
-    final DateFormat mmmmFormat = new SimpleDateFormat(MONTH.javaFmt, Locale.ROOT);
+    final DateFormat eeeFormat = new SimpleDateFormat(DY.javaFmt, Locale.US);
```
YiwenWu, line 304: Why use Locale.US?

mihaibudiu: SQL requires the output to be in English. Using Locale.ROOT gives the wrong result for the eeeeFormat at least (Mon instead of Monday).

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
mihaibudiu: I didn't know this trick, I will try it.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -281,9 +301,9 @@ static Work get() {
 
     /** Uses Locale.US instead of Locale.ROOT to fix formatting in Java 11 */
```
caicancai, line 302: Hello, Should we change the comment here?

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
caicancai: It looks good to me

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
caicancai: There are only Locale.ROOT comments here

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
mihaibudiu: it doesn't work, the subclass relationship goes the other way. sql.date extends util.date.
The calendar-based conversion seems to do the right thing and it's not too complicated.

caicancai (approved):
LGTM

PR #3682, opened 2024-02-11
[CALCITE-6257] StarRocks dialect implementation

YiwenWu:
Implementing StarRocks dialect in Calcite 
https://issues.apache.org/jira/browse/CALCITE-6257

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
@@ -1,0 +1,159 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.sql.dialect;
+
+import org.apache.calcite.avatica.util.TimeUnitRange;
+import org.apache.calcite.config.NullCollation;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeSystem;
+import org.apache.calcite.sql.SqlAbstractDateTimeLiteral;
+import org.apache.calcite.sql.SqlAlienSystemTypeNameSpec;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlDataTypeSpec;
+import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.SqlIntervalQualifier;
+import org.apache.calcite.sql.SqlLiteral;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlWriter;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -0,0 +1,136 @@
...
+      break;
+    default:
+      super.unparseCall(writer, call, leftPrec, rightPrec);
+      break;
+    }
+  }
+
+  @Override public @Nullable SqlNode getCastSpec(RelDataType type) {
+    switch (type.getSqlTypeName()) {
+    case TIMESTAMP:
```
mihaibudiu, line 119: I am confused what's going on here.
If the Calcite TIMESTAMP type is called "DATETIME" in StarRocks, I am not sure whether this is the only case that needs changing. Can you write some tests that use TIMESTAMP?

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -0,0 +1,136 @@
...
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static org.apache.calcite.util.RelToSqlConverterUtil.unparseHiveTrim;
+
+/**
+ * A <code>SqlDialect</code> implementation for the StarRocks database.
+ */
+public class StarRocksSqlDialect extends SqlDialect {
```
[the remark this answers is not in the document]
caicancai: As far as I know, the sql dialect of starrocks is compatible with mysql, but starrocks is an olap database that supports more data types and some special sql functions.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -4506,13 +4600,16 @@ private void checkLiteral2(String expression, String expected) {
     String expectedPresto = "SELECT DATE_TRUNC('MINUTE', \"hire_date\")\n"
         + "FROM \"foodmart\".\"employee\"";
     String expectedFirebolt = expectedPostgresql;
+    String expectedStarRocks = "SELECT DATE_FORMAT(`hire_date`, '%Y-%m-%d %H:%i:00')\n"
```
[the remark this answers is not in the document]
YiwenWu: Indeed, this test case directly uses `MysqlSqlDialect#unparseFloor`,  MySQL/StarRocks Function DATE_FORMAT returns a string. [date_format docs](https://docs.starrocks.io/docs/2.1/sql-reference/sql-functions/date-time-functions/date_format/)

StarRocks also support function [date_trunc](https://docs.starrocks.io/docs/2.1/sql-reference/sql-functions/date-time-functions/date_trunc/),  replace with DATE_TRUNC function

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
YiwenWu: Exactly,  Similar to MySQL, the Calcite TIMESTAMP type is called "DATETIME" in StarRocks.  This processing refers to `MysqlSqlDialect#getCastSpec`.

YiwenWu: Added additional TIMESTAMP test:
- `RelToSqlConverterTest#testCastToTimestamp`
- `RelToSqlConverterTest#testUnparseSqlIntervalQualifier`

macroguo-ghy:
Please solve the conflicts first.

from [starrocks doc](https://docs.starrocks.io/docs/introduction/what_is_starrocks/), Starrocks say 'Compatible with MySQL', so maybe `StarrocksSqlDialect` can extend from `MysqlSqldialect`?

YiwenWu:
> Please solve the conflicts first.
> 
> from [starrocks doc](https://docs.starrocks.io/docs/introduction/what_is_starrocks/), Starrocks say 'Compatible with MySQL', so maybe `StarrocksSqlDialect` can extend from `MysqlSqldialect`?

I'll try to handle it like this.

PR #3683, opened 2024-02-11
[CALCITE-6258] Map value constructor is unparsed incorrectly for PrestoSqlDialect

YiwenWu:
fixed Presto Map Function is unparsed incorrectly ：map(array<K>, array<V>) 
https://issues.apache.org/jira/browse/CALCITE-6258

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
@@ -20,18 +20,27 @@ import org.apache.calcite.avatica.util.Casing;
 import org.apache.calcite.config.NullCollation;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
+import org.apache.calcite.sql.SqlBasicCall;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlDialect;
 import org.apache.calcite.sql.SqlIntervalQualifier;
 import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlLiteral;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.fun.SqlArrayValueConstructor;
+import org.apache.calcite.sql.fun.SqlMapValueConstructor;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.util.RelToSqlConverterUtil;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
+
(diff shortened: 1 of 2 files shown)


JiajunBernoulli:
Good catch. 
I have a similar PR: https://github.com/linkedin/coral/pull/99/files

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -4169,7 +4171,10 @@ private SqlDialect nonOrdinalDialect() {
     final String query = "SELECT MAP['k1', 'v1', 'k2', 'v2']";
     final String expected = "SELECT MAP ('k1', 'v1', 'k2', 'v2')\n"
         + "FROM (VALUES (0)) t (ZERO)";
-    sql(query).withSpark().ok(expected);
+    final String expectedPresto = "SELECT MAP (ARRAY['k1', 'k2'], ARRAY['v1', 'v2'])\n"
```
caicancai, line 4174: Maybe we should add a comment with a jira link in the comment

core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
```diff
@@ -131,8 +136,30 @@ private static void unparseUsingLimit(SqlWriter writer, @Nullable SqlNode offset
       RelToSqlConverterUtil.specialOperatorByName("APPROX_DISTINCT")
           .unparse(writer, call, 0, 0);
     } else {
-      // Current impl is same with Postgresql.
-      PostgresqlSqlDialect.DEFAULT.unparseCall(writer, call, leftPrec, rightPrec);
+      switch (call.getKind()) {
```
JiajunBernoulli, line 139: Extract them to `createMapValueConstuctor` function is bettter.
> I like single responsibility principle

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
YiwenWu: done, added new tests

core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
YiwenWu: done, extract new function `convertMapValueCall`

JiajunBernoulli: Why don't we extract more thoroughly?

We can use more parameter like `unparseCall(writer, call, leftPrec, rightPrec)`

YiwenWu: done, extract map unparse  method `unparseMapValue`

macroguo-ghy (approved):
+1 for this pr.

But `MAP[]` will be rejected by `SqlMapValueConstructor#checkOperandTypes`, so I think we need add a new `SqlLibrary` and add (or enable) `map` function for presto.

macroguo-ghy:
Please squash your commits, I will merge it later. @YiwenWu

PR #3687, opened 2024-02-14
[CALCITE-6265] Type coercion is failing for numeric values in prepared statements

tindzk:
Given a column of type `INT`. When providing a `short` value as a placeholder in a prepared statement, a `ClassCastException` is thrown.

## Test case
```java
final String sql =
    "select \"empid\" from \"hr\".\"emps\" where \"empid\" in (?, ?)";
CalciteAssert.hr()
    .query(sql)
    .consumesPreparedStatement(p -> {
        p.setShort(1, (short) 100);
        p.setShort(2, (short) 110);
    })
    .returnsUnordered("empid=100", "empid=110");
```

## Stack trace
```
java.lang.ClassCastException: class java.lang.Short cannot be cast to class java.lang.Integer (java.lang.Short and java.lang.Integer are in module java.base of loader 'bootstrap')
     at Baz$1$1.moveNext(Unknown Source)
     at org.apache.calcite.linq4j.Linq4j$EnumeratorIterator.<init>(Linq4j.java:679)
```

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
@@ -31,6 +31,7 @@ import org.apache.calcite.linq4j.tree.Expressions;
 import org.apache.calcite.linq4j.tree.ParameterExpression;
 import org.apache.calcite.linq4j.tree.Primitive;
 import org.apache.calcite.linq4j.tree.Statement;
+import org.apache.calcite.linq4j.tree.Types;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexCall;
@@ -1374,11 +1375,26 @@ public class RexToLixTranslator implements RexVisitor<RexToLixTranslator.Result>
     }
     final Type storageType = currentStorageType != null
         ? currentStorageType : typeFactory.getJavaClass(dynamicParam.getType());
-    final Expression valueExpression =
+
+    final boolean isNumeric = SqlTypeFamily.NUMERIC.contains(dynamicParam.getType());
+
+    // For numeric types, use java.lang.Number to prevent cast exception
+    // when the parameter type differs from the target type
+    Expression argumentExpression =
         EnumUtils.convert(
             Expressions.call(root, BuiltInMethod.DATA_CONTEXT_GET.method,
                 Expressions.constant("?" + dynamicParam.getIndex())),
-            storageType);
+            isNumeric ? java.lang.Number.class : storageType);
+
+    // Short-circuit if the expression evaluates to null. The cast
+    // may throw a NullPointerException as it calls methods on the
+    // object such as longValue().
+    Expression valueExpression =
+        Expressions.condition(
+            Expressions.equal(argumentExpression, Expressions.constant(null)),
+            Expressions.constant(null),
+            Types.castIfNecessary(storageType, argumentExpression));
+
     final ParameterExpression valueVariable =
         Expressions.parameter(valueExpression.getType(),
             list.newName("value_dynamic_param"));
--- a/core/src/test/java/org/apache/calcite/test/JdbcTest.java
+++ b/core/src/test/java/org/apache/calcite/test/JdbcTest.java
@@ -84,6 +84,7 @@ import org.apache.calcite.sql.SqlSpecialOperator;
 import org.apache.calcite.sql.parser.SqlParser;
 import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.parser.impl.SqlParserImpl;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql2rel.SqlToRelConverter.Config;
 import org.apache.calcite.test.schemata.catchall.CatchallSchema;
 import org.apache.calcite.test.schemata.foodmart.FoodmartSchema;
@@ -8421,6 +8422,98 @@ public class JdbcTest {
             throw new RuntimeException(e);
           }
         });
+  }
+
+  @Test void bindByteParameter() {
+    for (SqlTypeName tpe : SqlTypeName.INT_TYPES) {
+      final String sql =
+          "with cte as (select cast(100 as " + tpe.getName() + ") as empid)"
+              + "select * from cte where empid = ?";
+      CalciteAssert.hr()
+          .query(sql)
+          .consumesPreparedStatement(p -> {
+            p.setByte(1, (byte) 100);
+          })
+          .returnsUnordered("EMPID=100");
+    }
+  }
+
+  @Test void bindShortParameter() {
+    for (SqlTypeName tpe : SqlTypeName.INT_TYPES) {
+      final String sql =
+          "with cte as (select cast(100 as " + tpe.getName() + ") as empid)"
+              + "select * from cte where empid = ?";
+
+      CalciteAssert.hr()
+          .query(sql)
+          .consumesPreparedStatement(p -> {
+            p.setShort(1, (short) 100);
+          })
+          .returnsUnordered("EMPID=100");
+    }
+  }
+
+  @Test void bindOverflowingTinyIntParameter() {
+    final String sql =
+        "with cte as (select cast(300 as smallint) as empid)"
+            + "select * from cte where empid = cast(? as tinyint)";
+
+    java.sql.SQLException t =
+        assertThrows(
+          java.sql.SQLException.class,
+          () -> CalciteAssert.hr()
+            .query(sql)
+            .consumesPreparedStatement(p -> {
+              p.setShort(1, (short) 300);
+            })
+            .returns(""));
+
+    assertThat(
+        "message matches",
+        t.getMessage().contains("value is outside the range of java.lang.Byte"));
+  }
+
+  @Test void bindIntParameter() {
+    for (SqlTypeName tpe : SqlTypeName.INT_TYPES) {
+      final String sql =
+          "with cte as (select cast(100 as " + tpe.getName() + ") as empid)"
+              + "select * from cte where empid = ?";
+
+      CalciteAssert.hr()
+          .query(sql)
+          .consumesPreparedStatement(p -> {
+            p.setInt(1, 100);
+          })
+          .returnsUnordered("EMPID=100");
+    }
+  }
+
+  @Test void bindLongParameter() {
+    for (SqlTypeName tpe : SqlTypeName.INT_TYPES) {
+      final String sql =
+          "with cte as (select cast(100 as " + tpe.getName() + ") as empid)"
+              + "select * from cte where empid = ?";
+
+      CalciteAssert.hr()
+          .query(sql)
+          .consumesPreparedStatement(p -> {
+            p.setLong(1, 100);
+          })
+          .returnsUnordered("EMPID=100");
+    }
+  }
+
+  @Test void bindNumericParameter() {
+    final String sql =
+        "with cte as (select cast(100 as numeric(5)) as empid)"
+            + "select * from cte where empid = ?";
+
+    CalciteAssert.hr()
+        .query(sql)
+        .consumesPreparedStatement(p -> {
+          p.setLong(1, 100);
+        })
+        .returnsUnordered("EMPID=100");
   }
 
   private static String sums(int n, boolean c) {
--- a/linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
+++ b/linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
@@ -622,9 +622,49 @@ public abstract class Expressions {
    * operation that throws an exception if the target type is
    * overflowed.
    */
-  public static UnaryExpression convertChecked(Expression expression,
+  public static Expression convertChecked(Expression expression,
       Type type) {
-    throw Extensions.todo();
+    if (type == Byte.class
+            || type == Short.class
+            || type == Integer.class
+            || type == Long.class) {
+      Class<?> typeClass = (Class<?>) type;
+
+      Object minValue;
+      Object maxValue;
+
+      try {
+        minValue = typeClass.getField("MIN_VALUE").get(null);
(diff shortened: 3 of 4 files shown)


core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -1374,11 +1374,35 @@ private Result toInnerStorageType(Result result, Type storageType) {
     }
     final Type storageType = currentStorageType != null
         ? currentStorageType : typeFactory.getJavaClass(dynamicParam.getType());
-    final Expression valueExpression =
+
+    // For numeric types, get the value using the following functions on the
```
mihaibudiu, line 1378: This may work for numeric values, but does not seem to be a general solution for other types. Is there a design that would handle that? It looks like any source/destination type combination is possible. So perhaps this could be handled like a cast expression? Although I don't see the destination type in this function.

tindzk: Indeed. The PR only fixes the issue for numeric values, but for other types the user would still encounter a `ClassCastException`. Ideally, we would have a function that handles all implicit conversions and fails with a user-friendly error for unsupported conversions. I am sure this functionality already exists in the Calcite in the form of `CAST` expressions. Possibly it could be leveraged here.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -8301,6 +8301,67 @@ private void checkGetTimestamp(Connection con) throws SQLException {
         });
   }
 
+  final String[] intTypes = {"tinyint", "smallint", "int", "bigint"};
```
asolimando, line 8304: You can use `SqlTypeFamily.INTEGER` to derive those type names

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
asolimando: As @mihaibudiu suggests, this is basically a cast, which is already available via the `Types.castIfNecessary` method, see below for how it can be used:

```
    final boolean isNumeric = SqlTypeFamily.NUMERIC.contains(dynamicParam.getType());
    final Expression valueExpression = Types.castIfNecessary(storageType,
        EnumUtils.convert(Expressions.call(root, BuiltInMethod.DATA_CONTEXT_GET.method,
                Expressions.constant("?" + dynamicParam.getIndex())),
            isNumeric ? java.lang.Number.class : storageType));
```

If you replace your code with this, it passes all the tests you added, and it might be working for non-numeric cases too (maybe with some adaptations).

Could you give it a try and add more tests non-numeric types to check?

asolimando (changes requested):

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
tindzk: Thanks for the suggestion! It greatly simplifies the fix.

I made the requested change, but kept the null check as the cast operation may fail. For instance, `o.intValue()` requires `o` to be non-null. Otherwise, `testPreparedStatement` would fail.

As for additional tests, do you have in mind assigning a non-numeric value to a numeric value? These would still raise an exception:

```java
  @Test void bindStringParameter() {
    for (SqlTypeName tpe : SqlTypeName.INT_TYPES) {
      final String sql =
          "with cte as (select cast(100 as " + tpe.getName() + ") as empid)"
              + "select * from cte where empid = ?";

      CalciteAssert.hr()
          .query(sql)
          .consumesPreparedStatement(p -> {
            p.setString(1, "100");
          })
          .returnsUnordered("EMPID=100");
    }
  }
```

This throws a `ClassCastException` which we could rewrite into a more user-friendly message. Alternatively, we could prevent the conversion happening in the first place by introducing a validation check.

```
java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Number
```

mihaibudiu: I personally would file a new Jira issue for the casts not covered and merge this PR as is.
You have to decide which casts can appear in this context.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -8301,6 +8302,65 @@ private void checkGetTimestamp(Connection con) throws SQLException {
...
+      CalciteAssert.hr()
+          .query(sql)
+          .consumesPreparedStatement(p -> {
+            p.setInt(1, 100);
+          })
+          .returnsUnordered("EMPID=100");
+    }
+  }
+
+  @Test void bindLongParameter() {
```
mihaibudiu, line 8349: Can you also add a test where the value overflows the type, e.g., TINYINT with a value of 300?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
tindzk: Agreed. It seems unrelated to this PR. I created a ticket for it: https://issues.apache.org/jira/browse/CALCITE-6284

core/src/test/java/org/apache/calcite/test/JdbcTest.java
mihaibudiu: CAST(300 as TINYINT)

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
asolimando: Sorry I haven't been very precise, I didn't mean to cast Strings to numbers, what I had in mind was values from the string family to be used in parameters of the same family, possibly with (e.g., `varchar(2)` value for `varchar(3)` parameter), but I don't know if it is even expressible with the APIs we use to pass literals in, so feel free to leave this outside the scope of the present PR in case.

asolimando: > I made the requested change, but kept the null check as the cast operation may fail. For instance, `o.intValue()` requires `o` to be non-null. Otherwise, `testPreparedStatement` would fail.

I might have made a mistake but I tried all your tests with the suggested piece of code and all passed.
Did you add a new test specifically for this maybe?

core/src/test/java/org/apache/calcite/test/JdbcTest.java
tindzk: Thanks for clarifying. `CAST(300 as TINYINT)` is correctly fine, but `CAST(? as TINYINT)` does not trigger an error when the value is out of bounds. I added a test and fixed it.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
tindzk: Yes, that makes sense. I added a similar test based on Mihai's request below. Since this PR is mostly about numeric types, `VARCHAR` conversions could be tackled in a separate ticket.

Indeed. All the tests I added were passing with your change, but `testPreparedStatement` existed before and started failing. Explicitly checking whether the value is `null`, fixed it.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
```diff
@@ -622,9 +622,49 @@ public static UnaryExpression convert_(Expression expression, Type type,
    * operation that throws an exception if the target type is
    * overflowed.
    */
-  public static UnaryExpression convertChecked(Expression expression,
+  public static Expression convertChecked(Expression expression,
       Type type) {
-    throw Extensions.todo();
+    if (type == Byte.class
```
mihaibudiu, line 627: It's funny, I implemented this myself in a separate PR: https://github.com/apache/calcite/pull/3589
There I am relying on an existing implementation in Primitive.numberValue.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
```diff
@@ -2829,6 +2869,18 @@ public static ThrowStatement throw_(Expression expression) {
     return new ThrowStatement(expression);
   }
 
+  /**
+   * Creates an expression that represents the throwing of an exception.
+   */
+  public static Expression throwExpr(Expression expression) {
```
mihaibudiu, line 2875: There's a ThrowStatement.
Maybe having an expression is useful, since it can be used in an expression context - you don't need to break the current expression to create a new statement.
I would document this feature.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
tindzk: Nice. Your approach seems to cover more scenarios.

I will keep `convertChecked` as-is for now. The inlined validations should be sufficient within the current PR's scope, but feel free to replace it once #3589 has been merged.

asolimando (changes requested):
The PR is getting to a mergeable status, but we need tests covering the new/modified expressions in [ExpressionTest](https://github.com/apache/calcite/blob/main/linq4j/src/test/java/org/apache/calcite/linq4j/test/ExpressionTest.java).

Could you also check Sonar issues and fix the ones which are legit?

linq4j/src/main/java/org/apache/calcite/linq4j/tree/CastExpression.java
```diff
@@ -0,0 +1,81 @@
...
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.lang.reflect.Type;
+import java.util.Objects;
+
+/**
+ * Represents an expression cast to the specified type.
+ */
+public class CastExpression extends Expression {
```
mihaibudiu, line 27: I thought I left a comment in my previous review that the functionality of Cast already seems to exist: it is a UnaryExpression with type Convert. I don't think we should have two ways to do the same thing unless you can explain clearly how this is different; this will be very hard to maintain, since people won't know where to look for the functionality of casts. Why can't you reuse the Convert?
For some reason I don't see this comment, maybe I forgot to type it.

asolimando: Agree, this was also suggested in https://github.com/apache/calcite/pull/3687#discussion_r1490150823 and adopted in a following commit, so the change was intentional I guess, but if possible let's reuse `convert`

tindzk: Thanks for pointing it out. I was not aware that `Convert` maps onto a cast operation. I removed `CastExpression` and changed the code to use `Expressions.convert_()` instead.

asolimando (approved):
LGTM, nice work @tindzk!

mihaibudiu:
the next step is to rebase on main and squash all the commits into a single one in preparation for merging

asolimando:
> the next step is to rebase on main and squash all the commits into a single one in preparation for merging

@tindzk please make sure the commit message is well formatted and it matches the title of the corresponding Jira ticket when you rebase, thanks!

tindzk:
@asolimando I squashed the commits and updated the title.

asolimando (changes requested):
@tindzk by checking the Jira ticket I just remembered we didn't follow up on Julian's suggestions to look into the Avatica side before trying to fix it into the Calcite side.

Not surprisingly he seems to be right: if you check [AvaticaSite.java#L93](https://github.com/apache/calcite-avatica/blob/275a082d364e00a9f2e9d50a4468fef0782e3c81/core/src/main/java/org/apache/calcite/avatica/AvaticaSite.java#L93), you have `setXYZ` methods, where we have [all the information](https://github.com/apache/calcite-avatica/blob/275a082d364e00a9f2e9d50a4468fef0782e3c81/core/src/main/java/org/apache/calcite/avatica/AvaticaSite.java#L66) we need to check overflows and possibly do type coercion.

It seems the right thing to do for two reasons:
1) it's early on, the earliest we fix this kind of issues, the better it is
2) kind of a consequence of 1), but the fix in `Linq` realms seems overly complicated compared to what we can do on the Avatica side, this suggests that we should do it before landing there

For these reasons I am removing my approval and I am asking you to check the aforementioned `AvaticaSide` and see if we can get a simpler fix there.

If after this due diligence we are still convinced that Calcite (Linq) is the better place to have this handled, let's make our point in Jira and try to have consensus there.

Apologies for forgetting to follow-up on Julian's message, I have also put the "discussion-in-jira" label as a reminder but it has slipped through the cracks nonetheless!

tindzk:
After checking the referenced code I am not sure whether the issue can be fixed in Avatica. In order to perform type coercion, the target type needs to be known which does not seem to be the available there. Without this information, we also cannot check for overflows.

Note that type conversions are already happening in `RexToLixTranslator` as part of the `EnumUtils.convert()` call. The main contribution of my PR is to correctly convert numeric types to the expected target type. Currently, these conversions are failing due to a Java limitation that prevents numeric object types (`java.lang.Long` etc.) from being [directly cast](https://mkyong.com/java/java-lang-classcastexception-class-java-lang-integer-cannot-be-cast-to-class-java-lang-long/) to another numeric type. This is solved by first deriving a primitive `long` (not `Long`) value that can then be safely cast to the target type. This indirection does require an additional check because the object on which `longValue()` gets called may be null. I have not been able to find a simpler workaround for this Java limitation.

The remaining changes are unrelated to the reported issue and fix tests that were requested during the PR review, notably passing values to `numeric(N)` placeholders, and preventing overflows that could arise in the conversion process above. I also expect that `convertChecked()` will be replaced by Mihai's implementation after https://github.com/apache/calcite/pull/3589 has been merged which will increase the type coverage.

mihaibudiu:
From my understanding of the codebase this is not an Avatica issue.

mihaibudiu:
> From my understanding of the codebase this is not an Avatica issue.

Maybe I am wrong. I guess this is a bug in the transfer of data between JDBC and the evaluator, and JDBC is the one that should enforce the constraints given by the type. I haven't looked at the flow of data enough to judge this.

asolimando (commented):
(removing the request for changes as my questions have been answered)

PR #3688, opened 2024-02-14
[CALCITE-6219] Must Filter validation

olivrlee:


core/src/main/java/org/apache/calcite/sql/validate/IdentifierNamespace.java
```diff
@@ -187,33 +188,31 @@ private SqlValidatorNamespace resolveImpl(SqlIdentifier id) {
...
   @Override public RelDataType validateImpl(RelDataType targetRowType) {
     resolvedNamespace = resolveImpl(id);
-    if (resolvedNamespace instanceof TableNamespace) {
-      SqlValidatorTable table = ((TableNamespace) resolvedNamespace).getTable();
-      if (validator.config().identifierExpansion()) {
+    validator.validateNamespace(resolvedNamespace, targetRowType);
+
+    if (validator.config().identifierExpansion()) {
+      SqlValidatorTable table = resolvedNamespace.getTable();
+      if (table != null) {
```
tanclary, line 195: It's hard to tell on github but could we remove a level of nesting if we change this to `if (table == null) { break; }`?

PR #3689, opened 2024-02-15
[CALCITE-6224] Add LOG2 function (enabled in Mysql, Spark library)

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6224

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -208,6 +208,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.JSON_TYPE;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LEFT;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LEVENSHTEIN;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG2;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_AND;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_OR;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LPAD;
@@ -637,6 +638,7 @@ public class RexImpTable {
       defineMethod(EXP, BuiltInMethod.EXP.method, NullPolicy.STRICT);
       defineMethod(POWER, BuiltInMethod.POWER.method, NullPolicy.STRICT);
       defineMethod(ABS, BuiltInMethod.ABS.method, NullPolicy.STRICT);
+      defineMethod(LOG2, BuiltInMethod.LOG2.method, NullPolicy.STRICT);
 
       map.put(LN, new LogImplementor());
       map.put(LOG, new LogImplementor());
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -2764,7 +2764,7 @@ public class SqlFunctions {
   }
 
 
-  // LN, LOG, LOG10
+  // LN, LOG, LOG10, LOG2
 
   /** SQL {@code LOG(number, number2)} function applied to double values. */
   public static double log(double d0, double d1) {
@@ -2786,6 +2786,17 @@ public class SqlFunctions {
   /** SQL {@code LOG(number, number2)} function applied to double values. */
   public static double log(BigDecimal d0, BigDecimal d1) {
     return Math.log(d0.doubleValue()) / Math.log(d1.doubleValue());
+  }
+
+  /** SQL {@code LOG2(number)} function applied to double values. */
+  public static @Nullable Double log2(double number) {
+    return (number <= 0) ? null : log(number, 2);
+  }
+
+  /** SQL {@code LOG2(number)} function applied to
+   * BigDecimal values. */
+  public static @Nullable Double log2(BigDecimal number) {
+    return log2(number.doubleValue());
   }
 
   // MOD
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -2149,6 +2149,14 @@ public abstract class SqlLibraryOperators {
           OperandTypes.NUMERIC_OPTIONAL_NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
+  /** The "LOG2(numeric)" function. Returns the base 2 logarithm of numeric. */
+  @LibraryOperator(libraries = {MYSQL, SPARK})
+  public static final SqlFunction LOG2 =
+      SqlBasicFunction.create("LOG2",
+          ReturnTypes.DOUBLE_FORCE_NULLABLE,
+          OperandTypes.NUMERIC,
+          SqlFunctionCategory.NUMERIC);
+
   @LibraryOperator(libraries = {BIG_QUERY, SPARK})
   public static final SqlFunction POW =
       SqlStdOperatorTable.POWER.withName("POW");
--- a/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
@@ -397,6 +397,13 @@ public abstract class ReturnTypes {
       DOUBLE.andThen(SqlTypeTransforms.TO_NULLABLE);
 
   /**
+   * Type-inference strategy whereby the result type of a call is a nullable
+   * Double.
+   */
+  public static final SqlReturnTypeInference DOUBLE_FORCE_NULLABLE =
+      DOUBLE.andThen(SqlTypeTransforms.FORCE_NULLABLE);
+
+  /**
    * Type-inference strategy whereby the result type of a call is a Char.
    */
   public static final SqlReturnTypeInference CHAR =
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -6222,6 +6222,43 @@ public class SqlOperatorTest {
     f.checkNull("log(10, cast(null as real))");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6224">[CALCITE-6224]
+   * Add LOG2 function (enabled in MYSQL, Spark library)</a>. */
+  @Test void testLog2Func() {
+    final SqlOperatorFixture f0 = fixture();
+    f0.checkFails("^log2(4)^",
+        "No match found for function signature LOG2\\(<NUMERIC>\\)", false);
+    f0.setFor(SqlLibraryOperators.LOG2);
+    final Consumer<SqlOperatorFixture> consumer = f -> {
+      f.checkScalarApprox("log2(2)", "DOUBLE",
+          isWithin(1.0, 0.000001));
+      f.checkScalarApprox("log2(4)", "DOUBLE",
+          isWithin(2.0, 0.000001));
+      f.checkScalarApprox("log2(65536)", "DOUBLE",
+          isWithin(16.0, 0.000001));
+      f.checkScalarApprox("log2(2.0/3)", "DOUBLE",
+          isWithin(-0.5849625007211561, 0.000001));
+      f.checkScalarApprox("log2(4.0/3)", "DOUBLE",
+          isWithin(0.4150374992788435, 0.000001));
+      f.checkScalarApprox("log2(0.5)", "DOUBLE",
+          isWithin(-1.0, 0.000001));
+      f.checkScalarApprox("log2(cast(10e8 as double))", "DOUBLE",
+          isWithin(29.897352853986263, 0.000001));
+      f.checkScalarApprox("log2(cast(10e8 as float))", "DOUBLE",
+          isWithin(29.897352853986263, 0.000001));
+      f.checkScalarApprox("log2(1e+52)", "DOUBLE",
+          isWithin(172.74026093414284, 0.000001));
+      f.checkNull("log2(0)");
+      f.checkNull("log2(-2)");
+      f.checkNull("log2(+0.0)");
+      f.checkNull("log2(-0.0)");
+      f.checkNull("log2(null)");
+      f.checkNull("log2(cast(null as real))");
+    };
+    f0.forEachLibrary(list(SqlLibrary.MYSQL, SqlLibrary.SPARK), consumer);
+  }
+
   @Test void testRandFunc() {
     final SqlOperatorFixture f = fixture();
     f.setFor(SqlStdOperatorTable.RAND, VmName.EXPAND);
--- a/core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
+++ b/core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
@@ -510,6 +510,7 @@ public enum BuiltInMethod {
   SAFE_MULTIPLY(SqlFunctions.class, "safeMultiply", double.class, double.class),
   SAFE_SUBTRACT(SqlFunctions.class, "safeSubtract", double.class, double.class),
   LOG(SqlFunctions.class, "log", long.class, long.class),
+  LOG2(SqlFunctions.class, "log2", long.class),
   SEC(SqlFunctions.class, "sec", double.class),
   SECH(SqlFunctions.class, "sech", double.class),
   SIGN(SqlFunctions.class, "sign", long.class),
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2783,6 +2783,7 @@ In the following:
 | b f s | LENGTH(string)                             | Equivalent to `CHAR_LENGTH(string)`
 | h s | LEVENSHTEIN(string1, string2)                | Returns the Levenshtein distance between *string1* and *string2*
 | b | LOG(numeric1 [, numeric2 ])                    | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present
+| m s | LOG2(numeric)                                | Returns the base 2 logarithm of *numeric*
 | b o s | LPAD(string, length [, pattern ])          | Returns a string or bytes value that consists of *string* prepended to *length* with *pattern*
 | b | TO_BASE32(string)                              | Converts the *string* to base-32 encoded form and returns an encoded string
 | b | FROM_BASE32(string)                            | Returns the decoded result of a base-32 *string* as a string


caicancai:
Regarding the case where the parameter is 0, it will be fixed in another jira
discussion: https://issues.apache.org/jira/browse/CALCITE-6259

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2149,6 +2149,18 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
+  /** The "LOG2(value)" function.
+   *
+   * @see SqlStdOperatorTable#LN
+   * @see SqlStdOperatorTable#LOG10
+   */
+  @LibraryOperator(libraries = {MYSQL, SPARK})
+  public static final SqlFunction LOG2 =
+      SqlBasicFunction.create("LOG2",
+          ReturnTypes.DOUBLE_NULLABLE,
+          OperandTypes.NUMERIC,
```
mihaibudiu, line 2161: Do there exist different functions for each numeric types or is there just one function for the DOUBLE type?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6218,10 +6226,43 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
         isWithin(9.0, 0.000001));
     f.checkScalarApprox("log(cast(10e-3 as real), 10)", "DOUBLE NOT NULL",
         isWithin(-2.0, 0.000001));
+    f.checkScalarApprox("log(2.0/3, 100)", "DOUBLE NOT NULL",
+        isWithin(-0.08804562952784059, 0.000001));
+    f.checkScalarApprox("log(0.5, 2)", "DOUBLE NOT NULL",
+        isWithin(-1.0, 0.000001));
```
mihaibudiu, line 6232: no tests for negative/zero?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6193,6 +6193,14 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
         isWithin(3.0, 0.000001));
     f.checkScalarApprox("log10(cast(10e-3 as real))", "DOUBLE NOT NULL",
         isWithin(-2.0, 0.000001));
+    f.checkScalarApprox("log10(2.0/3)", "DOUBLE NOT NULL",
+        isWithin(-0.17609125905568118, 0.000001));
+    f.checkScalarApprox("log10(0.5)", "DOUBLE NOT NULL",
+        isWithin(-0.30102999566398114, 0.000001));
+    f.checkScalarApprox("log10(100.0/3)", "DOUBLE NOT NULL",
+        isWithin(1.5228787452803372, 0.000001));
+    f.checkScalarApprox("log10(-1)", "DOUBLE NOT NULL",
```
mihaibudiu, line 6202: no test for 0?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6218,10 +6226,43 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+          isWithin(16.0, 0.000001));
+      f.checkScalarApprox("log2(2.0/3)", "DOUBLE NOT NULL",
+          isWithin(-0.5849625007211561, 0.000001));
+      f.checkScalarApprox("log2(4.0/3)", "DOUBLE NOT NULL",
+          isWithin(0.4150374992788435, 0.000001));
+      f.checkScalarApprox("log2(0.5)", "DOUBLE NOT NULL",
+          isWithin(-1.0, 0.000001));
+      f.checkScalarApprox("log2(-2)", "DOUBLE NOT NULL",
+          "NaN");
+      f.checkNull("log2(cast(null as real))");
```
mihaibudiu, line 6261: how about a test for 0

mihaibudiu: At least in MySQL 8.0 LOG2(0) returns NULL. So this does not seem to be an accurate implementation.

caicancai: > At least in MySQL 8.0 LOG2(0) returns NULL. So this does not seem to be an accurate implementation.

I will do the fix in another jira, I don't think it is appropriate to fix it in this jira.
https://issues.apache.org/jira/browse/CALCITE-6259

mihaibudiu: If you are implementing LOG2 as defined by MySQL you have no choice.
Why would we merge an incorrect implementation?

caicancai: OK, let me try to solve another jira and open this pr again. Thank you for your patient reply.

caicancai: done

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: Regarding log10 and log functions, I will not fix them in this jira, because the jira is only for log2 functions

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: I don't think there is any problem in keeping it the same as log10. I added test to explain it

caicancai:
@mihaibudiu @tanclary I added a function to log2 in Sqlfunctions, I can't think of a better solution, if you have, please let me know, I would be very grateful

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2788,6 +2788,18 @@ public static double log(BigDecimal d0, BigDecimal d1) {
...
 
+  /** SQL {@code LOG2(number)} function applied to double values. */
+  public static double log2(double d0) {
+    return (d0 == 0) ? Double.NaN : Math.log(d0) / Math.log(2);
+  }
+
+  /** SQL {@code LOG2(number)} function applied to
+   * BigDecimal and double values. */
+  public static double log2(BigDecimal d0) {
+    return (d0.doubleValue() == 0) ? Double.NaN : Math.log(d0.doubleValue()) / Math.log(2);
```
mihaibudiu, line 2799: just call the other function.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2788,6 +2788,18 @@ public static double log(BigDecimal d0, BigDecimal d1) {
     return Math.log(d0.doubleValue()) / Math.log(d1.doubleValue());
   }
 
+  /** SQL {@code LOG2(number)} function applied to double values. */
+  public static double log2(double d0) {
+    return (d0 == 0) ? Double.NaN : Math.log(d0) / Math.log(2);
```
mihaibudiu, line 2793: In general testing for equality FP values does not work.
There are also two kinds of zeros in FP, +0 and -0.
If the result from Math.log(0) is -Infinity, the more robust way to do this is to actually compute the log and check when the result is the negative infinity.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6222,6 +6222,41 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+          isWithin(-1.0, 0.000001));
+      f.checkScalarApprox("log2(cast(10e8 as double))", "DOUBLE NOT NULL",
+          isWithin(29.897352853986263, 0.000001));
+      f.checkScalarApprox("log2(cast(10e8 as float))", "DOUBLE NOT NULL",
+          isWithin(29.897352853986263, 0.000001));
+      f.checkScalarApprox("log2(-2)", "DOUBLE NOT NULL",
+          "NaN");
+      f.checkScalarApprox("log2(0)", "DOUBLE NOT NULL",
+          "NaN");
+      f.checkNull("log2(cast(null as real))");
```
mihaibudiu, line 6255: how about a test with just NULL, without the cast? that should work.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: Thank you, learned it

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: done, thanks

mihaibudiu: done perhaps, but I don't see a new commit yet

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: done，thank you

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2790,13 +2790,13 @@ public static double log(BigDecimal d0, BigDecimal d1) {
 
   /** SQL {@code LOG2(number)} function applied to double values. */
   public static double log2(double d0) {
-    return (d0 == 0) ? Double.NaN : Math.log(d0) / Math.log(2);
+    return (Double.isInfinite(log(d0, 2))) ? Double.NaN : log(d0, 2);
```
mihaibudiu, line 2793: I don't think this will work for the case of +infinity.
You should add a test case for `log2(CAST "+Infinity' AS NULL)`. I expect that the result will be NaN, whereas you probably expect Infinity in that case. So you should also check that the infinity is negative here.

mihaibudiu: that should work using < 0.0.

caicancai: > I don't think this will work for the case of +infinity. You should add a test case for `log2(CAST "+Infinity' AS NULL)`. I expect that the result will be NaN, whereas you probably expect Infinity in that case. So you should also check that the infinity is negative here.

Maybe I should check how mysql and spark work

caicancai: > that should work using < 0.0.

The result of log2(2.0/3) is also less than 0.0，

mihaibudiu: `Double.isInfinite(d0) && d0 < 0`

caicancai: Thank you

caicancai: wait， Double.isInfinite(d0) && d0 < 0 ， I think you mean Double.isInfinite(log(d0, 2)) && d0.doubleValue() < 0.0

mihaibudiu: just make sure it matches MySQL on all corner cases

mihaibudiu: cast as double

caicancai: Mysql seems to be unable to support the parsing of +Infinity and -Infinity. From a mathematical point of view, I am not sure whether this makes sense, but mysql obviously does not support it. Maybe we can only deal with 0 specifically. I tested +0.0 and - 0.0 returns null

mihaibudiu: you can try a very large value, such as 1e+52
https://dev.mysql.com/doc/refman/8.0/en/problems-with-float.html

caicancai: Thank you very much for answering my question so patiently

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6222,6 +6222,49 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+          isWithin(-0.5849625007211561, 0.000001));
+      f.checkScalarApprox("log2(4.0/3)", "DOUBLE NOT NULL",
+          isWithin(0.4150374992788435, 0.000001));
+      f.checkScalarApprox("log2(0.5)", "DOUBLE NOT NULL",
+          isWithin(-1.0, 0.000001));
+      f.checkScalarApprox("log2(cast(10e8 as double))", "DOUBLE NOT NULL",
+          isWithin(29.897352853986263, 0.000001));
+      f.checkScalarApprox("log2(cast(10e8 as float))", "DOUBLE NOT NULL",
+          isWithin(29.897352853986263, 0.000001));
+      f.checkScalarApprox("log2(1e+52)", "DOUBLE NOT NULL",
```
caicancai, line 6250: I added all the negative tests I could think of. At present, the test results are consistent with mysql.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2788,6 +2788,18 @@ public static double log(BigDecimal d0, BigDecimal d1) {
     return Math.log(d0.doubleValue()) / Math.log(d1.doubleValue());
   }
 
+  /** SQL {@code LOG2(number)} function applied to double values. */
+  public static double log2(double d0) {
+    return (Double.isInfinite(log(d0, 2)) && d0 == 0.0) ? Double.NaN : log(d0, 2);
```
caicancai, line 2793: Because - infinity and + infinity are special cases in SQL and do not support parsing, so I added the condition here that d0 == 0. At present, only the log2 (0) test is controversial.

tanclary: What do you mean they don't support parsing?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -637,6 +638,7 @@ Builder populate() {
       defineMethod(EXP, BuiltInMethod.EXP.method, NullPolicy.STRICT);
       defineMethod(POWER, BuiltInMethod.POWER.method, NullPolicy.STRICT);
       defineMethod(ABS, BuiltInMethod.ABS.method, NullPolicy.STRICT);
+      defineMethod(LOG2, BuiltInMethod.LOG2.method, NullPolicy.STRICT);
```
tanclary, line 641: Should this use the LogImplementor defined below?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2764,7 +2764,7 @@ public static double power(BigDecimal b0, BigDecimal b1) {
   }
 
 
-  // LN, LOG, LOG10
+  // LN, LOG, LOG2, LOG10
```
tanclary, line 2767: nit: LOG2 goes after LOG10

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2788,6 +2788,18 @@ public static double log(BigDecimal d0, BigDecimal d1) {
...
 
+  /** SQL {@code LOG2(number)} function applied to double values. */
+  public static double log2(double d0) {
+    return (Double.isInfinite(log(d0, 2)) && d0 == 0.0) ? Double.NaN : log(d0, 2);
+  }
+
+  /** SQL {@code LOG2(number)} function applied to
+   * BigDecimal and double values. */
+  public static double log2(BigDecimal d0) {
+    return (Double.isInfinite(log(d0, 2)) && d0.doubleValue() == 0.0) ? Double.NaN : log(d0, 2);
```
tanclary, line 2799: Is `NaN` definitely what we want to return?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6222,6 +6222,49 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6224">[CALCITE-6224]
+   * Add LOG2 function (enabled in MYSQL, Spark library)</a>. */
+  @Test void testLog2Func() {
+    final SqlOperatorFixture f0 = fixture();
+    f0.checkFails("^log2(4)^",
+        "No match found for function signature LOG2\\(<NUMERIC>\\)", false);
+    final Consumer<SqlOperatorFixture> consumer = f -> {
+      f.setFor(SqlLibraryOperators.LOG2);
```
tanclary, line 6233: You can set this outside of the consumer for f0 no?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: Which of these are negative tests?

tanclary (changes requested):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: At present, we mainly cannot return correct results for log2(0), log2(0.0), etc. In mysql, these return null. I think this corresponds to NaN.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: Log2(+Infinity) and log2(-Infinity) will report errors in mysql and cannot display the results. Calcite will also report errors in runtime.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: I tried it, but apparently it's not a good idea because mysql and postgres handle the log10(0) function inconsistently

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
mihaibudiu: if you are emulating MySQL you should return the exact same value in this case. Otherwise your code is wrong.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
mihaibudiu: I don't understand what you mean. For each function you have to return the same results as the corresponding function in MySQL.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: log2(1e+52) mainly wants to simulate the situation of log2(-Infinity), as introduced on the mysql official website

caicancai: [mihaibudiu](https://github.com/mihaibudiu) As mihaibudiu mentioned, I was worried that +0.0 and -0.0 might be different in mysql, so I also tested it in mysql and returned null. I also added the corresponding test in calcite.

mihaibudiu: 1e+52 should be converted to +infinity.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: yes, so I didn't put it in the LogImplementor defined below

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: I'm not sure whether 1e+52 will be converted into infinity in mysql, but as far as the results are concerned, currently the result of log2(1e+52) corresponds to mysql

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: If I put it in LogImplementor, I might have to do more work than now, because the log function in postgres and mysql obviously handles 0 differently, so I raised jira on this issue.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: I want to simulate infinity through some methods, such as pow(0,3) but this obviously fails in mysql

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: I may not understand what you mean. I wrote this based on how other functions are written.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: Maybe what you said makes sense, but I don't plan to solve it in this jira, because this may not only be a problem with the log function, but many Math functions may have similar problems. I will check the practices of other databases and mention one. Jira or email to discuss it

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: I don't know the reason for doing this either.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: I tested mysql's log(+infinity) and returned NULL. .Postgres's log(+infinity) returns infinity, which will report an error in calcite. It seems to be a bug in calcite.
```sql
select log10('-Infinity');
```

caicancai: It seems that calcite cannot convert infinity into the correct java type

caicancai: I think the infinity type should not be fixed in this jira. We discussed a lot about adapting the log2 function and I learned a lot. But at present, infinite adaptation does not seem to be a simple problem, or even just a problem of the log function. I will send an email or jira to discuss it.

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6277

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: You can set the operator outside of the consumer. There's no point resetting it each time.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: So it should return NULL?

tanclary: What does MySql return? Does it return `Null`? `NaN` is not the same as `null`. As Mihai said: `if you are emulating MySQL you should return the exact same value in this case. Otherwise your code is wrong`.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: I try to change the result to null in sqlfunctions, but a null pointer exception will be reported when comparingResultSet is checked.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: I try to change the result to null in sqlfunctions, but a null pointer exception will be reported when comparingResultSet is checked.

caicancai: comparingResultSet looks like it doesn't allow for nulls when iterating over the result set

caicancai: To a certain extent, I agree with your point of view, but if it is just for the case of log2(0)==null that a lot of calcite code needs to be changed, I am skeptical. I can even suggest that our handling of log2(0) should be stated in the document, and I feel that mysql handles this wrongly from the beginning. I studied presto and arrow-datafusion, and to be honest, they couldn't overcome this problem very well.

mihaibudiu: @caicancai if you think MySQL is wrong you should file a bug with MySQL.

caicancai: @mihaibudiu Maybe you are right, I would rather try to fix calcite than modify mysql, but I am not sure whether to modify the comparingResultSet method, because obviously we found a lot of problems when adapting the log2 function, and the problems are getting more and more complex.
Thank you

tanclary: Of course Julian can give his opinion but I'm confused what the issue is. MySQL has some behavior and we want to mimic that behavior. What's keeping us from doing this?

caicancai: Maybe I'm not sure how to modify it correctly and how to modify it appropriately. I hope to adapt it with minimal changes. At the moment, I'm not sure whether it's appropriate to modify comparingResultSet. I will try to find the best solution.

caicancai: @tanclary To be honest, I could have completed the changes and handed them over to you for review, but during the adaptation process, I found that I had a lot of doubts about this, which led me to ask you many questions. Perhaps this is the reason why this PR has been discussed for so long. It is very Thank you for your patient answers.

tanclary: If you upload your latest commit, and tell me what test to run, I can help.

caicancai: Thank you very much, just need rust test

caicancai:
I thought about it for a long time, maybe I think I already understand the doubts in my heart.
As @mihaibudiu  said, java syntax has its own type rules, and sql has its own type rules. What calcite is currently doing is to use java syntax to adapt to the sql rules of each database and complete execution optimization. The SQL rules of each database are inconsistent in some extreme scenarios. Calcite needs to ensure that these extreme scenarios are all adapted. However, if on a certain day, for example, mysql returns the result of log10(0) as an error instead of null, then does calcite need to adapt to the new version of mysql? If it adapts to the new version of mysql, then does the old version of mysql need to adapt to the new version of mysql? Need to adapt. This may be a paradox in my opinion. Can you please help me explain it?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: 捕获
@tanclary I'm not sure if this is a problem with calcite, but it seems like I have no idea about it.

caicancai: SAFE_* functions doesn't look like this will happen

caicancai:
@tanclary Thank you very much, let's wait for the cli

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2788,6 +2788,17 @@ public static double log(BigDecimal d0, BigDecimal d1) {
     return Math.log(d0.doubleValue()) / Math.log(d1.doubleValue());
   }
 
+  /** SQL {@code LOG2(number)} function applied to double values. */
+  public static @Nullable Double log2(double d0) {
+    return (d0 < 0.0 || d0 == 0.0) ? null : log(d0, 2);
```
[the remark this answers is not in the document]

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2788,6 +2788,17 @@ public static double log(BigDecimal d0, BigDecimal d1) {
     return Math.log(d0.doubleValue()) / Math.log(d1.doubleValue());
   }
 
+  /** SQL {@code LOG2(number)} function applied to double values. */
```
tanclary, line 2791: in the comment you call the parameter "number" but in the actual method you call it "d0", these should be consistent.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2788,6 +2788,17 @@ public static double log(BigDecimal d0, BigDecimal d1) {
...
+  public static @Nullable Double log2(double d0) {
+    return (d0 <= 0) ? null : log(d0, 2);
+  }
+
+  /** SQL {@code LOG2(number)} function applied to
```
tanclary, line 2796: same comment as above

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2149,6 +2149,18 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
           OperandTypes.NUMERIC_OPTIONAL_NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
+  /** The "LOG2(value)" function.
+   *
+   * @see SqlStdOperatorTable#LN
```
tanclary, line 2154: Why "@see" these other functions? They're not the same. I think a brief description would be more helpful.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -396,6 +396,13 @@ public static SqlCall stripSeparator(SqlCall call) {
   public static final SqlReturnTypeInference DOUBLE_NULLABLE =
       DOUBLE.andThen(SqlTypeTransforms.TO_NULLABLE);
 
+  /**
+   * Type-inference strategy whereby the result type of a call is a nullable
+   * Double.
```
tanclary, line 401: Going forward if someone writes some code for you it's good practice to add them as a co-author. Don't worry about it for me but other people may want their work attributed to them.

tanclary (changes requested):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: I like number

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: done

caicancai: Sorry, not done

caicancai: done

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2149,6 +2149,14 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
           OperandTypes.NUMERIC_OPTIONAL_NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
+  /** The "LOG2(value)" function. Returns the base 2 logarithm of numeric. */
```
tanclary, line 2152: Your comment is inconsistent. You said "value" as the parameter name and then later called it "numeric"

tanclary:
I think when you get a review, you should try to make sure you didn't make the same mistakes in other places in the PR. If you fix them all at once it will save you and the reviewer time. Especially when your reviewers are in different time zones, it will help reduce the time it takes to merge the commit.

caicancai:
> I think when you get a review, you should try to make sure you didn't make the same mistakes in other places in the PR. If you fix them all at once it will save you and the reviewer time. Especially when your reviewers are in different time zones, it will help reduce the time it takes to merge the commit.

Maybe what you said makes sense, but the commits that were merged before were written in a similar way. I learned from their writing methods. I don't understand why the previous PR can be merged if I modified it. I don't mean any harm, I just don't understand

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: the last time

tanclary:
> > I think when you get a review, you should try to make sure you didn't make the same mistakes in other places in the PR. If you fix them all at once it will save you and the reviewer time. Especially when your reviewers are in different time zones, it will help reduce the time it takes to merge the commit.
> 
> Maybe what you said makes sense, but the commits that were merged before were written in a similar way. I learned from their writing methods. I don't understand why the previous PR can be merged if I modified it. I don't mean any harm, I just don't understand

When someone writes code you use in your commit, it is polite to add them as a co-author. You used some code I wrote in this commit. Does that make sense?

caicancai:
> > > I think when you get a review, you should try to make sure you didn't make the same mistakes in other places in the PR. If you fix them all at once it will save you and the reviewer time. Especially when your reviewers are in different time zones, it will help reduce the time it takes to merge the commit.
> > 
> > 
> > Maybe what you said makes sense, but the commits that were merged before were written in a similar way. I learned from their writing methods. I don't understand why the previous PR can be merged if I modified it. I don't mean any harm, I just don't understand
> 
> I mean that I've noticed you make the same mistakes multiple times, even after they have been pointed out by a reviewer. You can avoid this by checking over the PR and making sure you fix all instances of a mistake so it saves everyone time. Does that make sense?



> > > I think when you get a review, you should try to make sure you didn't make the same mistakes in other places in the PR. If you fix them all at once it will save you and the reviewer time. Especially when your reviewers are in different time zones, it will help reduce the time it takes to merge the commit.
> > 
> > 
> > Maybe what you said makes sense, but the commits that were merged before were written in a similar way. I learned from their writing methods. I don't understand why the previous PR can be merged if I modified it. I don't mean any harm, I just don't understand
> 
> I mean that I've noticed you make the same mistakes multiple times, even after they have been pointed out by a reviewer. You can avoid this by checking over the PR and making sure you fix all instances of a mistake so it saves everyone time. Does that make sense?

My question, sorry

PR #3690, opened 2024-02-16
[CALCITE-6266] Parser should be able to parse cross join with left lateral

snuyanzin:


jeyhunkarimov (changes requested):
Thanks for the PR @snuyanzin . 
I have a few comments.

From the Calcite's parenthesized sub-clause in JOIN PR ([CALCITE-35](https://github.com/apache/calcite/commit/63a15128a332af7a641b26a3224392960882170a)) also from the `Parser.jj` file in the `main` branch, I see:

```
        // Comma joins should only occur at top-level in the FROM clause.
        // Valid:
        //  * FROM a, b
        //  * FROM (a CROSS JOIN b), c
        // Not valid:
        //  * FROM a CROSS JOIN (b, c)

```

So, this query - `select * from dept cross join  ((VALUES ('A'), ('B')), (VALUES ('A'), ('B')))` should not pass (and it fails in the `main` branch). However, with this PR it passes. 

Also, this query - `select * from dept cross join ( lateral table(ramp(dept.deptno)) cross join (VALUES ('A'), ('B')) )` also fails with the exception `Table 'DEPT' not found`. We might need to investigate if this is an expected behaviour.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -3606,9 +3606,7 @@ void checkPeriodPredicate(Checker checker) {
   @Test void testMixedFrom() {
     sql("select * from a join b using (x), c join d using (y)")
         .ok("SELECT *\n"
-            + "FROM `A`\n"
```
JiajunBernoulli, line 3609: Why change it?

snuyanzin:
Closed in favor of https://github.com/apache/calcite/pull/3871

PR #3694, opened 2024-02-16
[CALCITE-6255] Support BigQuery-style JSON_OBJECT invocation syntax

barrkel:
[CALCITE-6255](https://issues.apache.org/jira/browse/CALCITE-6255)

Currently, Calcite has two styles for invoking `JSON_OBJECT`:

```
    JSON_OBJECT(KEY 'foo' VALUE 'bar', KEY 'foo2' VALUE 'bar2')
```

and:

```
    JSON_OBJECT('foo' : 'bar', 'foo2' : 'bar')
```

However, [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#json_object) uses a lighter-weight syntax resembling the MAP<> constructor, with an even numbered argument list where every odd item is a key and the following even item is a value:

```
    JSON_OBJECT('foo', 'bar', 'foo2', 'bar2')
```

This PR updates the grammar to allow `,` as well as `:` for separating key/value pairs in the invocation.

It does not prevent mixing `,` and `:`, but retains the prohibition from mixing with `KEY`/`VALUE` syntax.

### DIFF
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -6732,6 +6732,12 @@ List<SqlNode> JsonNameAndValue() :
     (
         <VALUE>
     |
+        <COMMA> {
+            if (kvMode) {
+                throw SqlUtil.newContextException(getPos(), RESOURCE.illegalComma());
+            }
+        }
+    |
         <COLON> {
             if (kvMode) {
                 throw SqlUtil.newContextException(getPos(), RESOURCE.illegalColon());
--- a/core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
+++ b/core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
@@ -109,8 +109,11 @@ public interface CalciteResource {
   @BaseMessage("ROW expression encountered in illegal context")
   ExInst<CalciteException> illegalRowExpression();
 
-  @BaseMessage("Illegal identifier '':''. Was expecting ''VALUE''")
+  @BaseMessage("Unexpected symbol '':''. Was expecting ''VALUE''")
   ExInst<CalciteException> illegalColon();
+
+  @BaseMessage("Unexpected symbol '',''. Was expecting ''VALUE''")
+  ExInst<CalciteException> illegalComma();
 
   @BaseMessage("TABLESAMPLE percentage must be between 0 and 100, inclusive")
   @Property(name = "SQLSTATE", value = "2202H")
--- a/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
+++ b/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
@@ -8815,6 +8815,16 @@ public class SqlParserTest {
         .ok("JSON_OBJECT(KEY 'foo' VALUE "
             + "JSON_OBJECT(KEY 'foo' VALUE 'bar' NULL ON NULL) "
             + "FORMAT JSON NULL ON NULL)");
+    expr("json_object('foo', 'bar')")
+        .ok("JSON_OBJECT(KEY 'foo' VALUE 'bar' NULL ON NULL)");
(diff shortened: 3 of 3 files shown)


core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
```diff
@@ -112,6 +112,9 @@ public interface CalciteResource {
   @BaseMessage("Illegal identifier '':''. Was expecting ''VALUE''")
   ExInst<CalciteException> illegalColon();
 
+  @BaseMessage("Illegal identifier '',''. Was expecting ''VALUE''")
```
mihaibudiu, line 115: Comma is not really an identifier.  Maybe "symbol".
Yes, the previous message is probably wrong too.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -8815,6 +8815,10 @@ private static Consumer<List<? extends Throwable>> checkWarnings(
         .ok("JSON_OBJECT(KEY 'foo' VALUE "
             + "JSON_OBJECT(KEY 'foo' VALUE 'bar' NULL ON NULL) "
             + "FORMAT JSON NULL ON NULL)");
+    expr("json_object('foo', 'bar')")
```
mihaibudiu, line 8818: how about some other tests, e.g., with an odd number of elements?

core/src/main/codegen/templates/Parser.jj
```diff
@@ -6767,7 +6773,8 @@ SqlCall JsonObjectFunctionCall() :
 }
 {
     <JSON_OBJECT> { span = span(); }
-    <LPAREN> [
+    <LPAREN>
```
mihaibudiu, line 6776: why this change?

tanclary (commented):
format the PR title and commit so that they begin with `[CALCITE-6255]` 

otherwise LGTM

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
tanclary: It's kind of a delimiter too? Wdyt

mihaibudiu: delimiter is also fine

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
JiajunBernoulli: How about mixed syntax tests? e.g.
```
json_object('foo': json_object('foo', 'bar') format json)
```

core/src/main/codegen/templates/Parser.jj
barrkel: Stray edit from an earlier version of the code. I'll revert.

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
barrkel: Well `,` isn't a legal identifier - the message is not wrong. It's kind of a two-prong message; we were expecting an identifier, and for the identifier to be `VALUE`. We fell at the first hurdle because `,` isn't an identifier.

(AFAICT `VALUE` is usually a keyword not an identifier but it's a bit squishy because it could easily be a non-reserved word.)

A better message might be "Unexpected symbol" rather than "Illegal identifier", which has better symmetry with the expected clause following. I'll change both to that.

tanclary (approved):
LGTM, will merge soon if no other comments.

macroguo-ghy (approved):
+1

macroguo-ghy:
I think you should make the Jira ticket title, GitHub PR title, and commit message title consistent.

barrkel:
> I think you should make the Jira ticket title, GitHub PR title, and commit message title consistent.

Done.

tanclary (approved):
LGTM

PR #3696, opened 2024-02-20
[CALCITE-6274] Two Elasticsearch index join return empty result

zoov-w:
[CALCITE-6274](https://issues.apache.org/jira/browse/CALCITE-6274)
Two index of Elasticsearch join return empty result even if the data from both indexes can match。

create index test_01:
`
PUT /test_01/_doc/1
{
  "doc_id" : 1,
  "doc_desc" : "doc01"
} 
`

create index test_02:
`
PUT /test_02/_doc/1
{
  "doc_id" : 1,
  "doc_score" : 90
} 
`

execute sql:
`
select * from es.test_01 t1 join es.test_02 t2 on cast(t1._MAP['doc_id'] as bigint) = cast(t2._MAP['doc_id'] as bigint) 
`

the code generate by ElasticsearchToEnumerableConverter like this:
subquery of index test_01:
`
{
  return ((org.apache.calcite.adapter.elasticsearch.ElasticsearchTable.ElasticsearchQueryable) org.apache.calcite.schema.Schemas.queryable(root, root.getRootSchema().getSubSchema("es"), java.lang.Object[].class, "test_01")).find(java.util.Collections.EMPTY_LIST, java.util.Arrays.asList(new org.apache.calcite.util.Pair[] {
      new org.apache.calcite.util.Pair(
        "_MAP",
        java.util.Map.class),
      new org.apache.calcite.util.Pair(
        "_1",
        java.lang.Long.class)}), java.util.Arrays.asList(new org.apache.calcite.util.Pair[] {
      new org.apache.calcite.util.Pair(
        "doc_id",
        org.apache.calcite.rel.RelFieldCollation.Direction.ASCENDING)}), java.util.Collections.EMPTY_LIST, java.util.Arrays.asList(new org.apache.calcite.util.Pair[] {}), com.google.common.collect.ImmutableMap.of("$f1", "doc_id"), null, null);
}
`
project field names: `"_MAP", "_1"`

subquery of index test_02:
`
{
  return ((org.apache.calcite.adapter.elasticsearch.ElasticsearchTable.ElasticsearchQueryable) org.apache.calcite.schema.Schemas.queryable(root, root.getRootSchema().getSubSchema("es"), java.lang.Object[].class, "test_02")).find(java.util.Collections.EMPTY_LIST, java.util.Arrays.asList(new org.apache.calcite.util.Pair[] {
      new org.apache.calcite.util.Pair(
        "_MAP",
        java.util.Map.class),
      new org.apache.calcite.util.Pair(
        "_1",
        java.lang.Long.class)}), java.util.Arrays.asList(new org.apache.calcite.util.Pair[] {
      new org.apache.calcite.util.Pair(
        "doc_id",
        org.apache.calcite.rel.RelFieldCollation.Direction.ASCENDING)}), java.util.Collections.EMPTY_LIST, java.util.Arrays.asList(new org.apache.calcite.util.Pair[] {}), com.google.common.collect.ImmutableMap.of("$f1", "doc_id"), null, null);
}
`
project field names: `"_MAP", "_1"`

This org.apache.calcite.adapter.elasticsearch.ElasticsearchTable.ElasticsearchQueryable#find function actually execute request, subq-query result  projected according to second paramter fields. Field "_1" can not find from subq-query result.  "_1" not in mappings {"$f1":"doc_id"}, cause two sub-query join condition value is null, so the result of sql is empty.

This PR fix the problem, the sub-query project field names is: `"_MAP", "$f1"`. join condition field `$f1` can be find in mappings. The result of sql match expectations.

### DIFF
--- a/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchRules.java
+++ b/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchRules.java
@@ -113,8 +113,7 @@ class ElasticsearchRules {
     return SqlValidatorUtil.uniquify(
         new AbstractList<String>() {
           @Override public String get(int index) {
-            final String name = rowType.getFieldList().get(index).getName();
-            return name.startsWith("$") ? "_" + name.substring(2) : name;
+            return rowType.getFieldList().get(index).getName();
           }
 
           @Override public int size() {
--- a/elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/JoinTest.java
+++ b/elasticsearch/src/test/java/org/apache/calcite/adapter/elasticsearch/JoinTest.java
@@ -1,0 +1,120 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.adapter.elasticsearch;
+
+import org.apache.calcite.jdbc.CalciteConnection;
+import org.apache.calcite.schema.SchemaPlus;
+import org.apache.calcite.test.CalciteAssert;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.collect.ImmutableMap;
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.ResourceAccessMode;
+import org.junit.jupiter.api.parallel.ResourceLock;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * Testing Elasticsearch join query.
+ */
+@ResourceLock(value = "elasticsearch-scrolls", mode = ResourceAccessMode.READ)
+public class JoinTest {
+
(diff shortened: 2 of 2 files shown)


mihaibudiu:
The tools don't like the way you formatted your code; you can run ./gradlew build to find out what's wrong.

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchRules.java
```diff
@@ -113,8 +113,7 @@ static List<String> elasticsearchFieldNames(final RelDataType rowType) {
     return SqlValidatorUtil.uniquify(
         new AbstractList<String>() {
           @Override public String get(int index) {
-            final String name = rowType.getFieldList().get(index).getName();
-            return name.startsWith("$") ? "_" + name.substring(2) : name;
+            return rowType.getFieldList().get(index).getName();
```
rubenada, line 116: Before tearing down this chesterton fence... do we know what was the logic behind this piece of code that you propose to remove?

mihaibudiu:

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchRules.java
zoov-w: I think it's not works. This function similar to org.apache.calcite.adapter.mongodb.MongoRules#mongoFieldNames which commit-id is ce0eb9c4.

zoov-w: But elasticsearch adapter is different, In my above description,  when join condition push  down to project, _MAP['doc_id'] field name is "$f1",  correspond to "doc_id" field in _source of elasticsearch index. So the value of  fields is ["$f1"],  mappings values is  {"$f1": "doc_id"}, fields and mappings are paramter of   ElasticsearchTable.ElasticsearchQueryable#find.

Function elasticsearchFieldNames called by ElasticsearchToEnumerableConverter transform "$f1" of fields to "_1"  but dose not transform key of mappings. so column value of "$f1" is null.

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchRules.java
```diff
@@ -113,8 +113,7 @@ static List<String> elasticsearchFieldNames(final RelDataType rowType) {
     return SqlValidatorUtil.uniquify(
         new AbstractList<String>() {
           @Override public String get(int index) {
-            final String name = rowType.getFieldList().get(index).getName();
-            return name.startsWith("$") ? "_" + name.substring(2) : name;
```
xuzifu666, line 117: I had test join operation of elasticsearch，it would return empty result. here name.substring(2) is not right which maybe copy from mongodb(mongodb is $_f1 and elasticsearch is $f1),and here not necessary to do this operation，return rowType.getFieldList().get(index).getName() directly is OK

xuzifu666:
Due to this pr is stale for one more year and it is necessary to be fixed, currently maybe can merge this pr or I start a new pr to continue it?  @mihaibudiu @NobiGo

xuzifu666 (approved):

mihaibudiu (approved):

PR #3697, opened 2024-02-20
[CALCITE-6262] CURRENT_TIMESTAMP(P) ignores DataTypeSystem#getMaxPrecision

jduo:
Change datetime functions taking in a precision to validate against the max precision from the type system, rather than the SqlTypeName.

### DIFF
--- a/core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
+++ b/core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
@@ -19,11 +19,18 @@ package org.apache.calcite.sql.type;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeSystemImpl;
+import org.apache.calcite.runtime.CalciteException;
+import org.apache.calcite.runtime.Resources;
+import org.apache.calcite.sql.SqlLiteral;
+import org.apache.calcite.sql.SqlOperatorBinding;
 import org.apache.calcite.sql.fun.SqlLibraryOperators;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.validate.SqlValidatorException;
 
 import com.google.common.collect.Lists;
 
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.junit.jupiter.api.Test;
 
 import static org.hamcrest.CoreMatchers.is;
@@ -36,9 +43,13 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 class RelDataTypeSystemTest {
 
   /**
-   * Custom type system class that overrides the default decimal plus type derivation.
+   * Custom type system class that overrides the default decimal plus type derivation and
+   * overrides the max precision for timestamps.
    */
   private static final class CustomTypeSystem extends RelDataTypeSystemImpl {
+    // Arbitrarily choose a different maximum timestamp precision from the default.
+    private static final int CUSTOM_MAX_TIMESTAMP_PRECISION =
+        SqlTypeName.MAX_DATETIME_PRECISION + 3;
 
     @Override public RelDataType deriveDecimalPlusType(RelDataTypeFactory typeFactory,
         RelDataType type1, RelDataType type2) {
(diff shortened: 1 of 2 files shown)


jduo:
@tanclary

core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
```diff
@@ -118,6 +126,13 @@ private static final class CustomTypeSystem extends RelDataTypeSystemImpl {
     @Override public int getMaxNumericPrecision() {
       return 38;
     }
+
+    @Override public int getMaxPrecision(SqlTypeName typeName) {
+      if (typeName == SqlTypeName.TIMESTAMP) {
+        return SqlTypeName.MAX_DATETIME_PRECISION + 3; // Arbitrarily choose a different precision.
```
tanclary, line 132: Where is the +3 coming from?

core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
```diff
@@ -218,4 +233,32 @@ static class Fixture extends SqlTypeFixture {
     assertEquals(28, dataType.getPrecision());
     assertEquals(10, dataType.getScale());
   }
+
+  @Test void testCustomMaxTimestampPrecisionTimeFunctionReturnTypeInference() {
```
tanclary, line 237: would be helpful to leave a comment with the jira case. there are lots of examples with `Test case for ...` you can use

core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
```diff
@@ -118,6 +126,13 @@ private static final class CustomTypeSystem extends RelDataTypeSystemImpl {
     @Override public int getMaxNumericPrecision() {
       return 38;
     }
+
+    @Override public int getMaxPrecision(SqlTypeName typeName) {
+      if (typeName == SqlTypeName.TIMESTAMP) {
```
tanclary, line 131: why not just have a `MAX_TIMESTAMP_PRECISION`? Am I wrong in saying that everytime we look for the max timestamp precision we're just retrieving the max datetime and then adding 3? In that case we should just define it as a constant and then this becomes unnecessary.

tanclary (changes requested):

core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
jduo: It's just a made-up number that differs from the default max.

core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
jduo: I named this constant CUSTOM_MAX_TIMESTAMP_PRECISION just to be explicit that it's testing a custom max.

core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
jduo: Done

jduo:
@tanclary , the linter fixes are done. Please let me know if there are any other concerns.

tanclary (approved):

core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
```diff
@@ -36,9 +43,13 @@
...
 
   /**
-   * Custom type system class that overrides the default decimal plus type derivation.
+   * Custom type system class that overrides the default decimal plus type derivation and
+   * overrides the max precision for timestamps.
    */
   private static final class CustomTypeSystem extends RelDataTypeSystemImpl {
+    // Arbitrarily choose a different maximum timestamp precision from the default.
+    private static final int CUSTOM_MAX_TIMESTAMP_PRECISION =
+        SqlTypeName.MAX_DATETIME_PRECISION + 3;
```
JiajunBernoulli, line 52: I don't understand this, I'm curious why it is + `3`?

tanclary: I think he just chose an arbitrary number

jduo: Yes, just the default offset by an arbitrary number (to ensure it is distinct from the default).

jduo:
Hi, just wondering if there's anything needed on my side to get this one merged.
Thanks for the reviews.

mihaibudiu:
If there are no comments before the end of the day I plan to merge this PR, since it seems to have an approval.

PR #3698, opened 2024-02-21
[CALCITE-6275] Parser for data types ignores element nullability in collections

mihaibudiu:
This modifies the parser to accept types such as INTEGER NOT NULL ARRAY NULL for column types.
The only thing I don't like too much about this PR is the fact that this type is unparsed differently depending on where it appears: 

- in a column definition it is unparsed as INTEGER NOT NULL ARRAY
- in a CAST it is unparsed as INTEGER ARRAY (the type can appear in a cast due to type coercion, so this is a legal use case)

These two not the same type at all, but other databases, such as Postgres, also do not accept types with nullabilities in casts. Maybe that's a bug in Postgres as well, or maybe in the SQL handling of collection types in general? 

In order to provide this behavior I had to add some state to the SqlWriter which indicates the context where a type is being unparsed. Fortunately, types cannot contain other kinds of SqlNodes - e.g., expressions - so a stack of context is not necessary, a scalar will do.

### DIFF
--- a/server/src/test/java/org/apache/calcite/test/ServerParserTest.java
+++ b/server/src/test/java/org/apache/calcite/test/ServerParserTest.java
@@ -41,7 +41,6 @@ import org.junit.jupiter.api.Test;
  *
  * <li>during CREATE VIEW, check for a table and a materialized view
  * with the same name (they have the same namespace)
- *
  * </ul>
  */
 class ServerParserTest extends SqlParserTest {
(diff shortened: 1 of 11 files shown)


server/src/test/java/org/apache/calcite/test/ServerParserTest.java
```diff
@@ -116,6 +115,30 @@ class ServerParserTest extends SqlParserTest {
...
+        + "c int not null array, "
+        + "d int not null array not null, "
+        + "e int array null, " // same as column a
+        + "f int null array null, " // same as column a
+        + "g int null array not null, " // same as column b
+        + "h int not null array null)") // same as column c
+        .ok("CREATE TABLE `X` ("
+            + "`A` INTEGER ARRAY, "
+            + "`B` INTEGER ARRAY NOT NULL, "
+            + "`C` INTEGER NOT NULL ARRAY, "
```
YiwenWu, line 134: Is  `INTEGER NOT NULL ARRAY`supported in Postgresql?

mihaibudiu: you are right, it does not seem to be supported.
I will discuss this topic in the JIRA.

mihaibudiu:
Since no other SQL dialects seem to accept this kind of syntax, I am closing this PR.
The spec does not say anything about the nullability of a column with ARRAY type in a CREATE TABLE statement.

PR #3705, opened 2024-02-26
[CALCITE-5976] Function ARRAY_PREPEND/ARRAY_APPEND/ARRAY_INSERT gives exception when inserted element type not equals array component type

caicancai:
https://issues.apache.org/jira/browse/CALCITE-5976

PR #3706, opened 2024-02-27
[CALCITE-6283] Function array_append with a NULL array argument crashes with NullPointerException

mihaibudiu:
…


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6268,6 +6268,18 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6283">
+   * [CALCITE-6283] Function array_append with a NULL array argument crashes with
+   * NullPointerException</a>. */
+  @Test void testArrayNullFunc() {
+    final SqlOperatorFixture f = fixture().withLibrary(SqlLibrary.SPARK);
+    f.checkNull("array_append(null, 2)");
+    f.checkNull("array_prepend(null, 2)");
+    f.checkNull("array_remove(null, 2)");
+    f.checkNull("array_contains(null, 2)");
```
tanclary, line 6279: Do we have tests for appending null elements

core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
```diff
@@ -45,23 +45,27 @@ public class ArrayElementOperandTypeChecker implements SqlOperandTypeChecker {
       return false;
     }
 
-    RelDataType arrayComponentType =
-        getComponentTypeOrThrow(SqlTypeUtil.deriveType(callBinding, op0));
+    RelDataType op0Type = SqlTypeUtil.deriveType(callBinding, op0);
```
tanclary, line 48: can this be final

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: @tanclary Yes, similar to:
 f.checkNull("array_append(cast(null as integer array), 1)");

caicancai:
@mihaibudiu I have also recently been fixing the bug of array_append, which seems to conflict with this PR.
https://github.com/apache/calcite/pull/3705

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: I meant a test where the element is null, not the array. But i found that we do so that's good.

caicancai: I tested it in spark3.4. It seems that the first parameter of array_append is null. Why do we need to change it like this?

caicancai: 捕获

tanclary: I don't understand your question

caicancai: My problem is that array_append is to adapt to spark, but spark’s array_append(null, 2) will report an error.

caicancai: What I mean is that spark3.4 version does not support array_append(null, 1), it will report an error. Currently, array_append and array_prepend are adapted for spark in calcite. If spark does not support array_append(null , 1), why do we need to return null? I am not sure whether higher versions of spark support it.
I don't know if my opinion is useful, if it is problematic, please ignore it. But I still want to express my opinion. This is open source, thank you.
捕获

tanclary: Ah. Yes I agree with you. What do you think @mihaibudiu ?

mihaibudiu: I have changed the type checker to reject NULL literals for arrays.
However, it seems to be that Spark is inconsistent, since it allows NULL values for arrays at runtime, but not at compile time. This is probably because Spark uses a different type checking/inference algorithm.

caicancai: > I have changed the type checker to reject NULL literals for arrays. However, it seems to be that Spark is inconsistent, since it allows NULL values for arrays at runtime, but not at compile time. This is probably because Spark uses a different type checking/inference algorithm.

spark sql, after generating the Unresolved Logical Plan, runs a series of rules to turn it into the Logical Plan, which solves some type conversion issues there, and the type handling of array_append(null,2) seems to be in that part

chucheng92:
pls also see https://issues.apache.org/jira/browse/CALCITE-5893 to support array_except/array_intersect.  they also should reject NULL literals.

chucheng92:
FYI: One background is that many spark functions related to calcite's array are contributed by one contributor. One basic strategy is to turn on NullPolicy, which means that NULL literals, including **NULL** and **cast(null as xxx)** are returned null. But in fact, spark will throw an exception in the former. We need to address the former case.

mihaibudiu:
The situation may be even more complicated if one considers cases where the compiler needs to insert implicit casts. I don't have a spark engine handy to test what Spark does, but if the array element types and the added element do not have the same type, Spark may choose to insert implicit casts in different ways than Calcite does. 

Another fact I don't have a good handle on is the nullability of array elements: https://issues.apache.org/jira/browse/CALCITE-6275

chucheng92:
array_append needs to be capitalized in commit name/jira name/pr name.

chucheng92:
> The situation may be even more complicated if one considers cases where the compiler needs to insert implicit casts. I don't have a spark engine handy to test what Spark does, but if the array element types and the added element do not have the same type, Spark may choose to insert implicit casts in different ways than Calcite does.
> 
> Another fact I don't have a good handle on is the nullability of array elements: https://issues.apache.org/jira/browse/CALCITE-6275

make sense. I just read it wrong. Your PR only handles the situation when array is null.

mihaibudiu:
> make sense. I just read it wrong. Your PR only handles the situation when array arg is NULL. My point is that the element is also a NULL literal.

There are already tests for this case, e.g.: `testArrayAppendFunc`

mihaibudiu:
> pls also see https://issues.apache.org/jira/browse/CALCITE-5893 to support array_except/array_intersect. they also should reject NULL literals.
> 
> calcite spark:
> 
> // return null select array_contains(array[1, 2], null); // return null select array_except(array[1, 2, 3], null) // return null select array_intersect(array[1,2,3], null)
> 
> they should reject NULL.

That's great, we should merge that PR if it solves the problem. I will try to review it.

mihaibudiu:
> array_append needs to be capitalized in commit name/jira name/pr name.

I will update these when I squash the commits

chucheng92:
> > make sense. I just read it wrong. Your PR only handles the situation when array arg is NULL. My point is that the element is also a NULL literal.
> 
> There are already tests for this case, e.g.: `testArrayAppendFunc`

If the first array arg is NULL, then it is an exception for all these array functions. There is no doubt about this, but if the second arg is NULL, the semantics are different. `array_contains/array_except/array_intersect` is currently wrong. And you are right to mention `array_append/array_prepend` as it does insert null.

image

yes, I have fixed `array_contains/array_except/array_intersect` in that PR, and that PR will affect this PR.

mihaibudiu:
Let's merge [CALCITE-5893] first and I will rebase this one.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6277,6 +6277,19 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+   * [CALCITE-6283] Function array_append with a NULL array argument crashes with
+   * NullPointerException</a>. */
+  @Test void testArrayNullFunc() {
+    final String expected = "Illegal use of 'NULL'";
+    final SqlOperatorFixture f = fixture().withLibrary(SqlLibrary.SPARK);
+    f.checkFails("^array_append(null, 2)^", expected, false);
+    f.checkFails("^array_prepend(null, 2)^", expected, false);
+    f.checkFails("^array_remove(null, 2)^", expected, false);
+    f.checkFails("^array_contains(null, 2)^", expected, false);
+    f.checkFails("^array_position(null, 2)^", expected, false);
```
snuyanzin, line 6290: It seems there is a similar issue for `array_min/array_max` should we fix it within this issue or create a separate one?

mihaibudiu: since I got started I will add these too

mihaibudiu:
In my last commit I have fixed array_min and array_max as well.
I have slightly modified the helper function introduced by @chucheng92 in https://github.com/apache/calcite/pull/3355 and used it for typechecking more functions.
I have also corrected a test for array_contains, according to the Databricks manual, that test was incorrect.
https://docs.databricks.com/en/sql/language-manual/functions/array_contains.html

core/src/main/java/org/apache/calcite/sql/type/ArrayElementOperandTypeChecker.java
```diff
@@ -34,61 +33,64 @@
...
+  public ArrayElementOperandTypeChecker(boolean arrayMayBeNull, boolean elementMayBeNull) {
+    this.arrayMayBeNull = arrayMayBeNull;
+    this.elementMayBeNull = elementMayBeNull;
   }
 
   //~ Methods ----------------------------------------------------------------
 
   @Override public boolean checkOperandTypes(
       SqlCallBinding callBinding,
       boolean throwOnFailure) {
```
chucheng92, line 50: If array is not first argument? such as array_xxx(1, array, 2). It seems that this typechecker only works in this situation.

chucheng92:
overall looks good to me +1

mihaibudiu:
Interpreting the +1 as a "go ahead and merge"

PR #3685, opened 2024-02-12
[CALCITE-6248] Illegal dates are accepted by casts

mihaibudiu:
This PR is a pair to https://github.com/apache/calcite-avatica/pull/238 and should be merged before the other one can pass.
It only changes a few date literals in tests to conform to the expectation of the SQL standard.
Together these two PRs constitute a fix to [CALCITE-6248]

PR #3708, opened 2024-02-29
[CALCITE-6248] Illegal dates are accepted by casts

mihaibudiu:
This is the second attempt to provide a partial fix for [CALCITE-6248].
The fix is partial, because the bug is in Avatica, but Avatica CI fails if this fix is not applied.
I have already merged a PR for this issue, but here I revert the changes in that prior PR.
I apologize for this change of mind, it has become apparent that the previous choice for the fix was incorrect.
The fix is about CAST(string to DATE). The assumption is that this cast should be more lenient than DATE literals in parsing the string, but should still require the date to be correct.
The plan is to merge this PR, then attempt to merge the corresponding Avatica PR: https://github.com/apache/calcite-avatica/pull/238. If that PR passes, it can be merged. Then some leftover code in Calcite can be cleaned-up when a new Calcite release picks up a new Avatica release.
Unfortunately I don't know of an easy way to test this PR before it's merged.

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -1273,7 +1273,6 @@ public class SqlOperatorTest {
     if (Bug.CALCITE_6282_FIXED) {
       f.checkScalar("cast('1945-02-24 12:42:25.34' as TIMESTAMP(2))",
           "1945-02-24 12:42:25.34", "TIMESTAMP(2) NOT NULL");
-
       if (castType == CastType.CAST) {
         f.checkFails("cast('1945-2-2 12:2:5' as TIMESTAMP)",
             "Invalid DATE value, '1945-2-2 12:2:5'", true);
@@ -1305,7 +1304,7 @@ public class SqlOperatorTest {
     f.checkCastToString("DATE '1945-2-24'", null, "1945-02-24", castType);
 
     f.checkScalar("cast('1945-02-24' as DATE)", "1945-02-24", "DATE NOT NULL");
-    f.checkScalar("cast(' 1945-02-04 ' as DATE)", "1945-02-04", "DATE NOT NULL");
+    f.checkScalar("cast(' 1945-2-4 ' as DATE)", "1945-02-04", "DATE NOT NULL");
     f.checkScalar("cast('  1945-02-24  ' as DATE)",
         "1945-02-24", "DATE NOT NULL");
     if (castType == CastType.CAST) {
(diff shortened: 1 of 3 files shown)


mihaibudiu:
This is the PR which is being reverted in this one: https://github.com/apache/calcite/pull/3685

mihaibudiu:
It looks like this CI failure is to be expected, since it fails with Avatica main. That should pass once Avatica main contains the corresponding fix.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1329,8 +1329,14 @@ void testCastStringToDateTime(CastType castType, SqlOperatorFixture f) {
       f.checkNull("cast('notdate' as DATE)");
     }
 
-    f.checkScalar("cast('52534253' as DATE)", "7368-10-13", "DATE NOT NULL");
-    f.checkScalar("cast('1945-30-24' as DATE)", "1947-06-26", "DATE NOT NULL");
+    // Remove the if condition and the else block once CALCITE-6248 is fixed in Avatica
+    if (TestUtil.AVATICA_VERSION.startsWith("1.0.0-dev-main")) {
+      f.checkFails("cast('52534253' as DATE)", BAD_DATETIME_MESSAGE, true);
```
YiwenWu, line 1334: There is a unit test failed

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1320,7 +1320,7 @@ void testCastStringToDateTime(CastType castType, SqlOperatorFixture f) {
     f.checkCastToString("DATE '1945-2-24'", null, "1945-02-24", castType);
 
     f.checkScalar("cast('1945-02-24' as DATE)", "1945-02-24", "DATE NOT NULL");
-    f.checkScalar("cast(' 1945-02-04 ' as DATE)", "1945-02-04", "DATE NOT NULL");
```
YiwenWu, line 1323: Is it possible to add new unit tests instead of changing existing test cases?

mihaibudiu: I am afraid we have to change the tests because they are wrong.

JiajunBernoulli:
@mihaibudiu 
Maybe you can public test version in calcite-avatica.
Then update calcite-avatica in calcite to test.

mihaibudiu:
I sent a message on the dev list about this. I still have to figure out how to do this. We need simultaneous changes to avatica and calcite to make this pass.

mihaibudiu:
This is ready for review, to enable migration to Avatica 1.25.

mihaibudiu:
I am thinking to merge this PR, since there isn't much to review, so we can proceed with Avatica.

PR #3709, opened 2024-03-01
Improve documentation for Strong

chenzl25:
According to the definition, `c1 = 1 OR c2 IS NULL` cannot be strong in `[c1]`, because `c2 IS NULL` could be true.

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/Strong.java
+++ b/core/src/main/java/org/apache/calcite/plan/Strong.java
@@ -38,7 +38,7 @@ import java.util.Map;
 
 /** Utilities for strong predicates.
  *
- * <p>A predicate is strong (or null-rejecting) with regards to selected subset of inputs
+ * <p>A predicate is strong (or null-rejecting) with regard to selected subset of inputs
  * if it is UNKNOWN if all inputs in selected subset are UNKNOWN.
  *
  * <p>By the way, UNKNOWN is just the boolean form of NULL.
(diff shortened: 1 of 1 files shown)


caicancai:
Calcite has certain standard requirements for commit message. You can modify your commit message to pass cli. If you think it is necessary, you can mention it to jira

chenzl25:
> Calcite has certain standard requirements for commit message. You can modify your commit message to pass cli. If you think it is necessary, you can mention it to jira

Thanks! Let me mark it as [MINOR].

caicancai:
> > Calcite has certain standard requirements for commit message. You can modify your commit message to pass cli. If you think it is necessary, you can mention it to jira
> 
> Thanks! Let me mark it as [MINOR].

[MINOR] has been discontinued in recent discussions. Maybe we should use ‘Lint’ or ‘Typo’. The first letter of the commit message is not allowed to be fix. Maybe you should modify your commit message again.

core/src/main/java/org/apache/calcite/plan/Strong.java
```diff
@@ -53,8 +53,6 @@
  *   <li>{@code p1 AND p2} is strong in [p1, p2] (definitely null if either p1
  *   is null or p2 is null)
  *   <li>{@code p1 OR p2} is strong if p1 and p2 are strong
- *   <li>{@code c1 = 1 OR c2 IS NULL} is strong in [c1] (definitely null if c1
```
YiwenWu, line 56: LGTM.
By the way, can you also correct the spelling error of the idea prompt, changing `with regards to` to `with regard to`.

https://github.com/apache/calcite/blob/c546d4a59cf1cb20e6ec2a8981decf179d9a6b50/core/src/main/java/org/apache/calcite/plan/Strong.java#L41

YiwenWu (approved):

core/src/main/java/org/apache/calcite/plan/Strong.java
chenzl25: Sure.

caicancai (approved):
LGTM

caicancai:
@macroguo-ghy Hello，good to merge?

macroguo-ghy (changes requested):
Can you change the commit message to `Improve documentation for Strong`?

I think it is not a `Lint` commit.

PR #3711, opened 2024-03-01
[CALCITE-6259] The implementation of the Log library operator does not match the actual dialect behavior

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6259

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -208,9 +208,11 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.JSON_TYPE;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LEFT;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LEVENSHTEIN;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG10_MS;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG2;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_AND;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_OR;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG_MS;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LPAD;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_CONCAT;
@@ -638,11 +640,15 @@ public class RexImpTable {
       defineMethod(EXP, BuiltInMethod.EXP.method, NullPolicy.STRICT);
       defineMethod(POWER, BuiltInMethod.POWER.method, NullPolicy.STRICT);
       defineMethod(ABS, BuiltInMethod.ABS.method, NullPolicy.STRICT);
-      defineMethod(LOG2, BuiltInMethod.LOG2.method, NullPolicy.STRICT);
 
       map.put(LN, new LogImplementor());
       map.put(LOG, new LogImplementor());
       map.put(LOG10, new LogImplementor());
+
+      map.put(LN, new LogMSImplementor());
+      map.put(LOG_MS, new LogMSImplementor());
+      map.put(LOG2, new LogMSImplementor());
+      map.put(LOG10_MS, new LogMSImplementor());
 
       defineReflective(RAND, BuiltInMethod.RAND.method,
           BuiltInMethod.RAND_SEED.method);
@@ -4139,6 +4145,43 @@ public class RexImpTable {
         // fall through
       case "LN":
         return list.append(Expressions.constant(Math.exp(1)));
+      case "LOG10":
+        return list.append(Expressions.constant(BigDecimal.TEN));
+      default:
+        throw new AssertionError("Operator not found: " + call.getOperator());
+      }
+    }
+  }
+
+  /** Implementor for the {@code LN}, {@code LOG}, and {@code LOG10} operators.
+   *
+   * <p>Handles all logarithm functions using log rules to determine the
+   * appropriate base (i.e. base e for LN).
+   */
+  private static class LogMSImplementor extends AbstractRexCallImplementor {
+    LogMSImplementor() {
+      super("logMS", NullPolicy.STRICT, true);
+    }
+
+    @Override Expression implementSafe(final RexToLixTranslator translator,
+        final RexCall call, final List<Expression> argValueList) {
+      return Expressions.call(BuiltInMethod.LOGMS.method, args(call, argValueList));
+    }
+
+    private static List<Expression> args(RexCall call,
+        List<Expression> argValueList) {
+      Expression operand0 = argValueList.get(0);
+      final Expressions.FluentList<Expression> list = Expressions.list(operand0);
+      switch (call.getOperator().getName()) {
+      case "LOG":
+        if (argValueList.size() == 2) {
+          return list.append(argValueList.get(1));
+        }
+        // fall through
+      case "LN":
+        return list.append(Expressions.constant(Math.exp(1)));
+      case "LOG2":
+        return list.append(Expressions.constant(2));
       case "LOG10":
         return list.append(Expressions.constant(BigDecimal.TEN));
       default:
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -207,6 +207,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.JSON_STORAGE_SIZE;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.JSON_TYPE;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LEFT;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LEVENSHTEIN;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LN_MS;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG10_MS;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG2;
@@ -645,7 +646,7 @@ public class RexImpTable {
       map.put(LOG, new LogImplementor());
       map.put(LOG10, new LogImplementor());
 
-      map.put(LN, new LogMSImplementor());
+      map.put(LN_MS, new LogMSImplementor());
       map.put(LOG_MS, new LogMSImplementor());
       map.put(LOG2, new LogMSImplementor());
       map.put(LOG10_MS, new LogMSImplementor());
@@ -4153,7 +4154,8 @@ public class RexImpTable {
     }
   }
 
-  /** Implementor for the {@code LN}, {@code LOG}, and {@code LOG10} operators.
+  /** Implementor for the {@code LN}, {@code LOG}, {@code LOG2} and {@code LOG10} operators
+   *  on Mysql and Spark library
    *
    * <p>Handles all logarithm functions using log rules to determine the
    * appropriate base (i.e. base e for LN).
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2157,6 +2165,23 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
           OperandTypes.NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
+  /** The "LOG10(numeric)" function. Returns the base 10 logarithm of numeric. */
+  @LibraryOperator(libraries = {MYSQL, SPARK})
+  public static final SqlFunction LOG10_MS =
+      SqlBasicFunction.create("LOG10",
+          ReturnTypes.DOUBLE_FORCE_NULLABLE,
```
caicancai, line 2172: ReturnTypes.DOUBLE_FORCE_NULLABLE does not take effect

caicancai: I changed DOUBLE_NULLABLE in log10 in SqlStdOperatorTable into ReturnTypes.DOUBLE_FORCE_NULLABLE, and it took effect. Should I modify this? I'm not sure.

caicancai: @tanclary Do you have any ideas? You can uncomment me and try to debug it. Thank you

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6280,6 +6280,94 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+      f.checkNull("log(null)");
+      f.checkNull("log(cast(null as real))");
+    };
+    f0.forEachLibrary(list(SqlLibrary.MYSQL, SqlLibrary.SPARK), consumer);
+  }
+
+//  /** Test case for
+//   * <a href="https://issues.apache.org/jira/browse/CALCITE-6259">[CALCITE-6259]
+//   * Add LOG10 function (enabled in MYSQL, Spark library)</a>. */
+//  @Test void testLog10MSFunc() {
```
caicancai, line 6323: @tanclary Do you have any ideas? You can uncomment me and try to debug it. Thank you

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
```diff
@@ -510,7 +510,7 @@ public enum BuiltInMethod {
   SAFE_MULTIPLY(SqlFunctions.class, "safeMultiply", double.class, double.class),
   SAFE_SUBTRACT(SqlFunctions.class, "safeSubtract", double.class, double.class),
   LOG(SqlFunctions.class, "log", long.class, long.class),
-  LOG2(SqlFunctions.class, "log2", long.class),
+  LOGMS(SqlFunctions.class, "logMS", long.class, long.class),
```
YiwenWu, line 513: Why delete `LOG2` here, but `SqlLibraryOperators` has no deletion

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2157,6 +2165,23 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
+          OperandTypes.NUMERIC,
+          SqlFunctionCategory.NUMERIC);
```
YiwenWu, line 2174: I have a question, `LOG10` and `LN` have been defined in `SqlStdOperatorTable`. Defining functions with the same name and Sqlkind here may cause `SqlOperatorTable#lookupOperatorOverloads` to find multiple `overloads`, will this cause SQL validation exceptions?

site/_docs/reference.md
```diff
@@ -2783,7 +2783,10 @@ In the following:
 | b f s | LENGTH(string)                             | Equivalent to `CHAR_LENGTH(string)`
 | h s | LEVENSHTEIN(string1, string2)                | Returns the Levenshtein distance between *string1* and *string2*
 | b | LOG(numeric1 [, numeric2 ])                    | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present
+| m s | LN(numeric)                                  | Returns the natural logarithm (base *e*) of *numeric*
```
YiwenWu, line 2786: `LN` and `LOG10` are defined in both built-in and engine extension functions, which may confuse users.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: Thank you for your feedback. I have asked similar questions in jira. Because the behavior of log10 and ln functions is inconsistent between mysql and spark and the log10 and ln functions in SqlStdOperatorTable. The problem you mentioned does occur. I am trying to find a reasonable way to solve this problem. If you have a way to solve this problem without modifying SqlStdOperatorTable, that would be great.

caicancai: The problem I am currently encountering is that ReturnTypes.DOUBLE_FORCE_NULLABLE of LOG10_MS does not take effect.

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
caicancai: The log2 function was originally designed to adapt to mysql and spark functions. This function adapts to the behavior of all log functions in mysql and spark. I made some reasonable reconstructions. In order to improve the readability and maintainability of the code, you can Found that the log2 function is still effective

site/_docs/reference.md
caicancai: I don’t think so. Users can still choose the log10 and ln functions in SqlStdOperatorTable.

caicancai:
It looks like there is a class loading conflict

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
YiwenWu: You can try to change the function `SqlKind`  in `SqlLibraryOperators`, because the default `equals` method of `SqlOperator` is to compare function name and SqlKind.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
YiwenWu: It may be a function loading conflict. You can change the SqlKind of the custom function,  instead of modifying the `SqlStdOperatorTable` function.

caicancai:
@tanclary It seems like I don't have a good way to solve this problem. Can you give me some advice?

tanclary:
> @tanclary It seems like I don't have a good way to solve this problem. Can you give me some advice?

Can you be more specific? What problem?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -638,12 +640,16 @@ Builder populate() {
       defineMethod(EXP, BuiltInMethod.EXP.method, NullPolicy.STRICT);
       defineMethod(POWER, BuiltInMethod.POWER.method, NullPolicy.STRICT);
       defineMethod(ABS, BuiltInMethod.ABS.method, NullPolicy.STRICT);
-      defineMethod(LOG2, BuiltInMethod.LOG2.method, NullPolicy.STRICT);
 
       map.put(LN, new LogImplementor());
       map.put(LOG, new LogImplementor());
       map.put(LOG10, new LogImplementor());
 
+      map.put(LN, new LogMSImplementor());
```
mihaibudiu, line 648: map is a Java Map, you are overwriting the previous entry.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4147,6 +4153,43 @@ private static List<Expression> args(RexCall call,
     }
   }
 
+  /** Implementor for the {@code LN}, {@code LOG}, and {@code LOG10} operators.
+   *
+   * <p>Handles all logarithm functions using log rules to determine the
+   * appropriate base (i.e. base e for LN).
+   */
+  private static class LogMSImplementor extends AbstractRexCallImplementor {
```
mihaibudiu, line 4161: You should probably document that this is for MySQL and Spark.
The part of the comment about determining the base is not very useful.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: It doesn't look like this PR is ready for review yet.

caicancai:
> Can you be more specific? What problem?

You can find that I defined ReturnTypes.DOUBLE_FORCE_NULLABLE in LOG10_MS. The purpose is to make LOG10_MS return null when log10(0). However, during my debugging process, I found that the ReturnTypes.DOUBLE_FORCE_NULLABLE I defined 2 did not take effect.

```java
@LibraryOperator(libraries = {MYSQL, SPARK})
  public static final SqlFunction LOG10_MS =
      ((SqlBasicFunction) LOG10).withKind(SqlKind.LOG10_MS)
          .withReturnTypeInference(ReturnTypes.DOUBLE_FORCE_NULLABLE);
```

Regarding the ReturnType processing of LOG10, it still follows the ReturnType logic of LOG10 in `SqlStdOperatorTable`

```java
public static final SqlFunction LOG10 =
      SqlBasicFunction.create("LOG10",
          ReturnTypes.DOUBLE_NULLABLE,
          OperandTypes.NUMERIC,
          SqlFunctionCategory.NUMERIC);
```

Can you understand what I mean?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: Yes, I had some difficulties. Sorry

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: ohhh, thank you for reminding me, I made a serious mistake

caicancai:
@tanclary I feel that I cannot change SqlStdOperatorTable, which will affect some other tests (I tried)

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2142,13 +2143,21 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
    * @see SqlStdOperatorTable#LN
    * @see SqlStdOperatorTable#LOG10
    */
-  @LibraryOperator(libraries = {BIG_QUERY})
+  @LibraryOperator(libraries = {BIG_QUERY, POSTGRESQL})
```
jduo, line 2146: PostgreSQL has different semantics for its log function.
If it's only 1 argument, PostgreSQL is the same as LOG10() and BigQuery is the same as LN().
If it's 2-argument, PostgreSQL has the base as the first argument and BigQuery has the base as the second argument.

See:
https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#log
https://www.postgresql.org/docs/9.1/functions-math.html
Mentioned in https://issues.apache.org/jira/browse/CALCITE-6312

jduo:
> @tanclary I feel that I cannot change SqlStdOperatorTable, which will affect some other tests (I tried)

I raised a similar question on the mailing list and didn't hear back. See:
https://lists.apache.org/thread/dloo5q43k6b739gqmg1qf2hht3wzvt0t

I *think* we need to edit the parser so that when it encounters the LOG function, we call a method to look up the SqlOperator based on the SqlConformance. Similar to what looks to have been suggested for Floor/Ceil. @tanclary @julianhyde perhaps you know?

julianhyde:
No, the parser should not be making decisions about semantics. The validator (assisted by the operator table) should be doing that.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: get, As you can see, I encountered difficulties in implementing the log10 and ln functions. I think that only by solving this problem, everything will be solved.

caicancai:
> No, the parser should not be making decisions about semantics. The validator (assisted by the operator table) should be doing that.

Hi，I would like to ask if there are any previous cases similar to Jira that I can learn from. I don’t have a good idea at the moment. I should try Benchao’s idea https://lists.apache.org/list.html?dev@calcite.apache.org , but I’m not sure if I can complete it.

caicancai:
@tanclary In addition, I have to explain that currently calcite's log10(0) returns negative infinity, but many databases I tested returned NULL or errors. I am not sure whether I should modify the underlying implementation of the log function (Sqlfunctions.log) , or create a new plan

caicancai:
I created two corresponding subtasks because the PR code changes may be large, and I am not sure about the ln and log10 functions, so I will adapt the log function first. Thank you for your reviews and suggestions.

PR #3712, opened 2024-03-03
[CALCITE-6278] Add REGEXP, REGEXP_LIKE  function (enabled in Spark library)

YiwenWu:
https://issues.apache.org/jira/browse/CALCITE-6278

1. Add Function [REGEXP](https://spark.apache.org/docs/latest/api/sql/index.html#regexp)、[REGEXP_LIKE](https://spark.apache.org/docs/latest/api/sql/index.html#regexp_like) enabled in the Spark library.
Since this function has the same implementation as the Spark [RLIKE](https://spark.apache.org/docs/latest/api/sql/index.html#rlike) function. 
The implementation can be reused.

[Source Code](https://github.com/apache/spark/blob/ee15cbd3ac3cc98ae262834b4f16e74fe54d82c3/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala#L567C1-L569C54)
image

[undo] Discuss results in Jira:
> 2. Since Spark 2.0, string literals (including regex patterns) are unescaped in SQL parser, also fix this bug in calcite.  

2. Add spark query tests in BabelQuidemTest to ensure the execution correctness.

### DIFF
--- a/babel/src/test/resources/sql/spark.iq
+++ b/babel/src/test/resources/sql/spark.iq
@@ -1,0 +1,243 @@
+# spark.iq - Babel test for Spark dialect of SQL
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to you under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+!use scott-spark
+!set outputformat csv
+
+#####################################################################
+# String functions ######################
+
+#####################################################################
+# RLIKE
+#
+# string1 RLIKE string2
+# Returns true if str matches regexp, or false otherwise.
+#
+# Returns BOOLEAN
+
+select NULL RLIKE 'abc*';
+EXPR$0
+null
+!ok
+
+select 'abc' RLIKE NULL;
+EXPR$0
+null
+!ok
+
+select 'abc' RLIKE '';
+EXPR$0
+true
+!ok
+
+SELECT 'abc def ghi' RLIKE 'abc';
+EXPR$0
+true
+!ok
+
+SELECT 'abc def ghi' RLIKE 'abcd';
+EXPR$0
+false
+!ok
+
+select 'abc' RLIKE '^\abc$';
+EXPR$0
+false
+!ok
+
+select '\abc' RLIKE '\abc$';
+EXPR$0
+false
+!ok
+
+select '\abc' RLIKE '^\abc$';
+EXPR$0
+false
+!ok
+
+select '\abc' RLIKE '^\\abc$';
+EXPR$0
+true
+!ok
+
+select 'abc' RLIKE '^abc$';
+EXPR$0
+true
+!ok
+
+select 'abc' RLIKE 'abz*';
+EXPR$0
+true
+!ok
+
+SELECT '%SystemDrive%\\Users\\John' RLIKE '%SystemDrive%\\\\Users.*';
+EXPR$0
+true
+!ok
+
+select '%SystemDrive%\Users\John' RLIKE '%SystemDrive%\\Users.*';
+EXPR$0
+true
+!ok
+
+#####################################################################
+# REGEXP
+#
+# REGEXP(str, regexp)
+# Returns true if str matches regexp, or false otherwise.
+#
+# Returns BOOLEAN
+
+select REGEXP(NULL, 'abc*');
+EXPR$0
+null
+!ok
+
+select REGEXP('abc', NULL);
+EXPR$0
+null
+!ok
+
+select REGEXP('abc', '');
+EXPR$0
+true
+!ok
+
+SELECT REGEXP('abc def ghi', 'abc');
+EXPR$0
+true
+!ok
+
+SELECT REGEXP('abc def ghi', 'abcd');
+EXPR$0
+false
+!ok
+
+select REGEXP('abc', '^\abc$');
+EXPR$0
+false
+!ok
+
+select REGEXP('\abc', '\abc$');
+EXPR$0
+false
+!ok
+
+select REGEXP('\abc', '^\abc$');
+EXPR$0
+false
+!ok
+
+select REGEXP('\abc', '^\\abc$');
+EXPR$0
+true
+!ok
+
+select REGEXP('abc', '^abc$');
+EXPR$0
+true
+!ok
+
+select REGEXP('abc', 'abz*');
+EXPR$0
+true
+!ok
+
+SELECT REGEXP('%SystemDrive%\\Users\\John', '%SystemDrive%\\\\Users.*');
+EXPR$0
+true
+!ok
+
+SELECT REGEXP('%SystemDrive%\Users\John', '%SystemDrive%\\Users.*');
+EXPR$0
+true
+!ok
+
+#####################################################################
+# REGEXP_LIKE
+#
+# REGEXP_LIKE(str, regexp)
+# Returns true if str matches regexp, or false otherwise.
+#
+# Returns BOOLEAN
+
+select REGEXP_LIKE(NULL, 'abc*');
+EXPR$0
+null
+!ok
+
+select REGEXP_LIKE('abc', NULL);
+EXPR$0
+null
+!ok
+
+select REGEXP_LIKE('abc', '');
+EXPR$0
+true
+!ok
+
+SELECT REGEXP_LIKE('abc def ghi', 'abc');
+EXPR$0
+true
+!ok
+
+SELECT REGEXP_LIKE('abc def ghi', 'abcd');
+EXPR$0
+false
+!ok
+
+select REGEXP_LIKE('abc', '^\abc$');
+EXPR$0
+false
+!ok
+
+select REGEXP_LIKE('\abc', '\abc$');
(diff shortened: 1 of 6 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4928,6 +4928,37 @@ private static void checkIf(SqlOperatorFixture f) {
     f.checkNull("regexp_contains(cast(null as varchar), cast(null as varchar))");
   }
 
+  @Test void testRegexpFunc() {
+    final SqlOperatorFixture f = fixture().setFor(SqlLibraryOperators.REGEXP);
```
caicancai, line 4932: fixture().setFor(SqlLibraryOperators.REGEXP).withLibrary(SqlLibrary.Spark) ？

babel/src/test/resources/sql/spark.iq
```diff
@@ -0,0 +1,243 @@
+# spark.iq - Babel test for Spark dialect of SQL
```
caicancai, line 1: Are you sure this test only takes effect for semantics that conform to spark sql?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -691,7 +691,9 @@ private static int makeRegexpFlags(String stringFlags) {
 
     /** SQL {@code RLIKE} function. */
     public boolean rlike(String s, String pattern) {
-      return cache.getUnchecked(new Key(0, pattern)).matcher(s).find();
+      s = StringEscapeUtils.unescapeJava(s);
```
caicancai, line 694: Do we need to add a little comment?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
YiwenWu: `withLibrary` is optional.
`f0.forEachLibrary(list(functionAlias.libraries), consumer)` will automatically load the corresponding Library type.

babel/src/test/resources/sql/spark.iq
YiwenWu: This test will load Built-in functions and spark functions.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
YiwenWu: done, add comment

babel/src/test/resources/sql/spark.iq
JiajunBernoulli: I am not against using this test.

But I want to know the necessity of using QuidemTest for Spark.
Because your code may also become a future contribution specification.

Let's discuss two questions.
1. Why need `spark.iq`?
2. What features need to be tested and what features don't need to be tested?

caicancai: +1
At present, calcite has adapted many spark functions. Do we need to add all spark functions? What are the benefits after adding them?

YiwenWu: 1. Compared to `SqlOperatorTest`, I think the function tests in Babel are more accurate and complete. Therefore, 
references`big-query.iq` and `redshift.iq`,  `spark.iq` were added to verify the correctness of function execution.
2. I think that unit tests can be added to this file when adding spark functions, but it is unnecessary.

babel/src/test/resources/sql/spark.iq
```diff
@@ -0,0 +1,243 @@
...
+true
+!ok
+
+#####################################################################
+# REGEXP
+#
+# REGEXP(str, regexp)
+# Returns true if str matches regexp, or false otherwise.
+#
+# Returns STRING
```
julianhyde, line 103: no it doesn't

site/_docs/reference.md
```diff
@@ -2812,6 +2812,8 @@ In the following:
 | b | PARSE_TIMESTAMP(format, string[, timeZone])    | Uses format specified by *format* to convert *string* representation of timestamp to a TIMESTAMP WITH LOCAL TIME ZONE value in *timeZone*
 | h s | PARSE_URL(urlString, partToExtract [, keyToExtract] ) | Returns the specified *partToExtract* from the *urlString*. Valid values for *partToExtract* include HOST, PATH, QUERY, REF, PROTOCOL, AUTHORITY, FILE, and USERINFO. *keyToExtract* specifies which query to extract
 | b s | POW(numeric1, numeric2)                      | Returns *numeric1* raised to the power *numeric2*
+| s | REGEXP(string, regexp)                         | Returns true if *string* matches *regexp*, or false otherwise, string literals (including regex patterns) are unescaped
```
julianhyde, line 2815: can't you just say 'alias for RLIKE'?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4928,6 +4928,37 @@ private static void checkIf(SqlOperatorFixture f) {
...
+
+      f.checkQuery("select " + fn + "('abc def ghi', 'abc')");
+      f.checkQuery("select " + fn + "('foo@bar.com', '@[a-zA-Z0-9-]+\\\\.[a-zA-Z0-9-.]+')");
+      f.checkQuery("select " + fn + "('55as56664as422', '\\d{10}')");
+
+      f.checkNull(fn + "('abc def ghi', cast(null as varchar))");
+      f.checkNull(fn + "(cast(null as varchar), 'abc')");
+      f.checkNull(fn + "(cast(null as varchar), cast(null as varchar))");
+    };
+    f0.forEachLibrary(list(functionAlias.libraries), consumer);
```
julianhyde, line 4959: if RLIKE, REGEXP_LIKE and REGEXP are identical, you should devise a way to test all three with a single test

babel/src/test/resources/sql/spark.iq
YiwenWu: done, fixed to return Boolean

site/_docs/reference.md
YiwenWu: done, `Equivalent to string1 RLIKE string2`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
YiwenWu: Although `RLIKE` and `REGEXP` have the same implementation, their function calling methods differ, so they are divided into different tests.
- `RLIKE`:  'str' RLIKE 'regex'
- `REGEXP`|`REGEXP_LIKE` : fn('str',  'regex').

In addition, we can abstract a method to support different Function calling modes, but I may think it is not necessary at present

YiwenWu: If you think it's better to combine them into a single test, I'll modify the test case.

julianhyde: It is necessary. And it's easy to combine all three. `BinaryOperator<String>`. For rlike, pass `(a, b) -> a + " RLIKE " b`; for REGEXP, pass `(a, b) -> "REGEXP(" + a + ", " + b + ")".

YiwenWu: done,  combined test case of `RLIKE`、`REGEXP`、`REGEXP_LIKE`

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -691,7 +691,10 @@ private static int makeRegexpFlags(String stringFlags) {
 
     /** SQL {@code RLIKE} function. */
     public boolean rlike(String s, String pattern) {
-      return cache.getUnchecked(new Key(0, pattern)).matcher(s).find();
+      // Since Spark 2.0, string literals (including regex patterns) are unescaped in SQL parser
```
macroguo-ghy, line 694: `rlike` also enabled in HIVE library, does the `rlike` function have the same semantics in hive as in spark sql?

YiwenWu: Yes, judging by the engine execution behavior, Hive also performs the unescape, and the result is consistent with Spark.
image

macroguo-ghy (approved):
+1

jduo:
Would it be possible to merge these changes? I'm looking to implement CALCITE-6308 on top of this.

site/_docs/reference.md
```diff
@@ -2821,8 +2821,10 @@ In the following:
 | b m p s | REPEAT(string, integer)                  | Returns a string consisting of *string* repeated of *integer* times; returns an empty string if *integer* is less than 1
 | b m | REVERSE(string)                              | Returns *string* with the order of the characters reversed
 | b m p s | RIGHT(string, length)                    | Returns the rightmost *length* characters from the *string*
-| h s | string1 RLIKE string2                        | Whether *string1* matches regex pattern *string2* (similar to `LIKE`, but uses Java regex)
-| h s | string1 NOT RLIKE string2                    | Whether *string1* does not match regex pattern *string2* (similar to `NOT LIKE`, but uses Java regex)
+| h s | string1 RLIKE string2                        | Whether *string1* matches regex pattern *string2* (similar to `LIKE`, but uses Java regex), string literals (including regex patterns) are unescaped
```
JiajunBernoulli, line 2824: Should we maintain alphabetical order?
> `re` before than `ri`

YiwenWu: done, order adjusted.

YiwenWu:
> Would it be possible to merge these changes? I'm looking to implement [CALCITE-6309](https://issues.apache.org/jira/browse/CALCITE-6309) on top of this.

@jduo  I think this mr will be able to be merged recently, the details of the relevant issues have been discussed clearly in Jira.

macroguo-ghy:
If no other comment, I will merge it later.

jduo:
> If no other comment, I will merge it later.

Thanks, that'd be great. Once that's in I'll follow-up with the CALCITE-6309 work.

PR #3714, opened 2024-03-04
[CALCITE-6290] Incorrect return type for BigQuery TRUNC

tanclary:
The modified tests are a result of me changing the `ARG0_EXCEPT_INTEGER` return type. I wrote that code last June and I got the nullability wrong so that's why some FLOOR/CEIL tests got changed. I also realized I wasn't preserving the scale/precision of DECIMAL types so I fixed that too. Let me know if anyone has questions.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -456,7 +456,7 @@ public abstract class SqlLibraryOperators {
       .withName("CEIL_BIG_QUERY")
       .withReturnTypeInference(ReturnTypes.ARG0_EXCEPT_INTEGER_NULLABLE);
 
-  /** The "FLOOR(value)" function. Identical to the stadnard <code>FLOOR</code> function
+  /** The "FLOOR(value)" function. Identical to the standard <code>FLOOR</code> function
    * except the return type should be a double if the operand is an integer. */
   @LibraryOperator(libraries = {BIG_QUERY})
   public static final SqlFunction FLOOR_BIG_QUERY = new SqlFloorFunction(SqlKind.FLOOR)
@@ -2161,9 +2161,12 @@ public abstract class SqlLibraryOperators {
   public static final SqlFunction POW =
       SqlStdOperatorTable.POWER.withName("POW");
 
+  /** The "TRUNC(numeric1 [, integer2])" function. Identical to the standard <code>TRUNCATE</code>
+  * function except the return type should be a double if numeric1 is an integer. */
   @LibraryOperator(libraries = {BIG_QUERY})
-  public static final SqlFunction TRUNC =
-      SqlStdOperatorTable.TRUNCATE.withName("TRUNC");
+  public static final SqlFunction TRUNC_BIG_QUERY = SqlStdOperatorTable.TRUNCATE
+          .withName("TRUNC")
+          .withReturnTypeInference(ReturnTypes.ARG0_EXCEPT_INTEGER_NULLABLE);
 
   /** Infix "::" cast operator used by PostgreSQL, for example
    * {@code '100'::INTEGER}. */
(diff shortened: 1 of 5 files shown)


wnob (approved):

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -550,12 +550,12 @@ public static SqlCall stripSeparator(SqlCall call) {
    */
   public static final SqlReturnTypeInference ARG0_EXCEPT_INTEGER = opBinding ->  {
     RelDataTypeFactory typeFactory = opBinding.getTypeFactory();
-    SqlTypeName op = opBinding.getOperandType(0).getSqlTypeName();
-    if (SqlTypeName.INT_TYPES.contains(op)) {
+    RelDataType op = opBinding.getOperandType(0);
```
julianhyde, line 553: `op` is no longer a good name for this variable

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2161,9 +2161,12 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
   public static final SqlFunction POW =
       SqlStdOperatorTable.POWER.withName("POW");
 
+  /** The "TRUNC(numeric1 [, integer2])" function. Identical to the standard <code>TRUNCATE</code>
+  * function except the return type should be a double if numeric1 is an integer. */
   @LibraryOperator(libraries = {BIG_QUERY})
-  public static final SqlFunction TRUNC =
-      SqlStdOperatorTable.TRUNCATE.withName("TRUNC");
+  public static final SqlFunction TRUNC = SqlStdOperatorTable.TRUNCATE
```
julianhyde, line 2167: Consider renaming the function to TRUNC_BIG_QUERY. The BigQuery FLOOR and TRUNC functions are now in sync, but need to remain so, and linking their names will make that more likely. 

In Jira please link to CALCITE-5747. That's where these came adrift in the first place.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -592,6 +592,20 @@ private static String toSql(RelNode root, SqlDialect dialect,
         + "created_thing\nFROM foodmart.product");
   }
 
+    /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6290">[CALCITE-6290]
+   * Incorrect return type for BigQuery TRUNC</a>. */
+  @Test void testBigQueryTruncPreservesCast() {
+    final String query = "SELECT CAST(TRUNC(3) AS BIGINT) "
+        + "as created_thing\n FROM `foodmart`.`product`";
```
julianhyde, line 600: add line break after `\n`

julianhyde (changes requested):
In Jira and commit message please add a note that, in addition to changing BigQuery TRUNC to return DOUBLE, you are changing the nullability of CEIL and FLOOR in BigQuery.

+1 when the requested changes are done

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
tanclary: I think renaming it might be a bit redundant, I only renamed FLOOR and CEIL because they have standard counterparts with the same name, TRUNC however is unique from the standard TRUNCATE. What do you think? I don't feel very strongly either way. @julianhyde

tanclary: And I think the JIRA should already have a link to 5747 let me know if it looks wrong

julianhyde: maintainers will do the wrong thing if you don't make it very obvious. Find a way to make it obvious.

PR #3716, opened 2024-03-07
[CALCITE-6301] Extend ‘Must-filter’ columns to support a conditional …

olivrlee:
…bypass list

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/AggregateJoinTransposeRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/AggregateJoinTransposeRule.java
@@ -320,7 +320,10 @@ public class AggregateJoinTransposeRule
     // Update condition
     final Mapping mapping =
         (Mapping) Mappings.target(map::get,
-            join.getRowType().getFieldCount(),
+            join.getJoinType().projectsRight()
+                ? join.getRowType().getFieldCount()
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3894,32 +3894,37 @@ protected void validateSelect(
 
     // Deduce which columns must be filtered.
     ns.mustFilterFields = ImmutableBitSet.of();
+    ns.mustFilterBypassFields = ImmutableBitSet.of();
     if (from != null) {
       final Set<SqlQualified> qualifieds = new LinkedHashSet<>();
```
tanclary, line 3899: is `qualifieds` == `qualifiedFields` If so you could make it `qualifiedFields` and `qualifiedBypassFields` for consistency ?

tanclary (commented):
Aside from the style checks I think this looks really solid

mihaibudiu:
@olivrlee can you please fix the style issues which prevent the CI from passing?

olivrlee:
> @olivrlee can you please fix the style issues which prevent the CI from passing?

Sorry, I haven't had time to return to this.
There is one failing test case that I added in my last commit that I need to address before this will be ready for review again.

PR #3717, opened 2024-03-07
[CALCITE-6306] JDBC adapter should not generate FILTER (WHERE) in MySQL and StarRocks dialect

macroguo-ghy:
CALCITE-6306


YiwenWu (approved):
LGTM

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3190,6 +3190,14 @@ private SqlDialect nonOrdinalDialect() {
     sql(query).dialect(MysqlSqlDialect.DEFAULT).ok(expected);
   }
 
+  @Test void testMysqlWithAggregateFunctionFilter() {
```
julianhyde, line 3193: no need for a new test method. add MySQL to an existing test method.

PR #3720, opened 2024-03-08
[CALCITE-6314] Add RANDOM function (enabled in Postgres library)

jduo:


jduo:
@tanclary

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2257,4 +2258,16 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
   @LibraryOperator(libraries = {SPARK})
   public static final SqlFunction GETBIT =
       BIT_GET.withName("GETBIT");
+
+  /** The RANDOM() function. Equivalent to RAND(). */
+  @LibraryOperator(libraries = {POSTGRESQL})
+  public static final SqlFunction RANDOM = new SqlRandFunction() {
```
tanclary, line 2264: what's to keep you from doing RAND.withName("RANDOM").withReturnTypeInferance(...)` I think that's better than these overrides

site/_docs/reference.md
```diff
@@ -2813,6 +2813,7 @@ In the following:
 | b | PARSE_TIMESTAMP(format, string[, timeZone])    | Uses format specified by *format* to convert *string* representation of timestamp to a TIMESTAMP WITH LOCAL TIME ZONE value in *timeZone*
 | h s | PARSE_URL(urlString, partToExtract [, keyToExtract] ) | Returns the specified *partToExtract* from the *urlString*. Valid values for *partToExtract* include HOST, PATH, QUERY, REF, PROTOCOL, AUTHORITY, FILE, and USERINFO. *keyToExtract* specifies which query to extract
 | b s | POW(numeric1, numeric2)                      | Returns *numeric1* raised to the power *numeric2*
+| p | RANDOM()                                       | Generates a random double between 0 and 1 inclusive
```
tanclary, line 2816: Is there already documentation for RAND? If so, just say `Equivalent to RAND` (see other functions w multiple names for examples)

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6307,6 +6307,22 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+    final SqlOperatorFixture f = fixture();
+    f.setFor(SqlLibraryOperators.RANDOM, VmName.EXPAND);
+    f.checkFails("^random^", "Column 'RANDOM' not found in any table", false);
+    Consumer<SqlOperatorFixture> test = fixture -> {
+      for (int i = 0; i < 100; i++) {
+        // Result must always be between 0 and 1, inclusive.
+        fixture.checkScalarApprox("random()", "DOUBLE NOT NULL", isWithin(0.5, 0.5));
+      }
+    };
+    f.forEachLibrary(list(SqlLibrary.POSTGRESQL), test);
```
tanclary, line 6323: do any other main dialects used by Calcite support this function? If so, enable them too please

tanclary (changes requested):
Please adjust the JIRA and commit message to follow the style of other function additions, should look something like `[CALCITE-XXXX] Add RANDOM function (enabled in Postgres library)` look through commit history for examples

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
jduo: I'd prefer that as well. Those methods are part of SqlBasicFunction, but SqlRandFunction derives directly from SqlFunction.

site/_docs/reference.md
jduo: This is from the RAND documentation, minus the comment about an optional seed. In this case where the functions are not exactly the same due to the missing overload, should we still use "Equivalent to"...?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
jduo: Looks like Redshift is the only other DBMS supporting this.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
jduo: Done

tanclary: can't you make SqlRandFunction extend SqlBasicfunction

site/_docs/reference.md
tanclary: ah sorry didn't see there's a difference, this is fine

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
jduo: SqlBasicFunction has a private constructor.
I'll look into making RAND be an instance of SqlBasicFunction then RANDOM being a variant instead.

jduo: Done.

tanclary: the constructor is private because you're meant to use `.create()`

jduo: Yes, I wound up adding more with*() functions to SqlBasicFunction for setting dynamic and the OperandTypeValidator and made both RAND() and RANDOM() instances of it.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6307,6 +6307,22 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6314">[CALCITE-6314]
+   * Add RANDOM function (enabled in Postgres library)</a>. */
+  @Test void testRandomFunc() {
+    final SqlOperatorFixture f = fixture();
+    f.setFor(SqlLibraryOperators.RANDOM, VmName.EXPAND);
+    f.checkFails("^random^", "Column 'RANDOM' not found in any table", false);
+    Consumer<SqlOperatorFixture> test = fixture -> {
```
tanclary, line 6317: sorry this looks good but can we just call this `consumer` to fit the pattern of other tests? `test` doesn't tell us much

tanclary (changes requested):
One small comment, +1 to merge after

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
jduo: Sure, I've renamed test to consumer.

tanclary (commented):
Looks like some merge conflicts?

jduo:
> Looks like some merge conflicts?

Updated the PR. It's introduced from merging CALCITE-6278. I expect we'll see more of this since we have a few devs working on adding functions now.

tanclary (approved):
LGTM

PR #3721, opened 2024-03-08

caicancai:

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -1163,6 +1163,50 @@ public abstract class SqlLibraryOperators {
         false);
   }
 
+  private static RelDataType mapKeyReturnType(SqlOperatorBinding opBinding) {
+    List<RelDataType> operandType = new ArrayList<>();
+
+    operandType.add(opBinding.collectOperandTypes().get(0).getKeyType());
+    operandType.add(opBinding.collectOperandTypes().get(0).getValueType());
+    Pair<@Nullable RelDataType, @Nullable RelDataType> type =
+        getComponentTypes(
+            opBinding.getTypeFactory(), operandType);
+
+    requireNonNull(type.left, () -> "type left of " + type.left);
+    requireNonNull(type.right, () -> "type right of " + type.right);
+    if (type.left.getSqlTypeName() != SqlTypeName.UNKNOWN
+        && type.right.getSqlTypeName() != SqlTypeName.UNKNOWN) {
+      SqlValidatorUtil.adjustTypeForMapFunctionConstructor(type, opBinding);
+    }
+
+    return SqlTypeUtil.createArrayType(
+        opBinding.getTypeFactory(),
+        requireNonNull(type.left, "inferred key type"),
+        false);
+  }
+
+  private static RelDataType mapValueReturnType(SqlOperatorBinding opBinding) {
+    List<RelDataType> operandType = new ArrayList<>();
+
+    operandType.add(opBinding.collectOperandTypes().get(0).getKeyType());
+    operandType.add(opBinding.collectOperandTypes().get(0).getValueType());
+    Pair<@Nullable RelDataType, @Nullable RelDataType> type =
+        getComponentTypes(
+            opBinding.getTypeFactory(), operandType);
+
+    requireNonNull(type.left, () -> "type left of " + type.left);
+    requireNonNull(type.right, () -> "type right of " + type.right);
+    if (type.left.getSqlTypeName() != SqlTypeName.UNKNOWN
+        && type.right.getSqlTypeName() != SqlTypeName.UNKNOWN) {
+      SqlValidatorUtil.adjustTypeForMapFunctionConstructor(type, opBinding);
+    }
+
+    return SqlTypeUtil.createArrayType(
+        opBinding.getTypeFactory(),
+        requireNonNull(type.right, "inferred value type"),
+        false);
+  }
+
   private static Pair<@Nullable RelDataType, @Nullable RelDataType> getComponentTypes(
       RelDataTypeFactory typeFactory,
       List<RelDataType> argTypes) {
@@ -1514,14 +1558,14 @@ public abstract class SqlLibraryOperators {
   @LibraryOperator(libraries = {SPARK})
   public static final SqlFunction MAP_KEYS =
       SqlBasicFunction.create(SqlKind.MAP_KEYS,
-          ReturnTypes.TO_MAP_KEYS_NULLABLE,
+          SqlLibraryOperators::mapKeyReturnType,
           OperandTypes.MAP);
 
   /** The "MAP_VALUES(map)" function. */
   @LibraryOperator(libraries = {SPARK})
   public static final SqlFunction MAP_VALUES =
       SqlBasicFunction.create(SqlKind.MAP_VALUES,
-          ReturnTypes.TO_MAP_VALUES_NULLABLE,
+          SqlLibraryOperators::mapValueReturnType,
           OperandTypes.MAP);
 
   /** The "MAP_CONTAINS_KEY(map, key)" function. */
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -1163,6 +1163,7 @@ public abstract class SqlLibraryOperators {
         false);
   }
 
+  @SuppressWarnings("argument.type.incompatible")
   private static RelDataType mapKeyReturnType(SqlOperatorBinding opBinding) {
     List<RelDataType> operandType = new ArrayList<>();
 
@@ -1185,6 +1186,7 @@ public abstract class SqlLibraryOperators {
         false);
   }
 
+  @SuppressWarnings("argument.type.incompatible")
   private static RelDataType mapValueReturnType(SqlOperatorBinding opBinding) {
     List<RelDataType> operandType = new ArrayList<>();
 
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorUtil.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorUtil.java
@@ -1390,6 +1390,23 @@ public class SqlValidatorUtil {
   }
 
   /**
+   * When the map key or value does not equal the map component key type or value type,
+   * make explicit casting.
+   *
+   * @param componentType derived map pair component type
+   * @param opBinding description of call
+   */
+  public static void adjustTypeForMapFunctionConstructor(
+      Pair<RelDataType, RelDataType> componentType, SqlOperatorBinding opBinding) {
+    if (opBinding instanceof SqlCallBinding) {
+      requireNonNull(componentType.getKey(), "map key type");
+      requireNonNull(componentType.getValue(), "map value type");
+      adjustTypeForMapFunctionConstructor(
+          componentType.getKey(), componentType.getValue(), (SqlCallBinding) opBinding);
+    }
+  }
+
+  /**
    * Adjusts the types for operands in a SqlCallBinding during the construction of a sql collection
    * type such as Array or Map. This method iterates from the operands of a {@link SqlCall}
    * obtained from the provided {@link SqlCallBinding}.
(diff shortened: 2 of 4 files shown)


caicancai:
I may need to add some comments, but the code structure is basically determined

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1149,6 +1149,56 @@ private static RelDataType mapReturnType(SqlOperatorBinding opBinding) {
...
+  private static RelDataType mapKeyReturnType(SqlOperatorBinding opBinding) {
+    List<RelDataType> operandType = new ArrayList<>();
+
+    operandType.add(opBinding.collectOperandTypes().get(0).getKeyType());
+    operandType.add(opBinding.collectOperandTypes().get(0).getValueType());
+    Pair<@Nullable RelDataType, @Nullable RelDataType> type =
+        getComponentTypes(
+            opBinding.getTypeFactory(), operandType);
+
+    requireNonNull(type, "inferred map type");
```
macroguo-ghy, line 1162: type is never null

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1149,6 +1149,56 @@ private static RelDataType mapReturnType(SqlOperatorBinding opBinding) {
...
+    requireNonNull(type.getKey(), "inferred map key type");
+    requireNonNull(type.getValue(), "inferred value type");
+
+    if (!type.getValue().getSqlTypeName().getName().equals("UNKNOWN")
```
macroguo-ghy, line 1166: Compare the `SqlTypeName` enum, not string

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1458,14 +1508,14 @@ private static RelDataType deriveTypeMapConcat(SqlOperatorBinding opBinding) {
   @LibraryOperator(libraries = {SPARK})
   public static final SqlFunction MAP_KEYS =
       SqlBasicFunction.create(SqlKind.MAP_KEYS,
-          ReturnTypes.TO_MAP_KEYS_NULLABLE,
```
macroguo-ghy, line 1461: remove `ReturnTypes.TO_MAP_KEYS_NULLABLE` if we don't use it any more.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1458,14 +1508,14 @@ private static RelDataType deriveTypeMapConcat(SqlOperatorBinding opBinding) {
...
+          SqlLibraryOperators::mapKeyReturnType,
           OperandTypes.MAP);
 
   /** The "MAP_VALUES(map)" function. */
   @LibraryOperator(libraries = {SPARK})
   public static final SqlFunction MAP_VALUES =
       SqlBasicFunction.create(SqlKind.MAP_VALUES,
-          ReturnTypes.TO_MAP_VALUES_NULLABLE,
```
macroguo-ghy, line 1468: remove it if we don't use it any more.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1149,6 +1149,56 @@ private static RelDataType mapReturnType(SqlOperatorBinding opBinding) {
...
+  private static RelDataType mapValueReturnType(SqlOperatorBinding opBinding) {
+    List<RelDataType> operandType = new ArrayList<>();
+
+    operandType.add(opBinding.collectOperandTypes().get(0).getKeyType());
+    operandType.add(opBinding.collectOperandTypes().get(0).getValueType());
+    Pair<@Nullable RelDataType, @Nullable RelDataType> type =
+        getComponentTypes(
+            opBinding.getTypeFactory(), operandType);
+
+    requireNonNull(type, "inferred map type");
```
macroguo-ghy, line 1187: type is never null

macroguo-ghy (commented):
Before I read`adjustTypeForMapFunctionConstructor`, I want to know why we need this method? 
If `MAP` function throw an exception when component type not equal? Like `map('k1', cast(1 as tinyint), 'k2', cast(1 as bigint))`

caicancai:
> Before I read`adjustTypeForMapFunctionConstructor`, I want to know why we need this method? If `MAP` function throw an exception when component type not equal? Like `map('k1', cast(1 as tinyint), 'k2', cast(1 as bigint))`

Thank you for such a quick review, I will write a plan to illustrate it

caicancai:
Here you need to convert mapKeyType and mapValueType into the corresponding Biggest Type. According to the initial method, there is definitely no conversion of mapKeyType and mapValueType into the corresponding Biggest Type.
Why add an `adjustTypeForMapFunctionConstructor method`? Here I will give an example:

```java
map_values(map('foo', 1, 'bar', cast(1 as tinyint)))
```
test1
When performing node parsing, map('foo', 1, 'bar', cast(1 as tinyint)) will be regarded as a node, that is, `((SqlBasicCall) call.getOperandList().get(0)) `== map('foo', 1, 'bar', cast(1 as tinyint)),

I have to typecast mapKeyType and mapValueType via `((SqlBasicCall) call.getOperandList().get(0)).getOperandList()` to get what I want.

test2

I didn't find the corresponding method, so I wrote a new method

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: I have deleted

caicancai: done, thanks

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: done, thanks

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: done, thanks

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: done, thanks

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: done, thanks

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorUtil.java
```diff
@@ -1381,6 +1399,54 @@ private static void adjustTypeForMultisetConstructor(
...
+   * if map's value targetType is double, this method would ensure that the value of the
+   * first map are cast to double.
+   *
+   * @param evenType the {@link RelDataType} to which the operands at even positions should be cast
+   * @param oddType the {@link RelDataType} to which the operands at odd positions should be cast
+   * @param sqlCallBinding the {@link SqlCallBinding} containing the operands to be adjusted
+   */
+  private static void adjustTypeForMapFunctionConstructor(
+      RelDataType evenType, RelDataType oddType, SqlCallBinding sqlCallBinding) {
+    SqlCall call = sqlCallBinding.getCall();
```
macroguo-ghy, line 1426: Do you assume that `call` is a map constructor? I think the `call` can be a column , a complex expr (like `map_concat(map('foo', 1), map('bar', 2))`)...


If we have a table like
```sql
create table t (id int, map_column map<integer, integer>);
```

when we execute a query like
```sql
select map_keys(map_column) from t;
```

I think your code will throw an exeception. Can you add more tests for your PR?

macroguo-ghy (changes requested):

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorUtil.java
caicancai: Good catch. Although my function currently only has map_values and map_keys use, but I think I should consider the situation you mentioned

caicancai: @macroguo-ghy According to the official website, map_keys and map_values only allow map parameters.
捕获

I tried to execute this sql in spark-shell, but it reported an error.
捕获

caicancai: But map_contact does have the bug you mentioned. Map cannot be converted to the corresponding biggest type.

caicancai:
@macroguo-ghy I'm sorry, I made the wrong request. I thought I should resolve the conflict first

caicancai:
@macroguo-ghy If you have time, you can help me take a look. I think it is reasonable to only recognize map. Thank you.

macroguo-ghy:
> @macroguo-ghy If you have time, you can help me take a look. I think it is reasonable to only recognize map. Thank you.

Sorry, I'm very busy this week, I will review at the earliest this weekend. And you can ask others for help.

caicancai:
> > @macroguo-ghy If you have time, you can help me take a look. I think it is reasonable to only recognize map. Thank you.
> 
> @caicancai Sorry, I'm very busy this week, I will review at the earliest this weekend. And you can ask others for help.

Small problem, this is open source, maybe I have more questions

caicancai:
@mihaibudiu If you have time, can you review it for me? Thank you.

caicancai:
@mihaibudiu Sorry for following up on this PR so late. If you have time, can you help me review this PR, thank you

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1163,6 +1163,52 @@ private static RelDataType mapReturnType(SqlOperatorBinding opBinding) {
...
+    operandType.add(opBinding.collectOperandTypes().get(0).getValueType());
+    Pair<@Nullable RelDataType, @Nullable RelDataType> type =
+        getComponentTypes(
+            opBinding.getTypeFactory(), operandType);
+
+    requireNonNull(type.left, () -> "type left of " + type.left);
+    requireNonNull(type.right, () -> "type right of " + type.right);
+    if (type.left.getSqlTypeName() != SqlTypeName.UNKNOWN
+        && type.right.getSqlTypeName() != SqlTypeName.UNKNOWN) {
+      SqlValidatorUtil.adjustTypeForMapFunctionConstructor(type, opBinding);
```
mihaibudiu, line 1180: Is this function allowed to mutate the operators?
Can you point to another example which does that?
I would not expect type inference to modify the program.

caicancai: I think what you mean is whether there are other functions like map_values and map_key functions. I will test it later to determine whether there are similar functions.

caicancai: map_contains_keys?
```bash
scala> val df = spark.sql("select map_contains_key(map(1, 'a', 2, 'b'), 1.0)")
val df: org.apache.spark.sql.DataFrame = [map_contains_key(map(1, a, 2, b), 1.0): boolean]

scala> df.show()
+--------------------------------------+
|map_contains_key(map(1, a, 2, b), 1.0)|
+--------------------------------------+
|                                  true|
+--------------------------------------+
```

caicancai: I think I should modify this jira description, it seems that not only these two functions have problems.
We can listen to other people's opinions, what do you think?
@tanclary

mihaibudiu: I am asking whether the returnTypeInference function is allowed to change the opBinding.
Did you see other examples in the code where this happens?
Neither map_concat or map_entries do that.

caicancai: Sorry, I misunderstood what you meant. Of course there are other similar encodings
https://issues.apache.org/jira/browse/CALCITE-5976
https://issues.apache.org/jira/browse/CALCITE-5948

PR #3726, opened 2024-03-12
[CALCITE-5681] Support authorization via GRANT and REVOKE DDL commands

macroguo-ghy:
CALCITE-5681

### DIFF
--- a/core/src/main/java/org/apache/calcite/access/CalcitePrincipal.java
+++ b/core/src/main/java/org/apache/calcite/access/CalcitePrincipal.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.access;
+
+import org.apache.commons.lang.StringUtils;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.security.Principal;
+import java.util.Locale;
+
+import static org.apache.calcite.runtime.Utilities.hash;
+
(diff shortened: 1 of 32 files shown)


site/_docs/reference.md
```diff
@@ -3542,6 +3544,34 @@ dropFunctionStatement:
...
+
+revokeStatement:
+      REVOKE
+      {
+          privilege [, privilege ]*
+      |   ALL [ PRIVILEGES ]
+      }
+      ON
+      {
+          [ TABLE ] table [, table ]*
```
YiwenWu, line 3570: 1. Are only the `TABLE` type currently considered for resource types? Is this syntax more generic, such as using `resource_type`, which lists supported types: `TABLE`
2. Can `ALL TABLES IN SCHEMA` be equivalently represented as the  `SCHEMA` resource type?

site/_docs/reference.md
```diff
@@ -3595,3 +3625,31 @@ employee_typ(315, 'Francis', 'Logan', 'FLOGAN',
     '555.777.2222', DATE '2004-05-01', 'SA_MAN', 11000, .15, 101, 110,
      address_typ('376 Mission', 'San Francisco', 'CA', '94222'))
 {% endhighlight %}
+
+#### Granting and revoking privileges
+
+You can grant and revoke privileges on tables, views and materialized views.
```
YiwenWu, line 3631: 'tables, views and materialized views', do all their types use `TABLE`?

core/src/main/java/org/apache/calcite/access/CalcitePrincipal.java
```diff
@@ -0,0 +1,69 @@
...
+import static org.apache.calcite.runtime.Utilities.hash;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * CalcitePrincipal is a simple implementation of {@link Principal}
+ * that represents a user. CalcitePrincipal ignore the case of the
+ * name when comparing two principals.
+ */
+public class CalcitePrincipal implements Principal, Comparable<CalcitePrincipal> {
```
YiwenWu, line 35: To refer to a user can use `UserPrincipal`

core/src/main/java/org/apache/calcite/jdbc/CalciteSchema.java
```diff
@@ -71,6 +74,9 @@ public abstract class CalciteSchema {
   protected final NameSet functionNames;
   protected final NameMap<FunctionEntry> nullaryFunctionMap;
   protected final NameMap<CalciteSchema> subSchemaMap;
+
+  // tableName -> (user, access)
+  protected final NameMap<Map<Principal, SqlAccessType>> privilegeMap;
```
YiwenWu, line 79: Whether to use a separate class to maintain the relationship between `Principal`, `SqlAccessType`, and `resource_type(TABLE/SCHEMA/FUNCTION...)`. 
Using this way resource types can be more general and facilitate subsequent expansion.

PR #3728, opened 2024-03-12
Fix style check rule that breaks the build

mihaibudiu:
The main build seems to be broken, this is an attempt to fix it.
It seems to be caused by a buggy checkStyle rule.


snuyanzin (approved):
Thanks for looking into it 
LGTM once the tests are green

PR #3729, opened 2024-03-12
[CALCITE-6323] Serialize return type during RelJson.toJson(RexNode no…

tanclary:
…de) for SqlKind.SAFE_CAST

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
+++ b/core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
@@ -632,6 +632,7 @@ public class RelJson {
         switch (node.getKind()) {
         case MINUS:
         case CAST:
+        case SAFE_CAST:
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -632,6 +632,7 @@ public Object toJson(RexNode node) {
         switch (node.getKind()) {
         case MINUS:
         case CAST:
+        case SAFE_CAST:
```
mihaibudiu, line 635: How about TRY_CAST?

tanclary: TRY_CAST uses SqlKind.SAFE_CASt too

mihaibudiu (approved):

PR #3731, opened 2024-03-12
[CALCITE-6324] Type inferred for result of STDDEV, VAR_SAMP, etc. is incorrect

mihaibudiu:


mihaibudiu:
The issue solved by this PR has been marked as invalid.

PR #3732, opened 2024-03-13
[CALCITE-6325] Add LOG function (enabled in Mysql and Spark library)

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6325

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -211,7 +211,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG2;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_AND;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_OR;
-import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG_MYSQL_STYLE;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG_MYSQL_SPARK;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LPAD;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_CONCAT;
@@ -650,8 +650,8 @@ public class RexImpTable {
       map.put(LOG10, logImplementor);
 
       LogImplementor logMysqlImplementor =
-          new LogImplementor(BuiltInMethod.LOG_MYSQL.method);
-      map.put(LOG_MYSQL_STYLE, logMysqlImplementor);
+          new LogImplementor(BuiltInMethod.LOG_MYSQL_SPARK.method);
+      map.put(LOG_MYSQL_SPARK, logMysqlImplementor);
       map.put(LOG2, logMysqlImplementor);
 
       defineReflective(RAND, BuiltInMethod.RAND.method,
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -211,6 +211,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG2;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_AND;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_OR;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG_MS;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LPAD;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_CONCAT;
@@ -642,11 +643,13 @@ public class RexImpTable {
       defineMethod(EXP, BuiltInMethod.EXP.method, NullPolicy.STRICT);
       defineMethod(POWER, BuiltInMethod.POWER.method, NullPolicy.STRICT);
       defineMethod(ABS, BuiltInMethod.ABS.method, NullPolicy.STRICT);
-      defineMethod(LOG2, BuiltInMethod.LOG2.method, NullPolicy.STRICT);
 
       map.put(LN, new LogImplementor());
       map.put(LOG, new LogImplementor());
       map.put(LOG10, new LogImplementor());
+
+      map.put(LOG_MS, new LogMSImplementor());
+      map.put(LOG2, new LogMSImplementor());
 
       defineReflective(RAND, BuiltInMethod.RAND.method,
           BuiltInMethod.RAND_SEED.method);
@@ -4146,9 +4149,52 @@ public class RexImpTable {
         if (argValueList.size() == 2) {
           return list.append(argValueList.get(1));
         }
+        if (argValueList.size() == 1) {
+          return list.append(Expressions.constant(Math.exp(1)));
+        }
         // fall through
       case "LN":
         return list.append(Expressions.constant(Math.exp(1)));
+      case "LOG10":
+        return list.append(Expressions.constant(BigDecimal.TEN));
+      default:
+        throw new AssertionError("Operator not found: " + call.getOperator());
+      }
+    }
+  }
+
+  /** Implementor for the {@code LN}, {@code LOG}, and {@code LOG10} operators.
+   *
+   * <p>Handles all logarithm functions using log rules to determine the
+   * appropriate base (i.e. base e for LN).
+   */
+  private static class LogMSImplementor extends AbstractRexCallImplementor {
+    LogMSImplementor() {
+      super("logMS", NullPolicy.STRICT, true);
+    }
+
+    @Override Expression implementSafe(final RexToLixTranslator translator,
+        final RexCall call, final List<Expression> argValueList) {
+      return Expressions.call(BuiltInMethod.LOGMS.method, args(call, argValueList));
+    }
+
+    private static List<Expression> args(RexCall call,
+        List<Expression> argValueList) {
+      Expression operand0 = argValueList.get(0);
+      final Expressions.FluentList<Expression> list = Expressions.list(operand0);
+      switch (call.getOperator().getName()) {
+      case "LOG":
+        if (argValueList.size() == 2) {
+          return list.append(argValueList.get(1));
+        }
+        if (argValueList.size() == 1) {
+          return list.append(Expressions.constant(Math.exp(1)));
+        }
+        // fall through
+      case "LN":
+        return list.append(Expressions.constant(Math.exp(1)));
+      case "LOG2":
+        return list.append(Expressions.constant(2));
       case "LOG10":
         return list.append(Expressions.constant(BigDecimal.TEN));
       default:
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -644,12 +644,15 @@ public class RexImpTable {
       defineMethod(POWER, BuiltInMethod.POWER.method, NullPolicy.STRICT);
       defineMethod(ABS, BuiltInMethod.ABS.method, NullPolicy.STRICT);
 
-      map.put(LN, new LogImplementor(BuiltInMethod.LOG.method));
-      map.put(LOG, new LogImplementor(BuiltInMethod.LOG.method));
-      map.put(LOG10, new LogImplementor(BuiltInMethod.LOG.method));
+      LogImplementor logImplementor = new LogImplementor(BuiltInMethod.LOG.method);
+      map.put(LN, logImplementor);
+      map.put(LOG, logImplementor);
+      map.put(LOG10, logImplementor);
 
-      map.put(LOG_MYSQL_SPARK, new LogImplementor(BuiltInMethod.LOG_MYSQL_SPARK.method));
-      map.put(LOG2, new LogImplementor(BuiltInMethod.LOG_MYSQL_SPARK.method));
+      LogImplementor logMysqlSparkImplementor =
+          new LogImplementor(BuiltInMethod.LOG_MYSQL_SPARK.method);
+      map.put(LOG_MYSQL_SPARK, logMysqlSparkImplementor);
+      map.put(LOG2, logMysqlSparkImplementor);
 
       defineReflective(RAND, BuiltInMethod.RAND.method,
           BuiltInMethod.RAND_SEED.method);
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -211,7 +211,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG2;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_AND;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_OR;
-import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG_MS;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG_MYSQL_SPARK;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LPAD;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_CONCAT;
@@ -644,12 +644,12 @@ public class RexImpTable {
       defineMethod(POWER, BuiltInMethod.POWER.method, NullPolicy.STRICT);
       defineMethod(ABS, BuiltInMethod.ABS.method, NullPolicy.STRICT);
 
-      map.put(LN, new LogImplementor());
-      map.put(LOG, new LogImplementor());
-      map.put(LOG10, new LogImplementor());
+      map.put(LN, new LogImplementor("LOG"));
+      map.put(LOG, new LogImplementor("LOG"));
+      map.put(LOG10, new LogImplementor("LOG"));
 
-      map.put(LOG_MS, new LogMSImplementor());
-      map.put(LOG2, new LogMSImplementor());
+      map.put(LOG_MYSQL_SPARK, new LogImplementor("LOG_MYSQL_SPARK"));
+      map.put(LOG2, new LogImplementor("LOG_MYSQL_SPARK"));
 
       defineReflective(RAND, BuiltInMethod.RAND.method,
           BuiltInMethod.RAND_SEED.method);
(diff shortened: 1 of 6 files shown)


caicancai:
I also fixed some problems with bigquery's log function. For example, one parameter of the log function defaults to ln, and the same is true for mysql and spark.

caicancai:
@tanclary If you have time, can you help review the code? As you can see, I have not implemented the ln and log10 functions, thank you

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -211,6 +211,7 @@
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG2;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_AND;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_OR;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG_MS;
```
tanclary, line 214: We really should look into some other mechanism for operators implemented differently across different dialects. I feel that just modifying the name is confusing and not very sustainable (what if 2 dialects support 1 version and 2 support another? what do you name it then?) I know it's not in the scope of this change but just putting it out there.

tanclary: I think we could rethink how we use the @LibraryAnnotations because it seems like we have a straightforward way of referencing what dialects a given operator supports.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -642,12 +643,14 @@ Builder populate() {
       defineMethod(EXP, BuiltInMethod.EXP.method, NullPolicy.STRICT);
       defineMethod(POWER, BuiltInMethod.POWER.method, NullPolicy.STRICT);
       defineMethod(ABS, BuiltInMethod.ABS.method, NullPolicy.STRICT);
-      defineMethod(LOG2, BuiltInMethod.LOG2.method, NullPolicy.STRICT);
 
       map.put(LN, new LogImplementor());
       map.put(LOG, new LogImplementor());
       map.put(LOG10, new LogImplementor());
```
tanclary, line 650: Can you just create an instance of a LogMsImplementor that we can reuse? I think that's more efficient.

Would you mind doing the same for LogImplementor above?

Thanks.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4146,9 +4149,52 @@ private static List<Expression> args(RexCall call,
         if (argValueList.size() == 2) {
```
tanclary, line 4149: can the function only have 1 or 2 arguments? If so, let's rewrite this as an `if (size == 1) ... else ...`

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4146,9 +4149,52 @@ private static List<Expression> args(RexCall call,
...
+        return list.append(Expressions.constant(Math.exp(1)));
+      case "LOG10":
+        return list.append(Expressions.constant(BigDecimal.TEN));
+      default:
+        throw new AssertionError("Operator not found: " + call.getOperator());
+      }
+    }
+  }
+
+  /** Implementor for the {@code LN}, {@code LOG}, and {@code LOG10} operators.
```
tanclary, line 4166: This javadoc looks identical to the one above. If the comment can be the same there's a chance we probably don't need two different implementors.

Would you mind either: a) making the comment more clear so we know why this implementor is needed or b) putting it all in the original LogImplementor?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4146,9 +4149,52 @@ private static List<Expression> args(RexCall call,
...
+    LogMSImplementor() {
+      super("logMS", NullPolicy.STRICT, true);
+    }
+
+    @Override Expression implementSafe(final RexToLixTranslator translator,
+        final RexCall call, final List<Expression> argValueList) {
+      return Expressions.call(BuiltInMethod.LOGMS.method, args(call, argValueList));
+    }
+
+    private static List<Expression> args(RexCall call,
```
tanclary, line 4181: this looks duplicated can you point out what's different here?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2788,15 +2788,26 @@ public static double log(BigDecimal d0, BigDecimal d1) {
...
+  /** SQL {@code LOGMS(number, number2)} function applied to
+   * double and BigDecimal values. */
+  public static @Nullable Double logMS(double number, BigDecimal number2) {
+    return (number <= 0) ? null : log(number, number2);
+  }
+
+  /** SQL {@code LOGMS(number, number2)} function applied to
+   * BigDecimal and double values. */
+  public static @Nullable Double logMS(BigDecimal number, double number2) {
+    return (number.doubleValue() <= 0) ? null : log(number, number2);
```
tanclary, line 2805: just put this conditional logic in one of the methods, and then have the others cast to that type and use it.

example
public Double foo(int num) {
    return num * 2;
}

public Double foo (double num) {
    return foo((int) num);
}

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2205,6 +2205,14 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
           OperandTypes.NUMERIC_OPTIONAL_NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
+  /** The "LOG(numeric, numeric1)" function. Returns the base numeric1 logarithm of numeric. */
+  @LibraryOperator(libraries = {MYSQL, SPARK})
+  public static final SqlFunction LOG_MS =
```
tanclary, line 2210: This name is confusing to me, when I see it i think `MsSQL`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6295,6 +6295,20 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
     f.checkNull("log(10, cast(null as real))");
   }
 
+  @Test void testLogFuncByOneParameter() {
```
tanclary, line 6298: can you javadoc this method? Why not add these to `testLogFunc`?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6332,6 +6346,66 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
     f0.forEachLibrary(list(SqlLibrary.MYSQL, SqlLibrary.SPARK), consumer);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6325">[CALCITE-6325]
+   * Add LOG function (enabled in MYSQL, Spark library)</a>. */
```
tanclary, line 6351: nit: MySQL*, and can you make that consistent across this PR (commit message, jira, etc.)

tanclary (changes requested):

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: @tanclary I very much agree with your point of view, I am also thinking about this issue, maybe we should take the first step here with the log function. Do you think I should send an email to discuss it?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: Okay, looks like the abbreviation won't work, I'll change it to LOG_MYSQL_SPARK

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: done

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: remove

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: add testLogfunc

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: I'm not sure what it means, sorry

caicancai:
I added clary as co-author, thank you for his review and suggestions

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: Looks like there's something wrong with my approach, I'll fix it

caicancai:
Calcite's commit message does not seem to allow the fix keyword to begin with

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: fix

caicancai:
I tried to add LOG_Postgres(SqlFunctions.class, "log", long.class, long.class) to BuiltInMethod, trying to use method.name() to distinguish different databases, but the method of BuiltInMethod corresponds to a Sqlfunctions method, in the form of a map Save, it seems that it cannot be superimposed,

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
tanclary: I meant can you create an instance of the Log implementors and then reuse them? You're instantiating each implementor several times and they're all the same so we can just reuse the same one. 

Also would you mind not marking these as resolved until they've been reviewed? It's hard to tell what comments the newest changes address. Also would you mind waiting to squash your commits until the PR has been approved?

tanclary (changes requested):
@caicancai did you forget to push your changes? All the comments are marked as resolved but it looks identical to how it was last week. Re-ping me when it's ready!

caicancai:
> @caicancai did you forget to push your changes? All the comments are marked as resolved but it looks identical to how it was last week. Re-ping me when it's ready!

Really, I submitted two commits, can you see them?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: Yes, I did it like this. You can’t see it over there. It seems I have to submit a PR again.

tanclary: Oh very weird, they are showing up now. Sorry about that.

tanclary: It still doesn't look like you addressed the comment though so would you mind doing that? Thanks

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4127,14 +4130,14 @@ private static class LogicalNotImplementor extends AbstractRexCallImplementor {
    * <p>Handles all logarithm functions using log rules to determine the
    * appropriate base (i.e. base e for LN).
    */
-  private static class LogImplementor extends AbstractRexCallImplementor {
-    LogImplementor() {
-      super("log", NullPolicy.STRICT, true);
+  private static class LogImplementor extends MethodImplementor {
+    LogImplementor(Method method) {
+      super(method, NullPolicy.STRICT, true);
```
tanclary, line 4135: shouldn't we be passing a String here? Why change the class we're extending?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2788,15 +2788,26 @@ public static double log(BigDecimal d0, BigDecimal d1) {
     return Math.log(d0.doubleValue()) / Math.log(d1.doubleValue());
   }
 
-  /** SQL {@code LOG2(number)} function applied to double values. */
-  public static @Nullable Double log2(double number) {
-    return (number <= 0) ? null : log(number, 2);
+  /** SQL {@code LOGMySqlSpark(number, number2)} function applied to double values. */
```
tanclary, line 2791: Just call it LOG and mention it's for those libraries. Your comment it makes it seem like `LOGMySqlSpark` is a function.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2205,6 +2205,14 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
           OperandTypes.NUMERIC_OPTIONAL_NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
+  /** The "LOG(numeric, numeric1)" function. Returns the base numeric1 logarithm of numeric. */
```
tanclary, line 2208: call them `numeric1` and `numeric2` it's more consistent

site/_docs/reference.md
```diff
@@ -2785,6 +2785,7 @@ In the following:
 | b f s | LENGTH(string)                             | Equivalent to `CHAR_LENGTH(string)`
 | h s | LEVENSHTEIN(string1, string2)                | Returns the Levenshtein distance between *string1* and *string2*
 | b | LOG(numeric1 [, numeric2 ])                    | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present
+| m s | LOG(numeric1 [, numeric2 ])                  | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present
```
tanclary, line 2788: This description is the same as the one above. If it's different from the BQ function, the description should explain how.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
tanclary: Please write it as "MySQL", that's how they style it.

tanclary (changes requested):

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: Do you mean to solve everything with one new LogImplementor() instead of multiple instantiations?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: @tanclary I hope you can listen to my analysis here, because the functions of log and log_mysql_spark in sqlfunctions are different. If it is passed in through a string, I also need to judge whether the string is log or log_mysql_spark. If it is passed through the method method, I don’t seem to need to Do this. 

Later, we also need to consider the log adaptation of postgres. Its argValueList.size() == 1 situation is different from that of bigquery and mysql. At that time, I only need method.getname to judge and align for processing. I thought so

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai:
```java
LogImplementor logImplementor = new LogImplementor(BuiltInMethod.LOG.method);
map.put(LN, logImplementor);
map.put(LOG, logImplementor);
map.put(LOG10, logImplementor);

LogImplementor logMysqlSparkImplementor = new LogImplementor(BuiltInMethod.LOG_MYSQL_SPARK.method);
map.put(LOG_MYSQL_SPARK, logMysqlSparkImplementor);
map.put(LOG2, logMysqlSparkImplementor);
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: Add LOG function as MySql style  or Add LOG function (enabled in MYSQL*) ?

tanclary: Please write it as "MySQL"

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: done

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java

site/_docs/reference.md

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: Sorry, I was sick during this period and was delayed for some time.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
tanclary: That sounds fine but why are you extending MethodImplementor and not AbstractRexCallImplementor

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6332,6 +6338,53 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
     f0.forEachLibrary(list(SqlLibrary.MYSQL, SqlLibrary.SPARK), consumer);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6325">[CALCITE-6325]
+   * Add LOG function (enabled in MYSQL library)</a>. */
```
tanclary, line 6343: Please fix the capitalization

tanclary (changes requested):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: Mysql?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: @tanclary Hello，I have tried it before. Using AbstractRexCallImplementor means that I need to pass a parameter to determine whether the log function belongs to mysql or bigquery. I can completely locate it through method and method name. This does not require me to make a judgment (because method name can be directly mapped to Sqlfunctions method), I think MethodImplementor is more suitable in terms of code size and simplicity, thank you

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java

caicancai:
@tanclary PTAL

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2205,6 +2205,14 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
           OperandTypes.NUMERIC_OPTIONAL_NUMERIC,
           SqlFunctionCategory.NUMERIC);
 
+  /** The "LOG(numeric1, numeric2)" function. Returns the base numeric1 logarithm of numeric. */
+  @LibraryOperator(libraries = {MYSQL, SPARK})
```
tanclary, line 2209: the PR title just has MySQL.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2205,6 +2205,14 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
+  public static final SqlFunction LOG_MYSQL =
```
tanclary, line 2210: the name shouldn't just have MySQL if it's not specific to MySQL.

tanclary (changes requested):

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: @tanclary How should I reasonably describe this PR title?
Add LOG function (enabled in MYSQL, Spark library) or Add LOG function (enabled in MYSQL* library).
Before I was Add LOG function (enabled in MYSQL, Spark library), I thought you asked me to change it to Add LOG function (enabled in MYSQL library), it seems I didn’t understand what you meant, sorry

caicancai:
I use Mysql_Style to describe it. Do you think it is acceptable?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: I use Mysql_Style to describe it. Do you think it is acceptable?

tanclary: No. Please look at other similar function implementations to see how they handle multiple libraries.

caicancai: Other functions do not specifically adapt their log functions to the mysql and spark dialects like the log function.

caicancai: I found similar cases Add SHA256, SHA512 functions (enabled in BigQuery and PostgreSQL libraries), so I should change back to Add LOG function (enabled in MYSQL, Spark library)

tanclary: "Style" doesn't add much. The issue is you can't name a function`LOG_MYSQL` if that function is also used for Spark. Does that make sense?

caicancai: LOG_MYSQL_SPARK?

caicancai: What I wrote at the beginning was LOG_MYSQL_SPARK. It seems that I have always misunderstood what you meant. Sorry.

caicancai: @tanclary PTAL, thank you

caicancai:
@tanclary PATL, i think this pr can merge 1.37

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2788,15 +2788,28 @@ public static double log(BigDecimal d0, BigDecimal d1) {
...
+   * but return null when number2 is 0. */
+  public static @Nullable Double logMysqlSpark(double number, double number2) {
+    return (number <= 0) ? null : log(number, number2);
   }
 
-  /** SQL {@code LOG2(number)} function applied to
-   * BigDecimal values. */
-  public static @Nullable Double log2(BigDecimal number) {
-    return log2(number.doubleValue());
+  /** SQL {@code LOG(number, number2)} function applied to double and BigDecimal values.
```
tanclary, line 2797: Does this function only exist because MySQL and Spark handle null behavior differently? If so, you shouldn't need to add methods here. Just changed the NullPolicy for that operator.

tanclary (changes requested):
The problem with this PR is I don't know what I'm reviewing. I know you're adding functions but it's not clear why you need an operator separate from BigQuery. Right now, your JIRA case is completely blank so no reviewer, viewer, or user has any context on what you're doing.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: good idea， let me try. I'm very sorry that I didn't describe it clearly in the jira case, I will improve it.
closing.

PR #3735, opened 2024-03-15
[CALCITE-6333] NullPointerException in AggregateExpandDistinctAggregatesRule.doRewrite when rewriting filtered distinct aggregation

abhishekagarwal87:
Given that we are already creating a select distinct and moving the filter condition in the aggregate call to project, there is no need to set the filter argument in the aggregate call.

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -2149,6 +2149,27 @@ class RelOptRulesTest extends RelOptTestBase {
         .check();
   }
 
+  @Test void testDistinctWithFilterWithoutGroupByUsingJoin() {
+    final String sql = "SELECT SUM(comm), COUNT(DISTINCT sal) FILTER (WHERE sal > 1000)\n"
+                       + "FROM emp";
+    sql(sql)
+        .withRule(CoreRules.AGGREGATE_EXPAND_DISTINCT_AGGREGATES_TO_JOIN)
+        .check();
+  }
+
+  @Test void testMultipleDistinctWithSameArgsDifferentFilterUsingJoin() {
+    final String sql = "select deptno, "
+                       + "count(distinct sal) FILTER (WHERE sal > 1000), "
+                       + "count(distinct sal) FILTER (WHERE sal > 500) "
+                       + "from sales.emp group by deptno";
+    sql(sql)
+        .withRule(
+            CoreRules.AGGREGATE_EXPAND_DISTINCT_AGGREGATES_TO_JOIN,
+            CoreRules.AGGREGATE_PROJECT_MERGE
+        )
+        .check();
+  }
+
   @Test void testDistinctWithFilterWithoutGroupBy() {
     final String sql = "SELECT SUM(comm), COUNT(DISTINCT sal) FILTER (WHERE sal > 1000)\n"
         + "FROM emp";
--- a/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
@@ -2859,6 +2859,32 @@ LogicalAggregate(group=[{}], EXPR$0=[MIN($1) FILTER $3], EXPR$1=[COUNT($0) FILTE
 ]]>
     </Resource>
   </TestCase>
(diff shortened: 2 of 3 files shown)


caicancai:
Hi @abhishekagarwal87 Calcite's commit message does not seem to allow the fix keyword to begin with

abhishekagarwal87:
@caicancai I changed the commit message based on guidelines here - https://calcite.apache.org/develop/#contributing

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -2856,6 +2856,32 @@ LogicalAggregate(group=[{}], EXPR$0=[MIN($1) FILTER $3], EXPR$1=[COUNT($0) FILTE
...
+    <Resource name="planBefore">
+      <![CDATA[
+LogicalAggregate(group=[{}], EXPR$0=[SUM($0)], EXPR$1=[COUNT(DISTINCT $1) FILTER $2])
+  LogicalProject(COMM=[$6], SAL=[$5], $f2=[>($5, 1000)])
+    LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+]]>
+    </Resource>
+    <Resource name="planAfter">
+      <![CDATA[
+LogicalJoin(condition=[true], joinType=[inner])
```
mihaibudiu, line 2876: frankly, this plan looks wrong to me.
From my reading of the spec the filter should not be applied to the sum.

abhishekagarwal87: The filter is not being applied to the sum. There is an additional `project` for the count aggregator that looks at the result of the filter condition and projects another column that gets counted. If the filter were being applied to the sum, I would imagine this `case` project being on the left side of the join, too. 

I haven't looked at these plans before, but please let me know if I am misinterpreting them. I have also verified this fix in `apache/druid` and I could see the native druid plan adding the filter only to the count.

mihaibudiu: My mistake.

Should we strive to address [CALCITE-6332] in this PR as well? I saw you commented on it.
As far as I can tell, a rule can check some preconditions and refuse to be applied if they are not satisfied. That would be one way to solve 6332.

As for this particular fix, don't know what happens if you have multiple aggregates each with a different condition, as the other above comment from @kgyrtkirk points out.

One solution would be to refuse to apply this rule when such combinations of filters exist.

I also sent a question to the dev list about the ability to test such rules using Quidem tests. That would be great, but I don't know if it's possible yet.

abhishekagarwal87: That (refusing to let the JOIN flavour fire when query has grouping sets and distinct counts) should be done in a separate PR. I will try a test case for the scenario when multiple aggregates with different conditions are passed and see if Zoltan's suggestion fixes the issues.

mihaibudiu: What is the status of this PR? @abhishekagarwal87 are you still working on this?

abhishekagarwal87: It should be ready now.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -2149,6 +2149,29 @@ private void checkSemiOrAntiJoinProjectTranspose(JoinRelType type) {
...
+                       + "count(distinct sal) FILTER (WHERE sal > 500) "
+                       + "from sales.emp group by deptno";
+    sql(sql)
+        .withRule(
+            CoreRules.AGGREGATE_EXPAND_DISTINCT_AGGREGATES_TO_JOIN,
+            CoreRules.AGGREGATE_PROJECT_MERGE
+        )
+        .check();
+
+    testSortUnionTranspose();
```
kgyrtkirk, line 2172: this call to `testSortUnionTranspose` seem to invoke another test

now I see that this was there by mistake in [some earlier comment](https://github.com/apache/calcite/pull/3735#discussion_r1528352739) - its not necessary

kgyrtkirk (approved):
beyond that extra line it looks good :)

PR #3738, opened 2024-03-20
[CALCITE-6337] Distinguish naked measure inside and outside aggregation

barrkel:
Split nakedMeasures flag into two flags:
- nakedMeasuresInsideAggregatingQuery : allowed in group-by queries
- nakedMeasuresOutsideAggregatingQuery : allowed in non-group-by queries

The existing nakedMeasures flag is marked as deprecated. Its functionality is equivalent to setting the above two flags to the same value.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlValidator.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlValidator.java
@@ -912,15 +912,39 @@ public interface SqlValidator {
      */
     Config withLenientOperatorLookup(boolean lenient);
 
-    /** Returns whether the validator allows measures to be used without the
-     * AGGREGATE function. Default is true. */
-    @Value.Default default boolean nakedMeasures() {
+    /** Returns whether the validator allows measures to be used without
+     * AGGREGATE function in a non-aggregate query. Default is true.
+     */
+    @Value.Default default boolean nakedMeasuresInNonAggregateQuery() {
       return true;
     }
 
+    /** Sets whether the validator allows measures to be used without AGGREGATE
+     * function in a non-aggregate query.
+     */
+    Config withNakedMeasuresInNonAggregateQuery(boolean value);
+
+    /** Returns whether the validator allows measures to be used without
+     * AGGREGATE function in an aggregate query. Default is true.
+     */
+    @Value.Default default boolean nakedMeasuresInAggregateQuery() {
+      return true;
+    }
+
+    /** Sets whether the validator allows measures to be used without AGGREGATE
+     * function in an aggregate query.
+     */
+    Config withNakedMeasuresInAggregateQuery(boolean value);
+
     /** Sets whether the validator allows measures to be used without the
-     * AGGREGATE function. */
-    Config withNakedMeasures(boolean nakedMeasures);
+     * AGGREGATE function inside or outside aggregate queries.
+     * Deprecated: use the inside / outside variants instead.
+     */
+    @Deprecated // to be removed before 1.38
+    default Config withNakedMeasures(boolean nakedMeasures) {
+      return withNakedMeasuresInAggregateQuery(nakedMeasures)
+              .withNakedMeasuresInNonAggregateQuery(nakedMeasures);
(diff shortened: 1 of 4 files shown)


tanclary (changes requested):
Looks good to me aside from the test, can you surround the JIRA case in brackets in the PR description/commit message?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidator.java
```diff
@@ -912,15 +912,39 @@ interface Config {
      */
     Config withLenientOperatorLookup(boolean lenient);
 
-    /** Returns whether the validator allows measures to be used without the
-     * AGGREGATE function. Default is true. */
-    @Value.Default default boolean nakedMeasures() {
+    /** Returns whether the validator allows measures to be used without
+     * AGGREGATE function outside an aggregating query. Default is true.
+     */
```
julianhyde, line 917: move `*/` to end of previous line, to be consistent with formatting in this file

core/src/main/java/org/apache/calcite/sql/validate/SqlValidator.java
```diff
@@ -912,15 +912,39 @@ interface Config {
...
+    Config withNakedMeasuresInsideAggregatingQuery(boolean value);
+
     /** Sets whether the validator allows measures to be used without the
-     * AGGREGATE function. */
-    Config withNakedMeasures(boolean nakedMeasures);
+     * AGGREGATE function inside or outside aggregate queries.
+     * Deprecated: use the inside / outside variants instead.
+     */
+    @Deprecated
+    default Config withNakedMeasures(boolean nakedMeasures) {
```
julianhyde, line 944: Set a release for removal. I suggest `// to be removed before 1.38`. Measures are an experimental feature, so one minor release notice is sufficient.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidator.java
```diff
@@ -912,15 +912,39 @@ interface Config {
...
-    @Value.Default default boolean nakedMeasures() {
+    /** Returns whether the validator allows measures to be used without
+     * AGGREGATE function outside an aggregating query. Default is true.
+     */
+    @Value.Default default boolean nakedMeasuresOutsideAggregatingQuery() {
       return true;
     }
 
+    /** Sets whether the validator allows measures to be used without AGGREGATE
+     * function outside an aggregating query.
```
julianhyde, line 923: I think 'in a non-aggregate query' would be clearer than 'outside an aggregating query' (assuming I understand you correctly).

julianhyde: Also, we seem to use the term 'aggregate query' more than 'aggregating query' these days.

julianhyde (changes requested):

core/src/main/java/org/apache/calcite/sql/validate/SqlValidator.java
barrkel: There are 96 `/**` doc comments in this source file, 16 of which don't have `*/` on a separate line at the end, and of those, 10 are single-line doc comments. Only 6 doc comments are multi-line with `*/` at the end of the last line, rather than on a line on its own. I changed the style of this comment when editing to bring it into greater consistency with the majority of the rest of the file.

I didn't do an exahustive analysis, but a quick scan of other source files in this directory shows they generally have doc comments with `*/` at the end on a line of its own.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidator.java
barrkel: I think I addressed this phrasing.

barrkel:
> Looks good to me aside from the test, can you surround the JIRA case in brackets in the PR description/commit message?

This is done.

tanclary (approved):
Seems like you addressed all of my and @julianhyde 's comments, will merge soon !

PR #3740, opened 2024-03-22
[CALCITE-6282] Avatica ignores time precision when returning TIME results

mihaibudiu:
This is a counterpart to the Avatica fix for [CALCITE-6282] and will only work after that fix is merged and Calcite moves to using the new Avatica release.


YiwenWu (approved):
LGTM

PR #3741, opened 2024-03-26
[CALCITE-6343] Fix AS alias operator stripping MEASUREness from measures

barrkel:
When inferring the return type of an operator application, don't unwrap MEASURE types if the operator is an AS alias.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlOperator.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlOperator.java
@@ -539,8 +539,10 @@ public abstract class SqlOperator {
       }
 
       // MEASURE wrapper should be removed, e.g. MEASURE<DOUBLE> should just be DOUBLE
-      if (isMeasure(returnType) && returnType.getMeasureElementType() != null) {
-        returnType = Objects.requireNonNull(returnType.getMeasureElementType());
+      if (opBinding.getOperator().kind != SqlKind.AS
+          && isMeasure(returnType)
+          && returnType.getMeasureElementType() != null) {
+        returnType = returnType.getMeasureElementType();
       }
 
       if (operandTypeInference != null
--- a/core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
+++ b/core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
@@ -189,6 +189,17 @@ class RelDataTypeSystemTest {
     assertThat(dataType, is(innerType));
   }
 
+  /** <a href="https://issues.apache.org/jira/browse/CALCITE-6343">[CALCITE-6343]</a>
+   * Ensure that AS operator doesn't change return type of measures. */
+  @Test void testAsOperatorReturnTypeInferenceDoesNotRemoveMeasure() {
+    final SqlTypeFactoryImpl f = new Fixture().typeFactory;
+    RelDataType innerType = f.createSqlType(SqlTypeName.DOUBLE);
+    RelDataType measureType = f.createMeasureType(innerType);
+    RelDataType dataType =
+        SqlStdOperatorTable.AS.inferReturnType(f, Lists.newArrayList(measureType));
+    assertThat(dataType, is(measureType));
+  }
+
   @Test void testCustomDecimalPlusReturnTypeInference() {
     final SqlTypeFactoryImpl f = new Fixture().customTypeFactory;
     RelDataType operand1 = f.createSqlType(SqlTypeName.DECIMAL, 38, 10);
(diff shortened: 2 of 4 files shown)


core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
```diff
@@ -189,6 +189,15 @@ static class Fixture extends SqlTypeFixture {
     assertThat(dataType, is(innerType));
   }
 
+  @Test void testAsOperatorReturnTypeInferenceDoesNotRemoveMeasure() {
```
tanclary, line 192: Add javadoc above with JIRA link

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -4756,6 +4756,22 @@ void checkUserDefinedOrderByOver(NullCollation nullCollation) {
         .ok();
   }
 
+  /** Test case for:
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6343">[CALCITE-6343]
+   * AS alias operator strips MEASUREness from measures</a>.
```
tanclary, line 4761: Is "strips MEASURE wrapper" more descriptive?

core/src/main/java/org/apache/calcite/sql/SqlOperator.java
```diff
@@ -539,7 +539,9 @@ public RelDataType inferReturnType(
       }
 
       // MEASURE wrapper should be removed, e.g. MEASURE<DOUBLE> should just be DOUBLE
-      if (isMeasure(returnType) && returnType.getMeasureElementType() != null) {
+      if (opBinding.getOperator().kind != SqlKind.AS
+          && isMeasure(returnType)
+          && returnType.getMeasureElementType() != null) {
         returnType = Objects.requireNonNull(returnType.getMeasureElementType());
```
tanclary, line 545: Is this still required given that we've just checked the element type isn't null?

tanclary (changes requested):

YiwenWu:
test failed：
image

core/src/main/java/org/apache/calcite/sql/SqlOperator.java
barrkel: This your code! I'll remove the `requireNotNull` though.

barrkel:
I saw a failure with `SqlXmlFunctionsTest` which doesn't reproduce in the IDE and looks to be completely unrelated to my change, so I'm not sure what's going on there.

core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
barrkel: Done.

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
tanclary: Haha yeah, I meant is that a better issue summary?

core/src/main/java/org/apache/calcite/sql/SqlOperator.java
```diff
@@ -539,8 +539,10 @@ public RelDataType inferReturnType(
       }
 
       // MEASURE wrapper should be removed, e.g. MEASURE<DOUBLE> should just be DOUBLE
-      if (isMeasure(returnType) && returnType.getMeasureElementType() != null) {
-        returnType = Objects.requireNonNull(returnType.getMeasureElementType());
+      if (opBinding.getOperator().kind != SqlKind.AS
+          && isMeasure(returnType)
+          && returnType.getMeasureElementType() != null) {
+        returnType = returnType.getMeasureElementType();
```
tanclary, line 545: The checkerframework failure below does seem related to your change

tanclary (changes requested):

core/src/main/java/org/apache/calcite/sql/SqlOperator.java
barrkel: It's related to the change you asked me to make on your code, removing the not-null assertion.

tanclary: Might need to add it back. Seems like it's messing something up.

tanclary: Also I've found it can be a time-saver to try re-building before pushing new changes.

mihaibudiu:
This looks easy to fix, if you can handle the checker errors we could merge this in time for 1.38

PR #3746, opened 2024-04-01
[CALCITE-6348] ARRAY_OVERLAP with a NULL argument crashes the compiler

mihaibudiu:


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7126,6 +7126,9 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
         "Cannot apply 'ARRAYS_OVERLAP' to arguments of type 'ARRAYS_OVERLAP\\("
             + "<INTEGER ARRAY>, <BOOLEAN>\\)'. Supported form\\(s\\): 'ARRAYS_OVERLAP\\("
             + "<EQUIVALENT_TYPE>, <EQUIVALENT_TYPE>\\)'", false);
+    f.checkNull("arrays_overlap(null, null)");
```
caicancai, line 7129: @mihaibudiu Are you sure spark supports this writing method? My test failed in spark 3.4.

caicancai: 2024-04-02 23-10-05屏幕截图

mihaibudiu: I am not sure what you mean by that. This is a legal expression which shouldn't crash the compiler.

caicancai: 2024-04-02 23-10-05屏幕截图

caicancai: 2024-04-03 08-20-47屏幕截图

caicancai: > I am not sure what you mean by that. This is a legal expression which shouldn't crash the compiler.

It seems to be a bug in spark. I can help you test it on the latest code branch of spark to see if it can be fixed.

caicancai: @mihaibudiu I tested it on the spark master branch. It seems that the parameters in arrays_overlap are not allowed to be null, which will cause the program to crash.

mihaibudiu: which program crashes?

caicancai: There may be something wrong with my expression. It should not be said that the program crashed, but that the execution failed. sorry
2024-04-03 09-04-51屏幕截图

mihaibudiu: Whoever wrote that code is confusing the type "NULL" with the type "VOID". These are not the same thing at all. NULL is a type which has one value, whereas VOID has no values at all.
So there is at least one bug in their handling of types.
We can also reject this expression at compilation time in Calcite.
What happens if both arrays are NULL?

caicancai: The situation of arrays_overlap(null, null) is as follows
2024-04-03 09-12-28屏幕截图

caicancai:
By the way, there are still some bugs in Calcite's handling of map and array types. I'm fixing them, but no one seems to be reviewing [my jira.](https://issues.apache.org/jira/browse/CALCITE-6300)

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: then it's easy, we just reject arguments of type NULL for this function.
I will rework the implementation.

caicancai (approved):
LGTM

PR #3748, opened 2024-04-01
[CALCITE-6285] Function ARRAY_INSERT produces an incorrect result for…

mihaibudiu:
… negative indices

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -7008,18 +7008,20 @@ public class SqlOperatorTest {
         "[1, 2, null, 3]", "INTEGER ARRAY NOT NULL");
     f1.checkScalar("array_insert(array[2, 3, 4], 1, 1)",
         "[1, 2, 3, 4]", "INTEGER NOT NULL ARRAY NOT NULL");
-    f1.checkScalar("array_insert(array[1, 3, 4], -2, 2)",
+    f1.checkScalar("array_insert(array[1, 3, 4], -1, 2)",
+        "[1, 3, 4, 2]", "INTEGER NOT NULL ARRAY NOT NULL");
+    f1.checkScalar("array_insert(array[1, 3, 4], -3, 2)",
         "[1, 2, 3, 4]", "INTEGER NOT NULL ARRAY NOT NULL");
-    f1.checkScalar("array_insert(array[2, 3, null, 4], -5, 1)",
+    f1.checkScalar("array_insert(array[2, 3, null, 4], -6, 1)",
         "[1, null, 2, 3, null, 4]", "INTEGER ARRAY NOT NULL");
     // check complex type
     f1.checkScalar("array_insert(array[array[1,2]], 1, array[1])",
         "[[1], [1, 2]]", "INTEGER NOT NULL ARRAY NOT NULL ARRAY NOT NULL");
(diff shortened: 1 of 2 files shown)


caicancai (approved):
LGTM

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7008,18 +7008,20 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
         "[1, 2, null, 3]", "INTEGER ARRAY NOT NULL");
     f1.checkScalar("array_insert(array[2, 3, 4], 1, 1)",
         "[1, 2, 3, 4]", "INTEGER NOT NULL ARRAY NOT NULL");
-    f1.checkScalar("array_insert(array[1, 3, 4], -2, 2)",
+    f1.checkScalar("array_insert(array[1, 3, 4], -1, 2)",
```
caicancai, line 7011: I have a question, what will happen if the array length is 5, but I fill in -6? I can test it in spark

caicancai: @mihaibudiu Hello！
array_insert(array(1, 3, 4), -4, 2) Spark seems to allow this situation, should we add some test samples?
2024-04-02 23-10-05屏幕截图

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7008,18 +7008,20 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+        "[1, 3, 4, 2]", "INTEGER NOT NULL ARRAY NOT NULL");
+    f1.checkScalar("array_insert(array[1, 3, 4], -3, 2)",
         "[1, 2, 3, 4]", "INTEGER NOT NULL ARRAY NOT NULL");
-    f1.checkScalar("array_insert(array[2, 3, null, 4], -5, 1)",
+    f1.checkScalar("array_insert(array[2, 3, null, 4], -6, 1)",
```
caicancai, line 7015: Maybe we should write array(), spark does not allow array[]

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: the docs say that this is a prepend. there is a test for this below

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: these two forms are equivalent

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: Thank you for your reply

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: thank you for your reply

mihaibudiu:
@chucheng92 I think we agreed to follow the semantics from the newest version of spark, so if you have no objection I will merge this PR.

chucheng92 (approved):

PR #3757, opened 2024-04-11
[CALCITE-6340] RelBuilder drops traits when aggregating over duplicate projected fields

jduo:
Calculate the post-pruning RelTraitSet on the projection using TraitSet#apply(Mapping)

### DIFF
--- a/core/src/main/java/org/apache/calcite/tools/RelBuilder.java
+++ b/core/src/main/java/org/apache/calcite/tools/RelBuilder.java
@@ -2502,9 +2502,12 @@ public class RelBuilder {
           newProjects.add(project.getProjects().get(i));
           builder.add(project.getRowType().getFieldList().get(i));
         }
+
+        // This currently does not apply mappings correctly to the RelCollation due to
+        // https://issues.apache.org/jira/browse/CALCITE-6391
         r =
-            project.copy(cluster.traitSet(), project.getInput(), newProjects,
-                builder.build());
+            project.copy(project.getTraitSet().apply(targetMapping), project.getInput(),
+                newProjects, builder.build());
       } else {
         groupSetAfterPruning = groupSet;
         groupSetsAfterPruning = groupSets;
--- a/core/src/main/java/org/apache/calcite/util/Bug.java
+++ b/core/src/main/java/org/apache/calcite/util/Bug.java
@@ -215,6 +215,12 @@ public abstract class Bug {
    * Fix failing quidem tests for FORMAT in CAST</a> is fixed. */
   public static final boolean CALCITE_6375_FIXED = false;
 
+  /** Whether
+   * <a href="https://issues.apache.org/jira/browse/CALCITE/issues/CALCITE-6391">
+   * [CALCITE-6391] Apply mapping to RelCompositeTrait does not apply it to wrapped traits</a>
+   * is fixed. */
+  public static final boolean CALCITE_6391_FIXED = false;
+
   /**
    * Use this to flag temporary code.
    */
--- a/core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
@@ -26,6 +26,7 @@ import org.apache.calcite.plan.RelTraitDef;
 import org.apache.calcite.plan.RelTraitSet;
 import org.apache.calcite.rel.RelCollations;
 import org.apache.calcite.rel.RelDistributions;
+import org.apache.calcite.rel.RelFieldCollation;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.AggregateCall;
 import org.apache.calcite.rel.core.Correlate;
@@ -78,6 +79,7 @@ import org.apache.calcite.tools.RelRunner;
 import org.apache.calcite.tools.RelRunners;
 import org.apache.calcite.tools.RuleSet;
 import org.apache.calcite.tools.RuleSets;
+import org.apache.calcite.util.Bug;
 import org.apache.calcite.util.Holder;
 import org.apache.calcite.util.ImmutableBitSet;
 import org.apache.calcite.util.Pair;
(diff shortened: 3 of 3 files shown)


core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -2503,7 +2503,7 @@ private RelBuilder pruneAggregateInputFieldsAndDeduplicateAggCalls(
           builder.add(project.getRowType().getFieldList().get(i));
         }
         r =
-            project.copy(cluster.traitSet(), project.getInput(), newProjects,
+            project.copy(project.getTraitSet(), project.getInput(), newProjects,
```
adamkennedy, line 2506: Can we always propagate all traits? Since this happens when the fields of an existing project are pruned down, couldn't a Collation or Distribution trait end up being set incorrectly?

jduo: I guess to be safe we should just propagate the convention.

jduo: I've changed the code to only propagate the convention.

adamkennedy: I think there might be a slightly better way using Mapping, the column pruning is effectively a Mapping anyway, so if we were to make it actually a Mapping there's a method that will transform the TraitSet according to that mapping, I believe including Collation and Distribution

adamkennedy: This is what I was thinking of
`public RelTraitSet apply(Mappings.TargetMapping mapping)`

jduo: Done. The post-pruning RelTraitSet is now calculated using RelTraitSet.apply(Mapping) from the original project's RelTraitSet.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -2502,9 +2502,10 @@ private RelBuilder pruneAggregateInputFieldsAndDeduplicateAggCalls(
           newProjects.add(project.getProjects().get(i));
           builder.add(project.getRowType().getFieldList().get(i));
         }
+
         r =
-            project.copy(cluster.traitSet(), project.getInput(), newProjects,
-                builder.build());
+            project.copy(project.getTraitSet().apply(targetMapping), project.getInput(),
```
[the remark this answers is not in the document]
adamkennedy: If that's not the right one, it would result in corrupting the Collation of the underlying Project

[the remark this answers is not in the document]
jduo: It looks like it maps to the underlying project to me.
The same fields referenced in building the Mapping are used to construct the newProjects variable (the list of RexNodes in the new projection).

asolimando (changes requested):
@jduo overall LGTM but there are a couple of changes I'd like you to make before getting this in, please take a look

core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
```diff
@@ -547,6 +547,31 @@ private void checkSimplify(UnaryOperator<RelBuilder.Config> transform,
...
+                builder.alias(builder.field(1), "b"),
+                builder.alias(builder.field(2), "c"),
+                builder.alias(builder.field(1), "d"))
+            .aggregate(builder.groupKey(0, 1, 2, 1),
+                builder.aggregateCall(SqlStdOperatorTable.SUM,
+                    builder.field(0)))
+            .build();
+
+    // Verify that the project under the aggregate kept the EnumerableConvention.INSTANCE trait.
+    assertTrue(root.getInput(0).getTraitSet().contains(EnumerableConvention.INSTANCE));
```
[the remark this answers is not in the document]
jduo: Thanks! Good suggestions @asolimando

core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
```diff
@@ -547,6 +547,31 @@ private void checkSimplify(UnaryOperator<RelBuilder.Config> transform,
     assertThat(root, matcher);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6340">
+   * [CALCITE-6340] RelBuilder always creates Project with Convention.NONE during aggregate_.</a>.
```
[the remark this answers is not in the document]
jduo: That description better describes the problem without going into the weeds.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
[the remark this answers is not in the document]
jduo: The use of the mapping was to handle the collation and distribution traits.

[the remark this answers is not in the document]
asolimando: I see, then the current test is not complete enough, it must pass with your variant while it must fail if you remove the mapping application, can you rework it to make this happen?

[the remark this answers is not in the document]
jduo: I haven't figured out how to create a test that covers this -- do you know of a way to alter collations for nodes created by a RelBuilder?

I'm not entirely sure if using the mapping is necessary. When implementing this using the Mapping route vs. copy the traitset, I see a slight difference in the interner, but I don't think this would be user-facing:
image
The order of the entries in the MapMakerInternalMap differs a bit (the PRESERVE entry appears earlier).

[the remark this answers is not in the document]
jduo: @asolimando , what I've found now:
1. There doesn't appear to be a way to add a RelDistribution trait to the builder and have it apply to the projection. Adding a RelDistribution trait using builder.exchange() adds a LogicalExchange node on in between the mapped projections so the pruning block doesn't get triggered.
2. It is possible to use builder.sort() to add a RelCollation trait to the projection before the mapping. However the RelCollation is stored in a RelCompositeTrait, and that composite trait does not override apply(Mapping), it just uses the default implementation which is the identity mapping. This seems off though -- I don't get why it doesn't propagate to wrapped traits. I don't have enough background to tell if this is intentional or a bug. What are your thoughts here?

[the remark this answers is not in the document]
jduo: Hey @asolimando ,

I've changed the implementation to skip the mapping and removed the test for other traits as I had issues with them described above. Perhaps it'd be better to deal with other traits in a separate ticket.

core/src/main/java/org/apache/calcite/plan/RelCompositeTrait.java
```diff
@@ -116,4 +116,5 @@ public T trait(int i) {
   public int size() {
     return traits.length;
   }
+
```
asolimando, line 119: Can you remove this modification? It seems to have slipped through by mistake.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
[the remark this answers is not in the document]
asolimando: Can you file a ticket for the issue you encountered with composite traits and missing override of mapping?

You can add in the description the unit tests for collations, link this ticket and say that the mapping in such case would be needed but it is ineffective.

Regarding the current PR, once you have the other ticket, I'd just add a comment pointing to it where you have the `project.copy` part so to make sure that this is a known limitation for people down the line.

asolimando (changes requested):
Few modifications as described in the unresolved comments

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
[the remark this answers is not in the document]
jduo: I've created https://issues.apache.org/jira/browse/CALCITE-6391 for the propagation of apply() calls in RelCompositeTrait. Updating the code to reference this.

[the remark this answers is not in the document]
jduo: Done.

asolimando (commented):
@jduo, I have made some modifications to the PR in [this commit](https://github.com/asolimando/calcite/commit/27b5d879b8097bac8be5b53b27d776407e312094), can you take a look?

I have added a conditional check on the results based on the status of the ticket you filed (as it's customary in Calcite), and I have also refined some tests and added few more

jduo:
> @jduo, I have made some modifications to the PR in [this commit](https://github.com/asolimando/calcite/commit/27b5d879b8097bac8be5b53b27d776407e312094), can you take a look?
> 
> I have added a conditional check on the results based on the status of the ticket you filed (as it's customary in Calcite), and I have also refined some tests and added few more
> 
> EDIT: finally it seems that all traits are dropped by this bug, so I suggest to rename the jira ticket as "[[CALCITE-6340](https://issues.apache.org/jira/browse/CALCITE-6340)] RelBuilder drops traits when aggregating over duplicate projected fields" and update it in the final commit (when you will be squashing after getting approval) and wherever it is referenced in the code

I think the changes proposed are mostly OK, but the Bug entry should be for the new ticket (6391) instead of this one (6340).

core/src/main/java/org/apache/calcite/util/Bug.java
```diff
@@ -215,6 +215,12 @@ public abstract class Bug {
    * Fix failing quidem tests for FORMAT in CAST</a> is fixed. */
   public static final boolean CALCITE_6375_FIXED = false;
 
+  /** Whether
+   * <a href="https://issues.apache.org/jira/browse/CALCITE/issues/CALCITE-6391">
+   * [CALCITE-6340] Apply mapping to RelCompositeTrait does not apply it to wrapped traits</a>
```
asolimando, line 220:
```suggestion
   * [CALCITE-6391] Apply mapping to RelCompositeTrait does not apply it to wrapped traits</a>
```

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -2502,9 +2502,12 @@ private RelBuilder pruneAggregateInputFieldsAndDeduplicateAggCalls(
           newProjects.add(project.getProjects().get(i));
           builder.add(project.getRowType().getFieldList().get(i));
         }
+
+        // This currently does not apply mappings correctly to the RelCollation due to
+        // https://issues.apache.org/jira/browse/CALCITE-6391.
```
asolimando, line 2507:
```suggestion
        // https://issues.apache.org/jira/browse/CALCITE-6391
```

core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
```diff
@@ -1450,6 +1452,225 @@ private RexNode caseCall(RelBuilder b, RexNode ref, RexNode... nodes) {
...
+    if (Bug.CALCITE_6391_FIXED) {
+      assertThat(node.getInput(0).getTraitSet(), is(expectedTraitSet));
+    } else {
+      assertThat(node.getInput(0).getTraitSet().get(0), is(expectedTraitSet.get(0)));
+    }
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE/issues/CALCITE-6340">
+   * [CALCITE-6340] RelBuilder drops set conventions when aggregating over duplicate
```
asolimando, line 1504: The commit message here is stale

core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
```diff
@@ -1450,6 +1452,225 @@ private RexNode caseCall(RelBuilder b, RexNode ref, RexNode... nodes) {
...
```
asolimando, line 1556: The commit message here is stale too

asolimando (changes requested):
Few more minor changes and for me we are good to go.

Please don't force-push until the review process is on-going as it makes incremental reviewing more difficult.

jduo:
> Few more minor changes and for me we are good to go.
> 
> Please don't force-push until the review process is on-going as it makes incremental reviewing more difficult.

In this case I force-pushed because there was a merge conflict against the head. Would it be preferred if I waited for the review to finish before dealing with these?

core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
jduo: Fixed.

core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
jduo: Fixed.

asolimando:
> > Few more minor changes and for me we are good to go.
> > Please don't force-push until the review process is on-going as it makes incremental reviewing more difficult.
> 
> In this case I force-pushed because there was a merge conflict against the head. Would it be preferred if I waited for the review to finish before dealing with these?

Unless for cases in which this is breaking CI then it's indeed preferable for the review process to be over.

asolimando (approved):
LGTM, thanks @jduo for taking care of the comments, can you squash the commits into a single one and make sure the commit message matches exactly the title of the Jira ticket preceded by `[CALCITE-6340]` (you can check the guidelines [here](https://calcite.apache.org/develop/#contributing) if you are in doubt).

After that, I will wait for a couple more days in case someone has further comments, then I will merge.

jduo:
> LGTM, thanks @jduo for taking care of the comments, can you squash the commits into a single one and make sure the commit message matches exactly the title of the Jira ticket preceded by `[CALCITE-6340]` (you can check the guidelines [here](https://calcite.apache.org/develop/#contributing) if you are in doubt).
> 
> After that, I will wait for a couple more days in case someone has further comments, then I will merge.

Thanks @asolimando . I've squashed and updated the title.

PR #3758, opened 2024-04-12
[CALCITE-6265] Type coercion is failing for numeric values in prepared statements (follow-up)

rubenada:
Address some issues on the original patch (https://github.com/apache/calcite/commit/4e6a320bf476c4b3c313d86f8ff3c6edf4c2a578 , via PR https://github.com/apache/calcite/pull/3687 )
See details in CALCITE-6265

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
@@ -354,6 +354,9 @@ public class EnumUtils {
     if (!Types.needTypeCast(fromType, toType)) {
       return operand;
     }
+
+    // TODO use Expressions#convertChecked to throw exception in case of overflow (CALCITE-6366)
+
     // E.g. from "Short" to "int".
     // Generate "x.intValue()".
     final Primitive toPrimitive = Primitive.of(toType);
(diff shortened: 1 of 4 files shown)


rubenada:
@tindzk , could you please take a look at this PR, which addresses some issue from CALCITE-6265? (more info in the Jira ticket)

tindzk:
The changes look good to me and your approach is a lot cleaner. Thanks for looking into it.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
```diff
@@ -354,6 +354,9 @@ public static Expression convert(Expression operand, Type fromType,
     if (!Types.needTypeCast(fromType, toType)) {
       return operand;
     }
+
+    // TODO consider using Expressions#convertChecked to throw exception in case of overflow
```
asolimando, line 358: @rubenada we might want to file a ticket for this and replace `TODO` with the ticket reference

asolimando (approved):
LGTM, left only a small comment, thanks @rubenada for looking into this

rubenada:
Thanks for the feedback. I have created [CALCITE-6366](https://issues.apache.org/jira/browse/CALCITE-6366), and reference it on the PR code.
I'll merge in the next 24h (if there's no problem on CI), if no further remarks arrive

mihaibudiu (approved):

PR #3760, opened 2024-04-13
[CALCITE-6363] Introduce a rule to derive more filters from inner join …

frostruan:
[https://issues.apache.org/jira/browse/CALCITE-6363](CALCITE-6363)

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
@@ -71,8 +71,10 @@ import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexLocalRef;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexOver;
+import org.apache.calcite.rex.RexPermuteInputsShuttle;
 import org.apache.calcite.rex.RexProgram;
 import org.apache.calcite.rex.RexShuttle;
+import org.apache.calcite.rex.RexSimplify;
 import org.apache.calcite.rex.RexSqlStandardConvertletTable;
 import org.apache.calcite.rex.RexSubQuery;
 import org.apache.calcite.rex.RexToSqlNodeConverter;
@@ -93,6 +95,7 @@ import org.apache.calcite.sql.type.MultisetSqlType;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.tools.RelBuilder;
 import org.apache.calcite.tools.RelBuilderFactory;
+import org.apache.calcite.util.BitSets;
 import org.apache.calcite.util.ImmutableBitSet;
 import org.apache.calcite.util.Litmus;
 import org.apache.calcite.util.Pair;
@@ -121,6 +124,7 @@ import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.Deque;
 import java.util.HashMap;
@@ -130,7 +134,10 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.NavigableSet;
+import java.util.NoSuchElementException;
 import java.util.Set;
+import java.util.SortedMap;
+import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
```diff
@@ -0,0 +1,380 @@
...
+
+@Value.Enclosing
+public class JoinDeriveEquivalenceFilterRule
+    extends RelRule<JoinDeriveEquivalenceFilterRule.Config> implements TransformationRule {
+
+  public JoinDeriveEquivalenceFilterRule(Config config) {
+    super(config);
+  }
+
+  @Override public void onMatch(RelOptRuleCall call) {
```
jamesstarr, line 109: This should use the RelMetaDataQuery for extracting filters from below.  Please look at RelMetadataQuery::getPulledUpPredicates usage in ReduceExpressionsRule.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
```diff
@@ -0,0 +1,380 @@
...
+        deriveEquivalenceCondition(simplify, rexBuilder, originalCondition);
+
+    if (arePredicatesEquivalent(rexBuilder, simplify, originalCondition, newCondition)) {
+      // if originalCondition and newCondition are equivalent, it means that the current
+      // Filter has either been derived or there is no room for derivation. if so,
+      // then we can stop.
+      return;
+    }
+
+    final Filter newFilter = filter.copy(filter.getTraitSet(), filter.getInput(), newCondition);
```
jamesstarr, line 130: I think it preferable to generate:
```
FILTER
  JOIN
    FILTER
      ....
    FILTER
      ....
```

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
```diff
@@ -0,0 +1,380 @@
...
+import com.google.common.collect.Sets;
+
+import org.immutables.value.Value;
+
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ *  Planner rule that derives more equivalent predicates from inner
```
jamesstarr, line 51: You can transitively generate predicates for the null generating sides or left and right joins.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
```diff
@@ -0,0 +1,380 @@
...
+      // then we can stop.
+      return;
+    }
+
+    final Filter newFilter = filter.copy(filter.getTraitSet(), filter.getInput(), newCondition);
+
+    call.transformTo(newFilter);
+
+    // after derivation, the original filter can be pruned
+    call.getPlanner().prune(filter);
```
jamesstarr, line 135: I am pretty sure we do not want to do this.  If their is a large tree, then this could cause problems.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
```diff
@@ -0,0 +1,380 @@
...
+ *  <p>2. only support simple condition inference, such as: {$1 = $2, $2 = 10} =&gt; {$1 = 10},
+ *     can not handle complex condition inference, such as conditions with functions, like
+ *     {a = b, b = abs(c), c = 1} =&gt; {a = abs(1)}
+ *  <p>3. only support discomposed literal, for example
+ *     it can infer {$1 = $2, $1 = 10} =&gt; {$2 = 10}
+ *     it can not infer {$1 = $2, $1 = 10 + 10} =&gt; {$2 = 10 + 10}
+ */
+
+@Value.Enclosing
+public class JoinDeriveEquivalenceFilterRule
```
jamesstarr, line 102: This name is misleading since it is not just equivalency, no?

core/src/main/java/org/apache/calcite/rex/RexUtil.java
```diff
@@ -3191,4 +3231,75 @@ private static class SearchExpandingShuttle extends RexShuttle {
...
+
+  /**
+   * Reorders some of the operands in this expression so structural comparison.
+   *
+   * <p>This is mostly the same as {@link SubstitutionVisitor#canonizeNode}, except
+   * that comparator is based on the {@link SqlKind}'s ordinal rather than string
+   * representation lexicographic order, which allows constant to always be on the
+   * right side of the expression. See {@link RexNormalize#reorderOperands} for details.
+   */
+  public static RexNode canonizeNode(RexBuilder rexBuilder, RexNode expression) {
```
jamesstarr, line 3243: I think it preferable to have helper functions for interacting with the existing nodes then creating a copy.

jamesstarr (commented):
I would prefer a more complete solution that handled things like using predicate pull ups and handling the null producing joins.  If a partial solution is accepted, then it might discourage a more complete solution down the line due to the legacy rule providing more base for expanding the solution.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
frostruan: Thanks for pointing out. I'll check and learn.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
frostruan: After a new Filter is generated, FilterJoinRule will push down the predicate.

In fact, I agree with you and the current implementation is a compromise. In terms of the order of rule application, the rules in calcite are order-insensitive, which allows the rules to remain completely independent without any dependencies. But if the predicate is pushed down in this rule, then I think this rule will do too many things. I want to keep the rule as simple and atomic as possible, focusing on one thing, so Just generating a new Filter ends here.

If in the future, we can allow rules to declare their own application order (such as applying before/after other rules), and then perform topological sorting on the rules, then it may be enough to just generate a new Filter.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
frostruan: Ok. Let me try this.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
frostruan: ok, this might be too radical.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
frostruan: Yes, this is indeed a bit misleading. I originally wanted to express the derivation of rules based on equivalent rewriting. I will rename it.

core/src/main/java/org/apache/calcite/rex/RexUtil.java
frostruan: Ok. I'll fix this.

frostruan:
REALLY appreciate your review. @jamesstarr It's very helpful. I'll address these as soon as possible.

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
```diff
@@ -0,0 +1,380 @@
...
+    final RexSimplify simplify =
+        new RexSimplify(rexBuilder, RelOptPredicateList.EMPTY, RexUtil.EXECUTOR);
+
+    final RexNode originalCondition =
+        prepare(rexBuilder, filter.getCondition(), join.getCondition());
+
+    final RexNode newCondition =
+        deriveEquivalenceCondition(simplify, rexBuilder, originalCondition);
+
+    if (arePredicatesEquivalent(rexBuilder, simplify, originalCondition, newCondition)) {
```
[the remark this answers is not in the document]
frostruan: Would you mind give me an example ? Can't imagine situation where an infinite loop would occur now. :)

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
jamesstarr: If we had a better extraction method, then we would see an infinite loop.  This is my concern with such a partial solution.  It will be difficult to build on top.

`
SELECT * FROM t1, t2
WHERE t1.c1 = t2.c1
  AND ((t1.c1, t2.c1) IN ((1, 2), (3, 4), (5, 6), (7, 8), (9, 10))
    OR (t1.c2, t2.c1) IN ((3, 4), (5, 6), (7, 8))`
Given the above example ideally you would get something like the following tree:
`
  FILTER ((t1.c1, t2.c1) IN ((1, 2), (3, 4), (5, 6), (7, 8), (9, 10)) OR (t1.c2, t2.c1) IN ((3, 4), (5, 6), (7, 8)))
    JOIN on t1.c1 = t2.c1
      SCAN t1
      FILTER t2.c1 IN (2, 4, 6, 8, 10)
        SCAN t2
`

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
jamesstarr: Does this break anything if removed?  Why would it be to radical?

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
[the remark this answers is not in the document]
jamesstarr: See [here ](https://github.com/apache/calcite/pull/3760#discussion_r1581544227)

core/src/main/java/org/apache/calcite/rel/rules/JoinDeriveEquivalenceFilterRule.java
asolimando: What I see here reminds me of [HIVE-25758](https://issues.apache.org/jira/browse/HIVE-25758) and similar bugs around `JoinPushTransitivePredicatesRule`, for which you can find more details in this slide deck from a Calcite meetup from last year: https://www.slideshare.net/slideshow/debugging-planning-issues-using-calcites-builtin-loggers/256567632 (slides 45-53).

As @jamesstarr suggests, once you can pull up more predicates, the burden of converging falls onto `RexSimplify`, if it fails at simplifying the "redundant" part, the predicate will always be identified as a new predicate that you will push from one side of the join to the other, pushed down and merged with existing predicates, which will again be identified as new predicates and you go into a loop.

I agree that having extra power here could be dangerous, but as long as it's not part of the core rules and it stays optional, it would be good to have it IMO.

frostruan:
Hi all 
I tried to re-implement predicate derivation in a new commit. Compared with a formal PR, it is more like a POC now. Looking forward to your suggestions.

### Why this change was been made
As I commented on https://issues.apache.org/jira/browse/CALCITE-6363, predicate inference is not available in VolcanoPlanner currently, so this PR tries to propose a temporary solution to help implement predicate inference.

### Main changes
1. Introduce a new method in RelOptUtil to implement predicate inference, which basically copies RelMdPredicates.JoinConditionBasedPredicateInference, but makes some simplifications as follows:
  1.1 Because the predicate is passed directly from the join, rather than pulled up from the filter node after the predicate is pushed down, there is no need to convert RexInputRef
  1.2 FilterJoinRule will classify the generated predicates, so the new method does not need to classify the predicates, and just returns the generated predicates directly
2. Call this method in FilterJoinRule, just after simplifying Outer Joins
3. Introduced a new configuration item in FilterJoinRule to determine whether to enable predicate inference, and turned it off by default
4. Added some unit tests

### Impact
1. Predicate inference is turned off by default, so it has no impact on the current logic
2. Applicable to Inner Join, Outer Join and Semi-Anti Join

ViggoC:
@frostruan It would make the code easier to review if you could display  the diff of your implementation and RelMdPredicates.JoinConditionBasedPredicateInference,

PR #3761, opened 2024-04-15
[CALCITE-6269] Fix missing/broken BigQuery date-time format elements

Anthrino:


core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -253,12 +373,27 @@ public enum FormatElementEnum implements FormatElement {
...
       final Work work = Work.get();
       sb.append(work.yyFormat.format(date));
     }
   },
+  YYY("yyy", "Last 3 digits of year") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      String formattedYear = String.format(Locale.ROOT, "%d", calendar.get(Calendar.YEAR));
+      sb.append(formattedYear.substring(formattedYear.length() - 3));
```
mihaibudiu, line 394: what happens for a year like 50?

core/src/test/resources/sql/cast-with-format.iq
```diff
@@ -32,8 +32,15 @@ EXPR$0
...
+# Input that contains shuffled date without time
+select cast('12-2010-05' as timestamp format
+    'DD-YYYY-MM');
+EXPR$0
+2010-05-12 00:00:00
+!ok
+
 !if (false) {
-### disabled until Bug.CALCITE_6269_FIXED ###
+### unsupported format model/elements, test disabled ###
```
mihaibudiu, line 43: There is a way in Quidem tests to refer to a Bug, why don't we use that one?
The good thing about the previous version of this string is that you can grep for it, here it is not clear which bug tracks the problem.

core/src/test/resources/sql/cast-with-format.iq
```diff
@@ -361,7 +351,7 @@ EXPR$0
 select cast('2017-05-31' as
     timestamp FORMAT 'RRRR-MM-DD');
 EXPR$0
-2017-05-31 00:00:00
+0017-05-31 00:00:00
```
mihaibudiu, line 354: why is this correct?

core/src/test/resources/sql/cast-with-format.iq
```diff
@@ -979,7 +939,7 @@ EXPR$0
 select cast(date'2019-01-07' as varchar
     FORMAT 'WW');
 EXPR$0
-01
+02
```
mihaibudiu, line 942: what is happening here?

core/src/test/resources/sql/cast-with-format.iq
```diff
@@ -1155,7 +1108,7 @@ EXPR$0
 !ok
 
 !if (false) {
-### disabled until Bug.CALCITE_6269_FIXED ###
+### unsupported format model/elements, test disabled ###
```
mihaibudiu, line 1111: this seems the same as the test in the change above on line 1091 above which is enabled

Anthrino: Agreed I think I missed that one, good catch!

core/src/test/resources/sql/cast-with-format.iq
Anthrino: Thanks for catching that, I accidentally edited that line instead of one below.. have refactored some of this code.

core/src/test/resources/sql/cast-with-format.iq
Anthrino: @mihaibudiu good point, I was also unsure of this comment and ideally wanted to link JIRA issues which would describe the exact causes. But there are quite a number of problems with the current implementation which is causing these many tests to fail and would take time to document them all, what do you recommend be done in this scenario?

mihaibudiu: I think it's best to have an issue to keep track of all of them, even if they are different, and to have an easy way to find it in the code.

The syntax in .iq files is: `!if (fixed.calcite1048)`

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
Anthrino: You are right, the earlier method was reusing what was being done in `YYYY` and cropping that down to three digits. But the `YYYY` implementation was incorrect as it was reading an integer where the `Calendar` object stored the year, and that would return an int value and not a formatted year with 4 digits (showing the century number).
 Corrected the `YYYY` and `YYY` functions to return a proper formatted value in terms of digits, and added a method `pctY` to return values consistent to what BQ returns.

core/src/test/resources/sql/cast-with-format.iq
Anthrino: The `WW` operator returns the `week number of the year (00-53)`, and that depends on what is being set as the start day of a week (Sun v/s Mon). The variances you see here are because we set the week start on Sunday, but setting to Monday also fails a bunch of other tests, which is strange as you would expect all of them to be consistent. So we have to use either of them and update the tests accordingly.

mihaibudiu: Since this is BigQuery, they better give the same results.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -217,6 +291,34 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%d", (calendar.get(Calendar.MONTH) / 3) + 1));
     }
   },
+  AMPM("", "The time as Meridian Indicator in uppercase") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      sb.append(calendar.get(Calendar.AM_PM) == Calendar.AM ? "AM" : "PM");
```
tanclary, line 298: can you just call `.toString()` on calendar.get(am_pm)` or no

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -306,12 +446,22 @@ static Work get() {
...
-    final DateFormat ssFormat = new SimpleDateFormat(FF2.javaFmt, Locale.ROOT);
     final DateFormat sssFormat = new SimpleDateFormat(FF3.javaFmt, Locale.ROOT);
-    final DateFormat ssssFormat = new SimpleDateFormat(FF4.javaFmt, Locale.ROOT);
-    final DateFormat sssssFormat = new SimpleDateFormat(FF5.javaFmt, Locale.ROOT);
-    final DateFormat ssssssFormat = new SimpleDateFormat(FF6.javaFmt, Locale.ROOT);
     final DateFormat yyFormat = new SimpleDateFormat(YY.javaFmt, Locale.ROOT);
+    final DateFormat yyyyFormat = new SimpleDateFormat(YYYY.javaFmt, Locale.ROOT);
+
+    /** Util to return the full or abbreviated weekday name from date and expected TextStyle. */
+    public String getDayFromDate(Date date, TextStyle style) {
```
tanclary, line 454: should this be private

tanclary (commented):
Overall think this looks good ! I agree with @mihaibudiu 's comments and left a couple questions of my own.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
Anthrino: Checked, `Calendar` in an INT enum, so the returned value needs to be compared and output as a string explicitly, as the enum only contains its int value.

Anthrino: Apologies, not an enum but the class stores these calendar elements as member properties of type int, so the conversion to string is still necessary.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
Anthrino: Good catch, can be kept private updated it!

core/src/test/resources/sql/cast-with-format.iq
Anthrino: Updated the comments and disable condition to point to the calcite issue, thanks for suggesting this change!

Anthrino:
@mihaibudiu @tanclary addressed your comments, would appreciate another review on this, thanks!

tanclary (approved):
LGTM if @mihaibudiu doesn't have any comments I think this is good to go

mihaibudiu (approved):

PR #3762, opened 2024-04-16
[CALCITE-6313] Add POWER function for PostgreSQL

normanj-bitquill:
* The existing power function is moved to all non PostgreSQL libraries
* The new power function is only for PostgreSQL
* The new function returns a decimal if any argument is a decimal

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -2216,7 +2216,22 @@ public abstract class SqlLibraryOperators {
 
   @LibraryOperator(libraries = {BIG_QUERY, SPARK})
   public static final SqlFunction POW =
-      SqlStdOperatorTable.POWER.withName("POW");
+      SqlBasicFunction.create("POW",
+          ReturnTypes.DOUBLE_NULLABLE,
+          OperandTypes.NUMERIC_NUMERIC,
+          SqlFunctionCategory.NUMERIC);
+
+  /** The {@code POWER(numeric, numeric)} function.
+   *
+   * <p>The return type is {@code DECIMAL} if either argument is a
+   * {@code DECIMAL}. In all other cases, the return type is a double.
+   */
+  @LibraryOperator(libraries = { POSTGRESQL })
+  public static final SqlFunction POWER_PG =
+      SqlBasicFunction.create("POWER",
+          ReturnTypes.DECIMAL_OR_DOUBLE_NULLABLE,
+          OperandTypes.NUMERIC_NUMERIC,
+          SqlFunctionCategory.NUMERIC);
 
   /** The "TRUNC(numeric1 [, integer2])" function. Identical to the standard <code>TRUNCATE</code>
   * function except the return type should be a double if numeric1 is an integer. */
--- a/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
@@ -774,6 +774,32 @@ public abstract class ReturnTypes {
   };
 
   /**
+   * Type-inference strategy that returns DECIMAL if any of the arguments are DECIMAL. It
+   * will return DOUBLE in all other cases.
+   */
+  public static final SqlReturnTypeInference DECIMAL_OR_DOUBLE = opBinding -> {
+    boolean haveDecimal = false;
+    for (int i = 0; i < opBinding.getOperandCount(); i++) {
+      if (SqlTypeUtil.isDecimal(opBinding.getOperandType(i))) {
+        haveDecimal = true;
+        break;
+      }
+    }
+
+    if (haveDecimal) {
+      return opBinding.getTypeFactory().createSqlType(
+          SqlTypeName.DECIMAL,
+          17);
+    } else {
+      return RelDataTypeImpl.proto(SqlTypeName.DOUBLE, false)
+          .apply(opBinding.getTypeFactory());
+    }
+  };
+
+  public static final SqlReturnTypeInference DECIMAL_OR_DOUBLE_NULLABLE =
+      DECIMAL_OR_DOUBLE.andThen(SqlTypeTransforms.TO_NULLABLE);
+
+  /**
    * Type-inference strategy whereby the result type of a call is
    * {@link #DECIMAL_SCALE0} with a fallback to {@link #ARG0} This rule
    * is used for floor, ceiling.
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -232,6 +232,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.PARSE_TIME;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.PARSE_TIMESTAMP;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.PARSE_URL;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.POW;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.POWER_PG;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.RANDOM;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.REGEXP;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.REGEXP_CONTAINS;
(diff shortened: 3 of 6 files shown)


normanj-bitquill:
In order to separate the POWER function for PostgreSQL and non-PostgreSQL dialects, the POWER function was moved to the `SqlLibraryOperators` class. This downgrades the POWER function from a standard function to a function from a library. This is problematic for things like SQRT which is a standard function and gets rewritten to a POWER expression.

Is there a more clean way to separate the POWER function so that the PostgreSQL implementation is able to return Decimal values?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2216,7 +2216,39 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
+      SPARK })
+  public static final SqlFunction POWER =
+      SqlBasicFunction.create("POWER",
+          ReturnTypes.DOUBLE_NULLABLE,
+          OperandTypes.NUMERIC_NUMERIC,
+          SqlFunctionCategory.NUMERIC);
+
+  /** The {@code POWER(numeric, numeric)} function.
+   *
+   * <p>The return type is always {@code DOUBLE} since we don't know
```
mihaibudiu, line 2241: this comment seems wrong

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -410,6 +410,24 @@ public static SqlCall stripSeparator(SqlCall call) {
   public static final SqlReturnTypeInference DOUBLE_FORCE_NULLABLE =
       DOUBLE.andThen(SqlTypeTransforms.FORCE_NULLABLE);
 
+  /**
```
mihaibudiu, line 413: this comment seems wrong too

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -773,6 +791,28 @@ public static SqlCall stripSeparator(SqlCall call) {
     return null;
   };
 
+  public static final SqlReturnTypeInference DECIMAL_OR_DOUBLE = opBinding -> {
```
mihaibudiu, line 794: why have both DOUBLE_OR_DECIMAL and DECIMAL_OR_DOUBLE?

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -559,15 +558,6 @@ static SqlOperatorTable operatorTableFor(SqlLibrary library) {
   }
 
   @Test void testArithmeticOperatorsFails() {
-    expr("^power(2,'abc')^")
```
mihaibudiu, line 562: why delete these tests? What happens to them? They look like good tests.

core/src/test/resources/sql/misc.iq
```diff
@@ -2301,17 +2301,6 @@ where false;
 
 !ok
 
-# [CALCITE-2447] POWER, ATAN2 functions fail with NoSuchMethodException
```
mihaibudiu, line 2304: why delete this test?

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -2006,7 +2006,7 @@ void checkPeriodPredicate(Checker checker) {
     expr("1-2+3*4/5/6-7")
         .ok("(((1 - 2) + (((3 * 4) / 5) / 6)) - 7)");
     expr("power(2,3)")
-        .ok("POWER(2, 3)");
+        .ok("`POWER`(2, 3)");
```
mihaibudiu, line 2009: these quotes look suspicious

mihaibudiu (changes requested):

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: Updated the comment.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
normanj-bitquill: Moved the comment to `DECIMAL_OR_DOUBLE`. Removed `DOUBLE_OR_DECIMAL`.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
normanj-bitquill: Removed `DOUBLE_OR_DECIMAL`.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
normanj-bitquill: The `POWER` function is no longer available in `SqlStdOperatorTable`. The implementation to use depends on which library is active since PostgreSQL can use a different return type.

This test fails as it was since `POWER` is not in `StdOperatorTable`. Is there a way to specify the library to use for the test?

mihaibudiu: Yes, see the operatorTableFor() function.

normanj-bitquill: Updated.

core/src/test/resources/sql/misc.iq
normanj-bitquill: The `POWER` function is no longer available in `SqlStdOperatorTable`. Is there any way of specifying a library to use here?

mihaibudiu: I don't think there is, but the question is whether there should be a standard power function.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
normanj-bitquill: This will end up going through here:
https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/SqlUtil.java#L393

and only look in `SqlStdOperatorTable` for the function.

mihaibudiu: Then it's related to my previous question: should there be a variant of this function in the standard operator table?
If not, I guess this is fine.

normanj-bitquill: Is it possible to have a power function in the standard operator table and still have an override for PostgreSQL that is able to return a decimal in some cases?

mihaibudiu: I don't know the answer to this question. But I know that Calcite can load multiple libraries, and I believe a function will be resolved by trying the libraries in order. I also am assuming that there is nothing special about the standard operator table. So if you load the postgres operator table first, perhaps the right POWER will be used?

A quick experiment should be able answer this question.

If this works, we should probably document that the right way to implement a dialect is to load the dialect specific operator table first.

normanj-bitquill: I found where the `ChainedSqlOperatorTable` is created and was able to give the library operator table precedence over the standard operator table.

With this change, no changes are needed to existing tests.

mihaibudiu:
In general, until the changes are approved I would recommend adding new commits instead of force-pushing. It makes it easier for reviewers to see what's new. Then you can squash when the PR is approved.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperatorTableFactory.java
```diff
@@ -117,8 +117,7 @@ private SqlOperatorTable create(ImmutableSet<SqlLibrary> librarySet) {
     SqlOperatorTable operatorTable = SqlOperatorTables.of(list);
     if (standard) {
       operatorTable =
-          SqlOperatorTables.chain(SqlStdOperatorTable.instance(),
-              operatorTable);
+          SqlOperatorTables.chain(operatorTable, SqlStdOperatorTable.instance());
```
mihaibudiu, line 120: This may cause problems for other people. Ideally you want to do this just for your tests.

normanj-bitquill: Understood. I'll take a look.

normanj-bitquill: This has been fixed. The only changes needed now are in the unit test itself. I setup the operator table there such that PostgreSQL is first.

mihaibudiu (approved):
The second commit should have a descriptive name, e.g., "only update loading library order for tests." But that is moot now, since I think this is ready for merging, so the next step is to squash the commits. We can't merge this until 1.37 is released, though.

PR #3764, opened 2024-04-18
[CALCITE-6169] EnumUtils.convert does not implement the correct SQL cast semantics

mihaibudiu:
This is a piece of https://github.com/apache/calcite/pull/3589 which only contains the fix to CALCITE-6169
This also fixes CALCITE-6366
The main part of the fix is implementing Expressions.convertChecked.
There was a recent implementation of this method, but we replace it in the generated Java code with a call to Primitive.numberValue, which covers more cases than the existing implementation.

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
@@ -69,6 +69,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.sql.Date;
 import java.sql.Time;
 import java.sql.Timestamp;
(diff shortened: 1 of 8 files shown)


mihaibudiu:
It would be nice to get a review on this PR, this is part of a long series of fixes to the implementation of type conversions in Calcite.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
```diff
@@ -481,10 +482,23 @@ public static Expression convert(Expression operand, Type fromType,
...
             operand, toPrimitive.getPrimitiveClass());
       }
-      if (fromNumber || fromBox == Primitive.CHAR) {
+      if (fromType == BigDecimal.class && toPrimitive.isFixedNumeric()) {
+        // Conversion from decimal to an exact type
+        ConstantExpression zero = Expressions.constant(0);
+        // Elsewhere Calcite uses this rounding mode implicitly, so we have to be consistent.
+        // E.g., this is the rounding mode used by BigDecimal.longValue().
+        Expression rounding = Expressions.constant(RoundingMode.DOWN);
+        // rounded = operand.setScale(0, RoundingMode.DOWN);
```
NobiGo, line 497: Keep consistent with other comment, Maybe should start with 
```
// Generate
```

core/src/main/java/org/apache/calcite/sql/SqlNumericLiteral.java
```diff
@@ -99,15 +99,22 @@ public boolean isExact() {
...
-        SqlTypeName result;
-        long l = bd.longValue();
-        if ((l >= Integer.MIN_VALUE) && (l <= Integer.MAX_VALUE)) {
-          result = SqlTypeName.INTEGER;
-        } else {
-          result = SqlTypeName.BIGINT;
+        try {
+          BigDecimal bd = getValueNonNull();
+          SqlTypeName result;
+          // Will throw if the number cannot
```
NobiGo, line 105: Maybe here does not need a break.

core/src/main/java/org/apache/calcite/sql/SqlNumericLiteral.java
```diff
@@ -99,15 +99,22 @@ public boolean isExact() {
...
+          long l = bd.longValue();
+          if ((l >= Integer.MIN_VALUE) && (l <= Integer.MAX_VALUE)) {
+            result = SqlTypeName.INTEGER;
+          } else {
+            result = SqlTypeName.BIGINT;
+          }
+          return typeFactory.createSqlType(result);
+        } catch (ArithmeticException ex) {
+          // This indicates that the value does
+          // not fit in any integer type.  Fallback to DECIMAL.
```
NobiGo, line 116: Same as above we don't need break line and before `Fallback`  there are extra spaces.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
```diff
@@ -481,10 +482,23 @@ public static Expression convert(Expression operand, Type fromType,
     if (toPrimitive != null) {
       if (fromPrimitive != null) {
         // E.g. from "float" to "double"
-        return Expressions.convert_(
+        if (toPrimitive == Primitive.BOOLEAN) {
```
NobiGo, line 485: Maybe need to add code comment about Why boolean and other type is different?

NobiGo (commented):
Minor change need to confirm.

mihaibudiu:
@NobiGo I hope I have addressed your comments

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
NobiGo: I will check this.

NobiGo (commented):
Leave one question for confirmation. I will check.

NobiGo (approved):

PR #3766, opened 2024-04-19
[CALCITE-6361] Uncollect.deriveUncollectRowType throws AssertionFailure if the input …

mihaibudiu:
…data is not a collection


mihaibudiu:
This is a tiny PR which fixes a crash, should be easy to review.

PR #3770, opened 2024-04-24
[CALCITE-6384] Add ASF header to buildcache.yml, gradle-wrapper-validation.yml

zabetak:
…

### DIFF
--- a/.github/workflows/buildcache.yml
+++ b/.github/workflows/buildcache.yml
@@ -1,3 +1,17 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to you under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


asolimando (approved):
LGTM, thanks @zabetak for helping Calcite being compliant with ASF policies!

asolimando (approved):
LGTM, thanks @zabetak for helping Calcite being compliant with ASF policies!

.github/workflows/buildcache.yml
```diff
@@ -1,3 +1,17 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to you under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
```
snuyanzin, line 8: It is not directly related to this issue
However out of curiosity: should we change this url to `https` rather than having `http` ?
like https://www.apache.org/licenses/LICENSE-2.0

snuyanzin (approved):

.github/workflows/buildcache.yml
zabetak: We should follow https://www.apache.org/legal/src-headers.html#headers so not at this point. I recall some discussions about updating HTTP to HTTPS in the header but I don't remember in which mailing list that was.

zabetak:
@snuyanzin Since we are cutting another RC let me know if you want me to merge this to main or do it after.

snuyanzin:
@zabetak yep agree, I think it would be a good idea to merge it before another RC
thanks for working on it!

PR #3772, opened 2024-04-25
[CALCITE-6385] LintTest fails when run in source distribution

zabetak:
The testContributorsFileIsSorted and testMailmapFile tests fail cause they rely on git and there is no git info available in source distribution.

In fact the tests don't really need git and the paths to the .mailmap and contributor.yml files can be specified explicitly. This removes the git requirement and also makes the tests more efficient since there is no process calls and lookup involved.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -826,6 +826,7 @@ allprojects {
                 passProperty("user.country", "tr")
                 passProperty("user.timezone", "UTC")
                 passProperty("calcite.avatica.version", props.string("calcite.avatica.version"))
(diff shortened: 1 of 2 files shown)


zabetak:
I run the tests through IntelliJ, git repo, and source, and it works everytime.

rubenada (approved):

zabetak:
Thanks for the review @rubenada ! 

@snuyanzin I know that `main` is on code freeze so let me know if I can merge this or you have another workflow in mind.

PR #3773, opened 2024-04-25
[CALCITE-6358] Support all PostgreSQL 14 date/time patterns

normanj-bitquill:
* Splits the PostgreSQL toChar function off to its own function
* Does not implement FX or TM prefix
* Does not implement SP suffix
* Timezone patterns are supported but all datetimes are in local timezone


mihaibudiu:
The amount of testing is very impressive. These won't be easy to review. I assume they were checked against postgres.
You should address the failures reported by the CI.

normanj-bitquill:
@mihaibudiu 
> The amount of testing is very impressive. These won't be easy to review. I assume they were checked against postgres. You should address the failures reported by the CI.

These were tested against PostgreSQL 14. The CI issues have been addressed.

Is there anything that I can be done to make reviewing this easier?

mihaibudiu:
The statement that they were run against Postgres helps a lot. Thanks!

core/src/main/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatter.java
```diff
@@ -0,0 +1,660 @@
...
+    }
+  }
+
+  /**
+   * A format element that is able to produce a string from a date.
+   */
+  private interface FormatPattern {
+    /**
+     * Checks if the current position in the format string applies to this format
+     * element. Will produce a formatted string if matched that is the format element
```
mihaibudiu, line 59: I find the second phrase difficult to parse,maybe you can rephrase it.

core/src/main/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatter.java
```diff
@@ -0,0 +1,660 @@
...
+          "am", "pm"),
+      new StringFormatPattern(
+          dt -> dt.getHour() < 12 ? "A.M." : "P.M.",
+          "A.M.", "P.M."),
+      new StringFormatPattern(
+          dt -> dt.getHour() < 12 ? "a.m." : "p.m.",
+          "a.m.", "p.m."),
+      new NumberFormatPattern(dt -> {
+        final String formattedYear = String.format(Locale.ROOT, "%0,4d", dt.getYear());
+        if (formattedYear.length() == 4 && formattedYear.charAt(0) == '0') {
```
mihaibudiu, line 299: this implies that years above 10000 wont' be correct. Maybe that's within spec.

core/src/main/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatter.java
```diff
@@ -0,0 +1,660 @@
...
+      new NumberFormatPattern(
+          dt -> Integer.toString(dt.getDayOfWeek().getValue()),
+          "ID"),
+      new NumberFormatPattern(
+          dt -> {
+            final String yearString = Integer.toString(dt.get(IsoFields.WEEK_BASED_YEAR));
+            return yearString.substring(yearString.length() - 1);
+          },
+          "I"),
+      new StringFormatPattern(
```
mihaibudiu, line 391: I am not sure Calcite can handle dates BC, but maybe this code should not concern itself with that point.
But you may not be able to test this codepath end-to-end.

core/src/main/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatter.java
```diff
@@ -0,0 +1,660 @@
...
+          dt -> dt.get(ChronoField.ERA) == 0 ? "B.C." : "A.D.",
+          "B.C.", "A.D."),
+      new StringFormatPattern(
+          dt -> dt.get(ChronoField.ERA) == 0 ? "b.c." : "a.d.",
+          "b.c.", "a.d."),
+      new StringFormatPattern(
+          dt -> {
+            final String monthName =
+                dt.getMonth().getDisplayName(TextStyle.FULL,
+                Locale.getDefault());
```
mihaibudiu, line 407: Are you sure about the Locale?
I thought that the spec requires the month name to be in English.

core/src/main/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatter.java
```diff
@@ -0,0 +1,660 @@
...
+            return monthName.toLowerCase(Locale.ROOT);
+          },
+          "month"),
+      new StringFormatPattern(
+          dt -> {
+            final String monthName =
+                dt.getMonth().getDisplayName(TextStyle.SHORT,
+                Locale.ROOT);
```
mihaibudiu, line 432: Similar question

mihaibudiu (approved):

mihaibudiu:
I have approved, but also left some questions.

caicancai (commented):
I remember that calcite currently has support for the postgres to_char function. Should we remove the previous support?

babel/src/test/resources/sql/postgresql.iq
```diff
@@ -58,11 +58,341 @@ NAME, PAY_BY_QUARTER, SCHEDULE
...
+EXPR$0
+2
+!ok
+
+select to_char(timestamp '2022-06-03 13:15:48.678', 'RM');
+EXPR$0
+VI
+!ok
+
+select to_char(timestamp '2022-06-03 13:15:48.678', 'rm');
```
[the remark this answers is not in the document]
normanj-bitquill: There is a test for this on line 316. It tests Jan 1 of year 1 to see if it is a Monday.

core/src/test/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatterTest.java
```diff
@@ -0,0 +1,1238 @@
...
+
+  @Test void testDayFullUpperCase() {
+    final ZonedDateTime date1 = createDateTime(2024, 1, 1, 23, 0, 0, 0);
+    final ZonedDateTime date2 = createDateTime(2024, 3, 1, 23, 0, 0, 0);
+    final ZonedDateTime date3 = createDateTime(2024, 10, 1, 23, 0, 0, 0);
+
+    final Locale originalLocale = Locale.getDefault();
+    try {
+      Locale.setDefault(Locale.US);
+      assertEquals("   MONDAY", PostgresqlDateTimeFormatter.toChar("DAY", date1));
```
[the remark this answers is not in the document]
normanj-bitquill: From this table:
https://www.postgresql.org/docs/14/functions-formatting.html#FUNCTIONS-FORMATTING-DATETIME-TABLE

it says that days are blank padded to 9 characters.

core/src/main/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatter.java
normanj-bitquill: `%0,4d` means to print out at least 4 digits and formatted with commas. A value like `12345` will become `12,345`.

There is the special case, where we get a 4 digit string back that has at least one leading 0. In this case, Java will not add the comma. For example `1` will become `0001`. Need to insert the comma to match PostgreSQL behaviour.

core/src/main/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatter.java
normanj-bitquill: I see what you mean. It should default to English and I should only translate when the `TM` prefix is provided. I'll make the changes.

https://www.postgresql.org/docs/14/functions-formatting.html#FUNCTIONS-FORMATTING-DATETIME-TABLE

core/src/main/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatter.java
normanj-bitquill: See my answer above. I'll make the changes.

core/src/test/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatterTest.java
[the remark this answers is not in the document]
normanj-bitquill: The padding should be at the end. I'll update.

normanj-bitquill:
@caicancai 

> I remember that calcite currently has support for the postgres to_char function. Should we remove the previous support?

The existing TO_CHAR function is used by MySQL and Oracle, so it is left in here. This PR adds a new implementation that only PostgreSQL uses.

https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java#L1654

core/src/main/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatter.java
normanj-bitquill: Updated to make use of the TM flag.

core/src/main/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatter.java
normanj-bitquill: Updated to make use of the TM flag.

core/src/main/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatter.java
normanj-bitquill: I've tried to reword this.

core/src/test/java/org/apache/calcite/util/format/PostgresqlDateTimeFormatterTest.java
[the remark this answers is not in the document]
normanj-bitquill: Updated.

normanj-bitquill:
@mihaibudiu 

> Until https://issues.apache.org/jira/browse/CALCITE-6281 has a solution, the best way to test this may be to create a Quidem test file (.iq suffix). If that file is identical with the Postgres shell output, you're done. (I have no idea whether the quidem output is compatible with Postgres, though).

Thanks, I'll take a look.

normanj-bitquill:
@mihaibudiu  Here is the output of a PostgreSQL script running all of the `TO_CHAR` tests from the `postgresql.iq` file in this PR. I kept the expected output for Calcite in SQL comments to make comparison easier.

This is the SQL script.
[to_char.txt](https://github.com/apache/calcite/files/15168101/to_char.txt)

This is the output when running it on PostgreSQL 14.
[to_char_results.txt](https://github.com/apache/calcite/files/15168106/to_char_results.txt)

mihaibudiu:
This is pretty close, the existing file `core/src/test/resources/sql/sub-query.iq` already uses the postgres format.
What I would do is to make a small python script to convert the postgres output into a .iq file, which you can hopefully just add to the resources directory in java. Then, before you commit, make sure the tests in the file are being executed by introducing a typo. And please add a comment to the iq file stating the Postgres version that was used to generate it. I am not sure whether there is some place where the python script itself could be committed...

normanj-bitquill:
@mihaibudiu 
> This is pretty close, the existing file `core/src/test/resources/sql/sub-query.iq` already uses the postgres format. What I would do is to make a small python script to convert the postgres output into a .iq file, which you can hopefully just add to the resources directory in java. Then, before you commit, make sure the tests in the file are being executed by introducing a typo. And please add a comment to the iq file stating the Postgres version that was used to generate it. I am not sure whether there is some place where the python script itself could be committed...

Added the queries here:
https://github.com/Bit-Quill/calcite/blob/calcite-6358/core/src/test/resources/pg_to_char_queries.sql

Here is a script that is able to generate the *.iq file:
https://github.com/Bit-Quill/calcite/blob/calcite-6358/core/src/test/resources/to_char_generate_iq.py

Ran the `:core:test` target with the generated *.iq file. Verified that the file is run by adding an error. The generated file is executed successfully with no errors.

mihaibudiu:
This looks very good, when they unfreeze the main branch we can merge this.
Regarding the script, it's good for this particular use case, but I am not sure whether it will generalize for arbitrary queries. But we can take this discussion to the other JIRA ticket (the preferred way to design solutions in Calcite is to discuss in JIRA). Some Calcite-isms leak into the iq file, like $EXPR0, so a true solution would probably have a different architecture.

caicancai (approved):
LGTM

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1651,13 +1652,22 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
           ReturnTypes.VARCHAR,
           OperandTypes.TIMESTAMP_STRING,
           SqlFunctionCategory.TIMEDATE);
 
+  /** The "TO_CHAR(timestamp, format)" function;
+   * converts {@code timestamp} to string according to the given {@code format}. */
+  @LibraryOperator(libraries = {POSTGRESQL})
+  public static final SqlFunction TO_CHAR_PG =
+      new SqlBasicFunction("TO_CHAR", SqlKind.OTHER_FUNCTION,
+      SqlSyntax.FUNCTION, true, ReturnTypes.VARCHAR, null,
```
arkanovicz, line 1667: Why not ReturnTypes.VARCHAR_NULLABLE here? The original Calcite TO_CHAR function result should also be nullable, by the way.

normanj-bitquill: Good point, this has been updated.

mihaibudiu:
I don't understand why the CI fails.
Can you please rebase on main? Maybe it has something to do with the recent transition to 1.37 which had some  build system changes.

normanj-bitquill:
> I don't understand why the CI fails. Can you please rebase on main? Maybe it has something to do with the recent transition to 1.37 which had some build system changes.

@mihaibudiu I have rebased to the latest for main. CI still fails for "Windows (JDK 8)" and "Windows (JDK 17)".

normanj-bitquill:
I have no idea about the Windows build errors. It fails trying to do task `:buildSrc:autostyleKotlinGradleCheck`. It fails since it can't find the class `org.jetbrains.kotlin.com.intellij.psi.tree.IElementType`.

I don't see why that is missing due to the changes for this PR.

normanj-bitquill:
Locally with a fresh checkout I get:
```
Skipping task ':buildSrc:autostyleKotlinGradleCheck' as it is up-to-date.
```

For the Windows build, I get:
```
:buildSrc:autostyleKotlinGradleCheck (Thread[Execution worker,5,main]) started.

> Task :buildSrc:autostyleKotlinGradleCheck FAILED
Caching disabled for task ':buildSrc:autostyleKotlinGradleCheck' because:
  Caching has not been enabled for the task
Task ':buildSrc:autostyleKotlinGradleCheck' is not up-to-date because:
  No history is available.
The input changes require a full rebuild for incremental task ':buildSrc:autostyleKotlinGradleCheck'.
```

normanj-bitquill:
@mihaibudiu I tried creating a new PR here, that did build correctly.
https://github.com/apache/calcite/pull/3792

If you are OK with it, we should just use that PR and close this one.

I think that there are a couple of things going on.
* I had squashed commits, which could be interfering with how changes are detected in autostyle. Lesson learned, I won't squash commits.
* Some Kotlin libraries are missing from the classpath for autostyle. Not an issue unless some changes are detected to a Kotlin file. The issue here may show up again if someone makes changes to `build.gradle.kts`

mihaibudiu:
Anything that works correctly is fine for me. Please note that I am on vacation so I won't have time to review this soon.

mihaibudiu:
Closing in favor of #3792

PR #3777, opened 2024-04-27
[CALCITE-6388] PsTableFunction throws NumberFormatException when the …

asolimando:
…'user' column has spaces

### DIFF
--- a/plus/src/main/java/org/apache/calcite/adapter/os/PsTableFunction.java
+++ b/plus/src/main/java/org/apache/calcite/adapter/os/PsTableFunction.java
@@ -26,13 +26,16 @@ import org.apache.calcite.schema.ScannableTable;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.util.Util;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * Table function that executes the OS "ps" command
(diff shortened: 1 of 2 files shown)


snuyanzin:
I guess the problem here that other columns also could contain space separated values like e.g. `comm`

I tried this `./sqlsh select distinct \`user\` from ps` for the branch for this PR
and the output is
```
6.9
13.3
0.0
0.1
1.0
0.2
1.1
0.3
1.2
69.7
39.5
0.4
0.5
0.6
0.7
108
4.4
0.8
1.8
35.8
0.9

```
while for main same query is
```
message+
serg
avahi
lp
kernoops
root
rtkit
syslog
systemd+
colord
```
after some checks I noticed that `command` could also contain space separated values and probably this is the reason...

caicancai (commented):
Can we add some tests to illustrate the jira case?

asolimando:
@snuyanzin, you are right, both `user` and `comm` can contain spaces, I have looked up if any of the other fields could be used in an unambiguous manner as an anchor to understand when `user` is over, unfortunately it doesn't seem to be any.

What I came up with is to move `user` as first field, split by whitespace as before, and stop parsing `user` whenever a purely numeric token is found. For `comm` I am just composing all the tokens counting on the fact that the number of fields is known.

So at the moment we support spaces, but not when one of the tokens is purely numeric like `root 123`. I don't have better ideas, it seems already an improvement over the current implementation which doesn't properly support spaces neither in `user` nor in `comm`.

@caicancai I have added a test, I had to unpack a bit the current implementation to that aim, thanks for the reminder.

Eager to hear your thoughts, thanks for your comments!

asolimando:
> @snuyanzin, you are right, both `user` and `comm` can contain spaces, I have looked up if any of the other fields could be used in an unambiguous manner as an anchor to understand when `user` is over, unfortunately it doesn't seem to be any.
> 
> What I came up with is to move `user` as first field, split by whitespace as before, and stop parsing `user` whenever a purely numeric token is found. For `comm` I am just composing all the tokens counting on the fact that the number of fields is known.
> 
> So at the moment we support spaces, but not when one of the tokens is purely numeric like `root 123`. I don't have better ideas, it seems already an improvement over the current implementation which doesn't properly support spaces neither in `user` nor in `comm`.
> 
> @caicancai I have added a test, I had to unpack a bit the current implementation to that aim, thanks for the reminder.
> 
> Eager to hear your thoughts, thanks for your comments!

Hello, if you don't have further comments I will merge around the end of the week

caicancai:
I will review this PR as soon as possible in the next two days

plus/src/test/java/org/apache/calcite/adapter/os/PsTableFunctionTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * Unit tests for the ps (process status) table function.
```
caicancai, line 33: Add jira link

plus/src/main/java/org/apache/calcite/adapter/os/PsTableFunction.java
```diff
@@ -43,114 +46,185 @@ public class PsTableFunction {
...
+  }
+
+  /**
+   * Class for parsing, line by line, the output of the ps command for a
+   * predefined list of parameters.
+   */
+  @VisibleForTesting
+  protected static class LineParser implements Function1<String, Object[]> {
+
+    @Override public Object[] apply(String line) {
```
caicancai, line 89: Good job

plus/src/main/java/org/apache/calcite/adapter/os/PsTableFunction.java
```diff
@@ -43,114 +46,185 @@ public class PsTableFunction {
...
+        StringBuilder sb = new StringBuilder();
+        while (processedTokens < tokens.length) {
+          sb.append(tokens[processedTokens++]).append(" ");
+        }
+        values[fieldIdx] =
+            field(PS_FIELD_NAMES.get(fieldIdx), sb.deleteCharAt(sb.length() - 1).toString());
+      } else {
+        values[fieldIdx] = field(PS_FIELD_NAMES.get(fieldIdx), tokens[processedTokens]);
+      }
+
```
caicancai, line 138: Remove extra blank lines

caicancai (approved):
overall LGTM，I left a few small comments

asolimando:
The force push is just to change the latest commit message (forgot that it must now start with an uppercase letter), no impact

caicancai (approved):

PR #3778, opened 2024-04-29
[CALCITE-6390] Exclude Arrow project on Windows builds

zabetak:


arrow/build.gradle.kts
```diff
@@ -36,6 +37,7 @@ dependencies {
 plugins.withType<JavaPlugin> {
     tasks {
         configureEach<Test> {
+            enabled = !OperatingSystem.current().isWindows
```
rubenada, line 40: It seems it is sufficient to exclude just the arrow tests, so I'd vote to go with this less drastic approach 👍  

Minor: perhaps we should add some comments, e.g.:
```
// Arrow test is disabled because it is not supported on Windows
// See https://arrow.apache.org/docs/java/install.html#system-compatibility
```

And since we are touching this file, maybe worth adding also some comments to the `jvmArgs` below:
```
// The following option `--add-opens=java.base/java.nio=ALL-UNNAMED` is required in jdk17+
// to avoid error. See https://arrow.apache.org/docs/java/install.html#java-compatibility
```

snuyanzin:
Thanks for working on this
agree with @rubenada 's comments
the rest is ok to me

arrow/build.gradle.kts
zabetak: Good idea. I added comments in https://github.com/apache/calcite/pull/3778/commits/c156046482bc929539fad70ff4a59dc64da219ed

asolimando (approved):
LGTM, just for the records in Cassandra JUnit extensions are used to achieve the same: [CassandraExtension.java#L135 ](https://github.com/apache/calcite/blob/3e633f627814ca79d1e2c592cd2ce1384717d6ca/cassandra/src/test/java/org/apache/calcite/test/CassandraExtension.java#L135), the Arrow's equivalent would be [ArrowExtension.java#L50](https://github.com/apache/calcite/blob/3e633f627814ca79d1e2c592cd2ce1384717d6ca/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowExtension.java#L50) 

I don't have any strong feelings towards any of the two approaches, so the current PR looks good as-is

EDIT: I suggest to update the title of the Jira ticket/commit to reflect the new strategy (disabling just the tests), so something along the line of "Disable Arrow tests for Windows"

rubenada (approved):

PR #3779, opened 2024-04-29
[CALCITE-6389] RexBuilder.removeCastFromLiteral does not preserve semantics for some…

mihaibudiu:
… types of literal

There are at least two other issues that I attempted to fix which uncover this bug, so I decided to make a separate PR for fixing it. This should make the other PRs simpler and less controversial.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
@@ -51,6 +51,7 @@ import org.checkerframework.checker.nullness.qual.EnsuresNonNullIf;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.nio.charset.Charset;
 import java.util.AbstractList;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
```diff
@@ -1822,6 +1823,35 @@ public static RelDataType extractLastNFields(RelDataTypeFactory typeFactory,
...
+   * @param toType A DECIMAL type.
+   * @return whether the value is valid for the type
+   */
+  public static boolean canBeRepresentedExactly(@Nullable BigDecimal value, RelDataType toType) {
+    assert toType.getSqlTypeName() == SqlTypeName.DECIMAL;
+    if (value == null) {
+      return true;
+    }
+    value = value.stripTrailingZeros();
+    if (value.scale() < 0) {
```
snuyanzin, line 1845: it would be great also to have a test with negative scale

mihaibudiu: Frankly negative scales are something I don't understand well.
Many SQL dialects do not accept negative scales, but Calcite allegedly does, although there aren't many tests about that.
Maybe we can file a separate issue about testing negative scales in Calcite, I expect it may uncover quite a few bugs.

mihaibudiu: I also think that whether negative scales are accepted or not should be a property of the type system.

snuyanzin: I haven't checked it however if you think that it is large area to work with, then separate issue could be probably also ok

mihaibudiu: Yes, I think that many places in Calcite assume implicitly that decimal scales cannot be negative.

mihaibudiu: what do we do about this PR in particular?

snuyanzin: as you mentioned if there is a separate issue for negative scale is required, could you please file one in Calcite jira?
All the rest is ok to me, thanks

snuyanzin (approved):

PR #3780, opened 2024-04-30
[CALCITE-6376]Filtering CTE of at least 6 columns with QUALIFY operation results in exception. 

DashShen:
This is a bug fix for issue https://issues.apache.org/jira/browse/CALCITE-6376

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/FlatLists.java
+++ b/core/src/main/java/org/apache/calcite/runtime/FlatLists.java
@@ -82,6 +82,11 @@ public class FlatLists {
     return new Flat6List<>(t0, t1, t2, t3, t4, t5);
   }
 
+  /** Creates a flat list with 7 elements. */
+  public static <T> List<T> of(T t0, T t1, T t2, T t3, T t4, T t5, T t6) {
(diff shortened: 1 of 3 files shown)


mihaibudiu:
Does this fix the bug in general, or just for 6 arguments?
If the bug persists for 7 arguments, what should be done about it?
Your commit should have the same text as the issue and PR.

DashShen:
This bug fix addresses the issue that arises when there are seven parameters.
However,  it is theoretically possible to encounter the same problem if the number of elements exceeds eight.

mihaibudiu:
Can there anything be done to handle the case for 8 or more without creating a class for every case?

DashShen:
> Can there anything be done to handle the case for 8 or more without creating a class for every case?

No more, I didn't find a better way to fix these cases for 8 or more parameters.

mihaibudiu:
If hope that #3781 solves the underlying problem

DashShen:
> If hope that #3781 solves the underlying problem
Yes, that works too.

mihaibudiu:
Closing in favor of #3781

PR #3781, opened 2024-04-30
[CALCITE-6376] Selecting 6 columns with QUALIFY operation results in exception

mihaibudiu:
This fix is simpler than #3780 and should cover all sizes, not just 7


asolimando (approved):
LGTM

PR #3782, opened 2024-05-01
[CALCITE-6395] Significant precision loss when representing REAL literals

mihaibudiu:


mihaibudiu:
This PR actually *removes* some code to fix a bug. Should be easy to review!

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1887,11 +1887,6 @@ public RexNode makeLambdaCall(RexNode expr, List<RexLambdaRef> parameters) {
               type.getSqlTypeName());
       return new BigDecimal(((Number) o).longValue());
     case REAL:
-      if (o instanceof BigDecimal) {
-        return o;
-      }
-      return new BigDecimal(((Number) o).doubleValue(), MathContext.DECIMAL32)
-          .stripTrailingZeros();
```
snuyanzin, line 1894: I wonder whether will we have similar issue with 
```sql
CAST(FLOAT AS VARCHAR)
```
https://github.com/apache/calcite/blob/19debd3e90e391bc6ba8250ec81735490b4b0cee/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java#L3511-L3520

mihaibudiu: Frankly I don't understand why in that place the float has to be converted to a decimal and then to a string. One can do the conversion directly. The question is whether I fix this in this PR or file another issue. If it's small enough I will bundle it here.

mihaibudiu: The truth is that float only has about 7 decimal digits of precision, so one could argue that the original answer provided by Calcite was legal. However, it makes it much easier to test code if the result provided by Calcite for `CAST(CAST string AS REAL) AS BIGINT)` is the same as the Java result of `(long)Float.parseFloat(string)`, at least when running on the same CPU.

mihaibudiu: I have fixed this path as well in a subsequent commit.
Just because Calcite uses Decimal literals to represent FP values doesn't mean it has to convert to Decimal when converting to strings.

snuyanzin (approved):

snuyanzin:
Thanks for fixing that, also checked that there is no more usages of `MathContext.DECIMAL32` in code

PR #3783, opened 2024-05-02
[CALCITE-6393] Add bytecode verification check using ASM in the gradle build

zabetak:


zabetak:
Closing this as its superceded and fixed in other PRs.

PR #3784, opened 2024-05-03
[CALCITE-6396] Add ADD_MONTHS function (enabled in Spark library)

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6396

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -270,6 +270,13 @@ public abstract class SqlLibraryOperators {
               .andThen(SqlTypeTransforms.TO_NULLABLE_ALL),
           OperandTypes.SAME_SAME);
 
+  /** The "NVL(value, value)" function. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlBasicFunction ADD_MONTHS =
+      SqlBasicFunction.create(SqlKind.ADD_MONTHS,
+          ReturnTypes.VARCHAR_NULLABLE,
+          OperandTypes.STRING_INTEGER);
+
   /** The "IFNULL(value, value)" function. */
   @LibraryOperator(libraries = {BIG_QUERY, SPARK})
   public static final SqlFunction IFNULL = NVL.withName("IFNULL");
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2674,6 +2674,7 @@ In the following:
 | p | expr :: type                                   | Casts *expr* to *type*
 | m | expr1 <=> expr2                                | Whether two values are equal, treating null values as the same, and it's similar to `IS NOT DISTINCT FROM`
 | * | ACOSH(numeric)                                 | Returns the inverse hyperbolic cosine of *numeric*
+| s | ADD_MONTHS(string, months)                      | Returns the date that is *months* after *string*
 | s | ARRAY([expr [, expr ]*])                       | Construct an array in Apache Spark. The function allows users to use `ARRAY()` to create an empty array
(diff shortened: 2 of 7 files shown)


site/_docs/reference.md
```diff
@@ -2674,6 +2674,7 @@ In the following:
 | p | expr :: type                                   | Casts *expr* to *type*
 | m | expr1 <=> expr2                                | Whether two values are equal, treating null values as the same, and it's similar to `IS NOT DISTINCT FROM`
 | * | ACOSH(numeric)                                 | Returns the inverse hyperbolic cosine of *numeric*
+| s | ADD_MONTHS(start_date, num_months)             | Returns the date that is *num_months* after *start_date*.
```
mihaibudiu, line 2677: does this operate on dates or strings?
the code assumes strings

caicancai: strings

caicancai: @mihaibudiu I don't know why cli reports an error

mihaibudiu: what cli?

caicancai: 2024-05-07 08-30-46屏幕截图

mihaibudiu: that is not a CLI, that is the CI

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -270,6 +270,13 @@ private static SqlCall transformConvert(SqlValidator validator, SqlCall call) {
               .andThen(SqlTypeTransforms.TO_NULLABLE_ALL),
           OperandTypes.SAME_SAME);
 
+  /** The "NVL(value, value)" function. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlBasicFunction ADD_MONTHS =
```
mihaibudiu, line 275: According to the Databricks documentation ADD_MONTHS takes an expression that evaluates to a date, and not a string: https://docs.databricks.com/en/sql/language-manual/functions/add_months.html

The result is also supposed to be a DATE.

I have already asked this question in a prior review. Reviews take a lot of energy and time, so please listen to the comments if you expect people to review your PRs.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10418,6 +10418,23 @@ void checkNvl(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
 
+  @Test void testAddMonthsFunc() {
+    final SqlOperatorFixture f0 = fixture();
+    f0.setFor(SqlLibraryOperators.ADD_MONTHS);
+    f0.checkFails("^add_months('2016-08-31', 1)^",
+        "No match found for function signature "
+            + "ADD_MONTHS\\(<CHARACTER>, <NUMERIC>\\)", false);
+
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
+    f.checkScalar("ADD_MONTHS('2016-08-31', 1)", "2016-09-30",
```
mihaibudiu, line 10429: you should test with invalid dates too, and with dates prior to the Gregorian calendar introduction.
You should also test with negative values for the number of months.

mihaibudiu (changes requested):

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: Thanks for reminding me, I will improve this PR

caicancai: Thank you

caicancai: I encountered a weird bug that seems to take some time

PR #3785, opened 2024-05-03
[CALCITE-6397] Add NVL2 function (enabled in Oracle, Spark library)

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6397

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlKind.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlKind.java
@@ -421,10 +421,13 @@ public enum SqlKind {
   /** {@code DECODE} function (Oracle). */
   DECODE,
 
-  /** {@code NVL} function (Oracle). */
+  /** {@code NVL} function (Oracle, Spark). */
   NVL,
 
-  /** {@code GREATEST} function (Oracle). */
+  /** {@code NVL2} function (Oracle, Spark). */
+  NVL2,
+
+  /** {@code GREATEST} function (Oracle, Spark). */
   GREATEST,
 
   /** The two-argument {@code CONCAT} function (Oracle). */
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -270,6 +270,14 @@ public abstract class SqlLibraryOperators {
               .andThen(SqlTypeTransforms.TO_NULLABLE_ALL),
           OperandTypes.SAME_SAME);
 
+  /** The "NVL2(value, value, value)" function. */
+  @LibraryOperator(libraries = {ORACLE, SPARK})
+  public static final SqlBasicFunction NVL2 =
+      SqlBasicFunction.create(SqlKind.NVL2,
+          ReturnTypes.NVL2_RESTRICTIVE
+              .andThen(SqlTypeTransforms.TO_NULLABLE_ALL),
+          OperandTypes.SECOND_THIRD_SAME);
+
   /** The "IFNULL(value, value)" function. */
   @LibraryOperator(libraries = {BIG_QUERY, SPARK})
   public static final SqlFunction IFNULL = NVL.withName("IFNULL");
--- a/core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
@@ -986,6 +986,19 @@ public abstract class OperandTypes {
 
   public static final SqlSingleOperandTypeChecker ANY_ANY =
       family(SqlTypeFamily.ANY, SqlTypeFamily.ANY);
+
+  /**
+   * Operand type-checking strategy where the second and third operands must be comparable.
+   * This is used when the operator has three operands and only the
+   * second and third operands need to be comparable.
+   */
+  public static final SqlSingleOperandTypeChecker SECOND_THIRD_SAME =
+      new SameOperandTypeChecker(3) {
+        @Override protected List<Integer> getOperandList(int operandCount) {
+          // Only check the second and third operands
+          return ImmutableList.of(1, 2);
+        }
+      };
   public static final SqlSingleOperandTypeChecker ANY_IGNORE =
       family(SqlTypeFamily.ANY, SqlTypeFamily.IGNORE);
   public static final SqlSingleOperandTypeChecker IGNORE_ANY =
--- a/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
@@ -35,6 +35,7 @@ import org.apache.calcite.util.Glossary;
 import org.apache.calcite.util.Util;
 
 import java.util.AbstractList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.function.UnaryOperator;
 
@@ -550,6 +551,15 @@ public abstract class ReturnTypes {
   public static final SqlReturnTypeInference LEAST_RESTRICTIVE =
       opBinding -> opBinding.getTypeFactory().leastRestrictive(
           opBinding.collectOperandTypes());
+
+  /**
+   * Type-inference strategy for NVL2 function. It returns the least restrictive type
+   * between the second and third operands.
+   */
+  public static final SqlReturnTypeInference NVL2_RESTRICTIVE = opBinding -> {
+    return opBinding.getTypeFactory().leastRestrictive(
+        Arrays.asList(opBinding.getOperandType(1), opBinding.getOperandType(2)));
+  };
 
   /**
    * Type-inference strategy that returns the type of the first operand, unless it
--- a/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
@@ -229,6 +229,7 @@ public class StandardConvertletTable extends ReflectiveConvertletTable {
         registerOp(operator, StandardConvertletTable::convertQuantifyOperator));
 
     registerOp(SqlLibraryOperators.NVL, StandardConvertletTable::convertNvl);
+    registerOp(SqlLibraryOperators.NVL2, StandardConvertletTable::convertNvl2);
     registerOp(SqlLibraryOperators.DECODE,
         StandardConvertletTable::convertDecode);
     registerOp(SqlLibraryOperators.IF, StandardConvertletTable::convertIf);
@@ -416,6 +417,29 @@ public class StandardConvertletTable extends ReflectiveConvertletTable {
                 cx.getTypeFactory()
                     .createTypeWithNullability(type, operand1.getType().isNullable()),
                 operand1)));
+  }
+
+  /** Converts a call to the {@code NVL2} function. */
+  private static RexNode convertNvl2(SqlRexContext cx, SqlCall call) {
+    final RexBuilder rexBuilder = cx.getRexBuilder();
+    final List<RexNode> operands =
+        convertOperands(cx, call, call.getOperandList(), SqlOperandTypeChecker.Consistency.NONE);
+    final RelDataType type = cx.getValidator().getValidatedNodeType(call);
+
+    // Create a CASE expression equivalent to the NVL2 function
+    // NVL2(x, y, z) is equivalent to CASE WHEN x IS NOT NULL THEN y ELSE z END
+    return rexBuilder.makeCall(type, SqlStdOperatorTable.CASE,
+        ImmutableList.of(
+            rexBuilder.makeCall(SqlStdOperatorTable.IS_NOT_NULL,
+                operands.get(0)),
+            rexBuilder.makeCast(
+                cx.getTypeFactory()
+                    .createTypeWithNullability(type, operands.get(1).getType().isNullable()),
+                operands.get(1)),
+            rexBuilder.makeCast(
+                cx.getTypeFactory()
+                    .createTypeWithNullability(type, operands.get(2).getType().isNullable()),
+                operands.get(2))));
   }
 
   /** Converts a call to the INSTR function.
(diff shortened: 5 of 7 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -270,6 +270,14 @@ private static SqlCall transformConvert(SqlValidator validator, SqlCall call) {
               .andThen(SqlTypeTransforms.TO_NULLABLE_ALL),
           OperandTypes.SAME_SAME);
 
+  /** The "NVL2(value, value, value)" function. */
+  @LibraryOperator(libraries = {SPARK})
```
mihaibudiu, line 274: why does the first operand have to have the same type as the other two?

caicancai: I borrowed from NVL, it seems like NVL2 doesn't necessarily have to keep the types the same

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -986,6 +986,14 @@ public static SqlSingleOperandTypeChecker same(int operandCount,
 
   public static final SqlSingleOperandTypeChecker ANY_ANY =
       family(SqlTypeFamily.ANY, SqlTypeFamily.ANY);
+
+  public static final SqlSingleOperandTypeChecker SECOND_THIRD_SAME =
+      new SameOperandTypeChecker(3) {
```
mihaibudiu, line 991: If you happened to see https://issues.apache.org/jira/browse/CALCITE-6383 you may have noticed that the SameOperatndTypeChecker does not enforce that the operands have the same type, but only that they are comparable. Perhaps if this is what you want you should document this fact here.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10036,6 +10036,34 @@ void assertSubFunReturns(boolean binary, String s, int start,
...
+   * Add NVL2 function (enabled in Spark library). </a>.
+   */
+  @Test void testNvl2Func() {
+    final SqlOperatorFixture f = fixture();
+    f.setFor(SqlLibraryOperators.NVL2, VmName.EXPAND);
+    f.checkFails("^nvl2(NULL, 2, 1)^",
+        "No match found for function signature "
+            + "NVL2\\(<NULL>, <NUMERIC>, <NUMERIC>\\)", false);
+
+    final SqlOperatorFixture f12 = f
```
mihaibudiu, line 10050: the SameOperandTypeChecker allows different types for the operands 1 & 2, including numeric and string. Perhaps you should have some tests for these cases.

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
caicancai: I think I understand what you mean, like nvl2(2.0, 'abc', 2) is allowed in spark, but
nvl2(2.0, true, 2) is not allowed because Boolean and numeric are not comparable

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: done

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
caicancai: This is also the case with calcite currently. I have tested it.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10036,6 +10036,45 @@ void assertSubFunReturns(boolean binary, String s, int start,
...
+    f12.checkScalar("nvl2('ab', 'abc', 'def')", "abc", "CHAR(3) NOT NULL");
+    f12.checkScalar("nvl2('a', 3, 2)", "3", "INTEGER NOT NULL");
+
+    // check operand 1 and operand 2 are inconsistent
+    f12.checkScalar("NVL2(2.0, 'abc', 2)", "abc", "CHAR(3) NOT NULL");
+    f12.checkScalar("NVL2(NULL, 'abc', 2)", "2", "INTEGER NOT NULL");
+    f12.checkScalar("NVL2(NULL, 3.0, 4.0)", "4.0", "DECIMAL(2, 1) NOT NULL");
+    f12.checkScalar("NVL2(NULL, 'abc', 2)", "2", "INTEGER NOT NULL");
+    f12.checkScalar("NVL2('abc', 3.0, 4.0)", "3.0", "DECIMAL(2, 1) NOT NULL");
+    f12.checkScalar("NVL2(1, 3.0, 'abc')", "3.0", "DECIMAL(2, 1) NOT NULL");
```
mihaibudiu, line 10069: I assume you have checked these against Spark.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10036,6 +10036,45 @@ void assertSubFunReturns(boolean binary, String s, int start,
...
+    f12.checkScalar("NVL2(NULL, 3.0, 'abc')", "abc", "CHAR(3) NOT NULL");
+
+    f12.checkNull("nvl2('ab', CAST(NULL AS VARCHAR(6)), 'def')");
+    f12.checkNull("nvl2(NULL, 'abc', NULL)");
+    f12.checkNull("nvl2(NULL, NULL, NULL)");
```
mihaibudiu, line 10074: Can you add a few tests that fail in compilation as well because they don't typecheck?
Ideally these would also fail in Spark.

caicancai: @mihaibudiu Hello, sorry for replying so late. I found a problem.
NVL2 (2.0, 'abc', true) is allowed in type verification. The string type and the Boolean type are comparable. I am not sure about this. According to my understanding, it is not allowed. This is also not allowed in spark.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: yes

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: done

caicancai:
@mihaibudiu I would like to ask a question. The type check of NVL2 (2.0, 'abc', true) in calcite is allowed. The string type and the boolean type are comparable. I am not sure about this. According to my understanding, this is is not allowed

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -421,9 +421,12 @@ public enum SqlKind {
   /** {@code DECODE} function (Oracle). */
   DECODE,
 
-  /** {@code NVL} function (Oracle). */
+  /** {@code NVL} function (Oracle, Spark). */
   NVL,
 
+  /** {@code NVL2} function (Oracle). */
```
julianhyde, line 427: Are you saying that NVL2 doesn't exist on Spark?

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -551,6 +551,13 @@ public static SqlCall stripSeparator(SqlCall call) {
       opBinding -> opBinding.getTypeFactory().leastRestrictive(
           opBinding.collectOperandTypes());
 
+  public static final SqlReturnTypeInference NVL2_RESTRICTIVE = opBinding -> {
```
julianhyde, line 554: NEed javadoc

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -418,6 +419,32 @@ private static RexNode convertNvl(SqlRexContext cx, SqlCall call) {
...
+    final RexBuilder rexBuilder = cx.getRexBuilder();
+    final RexNode operand0 =
+        cx.convertExpression(call.getOperandList().get(0));
+    final RexNode operand1 =
+        cx.convertExpression(call.getOperandList().get(1));
+    final RexNode operand2 =
+        cx.convertExpression(call.getOperandList().get(2));
+    final RelDataType type =
+        cx.getValidator().getValidatedNodeType(call);
+    // Preserve Operand Nullability
```
julianhyde, line 433: Why preserve operand nullability? Result is nullable iff either operand 1 or 2 is nullable.

A comment that you translate NVL2(x, y, z) to CASE WHEN x is NULL THEN y ELSE z END would make this method much easier to understand.

Why convert operands one at a time? Use `convertOperands`.

julianhyde: This method could probably be 6 lines long.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -551,6 +551,13 @@ public static SqlCall stripSeparator(SqlCall call) {
...
+    RelDataType opType1 = opBinding.getOperandType(1);
+    RelDataType opType2 = opBinding.getOperandType(2);
+    return opBinding.getOperandType(0).getSqlTypeName() == SqlTypeName.NULL ? opType2 : opType1;
+  };
+
+
```
julianhyde, line 560: remove blank line

site/_docs/reference.md
```diff
@@ -2812,6 +2812,7 @@ In the following:
 | b m p s | MD5(string)                              | Calculates an MD5 128-bit checksum of *string* and returns it as a hex string
 | m | MONTHNAME(date)                                | Returns the name, in the connection's locale, of the month in *datetime*; for example, it returns '二月' for both DATE '2020-02-10' and TIMESTAMP '2020-02-10 10:10:10'
 | o s | NVL(value1, value2)                          | Returns *value1* if *value1* is not null, otherwise *value2*
+| s | NVL2(value1, value2, value3)                   | Returns *value2* if *value1* is not null, or *value3* otherwise.
```
julianhyde, line 2815: remove trailing '.'

Oracle and Spark

Why ', or value3 otherwise' and not ', otherwise value3' when line above says 'otherwise value2'?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10397,6 +10397,50 @@ void assertSubFunReturns(boolean binary, String s, int start,
     checkNvl(f, FunctionAlias.of(SqlLibraryOperators.NVL));
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6397">[CALCITE-6397]
+   * Add NVL2 function (enabled in Spark library). </a>.
```
julianhyde, line 10402: remove '. '.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
julianhyde: need cases like one argument is `DECIMAL(3, 1)` and another is `DECIMAL(4, 3)`

you only seem to be testing the easy cases e.g. two `CHAR(2)`, two `CHAR(3)`. The users won't test the easy cases.

core/src/main/java/org/apache/calcite/sql/SqlKind.java
caicancai: Sorry, I forgot to add

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
caicancai: done

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
caicancai: Thanks for the teaching, done

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
caicancai: done

site/_docs/reference.md
caicancai: fix

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: ok,  done

caicancai:
@mihaibudiu Hello, I opened jira to discuss the comparability of character and boolean types because I'm not sure if this is a bug.
https://issues.apache.org/jira/browse/CALCITE-6409

caicancai:
@mihaibudiu PATL

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
mihaibudiu: This looks wrong. If the first operand does not have the type null then you need to unify the types of operands 2 and 3. In fact, I think you always need to unify them.

caicancai: @mihaibudiu You mean I need to uniformly convert to biggesttype?

caicancai:
@YiwenWu If you have time, can you help me take a look at this PR? Thank you

core/src/main/java/org/apache/calcite/sql/SqlKind.java
caicancai: done

YiwenWu (approved):
LGTM

mihaibudiu (approved):

caicancai:
@mihaibudiu @julianhyde @YiwenWu 
Thanks for the review, squashed

PR #3786, opened 2024-05-03
[CALCITE-6352] The map_contains_key function may return true when the key and mapkeytype types are different

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6352

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -5529,7 +5529,21 @@ public class SqlFunctions {
 
   /** Support the MAP_CONTAINS_KEY function. */
   public static Boolean mapContainsKey(Map map, Object key) {
+    for (Object mapKey : map.keySet()) {
+      if (isNumericEqual(mapKey, key)) {
+        return true;
+      }
+    }
     return map.containsKey(key);
+  }
+
+  private static boolean isNumericEqual(Object obj1, Object obj2) {
+    if (obj1 instanceof Number && obj2 instanceof Number) {
+      double num1 = ((Number) obj1).doubleValue();
+      double num2 = ((Number) obj2).doubleValue();
+      return num1 == num2;
+    }
+    return false;
   }
 
   /** Support the MAP_FROM_ARRAYS function. */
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -5529,9 +5529,23 @@ public static List mapValues(Map map) {
...
+      if (isNumericEqual(mapKey, key)) {
+        return true;
+      }
+    }
     return map.containsKey(key);
   }
 
+  private static boolean isNumericEqual(Object obj1, Object obj2) {
+    if (obj1 instanceof Number && obj2 instanceof Number) {
+      double num1 = ((Number) obj1).doubleValue();
```
mihaibudiu, line 5542: this may not work for some decimal values which cannot be represented exactly as doubles
in general, it is not recommended to compare floating point values for equality, e.g, https://www.jetbrains.com/help/resharper/CompareOfFloatsByEqualityOperator.html

caicancai: you are right

caicancai: @mihaibudiu Should I add a layer of explicit conversion in case to convert the types in map and key to the biggest type

caicancai: Or convert BigDecimal for comparison, which one do you prefer?

caicancai:
@mihaibudiu Can it be merged?

caicancai:
2024-06-09 23-35-33屏幕截图
The number comparison of spark map_contains_key function seems to only support double, not bigdecimal.
Calcite's map_contains_key is also only adapted to spark. I think my change is the minimum change. Is it acceptable?
What do you think?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
mihaibudiu: As I said, I don't think you should convert the values to double and then compare them.

caicancai: OK, I will convert the numeric type to double by default for comparison

PR #3790, opened 2024-05-10
[CALCITE-6380] Casts from INTERVAL and STRING to DECIMAL are incorrect

mihaibudiu:


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -770,41 +773,39 @@ void testCastIntervalToNumeric(CastType castType, SqlOperatorFixture f) {
     f.setFor(SqlStdOperatorTable.CAST, VmName.EXPAND);
 
     // interval to decimal
```
caicancai, line 775: Interval to Decimal

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -449,6 +449,35 @@ private Expression getConvertExpression(
...
+              Expressions.constant(sourceType.getSqlTypeName().getEndUnit().multiplier));
+        } else if (sourceType.getFamily() == SqlTypeFamily.INTERVAL_YEAR_MONTH) {
+          return Expressions.call(
+              BuiltInMethod.LONG_INTERVAL_DECIMAL_CAST.method,
+              operand,
+              Expressions.constant(precision),
+              Expressions.constant(scale),
+              Expressions.constant(sourceType.getSqlTypeName().getEndUnit().multiplier));
+        }
+      }
```
caicancai, line 478: Should we throw an exception when a situation does not include the above three cases

caicancai (commented):
Overall LGTM

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -738,30 +739,32 @@ void testCastToExactNumeric(CastType castType, SqlOperatorFixture f) {
...
   void testCastStringToDecimal(CastType castType, SqlOperatorFixture f) {
     f.setFor(SqlStdOperatorTable.CAST, VmName.EXPAND);
-    if (!DECIMAL) {
-      return;
-    }
     // string to decimal
     f.checkScalarExact("cast('1.29' as decimal(2,1))",
         "DECIMAL(2, 1) NOT NULL",
-        "1.3");
+        "1.2");
```
YiwenWu, line 745: Different systems have different interception methods, I think the return value `1.3` is also reasonable. Most systems return `1.3`, such as MySQL, PG, Hive, and Spark, but ClickHouse returns `1.2`.

caicancai: There doesn't seem to be a standard set of sql rules here, so maybe we should think about what makes sense mathematically or what makes sense to follow the rules of most databases

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
mihaibudiu: I have left the previous default behavior unchanged.
The other PR #3733 handles more cases.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: @YiwenWu I have written about the rounding mode used in Calcite: https://issues.apache.org/jira/browse/CALCITE-6322. Here I am just consistent with the rest of the codebase, but I agree that I would prefer a different rounding mode. Maybe this should be part of the typesystem.

caicancai (approved):
LGTM

PR #3792, opened 2024-05-15
[CALCITE-6358] Support all PostgreSQL 14 date/time patterns for to_char

normanj-bitquill:
* Splits the PostgreSQL toChar function off to its own function
* Does not implement SP suffix
* Timezone patterns are supported but all datetimes are in local timezone

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -42,6 +42,7 @@ import org.apache.calcite.sql.type.SqlTypeFamily;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.type.SqlTypeTransforms;
 import org.apache.calcite.sql.type.SqlTypeUtil;
+import org.apache.calcite.sql.validate.SqlMonotonicity;
 import org.apache.calcite.sql.validate.SqlValidator;
 import org.apache.calcite.sql.validate.SqlValidatorUtil;
 import org.apache.calcite.util.Litmus;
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1651,13 +1652,22 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
-          ReturnTypes.VARCHAR,
+          ReturnTypes.VARCHAR_NULLABLE,
           OperandTypes.TIMESTAMP_STRING,
           SqlFunctionCategory.TIMEDATE);
 
+  /** The "TO_CHAR(timestamp, format)" function;
+   * converts {@code timestamp} to string according to the given {@code format}. */
+  @LibraryOperator(libraries = {POSTGRESQL})
+  public static final SqlFunction TO_CHAR_PG =
+      new SqlBasicFunction("TO_CHAR", SqlKind.OTHER_FUNCTION,
```
YiwenWu, line 1666: Can this be used directly?
```
SqlBasicFunction.create("TO_CHAR",
          ReturnTypes.VARCHAR_NULLABLE,
          OperandTypes.TIMESTAMP_STRING,
```

normanj-bitquill: No. The implementations of `hashCode` and `equals` are from `SqlOperator` and only looks at the name and kind.

There is also a `TO_CHAR` for MySQL/Oracle and this new one for PostgreSQL. Since they have the same name and kind they would conflict in a HashMap. Using a new class here gets around the collision. Another solution is to add `equals` and `hashCode` implementations to `SqlBasicFunction`.

YiwenWu: Understood. 

A small suggestion, I think the `SqlBasicFunction` constructor is designed to be as private as possible. Is it possible to add a create method to create the corresponding anonymous class, for example:

```
  public static SqlBasicFunction overloadCreate(String name,
      SqlReturnTypeInference returnTypeInference,
      SqlOperandTypeChecker operandTypeChecker, SqlFunctionCategory category) {
    return new SqlBasicFunction(name, SqlKind.OTHER_FUNCTION,
        SqlSyntax.FUNCTION, true, returnTypeInference, null,
        OperandHandlers.DEFAULT, operandTypeChecker, 0,
        category, call -> SqlMonotonicity.NOT_MONOTONIC, false){};
  }
```

mihaibudiu (approved):

mihaibudiu:
This looks good, we can probably merge this one and close #3773

PR #3798, opened 2024-05-24
[CALCITE-6417] Map value constructor and Array value constructor unpa…

NobiGo:
…rsed incorrectly for HiveSqlDialect

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/HiveSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/HiveSqlDialect.java
@@ -33,6 +33,8 @@ import org.apache.calcite.util.RelToSqlConverterUtil;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
+import static org.apache.calcite.util.RelToSqlConverterUtil.unparseSparkArrayAndMap;
+
 /**
  * A <code>SqlDialect</code> implementation for the Apache Hive database.
  */
@@ -80,6 +82,10 @@ public class HiveSqlDialect extends SqlDialect {
   @Override public void unparseCall(final SqlWriter writer, final SqlCall call,
       final int leftPrec, final int rightPrec) {
     switch (call.getKind()) {
+    case ARRAY_VALUE_CONSTRUCTOR:
+    case MAP_VALUE_CONSTRUCTOR:
+      unparseSparkArrayAndMap(writer, call, leftPrec, rightPrec);
+      break;
     case POSITION:
       final SqlWriter.Frame frame = writer.startFunCall("INSTR");
       writer.sep(",");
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/HiveSqlDialect.java
```diff
@@ -80,6 +82,10 @@ public HiveSqlDialect(Context context) {
   @Override public void unparseCall(final SqlWriter writer, final SqlCall call,
       final int leftPrec, final int rightPrec) {
     switch (call.getKind()) {
+    case ARRAY_VALUE_CONSTRUCTOR:
+    case MAP_VALUE_CONSTRUCTOR:
+      unparseSparkArrayAndMap(writer, call, leftPrec, rightPrec);
```
snuyanzin, line 87: minor: should we think about a different name since it is not only `Spark` after this PR?

NobiGo: Hi, Thanks for the review. This name means we use SparkSQL standard to handle Array and Map, Like SparkSqlDialect and StarRocksSqlDialect use unparseHiveTrim method to handle the Trim.  So I think it's Ok.

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
```diff
@@ -63,6 +63,36 @@ public static void unparseHiveTrim(
...
+   *
+   * <blockquote><pre>
+   * SELECT ARRAY[1, 2, 3] &rarr; SELECT ARRAY (1, 2, 3)
+   * SELECT MAP['k1', 'v1', 'k2', 'v2'] &rarr; SELECT MAP ('k1', 'v1', 'k2', 'v2')
+   * </pre></blockquote>
+   *
+   * @param writer writer
+   * @param call the call
+   */
+  public static void unparseSparkArrayAndMap(SqlWriter writer,
```
YiwenWu, line 79: minor: putting the unparse method in the `RelToSqlConverterUtil` tool class feels a bit ambiguous,  because unparse should belong to SqlNode to Sql rather than RelNode to SqlNode.

YiwenWu (approved):

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
NobiGo: Hi @YiwenWu ,I noticed this too. So how about when we merge this after , In another PR ,We can create new SqlNodeToSqlUtil , Then refactor the relative code ?

YiwenWu: I agree, it's a good idea

PR #3799, opened 2024-05-27
[CALCITE-6419] Invalid unparse for VARCHAR without precision in HiveS…

NobiGo:
…qlDialect And SparkSqlDialect

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/SparkSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/SparkSqlDialect.java
@@ -18,8 +18,11 @@ package org.apache.calcite.sql.dialect;
 
 import org.apache.calcite.avatica.util.TimeUnitRange;
 import org.apache.calcite.config.NullCollation;
+import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.JoinType;
+import org.apache.calcite.sql.SqlAlienSystemTypeNameSpec;
 import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlDataTypeSpec;
 import org.apache.calcite.sql.SqlDialect;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlLiteral;
@@ -28,6 +31,9 @@ import org.apache.calcite.sql.SqlUtil;
 import org.apache.calcite.sql.SqlWriter;
 import org.apache.calcite.sql.fun.SqlFloorFunction;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.type.BasicSqlType;
+import org.apache.calcite.sql.type.SqlTypeName;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/SparkSqlDialect.java
```diff
@@ -146,4 +152,15 @@ public SparkSqlDialect(SqlDialect.Context context) {
       super.unparseCall(writer, call, leftPrec, rightPrec);
     }
   }
+
+  @Override public @Nullable SqlNode getCastSpec(RelDataType type) {
+    if (type instanceof BasicSqlType) {
+      if (type.getSqlTypeName() == SqlTypeName.VARCHAR) {
```
YiwenWu, line 158: minor: can use a `switch case` block or use a single `if` judgment instead of `nested ifs`?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3519,9 +3519,57 @@ private SqlDialect nonOrdinalDialect() {
...
+    final String expectedSpark = "SELECT CAST(product_id AS STRING)\n"
+        + "FROM foodmart.product";
     sql(query)
         .withClickHouse().ok(expectedClickHouse)
-        .withMysql().ok(expectedMysql);
+        .withMysql().ok(expectedMysql)
+        .withHive().ok(expectedHive)
+        .withSpark().ok(expectedSpark);
+  }
+
```
YiwenWu, line 3532: can add corresponding jira link comments?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3519,9 +3519,57 @@ private SqlDialect nonOrdinalDialect() {
...
+        + "FROM foodmart.product";
+    sql(query)
+        .withMysql().ok(expectedMysql)
+        .withHive().ok(expectedHive)
+        .withSpark().ok(expectedSpark);
+  }
+
+  @Test void testCastToChar() {
+    String query = "select cast(\"product_id\" as char) from \"product\"";
+    final String expectedMysql = "SELECT CAST(`product_id` AS CHAR(1))\n"
```
YiwenWu, line 3549: A problem: MYSQL supports`select cast(\"product_id\" as char)`. Converting `char` to `char(1)` here will intercept characters. Is it reasonable?

core/src/main/java/org/apache/calcite/sql/dialect/SparkSqlDialect.java
NobiGo: Thanks for the review @YiwenWu . I wil use if to replace nested ifs.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
NobiGo: Yes, I will add it.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
NobiGo: I will check this.

core/src/main/java/org/apache/calcite/sql/dialect/SparkSqlDialect.java
NobiGo: Hi @YiwenWu , I have fix this.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
NobiGo: Hi @YiwenWu , I have fix this.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
NobiGo: This is a bug in Calcite, So I create another Jira issue to describe it, Please check https://issues.apache.org/jira/browse/CALCITE-6423

PR #3801, opened 2024-05-27
CALCITE-6310: Add REGEXP_REPLACE function (enabled in PostgreSQL library)

jduo:
PostgreSQL allows use of the flags string at any parameter after the replacement string.


jduo:
Initial PR with only validation implemented.

normanj-bitquill:
Updated with the following behaviour:

### Amazon Redshift
* Can provide only two arguments. The replacement string will be the empty string.

https://docs.aws.amazon.com/redshift/latest/dg/REGEXP_REPLACE.html

### PostgreSQL
* With 3 arguments, only the first occurrence is replaced
* With 4 arguments, the last argument is flags. If flags contains `g`, then all occurrences will be replaced.
* Cannot provide more than 4 arguments

### BigQuery, MySQL, Oracle, Amazon Redshift
* Must provide at least 3 arguments
* No change to behaviour

https://www.postgresql.org/docs/14/functions-matching.html#FUNCTIONS-POSIX-REGEXP

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -563,12 +563,58 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
+  @LibraryOperator(libraries = {BIG_QUERY, MYSQL, ORACLE, REDSHIFT})
+  public static final SqlFunction REGEXP_REPLACE_3 =
+      SqlBasicFunction.create("REGEXP_REPLACE", ReturnTypes.VARCHAR_NULLABLE,
+          OperandTypes.STRING_STRING_STRING, SqlFunctionCategory.STRING);
+
   /** The "REGEXP_REPLACE(value, regexp, rep [, pos [, occurrence [, matchType]]])"
    * function. Replaces all substrings of value that match regexp with
    * {@code rep} and returns modified value. */
   @LibraryOperator(libraries = {BIG_QUERY, MYSQL, ORACLE, REDSHIFT})
   public static final SqlFunction REGEXP_REPLACE = new SqlRegexpReplaceFunction();
```
normanj-bitquill, line 588: Remove the `SqlRegexpReplaceFunction` class.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -563,12 +563,58 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
+          SqlSyntax.FUNCTION, true, ReturnTypes.VARCHAR_NULLABLE, null,
+          OperandHandlers.DEFAULT, OperandTypes.STRING_STRING_STRING_STRING, 0,
+          SqlFunctionCategory.STRING, call -> SqlMonotonicity.NOT_MONOTONIC, false) { };
+
+  /** The PostgreSQL variant of
+   * "REGEXP_REPLACE(value, regexp, rep [, pos [, occurrence]] [, matchType])"
+   * function. Replaces all substrings of value that match regexp with
+   * {@code rep} and returns modified value. */
+  @LibraryOperator(libraries = {POSTGRESQL})
+  public static final SqlFunction PG_REGEXP_REPLACE = new SqlPgRegexpReplaceFunction();
```
normanj-bitquill, line 615: Remove this, since it is not used.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -563,12 +563,58 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
           OperandTypes.STRING_STRING_OPTIONAL_INTEGER_OPTIONAL_INTEGER_OPTIONAL_INTEGER,
           SqlFunctionCategory.STRING);
 
+  /** The "REGEXP_REPLACE(value, regexp)"
+   * function. Replaces all substrings of value that match regexp with
+   * {@code rep} and returns modified value. */
+  @LibraryOperator(libraries = {REDSHIFT})
```
normanj-bitquill, line 569: Also enable for PostgreSQL.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: Removed. This class was looking at the `SqlConformanceEnum` to decide on how capture groups are specified in the replacement. BigQuery uses `\\` and others use `$`. There is now a new LibraryOperator for BigQuery so that `\\` in the replacement is converted to `$` before doing the replacement.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: Removed.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: Enabled.

mihaibudiu:
If this PR is ready for review please make a sign

normanj-bitquill:
> If this PR is ready for review please make a sign

Yes, it is ready. Please take a look when you have time.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -630,6 +630,11 @@ public int regexpInstr(String value, String regex, int position,
       return matchIndex;
     }
 
+    /** SQL {@code REGEXP_REPLACE} function with 2 arguments. */
+    public String regexpReplace(String s, String regex) {
+      return regexpReplace(s, regex, "", 1, 0, null);
```
mihaibudiu, line 635: This is all predicated on regexpReplace doing the right thing.
I guess that is not in the scope of this PR, but all these regular expression languages have slight differences from each other, so it's not obvious to me that regexpReplace implements faithfully what each of these databases expects.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -563,11 +563,86 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
           OperandTypes.STRING_STRING_OPTIONAL_INTEGER_OPTIONAL_INTEGER_OPTIONAL_INTEGER,
           SqlFunctionCategory.STRING);
 
-  /** The "REGEXP_REPLACE(value, regexp, rep [, pos [, occurrence [, matchType]]])"
+  /** The "REGEXP_REPLACE(value, regexp)"
```
mihaibudiu, line 566: this comment cannot be right, since there is no "rep"

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -563,11 +563,86 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
+  /** The "REGEXP_REPLACE(value, regexp, rep)"
+   * function. Replaces all substrings of value that match regexp with
+   * {@code rep} and returns modified value. */
+  @LibraryOperator(libraries = {MYSQL, ORACLE, REDSHIFT})
+  public static final SqlFunction REGEXP_REPLACE_3 =
+      SqlBasicFunction.create("REGEXP_REPLACE", ReturnTypes.VARCHAR_NULLABLE,
+          OperandTypes.STRING_STRING_STRING, SqlFunctionCategory.STRING);
+
+  /** The "REGEXP_REPLACE(value, regexp, rep, pos)"
+   * function. Replaces all substrings of value that match regexp with
```
mihaibudiu, line 585: what is pos?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -563,11 +563,86 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
+  @LibraryOperator(libraries = {MYSQL, ORACLE, REDSHIFT})
+  public static final SqlFunction REGEXP_REPLACE_4 =
+      new SqlBasicFunction("REGEXP_REPLACE", SqlKind.OTHER_FUNCTION,
+          SqlSyntax.FUNCTION, true, ReturnTypes.VARCHAR_NULLABLE, null,
+          OperandHandlers.DEFAULT, OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.STRING,
+              SqlTypeFamily.STRING, SqlTypeFamily.INTEGER),
+          0, SqlFunctionCategory.STRING, call -> SqlMonotonicity.NOT_MONOTONIC, false) { };
+
+  /** The "REGEXP_REPLACE(value, regexp, rep, pos, occurrence)"
+   * function. Replaces all substrings of value that match regexp with
```
mihaibudiu, line 596: please document all arguments in these comments

core/src/main/java/org/apache/calcite/sql/fun/SqlPgRegexpReplaceFunction.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.calcite.sql.type.SqlTypeFamily;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * The PostgreSQL
+ * REGEXP_REPLACE(source_string, pattern, replacement [, pos, [, occurrence]] [, match_type])
+ * searches for a regular expression pattern and replaces every occurrence of the pattern
+ * with the specified string. It differs from the standard REGEXP_REPLACE in that there is
```
mihaibudiu, line 30: And here too

core/src/main/java/org/apache/calcite/sql/validate/SqlAbstractConformance.java
```diff
@@ -105,10 +105,6 @@ public abstract class SqlAbstractConformance implements SqlConformance {
     return SqlConformanceEnum.DEFAULT.isOffsetLimitAllowed();
   }
 
-  @Override public boolean isRegexReplaceCaptureGroupDollarIndexed() {
```
mihaibudiu, line 108: where did the functionality of this move?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5727,10 +5727,32 @@ private static void checkIf(SqlOperatorFixture f) {
     f.checkQuery("select regexp_instr('a9cadca5c4aecghi', 'a[0-9]c', 1, 3)");
   }
 
-  @Test void testRegexpReplaceFunc() {
+  @Test void testRegexpReplace2Func() {
     final SqlOperatorFixture f0 = fixture();
     final Consumer<SqlOperatorFixture> consumer = f -> {
-      f.setFor(SqlLibraryOperators.REGEXP_REPLACE);
+      f.setFor(SqlLibraryOperators.REGEXP_REPLACE_2);
```
mihaibudiu, line 5733: I am assuming these were validated against postgres. If true, please add a comment.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
normanj-bitquill: You are correct, there may be functional differences between the implementation of `regexp_replace` here and what the engines actually have. I would like to leave that out of scope for this PR.

mihaibudiu: I think that could easily evolve into a 1 year project just by itself.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: The comment has been updated.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: Added comments for each of the arguments.

core/src/main/java/org/apache/calcite/sql/fun/SqlPgRegexpReplaceFunction.java
normanj-bitquill: This class is not needed. It has been removed along with its parent class that is also no longer needed.

core/src/main/java/org/apache/calcite/sql/validate/SqlAbstractConformance.java
normanj-bitquill: There is a specific library function for BigQuery. BigQuery only supports the 3 argument signature. BigQuery was the only engine that uses dollar signs rather than backslash.

https://github.com/Bit-Quill/calcite/blob/calcite-6310-regex-replace/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java#L642

https://github.com/Bit-Quill/calcite/blob/calcite-6310-regex-replace/core/src/main/java/org/apache/calcite/util/BuiltInMethod.java#L590

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
normanj-bitquill: I took another look and realized that it is only Amazon Redshift that supports a 2 argument signature. I have updated accordingly. I have also tested other engines.
* With 4 arguments, all engines expect a signature of "string, string, string, int"
* Oracle only supports the signature "string, string, string, int, int" for 5 arguments
* MySQL and Amazon Redshift support two signature for 5 arguments
  * "string, string, string, int, int"
  * "string, string, string, int, string"

Updated to better match the signatures each engine supports.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: Updated the Javadoc comment here.

mihaibudiu (approved):
I haven't really verified that all these dialects behave as the PR says; if they do, this looks pretty good - modulo the fact that the regular expression libraries in all these languages may differ in minute aspects.

mihaibudiu:
From my pov the commits can be squashed

normanj-bitquill:
I have squashed the commits.

PR #3797, opened 2024-05-23
[CALCITE-6370] AS operator problems with USING clause

normanj-bitquill:
* Detect when a nested AS call is inserted and prevent it

PR #3802, opened 2024-05-27
[CALCITE-6369] Expanding "star" gives ArrayIndexOutOfBoundsException …

normanj-bitquill:
…with redundant columns and USING

Fixes expanding * when columns are also specified in the projection list

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
@@ -666,6 +666,7 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
     if (!identifier.isStar()) {
       return false;
     }
+    final int originalSize = selectItems.size();
     final SqlParserPos startPosition = identifier.getParserPosition();
     switch (identifier.names.size()) {
     case 1:
@@ -736,7 +737,7 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
       if (!hasDynamicStruct || Bug.CALCITE_2400_FIXED) {
         // If some fields before star identifier,
         // we should move offset.
-        int offset = calculatePermuteOffset(selectItems);
+        int offset = Math.min(calculatePermuteOffset(selectItems), originalSize);
         new Permute(from, offset).permute(selectItems, fields);
       }
       return true;
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -8318,6 +8318,74 @@ class RelToSqlConverterTest {
         .withPostgresql().ok(expectedQuery);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6369">[CALCITE-6369]
+   * Expanding "star" gives ArrayIndexOutOfBoundsException with redundant columns and USING</a>.
+   */
+  @Test void testUsingClauseWithStarInProjection() {
+    final String query = "select \"employee_id\", * from \"employee\" e0 join "
+        + "\"employee\" e1 using (\"employee_id\")";
+    final String expectedQuery = "SELECT \"employee\".\"employee_id\", "
+        + "\"employee\".\"employee_id\" AS \"employee_id0\", \"employee\".\"full_name\", "
+        + "\"employee\".\"first_name\", \"employee\".\"last_name\", \"employee\".\"position_id\", "
+        + "\"employee\".\"position_title\", \"employee\".\"store_id\", "
+        + "\"employee\".\"department_id\", \"employee\".\"birth_date\", "
+        + "\"employee\".\"hire_date\", \"employee\".\"end_date\", \"employee\".\"salary\", "
+        + "\"employee\".\"supervisor_id\", \"employee\".\"education_level\", "
+        + "\"employee\".\"marital_status\", \"employee\".\"gender\", "
+        + "\"employee\".\"management_role\", \"employee0\".\"full_name\" AS \"full_name0\", "
+        + "\"employee0\".\"first_name\" AS \"first_name0\", "
+        + "\"employee0\".\"last_name\" AS \"last_name0\", "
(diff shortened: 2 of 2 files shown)


normanj-bitquill:
The title is too long. It was taken from the Jira issue. Is it best to update the title for the Jira and then the commit message?

core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
```diff
@@ -297,8 +298,17 @@ protected void addColumnNames(
         namespace = map.get(tableName).namespace;
         break;
       default:
-        throw validator.newValidationError(identifier,
-            RESOURCE.columnAmbiguous(columnName));
+        // Allow if all namespaces are the same and the tables are not null
```
YiwenWu, line 301: I think that when this PR is merged, the column ambiguous problem will be solved, so no changes are needed here.
https://github.com/apache/calcite/pull/3797

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -633,6 +633,7 @@ private boolean expandStar(List<SqlNode> selectItems, Set<String> aliases,
     if (!identifier.isStar()) {
       return false;
     }
+    final int offset = selectItems.size();
```
YiwenWu, line 636: If `offset` is judged here, then all fields before * will not permutation.
I recommend improving the `calculatePermuteOffset` method. like
```
int originSize = selectItems.size();
......
int offset = Math.min(calculatePermuteOffset(selectItems), originSize);
```

normanj-bitquill: Updated.

normanj-bitquill: @YiwenWu Can you take another look?

core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java
YiwenWu: @normanj-bitquill here, is this modification necessary?

normanj-bitquill: @YiwenWu I'm not sure. If we have a query such as:

```
select x, * from foo f join bar b using (x)
```

Does it matter if we choose `foo.x` or `bar.x` in this context? Due to the join, they should have the same value.

YiwenWu: Without this modification, verified that the two examples of unit tests can also be executed successfully.

normanj-bitquill: @YiwenWu  Rebased this change off of the current main branch. As a result, this change is not needed and has been removed.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -736,7 +737,7 @@ private boolean expandStar(List<SqlNode> selectItems, Set<String> aliases,
       if (!hasDynamicStruct || Bug.CALCITE_2400_FIXED) {
         // If some fields before star identifier,
         // we should move offset.
-        int offset = calculatePermuteOffset(selectItems);
+        int offset = Math.min(calculatePermuteOffset(selectItems), originalSize);
```
mihaibudiu, line 740: I see this is gated on CALCITE_2400.
What is the relationship between this bug and CALCITE-2400?

normanj-bitquill: @mihaibudiu  It was possible to get an `IndexOutOfBoundsException` when correcting the field order (for CALCITE-2400). We need to make sure that the offset is not beyond the `selectItems` list.

YiwenWu (approved):

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
mihaibudiu: My question was whether the `|| Bug.CALCITE_2400_FIXED` is still needed after your fix.

mihaibudiu:
This looks fine, but I would have liked an answer to the question I asked about bug_2400.

mihaibudiu:
you can probably squash the commits, since this is a very small PR.

normanj-bitquill:
@mihaibudiu The commits have been squashed.

mihaibudiu:
can you say anything about bug_2400?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
normanj-bitquill: I don't see any use for the `Bug.CALCITE_2400_FIXED` flag anymore. Removed it.

normanj-bitquill:
> can you say anything about bug_2400?

@mihaibudiu I have removed the flag now, since it does not appear to be needed. 2400 linked to two other issues that are both now resolved.

mihaibudiu:
It's hard to tell from the JIRA description of [CALCITE-2400] whether this is fixing that issue as well.

mihaibudiu:
So maybe then you can also squash the commits here as well.

normanj-bitquill:
> It's hard to tell from the JIRA description of [[CALCITE-2400](https://issues.apache.org/jira/browse/CALCITE-2400)] whether this is fixing that issue as well.

@mihaibudiu I'll look more into this.

mihaibudiu:
> > It's hard to tell from the JIRA description of [[CALCITE-2400](https://issues.apache.org/jira/browse/CALCITE-2400)] whether this is fixing that issue as well.
> 
> @mihaibudiu I'll look more into this.

That is not necessary to merge this PR. But if you can figure it out with a small effort, even better.

normanj-bitquill:
@mihaibudiu I have reverted the change removing the flag for bug 2400. This PR does not fix it. At that point in the code, we are unable to expand the columns from a dynamic row. We could move the explicit fields in the select list to the front, but I don't know if that would fully solve CALCITE-2400.

mihaibudiu:
It is very unfortunate that there is no test for that issue.

PR #3803, opened 2024-05-29
[CALCITE-6411] Support Collect in ToLogicalConverter

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6411


YiwenWu (commented):
I think this PR is unnecessary now. 
First of all, there are no corresponding usage scenarios and unit tests. Secondly, this method is used by `PigConverter`.  Currently, Pig is rarely used, so there is no need to spend too much effort.

caicancai:
> I think this PR is unnecessary now. First of all, there are no corresponding usage scenarios and unit tests. Secondly, this method is used by `PigConverter`. Currently, Pig is rarely used, so there is no need to spend too much effort.

Thanks, closing

caicancai:
@YiwenWu I added it when I was learning this module and found it useful. I am not clear about its specific usage scenarios. Thank you for reminding me. Thank you very much

PR #3806, opened 2024-05-31
[CALCITE-6424] Enable RLIKE function in MySQL library

NobiGo:


mihaibudiu (approved):

site/_docs/reference.md
```diff
@@ -2833,8 +2833,8 @@ In the following:
 | b m p s | REPEAT(string, integer)                  | Returns a string consisting of *string* repeated of *integer* times; returns an empty string if *integer* is less than 1
 | b m | REVERSE(string)                              | Returns *string* with the order of the characters reversed
 | b m p s | RIGHT(string, length)                    | Returns the rightmost *length* characters from the *string*
-| h s | string1 RLIKE string2                        | Whether *string1* matches regex pattern *string2* (similar to `LIKE`, but uses Java regex)
-| h s | string1 NOT RLIKE string2                    | Whether *string1* does not match regex pattern *string2* (similar to `NOT LIKE`, but uses Java regex)
+| h m s | string1 RLIKE string2                        | Whether *string1* matches regex pattern *string2* (similar to `LIKE`, but uses Java regex)
```
caicancai, line 2836: Can you make the formatting align?

site/_docs/reference.md
```diff
@@ -2822,7 +2822,7 @@ In the following:
 | b s | POW(numeric1, numeric2)                      | Returns *numeric1* raised to the power *numeric2*
 | b c h q m o f s p | POWER(numeric1, numeric2) | Returns *numeric1* raised to the power of *numeric2*
 | p | RANDOM()                                       | Generates a random double between 0 and 1 inclusive
-| s | REGEXP(string, regexp)                         | Equivalent to `string1 RLIKE string2`
+| h s | REGEXP(string, regexp)                         | Equivalent to `string1 RLIKE string2`
```
caicancai, line 2825: above all

caicancai (approved):
Overall LGTM, I leave a few small comments

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -551,7 +551,7 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
   public static final SqlFunction REGEXP_SUBSTR = REGEXP_EXTRACT.withName("REGEXP_SUBSTR");
 
   /** The "REGEXP(value, regexp)" function, equivalent to {@link #RLIKE}. */
-  @LibraryOperator(libraries = {SPARK})
+  @LibraryOperator(libraries = {SPARK, HIVE})
```
YiwenWu, line 554: According to the official documentation, Hive does not support REGEXP(A, B).

https://cwiki.apache.org/confluence/display/Hive/LanguageManual+UDF
image

YiwenWu (changes requested):

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -974,12 +974,12 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
...
 
   /** The regex variant of the LIKE operator. */
-  @LibraryOperator(libraries = {SPARK, HIVE})
+  @LibraryOperator(libraries = {SPARK, HIVE, MYSQL})
   public static final SqlSpecialOperator RLIKE =
       new SqlLikeOperator("RLIKE", SqlKind.RLIKE, false, true);
 
   /** The regex variant of the NOT LIKE operator. */
-  @LibraryOperator(libraries = {SPARK, HIVE})
+  @LibraryOperator(libraries = {SPARK, HIVE, MYSQL})
```
YiwenWu, line 982: I think the `MySQL` engine should not be added. 
Although the SQL is the same, their results are not the same. Hive and Spark will return true/false, while MySQL will return 1/0

YiwenWu (changes requested):

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: @YiwenWu Thank you for your reminder. If we want to add RLIKE to MySQL, I think we should write an Operator, just like the log function, to adapt to different database dialects.

site/_docs/reference.md
NobiGo: Thanks for the review. I will format the code.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
NobiGo: Yes, I try A REGEXP B, not the REGEXP(A, B). Thanks for the review, I will rollback the code.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
NobiGo: @YiwenWu @caicancai In MySQL, Use 0 or 1 to replace the TRUE or false.
In MySQL Example:
`SELECT 1 = 1 FROM Product;`
will return 1 to replace the true.
And For example:
`SELECT * FROM Product where Name not rlike '.apper Plus';`
the NOT RLIKE can be used TRUE or False, So I think we don't add another operator for this.
Same as `REGEXP_LIKE` function in SqlLibraryOperators.

YiwenWu: Regarding the question of whether the return value needs to be completely consistent. I'm not sure, this is an ambiguous question. 
If Calcite allows the return value to be different from the corresponding engine, then it is okay to handle it this way.

PR #3808, opened 2024-06-03
[CALCITE-6426] Invalid unparse for INT and BIGINT in StarRocksDialect

fanluoo:
https://issues.apache.org/jira/browse/CALCITE-6426

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
@@ -21,6 +21,7 @@ import org.apache.calcite.config.NullCollation;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.SqlAbstractDateTimeLiteral;
 import org.apache.calcite.sql.SqlAlienSystemTypeNameSpec;
+import org.apache.calcite.sql.SqlBasicTypeNameSpec;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlDataTypeSpec;
 import org.apache.calcite.sql.SqlDialect;
@@ -109,6 +110,17 @@ public class StarRocksSqlDialect extends MysqlSqlDialect {
 
   @Override public @Nullable SqlNode getCastSpec(RelDataType type) {
     switch (type.getSqlTypeName()) {
+    case INTEGER:
+      return new SqlDataTypeSpec(
+          new SqlAlienSystemTypeNameSpec(
+              "INT",
+              type.getSqlTypeName(),
+              SqlParserPos.ZERO),
+          SqlParserPos.ZERO);
+    case BIGINT:
+      return new SqlDataTypeSpec(
+          new SqlBasicTypeNameSpec(SqlTypeName.BIGINT, SqlParserPos.ZERO),
+          SqlParserPos.ZERO);
     case TIMESTAMP:
       return new SqlDataTypeSpec(
           new SqlAlienSystemTypeNameSpec(
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -3072,24 +3072,44 @@ class RelToSqlConverterTest {
     sql(query).dialect(HiveSqlDialect.DEFAULT).ok(expected);
   }
 
-  @Test void testMysqlCastToBigint() {
-    // MySQL does not allow cast to BIGINT; instead cast to SIGNED.
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-2719">[CALCITE-2719]
+   * MySQL does not support cast to BIGINT type</a>
+   * and
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6426">[CALCITE-6426]
+   * Invalid unparse for INT and BIGINT in StarRocksDialect</a>. */
+  @Test void testCastToBigint() {
     final String query = "select cast(\"product_id\" as bigint) from \"product\"";
-    final String expected = "SELECT CAST(`product_id` AS SIGNED)\n"
+    // MySQL does not allow cast to BIGINT; instead cast to SIGNED.
+    final String expectedMysql = "SELECT CAST(`product_id` AS SIGNED)\n"
         + "FROM `foodmart`.`product`";
-    sql(query).withMysql().ok(expected);
+    final String expectedStarRocks = "SELECT CAST(`product_id` AS BIGINT)\n"
+        + "FROM `foodmart`.`product`";
+    sql(query)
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -109,6 +109,20 @@ public StarRocksSqlDialect(Context context) {
...
   @Override public @Nullable SqlNode getCastSpec(RelDataType type) {
     switch (type.getSqlTypeName()) {
+    case INTEGER:
+      return new SqlDataTypeSpec(
+          new SqlAlienSystemTypeNameSpec(
+              "INT",
+              type.getSqlTypeName(),
+              SqlParserPos.ZERO),
+          SqlParserPos.ZERO);
+    case BIGINT:
```
NobiGo, line 119: If we don't add this case, what's the generate SQL ?

NobiGo (commented):
Some questions need to confirm.

caicancai:
@YiwenWu If you have time, can you take a look? I remember that Starrocks was adapted by you before.

mihaibudiu (approved):
I am not familiar with starrocks, but assuming the claims about the dialect are correct, the code looks fine.

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
NobiGo: Yes, I get it. This dialect extends MySQL dialect. That's why I generated the signed type.

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -109,6 +109,20 @@ public StarRocksSqlDialect(Context context) {
...
+      return new SqlDataTypeSpec(
+          new SqlAlienSystemTypeNameSpec(
+              "BIGINT",
```
NobiGo, line 122: The BIGINT exists in SqlBasicTypeNameSpec. So maybe we can use :
SqlBasicTypeNameSpec like VARCHAR in MySQL dialect.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3072,24 +3072,33 @@ private SqlDialect nonOrdinalDialect() {
     sql(query).dialect(HiveSqlDialect.DEFAULT).ok(expected);
   }
 
-  @Test void testMysqlCastToBigint() {
```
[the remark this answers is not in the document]
caicancai: +1

caicancai (commented):
minor: You need to modify your commit message

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
fanluoo: > The BIGINT exists in SqlBasicTypeNameSpec. So maybe we can use : SqlBasicTypeNameSpec like VARCHAR in MySQL dialect.

Thank you for your suggestion, but I am not very clear about the essential difference between these two writing methods, or is the latter a standardized writing method

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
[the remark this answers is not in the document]
fanluoo: > minor: recommended to add new unit tests and annotate corresponding jira link.

Thank you for your suggestion. I have added corresponding jira link comments, but I think this is the same type conversion issue. It would be better to handle different dialect specificity in one test case, otherwise there may be many redundant test cases

fanluoo:
> minor: You need to modify your commit message

Thank you for your reminder. I have made the necessary modifications

fanluoo:
> minor: You need to modify your commit message

@caicancai Thank you for your reminder. I have made the necessary modifications

mihaibudiu:
In order to merge this PR you need to squash the commits to a single commit with the same title as the JIRA issue and the PR request.

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
NobiGo: Hi @fanluoo , AFAIK, If a datatype exists in SqlTypeName, Then we use the SqlBasicTypeNameSpec, meas this datatype is a basic type in Calcite.  SqlAlienSystemTypeNameSpec represents a type name for an alien system. For example, UNSIGNED is a built-in type in MySQL which is synonym of INTEGER. and STRING is synonym of VARCHAR in Spark. UNSIGNED and STRING should used SqlAlienSystemTypeNameSpec.  So BIGINT should used the SqlBasicTypeNameSpec

NobiGo (changes requested):
Minor changes required

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
fanluoo: > Hi @fanluoo , AFAIK, If a datatype exists in SqlTypeName, Then we use the SqlBasicTypeNameSpec, meas this datatype is a basic type in Calcite. SqlAlienSystemTypeNameSpec represents a type name for an alien system. For example, UNSIGNED is a built-in type in MySQL which is synonym of INTEGER. and STRING is synonym of VARCHAR in Spark. UNSIGNED and STRING should used SqlAlienSystemTypeNameSpec. So BIGINT should used the SqlBasicTypeNameSpec

@NobiGo The modifications have been made

NobiGo (approved):
LGTM

NobiGo:
Please squash the commits.

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -19,14 +19,7 @@
...
 import org.apache.calcite.rel.type.RelDataType;
-import org.apache.calcite.sql.SqlAbstractDateTimeLiteral;
-import org.apache.calcite.sql.SqlAlienSystemTypeNameSpec;
-import org.apache.calcite.sql.SqlCall;
-import org.apache.calcite.sql.SqlDataTypeSpec;
-import org.apache.calcite.sql.SqlDialect;
-import org.apache.calcite.sql.SqlLiteral;
-import org.apache.calcite.sql.SqlNode;
-import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.*;
```
caicancai, line 22: @fanluoo calcite does not allow import.*

fanluoo: > @fanluoo calcite does not allow import.*
Thank you for the reminder, Let me make some adjustments

NobiGo (approved):

PR #3809, opened 2024-06-04
[CALCITE-6427] Use a higher precision for DECIMAL intermediate results for some aggregate functions like STDDEV

mihaibudiu:
@zabetak hopefully this PR will fix the problem of STDDEV requiring a higher declared precision when combined with #3733

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
@@ -1591,12 +1591,13 @@ public class StandardConvertletTable extends ReflectiveConvertletTable {
       //     / (count(x1, x2) - 1)
       final SqlParserPos pos = SqlParserPos.ZERO;
       final SqlLiteral nullLiteral = SqlLiteral.createNull(SqlParserPos.ZERO);
+      final RelDataType highPrecision = AvgVarianceConvertlet.highPrecision(cx, varType);
 
       final RexNode arg0Rex = cx.convertExpression(arg0Input);
       final RexNode arg1Rex = cx.convertExpression(arg1Input);
 
-      final SqlNode arg0 = getCastedSqlNode(arg0Input, varType, pos, arg0Rex);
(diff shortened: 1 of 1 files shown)


mihaibudiu:
No tests were affected because today casts to decimal values are noops.

mihaibudiu:
I will try this commit in combination with #3733 to see whether it handles STDDEV. If it does, I will merge this one first, then amend #3733, and then hopefully merge that one too.

mihaibudiu:
Does anyone want to review this PR?
This is blocking some other more important PRs, and it's not too complicated.

NobiGo:
> Does anyone want to review this PR? This is blocking some other more important PRs, and it's not too complicated.

I will review in the next week.

NobiGo:
Strangely, we don't need to change any unit test.

NobiGo:
> I will try this commit in combination with #3733 to see whether it handles STDDEV. If it does, I will merge this one first, then amend #3733, and then hopefully merge that one too.

This PR works good with #3733 ?

NobiGo:
Use a higher precision for DECIMAL by precision*2 and scala*2, Do we refer to any database or  SQL standard？I just want to confirm.

mihaibudiu:
> Strangely, we don't need to change any unit test.

That is because casts to decimal have no effect

mihaibudiu:
> Use a higher precision for DECIMAL by precision_2 and scala_2, Do we refer to any database or SQL standard？I just want to confirm.

This is independent on the database. It is three enumerable implementation. Arguably, some of these choices should be part of the type system.

mihaibudiu:
> > I will try this commit in combination with #3733 to see whether it handles STDDEV. If it does, I will merge this one first, then amend #3733, and then hopefully merge that one too.
> 
> This PR works good with #3733 ?

The current version of #3733 includes this pr. This shows that this pr solves the problem for stddev but not for other aggregates like variance. I still want to merge this first.

mihaibudiu:
In the interest of unblocking #3733 I will merge this and rebase the other PR.

PR #3811, opened 2024-06-04
[CALCITE-5855] Implement frame exclusion in window functions

itiels:
This PR adds support for frame exclusion clause in window functions.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/core/Window.java
+++ b/core/src/main/java/org/apache/calcite/rel/core/Window.java
@@ -39,6 +39,7 @@ import org.apache.calcite.rex.RexLocalRef;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexSlot;
 import org.apache.calcite.rex.RexWindowBound;
+import org.apache.calcite.rex.RexWindowExclusion;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.util.ImmutableBitSet;
 import org.apache.calcite.util.ImmutableIntList;
@@ -223,8 +224,8 @@ public abstract class Window extends SingleRel implements Hintable {
   /**
    * Group of windowed aggregate calls that have the same window specification.
    *
-   * <p>The specification is defined by an upper and lower bound, and
-   * also has zero or more partitioning columns.
+   * <p>The specification is defined by an upper and lower bound, exclusion clause,
+   * and also has zero or more partitioning columns.
    *
    * <p>A window is either logical or physical. A physical window is measured
    * in terms of row count. A logical window is measured in terms of rows
@@ -247,6 +248,7 @@ public abstract class Window extends SingleRel implements Hintable {
     public final boolean isRows;
     public final RexWindowBound lowerBound;
     public final RexWindowBound upperBound;
+    public final RexWindowExclusion exclude;
     public final RelCollation orderKeys;
     private final String digest;
 
@@ -262,12 +264,14 @@ public abstract class Window extends SingleRel implements Hintable {
         boolean isRows,
         RexWindowBound lowerBound,
         RexWindowBound upperBound,
+        RexWindowExclusion exclude,
         RelCollation orderKeys,
         List<RexWinAggCall> aggCalls) {
       this.keys = Objects.requireNonNull(keys, "keys");
       this.isRows = isRows;
       this.lowerBound = Objects.requireNonNull(lowerBound, "lowerBound");
       this.upperBound = Objects.requireNonNull(upperBound, "upperBound");
+      this.exclude = exclude;
       this.orderKeys = Objects.requireNonNull(orderKeys, "orderKeys");
       this.aggCalls = ImmutableList.copyOf(aggCalls);
       this.digest = computeString();
@@ -319,6 +323,9 @@ public abstract class Window extends SingleRel implements Hintable {
         buf.append(lowerBound);
         buf.append(" and ");
         buf.append(upperBound);
+        if (exclude != RexWindowExclusion.EXCLUDE_NO_OTHER) {
+          buf.append(" ").append(exclude);
+        }
       }
       if (!aggCalls.isEmpty()) {
         if (buf.length() > i) {
@@ -358,7 +365,9 @@ public abstract class Window extends SingleRel implements Hintable {
     public boolean isAlwaysNonEmpty() {
       int lowerKey = lowerBound.getOrderKey();
       int upperKey = upperBound.getOrderKey();
-      return lowerKey > -1 && lowerKey <= upperKey;
+      return lowerKey > -1 && lowerKey <= upperKey
+            && (exclude == RexWindowExclusion.EXCLUDE_NO_OTHER
+               || exclude == RexWindowExclusion.EXCLUDE_TIES);
     }
 
     /**
--- a/core/src/main/java/org/apache/calcite/rel/logical/LogicalWindow.java
+++ b/core/src/main/java/org/apache/calcite/rel/logical/LogicalWindow.java
@@ -34,6 +34,7 @@ import org.apache.calcite.rex.RexProgram;
 import org.apache.calcite.rex.RexShuttle;
 import org.apache.calcite.rex.RexWindow;
 import org.apache.calcite.rex.RexWindowBound;
+import org.apache.calcite.rex.RexWindowExclusion;
 import org.apache.calcite.tools.RelBuilder;
 import org.apache.calcite.util.ImmutableBitSet;
 import org.apache.calcite.util.Litmus;
@@ -68,8 +69,8 @@ public final class LogicalWindow extends Window {
    *
    * @param cluster Cluster
    * @param traitSet Trait set
-   * @param hints   Hints for this node
-   * @param input   Input relational expression
+   * @param hints Hints for this node
+   * @param input Input relational expression
    * @param constants List of constants that are additional inputs
    * @param rowType Output row type
    * @param groups Window groups
@@ -199,6 +200,7 @@ public final class LogicalWindow extends Window {
               windowKey.isRows,
               windowKey.lowerBound.accept(toInputRefs),
               windowKey.upperBound.accept(toInputRefs),
+              windowKey.exclude,
               windowKey.orderKeys,
               aggCalls));
     }
@@ -334,22 +336,25 @@ public final class LogicalWindow extends Window {
     private final boolean isRows;
     private final RexWindowBound lowerBound;
     private final RexWindowBound upperBound;
+    private final RexWindowExclusion exclude;
 
     WindowKey(
         ImmutableBitSet groupSet,
         RelCollation orderKeys,
         boolean isRows,
         RexWindowBound lowerBound,
-        RexWindowBound upperBound) {
+        RexWindowBound upperBound,
+        RexWindowExclusion exclude) {
       this.groupSet = groupSet;
       this.orderKeys = orderKeys;
       this.isRows = isRows;
       this.lowerBound = lowerBound;
       this.upperBound = upperBound;
+      this.exclude = exclude;
     }
 
     @Override public int hashCode() {
-      return Objects.hash(groupSet, orderKeys, isRows, lowerBound, upperBound);
+      return Objects.hash(groupSet, orderKeys, isRows, lowerBound, upperBound, exclude);
     }
 
     @Override public boolean equals(@Nullable Object obj) {
@@ -359,6 +364,7 @@ public final class LogicalWindow extends Window {
           && orderKeys.equals(((WindowKey) obj).orderKeys)
           && Objects.equals(lowerBound, ((WindowKey) obj).lowerBound)
           && Objects.equals(upperBound, ((WindowKey) obj).upperBound)
+          && exclude == ((WindowKey) obj).exclude
           && isRows == ((WindowKey) obj).isRows;
     }
   }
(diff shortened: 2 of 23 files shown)


core/src/test/java/org/apache/calcite/rel/logical/ToLogicalConverterTest.java
```diff
@@ -467,6 +467,22 @@ private void verify(RelNode rel, String expectedPhysical, String expectedLogical
...
+        + " UNBOUNDED PRECEDING and CURRENT ROW EXCLUDE GROUP aggs [SUM($11)])])\n"
+        + "    JdbcToEnumerableConverter\n"
+        + "      JdbcTableScan(table=[[foodmart, employee]])\n";
+    String expectedLogical = ""
+        + "LogicalProject($0=[$17])\n"
+        + "  LogicalWindow(window#0=[window(order by [9] rows between"
+        + " UNBOUNDED PRECEDING and CURRENT ROW EXCLUDE GROUP aggs [SUM($11)])])\n"
+        + "    LogicalTableScan(table=[[foodmart, employee]])\n";
+    verify(rel(sql), expectedPhysical, expectedLogical);
+  }
```
mihaibudiu, line 485: missing empty line?
do you want to test the other kinds of exclusions too?

core/src/main/java/org/apache/calcite/rel/core/Window.java
```diff
@@ -358,7 +365,8 @@ public RelCollation collation() {
     public boolean isAlwaysNonEmpty() {
       int lowerKey = lowerBound.getOrderKey();
       int upperKey = upperBound.getOrderKey();
-      return lowerKey > -1 && lowerKey <= upperKey;
+      return lowerKey > -1 && lowerKey <= upperKey
```
mihaibudiu, line 368: wouldn't the "no ties" also apply here?

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -389,6 +389,7 @@ public RexNode makeOver(
       ImmutableList<RexFieldCollation> orderKeys,
       RexWindowBound lowerBound,
       RexWindowBound upperBound,
+      RexWindowExclusion exclude,
```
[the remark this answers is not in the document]
asolimando: The current policy is exactly what @mihaibudiu suggests, you should leave the old API around with a default value for the new parameter (exclude no others?), at that point no need to document in `history.md` IMO.

By doing this, the number of files to be updated in the scope of this PR will drop.

If you want to deprecate the old API, you can do so with the `@Deprecated` annotation (see [SqlDrop.java#L37](https://github.com/apache/calcite/blob/3c4d54a82e21ec6926e6aa724ce94ead63d19083/core/src/main/java/org/apache/calcite/sql/SqlDrop.java#L37) for an example), but I don't think it's necessary here.

asolimando (changes requested):
Please leave the old API as suggested, in order to avoid unnecessary breaking changes.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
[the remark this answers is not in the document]
itiels: Thanks for reviewing @mihaibudiu and @asolimando . 
So should I add overload option for each public method that I've changed? And how do I decide if the old versions should be deprecated or not?

core/src/main/java/org/apache/calcite/rel/core/Window.java
itiels: You are correct, great catch.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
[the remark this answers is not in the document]
asolimando: From a cursory look, the SQL parser still accepts the same syntax (without `exclude` statement) after your modifications.

If that's confirmed, then I think there is no need to deprecate the old APIs, as they would basically mimic the same (legal) old behaviour.

In a nut-shell, you are adding an optional functionality, what was working before should still be working as-is, it's not a change in behaviour.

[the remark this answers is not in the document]
itiels: Ok. I've pushed the changes, please let me know if that is what you had in mind.

core/src/main/java/org/apache/calcite/rel/logical/LogicalWindow.java
```diff
@@ -66,13 +67,13 @@ public final class LogicalWindow extends Window {
    *
    * <p>Use {@link #create} unless you know what you're doing.
    *
-   * @param cluster Cluster
-   * @param traitSet Trait set
-   * @param hints   Hints for this node
-   * @param input   Input relational expression
+   * @param cluster   Cluster
```
asolimando, line 70: In the codebase we generally don't format javadoc as you do here, can you please revert these changes?

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -958,6 +959,16 @@ public SqlNode toSql(RexWindowBound rexWindowBound) {
       }
     }
 
+    public SqlLiteral toSql(RexWindowExclusion exclude) {
+      switch (exclude) {
+      case EXCLUDE_CURRENT_ROW: return SqlWindow.createExcludeCurrentRow(POS);
+      case EXCLUDE_GROUP: return SqlWindow.createExcludeGroup(POS);
+      case EXCLUDE_TIES: return SqlWindow.createExcludeTies(POS);
+      }
+      assert exclude == RexWindowExclusion.EXCLUDE_NO_OTHER;
```
asolimando, line 968: Nit: why not a regular case and throwing on `default`? 

I am also wondering if we want to omit printing the exclusion when there is no exclusion, it would keep the generated SQL backward compatible to what we have been generating now. WDYT?

core/src/main/java/org/apache/calcite/rex/RexWindowExclusion.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Creates a window exclusion from a {@link SqlNode}.
+   *
+   * @param node SqlLiteral of the exclusion clause
+   * @return RexWindowExclusion
+   */
+  public static RexWindowExclusion create(SqlNode node) {
+    SqlLiteral exclude = (SqlLiteral) node;
+    if (SqlWindow.isExcludeCurrentRow(exclude)) {
```
asolimando, line 50: Nit: what about a switch similar to what you wrote for `SqlImplementor`?

core/src/test/resources/sql/winagg.iq
```diff
@@ -839,5 +839,66 @@ EnumerableCalc(expr#0..3=[{inputs}], GENDER=[$t2], $1=[$t3])
...
+!ok
+
+# Test excluding. Verified against Postgres
+!use scott
+
+select empno,
+  avg(comm) over (order by empno rows 2 preceding exclude current row) as e1,
+  avg(comm) over (order by empno rows 2 preceding exclude group) as e2,
+  avg(comm) over (order by empno rows 2 preceding exclude ties) as e3,
+  avg(comm) over (order by empno rows 2 preceding exclude no others) as e4
```
asolimando, line 851: Can you also add the same line but without "exclude no others" to verify they results in the same resultset? (here and in similar added tests)

site/_docs/algebra.md
```diff
@@ -530,19 +530,20 @@ call and then call its `over()` method, for instance `count().over()`.
...
+| `rowsTo(upper)`                                    | Creates a row-based window bounded above, `ROWS BETWEEN CURRENT ROW AND upper`
+| `exclude(excludeType)`                             | Exclude certain rows from the frame (see SQL `EXCLUDE`)
+| `rowsBetween(lower, upper)`                        | Creates a rows-based window, `ROWS BETWEEN lower AND upper`
 | `partitionBy(expr...)`<br/>`partitionBy(exprList)` | Partitions the window on the given expressions (see SQL `PARTITION BY`)
-| `orderBy(expr...)`<br/>`sort(exprList)` | Sorts the rows in the window (see SQL `ORDER BY`)
-| `allowPartial(b)`    | Sets whether to allow partial width windows; default true
-| `nullWhenCountZero(b)` | Sets whether whether the aggregate function should evaluate to null if no rows are in the window; default false
-| `as(alias)`          | Assigns a column alias (see SQL `AS`) and converts this `OverCall` to a `RexNode`
-| `toRex()`            | Converts this `OverCall` to a `RexNode`
+| `orderBy(expr...)`<br/>`sort(exprList)`            | Sorts the rows in the window (see SQL `ORDER BY`)
```
asolimando, line 545: Nit: can you remove the extra spaces at the end of these lines?

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -5780,6 +5780,22 @@ private static Matcher<SqlNode> isCharLiteral(String s) {
             + "FROM `BIDS`");
   }
 
+  @Test void testWindowSpecExclusion() {
+    // parses correctly
```
asolimando, line 5784: The comment is probably not needed as we have the `ok` method just after

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -5780,6 +5780,22 @@ private static Matcher<SqlNode> isCharLiteral(String s) {
...
+    sql("select sum(x) over (order by x rows between 2 preceding and 2 following exclude group) "
+        + "from emp")
+        .ok("SELECT (SUM(`X`) OVER (ORDER BY `X` ROWS BETWEEN 2 PRECEDING AND 2 "
+            + "FOLLOWING EXCLUDE GROUP))\n"
+            + "FROM `EMP`");
+    // doesn't parse without frame definition
+    sql("select sum(x) over (order by x ^exclude^ current row) from bids")
```
asolimando, line 5791: Why do we need the `'^'` here?

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -5780,6 +5780,22 @@ private static Matcher<SqlNode> isCharLiteral(String s) {
...
+        .fails("(?s).*Encountered \"exclude\".*");
+    // EXCLUDE NO OTHERS is the default behavior, and ommited from UNPARSE
+    sql("select sum(x) over (order by x rows between 2 preceding and 2 following exclude no others)"
+        + " from emp")
+        .ok("SELECT (SUM(`X`) OVER (ORDER BY `X` ROWS BETWEEN 2 PRECEDING AND 2 FOLLOWING))\n"
+            + "FROM `EMP`");
+  }
```
asolimando, line 5798: Please add a newline here

asolimando (changes requested):
LGTM, some more comments but all minor

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
mihaibudiu: @asolimando this is a convention used to indicate the error position reported by the parser in the statement

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
itiels: I've replace this assert with exception. 
Regarding generated SQL, we do ignore the exclude in case of `Exclude no others` in `SqlWindow.unparse`.

core/src/main/java/org/apache/calcite/rex/RexWindowExclusion.java
itiels: Maybe I'm missing something, but I think I can't do that here because I have `SqlLiteral` and not an Enum.

itiels:
Thanks again @asolimando! 
I've fixed according to your requests, and fixed some of the sonar cloud issues (some were indeed out of scope).

asolimando (approved):
LGTM, @itiels thanks for taking care of the comments, since @mihaibudiu already approved I think you can optimistically squash the commits for preparing to merge the PR (please make sure the only commit message aligns with the Jira title etc.). 

I generally wait one or more extra days before merging to give time for other people to react.

mihaibudiu (approved):

site/_docs/history.md
```diff
@@ -49,6 +49,8 @@ Guava versions 21.0 to 32.1.3-jre;
 other software versions as specified in gradle.properties.
 
 #### New features
+* [<a href="https://issues.apache.org/jira/browse/CALCITE-5855">CALCITE-5855</a>]
```
NobiGo, line 52: This file is a template file, we don't need to change this. The release manager will modify it.

site/_docs/algebra.md
```diff
@@ -530,8 +530,6 @@ call and then call its `over()` method, for instance `count().over()`.
 
 To further modify the `OverCall`, call its methods:
 
-| Method               | Description
```
NobiGo, line 533: This change is necessary?

core/src/test/resources/sql/winagg.iq
```diff
@@ -839,5 +839,68 @@ EnumerableCalc(expr#0..3=[{inputs}], GENDER=[$t2], $1=[$t3])
 +--------+--------+
 (6 rows)
 
+!ok
+
+# Test excluding. Verified against Postgres
```
NobiGo, line 844: PostgreSQL may be better.

NobiGo: Adding the Jira summary is better.

NobiGo (changes requested):
Need some change to confirm.

itiels:
Thanks for the review @NobiGo , I've fixes the issues. 
@asolimando I've squashed the commits. Thank you.

mihaibudiu:
@NobiGo if you are satisfied we can merge this

PR #3817, opened 2024-06-08
[CALCITE-6431] Implement the SINGLE_VALUE aggregation in HiveSqlDiale…

NobiGo:
…ct And SparkSQLDialect


mihaibudiu:
There are lots of changes in the quoting of identifiers.
Is this a bug that is being fixed?
If so, can it be a separate PR?

NobiGo:
> There are lots of changes in the quoting of identifiers. Is this a bug that is being fixed? If so, can it be a separate PR?

Yes,If we don't add the quating, Spark will parse failed the alias `$f0`, So we need add this. And Others database dialect almost have added this. Of course, We can separate this PR.

mihaibudiu:
Can you make a separate issue about quoting then?
Too many unrelated changes in this PR.

NobiGo:
> Can you make a separate issue about quoting then? Too many unrelated changes in this PR.

Please check the CALCITE-6434.

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
```diff
@@ -93,6 +101,41 @@ public static void unparseSparkArrayAndMap(SqlWriter writer,
...
 
+
+  /**
+   * Rewrite SINGLE_VALUE use SparkSQL Mode.
+   *
+   * <blockquote><pre>
+   * CASE COUNT(*)
+   * WHEN 0 THEN NULL
+   * WHEN 1 THEN MIN(&lt;result&gt;)
+   * ELSE (SELECT NULL UNION ALL SELECT NULL))
```
YiwenWu, line 112: 1. A problem: `SELECT NULL UNION ALL SELECT NULL` will return two NULL values ​​instead of one.
2. Suggestion: Use `COUNT(field)` instead of `COUNT(*)`

NobiGo: > 1. SELECT NULL UNION ALL SELECT NULL

this is intentional to make sure when the subquery include more then one value ,This SQL will throw exception.

2: Suggestion: Use COUNT(field) instead of COUNT(*)

Please check CALCITE-6430.

mihaibudiu: You should add a comment in the code about the UNION ALL. Your intent is not obvious at all.

I am not sure how the runtime error will look; if the error will mention a UNION ALL operator, the users may be confused, since there is no UNION ALL in the original query, but I guess there isn't much you can do about this.

YiwenWu: > > 1. SELECT NULL UNION ALL SELECT NULL
> 
> this is intentional to make sure when the subquery include more then one value ,This SQL will throw exception.
> 
> 2: Suggestion: Use COUNT(field) instead of COUNT(*)
> 
> Please check [CALCITE-6430](https://issues.apache.org/jira/browse/CALCITE-6430).

got it, thanks. Based on count(*) and union all, the result of the subquery is guaranteed to be only one row.

YiwenWu (approved):
LGTM

mihaibudiu:
why is this a draft?

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
mihaibudiu: Please add the explanations you gave here as comments in the code.
It will make maintaining the code much easier.

NobiGo:
> why is this a draft?

Hi @mihaibudiu . I have add comment in the JIRA. This SQL can't run as expected in Spark And Hive.

PR #3821, opened 2024-06-12
[CALCITE-6437]For druid sql JSON_OBJECT() function results in RUNTIME_FAILURE when querying INFORMATION_SCHEMA.COLUMNS

AlbericByte:
# Fix the druid json_object issue.
# Description 
jira: [CALCITE-6437](https://issues.apache.org/jira/browse/CALCITE-6437)
druid issue: https://github.com/apache/druid/issues/16356
1. Now in druid, we will construct the SqlFunction instance as following:

```
private static final SqlFunction SQL_FUNCTION = OperatorConversions
        .operatorBuilder(FUNCTION_NAME)
        .operandTypeChecker(OperandTypes.variadic(SqlOperandCountRanges.from(1)))
        .operandTypeInference((callBinding, returnType, operandTypes) -> {
          RelDataTypeFactory typeFactory = callBinding.getTypeFactory();
          for (int i = 0; i < operandTypes.length; i++) {
            if (i % 2 == 0) {
              operandTypes[i] = typeFactory.createSqlType(SqlTypeName.VARCHAR);
              continue;
            }
            operandTypes[i] = typeFactory.createTypeWithNullability(
                typeFactory.createSqlType(SqlTypeName.ANY),
                true
            );
          }
        })
        .returnTypeInference(NESTED_RETURN_TYPE_INFERENCE)
        .functionCategory(SqlFunctionCategory.SYSTEM)
        .build();
```

2. We try to get SqlJsonObjectFunction as following:
```
 public @Nullable RexCallImplementor get(final SqlOperator operator) {
    if (operator instanceof SqlUserDefinedFunction) {
      org.apache.calcite.schema.Function udf =
          ((SqlUserDefinedFunction) operator).getFunction();
      if (!(udf instanceof ImplementableFunction)) {
        throw new IllegalStateException("User defined function " + operator
            + " must implement ImplementableFunction");
      }
      CallImplementor implementor =
          ((ImplementableFunction) udf).getImplementor();
      return wrapAsRexCallImplementor(implementor);
    } else if (operator instanceof SqlTypeConstructorFunction) {
      return map.get(SqlStdOperatorTable.ROW);
    }
    return map.get(operator);
  }
```
3. Here is issue, the operator is SqlFunction, but in the map, the key instance is SqlJsonObjectFunction, the type is not equals, so we can naver get from the map
4. So i try to fix this in this cr as following in SqlToRelConverter.
We can overwrite to be SqlJsonObjectFunction
5. And also probably there is other way to fix:
add one more construction of SqlJsonObjectFunction as following 
```
public SqlJsonObjectFunction(SqlFunction baseFunction)
```
and  update JsonObjectOperatorConversion in druid as following:
```
SQL_FUNCTION = SqlJsonObjectFunction(OperatorConversions
        .operatorBuilder(FUNCTION_NAME)
        .operandTypeChecker(OperandTypes.variadic(SqlOperandCountRanges.from(1)))
        .operandTypeInference((callBinding, returnType, operandTypes) -> {
          RelDataTypeFactory typeFactory = callBinding.getTypeFactory();
          for (int i = 0; i < operandTypes.length; i++) {
            if (i % 2 == 0) {
              operandTypes[i] = typeFactory.createSqlType(SqlTypeName.VARCHAR);
              continue;
            }
            operandTypes[i] = typeFactory.createTypeWithNullability(
                typeFactory.createSqlType(SqlTypeName.ANY),
                true
            );
          }
        })
        .returnTypeInference(NESTED_RETURN_TYPE_INFERENCE)
        .functionCategory(SqlFunctionCategory.SYSTEM)
        .build());
```
6. I am not sure which one is better, could you guy help to look, or better solution?

### DIFF
--- a/core/src/test/resources/sql/agg.iq
+++ b/core/src/test/resources/sql/agg.iq
@@ -3247,7 +3247,7 @@ select json_object('deptno': deptno, 'employees': json_arrayagg(json_object('ena
 +-------------------------------------------------------------------------------------------+
 (6 rows)
 
-!ok2
+!ok
 
 # [CALCITE-2786] Add order by clause support for JSON_ARRAYAGG
 select gender,
--- a/core/src/test/resources/sql/agg.iq
+++ b/core/src/test/resources/sql/agg.iq
@@ -3247,7 +3247,7 @@ select json_object('deptno': deptno, 'employees': json_arrayagg(json_object('ena
 +-------------------------------------------------------------------------------------------+
 (6 rows)
 
-!ok
+!ok2
 
 # [CALCITE-2786] Add order by clause support for JSON_ARRAYAGG
 select gender,
@@ -3776,6 +3776,23 @@ select distinct sum(deptno + '1') as deptsum from dept order by 1;
 |     104 |
 +---------+
 (1 row)
+
+!ok
+
+# Test cases for [CALCITE-6437] For druid sql JSON_OBJECT() function results in RUNTIME_FAILURE when querying
+# INFORMATION_SCHEMA.COLUMNS
+select json_object('name': job, 'value': deptno) as options
+from "scott".emp
+where deptno = 10
+group by deptno, job;
++---------------------------------+
+| OPTIONS                         |
++---------------------------------+
+| {"name":"CLERK","value":10}     |
+| {"name":"MANAGER","value":10}   |
+| {"name":"PRESIDENT","value":10} |
++---------------------------------+
+(3 rows)
 
 !ok
 # End agg.iq
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -6474,7 +6474,8 @@ public class SqlToRelConverter {
     }
 
     @Override public RexNode visitCall(RexCall call) {
-      if (call.getOperator() == SqlStdOperatorTable.JSON_OBJECT) {
+      if (call.getOperator().getName().equals(SqlStdOperatorTable.JSON_OBJECT.getName())
+          && call.getOperator().getKind() == SqlStdOperatorTable.JSON_OBJECT.getKind()) {
         final ImmutableList.Builder<RexNode> builder = ImmutableList.builder();
         for (int i = 0; i < call.operands.size(); ++i) {
           if ((i & 1) == 0 && i != 0) {
@@ -6485,7 +6486,8 @@ public class SqlToRelConverter {
         }
         return rexBuilder.makeCall(SqlStdOperatorTable.JSON_OBJECT, builder.build());
       }
-      if (call.getOperator() == SqlStdOperatorTable.JSON_ARRAY) {
+      if (call.getOperator().getName().equals(SqlStdOperatorTable.JSON_ARRAY.getName())
+          && call.getOperator().getKind() == SqlStdOperatorTable.JSON_ARRAY.getKind()) {
         final ImmutableList.Builder<RexNode> builder = ImmutableList.builder();
         builder.add(call.operands.get(0));
         for (int i = 1; i < call.operands.size(); ++i) {
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -6330,6 +6330,12 @@ public class SqlOperatorTest {
         "{\"foo\":{\"foo\":\"bar\"}}", "VARCHAR(2000) NOT NULL");
     f.checkString("json_object('foo': json_object('foo': 'bar') format json)",
         "{\"foo\":{\"foo\":\"bar\"}}", "VARCHAR(2000) NOT NULL");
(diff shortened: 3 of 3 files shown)


NobiGo:
@AlbericByte Please create a new issue in [JIRA](https://issues.apache.org/jira/browse/CALCITE).

AlbericByte:
> @AlbericByte Please create a new issue in [JIRA](https://issues.apache.org/jira/browse/CALCITE).

add the jira in the description : CALCITE-6437
Thanks for notification

mihaibudiu:
In order to make it easier to keep track of the correspondence of issues in JIRA and github the issue title should match exactly the JIRA title.
The commit that fixes the issue should also have the same message.

mihaibudiu:
Please add a unit test that fails before the fix.
The file SqlOperatorTest may be the right place for it.

AlbericByte:
> In order to make it easier to keep track of the correspondence of issues in JIRA and github the issue title should match exactly the JIRA title. The commit that fixes the issue should also have the same message.

Got it, thanks for the tips

AlbericByte:
> Please add a unit test that fails before the fix. The file SqlOperatorTest may be the right place for it.
@mihaibudiu and @NobiGo 
+ the issue only happen in a BindableProject.BINDABLE.[](input=BindableTableScan#983,exprs=[JSON_OBJECT(FLAG(NULL_ON_NULL), 'name', $3, 'type', $7)]).
`select JSON_OBJECT('name': COLUMN_NAME, 'type': DATA_TYPE)
from INFORMATION_SCHEMA.COLUMNS`

+ And i try to create a sql as following, but it is not a BindableProject.BINDABLE
`select json_object('key':col1, 'value':col2) from (SELECT *
FROM (VALUES (1, 'a'), (2, 'b'), (3, 'c'),(1, 'a'), (2, 'b'), (3, 'c')) AS t(col1, col2))`

+ Is there any meta table I can create a test query, then i can create a test case.

core/src/test/resources/sql/agg.iq
```diff
@@ -3247,7 +3247,7 @@ select json_object('deptno': deptno, 'employees': json_arrayagg(json_object('ena
 +-------------------------------------------------------------------------------------------+
 (6 rows)
 
-!ok
+!ok2
```
mihaibudiu, line 3250: this change looks strange

mihaibudiu (approved):
Besides the comment I left, this looks fine

core/src/test/resources/sql/agg.iq
```diff
@@ -3777,5 +3777,22 @@ select distinct sum(deptno + '1') as deptsum from dept order by 1;
 +---------+
 (1 row)
 
+!ok
+
+# Test cases for [CALCITE-6437] For druid sql JSON_OBJECT() function results in RUNTIME_FAILURE when querying
```
NobiGo, line 3782: This SQL can run success without this PR. And according to this change, I doubt this is a good way to fix this issue in Calcite.

NobiGo (changes requested):

AlbericByte:
> the issue arises from the fact that Druid tries to implement `JSON_OBJECT` with its own `SqlFunction`
> 
> Comparing against `SqlStdFunctions.X` enforces users of Calcite to use those functions. I think instead of resorting to String comparision it would be more straightforward to compare them based on their `SqlKind` instead.
> 
> So I think another way to solve this is to introduce `SqlKind.JSON_OBJECT` (and possibly do that for all `SqlStdOperators` ? ) How does that sound?
I have tried this solution to add SqlKind.JSON_OBJECT, and add this as following:
` public static final Set<SqlKind> FUNCTION =
      EnumSet.of(OTHER_FUNCTION, ROW, TRIM, LTRIM, RTRIM, CAST, REVERSE,
          JDBC_FN, JSON_OBJECT, JSON_ARRAY POSITION, CONVERT);`
But after modification, the test case will be failed
` expr("json_arrayagg(json_array(\"column\") format json)")
        .ok("JSON_ARRAYAGG(JSON_ARRAY(`column` ABSENT ON NULL) FORMAT JSON ABSENT ON NULL)");`
The expected result will be 
`"JSON_ARRAYAGG((JSON_ARRAY(`column` ABSENT ON NULL)) FORMAT JSON ABSENT ON NULL)"`
i need time to figure why there is one more parentheses， still need @NobiGo 'suggestion, does this solution is good or not?

AlbericByte:
@NobiGo @kgyrtkirk @mihaibudiu  in Druid the return type of json_object function is ExpressionType.NESTED_DATA, but in Calcite, the return type of json_object is VARCHAR(2000).

Follow current calcite logic, if can not find the sqlfunction in RexImpTable, calcite will throw exception. And actually in druid, we have implement json_object in JsonObjectExprMacro. is there any api to register new SqlJsonObjectFunction with NESTED_DATA in druid? so i can register a new SqlJsonObjectFunction with NESTED_DATA output.

i tried to reused SqlStdOperatorTable.JSON_OBJECT in druid, and also add cast or json_query to wrap the json_object, but it will fail because the SqlStdOperatorTable.JSON_OBJECT return varchar, or cast donot support NESTED_DATA.

mihaibudiu:
What is the status of this PR? Is it ready or more work is needed?

mihaibudiu:
There are many functions in Calcite which are "overloaded", which behave differently depending on the SQL dialect. You can define a new function, with the same name, but a different implementation for Druid.

PR #3824, opened 2024-06-16
[CALCITE-6440] Add optimization to returns empty when the Order By parameter in the Sort operator is empty

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6440


caicancai:
If there is a corresponding rule please let me know so I can close this jira case

PR #3826, opened 2024-06-18
[CALCITE-3094] Code of method grows beyond 64 KB when joining two tables with many fields

jduo:
Fix errors about the code size being too large for a method when joining thousands of fields.

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
@@ -22,6 +22,7 @@ import org.apache.calcite.linq4j.AbstractEnumerable;
 import org.apache.calcite.linq4j.Enumerable;
 import org.apache.calcite.linq4j.Enumerator;
 import org.apache.calcite.linq4j.JoinType;
+import org.apache.calcite.linq4j.Nullness;
 import org.apache.calcite.linq4j.Ord;
 import org.apache.calcite.linq4j.function.Function1;
 import org.apache.calcite.linq4j.function.Function2;
@@ -30,12 +31,14 @@ import org.apache.calcite.linq4j.tree.BlockBuilder;
 import org.apache.calcite.linq4j.tree.BlockStatement;
 import org.apache.calcite.linq4j.tree.ConstantExpression;
 import org.apache.calcite.linq4j.tree.ConstantUntypedNull;
+import org.apache.calcite.linq4j.tree.DeclarationStatement;
 import org.apache.calcite.linq4j.tree.Expression;
 import org.apache.calcite.linq4j.tree.ExpressionType;
 import org.apache.calcite.linq4j.tree.Expressions;
 import org.apache.calcite.linq4j.tree.FunctionExpression;
 import org.apache.calcite.linq4j.tree.MethodCallExpression;
 import org.apache.calcite.linq4j.tree.MethodDeclaration;
+import org.apache.calcite.linq4j.tree.NewArrayExpression;
 import org.apache.calcite.linq4j.tree.ParameterExpression;
 import org.apache.calcite.linq4j.tree.Primitive;
 import org.apache.calcite.linq4j.tree.Types;
@@ -61,6 +64,7 @@ import com.google.common.collect.ImmutableMap;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
+import java.lang.reflect.Array;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
@@ -160,6 +164,34 @@ public class EnumUtils {
     // Generate all fields.
     final List<Expression> expressions = new ArrayList<>();
     final int outputFieldCount = physType.getRowType().getFieldCount();
+
+    // If there are many output fields, create the output dynamically so that the code size stays
+    // below the limit. See CALCITE-3094.
+    final boolean generateCompactCode = outputFieldCount >= 100;
+    final ParameterExpression compactOutputVar;
+    final BlockBuilder compactCode = new BlockBuilder();
+    if (generateCompactCode) {
+      Class<?> fieldClass = physType.fieldClass(0);
+      // If all fields have the same type, use the specific type. Otherwise just use Object.
+      for (int fieldIndex = 1; fieldIndex < outputFieldCount; ++fieldIndex) {
+        if (fieldClass != physType.fieldClass(fieldIndex)) {
+          fieldClass = Object.class;
+          break;
+        }
+      }
+
+      final Class<?> arrayClass = Array.newInstance(fieldClass, 0).getClass();
+      compactOutputVar = Expressions.variable(arrayClass, "outputArray");
+      final DeclarationStatement exp =
+          Expressions.declare(
+              0, compactOutputVar, new NewArrayExpression(fieldClass, 1,
+              Expressions.constant(outputFieldCount), null));
+      compactCode.add(exp);
+    } else {
+      compactOutputVar = null;
+    }
+
+    int outputField = 0;
     for (Ord<PhysType> ord : Ord.zip(inputPhysTypes)) {
       final PhysType inputPhysType =
           ord.e.makeNullable(joinType.generatesNullsOn(ord.i));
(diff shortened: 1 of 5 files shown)


core/src/test/java/org/apache/calcite/test/LargeGeneratedJoinTest.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Test class for CALCITE-3094.
```
mihaibudiu, line 58: there is a certain style expected from the javadoc for such tests.
E.g.:

```
/** Test case for
  * <a href="https://issues.apache.org/jira/browse/CALCITE-5160">[CALCITE-5160]
  * ANY/SOME, ALL operators should support collection expressions</a>. */
```

core/src/test/java/org/apache/calcite/test/LargeGeneratedJoinTest.java
```diff
@@ -0,0 +1,156 @@
...
+      }
+    };
+
+    final CalciteSchema sp = CalciteSchema.createRootSchema(false, true);
+    sp.add("ROOT", rootSchema);
+
+    String sql = "SELECT * \n"
+        + "FROM ROOT.T0 \n"
+        + "JOIN ROOT.T1 \n"
+        // + "ON ROOT.T0.F_0 = ROOT.T1.F_0";
```
mihaibudiu, line 131: if this is not used maybe it can be deleted

core/src/main/java/org/apache/calcite/adapter/enumerable/JavaRowFormat.java
```diff
@@ -256,6 +262,25 @@ public enum JavaRowFormat {
...
+    @Override public Expression setFieldDynamic(Expression expression, Expression field,
+        Expression value) {
+      final IndexExpression e =
+          Expressions.arrayIndex(expression, Expressions.constant(field));
+      return Expressions.assign(e, value);
+    }
+
+    @Override public @Nullable Expression copy(ParameterExpression parameter,
+        ParameterExpression outputArray, int outputStartIndex, int length) {
+
```
mihaibudiu, line 279: why the empty lines?

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
```diff
@@ -189,6 +233,15 @@ static Expression joinSelector(JoinRelType joinType, PhysType physType,
         expressions.add(expression);
       }
     }
+
+    if (generateCompactCode) {
```
mihaibudiu, line 237: if you could add a comment showing here the shape of the generated code it could help people understand easier what is going on.

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
normanj-bitquill: @thomasrebele 

Here is the code generated from a unit test.
```
{
  return org.apache.calcite.linq4j.EnumerableDefaults.nestedLoopJoin(org.apache.calcite.schema.Schemas.queryable(root, root.getRootSchema().getSubSchema("ROOT"), org.apache.calcite.interpreter.Row.class, "T1").asEnumerable(), org.apache.calcite.schema.Schemas.queryable(root, root.getRootSchema().getSubSchema("ROOT"), org.apache.calcite.interpreter.Row.class, "T0").asEnumerable(), new org.apache.calcite.linq4j.function.Predicate2() {
      public boolean apply(org.apache.calcite.interpreter.Row left, org.apache.calcite.interpreter.Row right) {
        return true;
      }
      public boolean apply(Object left, Object right) {
        return apply(
          (org.apache.calcite.interpreter.Row) left,
          (org.apache.calcite.interpreter.Row) right);
      }
    }
    , new org.apache.calcite.linq4j.function.Function2() {
      public String[] apply(org.apache.calcite.interpreter.Row left, org.apache.calcite.interpreter.Row right) {
        String[] outputArray = new String[201];
        return outputArray;
      }
      public String[] apply(Object left, Object right) {
        return apply(
          (org.apache.calcite.interpreter.Row) left,
          (org.apache.calcite.interpreter.Row) right);
      }
    }
    , org.apache.calcite.linq4j.JoinType.INNER);
}
```

This is just returning an initialized String array of the correct size. The data isn't copied.

The issue could be from here:
https://github.com/Bit-Quill/calcite/blob/calcite-3094-method-too-large/core/src/main/java/org/apache/calcite/adapter/enumerable/JavaRowFormat.java#L347

What should `JavaRowFormat.copy()` return?

core/src/test/java/org/apache/calcite/test/LargeGeneratedJoinTest.java
normanj-bitquill: Updated.

core/src/test/java/org/apache/calcite/test/LargeGeneratedJoinTest.java
normanj-bitquill: Removed.

core/src/main/java/org/apache/calcite/adapter/enumerable/JavaRowFormat.java
normanj-bitquill: Removed.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumUtils.java
thomasrebele: Indeed, I had not implemented `JavaRowFormat.copy()`, as I didn't know how to do that. Seems that the input would be `org.apache.calcite.interpreter.Row` according to the generated code. I'm astonished that the return type is not of the same type.
The `JavaRowFormat.copy()` would need to return an `Expression` that represents the code to copy the data from the left and right `org.apache.calcite.interpreter.Row`. Not sure whether an `ARRAY_COPY` could be used here. Alternatively, the generated code could be a for loop, here some pseudo-code:
```
int outIdx = 0;
for(int leftIdx = 0; leftIdx < left length; leftIdx++) {
  outputArray[outIdx++] = left item at position leftIdx;
}
for(int rightIdx = 0; rightIdx < right length; rightIdx++) {
  outputArray[outIdx++] = right item at position rightIdx;
}
```

jduo: I've changed copy to generate the following code:

`new org.apache.calcite.linq4j.function.Function2() {
  public String[] apply(org.apache.calcite.interpreter.Row left, org.apache.calcite.interpreter.Row right) {
    String[] outputArray = new String[201];
    System.arraycopy(left.copyValues(), 0, outputArray, 0, 101);
    System.arraycopy(right.copyValues(), 0, outputArray, 101, 100);
    return outputArray;
  }
  public String[] apply(Object left, Object right) {
    return apply(
      (org.apache.calcite.interpreter.Row) left,
      (org.apache.calcite.interpreter.Row) right);
  }
}
`

I'll add that to the comments.

jduo: @mihaibudiu I think your comments are now addressed. Thanks!

core/src/main/java/org/apache/calcite/interpreter/Row.java
```diff
@@ -86,6 +86,8 @@ public static Row of(@Nullable Object...values) {
   }
 
   /** Returns a copy of the values. */
+  // Note: This implements BuildInMethod.ROW_COPY_VALUES.
+  @SuppressWarnings("unused")
```
mihaibudiu, line 90: There are lots of methods like this in the code, which are called using reflection.
I like the idea of using this annotation, but is there a guideline about which annotations should be used and which ones shouldn't?

mihaibudiu (approved):
Is this now working end-to-end?

core/src/main/java/org/apache/calcite/interpreter/Row.java
jduo: I'm not sure if there's a guideline for this in Calcite.
In other projects I've seen this annotation used to mark methods that are called via reflection or not used internally but needed in the public API to prevent accidental removal due to refactoring. It also has the nice side-effect of un-graying out in IDEs such as IntelliJ.

jduo:
> Is this now working end-to-end?

This is working end-to-end. I've also beefed up the supplied test to validate the data from the join result.

rubenada:
@mihaibudiu (or the others), is there any further comment? I think it looks in a good shape for merge (after squashing commits into a single one)

mihaibudiu:
I have no other comments.

PR #3828, opened 2024-06-21
[CALCITE-5634] Enable GREATEST, LEAST functions in PostgreSQL library

normanj-bitquill:
* Add new GREATEST and LEAST functions for PostgreSQL
* The new functions only return null if all arguments are null
* Created a new library for RedShift since it is not truly a union of Oracle and PostgreSQL functions

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -446,16 +446,30 @@ public abstract class SqlLibraryOperators {
           SqlFunctionCategory.STRING);
 
   /** The "GREATEST(value, value)" function. */
-  @LibraryOperator(libraries = {BIG_QUERY, ORACLE, REDSHIFT, SPARK})
+  @LibraryOperator(libraries = {BIG_QUERY, ORACLE, SPARK})
   public static final SqlFunction GREATEST =
       SqlBasicFunction.create(SqlKind.GREATEST,
           ReturnTypes.LEAST_RESTRICTIVE.andThen(SqlTypeTransforms.TO_NULLABLE),
           OperandTypes.SAME_VARIADIC);
 
+  /** The "GREATEST(value, value)" function. */
+  @LibraryOperator(libraries = {POSTGRESQL})
+  public static final SqlFunction GREATEST_PG =
+      SqlBasicFunction.create("GREATEST", SqlKind.GREATEST_PG,
+          ReturnTypes.LEAST_RESTRICTIVE.andThen(SqlTypeTransforms.TO_NULLABLE),
+          OperandTypes.SAME_VARIADIC);
+
   /** The "LEAST(value, value)" function. */
-  @LibraryOperator(libraries = {BIG_QUERY, ORACLE, REDSHIFT, SPARK})
+  @LibraryOperator(libraries = {BIG_QUERY, ORACLE, SPARK})
   public static final SqlFunction LEAST =
       SqlBasicFunction.create(SqlKind.LEAST,
+          ReturnTypes.LEAST_RESTRICTIVE.andThen(SqlTypeTransforms.TO_NULLABLE),
+          OperandTypes.SAME_VARIADIC);
+
+  /** The "GREATEST(value, value)" function. */
+  @LibraryOperator(libraries = {POSTGRESQL})
+  public static final SqlFunction LEAST_PG =
+      SqlBasicFunction.create("LEAST", SqlKind.LEAST_PG,
           ReturnTypes.LEAST_RESTRICTIVE.andThen(SqlTypeTransforms.TO_NULLABLE),
           OperandTypes.SAME_VARIADIC);
 
--- a/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
@@ -193,7 +193,9 @@ public class StandardConvertletTable extends ReflectiveConvertletTable {
         new TrimConvertlet(SqlTrimFunction.Flag.TRAILING));
 
     registerOp(SqlLibraryOperators.GREATEST, new GreatestConvertlet());
+    registerOp(SqlLibraryOperators.GREATEST_PG, new GreatestPgConvertlet());
     registerOp(SqlLibraryOperators.LEAST, new GreatestConvertlet());
+    registerOp(SqlLibraryOperators.LEAST_PG, new GreatestPgConvertlet());
     registerOp(SqlLibraryOperators.SUBSTR_BIG_QUERY,
         new SubstrConvertlet(SqlLibrary.BIG_QUERY));
     registerOp(SqlLibraryOperators.SUBSTR_MYSQL,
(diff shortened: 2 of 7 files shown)


mihaibudiu:
What's the relationship with #3142?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
```diff
@@ -72,6 +72,8 @@ public enum SqlLibrary {
   /** A collection of operators that are in PostgreSQL but not in standard
    * SQL. */
   POSTGRESQL("p", "postgresql"),
+  /** A collection of operators that are in RedShift but not in standard SQL. */
+  REDSHIFT("r", "redshift"),
```
mihaibudiu, line 76: This seems only tangentially related to this PR.
I wonder whether this deserves a separate issue and PR.

mihaibudiu (changes requested):
I haven't read everything, but I think you should break this PR into (at least) two separate PRs:
- adding support for Redshift
- implementing GREATEST, LEAST
You probably need to create a new issue for redshift.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
normanj-bitquill: @mihaibudiu I ran into a problem with this:
https://github.com/apache/calcite/blob/main/babel/src/test/java/org/apache/calcite/test/BabelQuidemTest.java#L110

Both Oracle and PostgreSQL have `GREATEST` and `LEAST` functions but they behave differently. I needed some way for RedShift to be able to choose only one implementation.

normanj-bitquill:
@mihaibudiu I'll split this as you suggest. The RedShift library change will go into a new PR.

normanj-bitquill:
@mihaibudiu This PR is now based on top of
https://github.com/apache/calcite/pull/3829

When 3829 is merged in, I can rebase this off of main.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
normanj-bitquill: This has been moved to another PR.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -101,7 +102,7 @@ private SqlLibraryOperators() {
    * converts the timezone of {@code datetime} from {@code tz1} to {@code tz2}.
    * This function is only on Redshift, but we list it in PostgreSQL
    * because Redshift does not have its own library. */
-  @LibraryOperator(libraries = {POSTGRESQL})
+  @LibraryOperator(libraries = {POSTGRESQL, REDSHIFT})
```
mihaibudiu, line 105: I assume all these REDSHIFT should also move to the other PR.

site/_docs/reference.md
```diff
@@ -2707,18 +2708,18 @@ In the following:
...
 | m p | CONCAT_WS(separator, str1 [, string ]*)      | Concatenates one or more strings, returns null only when separator is null, otherwise treats null arguments as empty strings
 | q | CONCAT_WS(separator, str1, str2 [, string ]*)  | Concatenates two or more strings, requires at least 3 arguments (up to 254), treats null arguments as empty strings
 | m | COMPRESS(string)                               | Compresses a string using zlib compression and returns the result as a binary string
 | b | CONTAINS_SUBSTR(expression, string [ , json_scope =&gt; json_scope_value ]) | Returns whether *string* exists as a substring in *expression*. Optional *json_scope* argument specifies what scope to search if *expression* is in JSON format. Returns NULL if a NULL exists in *expression* that does not result in a match
 | q | CONVERT(type, expression [ , style ])          | Equivalent to `CAST(expression AS type)`; ignores the *style* operand
-| p | CONVERT_TIMEZONE(tz1, tz2, datetime)           | Converts the timezone of *datetime* from *tz1* to *tz2*
+| p r | CONVERT_TIMEZONE(tz1, tz2, datetime)         | Converts the timezone of *datetime* from *tz1* to *tz2*
```
mihaibudiu, line 2722: I also assume all these "r" changes will move to the other PR as well.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -1900,6 +1902,60 @@ private static class GreatestConvertlet implements SqlRexConvertlet {
     }
   }
 
+  /** Convertlet that converts {@code GREATEST} and {@code LEAST}. */
+  private static class GreatestPgConvertlet implements SqlRexConvertlet {
+    @Override public RexNode convertCall(SqlRexContext cx, SqlCall call) {
```
mihaibudiu, line 1907: this expansion is quadratic in the list size.
is this something we should worry about?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: They will for everything except `GREATEST` and `LEAST`, since Redshift should also use the new implementation.

site/_docs/reference.md
normanj-bitquill: That is correct. For this PR, they will only be added to `GREATEST` and `LEAST`.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
normanj-bitquill: I based this on the existing `GreatestConvertlet` class. It converts in a very similar way. The original `GreatestConvertlet` was written 8 years ago, so a different solution may make more sense now.

In general, the user calls the function with a list of scalars, such as `GREATEST(1, 2, 3, 4)`. There must be an upper limit, since it must fit in a SQL call that can be parsed.

mihaibudiu: There are some PRs attempting to solve https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-3094, which happens when a table has too many columns. So what we assume as a small constant can actually be big.

I could see cases where compiler-generated code could produce a long list as an argument for GREATEST. 
So I wonder whether throwing an exception for a very long list of arguments (e.g., >100) would make sense.

mihaibudiu:
In any case, I will review this again after it's rebased on #3829

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
normanj-bitquill: It would be preferable to give a useful error message to the caller, such as "too many arguments". I don't have a good idea what would make sense for the limit.

I tested `CONCAT` in PostgreSQL and it gives an error after 100 arguments. It probably makes sense to bring that limitation into Calcite.

I also tested `GREATEST` in PostgreSQL. PostgreSQL must do something clever with the arguments, since it works up to at least 1001 arguments.

mihaibudiu: The problem is that you want to compile to sql. In you use a language with local variables you can do it in size O(N). And postgres is not constrained to compile to sql.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: This has been cleaned up. No need to specify Redshift here since it will by default get PostgreSQL functions.

site/_docs/reference.md
normanj-bitquill: This is now cleaned up and only GREATEST and LEAST have changes.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
normanj-bitquill: An expression like `GREATEST(1, 2, ...)` is compiled and becomes a `SqlCall` object with arguments. The convertlet will then replace the `SqlCall` object with a new `SqlCall` for `CASE`. We are only manipulating objects after they are parsed, we do not generate SQL text and parse it again.

This will still explode original expression and use potentially much more memory.

With simple integer arguments, this PR works with at least 1000 values.

If no limit is in place, the user is still restricted by the max query length and what will fit in memory after the convertlet is run.

mihaibudiu: yes, your intermediate language is as expressive as SQL. That's what I mean.
But if you have let expressions you can do much better.

mihaibudiu (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10747,8 +10747,30 @@ void assertSubFunReturns(boolean binary, String s, int start,
...
+      f.checkString("greatest('on', 'earth')", "on   ", "CHAR(5) NOT NULL");
+      f.checkString("greatest('show', 'on', 'earth')", "show ",
+          "CHAR(5) NOT NULL");
+      f.checkScalar("greatest(12, CAST(NULL AS INTEGER), 3)", "12",
+          "INTEGER");
+      f.checkScalar("greatest(false, true)", true, "BOOLEAN NOT NULL");
+
+      final SqlOperatorFixture f12 = f.forOracle(SqlConformanceEnum.ORACLE_12);
+      f12.checkString("greatest('on', 'earth')", "on", "VARCHAR(5) NOT NULL");
+      f12.checkString("greatest('show', 'on', 'earth')", "show",
```
caicancai, line 10770: You should add negative test if the function returns null or fails

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10772,8 +10794,30 @@ void assertSubFunReturns(boolean binary, String s, int start,
...
+      f.checkString("least('on', 'earth')", "earth", "CHAR(5) NOT NULL");
+      f.checkString("least('show', 'on', 'earth')", "earth",
+          "CHAR(5) NOT NULL");
+      f.checkScalar("least(12, CAST(NULL AS INTEGER), 3)", "3",
+          "INTEGER");
+      f.checkScalar("least(false, true)", false, "BOOLEAN NOT NULL");
+
+      final SqlOperatorFixture f12 = f.forOracle(SqlConformanceEnum.ORACLE_12);
+      f12.checkString("least('on', 'earth')", "earth", "VARCHAR(5) NOT NULL");
+      f12.checkString("least('show', 'on', 'earth')", "earth",
```
caicancai, line 10817: above all

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10747,8 +10747,30 @@ void assertSubFunReturns(boolean binary, String s, int start,
           "VARCHAR(5) NOT NULL");
     };
     final List<SqlLibrary> libraries =
-        list(SqlLibrary.BIG_QUERY, SqlLibrary.ORACLE, SqlLibrary.REDSHIFT,
-            SqlLibrary.SPARK);
+        list(SqlLibrary.BIG_QUERY, SqlLibrary.ORACLE, SqlLibrary.SPARK);
+    f0.forEachLibrary(libraries, consumer);
+  }
+
+  @Test void testGreatestPgFunc() {
```
caicancai, line 10754: Add jira link

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
normanj-bitquill: You are right. Let expressions would avoid repeated evaluation of the arguments. Does Calcite support LET expressions?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
normanj-bitquill: Added a test where both arguments are null.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
normanj-bitquill: Added Javadoc comments for the Jira ticket.

caicancai (approved):
LGTM

normanj-bitquill:

mihaibudiu (approved):

PR #3829, opened 2024-06-21
[CALCITE-6444] Add a function library for Amazon Redshift

normanj-bitquill:


mihaibudiu (approved):
I haven't checked that indeed each of the affected SQL dialects has the exact functions modified in this PR. But otherwise the PR looks great.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -233,7 +234,7 @@ private static SqlCall transformConvert(SqlValidator validator, SqlCall call) {
       };
 
   /** The "DECODE(v, v1, result1, [v2, result2, ...], resultN)" function. */
-  @LibraryOperator(libraries = {ORACLE, SPARK})
+  @LibraryOperator(libraries = {ORACLE, REDSHIFT, SPARK})
```
caicancai, line 237: Should you add tests to SqlOperatorTest?

NobiGo (approved):

caicancai (approved):
Overall, LGTM

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: I have updated the `SqlOperatorTest` tests to also test for the REDSHIFT library where it makes sense.

mihaibudiu (approved):
Since you have 3 approvals, perhaps you can go ahead and squash the commits.

normanj-bitquill:
@mihaibudiu I have squashed the commits. Let's give this another day just in case anyone has further thoughts on the mailing list.

mihaibudiu:
I have updated the PR title

normanj-bitquill:
I have reworked this based on comments from @tanclary . A `SqlLibrary` can specify a parent. It will by default inherit all functions from the parent. The `SqlOperator` can now specify libraries that it should be excluded from. Used together it is possible to:
* Have a `SqlOperator` that is only in Redshift
* Have a `SqlOperator` that supports both PostgreSQL and Redshift, but only explicitly state PostgreSQL
* Have a `SqlOperator` that is only in PostgreSQL, not Redshift

mihaibudiu:
I have reviewed the updated code. Maybe you can squash the commits again.

normanj-bitquill:
@mihaibudiu I have squashed the commits.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
```diff
@@ -172,5 +187,20 @@ private static void addParent(Set<SqlLibrary> set, SqlLibrary library) {
       builder.put(value.fun, value);
     }
     MAP = builder.build();
+    final Map<SqlLibrary, ImmutableSet.Builder<SqlLibrary>> childrenMap = new HashMap<>();
```
julianhyde, line 190: A simpler - slightly less efficient - implementation would use a function
```
  Set<SqlLibrary> children(SqlLibrary library) {
    return Stream.of(values()).filter(lib -> lib.parent == library).collect(toSet());
  }
```
and then you could build the map directly (without an intermediate map of `ImmutableSet.builder`)

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperatorTableFactory.java
```diff
@@ -137,10 +139,21 @@ private static boolean operatorIsInLibrary(String operatorName, Field field,
       throw new AssertionError("Operator must belong to at least one library: "
           + operatorName);
     }
+    final Set<SqlLibrary> excludeLibrarySet = new HashSet<>();
```
julianhyde, line 142: fold these two lines into one.

use `ImmutableSet.of` (which, I believe, optimizes for enums)

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -375,8 +376,8 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
           OperandTypes.STRING_INTEGER_OPTIONAL_INTEGER,
           SqlFunctionCategory.STRING);
 
-  /** The "ENDS_WITH(value1, value2)" function (BigQuery, PostgreSQL). */
-  @LibraryOperator(libraries = {BIG_QUERY, POSTGRESQL})
+  /** The "ENDS_WITH(value1, value2)" function (BigQuery). */
```
julianhyde, line 379: Please amend your commit message with a list of functions added + removed from which libraries. For instance, it's important that you note that after this change the Postgres library no longer contains ENDS_WITH.

There should also be a line "Redshift library is as Postgres, but adds GREATEST, LEAST, ... and removes ARRAY_AGG, ...".

site/_docs/reference.md
```diff
@@ -2650,6 +2650,7 @@ The 'C' (compatibility) column contains value:
 * 'q' for Microsoft SQL Server ('fun=mssql' in the connect string),
 * 'o' for Oracle ('fun=oracle' in the connect string),
 * 'p' for PostgreSQL ('fun=postgresql' in the connect string),
+* 'r' for Amazon RedShift ('fun=redshift' in the connect string),
```
julianhyde, line 2653: I see that you explicitly say when a function is in redshift. That's good.

Does the test check that the table in reference.md is consistent with the functions defined in SqlLibaryOperators?

julianhyde (commented):
Looks good. A few changes needed, but we should get this into 1.38.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
normanj-bitquill: Updated as you suggested.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperatorTableFactory.java
normanj-bitquill: Updated.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: I have updated the commit message to reflect what is added and removed from function libraries. Also mentioned that the Amazon Redshift library extends the PostgreSQL function library.

site/_docs/reference.md
normanj-bitquill: Yes, this does verify that `reference.md` is consistent with the function libraries (including Redshift). I tested removing `r` from the `CHR` function in `reference.md` and there was a failed test as expected.

normanj-bitquill:
@julianhyde I have updated this PR, can you take another look?

PR #3830, opened 2024-06-23
[CALCITE-6445] Add REVERSE function (enabled in Spark library)

YiwenWu:
https://issues.apache.org/jira/browse/CALCITE-6445

Add Function [REVERSE](https://spark.apache.org/docs/latest/api/sql/#reverse) in Spark library.

### DIFF
--- a/babel/src/test/resources/sql/spark.iq
+++ b/babel/src/test/resources/sql/spark.iq
@@ -355,4 +355,97 @@ EXPR$0
 !ok
 
 
+#####################################################################
+# REVERSE
+#
+# REVERSE(str | array)
+# Returns a reversed string or an array with reverse order of elements.
+#
+# Returns STRING or ARRAY
+
+SELECT REVERSE('Spark SQL');
+EXPR$0
+LQS krapS
+!ok
+
+SELECT REVERSE('#c@b!a');
+EXPR$0
+a!b@c#
+!ok
+
+SELECT REVERSE(5);
+EXPR$0
+5
+!ok
+
+SELECT REVERSE(array(2, 1, 4, 3));
+EXPR$0
+[3, 4, 1, 2]
+!ok
+
+SELECT REVERSE(NULL);
+EXPR$0
+null
+!ok
+
+SELECT REVERSE(cast(NULL as tinyint));
+EXPR$0
+null
+!ok
+
+SELECT REVERSE(array(NULL));
+EXPR$0
+[null]
+!ok
+
+SELECT REVERSE(cast(NULL as integer array));
+EXPR$0
+null
+!ok
+
+SELECT REVERSE(array(cast(NULL as tinyint)));
+EXPR$0
+[null]
+!ok
+
+SELECT REVERSE('');
+EXPR$0
+
+!ok
+
+SELECT REVERSE(array(cast(1 as tinyint), 2));
+EXPR$0
+[2, 1]
+!ok
+
+SELECT REVERSE(array(2.1, 2.2));
+EXPR$0
+[2.2, 2.1]
+!ok
+
+SELECT REVERSE(array(2.1, NULL));
+EXPR$0
+[null, 2.1]
+!ok
+
+SELECT REVERSE(array(2, CAST(2.2 AS double)));
+EXPR$0
+[2.2, 2.0]
+!ok
+
+SELECT REVERSE(array(2, CAST(NULL AS double)));
+EXPR$0
+[null, 2.0]
+!ok
+
+SELECT REVERSE(array(CAST(2.1 as decimal(17)), 2.1111111111111119));
+EXPR$0
+[2.1111111111111119, 2.1]
+!ok
+
+SELECT REVERSE(array(CAST(2.1 as double), 2.1111111111111119));
+EXPR$0
+[2.111111111111112, 2.1]
+!ok
+
 # End spark.iq
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2851,6 +2851,7 @@ In the following:
 | b | REGEXP_SUBSTR(string, regexp [, position [, occurrence]]) | Synonym for REGEXP_EXTRACT
 | b m p r s | REPEAT(string, integer)                | Returns a string consisting of *string* repeated of *integer* times; returns an empty string if *integer* is less than 1
 | b m | REVERSE(string)                              | Returns *string* with the order of the characters reversed
+| s | REVERSE(string | array)                        | Returns *string* with the characters in reverse order or array with elements in reverse order
 | b m p r s | RIGHT(string, length)                  | Returns the rightmost *length* characters from the *string*
 | h m s | string1 RLIKE string2                      | Whether *string1* matches regex pattern *string2* (similar to `LIKE`, but uses Java regex)
 | h m s | string1 NOT RLIKE string2                  | Whether *string1* does not match regex pattern *string2* (similar to `NOT LIKE`, but uses Java regex)
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -5502,22 +5502,30 @@ public class SqlOperatorTest {
   }
 
   @Test void testReverseFunc() {
-    final SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.REVERSE);
+    final SqlFunction func = SqlLibraryOperators.REVERSE;
+    final SqlOperatorFixture f0 = fixture().setFor(func);
     f0.checkFails("^reverse('abc')^",
         "No match found for function signature REVERSE\\(<CHARACTER>\\)",
         false);
+    checReverseFunc(f0, func, list(SqlLibrary.BIG_QUERY, SqlLibrary.MYSQL));
+  }
+
+  void checReverseFunc(SqlOperatorFixture f0, SqlFunction function,
+      Iterable<? extends SqlLibrary> libraries) {
+    final String fn = function.getName();
     final Consumer<SqlOperatorFixture> consumer = f -> {
-      f.checkString("reverse('')", "", "VARCHAR(0) NOT NULL");
-      f.checkString("reverse('123')", "321", "VARCHAR(3) NOT NULL");
-      f.checkString("reverse('abc')", "cba", "VARCHAR(3) NOT NULL");
-      f.checkString("reverse('ABC')", "CBA", "VARCHAR(3) NOT NULL");
-      f.checkString("reverse('Hello World')", "dlroW olleH",
+      f.checkString(fn + "('')", "", "VARCHAR(0) NOT NULL");
+      f.checkString(fn + "('123')", "321", "VARCHAR(3) NOT NULL");
+      f.checkString(fn + "('abc')", "cba", "VARCHAR(3) NOT NULL");
+      f.checkString(fn + "('ABC')", "CBA", "VARCHAR(3) NOT NULL");
+      f.checkString(fn + "('Hello World')", "dlroW olleH",
           "VARCHAR(11) NOT NULL");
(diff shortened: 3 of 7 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7464,30 +7471,51 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+          "INTEGER NOT NULL ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[null, 1, cast(2 as tinyint)])", "[2, 1, null]",
+          "INTEGER ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[cast(1 as bigint), 2])", "[2, 1]",
+          "BIGINT NOT NULL ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[null, 1, cast(2 as bigint)])", "[2, 1, null]",
+          "BIGINT ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[cast(1 as decimal), 2])", "[2, 1]",
+          "DECIMAL(19, 0) NOT NULL ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[null, 1, cast(2 as decimal)])", "[2, 1, null]",
```
caicancai, line 7515: I think we should add some double type tests. Spark has its own set of measures for double type processing.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7464,30 +7471,51 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
-        "BIGINT ARRAY NOT NULL");
-    f.checkScalar("array_reverse(array[cast(1 as decimal), 2])", "[2, 1]",
-        "DECIMAL(19, 0) NOT NULL ARRAY NOT NULL");
-    f.checkScalar("array_reverse(array[null, 1, cast(2 as decimal)])", "[2, 1, null]",
-        "DECIMAL(19, 0) ARRAY NOT NULL");
+    checkArrayReverseFunc(f0, func, list(SqlLibrary.BIG_QUERY));
+  }
+
+  /**
+   * Tests {@code REVERSE} function from Spark.
```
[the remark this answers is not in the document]
YiwenWu: done

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
YiwenWu: More test cases have been added in `spark.iq`
image

babel/src/test/resources/sql/spark.iq
```diff
@@ -240,4 +240,47 @@ EXPR$0
 true
 !ok
 
+#####################################################################
+# REVERSE
+#
+# REVERSE(str | array)
```
NobiGo, line 246: This SQL is valid in Spark:
```
spark-sql> SELECT reverse(5);
5
Time taken: 0.071 seconds, Fetched 1 row(s)
```
5 is a number.

NobiGo (changes requested):

babel/src/test/resources/sql/spark.iq
YiwenWu: This support is due to spark type implicit conversion. In addition, Calcite also has a corresponding ability, although the abilities of the two are not exactly the same.

This test case has been added in `spark.iq`.

site/_docs/reference.md
```diff
@@ -2850,6 +2850,7 @@ In the following:
 | b | REGEXP_SUBSTR(string, regexp [, position [, occurrence]]) | Synonym for REGEXP_EXTRACT
 | b m p s | REPEAT(string, integer)                  | Returns a string consisting of *string* repeated of *integer* times; returns an empty string if *integer* is less than 1
 | b m | REVERSE(string)                              | Returns *string* with the order of the characters reversed
+| s | REVERSE( string | array )                      | Returns *string* or *array* with the order of the characters reversed
```
mihaibudiu, line 2853: Returns *string* with the characters in reverse order or *array* with elements in reverse order

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: technically these are decimals, but I don't think this makes a difference

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7464,30 +7471,51 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+          "INTEGER NOT NULL ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[null, 1, cast(2 as tinyint)])", "[2, 1, null]",
+          "INTEGER ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[cast(1 as bigint), 2])", "[2, 1]",
+          "BIGINT NOT NULL ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[null, 1, cast(2 as bigint)])", "[2, 1, null]",
+          "BIGINT ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[cast(1 as decimal), 2])", "[2, 1]",
+          "DECIMAL(19, 0) NOT NULL ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[null, 1, cast(2 as decimal)])", "[2, 1, null]",
```
mihaibudiu, line 7515: can you also check for a NULL array?
this can be done in two ways: a NULL literal, or a NULL literal cast to an array type.

mihaibudiu (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
YiwenWu: "NULL literal cast to an array type." I don’t quite get this.

Add more cases to validate NULL: 
image

site/_docs/reference.md
YiwenWu: Done, updated the description

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: there are many examples in SqlOperatorTest: `f.checkNull("array_remove(cast(null as integer array),`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
YiwenWu: Added more boundary test cases：
image

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
YiwenWu: Added cases in `spark.iq` and `SqlOperatorTest`

site/_docs/reference.md
```diff
@@ -2850,6 +2850,7 @@ In the following:
 | b | REGEXP_SUBSTR(string, regexp [, position [, occurrence]]) | Synonym for REGEXP_EXTRACT
 | b m p s | REPEAT(string, integer)                  | Returns a string consisting of *string* repeated of *integer* times; returns an empty string if *integer* is less than 1
 | b m | REVERSE(string)                              | Returns *string* with the order of the characters reversed
+| s | REVERSE(string | array)                        | Returns *string* with the characters in reverse order or array with elements in reverse order
```
NobiGo, line 2853: Because has support the `SELECT REVERSE(5); `So need to update the doc REVERSE(numeric) ?

NobiGo (changes requested):

site/_docs/reference.md
mihaibudiu: If this is an implicit cast to string, no.

YiwenWu: Unnecessary, types supported  through implicit conversion, not specified separately by default. 
For example, the spark function document does not specifically explain.

caicancai (commented):
Most functions in Spark actually allow mixed use of numeric types, such as int and double, which will eventually be converted to the biggest type. For example
```
scala> val df = spark.sql("SELECT REVERSE(array(2, 1, 4, 3.0))")
df: org.apache.spark.sql.DataFrame = [reverse(array(2, 1, 4, 3.0)): array<decimal(11,1)>]

scala> df.show()
+----------------------------+
|reverse(array(2, 1, 4, 3.0))|
+----------------------------+
|        [3.0, 4.0, 1.0, 2.0]|
+----------------------------+
```
Maybe we should write a method to imitate the operation of spark later.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7464,30 +7471,57 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+          "DECIMAL(19, 0) ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[CAST(2.1 as decimal(17)), 2.1111111111111119])",
+          "[2.1111111111111119, 2.1]", "DECIMAL(19, 2) NOT NULL ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[CAST(2.1 as double), 2.1111111111111119])",
+          "[2.111111111111112, 2.1]", "DOUBLE NOT NULL ARRAY NOT NULL");
+      f.checkScalar(fn + "(array[null])", "[null]", "NULL ARRAY NOT NULL");
+      f.checkNull(fn + "(cast(null as integer array))");
```
caicancai, line 7522: f.checkNull(fn + null)

caicancai: 2024-06-30 23-48-36屏幕截图
 REVERSE(cast(null as integer array)) should fail

YiwenWu: More tests at `spark.iq`

site/_docs/reference.md
NobiGo: Sorry, I don't understand why implicit conversion don't need to log the doc. The doc pay close attention to how to use the func, If this func support , Then we  should illustrate it.

caicancai (approved):
LGTM

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
XuQianJin-Stars: Add a test Case for Reversing a String with Special Characters:
Purpose: To verify the functionality of the string reversal for strings containing special characters.
Input: 'a!b@c#'
Expected Output: '#c@b!a'
This test case aims to ensure that the `REVERSE` function can correctly reverse the order of characters in a string, taking into account any special or non-alphanumeric characters that may be present.

YiwenWu: added.

site/_docs/reference.md
YiwenWu: I do not think so, you can arbitrarily find a func with a string as args, and funcName(integer) is supported.
image

Implicit conversions are supported because `FamilyOperandTypeChecker#checkOperandTypes` is handled:
image

More references: [Design of Calcite Implicit Type Cast](https://docs.google.com/document/d/1g2RUnLXyp_LjUlO-wbblKuP5hqEu3a_2Mt2k4dh6RwU/edit#heading=h.bnu0wb9cokyk)

NobiGo: +1. Thanks for the explaination.

NobiGo (approved):

YiwenWu:
@NobiGo @mihaibudiu Could you please merge this PR?

PR #3832, opened 2024-06-24
[CALCITE-6447] Extract common expressions for disjunctions in Join

frostruan:


xiedeyantu:
Can you fix the failed case?

PR #3833, opened 2024-06-26
[CALCITE-6449] Enable PostgreSQL implementations of to_date/to_timestamp

normanj-bitquill:
* Supports all date/time format patterns of PostreSQL 14
* Oracle will use the existing implementations


mihaibudiu:
@normanj-bitquill I find it very convenient to first run `./gradlew build` on my local machine before submitting a commit.
This catches most of the problems, and is faster than the CI.

normanj-bitquill:
> @normanj-bitquill I find it very convenient to first run `./gradlew build` on my local machine before submitting a commit. This catches most of the problems, and is faster than the CI.

I run that. The errors here are when it is run with the `-PenableCheckerframework` flag. I am running that locally now.

mihaibudiu:
the checkerframework is still not happy
Please leave a note when this is ready for review

normanj-bitquill:
@mihaibudiu Looks like I have resolved the checkerframework issues. Can you take another look?

Thank you for all of the help.

core/src/main/java/org/apache/calcite/util/format/postgresql/DateStringFormatPattern.java
```diff
@@ -40,19 +41,33 @@ public class DateStringFormatPattern<T> extends StringFormatPattern {
...
   private interface DateStringConverter<T> {
+    ChronoUnitEnum getChronoUnit();
+
     T getValueFromDateTime(ZonedDateTime dateTime);
 
+    T[] values();
+
     String getDisplayName(T value, TextStyle textStyle, boolean haveFillMode, Locale locale);
+
+    int getValue(T value);
```
mihaibudiu, line 52: These functions could use some documentation.
What is the integer result?

core/src/main/java/org/apache/calcite/util/format/postgresql/EnumStringFormatPattern.java
```diff
@@ -31,12 +32,27 @@ public class EnumStringFormatPattern extends StringFormatPattern {
...
+    final String inputTrimmed = input.substring(inputPosition.getIndex());
+
+    for (int i = 0; i < enumValues.length; i++) {
+      if (inputTrimmed.startsWith(enumValues[i])) {
+        inputPosition.setIndex(inputPosition.getIndex() + enumValues[i].length());
+        return i;
+      }
+    }
+
+    throw new Exception();
```
mihaibudiu, line 53: this exception needs a message

core/src/main/java/org/apache/calcite/util/format/postgresql/NumberFormatPattern.java
```diff
@@ -27,13 +27,35 @@
...
+  private final ChronoUnitEnum chronoUnit;
+  private final long minValue;
+  private final long maxValue;
+  private final int preferredLength;
   private final Function<ZonedDateTime, String> converter;
+  private final @Nullable Function<Integer, Integer> valueAdjuster;
 
-  protected NumberFormatPattern(Function<ZonedDateTime, String> converter, String... patterns) {
+  public NumberFormatPattern(ChronoUnitEnum chronoUnit, long minValue, long maxValue,
+      int preferredLength, Function<ZonedDateTime, String> converter, String... patterns) {
```
mihaibudiu, line 39: what is preferredLength?

core/src/main/java/org/apache/calcite/util/format/postgresql/NumberFormatPattern.java
```diff
@@ -143,4 +165,40 @@ protected String trimLeadingZeros(String value) {
...
+      }
+    }
+
+    if (endIndex == inputPosition.getIndex()) {
+      throw new Exception();
+    }
+
+    int value = Integer.parseInt(input.substring(inputPosition.getIndex(), endIndex));
+    if (value < minValue || value > maxValue) {
+      throw new Exception();
```
mihaibudiu, line 190: a better error message is needed

core/src/main/java/org/apache/calcite/util/format/postgresql/RomanNumeralMonthFormatPattern.java
```diff
@@ -81,4 +82,49 @@ public RomanNumeralMonthFormatPattern(boolean upperCase, String... patterns) {
...
+      return 12;
+    } else if (inputTrimmed.startsWith(upperCase ? "XI" : "xi")) {
+      inputPosition.setIndex(inputPosition.getIndex() + 2);
+      return 11;
+    } else if (inputTrimmed.startsWith(upperCase ? "X" : "x")) {
+      inputPosition.setIndex(inputPosition.getIndex() + 1);
+      return 10;
+    }
+
+    throw new Exception();
```
mihaibudiu, line 128: again

core/src/main/java/org/apache/calcite/util/format/postgresql/TimeZoneHoursFormatPattern.java
```diff
@@ -29,14 +30,57 @@
...
+      final Locale locale, final boolean haveFillMode, boolean enforceLength) throws Exception {
+
+    int inputOffset = inputPosition.getIndex();
+    String inputTrimmed = input.substring(inputOffset);
+
+    boolean isPositive = true;
+    if (inputTrimmed.charAt(0) == '-') {
+      isPositive = false;
+    } else if (inputTrimmed.charAt(0) != '+') {
+      throw new Exception();
```
mihaibudiu, line 46: again

core/src/main/java/org/apache/calcite/util/format/postgresql/TimeZoneHoursFormatPattern.java
```diff
@@ -29,14 +30,57 @@
...
+    }
+
+    inputOffset++;
+    inputTrimmed = inputTrimmed.substring(inputOffset);
+
+    if (!Character.isDigit(inputTrimmed.charAt(0))) {
+      throw new Exception();
```
mihaibudiu, line 53: I will stop commenting on the exceptions, you should review all of them

core/src/main/java/org/apache/calcite/util/format/postgresql/ChronoUnitEnum.java
```diff
@@ -0,0 +1,232 @@
...
+  YEARS_IN_MILLENIA_ISO_8601(
+      ChronoUnit.YEARS,
+      ISO_8601,
+      CENTURIES),
+  YEARS_IN_CENTURY_ISO_8601(
+      ChronoUnit.YEARS,
+      ISO_8601,
+      CENTURIES),
+  DAYS_IN_YEAR_ISO_8601(
+      ChronoUnit.YEARS,
```
mihaibudiu, line 56: is this unit right?

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
```diff
@@ -38,90 +45,164 @@ public class PostgresqlDateTimeFormatter {
...
+          0,
+          24 * 60 * 60 - 1,
+          5,
           dt -> Integer.toString(dt.get(ChronoField.SECOND_OF_DAY)),
-          "SSSSS", "SSSS"),
+          "SSSSS"),
       new NumberFormatPattern(
+          ChronoUnitEnum.SECONDS_IN_DAY,
+          0,
+          9999,
```
mihaibudiu, line 81: why is this 9999?

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
```diff
@@ -38,90 +45,164 @@ public class PostgresqlDateTimeFormatter {
...
+          ChronoField.AMPM_OF_DAY,
+          "A.M.", "P.M."),
+      new EnumStringFormatPattern(
+          ChronoUnitEnum.HALF_DAYS,
+          ChronoField.AMPM_OF_DAY,
+          "a.m.", "p.m."),
+      new YearWithCommasFormatPattern(),
+      new NumberFormatPattern(
+          ChronoUnitEnum.YEARS,
+          0,
```
mihaibudiu, line 167: is 0 a legal value?

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
```diff
@@ -38,90 +45,164 @@ public class PostgresqlDateTimeFormatter {
...
           dt -> {
             final String yearString =
                 String.format(Locale.ROOT, "%03d", dt.get(IsoFields.WEEK_BASED_YEAR));
             return yearString.substring(yearString.length() - 3);
           },
           "IYY"),
       new NumberFormatPattern(
+          ChronoUnitEnum.YEARS_IN_CENTURY_ISO_8601,
+          0,
+          Integer.MAX_VALUE,
```
mihaibudiu, line 193: why not 100?

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
```diff
@@ -38,90 +45,164 @@ public class PostgresqlDateTimeFormatter {
...
           dt -> {
             final String yearString =
                 String.format(Locale.ROOT, "%02d", dt.get(IsoFields.WEEK_BASED_YEAR));
             return yearString.substring(yearString.length() - 2);
           },
           "IY"),
       new NumberFormatPattern(
+          ChronoUnitEnum.YEARS_IN_MILLENIA,
+          0,
+          Integer.MAX_VALUE,
```
mihaibudiu, line 204: why not 1000?

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
```diff
@@ -132,6 +213,10 @@ public class PostgresqlDateTimeFormatter {
           },
           "YYY"),
       new NumberFormatPattern(
+          ChronoUnitEnum.YEARS_IN_CENTURY,
+          0,
+          Integer.MAX_VALUE,
```
mihaibudiu, line 218: 100?

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
```diff
@@ -142,6 +227,10 @@ public class PostgresqlDateTimeFormatter {
           },
           "YY"),
       new NumberFormatPattern(
+          ChronoUnitEnum.YEARS_IN_CENTURY,
+          0,
+          Integer.MAX_VALUE,
```
mihaibudiu, line 232: 1000?

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
```diff
@@ -175,25 +272,37 @@ public class PostgresqlDateTimeFormatter {
...
+          ChronoUnitEnum.DAYS_IN_WEEK,
+          1,
+          7,
+          1,
           dt -> Integer.toString(dt.getDayOfWeek().getValue()),
           "ID"),
       new NumberFormatPattern(
+          ChronoUnitEnum.YEARS_IN_CENTURY_ISO_8601,
+          0,
+          9,
```
mihaibudiu, line 284: these look strange

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
```diff
@@ -308,4 +481,362 @@ public static String toChar(String formatString, ZonedDateTime dateTime) {
...
+    }
+
+    if (dateParts.containsKey(ChronoUnitEnum.DAYS_IN_MONTH)) {
+      updatedDateTime =
+          updatedDateTime.withDayOfMonth(dateParts.get(ChronoUnitEnum.DAYS_IN_MONTH).intValue());
+    }
+
+    if (dateParts.containsKey(ChronoUnitEnum.WEEKS_IN_MONTH)) {
+      updatedDateTime =
+          updatedDateTime.withDayOfMonth(
```
mihaibudiu, line 638: this doesn't look right

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
```diff
@@ -308,4 +481,362 @@ public static String toChar(String formatString, ZonedDateTime dateTime) {
...
+              dateParts.get(ChronoUnitEnum.WEEKS_IN_MONTH).intValue() * 7 - 6);
+    }
+
+    if (dateParts.containsKey(ChronoUnitEnum.WEEKS_IN_YEAR)) {
+      updatedDateTime =
+          updatedDateTime.withDayOfYear(
```
mihaibudiu, line 644: this doesn't look right either

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
```diff
@@ -308,4 +481,362 @@ public static String toChar(String formatString, ZonedDateTime dateTime) {
...
+
+      return yearSign * year;
+    }
+
+    if (years != null) {
+      return yearSign * years.intValue();
+    }
+
+    if (yearsInCentury != null) {
+      int year = yearsInCentury.intValue();
```
mihaibudiu, line 744: is this part of the postgres spec?

core/src/test/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatterTest.java
```diff
@@ -1311,6 +1314,610 @@ void testEraLowerCaseWithDots(String pattern) {
...
+        createDateTime(2001, 1, 1, 0, 0, 0, 0),
+        PostgresqlDateTimeFormatter.toTimestamp("1", "YY", TIME_ZONE));
+    assertEquals(
+        createDateTime(2024, 1, 1, 0, 0, 0, 0),
+        PostgresqlDateTimeFormatter.toTimestamp("24", "YY", TIME_ZONE));
+  }
+
+  @Test void testToTimestampY() throws Exception {
+    assertEquals(
+        createDateTime(2001, 1, 1, 0, 0, 0, 0),
```
mihaibudiu, line 1364: do you want to make this and other frequently used dates into constants?

core/src/test/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatterTest.java
```diff
@@ -1311,6 +1314,610 @@ void testEraLowerCaseWithDots(String pattern) {
...
+        PostgresqlDateTimeFormatter.toTimestamp("1", "I", TIME_ZONE));
+    assertEquals(
+        createDateTime(2004, 1, 1, 0, 0, 0, 0),
+        PostgresqlDateTimeFormatter.toTimestamp("4", "I", TIME_ZONE));
+  }
+
+  @Test void testToTimestampBCAD() throws Exception {
+    assertEquals(
+        0,
+        PostgresqlDateTimeFormatter.toTimestamp("1920BC", "YYYYBC", TIME_ZONE)
```
mihaibudiu, line 1425: the rest of Calcite has trouble with BC dates, but that's no reason not to make this work

mihaibudiu (commented):
Will you also add one of these nice .iq scripts that has been compared with Postgres?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4004,6 +4004,19 @@ public static long unixDateExtract(String rangeString, long date) {
    * {@code FORMAT_DATETIME}, {@code FORMAT_TIME}, {@code TO_CHAR} functions. */
   @Deterministic
   public static class DateFormatFunction {
+    // Timezone to use for PostgreSQL parsing of timestamps
+    private static final ZoneId LOCAL_ZONE;
+    static {
+      ZoneId zoneId;
+      try {
+        // Currently the parsed timestamps are expected to be in UTC
```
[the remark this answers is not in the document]
normanj-bitquill: The value returned from the existing `to_timestamp` is expected to be the millis from epoch in UTC.
https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java#L4065

At some point in the future, this could be changed to support timezones, since you could do something like:
```
to_timestamp('2024-06-03 15:32:46+03:30')
```

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4061,11 +4074,40 @@ public int toDate(String dateString, String fmtString) {
...
+    }
+
     public long toTimestamp(String timestampString, String fmtString) {
       return toLong(
           new java.sql.Timestamp(internalToDateTime(timestampString, fmtString)));
     }
 
+    public long toTimestampPg(String timestampString, String fmtString) {
+      try {
+        return PostgresqlDateTimeFormatter.toTimestamp(timestampString, fmtString, LOCAL_ZONE)
```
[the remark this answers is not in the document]
normanj-bitquill: Matching the behaviour of the existing `to_timestamp` method. It will call `toLong()`
https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java#L3743

which will always use the TimeZone `LOCAL_TZ`.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
mihaibudiu: Please add this information as comments in the code.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
mihaibudiu: Is this what Postgres expects?

core/src/main/java/org/apache/calcite/util/format/postgresql/NumberFormatPattern.java
normanj-bitquill: Some format patterns such as `YYYY` are usually 4 characters long, but can be shorter. `YYYY` can also be used to parse a year value of `56`. `preferredLength` is the expected number of characters that the pattern consume, but it may actually consume fewer.

Consider this:
```
to_date('56-03-15', 'YYYY-MM-DD')
```

mihaibudiu: I think this would be a very useful comment in the code as well.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
normanj-bitquill: PostgreSQL will parse the timestamp in the default timezone unless the format and input have timezone information.

From my understanding, I need to eventually return a long value (millis since epoch in UTC) for a timestamp value. Is it possible to return a timestamp with timezone information?
https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java#L4064

Since I am returning a long value (millis since epoch in UTC), I am using UTC as the default timezone.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1690,22 +1690,40 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
           ReturnTypes.TIMESTAMP_NULLABLE,
           OperandTypes.STRING_STRING,
           SqlFunctionCategory.TIMEDATE);
 
+  /** The "TO_TIMESTAMP(string1, string2)" function for PostgreSQL; casts string1
+   * to a TIMESTAMP using the format specified in string2. */
+  @LibraryOperator(libraries = {POSTGRESQL})
+  public static final SqlFunction TO_TIMESTAMP_PG =
+      new SqlBasicFunction("TO_TIMESTAMP", SqlKind.OTHER_FUNCTION,
+          SqlSyntax.FUNCTION, true, ReturnTypes.TIMESTAMP_NULLABLE, null,
```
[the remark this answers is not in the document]
normanj-bitquill: You are correct. At this point in the code, am I able to return a timestamp with timezone?
https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java#L4064

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
normanj-bitquill: Yeah, `YY`. You could do something like `CCYY` if you wanted to enter this year as `2124`.

```
postgres=# select to_date('2124', 'CCYY');
  to_date
------------
 2024-01-01
(1 row)
```

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
normanj-bitquill: This is matching PostgreSQL behaviour.

```
postgres=# select to_date('2024-03-2', 'YYYY-MM-W');
  to_date
------------
 2024-03-08
(1 row)
```

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
normanj-bitquill: This is also to match PostgreSQL behaviour.

```
postgres=# select to_date('2024-15', 'YYYY-WW');
  to_date
------------
 2024-04-08
(1 row)

postgres=# select to_date('2024-99', 'YYYY-DDD');
  to_date
------------
 2024-04-08
(1 row)
```

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
mihaibudiu: I have added recently support for timestamps with time zone

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
normanj-bitquill: Improved the comment here.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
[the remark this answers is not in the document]
normanj-bitquill: I don't see how to return a timestamp with a timezone. This may not be an issue though. If the timestamp format does not have timezone patterns, then it will be assumed to be `LOCAL_ZONE` (UTC). If the pattern has timezone patterns, then it is parsed using the timezone from the input string. After parsing, it is converted to `LOCAL_ZONE` (UTC).

As an example, if you have the input `2024-04-17 00:00:00-07:00` and pattern `YYYY-MM-DD HH24:MI:SSTZH:TZM`, you would end up with `2024-04-17 07:00:00`.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5131,6 +5164,41 @@ void testBitGetFunc(SqlOperatorFixture f, String functionName) {
     f.checkNull("to_timestamp(NULL, NULL)");
   }
 
+  @Test void testToTimestampPg() {
+    final SqlOperatorFixture f = fixture().withLibrary(SqlLibrary.POSTGRESQL);
+    f.setFor(SqlLibraryOperators.TO_TIMESTAMP_PG);
+
+    f.checkString("to_timestamp('2022-06-03 18:34:56', 'YYYY-MM-DD HH24:MI:SS')",
```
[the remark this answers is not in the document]
normanj-bitquill: Apparently milliseconds will not work (at least for a test). You can successfully evaluate the expression

```
to_timestamp('2022-06-03 18:34:56.965', 'YYYY-MM-DD HH24:MI:SS.MS')
```

The method `toTimestampPg()` will return a long value with the milliseconds. When that is later converted to a string, the milliseconds are lost.

core/src/main/java/org/apache/calcite/util/format/postgresql/DateStringFormatPattern.java
normanj-bitquill: Added more Javadoc comments, including here. Hopefully this is explained better now.

core/src/main/java/org/apache/calcite/util/format/postgresql/EnumStringFormatPattern.java
normanj-bitquill: Changed to a ParseException and added a message.

core/src/main/java/org/apache/calcite/util/format/postgresql/NumberFormatPattern.java
normanj-bitquill: Added more Javadoc comments and tried to explain this better in the Javadoc comment.

core/src/main/java/org/apache/calcite/util/format/postgresql/NumberFormatPattern.java
normanj-bitquill: Changed to a ParseException and added a message.

core/src/main/java/org/apache/calcite/util/format/postgresql/RomanNumeralMonthFormatPattern.java
normanj-bitquill: Changed to a ParseException and added a message.

core/src/main/java/org/apache/calcite/util/format/postgresql/TimeZoneHoursFormatPattern.java
normanj-bitquill: Changed to a ParseException and added a message.

core/src/main/java/org/apache/calcite/util/format/postgresql/ChronoUnitEnum.java
normanj-bitquill: It was not, good catch. This is corrected.

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
normanj-bitquill: A mistake on my part. This should be parsed the same as `SSSSS`. Updated.

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
normanj-bitquill: Turns out that PostgreSQL does not limit this. It will accept 4 digit numbers.

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
normanj-bitquill: Turns out that PostgreSQL does not limit this. It will accept 4 digit numbers.

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
normanj-bitquill: Turns out that PostgreSQL does not limit this. It will accept 4 digit numbers.

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
normanj-bitquill: Turns out that PostgreSQL does not limit this. It will accept 4 digit numbers.

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
normanj-bitquill: A mistake on my part. PostgreSQL will accept any number, even a 4 digit number.

core/src/test/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatterTest.java
normanj-bitquill: Added some constants and used those in most places.

core/src/main/java/org/apache/calcite/util/format/postgresql/PostgresqlDateTimeFormatter.java
normanj-bitquill: PostgreSQL will accept `0` for year and silently convert it to 1 CE. I'll see if I have the behaviour right for this.

normanj-bitquill: Fixed up the behaviour here and added a test.

normanj-bitquill:
> Will you also add one of these nice .iq scripts that has been compared with Postgres?

@mihaibudiu I have added many tests for `to_date` and `to_timestamp` to the `postgresql.iq` file.

core/src/main/java/org/apache/calcite/util/format/postgresql/FormatPattern.java
```diff
@@ -36,6 +59,152 @@ public interface FormatPattern {
...
+  /**
+   * Get the ChronoUnitEnum value that this format pattern represents. For example, the
+   * pattern YYYY is for YEAR.
+   *
+   * @return a ChronoUnitEnum value
+   */
+  protected abstract ChronoUnitEnum getChronoUnit();
+
+  /**
+   * Attempts to parse a single value from the input for this pattern. It will start parsing
```
mihaibudiu, line 74: this is much better

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/util/format/postgresql/FormatPattern.java
normanj-bitquill: @mihaibudiu I see what you mean now. I have updated the return type and updated any tests that were affected.

mihaibudiu:
I think that you can squash the commits. I don't know if you want more than 1 - up to you.

normanj-bitquill:

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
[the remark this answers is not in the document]
normanj-bitquill: This has been addressed in this PR. The return type is now `ReturnTypes.TIMESTAMP_TZ_NULLABLE`.

PR #3835, opened 2024-06-27
[CALCITE-6435] SqlToRel conversion of IN expressions may lead to incorrect simplifications

kgyrtkirk:
Conversion path for comparisions generated from IN expressions was handling types differently. This may have lead to some over-simplification in some cases.

Altered the conversion to do the full SqlToRex conversion steps for these generated nodes as well. Added an extra safeguard check to RexSimplify to prevent the bug from being triggered.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexSimplify.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexSimplify.java
@@ -1682,7 +1682,10 @@ public class RexSimplify {
             final RexLiteral literal = comparison.literal;
             final RexLiteral prevLiteral =
                 equalityConstantTerms.put(comparison.ref, literal);
-            if (prevLiteral != null && !literal.equals(prevLiteral)) {
+
+            if (prevLiteral != null
+                && literal.getType().equals(prevLiteral.getType())
+                && !literal.equals(prevLiteral)) {
               return rexBuilder.makeLiteral(false);
             }
           } else if (RexUtil.isReferenceOrAccess(left, true)
@@ -1753,7 +1756,7 @@ public class RexSimplify {
         if (literal2 == null) {
           continue;
         }
-        if (!literal1.equals(literal2)) {
+        if (literal1.getType().equals(literal2.getType()) && !literal1.equals(literal2)) {
           // If an expression is equal to two different constants,
           // it is not satisfiable
           return rexBuilder.makeLiteral(false);
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -175,6 +175,8 @@ import org.apache.calcite.sql.validate.SqlValidatorScope;
 import org.apache.calcite.sql.validate.SqlValidatorTable;
 import org.apache.calcite.sql.validate.SqlValidatorUtil;
 import org.apache.calcite.sql.validate.SqlWithItemTableRef;
+import org.apache.calcite.sql2rel.SqlToRelConverter.Blackboard;
+import org.apache.calcite.sql2rel.SqlToRelConverter.SqlIdentifierFinder;
 import org.apache.calcite.tools.RelBuilder;
 import org.apache.calcite.tools.RelBuilderFactory;
 import org.apache.calcite.util.ImmutableBitSet;
@@ -1184,16 +1186,16 @@ public class SqlToRelConverter {
       }
       final SqlNode leftKeyNode = call.operand(0);
 
-      final List<RexNode> leftKeys;
+      final List<SqlNode> leftSqlKeys;
       switch (leftKeyNode.getKind()) {
       case ROW:
-        leftKeys = new ArrayList<>();
+        leftSqlKeys = new ArrayList<>();
         for (SqlNode sqlExpr : ((SqlBasicCall) leftKeyNode).getOperandList()) {
-          leftKeys.add(bb.convertExpression(sqlExpr));
+          leftSqlKeys.add(sqlExpr);
         }
         break;
       default:
-        leftKeys = ImmutableList.of(bb.convertExpression(leftKeyNode));
+        leftSqlKeys = ImmutableList.of(leftKeyNode);
       }
 
       if (query instanceof SqlNodeList) {
@@ -1204,7 +1206,7 @@ public class SqlToRelConverter {
           subQuery.expr =
               convertInToOr(
                   bb,
-                  leftKeys,
+                  leftSqlKeys,
                   valueList,
                   (SqlInOperator) call.getOperator());
           return;
@@ -1214,6 +1216,10 @@ public class SqlToRelConverter {
         // values list into an inline table for the
         // reference to Q below.
       }
+
+      final List<RexNode> leftKeys = leftSqlKeys.stream()
+          .map(bb::convertExpression)
+          .collect(toImmutableList());
 
       // Project out the search columns from the left side
 
@@ -1718,12 +1724,11 @@ public class SqlToRelConverter {
    */
   private @Nullable RexNode convertInToOr(
       final Blackboard bb,
-      final List<RexNode> leftKeys,
+      final List<SqlNode> leftKeys,
       SqlNodeList valuesList,
       SqlInOperator op) {
     final List<RexNode> comparisons = new ArrayList<>();
     for (SqlNode rightVals : valuesList) {
-      RexNode rexComparison;
       final SqlOperator comparisonOp;
       if (op instanceof SqlQuantifyOperator) {
         comparisonOp =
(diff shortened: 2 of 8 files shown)


core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -1682,7 +1682,10 @@ private <C extends Comparable<C>> RexNode simplifyAnd2ForUnknownAsFalse(
             final RexLiteral literal = comparison.literal;
             final RexLiteral prevLiteral =
                 equalityConstantTerms.put(comparison.ref, literal);
-            if (prevLiteral != null && !literal.equals(prevLiteral)) {
+
+            if (prevLiteral != null
+                && literal.getType().equals(prevLiteral.getType())
+                && !literal.equals(prevLiteral)) {
```
mihaibudiu, line 1688: so two literals can be equal but their types can be different?
one could argue that this is a bug in SqlLiteral::equals.

druid/src/test/java/org/apache/calcite/test/DruidAdapter2IT.java
```diff
@@ -1030,7 +1030,7 @@ private void checkGroupBySingleSortLimit(boolean approx) {
         + "intervals=[[1900-01-09T00:00:00.000Z/2992-01-10T00:00:00.000Z]], "
         + "filter=[AND("
         + "=($3, 'High Top Dried Mushrooms'), "
-        + "SEARCH($87, Sarg['Q2', 'Q3']:CHAR(2)), "
+        + "SEARCH($87, Sarg['Q2':VARCHAR, 'Q3':VARCHAR]:VARCHAR), "
```
mihaibudiu, line 1033: I find this change surprising.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -1711,19 +1717,18 @@ public RelNode convertToSingleValueSubq(
   /**
    * Converts "x IN (1, 2, ...)" to "x=1 OR x=2 OR ...".
    *
-   * @param leftKeys   LHS
+   * @param leftSqlKeys   LHS
```
mihaibudiu, line 1720: does this parameter need to be renamed?

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
kgyrtkirk: not necessarily - I've undone the rename

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
kgyrtkirk: the situation was caused by two literals which had the same value; but different types (`VARCHAR`; `CHAR(n)`)

actually the predicate based simplification also does this (beyond other things); but there are situations in which its disabled.

druid/src/test/java/org/apache/calcite/test/DruidAdapter2IT.java
kgyrtkirk: I was surprised too ; however this is the same kind of expressions `SqlToRel` would produce if the input would have been a [set of comparisison](https://github.com/apache/calcite/pull/3835/files#diff-26cbb70deb4731f2c84faf803724dc40a9cdf6e2f2c67f8724e1afbc98fc8950R3226-R3248) ; so the `IN` rewrite now [matches that](https://github.com/apache/calcite/pull/3835/files#diff-26cbb70deb4731f2c84faf803724dc40a9cdf6e2f2c67f8724e1afbc98fc8950R3221)

mihaibudiu:
Let's wait a bit to see whether anyone else has comments.

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
YiwenWu: I think `VARCHAR` and `CHAR` should be considered equivalent if RexLiteral.value is the same

YiwenWu:
I think it can be simpler to modify the judgment of `RexSimplify#simplifyAnd2ForUnknownAsFalse`?
```
if (prevLiteral != null && !literal.equalsWithTypeName(prevLiteral)) {
     return rexBuilder.makeLiteral(false);
 }
```
where `RexLiteral#equalsWithTypeName` compares to `TypeName` instead of `Type`. In this case, varchar `Sebastian` and char `Sebastian` can be equivalent.

kgyrtkirk:
> I think it can be simpler to modify the judgment of `RexSimplify#simplifyAnd2ForUnknownAsFalse`?
> ```
> if (prevLiteral != null && !literal.equalsWithTypeName(prevLiteral)) {
>      return rexBuilder.makeLiteral(false);
>  }
> ```
> where `RexLiteral#equalsWithTypeName` compares to `TypeName` instead of `Type` in `RexLiteral#equals` . In this case, varchar `Sebastian` and char `Sebastian` can be equivalent.
> 

I don't see any such method in
[RexLiteral](https://github.com/apache/calcite/blob/d74b283f56a0043202af8ad95080ca21c70b8baa/core/src/main/java/org/apache/calcite/rex/RexLiteral.java) did I missed it?

I would be kinda against of the introduction of such method because it would place custom type comparision logic inside a Rex implementation...

I might have not mentioned this here...but 
weakening these in RexSimplify has no adverse effects because predicate based simplification can do this too(and more) ; I think these map based things can even be removed....

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
mihaibudiu: CHAR(N) indicates that the literal has to be padded with spaces, whereas VARCHAR indicates that trailing spaces are to be ignored. So two literals with identical strings but different types can have different values.

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -1753,7 +1756,7 @@ private <C extends Comparable<C>> RexNode simplifyAnd2ForUnknownAsFalse(
         if (literal2 == null) {
           continue;
         }
-        if (!literal1.equals(literal2)) {
+        if (literal1.getType().equals(literal2.getType()) && !literal1.equals(literal2)) {
```
YiwenWu, line 1759: I'm puzzled about the necessity of adding extra type equals checks. What specific issues might arise without them in SQL simplification?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -1732,25 +1737,23 @@ public RelNode convertToSingleValueSubq(
...
             && call.operandCount() == leftKeys.size();
         rexComparison =
-            RexUtil.composeConjunction(rexBuilder,
-                Util.transform(
-                    Pair.zip(leftKeys, call.getOperandList()),
-                    pair -> rexBuilder.makeCall(comparisonOp, pair.left,
-                        // TODO: remove requireNonNull when checkerframework issue resolved
-                        ensureSqlType(requireNonNull(pair.left, "pair.left").getType(),
-                            bb.convertExpression(pair.right)))));
+            RexUtil.composeConjunction(
```
YiwenWu, line 1751: Is the focus solely on the SQL IN case, or might there be a need for adjustments elsewhere too?

YiwenWu:
> > I think it can be simpler to modify the judgment of `RexSimplify#simplifyAnd2ForUnknownAsFalse`?
> > ```
> > if (prevLiteral != null && !literal.equalsWithTypeName(prevLiteral)) {
> >      return rexBuilder.makeLiteral(false);
> >  }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > where `RexLiteral#equalsWithTypeName` compares to `TypeName` instead of `Type` in `RexLiteral#equals` . In this case, varchar `Sebastian` and char `Sebastian` can be equivalent.
> 
> I don't see any such method in [RexLiteral](https://github.com/apache/calcite/blob/d74b283f56a0043202af8ad95080ca21c70b8baa/core/src/main/java/org/apache/calcite/rex/RexLiteral.java) did I missed it?
> 
> I would be kinda against of the introduction of such method because it would place custom type comparision logic inside a Rex implementation...
> 
> I might have not mentioned this here...but weakening these in RexSimplify has no adverse effects because predicate based simplification can do this too(and more) ; I think these map based things can even be removed....



> > I think it can be simpler to modify the judgment of `RexSimplify#simplifyAnd2ForUnknownAsFalse`?
> > ```
> > if (prevLiteral != null && !literal.equalsWithTypeName(prevLiteral)) {
> >      return rexBuilder.makeLiteral(false);
> >  }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > where `RexLiteral#equalsWithTypeName` compares to `TypeName` instead of `Type` in `RexLiteral#equals` . In this case, varchar `Sebastian` and char `Sebastian` can be equivalent.
> 
> I don't see any such method in [RexLiteral](https://github.com/apache/calcite/blob/d74b283f56a0043202af8ad95080ca21c70b8baa/core/src/main/java/org/apache/calcite/rex/RexLiteral.java) did I missed it?
> 
> I would be kinda against of the introduction of such method because it would place custom type comparision logic inside a Rex implementation...
> 
> I might have not mentioned this here...but weakening these in RexSimplify has no adverse effects because predicate based simplification can do this too(and more) ; I think these map based things can even be removed....

Currently, there is no `RexLiteral#equalsWithTypeName` method.

It seems two distinct approaches. This PR handles `CHAR` to `VARCHAR` conversion early in `SqlToRelConverter` , enabling direct conjunction in the simplification stage.  Alternatively, refine the `simplifyAnd2ForUnknownAsFalse` method to bypass false judgments for `RexLiteral` with identical value and typeName(`SqlTypeName`).

There is no problem based on the approach of this PR. Only two small points that need to be confirmed: 
1. Addressing range concerns, the `CHAR` to `VARCHAR` conversion is specific to IN conditions?
2. Is an update to `simplifyAnd2ForUnknownAsFalse` necessary, what scenarios would benefit from this change?

kgyrtkirk:
> 1.   Addressing range concerns, the CHAR to VARCHAR conversion is specific to IN conditions?

the issue arised from the fact that `SqlToRel` had a different route to produce `=` checks when `IN` was rewritten to a set of OR conditions; this PR only makes that extra route go away - and shows the converter the full `=` expression to do the same conversion as it would have done in case it would have been part of the original query

> 2. Is an update to simplifyAnd2ForUnknownAsFalse necessary, what scenarios would benefit from this change?

not mandatory for the current bug ; as `1` solves that ;  but I would rather restrict the cases in which `RexSimplify` could over-simplify things; as its a quite frequently called part.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
kgyrtkirk: not really; this `conjunction` is there for complex expressions; which will be there for `NOT IN` as well:
```
(a,b) NOT IN ((1,2), (3,4))
NOT ((a=1 && b=2) || (a=3 && b=4) )
```

caicancai (approved):
LGTM

PR #3838, opened 2024-07-01
Add testSortRemoveConstantKeyWhenOrderByIsNull on RelOptRulesTest

caicancai:
This test is to show that SortRemoveConstantKeys can optimize Order By null

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -1388,6 +1388,17 @@ class RelOptRulesTest extends RelOptTestBase {
   }
 
   /** Test case for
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):

julianhyde:
The method needs a javadoc comment.

Why does the method name use the word 'empty'? Nothing is empty in this case.

caicancai:
> The method needs a javadoc comment.
> 
> Why does the method name use the word 'empty'? Nothing is empty in this case.

I modified the appropriate method name and added comments and associated the previous jira case, thank you

YiwenWu:
What is the reason for adding this test, is there a jira link?

caicancai:
> What is the reason for adding this test, is there a jira link?

https://issues.apache.org/jira/browse/CALCITE-6440

YiwenWu (approved):
LGTM

PR #3839, opened 2024-07-01
[CALCITE-6312] Add LOG function (enabled in PostgreSQL library)

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6312
https://issues.apache.org/jira/browse/CALCITE-6456

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -65,6 +65,7 @@ import org.apache.calcite.sql.SqlWindowTableFunction;
 import org.apache.calcite.sql.fun.SqlItemOperator;
 import org.apache.calcite.sql.fun.SqlJsonArrayAggAggFunction;
 import org.apache.calcite.sql.fun.SqlJsonObjectAggAggFunction;
+import org.apache.calcite.sql.fun.SqlLibrary;
 import org.apache.calcite.sql.fun.SqlQuantifyOperator;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.fun.SqlTrimFunction;
@@ -75,6 +76,7 @@ import org.apache.calcite.sql.validate.SqlUserDefinedFunction;
 import org.apache.calcite.sql.validate.SqlUserDefinedTableFunction;
 import org.apache.calcite.sql.validate.SqlUserDefinedTableMacro;
 import org.apache.calcite.util.BuiltInMethod;
+import org.apache.calcite.util.Pair;
 import org.apache.calcite.util.Util;
 
 import com.google.common.collect.ImmutableList;
@@ -216,6 +218,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG2;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_AND;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOGICAL_OR;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG_MYSQL;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG_POSTGRES;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LPAD;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.MAP_CONCAT;
@@ -654,12 +657,13 @@ public class RexImpTable {
       defineMethod(POWER_PG, BuiltInMethod.POWER_PG.method, NullPolicy.STRICT);
       defineMethod(ABS, BuiltInMethod.ABS.method, NullPolicy.STRICT);
 
-      map.put(LN, new LogImplementor());
-      map.put(LOG, new LogImplementor());
-      map.put(LOG10, new LogImplementor());
+      map.put(LN, new LogImplementor(SqlLibrary.BIG_QUERY));
+      map.put(LOG, new LogImplementor(SqlLibrary.BIG_QUERY));
+      map.put(LOG10, new LogImplementor(SqlLibrary.BIG_QUERY));
 
-      map.put(LOG_MYSQL, new LogMysqlImplementor());
-      map.put(LOG2, new LogMysqlImplementor());
+      map.put(LOG_POSTGRES, new LogImplementor(SqlLibrary.POSTGRESQL));
+      map.put(LOG_MYSQL, new LogImplementor(SqlLibrary.MYSQL));
+      map.put(LOG2, new LogImplementor(SqlLibrary.MYSQL));
 
       defineReflective(RAND, BuiltInMethod.RAND.method,
           BuiltInMethod.RAND_SEED.method);
@@ -4202,67 +4206,57 @@ public class RexImpTable {
    * appropriate base (i.e. base e for LN).
    */
   private static class LogImplementor extends AbstractRexCallImplementor {
-    LogImplementor() {
+    private final SqlLibrary library;
+    LogImplementor(SqlLibrary library) {
       super("log", NullPolicy.STRICT, true);
+      this.library = library;
     }
 
     @Override Expression implementSafe(final RexToLixTranslator translator,
         final RexCall call, final List<Expression> argValueList) {
-      return Expressions.call(BuiltInMethod.LOG.method, args(call, argValueList));
+      return Expressions.
+          call(BuiltInMethod.LOG.method, args(call, argValueList, library));
     }
 
+    /**
+     * This method is used to handle the implementation of different log functions.
+     * It generates the corresponding expression list based on the input function name
+     * and argument list.
+     *
+     * @param call The RexCall that contains the function call information.
+     * @param argValueList The list of argument expressions.
+     * @param library The SQL library that the function belongs to.
+     * @return A list of expressions that represents the implementation of the log function.
+     */
     private static List<Expression> args(RexCall call,
-        List<Expression> argValueList) {
-      Expression operand0 = argValueList.get(0);
-      final Expressions.FluentList<Expression> list = Expressions.list(operand0);
-      switch (call.getOperator().getName()) {
-      case "LOG":
-        if (argValueList.size() == 2) {
-          return list.append(argValueList.get(1)).append(Expressions.constant(0));
-        }
-        // fall through
-      case "LN":
-        return list.append(Expressions.constant(Math.exp(1))).append(Expressions.constant(0));
-      case "LOG10":
-        return list.append(Expressions.constant(BigDecimal.TEN)).append(Expressions.constant(0));
-      default:
-        throw new AssertionError("Operator not found: " + call.getOperator());
+        List<Expression> argValueList, SqlLibrary library) {
+      Pair<Expression, Expression> operands;
+      Expression operand0;
+      Expression operand1;
+      if (argValueList.size() == 1) {
+        operands = library == SqlLibrary.POSTGRESQL
+            ? Pair.of(argValueList.get(0), Expressions.constant(BigDecimal.TEN))
+            : Pair.of(argValueList.get(0), Expressions.constant(Math.exp(1)));
+      } else {
+        operands = library == SqlLibrary.BIG_QUERY
+            ? Pair.of(argValueList.get(0), argValueList.get(1))
+            : Pair.of(argValueList.get(1), argValueList.get(0));
       }
-    }
-  }
-
-  /** Implementor for the {@code LN}, {@code LOG}, {@code LOG2} and {@code LOG10} operators
-   *  on Mysql and Spark library
-   *
-   * <p>Handles all logarithm functions using log rules to determine the
-   * appropriate base (i.e. base e for LN).
-   */
-  private static class LogMysqlImplementor extends AbstractRexCallImplementor {
-    LogMysqlImplementor() {
-      super("log", NullPolicy.STRICT, true);
-    }
-
-    @Override Expression implementSafe(final RexToLixTranslator translator,
-        final RexCall call, final List<Expression> argValueList) {
-      return Expressions.call(BuiltInMethod.LOG.method, args(call, argValueList));
-    }
-
-    private static List<Expression> args(RexCall call,
-        List<Expression> argValueList) {
-      Expression operand0 = argValueList.get(0);
+      operand0 = operands.left;
+      operand1 = operands.right;
+      boolean nonPositiveIsNull = library == SqlLibrary.MYSQL ? true : false;
       final Expressions.FluentList<Expression> list = Expressions.list(operand0);
       switch (call.getOperator().getName()) {
       case "LOG":
-        if (argValueList.size() == 2) {
-          return list.append(argValueList.get(1)).append(Expressions.constant(1));
-        }
-        // fall through
+        return list.append(operand1).append(Expressions.constant(nonPositiveIsNull));
       case "LN":
-        return list.append(Expressions.constant(Math.exp(1))).append(Expressions.constant(1));
+        return list.append(Expressions.constant(Math.exp(1)))
+            .append(Expressions.constant(nonPositiveIsNull));
       case "LOG2":
-        return list.append(Expressions.constant(2)).append(Expressions.constant(1));
+        return list.append(Expressions.constant(2)).append(Expressions.constant(nonPositiveIsNull));
       case "LOG10":
-        return list.append(Expressions.constant(BigDecimal.TEN)).append(Expressions.constant(1));
+        return list.append(Expressions.constant(BigDecimal.TEN))
+            .append(Expressions.constant(nonPositiveIsNull));
       default:
         throw new AssertionError("Operator not found: " + call.getOperator());
       }
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -2785,41 +2785,66 @@ public class SqlFunctions {
     return Math.pow(b0.doubleValue(), b1.doubleValue());
   }
 
-
   // LN, LOG, LOG10, LOG2
 
-  /** SQL {@code LOG(number, number2)} function applied to double values. */
-  public static @Nullable Double log(double number, double number2, int nullFlag) {
-    if (nullFlag == 1 && number <= 0) {
+  /**
+   * SQL {@code LOG(number, base)} function applied to double values.
+   *
+   * @param nonPositiveIsNull if true return null for non-positive values
+   */
+  public static @Nullable Double log(double number, double base, boolean nonPositiveIsNull) {
+    if (nonPositiveIsNull && number <= 0) {
       return null;
     }
-    return Math.log(number) / Math.log(number2);
+    if (number <= 0 || base <= 0) {
+      throw new IllegalArgumentException("Cannot take logarithm of zero or negative number");
+    }
+    return Math.log(number) / Math.log(base);
   }
 
-  /** SQL {@code LOG(number, number2)} function applied to
-   * double and BigDecimal values. */
-  public static @Nullable Double log(double number, BigDecimal number2, int nullFlag) {
-    if (nullFlag == 1 && number <= 0) {
+  /** SQL {@code LOG(number, base)} function applied to
+   * double and BigDecimal values.
+   *
+   * @param nonPositiveIsNull if true return null for non-positive values
+   */
+  public static @Nullable Double log(double number, BigDecimal base, boolean nonPositiveIsNull) {
+    if (nonPositiveIsNull && number <= 0) {
       return null;
     }
-    return Math.log(number) / Math.log(number2.doubleValue());
+    if (number <= 0 || base.doubleValue() <= 0) {
+      throw new IllegalArgumentException("Cannot take logarithm of zero or negative number");
+    }
+    return  Math.log(number) / Math.log(base.doubleValue());
   }
 
-  /** SQL {@code LOG(number, number2)} function applied to
-   * BigDecimal and double values. */
-  public static @Nullable Double log(BigDecimal number, double number2, int nullFlag) {
-    if (nullFlag == 1 && number.doubleValue() <= 0) {
+  /** SQL {@code LOG(number, base)} function applied to
+   * BigDecimal and double values.
+   *
+   * @param nonPositiveIsNull if true return null for non-positive values
+   */
+  public static @Nullable Double log(BigDecimal number, double base, Boolean nonPositiveIsNull) {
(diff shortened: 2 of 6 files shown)


caicancai:
In order to prevent LogImplementor from making the code redundant, I added SqlLibrary for identification

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4202,67 +4205,43 @@ private static class LogicalNotImplementor extends AbstractRexCallImplementor {
...
+        List<Expression> argValueList, int flag) {
       Expression operand0 = argValueList.get(0);
       final Expressions.FluentList<Expression> list = Expressions.list(operand0);
       switch (call.getOperator().getName()) {
       case "LOG":
         if (argValueList.size() == 2) {
-          return list.append(argValueList.get(1)).append(Expressions.constant(1));
+          return list.append(argValueList.get(1)).append(Expressions.constant(flag));
+        }
+        if (flag == 2) {
```
YiwenWu, line 4234: Whether to consider BigQuery case, flag=0 ?

caicancai: Thanks for the review, I will improve the annotations to explain it later.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7045,6 +7049,33 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+  @Test void testPostgresLogFunc() {
+    final SqlOperatorFixture f0 = fixture()
+        .setFor(SqlLibraryOperators.LOG_POSTGRES, VmName.EXPAND);
+    f0.checkFails("^log(100, 10)^",
+        "No match found for function signature LOG\\(<NUMERIC>, <NUMERIC>\\)", false);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.POSTGRESQL);
+    f.checkScalarApprox("log(10, 10)", "DOUBLE NOT NULL",
+        isWithin(1.0, 0.000001));
+    f.checkScalarApprox("log(64, 8)", "DOUBLE NOT NULL",
+        isWithin(2.0, 0.000001));
```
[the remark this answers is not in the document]

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7045,6 +7049,33 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
     f0.forEachLibrary(list(SqlLibrary.MYSQL, SqlLibrary.SPARK), consumer);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6321">[CALCITE-6321]
```
YiwenWu, line 7053: Jira link error

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7045,6 +7049,33 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+   * Add LOG function (enabled in PostgreSQL library)</a>. */
+  @Test void testPostgresLogFunc() {
+    final SqlOperatorFixture f0 = fixture()
+        .setFor(SqlLibraryOperators.LOG_POSTGRES, VmName.EXPAND);
+    f0.checkFails("^log(100, 10)^",
+        "No match found for function signature LOG\\(<NUMERIC>, <NUMERIC>\\)", false);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.POSTGRESQL);
+    f.checkScalar("log(10, 10)", 1.0,
+        "DOUBLE NOT NULL");
+    f.checkScalar("log(64, 8)", 2.0,
```
YiwenWu, line 7063: LOG(base, value), the result of log(64, 8) is not 2.0

YiwenWu (changes requested):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: ohhh, seems to be a bug in sqlfunctions, I open a new jira case, bigquery is log(64, 8) == 2?
I overlooked the problem, and I'm sorry

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4202,67 +4205,43 @@ private static class LogicalNotImplementor extends AbstractRexCallImplementor {
...
-    }
-
-    @Override Expression implementSafe(final RexToLixTranslator translator,
-        final RexCall call, final List<Expression> argValueList) {
-      return Expressions.call(BuiltInMethod.LOG.method, args(call, argValueList));
-    }
 
     private static List<Expression> args(RexCall call,
-        List<Expression> argValueList) {
+        List<Expression> argValueList, int flag) {
```
XuQianJin-Stars, line 4226: The args has been changed and the default value of flag is 0. Have compatibility issues been considered here?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: fix

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4202,67 +4205,54 @@ private static class LogicalNotImplementor extends AbstractRexCallImplementor {
...
-      return Expressions.call(BuiltInMethod.LOG.method, args(call, argValueList));
-    }
 
     private static List<Expression> args(RexCall call,
-        List<Expression> argValueList) {
-      Expression operand0 = argValueList.get(0);
+        List<Expression> argValueList, int flag) {
+      Expression operand0;
+      Expression operand1;
+      if (flag == 0 && argValueList.size() == 2) {
```
caicancai, line 4229: I don't know if this part of the code can be optimized, but it seems to be the only way to match it

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: fix

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: done

YiwenWu (changes requested):

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4202,67 +4205,55 @@ private static class LogicalNotImplementor extends AbstractRexCallImplementor {
...
+        operand0 = argValueList.get(0);
+        operand1 = argValueList.get(1);
+      } else if ((flag == 1 || flag == 2) && argValueList.size() == 2) {
```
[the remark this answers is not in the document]
YiwenWu:
```
  private static class LogImplementor extends AbstractRexCallImplementor {

    @Override Expression implementSafe(final RexToLixTranslator translator,
      return Expressions.call(BuiltInMethod.LOG.method, args(call, argValueList, library));
    }

    private static List<Expression> args(RexCall call,
        List<Expression> argValueList, SqlLibrary library) {
      Pair<Expression, Expression> baseAndValuePair;
      if (argValueList.size() == 1) {
        ConstantExpression defaultBase = library == SqlLibrary.POSTGRESQL
            ? Expressions.constant(BigDecimal.TEN) : Expressions.constant(Math.exp(1));
        baseAndValuePair = Pair.of(defaultBase, argValueList.get(0));
      } else {
        boolean isBaseFirst = library == SqlLibrary.MYSQL || library == SqlLibrary.POSTGRESQL;
        baseAndValuePair = isBaseFirst ? Pair.of(argValueList.get(0), argValueList.get(1))
            : Pair.of(argValueList.get(1), argValueList.get(0));
      }
      Expression base = baseAndValuePair.left;
      Expression value = baseAndValuePair.right;
      final Expressions.FluentList<Expression> list = Expressions.list();
      switch (call.getOperator().getName()) {
      case "LOG":
        return list.append(value).append(base);
      case "LN":
        return list.append(value).append(Expressions.constant(Math.exp(1)));
      case "LOG2":
        return list.append(value).append(Expressions.constant(2));
      case "LOG10":
        return list.append(value).append(Expressions.constant(BigDecimal.TEN));
      default:
        throw new AssertionError("Operator not found: " + call.getOperator());
      }
    }
  }
```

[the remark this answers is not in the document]
caicancai: thanks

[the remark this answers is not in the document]
caicancai: Thank you for your code. I will optimize this part based on your code. Thank you very much.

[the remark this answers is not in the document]
caicancai: You also need to consider the flag, because mysql's log(0) will return null, but bigqeury and pg will not.

caicancai:
@XuQianJin-Stars @YiwenWu I optimized the implementation of the adapted log function. Can you review it for me when you have time? Thank you @YiwenWu for your suggestions.

caicancai:
@YiwenWu @XuQianJin-Stars @NobiGo PATL,  thanks

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -2237,13 +2237,21 @@ private static RelDataType deriveTypeMapFromEntries(SqlOperatorBinding opBinding
...
 
-  /** The "LOG(numeric, numeric1)" function. Returns the base numeric1 logarithm of numeric. */
+  /** The "LOG(numeric[,numeric1])" function. Returns the base numeric1 logarithm of numeric. */
   @LibraryOperator(libraries = {MYSQL, SPARK})
   public static final SqlFunction LOG_MYSQL =
       SqlBasicFunction.create(SqlKind.LOG,
           ReturnTypes.DOUBLE_FORCE_NULLABLE,
           OperandTypes.NUMERIC_OPTIONAL_NUMERIC);
 
+  /** The "LOG(numeric[,numeric1])" function. Returns the base numeric1 logarithm of numeric. */
```
YiwenWu, line 2247: The description is a bit confusing.
'Returns the logarithm of numeric1 using numeric as the base'?

site/_docs/reference.md
```diff
@@ -2805,8 +2805,9 @@ In the following:
 | f s | LEN(string)                                  | Equivalent to `CHAR_LENGTH(string)`
 | b f s | LENGTH(string)                             | Equivalent to `CHAR_LENGTH(string)`
 | h s | LEVENSHTEIN(string1, string2)                | Returns the Levenshtein distance between *string1* and *string2*
-| b | LOG(numeric1 [, numeric2 ])                    | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present
+| b | LOG(numeric1 [, numeric2 ])                    | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present if 1-arg version uses base e(ln)
 | m s | LOG(numeric1 [, numeric2 ])                  | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present, or null if *numeric1* is 0 or negative
+| p | LOG(numeric1 [, numeric2 ])                    | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present if 1-arg version uses base 10
```
YiwenWu, line 2810: "Returns the logarithm of *numeric1* to base *numeric2*",  → "numeric2" is base

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: fix

YiwenWu (approved):
LGTM

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: I modified it according to the official MySQL

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4202,67 +4206,46 @@ private static class LogicalNotImplementor extends AbstractRexCallImplementor {
...
+      operand1 = operands.right;
       final Expressions.FluentList<Expression> list = Expressions.list(operand0);
       switch (call.getOperator().getName()) {
       case "LOG":
-        if (argValueList.size() == 2) {
-          return list.append(argValueList.get(1)).append(Expressions.constant(1));
-        }
-        // fall through
+        return list.append(operand1).append(Expressions.constant(flag));
+          // fail through
```
mihaibudiu, line 4241: it is 'fall', not 'fail'

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2788,38 +2788,50 @@ public static double power(BigDecimal b0, BigDecimal b1) {
 
   // LN, LOG, LOG10, LOG2
 
-  /** SQL {@code LOG(number, number2)} function applied to double values. */
-  public static @Nullable Double log(double number, double number2, int nullFlag) {
-    if (nullFlag == 1 && number <= 0) {
+  /** SQL {@code LOG(number, base)} function applied to double values. */
```
mihaibudiu, line 2791: The meaning of 'flag' is very hard to follow from the call to the callee.
I think that a better way is to really have different functions, corresponding to the 3 different versions of 'flag'. Each function can have a nice descriptive name, like log_base_value

mihaibudiu (changes requested):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: I'm worried that this makes the code redundant, so I thought I could add a comment to make this clear.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: fall through will cause cli to fail, I don't know why, but fail through will not

mihaibudiu: this actually does not fall through, so you should just delete the comment.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2785,41 +2785,69 @@ public static double power(BigDecimal b0, BigDecimal b1) {
...
-  public static @Nullable Double log(double number, double number2, int nullFlag) {
-    if (nullFlag == 1 && number <= 0) {
+  /**
+   * SQL {@code LOG(number, base)} function applied to double values.
+   *
+   * @param dialectflag This parameter is used to specify the SQL dialect for the LOG function.
+   * 0 represents MySQL dialect, 1 represents BigQuery dialect, and 2 represents PostgreSQL dialect.
+   */
+  public static @Nullable Double log(double number, double base, int dialectflag) {
+    if (dialectflag == 0 && number <= 0) {
```
mihaibudiu, line 2797: you only compare dialectflag with 0.
So this flag should be named something like "nonPositiveIsNull".
The JavaDoc should say: "if true return null for non-positive values"

mihaibudiu (changes requested):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: yes, you are right

caicancai: done, thank you

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4202,67 +4206,48 @@ private static class LogicalNotImplementor extends AbstractRexCallImplementor {
...
+    private final SqlLibrary library;
+    LogImplementor(SqlLibrary library) {
       super("log", NullPolicy.STRICT, true);
+      this.library = library;
     }
 
     @Override Expression implementSafe(final RexToLixTranslator translator,
         final RexCall call, final List<Expression> argValueList) {
-      return Expressions.call(BuiltInMethod.LOG.method, args(call, argValueList));
+      boolean nonPositiveIsNull = library == SqlLibrary.MYSQL ? true : false;
```
mihaibudiu, line 4217: this can be simplified by removing the ? and everything after it.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -6965,8 +6965,16 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
         isWithin(9.0, 0.000001));
     f.checkScalarApprox("log(cast(10e-3 as real), 10)", "DOUBLE NOT NULL",
         isWithin(-2.0, 0.000001));
+    f.checkScalarApprox("log(10)", "DOUBLE NOT NULL",
+        isWithin(2.302585092994046, 0.000001));
     f.checkNull("log(cast(null as real), 10)");
     f.checkNull("log(10, cast(null as real))");
+    f.checkFails("log(0)",
+        "Cannot take logarithm of zero or negative number", true);
+    f.checkFails("log(0, 64)",
```
mihaibudiu, line 6974: The error message could probably be more descriptive, showing the value which is illegal.
Also, in one of these two cases it's not the number which is illegal, but the base.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7017,24 +7025,26 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+      f.checkScalarApprox("log(3,27)", "DOUBLE",
           isWithin(3.0, 0.000001));
-      f.checkScalarApprox("log(100, 10)", "DOUBLE",
-          isWithin(2.0, 0.000001));
       f.checkScalarApprox("log(10, 100)", "DOUBLE",
+          isWithin(2.0, 0.000001));
+      f.checkScalarApprox("log(100, 10)", "DOUBLE",
           isWithin(0.5, 0.000001));
-      f.checkScalarApprox("log(cast(10e6 as double), 10)", "DOUBLE",
+      f.checkScalarApprox("log(10, cast(10e6 as double))", "DOUBLE",
```
mihaibudiu, line 7036: I think these tests would be easier to read if you wrote 1e7.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -7017,24 +7025,26 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
           isWithin(7.0, 0.000001));
-      f.checkScalarApprox("log(cast(10e8 as float), 10)", "DOUBLE",
+      f.checkScalarApprox("log(10, cast(10e8 as float))", "DOUBLE",
           isWithin(9.0, 0.000001));
-      f.checkScalarApprox("log(cast(10e-3 as real), 10)", "DOUBLE",
+      f.checkScalarApprox("log(10, cast(10e-3 as real))", "DOUBLE",
```
mihaibudiu, line 7040: maybe the fact that the log of a REAL is a DOUBLE should be documented

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4202,67 +4206,48 @@ private static class LogicalNotImplementor extends AbstractRexCallImplementor {
...
-          return list.append(argValueList.get(1)).append(Expressions.constant(0));
-        }
-        // fall through
-      case "LN":
-        return list.append(Expressions.constant(Math.exp(1))).append(Expressions.constant(0));
-      case "LOG10":
-        return list.append(Expressions.constant(BigDecimal.TEN)).append(Expressions.constant(0));
-      default:
-        throw new AssertionError("Operator not found: " + call.getOperator());
+        List<Expression> argValueList, SqlLibrary library, boolean nonPositiveIsNull) {
```
mihaibudiu, line 4223: I would add JavaDoc here.
Since you are passing the library, you  as an argument, you can also compute nonPositiveIsNull in this function instead of passing it as an argument. I think this would be more consistent with the way you look at the other libraries.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: In fact, in LOG(numeric1 [, base ]) these two parameters cannot both be 0 or negative, otherwise errors will be reported in postgres and bigqeury.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: done

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
caicancai: done

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java

caicancai:
@mihaibudiu PATL, thank you

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4202,67 +4206,65 @@ private static class LogicalNotImplementor extends AbstractRexCallImplementor {
...
         final RexCall call, final List<Expression> argValueList) {
-      return Expressions.call(BuiltInMethod.LOG.method, args(call, argValueList));
-    }
-
+      return Expressions.
+          call(BuiltInMethod.LOG.method, args(call, argValueList, library));
+    }
+
+    /**
+     * This method, `args`, is used to handle the arguments for logarithmic functions in SQL.
```
mihaibudiu, line 4222: what is "`args`" here?
If you want to document the parameters please use the proper JavaDoc format using `@parameter` to describe what each parameter does.
This is almost ready.

caicancai: done

mihaibudiu (approved):

caicancai:
@mihaibudiu @YiwenWu @XuQianJin-Stars Thank you for your review

PR #3840, opened 2024-07-02
[CALCITE-6453] Simplify casts which are result of constant reduction

kasakrisz:
In `RexSimplify.simplifyCast` after constant reduction return the reduced expressions operand as result if the reduced expression and it's operand type is equal but different in nullability.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexSimplify.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexSimplify.java
@@ -2272,8 +2272,14 @@ public class RexSimplify {
           (RexLiteral) e.getOperands().get(1))
           : rexBuilder.makeCast(e.getType(), operand, safe, safe);
       executor.reduce(rexBuilder, ImmutableList.of(simplifiedExpr), reducedValues);
-      return requireNonNull(
-          Iterables.getOnlyElement(reducedValues));
+      RexNode reducedRexNode = requireNonNull(Iterables.getOnlyElement(reducedValues));
+      if (reducedRexNode.isA(SqlKind.CAST)) {
+        RexNode reducedOperand = ((RexCall) reducedRexNode).getOperands().get(0);
+        if (sameTypeOrNarrowsNullability(reducedRexNode.getType(), reducedOperand.getType())) {
+          return reducedOperand;
+        }
+      }
+      return reducedRexNode;
     default:
       if (operand == e.getOperands().get(0)) {
         return e;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -2272,8 +2272,14 @@ && sameTypeOrNarrowsNullability(e.getType(), intExpr.getType())) {
           (RexLiteral) e.getOperands().get(1))
           : rexBuilder.makeCast(e.getType(), operand, safe, safe);
       executor.reduce(rexBuilder, ImmutableList.of(simplifiedExpr), reducedValues);
-      return requireNonNull(
-          Iterables.getOnlyElement(reducedValues));
+      RexNode reducedRexNode = requireNonNull(Iterables.getOnlyElement(reducedValues));
+      if (reducedRexNode instanceof RexCall) {
```
mihaibudiu, line 2276: it looks to me dangerous to not check the call type here. It looks like you assume it's always a Cast.
Maybe at least assert that.

YiwenWu: +1 
add CAST judgement `RexNode.isA(SqlKind.CAST)`

kasakrisz: @mihaibudiu, @YiwenWu 
Thanks for the review, I changed this to check whether the reduced expression is as CAST.

core/src/test/java/org/apache/calcite/rex/RexProgramTest.java
```diff
@@ -3837,4 +3837,12 @@ private SqlSpecialOperatorWithPolicy(String name, SqlKind kind, int prec, boolea
     checkSimplify(add(zero, sub(nullInt, nullInt)), "null:INTEGER");
   }
 
+  @Test void testSimplifyCastWithConstantReduction() {
```
caicancai, line 3840: Add jira link

caicancai (approved):
overall LGTM

core/src/main/java/org/apache/calcite/rex/RexSimplify.java
caicancai: @kasakrisz You should not squash directly. You need to wait until all reviewers have reviewed before squashing. Otherwise, the reviewer will not know what you changed.

YiwenWu (approved):

NobiGo (approved):
LGTM

kgyrtkirk (approved):
I was wondering why `safe` is passed as `matchNullability` for `makeCast` [here](https://github.com/apache/calcite/blob/dad90736b0dbf30eae0d8f215112e22f64c90235/core/src/main/java/org/apache/calcite/rex/RexBuilder.java#L611) ; changing it to `false` has no effect since at that point the literal is still a `char`; only the executor makes it a `date`...

+1

asolimando:
I see several approvals (including committers) and there seems to be no pending issues, can this be merged?

mihaibudiu:
I have no other comments.

kasakrisz:
Thanks you guys for the review and merging this patch.

PR #3841, opened 2024-07-02
[CALCITE-6436] JDBC adapter generates SQL missing parentheses when co…

NobiGo:
…mparing 3 values with the same precedence like (a=b)=c


caicancai (approved):
LGTM

mihaibudiu (approved):

PR #3843, opened 2024-07-04
[CALCITE-6401] JDBC adapter cannot push down JOIN with condition includes IS TRUE、IS NULL、Dynamic parameter、CAST、Literal comparison

NobiGo:
…n include IS TRUE、IS NULL、Dynamic parameter、CAST、Literal comparison


caicancai (approved):
LGTM

PR #3845, opened 2024-07-05
[CALCITE-6451] Improve Nullability Derivation for Intersect and Minus

vbarua:
Issue: [CALCITE-6451](https://issues.apache.org/jira/browse/CALCITE-6451)

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
@@ -122,6 +122,7 @@ import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
 
 import static org.apache.calcite.test.Matchers.hasHints;
+import static org.apache.calcite.test.Matchers.hasRelDataType;
 import static org.apache.calcite.test.Matchers.hasTree;
 
 import static org.hamcrest.CoreMatchers.allOf;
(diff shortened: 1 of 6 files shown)


core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
```diff
@@ -2317,6 +2318,69 @@ private static RelNode groupIdRel(RelBuilder builder, boolean extra) {
     assertThat(root, hasTree(expected));
   }
 
+  @ParameterizedTest
+  @ValueSource(booleans = {false, true})
+  void testUnionTypeDerivation(boolean all) {
+    final RelBuilder builder = RelBuilder.create(config().build());
```
vbarua, line 2324: I'm not sure if this is the best place for these tests. I'm open to suggestions about where they should go.

vbarua:
This interacts interestingly/poorly with the [IntersectToDistinctRule](https://github.com/apache/calcite/blob/8a96095a64bc1cc955438d219d5f1fbcbc5762b7/core/src/main/java/org/apache/calcite/rel/rules/IntersectToDistinctRule.java), which rewrites Intersects to Unions using something like:

```
LogicalIntersect(all=[false])
  LogicalProject(a=[$0], b=[$1], c=[$2])
    LogicalTableScan(table=[[CATALOG, TEMP, FOO]])
  LogicalProject(a=[$0], b=[$1], c=[$2])
    LogicalTableScan(table=[[CATALOG, TEMP, BAR]])
```
which becomes
```
LogicalProject(a=[$0], b=[$1], c=[$2])
  LogicalFilter(condition=[=($9, 2)])
    LogicalAggregate(group=[{0, 1, 2}], agg#0=[COUNT()])
      LogicalUnion(all=[true])
        LogicalAggregate(group=[{0, 1, 2}], agg#0=[COUNT()])
          LogicalProject(a=[$0], b=[$1], c=[$2])
            LogicalTableScan(table=[[CATALOG, TEMP, FOO]])
        LogicalAggregate(group=[{0, 1, 2}], agg#0=[COUNT()])
          LogicalProject(a=[$0], b=[$1], c=[$2])
            LogicalTableScan(table=[[CATALOG, TEMP, BAR]])
```

What's interesting about this is that while it returns the same results, it's actually somewhat lossy with regards to the nullability information because we can't infer as tight nullability bounds for the second form as we can the first. That is to say, if `FOO.a` is nullable but `BAR.a` is not nullable, then with the first form we can say that in the output `a` will not be nullable but this is not the case with the second form.

One way I could see to work around this would be to improve the [IntersectToDistinctRule](https://github.com/apache/calcite/blob/8a96095a64bc1cc955438d219d5f1fbcbc5762b7/core/src/main/java/org/apache/calcite/rel/rules/IntersectToDistinctRule.java) by including filtering information when possible. That is, when a column is not nullable in all Intersect branches, we add an `IS NOT NULL` filter to exclude nulls:
```
LogicalProject(a=[$0], b=[$1], c=[$2])
  LogicalFilter(condition=[=($9, 2)])
    LogicalAggregate(group=[{0, 1, 2}], agg#0=[COUNT()])
      LogicalUnion(all=[true])
        LogicalAggregate(group=[{0, 1, 2}], agg#0=[COUNT()])
          LogicalFilter(condition=[IS NOT NULL($0)] <-- Filter here excludes NULL columns
            LogicalProject(a=[$0], b=[$1], c=[$2])
              LogicalTableScan(table=[[CATALOG, TEMP, FOO]])
        LogicalAggregate(group=[{0, 1, 2}], agg#0=[COUNT()])
          LogicalProject(a=[$0], b=[$1], c=[$2])
            LogicalTableScan(table=[[CATALOG, TEMP, BAR]])
```

However that might also require additional work because I don't believe that Calcite can use the presence of an IS NOT NULL filter to change the output nullability.

The reason this is a problem now is that when the IntersectToDisctinctRule is applied that types of the RelNode going in and RelNode coming out no longer match.

core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
NobiGo: Hi @vbarua , I find some test in https://github.com/apache/calcite/blob/main/core/src/test/java/org/apache/calcite/test/JdbcTest.java#L4525-L4526 . Looks it is a good please for this tests.

mihaibudiu (approved):
I will edit the PR title to match the JIRA ticket title.
You will have to do the same for the commit message.

mihaibudiu:
I think you will need to fix the IntersectToDistinct rule before we can merge this PR.

vbarua:
I've pushed some changes that address the issue with the IntersectToDistinctRule. Unfortunately, it looks there's a different issue as well. The test in:
https://github.com/apache/calcite/blob/f44ed0af8efb10d326d927418110fe0ab40e76ca/core/src/test/resources/sql/misc.iq#L1654-L1665

is failing now because the output type of the relational tree no longer matches the type of the validated SQL tree. I suspect I need to update the SQL validation logic as well to address that. I will continue to look at this when I have a chance.

mihaibudiu:
I will convert this to a draft, since it's not yet ready

PR #3847, opened 2024-07-06

caicancai:

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -3580,11 +3580,25 @@ public class SqlToRelConverter {
       final RelNode inputRel = bb.root();
 
       // Project the expressions required by agg and having.
-      bb.setRoot(
-          relBuilder.push(inputRel)
-              .projectNamed(preExprs.leftList(), preExprs.rightList(), false)
-              .build(),
-          false);
+      RelNode intermediateProject = relBuilder.push(inputRel)
+          .projectNamed(preExprs.leftList(), preExprs.rightList(), false)
+          .build();
+      final RelNode r2;
+      // deal with correlation
+      final CorrelationUse p = getCorrelationUse(bb, intermediateProject);
+      if (p != null) {
+        assert p.r instanceof Project;
+        // correlation variables have been normalized in p.r, we should use expressions
+        // in p.r instead of the original exprs
+        Project project1 = (Project) p.r;
+        r2 = relBuilder.push(bb.root())
+            .projectNamed(project1.getProjects(), project1.getRowType().getFieldNames(),
+                true, ImmutableSet.of(p.id))
+            .build();
+      } else {
+        r2 = intermediateProject;
+      }
+      bb.setRoot(r2, false);
       bb.mapRootRelToFieldProjection.put(bb.root(), r.groupExprProjection);
 
       // REVIEW jvs 31-Oct-2007:  doesn't the declaration of
--- a/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
@@ -3770,6 +3770,46 @@ class SqlToRelConverterTest extends SqlToRelTestBase {
     sql(sql).withExpand(false).withDecorrelate(false).ok();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6554">[CALCITE-6554]
+   * Nested correlated sub-query in aggregation does not have inner correlation variable bound
+   * to inner projection</a>. */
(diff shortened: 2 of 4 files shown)


caicancai:

caicancai:
I think I can encapsulate a unified interface during parameter verification to convert all spark map functions

caicancai:
@mihaibudiu Hello, I converted the callbinding after parameter verification, which only applies to the map function of spark.
I debugged the spark source code and found that the type has been converted before entering the logical plan.

What do you think

core/src/main/java/org/apache/calcite/sql/type/NonNullableAccessors.java
```diff
@@ -58,4 +58,10 @@ public static RelDataType getKeyTypeOrThrow(RelDataType type) {
     return requireNonNull(type.getKeyType(),
         () -> "keyType is null for " + type);
   }
+
+  @API(since = "1.37", status = API.Status.EXPERIMENTAL)
```
mihaibudiu, line 62: Is this annotation necessary?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8182,6 +8182,19 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
         "CHAR(3) NOT NULL ARRAY NOT NULL");
     f1.checkScalar("map_keys(map('foo', 1, null, 2))", "[foo, null]",
         "CHAR(3) ARRAY NOT NULL");
+
+    f1.checkScalar("map_keys(map('foo', 1, null, 2))", "[foo, null]",
```
mihaibudiu, line 8186: I am assuming these have been checked against spark.
Perhaps you can add a comment, also listing the spark version - Spark has been known to change function behaviors in the past.

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -1491,6 +1496,43 @@ private static class PeriodOperandTypeChecker
...
+      }
+
+      final RelDataType mapKeyType =
+          getKeyTypeOrThrow(SqlTypeUtil.deriveType(callBinding, op0));
+      final RelDataType mapValueType =
+          getValueTypeOrThrow(SqlTypeUtil.deriveType(callBinding, op0));
+
+      // If the key and value types are not unknown,
+      // adjust their types to suit the Map function's constructor.
+      if (mapKeyType.getSqlTypeName() != SqlTypeName.UNKNOWN
```
mihaibudiu, line 1528: I think we have discussed this in the past: I don't think that the type checker is supposed to modify the node that is being checked.

There are type checking strategies that will backtrack (e.g., SqlOperandTypeChecker.or), and they do not expect the checked data structure to change.

I think this pattern is very dangerous, and can lead to hard to diagnose bugs.

On the other hand, I don't know how this should be done properly. The map function does need to introduce type casts. But maybe the typechecking stage is not the right stage.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: done, I tested all versions after spark3.4

core/src/main/java/org/apache/calcite/sql/type/NonNullableAccessors.java
caicancai: done

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
caicancai: Currently, this type check is specially prepared for Spark's map function, but there may be some problems that are difficult to eliminate as you said. However, Spark currently performs some type conversions during parameter checking. I have to admit that Spark does not do this very well. 
 
I believe that after I have a deeper understanding of Calcite and the type system in the future, I can improve this part. I hope that day will not be too far away.

mihaibudiu: As Julian said in the Jira comments, you should insert casts during validation, before type checking.

caicancai: ok, thank you.

core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -1532,6 +1532,7 @@ private static class MapKeyOperandTypeChecker extends SameOperandTypeChecker {
     }
   }
 
+
```
rubenada, line 1535: This new line seems an unnecessary change, please undo.

rubenada:
@mihaibudiu do you have further comments here? IMO the proposed solution seems a bit ad-hoc, but seems to do the trick (and the new code is very well located). Shall we consider this for 1.38? (after resolving conflicts, squashing commits, etc.)

caicancai:
@rubenada Looks like my local git crashed for me, I will open a new PR

PR #3849, opened 2024-07-08
[CALCITE-6448] FilterReduceExpressionsRule returns empty RelNode for R…

soumyakanti3578:
…exLiterals

Add a check in Join and Filter to require conditions to be of type BOOLEAN

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/core/Filter.java
+++ b/core/src/main/java/org/apache/calcite/rel/core/Filter.java
@@ -86,8 +86,6 @@ public abstract class Filter extends SingleRel implements Hintable {
       RexNode condition) {
     super(cluster, traits, child);
     this.condition = requireNonNull(condition, "condition");
-    assert SqlTypeName.BOOLEAN == condition.getType().getSqlTypeName()
-        : "condition should be of BOOLEAN type, but was " + condition.getType();
     assert RexUtil.isFlat(condition) : "RexUtil.isFlat should be true for condition " + condition;
     assert isValid(Litmus.THROW, null);
     this.hints = ImmutableList.copyOf(hints);
@@ -146,6 +144,9 @@ public abstract class Filter extends SingleRel implements Hintable {
   }
 
   @Override public boolean isValid(Litmus litmus, @Nullable Context context) {
+    if (condition.getType().getSqlTypeName() != SqlTypeName.BOOLEAN) {
+      return litmus.fail("condition must be boolean: {}", condition.getType());
+    }
     if (RexUtil.isNullabilityCast(getCluster().getTypeFactory(), condition)) {
       return litmus.fail("Cast for just nullability not allowed");
     }
--- a/core/src/main/java/org/apache/calcite/rel/core/Filter.java
+++ b/core/src/main/java/org/apache/calcite/rel/core/Filter.java
@@ -34,6 +34,7 @@ import org.apache.calcite.rex.RexOver;
 import org.apache.calcite.rex.RexProgram;
 import org.apache.calcite.rex.RexShuttle;
 import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.util.Litmus;
 
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/calcite/rel/core/Filter.java
```diff
@@ -85,6 +86,8 @@ protected Filter(
       RexNode condition) {
     super(cluster, traits, child);
     this.condition = requireNonNull(condition, "condition");
+    assert SqlTypeName.BOOLEAN == condition.getType().getSqlTypeName()
```
mihaibudiu, line 89: I am not sure this is always true; for example, before validation where casts are introduced this may not hold.
Also, Calcite does not always represent SQL implicit casts explicitly, so the cast to Boolean which is implied here may not be in the program.

core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
```diff
@@ -2650,19 +2650,17 @@ private static RelBuilder assertSize(RelBuilder b,
         b.scan("EMP")
             .scan("DEPT")
             .join(JoinRelType.INNER,
-                b.or(b.literal(null),
-                    b.and(b.equals(b.field(2, 0, "DEPTNO"), b.literal(1)),
-                        b.equals(b.field(2, 0, "DEPTNO"), b.literal(2)),
-                        b.equals(b.field(2, 1, "DEPTNO"),
-                            b.field(2, 0, "DEPTNO")))))
+                b.and(b.equals(b.field(2, 0, "DEPTNO"), b.literal(1)),
```
mihaibudiu, line 2653: why are you modifying this test?

piglet/src/test/java/org/apache/calcite/test/PigRelExTest.java
```diff
@@ -145,76 +145,76 @@ public void testMatch() {
   }
 
   @Test void testAdd() {
-    checkTranslation("b + 3", inTree("+($1, 3)"));
+    checkTranslation("(boolean)(b + 3)", inTree("+($1, 3)"));
```
mihaibudiu, line 148: what is the meaning of this expression?

core/src/test/java/org/apache/calcite/test/RelBuilderTest.java
soumyakanti3578: Keeping a NULL literal makes the type of the condition to be of type NULL. This PR is not ready for review, and I am just testing things as tests are failing on my local machine with unrelated errors. I will mark this PR as draft for now.

core/src/main/java/org/apache/calcite/rel/core/Filter.java
caicancai: +1

zabetak: The SQL level (SqlNode) and the algebraic level (Rel/RexNode) are handled differently. The validator as well as the implicit casts logic apply on the SQL level and not at the algebraic level. The algebra is explicit so the assertion for Boolean type makes sense to me.

Other than that it may be better to put the check inside the `isValid` method which is the canonical way of checking validity and allows also overriding in case some specialized implementation wants to perform the validity check differently.

core/src/main/java/org/apache/calcite/rel/core/Join.java
```diff
@@ -100,6 +100,8 @@ protected Join(
       JoinRelType joinType) {
     super(cluster, traitSet, left, right);
     this.condition = Objects.requireNonNull(condition, "condition");
+    assert SqlTypeName.BOOLEAN == condition.getType().getSqlTypeName()
+        : "condition should be of BOOLEAN type, but was " + condition.getType();
```
zabetak, line 104: The check already exists inside the `isValid` method. What we may want to do here is include
`assert isValid` in the constructor as it is done in other RelNodes.

soumyakanti3578: For `Join`, I tried adding `assert isValid` in the constructor, but `CheckerFramework` throws an error:
```
> Task :core:compileJava
/home/runner/work/calcite/calcite/core/src/main/java/org/apache/calcite/rel/core/Join.java:107: error: [method.invocation.invalid] call to isValid(org.apache.calcite.util.Litmus,@org.checkerframework.checker.nullness.qual.Nullable org.apache.calcite.rel.RelNode.Context) not allowed on the given receiver.
    assert isValid(Litmus.THROW, null);
                  ^
  found   : @UnderInitialization(org.apache.calcite.rel.core.Join.class) @NonNull Join
  required: @Initialized @NonNull Join
```

It looks like we can't call it from the constructor without changing some config? Or it may be simpler to assert that the condition type is boolean in the constructor - just for Join.

I encountered another issue with calling `isValid` from the constructor. During `LogicalJoin` instantiation, it calls super and tries to init `Join` and then it goes through `isValid` -> `getRowType` -> `deriveRowType` -> `getSystemFieldList`. But `LogicalJoin#getSystemFieldList` simply returns `systemFieldList` and its value is `null` because the object is still initializing and unfortunately `super` has to be the first call in a constructor (for now). Finally we get an error because of the `requireNonNull` check on `systemFieldList` in `SqlValidatorUtil#deriveJoinRowType`.

PR #3850, opened 2024-07-09
[CALCITE-6460] SortRemoveConstantKeysRule fails with AssertionError due to mismatched collation on resulting Sort

rubenada:
See more info in [CALCITE-6460](https://issues.apache.org/jira/browse/CALCITE-6460)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/SortRemoveConstantKeysRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SortRemoveConstantKeysRule.java
@@ -19,6 +19,7 @@ package org.apache.calcite.rel.rules;
 import org.apache.calcite.plan.RelOptPredicateList;
 import org.apache.calcite.plan.RelOptRuleCall;
 import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelCollation;
 import org.apache.calcite.rel.RelCollationTraitDef;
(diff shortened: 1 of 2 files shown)


caicancai:
I recently learned about this rule and I thought I could help review it.

core/src/main/java/org/apache/calcite/rel/rules/SortRemoveConstantKeysRule.java
```diff
@@ -77,8 +78,13 @@ protected SortRemoveConstantKeysRule(Config config) {
       return;
     }
 
+    final RelCollation collation =
+        RelCollationTraitDef.INSTANCE.canonize(RelCollations.of(collationsList));
     final Sort result =
-        sort.copy(sort.getTraitSet(), input, RelCollations.of(collationsList));
+        sort.copy(
```
caicancai, line 84: Can we add a test in RelOptRulesTest.xml to illustrate this?

rubenada: `RelOptRulesTest#testSortRemovalOneKeyConstant` with the proposed adjustment tests this (otherwise the adjusted test fails). I have described it a bit more in detail on the Jira ticket.

caicancai (approved):
LGTM, I tested it locally

PR #3851, opened 2024-07-10
[CALCITE-6462] VolcanoPlanner internal valid may throw exception when…

NobiGo:
… log trace is enabled


core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
```diff
@@ -982,11 +982,17 @@ void propagateCostImprovements(RelNode rel) {
         if (!relNode.getTraitSet().satisfies(subset.getTraitSet())) {
           continue;
         }
-        if (!cost.isLt(subset.bestCost)) {
+
+        // Update subset best and best's cost when we find a cheaper rel
```
mihaibudiu, line 986: these two comments do not really seem to describe what the code next to them is doing

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
NobiGo: This is the condition when we need to update the best and best's cost.

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
```diff
@@ -982,11 +982,17 @@ void propagateCostImprovements(RelNode rel) {
...
+        if (relNode != subset.best && !cost.isLt(subset.bestCost)) {
           continue;
         }
-        // Update subset best cost when we find a cheaper rel or the current
-        // best's cost is changed
+
+        // Update the current best's cost when we find cost is changed
+        if (relNode == subset.best && cost.equals(subset.bestCost)) {
```
YiwenWu, line 992: A question, why `relNode == subset.best`, but the cost is different?

NobiGo: Hi  @YiwenWu , This is because RelSet's one of the subset find a cheaper rel's cost,  We need to update the  parents the best rel and best cost. In theory, this cost will become smaller, But according to the debug info, The SQL I added in the JdbcAdapterTest shows same rel node 's cost sometimes will become bigger. So I update it.

mihaibudiu: Maybe you could add this information in a comment?

NobiGo: @YiwenWu @mihaibudiu Thanks for the review. I have added the comment, please review it. Thanks.

YiwenWu (approved):

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
mihaibudiu: In general I would avoid using "I" in comments. 
I am not sure what "the same RelNode's" means here.
Can you make sure that the comment is grammatically correct?
For example, ", We" is not right, but that's not the only problem.

NobiGo: @mihaibudiu Hi, Thanks for your reminder. I update the Java doc. Hopefully, there is no ambiguity now.

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
```diff
@@ -982,11 +982,25 @@ void propagateCostImprovements(RelNode rel) {
...
-        // Update subset best cost when we find a cheaper rel or the current
-        // best's cost is changed
+
+        // Update the RelNode's cost when we find that the cost is changed.
+        //
+        // Why do we need to update it?
+        // When RelSet's one of the subsets find a cheaper RelNode, we need to update the
+        // parents of the subset to have the best RelNode and best cost.
+        // In theory, this cost will become smaller, But according to
+        // The SQL we added in the JdbcAdapterTest {@link testVolcanoPlannerInternalValid},
```
caicancai, line 997: The SQL -> the SQL

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
```diff
@@ -982,11 +982,25 @@ void propagateCostImprovements(RelNode rel) {
...
+        // it shows RelNode's cost will become bigger sometimes.
+        // So we update it.
```
caicancai, line 999: I'm not sure if the keyword "we" is appropriate to describe

caicancai (commented):
LGTM

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
mihaibudiu: Yes, I think comments should be impersonal - if you want to say "I" you should probably sign it in the code as well.

The comment could be just `cost X can be bigger than cost Y, as shown by the test @link ...`.

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
NobiGo: Update

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
mihaibudiu: The line commented does not update anything, it just calls continue.
The comment should say the opposite: if the cost is higher, continue.

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
NobiGo: @mihaibudiu  I think the java doc is ok now, WDYT?

core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
mihaibudiu: First, this is not JavaDoc, it's a regular comment.

Please make the entire comment as simple as possible. 

I still don't understand what the comment says and how it relates to the code. Please refer in the comment to variables in the code by name. This comparison is for equality, but the comments talks about "smaller". I don't understand how the comment describes the code. No need to make a story about "I" or "we", just say what the code does.

NobiGo: @mihaibudiu , To my knowledge, Java documentation and Java annotation are the same as code comments.  So maybe as you say comment is more eloquently here. And if we just describe the code code does, I will change the doc to :
`The Best RelNode's cost can be  changed in propagate progress, as shown by the test {@link JdbcAdapterTest.testVolcanoPlannerInternalValid}`

NobiGo: @mihaibudiu Hi , I think the comment is good now, So if you think still need to improve, Please submit another PR to improve it. I am ok with this.

caicancai: @NobiGo I'd be happy to test this PR in detail and then submit a PR modification comment

PR #3852, opened 2024-07-10
[CALCITE-6454] Implement array comparison operators

normanj-bitquill:
* <, <=, >, >= now work for arrays and rows
* Can also sort arrays and rows
* Comparison is performed along corresponding indexes
* Longer arrays are considered greater
* null is considered greater than anything
* Cannot change whether nulls are first or last

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -31,6 +31,7 @@ import org.apache.calcite.linq4j.Ord;
 import org.apache.calcite.linq4j.function.Deterministic;
 import org.apache.calcite.linq4j.function.Experimental;
 import org.apache.calcite.linq4j.function.Function1;
+import org.apache.calcite.linq4j.function.Functions;
 import org.apache.calcite.linq4j.function.NonDeterministic;
 import org.apache.calcite.linq4j.function.Predicate1;
 import org.apache.calcite.linq4j.tree.Primitive;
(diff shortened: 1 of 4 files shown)


mihaibudiu (commented):
the natural question is whether this works for other recursive types, such as `int array array`...

normanj-bitquill:
> the natural question is whether this works for other recursive types, such as `int array array`...

@mihaibudiu I'm not sure. This PR currently handles arrays and rows. The arrays and rows can contain scalar values as well as arrays and rows.

Do you have any examples of simple queries that could generate other types? I can generate an array in a query with `SELECT array[1, 2];` and a row with `SELECT row(1, 2);`

NobiGo:
@normanj-bitquill

1) I find different data source has different behaviour about NULL value:

```
spark-sql (default)> select array(1,2)>array(1,null);
true
Time taken: 0.192 seconds, Fetched 1 row(s)
spark-sql (default)> select array(1,2)<array(1,null);
false
Time taken: 0.127 seconds, Fetched 1 row(s)
```
But this maybe another issue to resolve it.

2) Please add the code annotation before the test to make the test more clearly. Others good to me .

mihaibudiu:
> Do you have any examples of simple queries that could generate other types? I can generate an array in a query with `SELECT array[1, 2];` and a row with `SELECT row(1, 2);`

There are some examples in SqlOperatorTest.java, for instance nested arrays: https://github.com/apache/calcite/blob/374091b0a7c24e61a5850e763c20587ff79e04ac/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java#L7248

normanj-bitquill:
@mihaibudiu I have fixed an issue with comparing rows and added tests with nesting (arrays in arrays, rows in arrays, etc).

normanj-bitquill:
> @normanj-bitquill
> 
> 1. I find different data source has different behaviour about NULL value:
> 
> ```
> spark-sql (default)> select array(1,2)>array(1,null);
> true
> Time taken: 0.192 seconds, Fetched 1 row(s)
> spark-sql (default)> select array(1,2)<array(1,null);
> false
> Time taken: 0.127 seconds, Fetched 1 row(s)
> ```
> 
> But this maybe another issue to resolve it.
> 
> 2. Please add the code annotation before the test to clarify the test. Other good to me.

1. I don't see any obvious way to support specifying the null order. It might help to consider how it should be configured. When building a Spark engine using Calcite, what would the ideal way of specifying whether NULLs should be first or last in array comparisons?

2. I'm not sure what you mean. I added a comment before the tests to say what they are for.

mihaibudiu (approved):

PR #3854, opened 2024-07-12
[CALCITE-6468] RelDecorrelator throws AssertionError if correlated variable is used as Aggregate group key

rubenada:
See more info in CALCITE-6468

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
@@ -560,11 +560,17 @@ public class RelDecorrelator implements ReflectiveVisitor {
       // Now add the corVars from the input, starting from
       // position oldGroupKeyCount.
       for (Map.Entry<CorDef, Integer> entry : frame.corDefOutputs.entrySet()) {
-        RexInputRef.add2(projects, entry.getValue(), newInputOutput);
-
-        corDefOutputs.put(entry.getKey(), newPos);
-        mapNewInputToProjOutputs.put(entry.getValue(), newPos);
-        newPos++;
+        // Verify if the CorDef position was already added to the mapNewInputToProjOutputs
+        // during the previous group key processing
+        final Integer pos = mapNewInputToProjOutputs.get(entry.getValue());
+        if (pos == null) {
+          RexInputRef.add2(projects, entry.getValue(), newInputOutput);
+          corDefOutputs.put(entry.getKey(), newPos);
+          mapNewInputToProjOutputs.put(entry.getValue(), newPos);
+          newPos++;
+        } else {
+          corDefOutputs.put(entry.getKey(), pos);
+        }
       }
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -560,11 +560,15 @@ protected RexNode removeCorrelationExpr(
       // Now add the corVars from the input, starting from
       // position oldGroupKeyCount.
       for (Map.Entry<CorDef, Integer> entry : frame.corDefOutputs.entrySet()) {
-        RexInputRef.add2(projects, entry.getValue(), newInputOutput);
-
-        corDefOutputs.put(entry.getKey(), newPos);
-        mapNewInputToProjOutputs.put(entry.getValue(), newPos);
-        newPos++;
+        final Integer pos = mapNewInputToProjOutputs.get(entry.getValue());
```
mihaibudiu, line 563: I think a comment would help maintainers understand why this can happen

core/src/test/java/org/apache/calcite/sql2rel/RelDecorrelatorTest.java
```diff
@@ -86,4 +99,75 @@ public static Frameworks.ConfigBuilder config() {
...
+          + "    LogicalAggregate(group=[{0}], EXPR$0=[AVG($1)])\n"
+          + "      LogicalProject(DEPTNO=[$0], TOTAL=[$1])\n"
+          + "        LogicalAggregate(group=[{0}], TOTAL=[SUM($1)])\n"
+          + "          LogicalProject(DEPTNO=[$0], SAL=[$1])\n"
+          + "            LogicalFilter(condition=[IS NOT NULL($0)])\n"
+          + "              LogicalProject(DEPTNO=[$7], SAL=[$5])\n"
+          + "                LogicalTableScan(table=[[scott, EMP]])\n";
+      assertThat(after, hasTree(planAfter));
+    } catch (Exception e) {
+      throw TestUtil.rethrow(e);
```
mihaibudiu, line 169: why do you have to catch and rethrow?

mihaibudiu (approved):

core/src/test/java/org/apache/calcite/sql2rel/RelDecorrelatorTest.java
rubenada: It was just because of the exceptions declared by parse and validation. I have isolated the try-catch to the place where it is required, to make things more clear.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
rubenada: Agree. Comment added.

julianhyde:
Change looks good. Can you also add a Quidem test, so we can make sure that the plan runs? `sub-query.iq` might be a good place.

caicancai (approved):

PR #3855, opened 2024-07-14
[CALCITE-6464] Type inference for DECIMAL division seems incorrect

timgrein:
The javadocs of `RelDataTypeSystem.deriveDecimalDivideType` state the following algorithm for deriving the decimal type of a decimal division:
 
```
d = p1 - s1 + s2
s < max(6, s1 + p2 + 1)
p = d + s
```

The inequality here doesn't make 100% sense to me as the result of an inequality is not a number, but a boolean. Should it be the following?

```
d = p1 - s1 + s2
s = max(6, s1 + p2 + 1) 
p = d + s 
```

I don't have access to the original SQL standard from 2003, but this would align at least with how MS SQL Server defines type derivation for decimal division: [Precision, Scale and Length (Transact SQL)](https://learn.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql?view=sql-server-ver16).

If so I propose to fix and follow the docs and adjust the previous implementation from:

```
final int maxNumericPrecision = getMaxNumericPrecision();
int dout =
    Math.min(
        p1 - s1 + s2,
        maxNumericPrecision);

int scale = Math.max(6, s1 + p2 + 1);
scale =
    Math.min(
        scale,
        maxNumericPrecision - dout);
scale = Math.min(scale, getMaxNumericScale());

int precision = dout + scale; 
```
 
to:

```
int scale = Math.max(6, s1 + p2 + 1);
int precision = p1 - s1 + s2 + scale;
precision = Math.min(precision, getMaxNumericPrecision());
scale = Math.min(scale, getMaxNumericScale()); 
```

This also fixes the issue described in [[CALCITE-6464] Type inference for DECIMAL division seems incorrect](https://issues.apache.org/jira/browse/CALCITE-6464). I added an explicit regression test to check this. 

I've also adjusted one test case name to reflect that it tests subtraction and not addition and added an explicit test case for addition. I can also remove the test changes from this PR and open a separate one, if needed.

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -2595,7 +2595,7 @@ public class SqlOperatorTest {
     }
     f.checkNull("1e1 / cast(null as float)");
 
-    f.checkScalarExact("100.1 / 0.00000000000000001", "DECIMAL(19, 0) NOT NULL",
+    f.checkScalarExact("100.1 / 0.00000000000000001", "DECIMAL(19, 19) NOT NULL",
         "1.001E+19");
   }
 
@@ -9420,9 +9420,9 @@ public class SqlOperatorTest {
     f.checkScalar("safe_divide(cast(2 as bigint), cast(4 as bigint))",
         "0.5", "DOUBLE");
     f.checkScalar("safe_divide(cast(15 as bigint), cast(1.2 as decimal(2,1)))",
-        "12.5", "DECIMAL(19, 0)");
+        "12.5", "DECIMAL(19, 6)");
     f.checkScalar("safe_divide(cast(4.5 as decimal(2,1)), cast(3 as bigint))",
-        "1.5", "DECIMAL(19, 18)");
+        "1.5", "DECIMAL(19, 19)");
     f.checkScalar("safe_divide(cast(4.5 as decimal(2,1)), "
         + "cast(1.5 as decimal(2, 1)))", "3", "DECIMAL(8, 6)");
     f.checkScalar("safe_divide(cast(3 as double), cast(3 as bigint))",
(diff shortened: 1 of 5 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -2595,7 +2595,7 @@ void checkModOperatorDivByZero(SqlOperatorFixture f) {
     }
     f.checkNull("1e1 / cast(null as float)");
 
-    f.checkScalarExact("100.1 / 0.00000000000000001", "DECIMAL(19, 0) NOT NULL",
+    f.checkScalarExact("100.1 / 0.00000000000000001", "DECIMAL(19, 19) NOT NULL",
```
mihaibudiu, line 2598: This looks wrong, DECIMAL(19, 19) cannot represent the result of this computation.
Mysql infers a result of DECIMAL(25, 5) for the result.

timgrein: I'll take a look, thanks for pointing it out. Do you've any guidance, which vendors/dbs one should try? (which do you usually try?) I guess some deviate more than others from the standard.

asolimando: Postgres is the most referenced one in discussions here around, but it's generally a good practice to check SQL Server, Oracle, MySQL as well (https://www.db-fiddle.com/ and https://sqlfiddle.com/ are a convenient way to do so in general).

mihaibudiu: In this case Postgres is not a good reference, because it supports arbitrary precision DECIMAL numbers. I have found a relatively precise description for SQL server and I will submit an alternate PR for this issue.

mihaibudiu:
Closing in favor of #3906

PR #3856, opened 2024-07-15
[CALCITE-6218] RelToSqlConverter fails to convert correlated lateral …

YiwenWu:
```
SELECT 
  "a", "x"
FROM 
  "myDb"."myTable",
  unnest("xs") as "x";
```

original obtained sql:
```
SELECT
  "$cor0"."a",
  "$cor0"."xs0" AS "x" -- <-- xs0 ?
FROM 
  (
    SELECT "a", "n1"."n11"."b", "n1"."n12"."c", "n2"."d", "xs", "e" FROM "myDb"."myTable"
  ) AS "$cor0",
  LATERAL UNNEST (
   SELECT "$cor0"."xs" FROM (VALUES (0)) AS "t" ("ZERO")
  ) AS "t1" ("xs") AS "t10"
```

this PR fixed obtained sql:
```
SELECT
  "$cor0"."a",
  "t10."xs" AS "x"
FROM 
  (
    SELECT "a", "n1"."n11"."b", "n1"."n12"."c", "n2"."d", "xs", "e" FROM "myDb"."myTable"
  ) AS "$cor0",
  LATERAL UNNEST (
   SELECT "$cor0"."xs" FROM (VALUES (0)) AS "t" ("ZERO")
  ) AS "t1" ("xs") AS "t10"
```

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
@@ -397,7 +397,7 @@ public class RelToSqlConverter extends SqlImplementor
   public Result visit(Correlate e) {
     final Result leftResult =
         visitInput(e, 0)
-            .resetAlias(e.getCorrelVariable(), e.getRowType());
+            .resetAlias(e.getCorrelVariable(), e.getInput(0).getRowType());
     parseCorrelTable(e, leftResult);
     final Result rightResult = visitInput(e, 1);
     final SqlNode rightLateral =
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterStructsTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterStructsTest.java
@@ -43,7 +43,7 @@ class RelToSqlConverterStructsTest {
     String query = "SELECT * FROM \"myTable\"";
     String expected = "SELECT \"a\", "
         + "ROW(ROW(\"n1\".\"n11\".\"b\"), ROW(\"n1\".\"n12\".\"c\")) AS \"n1\", "
-        + "ROW(\"n2\".\"d\") AS \"n2\", "
+        + "ROW(\"n2\".\"d\") AS \"n2\", \"xs\", "
         + "\"e\"\n"
         + "FROM \"myDb\".\"myTable\"";
     sql(query).ok(expected);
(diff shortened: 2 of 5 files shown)


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterStructsTest.java
```diff
@@ -41,11 +41,9 @@ private RelToSqlConverterTest.Sql sql(String sql) {
 
   @Test void testNestedSchemaSelectStar() {
     String query = "SELECT * FROM \"myTable\"";
-    String expected = "SELECT \"a\", "
-        + "ROW(ROW(\"n1\".\"n11\".\"b\"), ROW(\"n1\".\"n12\".\"c\")) AS \"n1\", "
-        + "ROW(\"n2\".\"d\") AS \"n2\", "
-        + "\"e\"\n"
-        + "FROM \"myDb\".\"myTable\"";
+    String expected = "SELECT \"a\", ROW(ROW(\"n1\".\"n11\".\"b\"), "
```
mihaibudiu, line 44: I would refrain for reformatting the string to make the comparison easier

mihaibudiu: In fact, is there any difference here?

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -397,7 +397,7 @@ private boolean isCommaJoin(Join join) {
   public Result visit(Correlate e) {
     final Result leftResult =
         visitInput(e, 0)
-            .resetAlias(e.getCorrelVariable(), e.getRowType());
+            .resetAlias(e.getCorrelVariable(), e.getInput(0).getRowType());
```
mihaibudiu, line 400: this seems like a pretty big bug, I wonder how it wasn't discovered sooner
Is it because for correlated queries the type of e is often the same as the type of e.getInput(0)?

mihaibudiu (approved):

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterStructsTest.java
YiwenWu: Yes, there is no difference, it has been rolled back.

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
YiwenWu: It may be that fewer cases are used. Correlate is similar to Join, with leftNode and rightNode, among which e.getInput(0) is equivalent to leftNode. Therefore the RowType of e.getInput(0) is a subset of Correlate.

PR #3858, opened 2024-07-16
[CALCITE-6296] Support IS NULL in Arrow adapter

timgrein:
Add the unary `isnull` Gandiva operator to `ArrowTranslator`.

### DIFF
--- a/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
+++ b/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
@@ -122,10 +122,16 @@ public class ArrowTable extends AbstractTable
         treeNodes.add(
             TreeBuilder.makeField(schema.getFields()
                 .get(schema.getFields().indexOf(schema.findField(data[0])))));
-        treeNodes.add(makeLiteralNode(data[2], data[3]));
-        String equality = data[1];
+
+        // if the split condition has more than two parts it's a binary operator
+        // with an additional literal node
+        if (data.length > 2) {
+          treeNodes.add(makeLiteralNode(data[2], data[3]));
+        }
+
+        String operator = data[1];
         conditionNodes.add(
-            TreeBuilder.makeFunction(equality, treeNodes, new ArrowType.Bool()));
+            TreeBuilder.makeFunction(operator, treeNodes, new ArrowType.Bool()));
       }
       final Condition filterCondition;
       if (conditionNodes.size() == 1) {
--- a/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
+++ b/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
@@ -114,7 +114,7 @@ class ArrowTranslator {
     return predicates;
   }
 
-  /** Translate a binary relation. */
+  /** Translate a binary or unary relation. */
   private String translateMatch2(RexNode node) {
     switch (node.getKind()) {
     case EQUALS:
@@ -127,8 +127,10 @@ class ArrowTranslator {
       return translateBinary("greater_than", "<", (RexCall) node);
     case GREATER_THAN_OR_EQUAL:
       return translateBinary("greater_than_or_equal_to", "<=", (RexCall) node);
+    case IS_NULL:
+      return translateUnary("isnull", (RexCall) node);
     default:
-      throw new UnsupportedOperationException("Unsupported binary operator " + node);
+      throw new UnsupportedOperationException("Unsupported operator " + node);
     }
   }
 
@@ -183,6 +185,34 @@ class ArrowTranslator {
       }
     }
     return name + " " + op + " " + valueString + " " + valueType;
+  }
+
+  /** Translates a call to a unary operator. */
+  private String translateUnary(String op, RexCall call) {
+    final RexNode opNode = call.operands.get(0);
+    @Nullable String expression = translateUnary2(op, opNode);
+
+    if (expression != null) {
+      return expression;
+    }
+
+    throw new UnsupportedOperationException("Unsupported unary operator " + call);
+  }
+
+  /** Translates a call to a unary operator. Returns null on failure. */
+  private @Nullable String translateUnary2(String op, RexNode opNode) {
+    if (opNode.getKind() == SqlKind.INPUT_REF) {
+      final RexInputRef inputRef = (RexInputRef) opNode;
+      final String name = fieldNames.get(inputRef.getIndex());
+      return translateUnaryOp(op, name);
+    }
+
+    return null;
+  }
+
+  /** Combines a field name and a unary operator to produce a predicate string. */
+  private String translateUnaryOp(String op, String name) {
+    return name + " " + op;
   }
 
   private static String getLiteralType(Object literal) {
--- a/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
+++ b/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
@@ -325,23 +325,34 @@ class ArrowAdapterTest {
         .explainContains(plan);
   }
 
+  /** Test case for
+  * <a href="https://issues.apache.org/jira/browse/CALCITE-6296">[CALCITE-6296]
+  * Support IS NULL in Arrow adapter</a>. */
   @Test void testArrowProjectFieldsWithIsNullFilter() {
     String sql = "select \"intField\", \"stringField\"\n"
         + "from arrowdata\n"
         + "where \"intField\" is null";
-    String plan;
-    if (Bug.CALCITE_6296_FIXED) {
-      plan = "ArrowToEnumerableConverter\n"
+    String plan = "ArrowToEnumerableConverter\n"
           + "  ArrowProject(intField=[$0], stringField=[$1])\n"
-          + "    ArrowFilter(condition=[IS NOT NULL($0)])\n"
+          + "    ArrowFilter(condition=[IS NULL($0)])\n"
           + "      ArrowTableScan(table=[[ARROW, ARROWDATA]], fields=[[0, 1, 2, 3]])\n\n";
-    } else {
-      plan = "PLAN=EnumerableCalc(expr#0..1=[{inputs}],"
-          + " expr#2=[IS NULL($t0)], proj#0..1=[{exprs}], $condition=[$t2])\n"
-          + "  ArrowToEnumerableConverter\n"
-          + "    ArrowProject(intField=[$0], stringField=[$1])\n"
-          + "      ArrowTableScan(table=[[ARROW, ARROWDATA]], fields=[[0, 1, 2, 3]])\n\n";
-    }
+
+    CalciteAssert.that()
+        .with(arrow)
+        .query(sql)
+        .returnsCount(0)
+        .explainContains(plan);
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6296">[CALCITE-6296]
+   * Support IS NULL in Arrow adapter</a>. */
+  @Test void testConjunctiveIsNullFilters() {
+    String sql = "select * from arrowdata\n"
+        + "where \"intField\" is null and \"stringField\" is null";
+    String plan = "PLAN=ArrowToEnumerableConverter\n"
+        + "  ArrowFilter(condition=[AND(IS NULL($0), IS NULL($1))])\n"
+        + "    ArrowTableScan(table=[[ARROW, ARROWDATA]], fields=[[0, 1, 2, 3]])\n\n";
 
     CalciteAssert.that()
         .with(arrow)
--- a/core/src/main/java/org/apache/calcite/util/Bug.java
+++ b/core/src/main/java/org/apache/calcite/util/Bug.java
@@ -236,11 +236,6 @@ public abstract class Bug {
    * [CALCITE-6295] Support IS NOT NULL in Arrow adapter</a> is fixed. */
   public static final boolean CALCITE_6295_FIXED = false;
 
-  /** Whether
-   * <a href="https://issues.apache.org/jira/browse/CALCITE/issues/CALCITE-6296">
-   * [CALCITE-6296] Support IS NULL in Arrow adapter</a> is fixed. */
-  public static final boolean CALCITE_6296_FIXED = false;
-
   /**
    * Use this to flag temporary code.
    */


asolimando (approved):
LGTM

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
```diff
@@ -329,19 +329,10 @@ static void initializeArrowState(@TempDir Path sharedTempDir) throws IOException
     String sql = "select \"intField\", \"stringField\"\n"
         + "from arrowdata\n"
         + "where \"intField\" is null";
-    String plan;
-    if (Bug.CALCITE_6296_FIXED) {
-      plan = "ArrowToEnumerableConverter\n"
+    String plan = "ArrowToEnumerableConverter\n"
```
caicancai, line 332: Just a question, why not add a new test

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
```diff
@@ -185,6 +187,36 @@ private String translateOp2(String op, String name, RexLiteral right) {
     return name + " " + op + " " + valueString + " " + valueType;
   }
 
+  /**
+   * Translates a call to a unary operator.
+   */
```
caicancai, line 192: Your javadoc format should be consistent

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
asolimando: That test was already covering exactly this missing feature, and was meant to be fixed upon implementing it.
I think it's OK unless you see the need of a new test around `IS NULL` which is not covered here.

caicancai: Because this test involves two jira cases, and both are bug fixes, I think it should be tested separately and add jira links, which may help subsequent developers understand

asolimando: When you say it involves two Jira cases, you mean that it was added in one past ticket and that we are modifying it in the context of this one?

The ticket has been originally inserted as disabled (per my request), I have subsequently enabled it when the fallback on `Enumerable` was added, with the guard on the present ticket exactly to have the test fixed as in this PR.

The value of the previous version of test was to exercise the fallback while the feature wasn't supported, but now that part is not needed anymore and it can be removed.

For clarity we can add a link to the present Jira ticket as it fully and exclusively covers the feature, and the old tickets don't apply anymore.

Would that be OK for you, @caicancai?

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
asolimando: The comment applies also to the other functions that were introduced

timgrein: Adjusted with [Adjust comment to follow the format of the other comments](https://github.com/apache/calcite/pull/3858/commits/3c48ffe1b54e1c2e1db5ff2a40089638bd67106a). I guess you meant putting it on one line?

timgrein:
> Edit: there seems to a problem with Sonar in CI which doesn't look like a transient failure, can you take a look @timgrein? Did you experience any issues locally?

I couldn't reproduce it locally. I ran the following commands:

`./gradlew clean ':arrow:test' --no-build-cache`

and

`./gradlew build`. Any other commands you want to share, which might be helpful to investigate this? @asolimando 

Both succeeded without the issue from CI, I'll check, if it appears again.

caicancai:
> > Edit: there seems to a problem with Sonar in CI which doesn't look like a transient failure, can you take a look @timgrein? Did you experience any issues locally?
> 
> I couldn't reproduce it locally. I ran the following commands:
> 
> `./gradlew clean ':arrow:test' --no-build-cache`
> 
> and
> 
> `./gradlew build`. Any other commands you want to share, which might be helpful to investigate this? @asolimando
> 
> Both succeeded without the issue from CI, I'll check, if it appears again.

2024-07-16 23-36-14屏幕截图

You can look at the cli log.

timgrein:
> > > Edit: there seems to a problem with Sonar in CI which doesn't look like a transient failure, can you take a look @timgrein? Did you experience any issues locally?
> > 
> > 
> > I couldn't reproduce it locally. I ran the following commands:
> > `./gradlew clean ':arrow:test' --no-build-cache`
> > and
> > `./gradlew build`. Any other commands you want to share, which might be helpful to investigate this? @asolimando
> > Both succeeded without the issue from CI, I'll check, if it appears again.
> 
> 2024-07-16 23-36-14屏幕截图
> 
> You can look at the cli log. Integration/jenkins/pr-head may occasionally fail. If the log exception is not very strange, I think this will not affect the merge.

Thanks for the context! Seems like it succeeded now.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
```diff
@@ -325,23 +325,15 @@ static void initializeArrowState(@TempDir Path sharedTempDir) throws IOException
         .explainContains(plan);
   }
 
+  /** <a href="https://issues.apache.org/jira/browse/CALCITE-6296">[CALCITE-6296]</a>. */
   @Test void testArrowProjectFieldsWithIsNullFilter() {
```
NobiGo, line 329: Maybe we can add another test for "a is null and b is null" or "a is null or b=c", To cover when the IS NULL is in a complicated condition.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
```diff
@@ -127,6 +127,8 @@ private String translateMatch2(RexNode node) {
       return translateBinary("greater_than", "<", (RexCall) node);
     case GREATER_THAN_OR_EQUAL:
       return translateBinary("greater_than_or_equal_to", "<=", (RexCall) node);
+    case IS_NULL:
+      return translateUnary("isnull", (RexCall) node);
     default:
       throw new UnsupportedOperationException("Unsupported binary operator " + node);
```
asolimando, line 133: I realize now that "binary" is not correct, it can be unary too, can we amend as `"Unsupported operator " + node`?

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
timgrein: Good point! Added tests with [Add Conjunctive IsNull Filter tests](https://github.com/apache/calcite/pull/3858/commits/6d31f70295e36529ff5691ae1d52f1a4c95389c8)

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
asolimando, line 133: I realize now that "binary" is not correct, it can be unary too, can we amend it as follows?
```suggestion
      throw new UnsupportedOperationException("Unsupported operator " + node);
```

timgrein: Good catch, adjusted with [Adjust UnsupportedOperationException message](https://github.com/apache/calcite/pull/3858/commits/96ddd0ce69fe9f70f96862d4bdc77a304b02aab5) (saw your old comment first :) )

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
```diff
@@ -122,7 +122,13 @@ public Enumerable<Object> query(DataContext root, ImmutableIntList fields,
...
                 .get(schema.getFields().indexOf(schema.findField(data[0])))));
-        treeNodes.add(makeLiteralNode(data[2], data[3]));
+
+        // if the split condition has more than two parts it's a binary operator
+        // with an additional literal node
+        if (data.length > 2) {
+          treeNodes.add(makeLiteralNode(data[2], data[3]));
+        }
+
         String equality = data[1];
```
asolimando, line 132: This is not necessarily an equality, right?

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
```diff
@@ -325,23 +325,29 @@ static void initializeArrowState(@TempDir Path sharedTempDir) throws IOException
         .explainContains(plan);
   }
 
+  /** <a href="https://issues.apache.org/jira/browse/CALCITE-6296">[CALCITE-6296]</a>. */
```
asolimando, line 328: We generally use this form to add Jira links for test cases (you can grep the codebase to see, if needed):
```suggestion
  /** Test case for
  * <a href="https://issues.apache.org/jira/browse/CALCITE-6296">[CALCITE-6296]
  * Support IS NULL in Arrow adapter</a>. */
```

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
asolimando: @timgrein please check https://github.com/apache/calcite/pull/3858/files#r1680614476 for the comment format

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
```diff
@@ -325,23 +325,29 @@ static void initializeArrowState(@TempDir Path sharedTempDir) throws IOException
...
-    }
+
+    CalciteAssert.that()
+        .with(arrow)
+        .query(sql)
+        .returnsCount(0)
+        .explainContains(plan);
+  }
+
+  @Test void testConjunctiveIsNullFilters() {
```
asolimando, line 345: Let's add the jira link here too, for uniformity

asolimando (changes requested):
Few minor changes emerged during review, let's address them, but they are minor and we are converging

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
timgrein: Nope, it isn't. Gandiva refers to it as a function, but I think `operator` definitely makes way more sense. Adjusted with [Rename equality to operator in ArrowTable](https://github.com/apache/calcite/pull/3858/commits/a7c195b43cc0e7b1d6c84bcdd1873efebb3be0fb)

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
timgrein: [Add test case comment to testConjunctiveIsNullFilters](https://github.com/apache/calcite/pull/3858/commits/634467a9e6bf98db8d4cf07f9fab19edf86eec0c)

timgrein:
> Few minor changes emerged during review, let's address them, but they are minor and we are converging

Thanks for the detailed review @asolimando, already helped a lot! Addressed your changes, let me know, if there's something else.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
```diff
@@ -185,6 +187,34 @@ private String translateOp2(String op, String name, RexLiteral right) {
...
+      return expression;
+    }
+
+    throw new UnsupportedOperationException("Unsupported unary operator " + call);
+  }
+
+  /** Translates a call to a unary operator. Returns null on failure. */
+  private @Nullable String translateUnary2(String op, RexNode opNode) {
+    if (opNode.getKind() == SqlKind.INPUT_REF) {
+      final RexInputRef inputRef = (RexInputRef) opNode;
```
caicancai, line 205: Are you sure this conversion will be successful? Do you need to make a judgment?

caicancai:
Mostly looks good, just a few minor comments

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
timgrein: I checked other places, where the same cast happens:
- [Cast in DruidExpressions](https://github.com/apache/calcite/blob/7d9b639cd41ff7b3e3c81a705da5a38308c0b8ac/druid/src/main/java/org/apache/calcite/adapter/druid/DruidExpressions.java#L109-L110)
- [Cast in DruidQuery](https://github.com/apache/calcite/blob/7d9b639cd41ff7b3e3c81a705da5a38308c0b8ac/druid/src/main/java/org/apache/calcite/adapter/druid/DruidQuery.java#L403-L404)
- [Cast in GeodeFilter](https://github.com/apache/calcite/blob/7d9b639cd41ff7b3e3c81a705da5a38308c0b8ac/geode/src/main/java/org/apache/calcite/adapter/geode/rel/GeodeFilter.java#L372-L374)
- [Casts in ProjectAggregateMergeRule](https://github.com/apache/calcite/blob/7d9b639cd41ff7b3e3c81a705da5a38308c0b8ac/core/src/main/java/org/apache/calcite/rel/rules/ProjectAggregateMergeRule.java#L95-L106)

I can add an `instanceof` check, but this might be unnecessary? At least that's the impression I get from the examples above.

asolimando: I don't think it's necessary, as you have noticed it's a common pattern to rely on `SqlKind` to infer the type of an object instead of `instanceof`, for me it's good as-is

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
```diff
@@ -185,6 +187,34 @@ private String translateOp2(String op, String name, RexLiteral right) {
...
+      final String name = fieldNames.get(inputRef.getIndex());
+      return translateUnaryOp(op, name);
+    }
+
+    return null;
+  }
+
+  /** Combines a field name and a unary operator to produce a predicate string. */
+  private String translateUnaryOp(String op, String name) {
```
asolimando, line 214: Nit: I just noticed this is used in a single place, I guess we can inline as it's a single line and the size would be almost the same.

asolimando (approved):
LGTM, one final nitpick but it's optional, all other comments seem to have been addressed.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
timgrein: I think we'll need at least one more unary operator afterwards, which I'll also look into: [Support IS NOT NULL in Arrow adapter](https://issues.apache.org/jira/browse/CALCITE-6295?jql=project%20%3D%20CALCITE%20AND%20component%20%3D%20arrow-adapter), so I guess we can keep it? WDYT?

caicancai (approved):
LGTM

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
caicancai: @asolimando thank you for the explanation

asolimando:
If @NobiGo is fine as well with the current status of the PR, I will ask you @timgrein to squash everything into a single commit and I will then merge. Please wait for the final confirmation before squashing as it makes incremental reviewing way harder, thanks!

NobiGo (approved):

asolimando:
@timgrein, can you please squash the commits into a single one, having "[CALCITE-6296] Support IS NULL in Arrow adapter" as commit message? After that I will merge, thanks!

PR #3859, opened 2024-07-16
[CALCITE-6473] HAVING clauses may not contain window functions

kgyrtkirk:


core/src/main/resources/org/apache/calcite/runtime/CalciteResource.properties
```diff
@@ -359,4 +359,5 @@ NoOperator=No operator for ''{0}'' with kind: ''{1}'', syntax: ''{2}'' during JS
 MustFilterFieldsMissing=SQL statement did not contain filters on the following fields: {0}
 IllegalNegativeBitGetPosition=BIT_GET/GETBIT error: negative position {0,number} not allowed
 IllegalBitGetPositionExceedsLimit=BIT_GET/GETBIT error: position {0,number} exceeds the bit upper limit {1,number}
+HavingMayNotContainWindow=Window expressions are not allowed in HAVING
```
StevenMPhillips, line 362: Would it make sense to suggest using QUALIFY in the error message?

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -6603,6 +6608,11 @@ void testGroupExpressionEquivalenceParams() {
     // OVER in clause
     sql("select ^sum(max(empno) OVER (order by deptno ROWS 2 PRECEDING))^ from emp")
         .fails(ERR_NESTED_AGG);
+
+    // OVER in HAVING
```
mihaibudiu, line 6612: is it worth mentioning the JIRA case?

mihaibudiu (approved):

core/src/main/resources/org/apache/calcite/runtime/CalciteResource.properties
NobiGo, line 362: I think the exception name "HavingMayNotContainWindow" is not very clear. Maybe we can use "WindowInHavingNotAllowed"

caicancai (approved):

kgyrtkirk:
I don't think we need to document this - it was just a bit cryptic how it failed with a planning error including a `LogicalFilter` with an `OVER` expression in it.

kgyrtkirk:
Thank you all for reviewing the changes!

PR #3861, opened 2024-07-16
[CALCITE-6474] Aggregate with constant key can get a RowCount greater than its MaxRowCount

rubenada:
More info: CALCITE-6474

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdMaxRowCount.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdMaxRowCount.java
@@ -159,7 +159,7 @@ public class RelMdMaxRowCount
     return rowCount * rel.getGroupSets().size();
   }
 
-  private static boolean allGroupKeysAreConstant(Aggregate aggregate,
+  static boolean allGroupKeysAreConstant(Aggregate aggregate,
       RelOptPredicateList predicateList) {
     final RexBuilder rexBuilder = aggregate.getCluster().getRexBuilder();
     for (int key : aggregate.getGroupSet()) {
--- a/core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
@@ -788,6 +788,23 @@ public class RelMetadataTest {
     sql(sql).assertThatRowCount(is(1D), is(0D), is(1D));
   }
 
+  /** Test case for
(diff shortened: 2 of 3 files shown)


core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -788,6 +788,20 @@ final RelMetadataFixture sql(String sql) {
     sql(sql).assertThatRowCount(is(1D), is(0D), is(1D));
   }
 
+  @Test void testRowCountAggregateConstantKeysOnBigInput() {
```
mihaibudiu, line 791: It's not obvious why this is an aggregate, I assume some rewrite step converts it into one.
That could use a comment.
Do you want to mention the JIRA case?

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/rel/metadata/RelMdMaxRowCount.java
```diff
@@ -159,7 +159,7 @@ && allGroupKeysAreConstant(rel, predicateList)) {
     return rowCount * rel.getGroupSets().size();
   }
 
-  private static boolean allGroupKeysAreConstant(Aggregate aggregate,
+  static boolean allGroupKeysAreConstant(Aggregate aggregate,
```
caicancai, line 162: protected or public might be better?

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
caicancai, line 791: Add jira link

caicancai: +1

core/src/main/java/org/apache/calcite/rel/metadata/RelMdMaxRowCount.java
rubenada: We want to reuse this in RelMdRowCount, so I think it's better to have it package-private.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
rubenada: done

rubenada: I think it is arguably clear that the aggregate comes from the count (I'm just replicating the same patter of test method name and test code from the existing ones).
Added javadoc with Jira case.

caicancai (approved):
LGTM

PR #3862, opened 2024-07-16
[CALCITE-6472] Add degree based trigonometric functions to PostgreSQL function…

normanj-bitquill:
… library

Added the following functions to only the PostgreSQL function library:
* COSD
* SIND
* TAND
* ACOSD
* ASIND
* ATAND


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3053,6 +3053,17 @@ public static double acos(double b0) {
     return Math.acos(b0);
   }
 
+  // ACOSD
+  /** SQL <code>ACOSD</code> operator applied to BigDecimal values. */
+  public static double acosd(BigDecimal b0) {
+    return Math.toDegrees(Math.acos(b0.doubleValue()));
```
mihaibudiu, line 3059: acos will supposedly throw if the value is out of range.
But the value of decimal can also be out of range for double, which will cause doubleValue to throw, and you will get a different error depending on its value.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3078,6 +3089,17 @@ public static double asin(double b0) {
     return Math.asin(b0);
   }
 
+  // ASIND
```
mihaibudiu, line 3092: Frankly, the little comment is not particularly useful

site/_docs/reference.md
```diff
@@ -2689,6 +2689,7 @@ In the following:
 |:- |:-----------------------------------------------|:-----------
 | p | expr :: type                                   | Casts *expr* to *type*
 | m | expr1 <=> expr2                                | Whether two values are equal, treating null values as the same, and it's similar to `IS NOT DISTINCT FROM`
+| p | ACOSD(numeric)                                 | Returns the inverse cosine of *numeric* in degrees
```
mihaibudiu, line 2692: I personally would like it if the return type was documented as double

mihaibudiu: the behavior for out of range values should also be documented

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8628,6 +8628,27 @@ void checkArrayReverseFunc(SqlOperatorFixture f0, SqlFunction function,
...
+                + "'ACOSD\\(<CHAR\\(3\\)>\\)'\\. Supported form\\(s\\): "
+                + "'ACOSD\\(<NUMERIC>\\)'",
+            false);
+    f.checkType("acosd('abc')", "DOUBLE NOT NULL");
+    f.checkScalarApprox("acosd(0.5)", "DOUBLE NOT NULL",
+        isWithin(59.99d, 60.01d));
+    f.checkScalarApprox("acosd(cast(0.5 as decimal(3, 1)))", "DOUBLE NOT NULL",
+        isWithin(59.9d, 60.1d));
+    f.checkNull("acosd(cast(null as integer))");
+    f.checkNull("acosd(cast(null as double))");
```
mihaibudiu, line 8649: how about some tests with values outside the range -1..1?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
normanj-bitquill: I have reworked the logic for these functions to try to better match the behaviour of the radian versions. The summary is:
ACOSD - returns NaN if the value is NaN or outside of -1..1
ASIND - returns NaN if the value is NaN or outside of -1..1
ATAND - returns NaN if the value is NaN or greater than `Double.MAX_VALUE`
COSD - returns NaN if the value is NaN or greater than `Double.MAX_VALUE`
SIND - returns NaN if the value is NaN or greater than `Double.MAX_VALUE`
TAND - returns NaN if the value is NaN or greater than `Double.MAX_VALUE`

When converting a `BigDecimal` value to a double, if it is larger than `Double.MAX_VALUE` then it becomes `Double.Infinity` and will be handled in the cases above.

mihaibudiu: Is this what postgres does?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
normanj-bitquill: This is following a pattern in the `SqlFunctions` source file. It is meant to start the group of `ASIND` methods. Lower down there will be a `// ASINH` comment that starts the list of `ASINH` methods.

site/_docs/reference.md
normanj-bitquill: Updated the documentation to say the return type and how out of range values are handled.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
normanj-bitquill: Added some more tests that are outside of the normal range to verify that `NaN` is returned when expected.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3053,6 +3053,22 @@ public static double acos(double b0) {
...
+  // ACOSD
+  /** SQL <code>ACOSD</code> operator applied to BigDecimal values. */
+  public static double acosd(BigDecimal b0) {
+    return acosd(b0.doubleValue());
+  }
+
+  /** SQL <code>ACOSD</code> operator applied to double values. */
+  public static double acosd(double b0) {
+    final double radians = Math.acos(b0);
+    if (Double.isFinite(radians)) {
```
rubenada, line 3065: How about extracting all these blocks of:
```
if (Double.isFinite(radians)) {
  return Math.toDegrees(radians);
} else {
  return Double.NaN;
}
```
into a single common auxiliary method?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
normanj-bitquill: I did some testing and no this doesn't match PostgreSQL. I have updated to match the behaviour I observed in PostgreSQL.

* ACOSD - returns NaN if input was NaN. Fails if outside of -1..1
* ASIND - returns NaN if input was NaN. Fails if outside of -1..1
* ATAND - returns NaN if input was NaN.
* COSD - returns NaN if input was NaN. Fails if input is greater than max double.
* SIND - returns NaN if input was NaN. Fails if input is greater than max double.
* TAND - returns NaN if input was NaN. Fails if input is greater than max double.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
normanj-bitquill: The methods have been reworked and this doesn't apply. There is perhaps room for a similar change for "cosd", "sind" and "tand", but they are rather small.

mihaibudiu (approved):

normanj-bitquill:
I have squashed the commits.

PR #3863, opened 2024-07-17
[CALCITE-6482] Oracle dialect convert boolean literal when version < 23

suibianwanwank:


core/src/main/java/org/apache/calcite/sql/SqlBoolLiteral.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * A SQL literal representing a BOOLEAN value,
+ * for example <code>BOOLEAN 'TRUE'</code>.
+ *
+ * <p>Create values using {@link SqlLiteral#createBoolean}.
+ */
+public class SqlBoolLiteral extends SqlLiteral {
```
mihaibudiu, line 30: I don't know if it's this easy to create this class.
It looks that currently SqlLiteral is used to represent Boolean literals.
With this change there will be two possible representations.
Unless you audit the whole codebase and make sure that SqlLiteral is never used to represent Booleans, everybody will handle both cases.
You can do that by asserting in SqlLiteral that the value is never a Boolean, and then you can see how many tests fail.

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
```diff
@@ -466,6 +466,12 @@ public void unparseCall(SqlWriter writer, SqlCall call, int leftPrec,
     }
   }
 
+  public void unparseBoolLiteral(SqlWriter writer,
+      SqlBoolLiteral literal, int leftPrec, int rightPrec) {
+    writer.keyword(
+        literal.getValue() == null ? "UNKNOWN" : (Boolean) literal.getValue() ? "TRUE" : "FALSE");
```
mihaibudiu, line 472: if you use a separate class you can probably add a method getBooleanValue in SqlBooleanLiteral

core/src/main/java/org/apache/calcite/sql/dialect/OracleSqlDialect.java
```diff
@@ -126,6 +139,19 @@ public OracleSqlDialect(Context context) {
...
 
+  @Override public void unparseBoolLiteral(SqlWriter writer,
+      SqlBoolLiteral literal, int leftPrec, int rightPrec) {
+    Boolean value = (Boolean) literal.getValue();
+    if (value == null || majorVersion >= 23) {
+      super.unparseBoolLiteral(writer, literal, leftPrec, rightPrec);
+      return;
+    }
+    // low version oracle not support bool literal
+    SqlCall call =
```
mihaibudiu, line 150: Creating a call seems overkill, you can just emit the expected output directly.
The expression emitted would also require parentheses.

core/src/main/java/org/apache/calcite/sql/SqlBoolLiteral.java
suibianwanwank: You're right, there doesn't seem to be a need to separate a class. I removed this in the new commit.

core/src/main/java/org/apache/calcite/sql/dialect/OracleSqlDialect.java
mihaibudiu: I think you should write `(1 = 1)`. You don't know what other operators will surround this expression

suibianwanwank: This is a very good reminder, and I noticed that someone also pointed this out in [CALCITE-6480](https://issues.apache.org/jira/browse/CALCITE-6480?jql=project%20%3D%20CALCITE%20AND%20status%20%3D%20Open%20ORDER%20BY%20created%20DESC). I will fix this.

mihaibudiu (approved):

mihaibudiu:
If the CI passes I think you can squash the commits into a single one.

suibianwanwank:
> 如果 CI 通过，我认为您可以将提交压缩为一个提交。

thanks, It's done

suibianwanwank:
> If the CI passes I think you can squash the commits into a single one.

Yes. It's done.

NobiGo (approved):

suibianwanwank:
thanks，Can this code be merged?  @mihaibudiu @NobiGo

asolimando (approved):
LGTM

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7700,6 +7700,27 @@ private void checkLiteral2(String expression, String expected) {
         .withOracle(11).throws_("Lower Oracle version(<12) doesn't support offset/fetch syntax!");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6482">[CALCITE-6482]
+   * Oracle dialect convert bool literal when version < 23</a>.*/
```
NobiGo, line 7705: The Doc comment should stay the same as the JIRA summary. And we try not to use abbreviations as much as possible.

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
```diff
@@ -466,6 +466,13 @@ public void unparseCall(SqlWriter writer, SqlCall call, int leftPrec,
     }
   }
 
+  public void unparseBoolLiteral(SqlWriter writer,
+      SqlLiteral literal, int leftPrec, int rightPrec) {
+    Object value = literal.getValue();
+    writer.keyword(
+        !(value instanceof Boolean) ? "UNKNOWN" : (Boolean) value ? "TRUE" : "FALSE");
```
NobiGo, line 473: Why do we use `value instanceof Boolean`, not the `value == null`？CASE condition has done the same thing, so I think `value == null` is clearer to me.

NobiGo (commented):
Some minor changes need to be confirmed.

NobiGo (approved):

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7700,6 +7700,27 @@ private void checkLiteral2(String expression, String expected) {
         .withOracle(11).throws_("Lower Oracle version(<12) doesn't support offset/fetch syntax!");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6482">[CALCITE-6482]
+   * Oracle dialect convert boolean literal when version < 23</a>.*/
```
NobiGo, line 7705: `</a>. */`  need to modify it. There is a space in the middle.

NobiGo (changes requested):

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
suibianwanwank: > `</a>. */` need to modify it. There is a space in the middle.

sry, I've fixed it.

NobiGo (approved):

PR #3864, opened 2024-07-18
[CALCITE-6484] Sonar analysis fails intermittently due to OOM

asolimando:
All details in the ticket, draft PR for now

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-org.gradle.jvmargs=-XX:+UseG1GC -Xmx1g -XX:MaxMetaspaceSize=512m
+org.gradle.jvmargs=-XX:+UseG1GC -Xmx1536m -XX:MaxMetaspaceSize=512m
 org.gradle.parallel=true
 # Build cache can be disabled with --no-build-cache option
 org.gradle.caching=true
@@ -173,6 +173,7 @@ xalan.version=2.7.1
 xercesImpl.version=2.9.1
 
 # sonar properties
+systemProp.sonar.gradle.skipCompile=true
+systemProp.sonar.host.url=https://sonarcloud.io
 systemProp.sonar.organization=apache
 systemProp.sonar.projectKey=apache_calcite
-systemProp.sonar.host.url=https://sonarcloud.io
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -44,9 +44,9 @@ node('ubuntu') {
       withEnv(["Path+JDK=$JAVA_JDK_17/bin","JAVA_HOME=$JAVA_JDK_17"]) {
         withCredentials([string(credentialsId: 'SONARCLOUD_TOKEN', variable: 'SONAR_TOKEN')]) {
           if ( env.BRANCH_NAME.startsWith("PR-") ) {
-            sh './gradlew --no-parallel --no-daemon jacocoAggregateTestReport sonar -PenableJacoco -Porg.sonarqube.version=4.4.1.3373 -Dsonar.pullrequest.branch=${CHANGE_BRANCH} -Dsonar.pullrequest.base=${CHANGE_TARGET} -Dsonar.pullrequest.key=${CHANGE_ID} -Dsonar.login=${SONAR_TOKEN}'
+            sh './gradlew build --no-parallel --no-daemon jacocoAggregateTestReport sonar -PenableJacoco -Porg.sonarqube.version=4.4.1.3373 -Dsonar.pullrequest.branch=${CHANGE_BRANCH} -Dsonar.pullrequest.base=${CHANGE_TARGET} -Dsonar.pullrequest.key=${CHANGE_ID} -Dsonar.token=${SONAR_TOKEN}'
           } else {
-            sh './gradlew --no-parallel --no-daemon jacocoAggregateTestReport sonar -PenableJacoco -Porg.sonarqube.version=4.4.1.3373 -Dsonar.branch.name=${BRANCH_NAME} -Dsonar.login=${SONAR_TOKEN}'
+            sh './gradlew build --no-parallel --no-daemon jacocoAggregateTestReport sonar -PenableJacoco -Porg.sonarqube.version=4.4.1.3373 -Dsonar.branch.name=${BRANCH_NAME} -Dsonar.token=${SONAR_TOKEN}'
           }
         }
       }


caicancai:
Maybe the result of the cli cannot be changed to be effective, but it is indeed improved a little from the code point of view

asolimando:
> Maybe the result of the cli cannot be changed to be effective, but it is indeed improved a little from the code point of view

I am not sure I get your point, sonar is failing with OOM sometimes, the idea is to increase memory a bit and try enough runs to see if it stops failing with the new proposed values.

Determining how many runs are enough is a bit tricky, but from observing the frequency of failures, I guess 20 are enough.

I have seen it failing for commits with little or no changes in the past, so my assumption is that empty commits would trigger a failure too if repeated enough times with an insufficient value, but that's hard to prove.

caicancai:
> > Maybe the result of the cli cannot be changed to be effective, but it is indeed improved a little from the code point of view
> 
> I am not sure I get your point, sonar is failing with OOM sometimes, the idea is to increase memory a bit and try enough runs to see if it stops failing with the new proposed values.
> 
> Determining how many runs are enough is a bit tricky, but from observing the frequency of failures, I guess 20 are enough.
> 
> I have seen it failing for commits with little or no changes in the past, so my assumption is that empty commits would trigger a failure too if repeated enough times with an insufficient value, but that's hard to prove.

Based on my recent experience with calcite, I guess that the triggering of OOM is that one PR or several PRs are submitted frequently in a period of time, which will cause OOM to occur.

caicancai:
@asolimando Do you mind if I help you test this PR?

asolimando:
> @asolimando Do you mind if I help you test this PR?

Any help is definitely welcome!

gradle.properties
```diff
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-org.gradle.jvmargs=-XX:+UseG1GC -Xmx1g -XX:MaxMetaspaceSize=512m
```
zabetak, line 17: I suppose that these jvmargs are also picked up in local runs. If that's the case then the OOM should be reproable locally if we use the same JDK. Having a local repro would make it much easier to test and prove that changes improve the situation. In fact, we could even lower the heap on local runs to see if this is what causes the problem.

asolimando: Yes, these `jvmargs` are used also locally, by setting `-Xmx300m` I get the same error as in CI:

> > Task :linq4j:compileJava
> Note: Some input files use unchecked or unsafe operations.
> Note: Recompile with -Xlint:unchecked for details.
> The Daemon will expire immediately since the JVM garbage collector is thrashing.
> The project memory settings are likely not configured or are configured to an insufficient value.
> The memory settings for this project must be adjusted to avoid this failure.
> These settings can be adjusted by setting 'org.gradle.jvmargs' in 'gradle.properties'.
> The currently configured max heap space is '304 MiB' and the configured max metaspace is '512 MiB'.
> For more information on how to set these values, please refer to https://docs.gradle.org/8.7/userguide/build_environment.html#sec:configuring_jvm_memory in the Gradle documentation.
> To disable this warning, set 'org.gradle.daemon.performance.disable-logging=true'.

So it's pretty much clear that low memory for the JVM can cause this error. What's troubling is that I had a build passing with `-Xmx512m` locally, while we had OOM with `-Xmx1g` in CI...

Tangentially, I have also noticed this warning:

> Configure project :

> The 'sonarqube' task depends on compile tasks. This behavior is now deprecated and will be removed in version 5.x. To avoid implicit compilation, set property 'sonar.gradle.skipCompile' to 'true' and make sure your project is compiled, before analysis has started.
> 
> The 'sonar' task depends on compile tasks. This behavior is now deprecated and will be removed in version 5.x. To avoid implicit compilation, set property 'sonar.gradle.skipCompile' to 'true' and make sure your project is compiled, before analysis has started.

I am wondering if setting the suggested property, and building and running sonar in two separate Gradle commands would help by spawning a fresh single-use daemon:

`./gradlew build --no-parallel --no-daemon` (or even `./gradlew build` as it's separate from Sonar?)

and 

`./gradlew --no-parallel --no-daemon jacocoAggregateTestReport sonar -PenableJacoco -Porg.sonarqube.version=4.4.1.3373 -Dsonar.branch.name=${BRANCH_NAME} -Dsonar.login=${SONAR_TOKEN}`

If all is good, subsequent tasks should not interfere with each others, so ideally there should not be any difference, but in practice the situation might be more complicated (my knowledge of Sonar is limited).

WDYT?

caicancai: I also tested it many times, and fortunately I did not trigger OOM

asolimando: https://ci-builds.apache.org/job/Calcite/job/Calcite-sonar/job/PR-3864/14/cloudbees-pipeline-explorer/ failed again with OOM with `org.gradle.jvmargs=-XX:+UseG1GC -Xmx768m -XX:MaxMetaspaceSize=512m`, will retrigger another time with these same parameters to see if the failure is confirmed.

If that's the case, both locally and in CI we have found that the same transient OOM error we generally see depends on these parameters, and hopefully a 50% increase can stabilize CI.

asolimando:
I plan to merge the PR in approximately 24 hours if there are no further comments

zabetak (approved):
LGTM, I would only highlight the fact that changing `org.gradle.jvmargs` is not something that will affect only the Sonar CI. All CI and local builds will be now running with more heap.

caicancai (approved):
LGTM，Maybe this can't solve the fundamental problem, but it may be a little more stable.

zabetak:
Hey @asolimando , I have seen (https://github.com/apache/calcite/pull/3864/commits/5165fab9438c073fef26e699f58a0ed8c8474325) that you are trying to change the heap size only for Sonar but I don't think its necessary. The 0.5g increase is very modest and I feel that it will be a positive thing overall. My previous comment was just for clarity not for requesting changes.

asolimando:
> Hey @asolimando , I have seen ([5165fab](https://github.com/apache/calcite/commit/5165fab9438c073fef26e699f58a0ed8c8474325)) that you are trying to change the heap size only for Sonar but I don't think its necessary. The 0.5g increase is very modest and I feel that it will be a positive thing overall. My previous comment was just for clarity not for requesting changes.

Thanks @zabetak for the clarification, I wanted to check if it was feasible (and it is, I had it working on a commit and broke it while testing other stuff).

I agree that we can just increase of 0.5g and being done with it, it was informative to see that it's possible to just increase the sonar job separately in CI, in case we will need it in the future.

I will clean the code back to the state it had when you guys approved and force-push it before merging.

PR #3865, opened 2024-07-18
[CALCITE-6483] Enable LEN and LENGTH in the correct function libraries

normanj-bitquill:
Enabled the LEN function for:
* Amazon Redshift
* Snowflake
* Spark

Enabled the LENGTH function for:
* BigQuery
* Hive
* PostgreSQL
* Amazon Redshift
* Snowflake
* Spark

### DIFF
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2803,8 +2803,8 @@ In the following:
 | m | JSON_STORAGE_SIZE(jsonValue)                   | Returns the number of bytes used to store the binary representation of *jsonValue*
 | b o p r s | LEAST(expr [, expr ]* )                | Returns the least of the expressions
 | b m p r s | LEFT(string, length)                   | Returns the leftmost *length* characters from the *string*
-| f s | LEN(string)                                  | Equivalent to `CHAR_LENGTH(string)`
-| b f s | LENGTH(string)                             | Equivalent to `CHAR_LENGTH(string)`
+| f r s | LEN(string)                                | Equivalent to `CHAR_LENGTH(string)`
+| b f h p r s | LENGTH(string)                       | Equivalent to `CHAR_LENGTH(string)`
 | h s | LEVENSHTEIN(string1, string2)                | Returns the Levenshtein distance between *string1* and *string2*
 | b | LOG(numeric1 [, numeric2 ])                    | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present
 | m s | LOG(numeric1 [, numeric2 ])                  | Returns the logarithm of *numeric1* to base *numeric2*, or base e if *numeric2* is not present, or null if *numeric1* is 0 or negative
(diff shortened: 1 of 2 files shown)


caicancai:
Should we add tests to SqlOperatorTest?

site/_docs/reference.md
```diff
@@ -2803,8 +2803,8 @@ In the following:
 | m | JSON_STORAGE_SIZE(jsonValue)                   | Returns the number of bytes used to store the binary representation of *jsonValue*
 | b o p r s | LEAST(expr [, expr ]* )                | Returns the least of the expressions
 | b m p r s | LEFT(string, length)                   | Returns the leftmost *length* characters from the *string*
-| f s | LEN(string)                                  | Equivalent to `CHAR_LENGTH(string)`
-| b f s | LENGTH(string)                             | Equivalent to `CHAR_LENGTH(string)`
+| f r s | LEN(string)                                | Equivalent to `CHAR_LENGTH(string)`
+| b f h p r s | LENGTH(string)                       | Equivalent to `CHAR_LENGTH(string)`
```
rubenada, line 2807: I think the `r` (from redshift) slipped in this line by mistake?

rubenada:
> Should we add tests to SqlOperatorTest?

@caicancai I think there is no need to touch `SqlOperatorTest`. The way that class works, the already existing tests `testLenFunc` and `testLengthFunc` will automatically consider the libraries in the `@LibraryOperator` annotation of `LEN` and `LENGTH` (including the newly added ones). So new checks are added transparently for the new libraries.

rubenada:
@normanj-bitquill overall LGTM. I think that current patch is wider than the original intent, so perhaps we should adjust the Jira title (and commit message) to something broader, e.g. "Add LEN and LENGTH functions to missing libraries", wdyt?

caicancai:
> @normanj-bitquill overall LGTM. I think that current patch is wider than the original intent, so perhaps we should adjust the Jira title (and commit message) to something broader, e.g. "Add LEN and LENGTH functions to missing libraries", wdyt?

@rubenada Thank you for your reply. I am worried that in some negative tests, the results of each dialect may be different. Do you mind if I test it on some databases first?

caicancai (approved):
LGTM, I test negative test

site/_docs/reference.md
normanj-bitquill: Amazon Redshift supports both `LEN()` and `LENGTH()`.
https://docs.aws.amazon.com/redshift/latest/dg/r_LEN.html

The Redshift function library extends the PostgreSQL function library and by default has all PostgreSQL functions (unless specifically excluded).

normanj-bitquill:
@rubenada I have updated the ticket title and commit message. Let me know if it still doesn't match the change.

site/_docs/reference.md
rubenada: True! I hadn't realized. Thanks for the clarification.

rubenada (approved):
LGTM

rubenada:
Thanks @normanj-bitquill , LGTM. I'll merge in the next 24-48h if no further comment appears.

NobiGo (approved):

PR #3869, opened 2024-07-19
[CALCITE-6145] Function 'TRIM' without parameters throw NullPointerEx…

timgrein:
Added an explicit check at parsing time to make sure that the string to trim is not `null`

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
@@ -131,7 +131,7 @@ public class SqlTrimFunction extends SqlFunction {
         operands[1] = SqlLiteral.createCharString(" ", pos);
       }
       if (operands[2] == null) {
-        throw new IllegalArgumentException("String to trim cannot be null");
+        throw new IllegalArgumentException("trim cannot be called without arguments");
       }
       break;
     default:
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
@@ -130,6 +130,9 @@ public class SqlTrimFunction extends SqlFunction {
       if (operands[1] == null) {
         operands[1] = SqlLiteral.createCharString(" ", pos);
       }
+      if (operands[2] == null) {
+        throw new IllegalArgumentException("String to trim cannot be null");
+      }
       break;
     default:
       throw new IllegalArgumentException(
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
@@ -134,7 +134,8 @@ public class SqlTrimFunction extends SqlFunction {
         // This variant occurs, when someone writes TRIM() without any arguments as the first two
         // absent arguments are set to default values and the third argument (string to trim)
         // stays null
-        throw new IllegalArgumentException("Invalid number of arguments to function 'TRIM'. Was expecting at least 2 arguments");
+        throw new IllegalArgumentException(
+            "Invalid number of arguments to function 'TRIM'. Was expecting at least 2 arguments");
       }
       break;
     default:
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
@@ -131,7 +131,10 @@ public class SqlTrimFunction extends SqlFunction {
         operands[1] = SqlLiteral.createCharString(" ", pos);
       }
       if (operands[2] == null) {
-        throw new IllegalArgumentException("trim cannot be called without arguments");
+        // This variant occurs, when someone writes TRIM() without any arguments as the first two
+        // absent arguments are set to default values and the third argument (string to trim)
+        // was absent, too
+        throw new IllegalArgumentException("Invalid number of arguments to function 'TRIM'. Was expecting at least 2 arguments");
       }
       break;
     default:
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
@@ -133,7 +133,7 @@ public class SqlTrimFunction extends SqlFunction {
       if (operands[2] == null) {
         // This variant occurs, when someone writes TRIM() without any arguments as the first two
         // absent arguments are set to default values and the third argument (string to trim)
-        // was absent, too
+        // stays null
         throw new IllegalArgumentException("Invalid number of arguments to function 'TRIM'. Was expecting at least 2 arguments");
       }
       break;
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -10857,6 +10857,7 @@ public class SqlOperatorTest {
     f.checkFails("trim('' from 'abcde')",
         "Trim error: trim character must be exactly 1 character",
         true);
+    f.checkFails("trim()", "String to trim cannot be null", false);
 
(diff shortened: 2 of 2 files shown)


caicancai (commented):
You should improve the description in reference.md

timgrein:
> You should improve the description in reference.md

Do you've something explicit on your mind? I've checked other functions from `Character string operators and functions` inside `reference.md` and it doesn't seem to be common to explicitly specify behavior, if certain args are `null`. I've fixed one issue though in another change: [TRIM function documentation refers to string1 two times](https://github.com/apache/calcite/commit/f44ed0af8efb10d326d927418110fe0ab40e76ca). Did you have this on your mind?

caicancai:
> > You should improve the description in reference.md
> 
> Do you've something explicit on your mind? I've checked other functions from `Character string operators and functions` inside `reference.md` and it doesn't seem to be common to explicitly specify behavior, if certain args are `null`. I've fixed one issue though in another change: [TRIM function documentation refers to string1 two times](https://github.com/apache/calcite/commit/f44ed0af8efb10d326d927418110fe0ab40e76ca). Did you have this on your mind?

Thanks for reminding me, I didn't notice this.

caicancai (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10857,6 +10857,7 @@ void assertSubFunReturns(boolean binary, String s, int start,
     f.checkFails("trim('' from 'abcde')",
         "Trim error: trim character must be exactly 1 character",
         true);
+    f.checkFails("trim()", "String to trim cannot be null", false);
```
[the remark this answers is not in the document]
timgrein: The creator of the original https://issues.apache.org/jira/browse/CALCITE-6145 suggested to throw an exception with an exact description.

I've checked the behavior of PostgreSQL (v15):

```sql
SELECT TRIM() AS trimmed_string;
```

leads to a syntax error: `Query Error: error: syntax error at or near ")"`.

```sql
SELECT TRIM("") AS trimmed_string;
```

leads to an error: `Query Error: error: zero-length delimited identifier at or near """"`.

```sql
SELECT TRIM(null) AS trimmed_string;
```

returns `null`.

So I think this aligns with the behavior of PostgreSQL throwing an error during query parsing, _but_ I think the message could be a bit better: `trim cannot be called without arguments` maybe? 

I've added an explicit test for when the value is `null`, which indeed returns `null` (aligns with PostgreSQL's behavior).

WDYT? @mihaibudiu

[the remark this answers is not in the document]
mihaibudiu: If the function trim() need to have arguments, then the type checker should have complained about not finding a function with a signature without arguments. I don't expect the parser to complain about it, that is strange. The parser should just accept the function, and the type checker should validate it.

The original message is not good, since clearly there is no 'null' involved. However, if the error comes from the parser, there isn't much you can do about it.

core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
```diff
@@ -130,6 +130,9 @@ public SqlTrimFunction(String name, SqlKind kind,
       if (operands[1] == null) {
         operands[1] = SqlLiteral.createCharString(" ", pos);
       }
+      if (operands[2] == null) {
+        throw new IllegalArgumentException("String to trim cannot be null");
```
mihaibudiu: "The second argument of trim cannot be null"
It's not clear what "string" you are referring to.

core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
```diff
@@ -130,6 +130,9 @@ public SqlTrimFunction(String name, SqlKind kind,
       if (operands[1] == null) {
         operands[1] = SqlLiteral.createCharString(" ", pos);
       }
+      if (operands[2] == null) {
```
mihaibudiu, line 133: something is wrong here. If you reached this point you have one argument.

timgrein: I think I've two, right?

For `trim(both 'a' from 'aAa')` you've three `operands`:

```
operand[0]: "BOTH"
operand[1]: "a"
operand[2]: "aAa"
```

mihaibudiu: yes, and the message says that there are "no arguments"

core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
timgrein: I think it's either the first or the third:

You can specify one argument (only the string to trim) or three (where to trim (leading, trailing both sides), what to trim from the string, the string to trim).

timgrein: > `"The second argument of trim cannot be null"`

This is not 100% correct I think as the string to trim from can be `null`: "trim('a' from cast(null as varchar(1)))" returns `null`, which is valid. I think it's more about the absence of the argument, which feels to me like a different thing than a value/field being `null`.

Ideally we should correctly fall through to the default case or handle the cases, where the string to trim is absent explicitly.

core/src/main/java/org/apache/calcite/sql/fun/SqlTrimFunction.java
timgrein: Ok so I've debugged through this:

If you call `trim()` the following happens:

- Enters the block for 3 operands (first gets a default value, second and third are `null`, not in the sense of their value, but absent)
- `operands[0]` gets set to `BOTH`
- `operands[1]` gets so to `" "`
- `operands[2]` is `null` (indicating absence of the arg not that it's `null`)

So the actual method call from a user perspective was one without arguments (`trim()`)

`trim(" a ")` works fine
`trim(both " a ")` works fine
`trim("a" from)` syntax error (string to trim is absent)
`trim(both "a" from)` syntax error (string to trim is absent)

So to handle the no-args case I think it's correct to check for `if (operands[2] == null)` in the block with `3` operands. I'll add a comment explaining it, otherwise it's confusing I think.

mihaibudiu: yes, the first two arguments have some default values

NobiGo: @mihaibudiu @timgrein Perhaps we can refer to SparkSQL's exception information.
SparkSQL:
```
spark-sql> select trim();
Error in query: Invalid number of arguments for function trim. Expected: one of 1 and 2; Found: 0; line 1 pos 7
```

timgrein:
> Please let us know when this is again ready for review.

Looking into this again, is there any guidance on where these type of validations/semantic analysis should live? I see there's another method called `validateCall`, which feels like a more appropriate place for these kind of validations.

mihaibudiu:
I think your code is in the right place.
the validation methods do not know anything about the function being called, just its type signature.

mihaibudiu:
So it this ready for review?

timgrein:
> So it this ready for review?

Sorry I was on vacation the last week, I'll give it a read again and let you know, whether it's ready to review!

timgrein:
Thank you for your patience!

I've adjusted the error message to follow what Calcite uses at a lot of other places `Invalid number of arguments to function...`. @mihaibudiu also added an explanatory comment, so it's ready to review again. Let me know, if you want to use a different approach or if this fine as a bug fix.

mihaibudiu:
Looks like you have some lines that are too long.
You can validate your commits before submitting them by running `./gradlew build`

mihaibudiu:
@timgrein this is almost ready, with just a little more work we can merge it

xiedeyantu:
I haven't looked at this PR carefully, but according to @mihaibudiu 's review suggestions, this PR is generally ready. How can I help him make some small changes with him as the author instead of resubmitting a PR?

mihaibudiu:
you can try to push another commit to his branch
or you can just give him credit in your own PR linking to this one

mihaibudiu:
Subsumed by https://github.com/apache/calcite/pull/4068

PR #3871, opened 2024-07-20
[CALCITE-6266] SqlValidatorException with LATERAL TABLE and JOIN

julianhyde:
@snuyanzin Can you please review?


mihaibudiu (approved):
I am not really familiar with the priority of operators in the grammar, but based on the tests this seems to be fine.

snuyanzin (approved):
I tested with tests (also Flink tests) and it seems ok

PR #3873, opened 2024-07-20
[CALCITE-6485] AssertionError When an IN list containing NULL has an …

NobiGo:
…implicit coercion type converter

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
+++ b/core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
@@ -105,21 +105,21 @@ class TypeCoercionTest {
     final Fixture f = fixture();
     f.checkCommonType(f.nullType, f.nullType, f.nullType, true);
     // BOOLEAN
-    f.checkCommonType(f.nullType, f.booleanType, f.booleanType, true);
+    f.checkCommonType(f.nullType, f.booleanType, f.nullableBooleanType, true);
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
```diff
@@ -638,22 +638,39 @@ static class Fixture {
     // single types
     final RelDataType nullType;
     final RelDataType booleanType;
+    final RelDataType nullableBooleanType;
```
caicancai, line 641: Should JavaDoc be added here to explain the difference from nulltype?

core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
```diff
@@ -105,21 +105,21 @@ private static ImmutableList<RelDataType> combine(
     final Fixture f = fixture();
     f.checkCommonType(f.nullType, f.nullType, f.nullType, true);
     // BOOLEAN
-    f.checkCommonType(f.nullType, f.booleanType, f.booleanType, true);
+    f.checkCommonType(f.nullType, f.booleanType, f.nullableBooleanType, true);
```
NobiGo, line 108: @caicancai Please check this test. Before this PR, we thought NULL and Boolean's common type was Boolean. After this PR. The common types will be nullableBoolean.

caicancai (approved):

NobiGo:
If there are no other comments, I plan to merge this PR in two days.

PR #3874, opened 2024-07-20
[CALCITE-6488] Ensure collations created by RelCollations are canonized once

rubenada:
More info: CALCITE-6488

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/RelCollations.java
+++ b/core/src/main/java/org/apache/calcite/rel/RelCollations.java
@@ -270,7 +270,7 @@ public class RelCollations {
     for (RelFieldCollation fc : collation.getFieldCollations()) {
       fieldCollations.add(fc.shift(offset));
     }
-    return new RelCollationImpl(fieldCollations.build());
+    return RelCollationTraitDef.INSTANCE.canonize(new RelCollationImpl(fieldCollations.build()));
   }
 
   /** Creates a copy of this collation that changes the ordinals of input
--- a/core/src/main/java/org/apache/calcite/rel/rules/SortJoinCopyRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SortJoinCopyRule.java
@@ -96,9 +96,7 @@ public class SortJoinCopyRule
     if (leftFieldCollation.isEmpty()) {
       newLeftInput = join.getLeft();
     } else {
-      final RelCollation leftCollation =
-          RelCollationTraitDef.INSTANCE.canonize(
-              RelCollations.of(leftFieldCollation));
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/RelCollations.java
```diff
@@ -270,7 +270,7 @@ public static RelCollation shift(RelCollation collation, int offset) {
     for (RelFieldCollation fc : collation.getFieldCollations()) {
       fieldCollations.add(fc.shift(offset));
     }
-    return new RelCollationImpl(fieldCollations.build());
+    return RelCollationTraitDef.INSTANCE.canonize(new RelCollationImpl(fieldCollations.build()));
```
caicancai, line 273: Looks good, maybe some JavaDoc could be added here

core/src/main/java/org/apache/calcite/rel/rules/SortJoinCopyRule.java
```diff
@@ -96,9 +96,7 @@ public SortJoinCopyRule(Class<? extends Sort> sortClass,
     if (leftFieldCollation.isEmpty()) {
       newLeftInput = join.getLeft();
     } else {
-      final RelCollation leftCollation =
-          RelCollationTraitDef.INSTANCE.canonize(
-              RelCollations.of(leftFieldCollation));
+      final RelCollation leftCollation = RelCollations.of(leftFieldCollation);
```
caicancai, line 99: One question, why did you change it? I observed that you only changed the shift interface, and the shift interface was not called here because this part itself is redundant.

caicancai (approved):
LGTM

core/src/main/java/org/apache/calcite/rel/rules/SortJoinCopyRule.java
asolimando: [RelCollations.of](https://github.com/apache/calcite/pull/3874/files#diff-d50ce55582ce0f63226670b6234a635b70884a7055fce75df03043b4a77f9f62L85) is already canonizing, only shift needs to be changed to avoid the need of canonization in the callers (this is explained in details in the Jira ticket), so it looks good to me

asolimando (approved):
LGTM

core/src/main/java/org/apache/calcite/rel/rules/SortJoinCopyRule.java
caicancai: Thanks for your answer, I also noticed this when I looked at the code

NobiGo (approved):

rubenada:
Thanks @caicancai , @asolimando , @NobiGo  for the review!

PR #3876, opened 2024-07-21
Add testDateProject and testDecimalProject for the Arrow adapter

caicancai:
part of [6305](https://issues.apache.org/jira/browse/CALCITE-6305)

### DIFF
--- a/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
+++ b/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
@@ -518,6 +518,38 @@ class ArrowAdapterTest {
         .explainContains(plan);
   }
 
+  @Test void testDateProject() {
+    String sql = "select HIREDATE from EMP";
+    String plan = "PLAN=ArrowToEnumerableConverter\n"
+        + "  ArrowProject(HIREDATE=[$4])\n"
+        + "    ArrowTableScan(table=[[ARROW, EMP]], fields=[[0, 1, 2, 3, 4, 5, 6, 7]])\n\n";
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):
This looks fine to me, but if someone from the Arrow project could review we could be more confident.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
```diff
@@ -518,6 +518,38 @@ static void initializeArrowState(@TempDir Path sharedTempDir) throws IOException
         .explainContains(plan);
   }
 
+  @Test void testDateProject() {
+    String sql = "select HIREDATE from EMP";
+    String plan = "PLAN=ArrowToEnumerableConverter\n"
+        + "  ArrowProject(HIREDATE=[$4])\n"
+        + "    ArrowTableScan(table=[[ARROW, EMP]], fields=[[0, 1, 2, 3, 4, 5, 6, 7]])\n\n";
+    String result = "HIREDATE=1970-01-01 00:00:04\n"
+        + "HIREDATE=1970-01-01 00:00:04\n"
```
NobiGo, line 527: Is this result correct? The date type should not include hours, minutes, and seconds.

NobiGo (changes requested):

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
caicancai: It seems that calcite has a bug in arrow's handling of date type. I will open a new jira to fix it

PR #3883, opened 2024-07-25
[CALCITE-6372] Support ASOF joins

mihaibudiu:
This PR is probably easiest to review commit by commit. The 4 commits correspond to various Calcite program representations.


mihaibudiu:
The CI failures are not fundamental, I will try to fix them today

caicancai:
I'm very interested in this PR. Do you mind if I spend some time reviewing it?

mihaibudiu:
Go ahead

core/src/main/java/org/apache/calcite/rel/metadata/RelMdUtil.java
```diff
@@ -831,6 +831,9 @@ public static double getMinusRowCount(RelMetadataQuery mq, Minus minus) {
     }
     double innerRowCount = left * right * selectivity;
     switch (join.getJoinType()) {
+    case ASOF:
+    case LEFT_ASOF:
```
YiwenWu, line 835: Is ASOF similar to inner join or left join?

mihaibudiu: In this implementation we have both ASOF and LEFT ASOF.
ASOF itself is an INNER join.

core/src/main/java/org/apache/calcite/sql/SqlAsofJoin.java
```diff
@@ -0,0 +1,164 @@
...
+      join.right.unparse(writer, getRightPrec(), rightPrec);
+      SqlNode matchCondition = join.matchCondition;
+      writer.keyword("MATCH_CONDITION");
+      matchCondition.unparse(writer, 0, 0);
+
+      SqlNode joinCondition = join.condition;
+      if (joinCondition != null) {
+        switch (join.getConditionType()) {
+        case ON:
+          writer.keyword("ON");
```
YiwenWu, line 146: Does this restriction need to be processed in sql `Parser.jj`?

core/src/main/java/org/apache/calcite/rel/core/RelFactories.java
```diff
@@ -398,6 +402,23 @@ RelNode createJoin(RelNode left, RelNode right, List<RelHint> hints,
...
+   */
+  public interface AsofJoinFactory {
+    /**
+     * Creates an ASOF join.
+     *
+     * @param left             Left input
+     * @param right            Right input
+     * @param hints            Hints
+     * @param condition        Join condition
+     * @param matchCondition   ASOF join match condition
```
YiwenWu, line 416: missing `joinType`

core/src/main/java/org/apache/calcite/rel/logical/LogicalAsofJoin.java
```diff
@@ -0,0 +1,170 @@
...
+      List<RelHint> hints,
+      RelNode left,
+      RelNode right,
+      RexNode condition,
+      RexNode matchCondition,
+      JoinRelType joinType,
+      ImmutableList<RelDataTypeField> systemFieldList) {
+    super(cluster, traitSet, hints, left, right, condition, ImmutableSet.of(), joinType);
+    this.systemFieldList = requireNonNull(systemFieldList, "systemFieldList");
+    this.matchCondition = matchCondition;
```
YiwenWu, line 82: limit `requireNonNull(matchCondition)`?

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -386,6 +386,8 @@ public static JoinType joinType(JoinRelType joinType) {
       return JoinType.INNER;
     case FULL:
       return JoinType.FULL;
+    case ASOF:
+      return JoinType.ASOF;
```
YiwenWu, line 390: missing `LEFT_ASOF` case?

caicancai:
I will review this PR this weekend, sorry for the delay

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableAsofJoin.java
```diff
@@ -0,0 +1,253 @@
...
+
+import com.google.common.collect.ImmutableList;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+/** Implementation of {@link LogicalAsofJoin} in
```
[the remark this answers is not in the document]
mihaibudiu: well, this doesn't implement a generic join, it only implements an ASOF join.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableAsofJoin.java
```diff
@@ -0,0 +1,253 @@
...
+ * {@link EnumerableConvention enumerable calling convention}. */
+public class EnumerableAsofJoin extends Join implements EnumerableRel {
+  // This implementation is based on {@link EnumerableHashJoin}
```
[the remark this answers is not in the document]
mihaibudiu: I don't think there would be much code reuse.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableAsofJoin.java
```diff
@@ -0,0 +1,253 @@
...
+            .replaceIfs(RelCollationTraitDef.INSTANCE,
+                () -> RelMdCollation.enumerableHashJoin(mq, left, right, joinType));
+    return new EnumerableAsofJoin(cluster, traitSet, left, right, condition, matchCondition,
+        variablesSet, joinType);
+  }
+
+  @Override public EnumerableAsofJoin copy(RelTraitSet traitSet, RexNode condition,
+                                           RelNode left, RelNode right, JoinRelType joinType,
+                                           boolean semiJoinDone) {
+    // This method does not know about the matchCondition, so it should not be called
```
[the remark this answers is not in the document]
mihaibudiu: Certainly this method cannot be right, so it's better to catch attempts to call it.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableAsofJoin.java
```diff
@@ -0,0 +1,253 @@
...
+                    rightResult.physType.generateAccessorWithoutNulls(joinInfo.rightKeys),
+                    // result selector
+                    EnumUtils.joinSelector(joinType,
+                        physType,
+                        ImmutableList.of(
+                            leftResult.physType, rightResult.physType)))
+                    // match comparator
+                    .append(matchPredicate)
+                    // timestamp comparator
+                    .append(timestampComparator)
```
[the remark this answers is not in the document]
mihaibudiu: "timestamp" is just a name for a pair of columns from each side that are being compared.
It has nothing to do with the TIMESTAMP data type.
The MATCH_CONDITION is restricted to one comparison between a pair of columns from either side. That's enforced by the validator, and part of the spec.

core/src/main/java/org/apache/calcite/sql/SqlAsofJoin.java
mihaibudiu: I think it is, right now only ON is supported in the parser.
I plan to add USING later.

mihaibudiu:
@YiwenWu thank you for the review, I have pushed a commit with fixes for problems you found.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -2092,13 +2096,32 @@ SqlNode JoinTable(SqlNode e) :
     //
     // We allow CROSS JOIN (joinType = CROSS_JOIN) to have a join condition,
     // even though that is not valid SQL; the validator will catch it.
-    LOOKAHEAD(3)
+    LOOKAHEAD(4)
     natural = Natural()
     joinType = JoinType()
     e2 = TableRef1(ExprContext.ACCEPT_QUERY_OR_JOIN)
     (
+        [ <MATCH_CONDITION> matchCondition = Expression(ExprContext.ACCEPT_SUB_QUERY) ]
```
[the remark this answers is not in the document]
mihaibudiu: (1) This grammar does not require parentheses, I think it's more general this way. We can add them if desired. 

(2) The grammar accepts any expression, but the validator will reject complex expressions, requiring a single comparison

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -786,6 +788,9 @@ public TrimResult trimFields(
         + join.getLeft().getRowType().getFieldCount()
         + join.getRight().getRowType().getFieldCount();
     final RexNode conditionExpr = join.getCondition();
+    final RexNode matchConditionExpr = (join instanceof LogicalAsofJoin)
+        ? ((LogicalAsofJoin) join).getMatchCondition()
```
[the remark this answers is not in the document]
mihaibudiu: Do you suggest an Objects.requireNonNull? I am not sure what you would gain from that.

mihaibudiu:
I have added a base `AsofJoin` class as suggested

YiwenWu (approved):
LGTM

core/src/main/codegen/templates/Parser.jj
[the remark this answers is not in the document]
YiwenWu: 1. This is a minor issue. The current implementation is based on Snowflake syntax, so keeping it consistent can help more users avoid confusion and learning new syntax.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3774,11 +3780,130 @@ protected void validateJoin(SqlJoin join, SqlValidatorScope scope) {
...
+    case LEFT_ASOF:
+    case ASOF: {
+      // In addition to the standard join checks, the ASOF join requires the
+      // ON conditions to be a conjunction of simple equalities from both relations.
+      SqlAsofJoin asof = (SqlAsofJoin) join;
+      SqlNode matchCondition = getMatchCondition(asof);
+      matchCondition = expand(matchCondition, joinScope);
+      join.setOperand(6, matchCondition);
+      validateWhereOrOn(joinScope, matchCondition, "MATCH_CONDITION");
+      ConjunctionOfEqualities conj = new ConjunctionOfEqualities();
```
suibianwanwank, line 3792: Would it be better to replace RelOptUtil.conjunctions() and match equal here.
Maybe the bool filed type affects like bool_column3 = (bool_column1 and bool_column2). WDYT

suibianwanwank: Of course, this is a rare case

core/src/main/java/org/apache/calcite/rel/metadata/RelMdUtil.java
suibianwanwank: I agree that LEFT_ASOF returns left. Would it be more accurate for asof to return (left * selectivity)?

core/src/main/codegen/templates/Parser.jj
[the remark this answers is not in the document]
mihaibudiu: This implementation accepts the Snowflake syntax, so users familiar with that syntax don't have to learn anything new. I think in general it's ok to be put as few constraints as possible on the programs you accept (the Jon Postel principle: https://en.wikipedia.org/wiki/Robustness_principle).

Also, please note that the Snowflake construct is deemed experimental in their documentation, so I don't think we should strive too much to adhere to it, maybe they will change their syntax too. (I hope they will, and, who knows, maybe we can even influence them?)

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
mihaibudiu: "and' is not the same as "=" for booleans, since "and" returns "false" for "false = false".
But maybe I misunderstand your question.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdUtil.java
mihaibudiu: I am not familiar with this part of the code, I see that other places use 1 - selectivity.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
suibianwanwank: I'm guessing that the checks here refer to snowflake's specification,  i.e. "ON clause for ASOF JOIN must contain conjunctions of equality conditions only. Each side of an equality condition must only refer to either the left table or the right table. (S.STATE = P.STATE) OR (S.LOCATION = P.LOCATION) is invalid."
We expect that condition needs to satisfy 
`l.c1 = r.c1 and l.c2 = r.c2`
But if you just check if sqlcall is equal / and, may a case like the following would also pass, if the types of c1 and c2 would both be boolean
`(l.c1 and r.c1 ) and  (l.c2 and r.c2 )`

mihaibudiu: Good point. I will add a test case to make sure this doesn't happen.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdUtil.java
suibianwanwank: My understanding is that for a regular inner join, selectivity means the probability of the condition being true in the result, and 1 - selectivity is because for an outer join we need to add the unmatched rows on the null side.
For asof join, we should consider how many matches there are against the left table, and we also could to consider the matchCondition, which is more complicated. Perhaps this optimization can be discussed in another issue.

caicancai (approved):

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
mihaibudiu: Unfortunately I can't reuse RelOptUtil because that works on RexNode, and here I only have SqlNode objects.
I will improve the ConjunctionOfEqualities visitor.

mihaibudiu:
Thank you @suibianwanwank, I hope I have fixed the problems you highlighted.

mihaibudiu:
Does anyone else plan to review this PR or can I merge it?
More work may be needed, but this implementation is fairly functional and useful.
The only thing I am aware that can be improved is handing USING joins, which can be done in a separate PR - this one is big enough as it is.

mihaibudiu:
I have tagged this with "will merge soon", please untag if you plan to review or disagree.

suibianwanwank (approved):
LGTM

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3774,11 +3780,136 @@ protected void validateJoin(SqlJoin join, SqlValidatorScope scope) {
...
+    boolean illegal = false;
+
+    @Override public @Nullable SqlNode visit(final org.apache.calcite.sql.SqlCall call) {
+      SqlKind kind = call.getKind();
+      if (kind != SqlKind.AND && kind != SqlKind.EQUALS) {
+        illegal = true;
+      }
+      if (kind == SqlKind.AND) {
+        List<SqlNode> operands = call.getOperandList();
+        for (SqlNode operand : operands) {
```
suibianwanwank, line 3903: It seems that there might still be an issue here. If the condition is a.id = b.id AND a.id2 = b.id2 AND a.id3 = b.id3, this might not pass as expected because the conditions in SqlNode are not in the same operandList.

mihaibudiu: In Calcite AND can have N operands and I believe that there is a pass prior to this which will make sure that a sequence of AND operators is flattened into a single operator.

mihaibudiu: It look like that is only true for RexNodes, so you are right, I will have to handle this case too.

mihaibudiu: I have fixed this case and added a test. Great observation.

suibianwanwank: Looks good✅

core/src/main/codegen/templates/Parser.jj
```diff
@@ -2092,13 +2096,32 @@ SqlNode JoinTable(SqlNode e) :
...
         <ON> { on = JoinConditionType.ON.symbol(getPos()); }
```
yashmayya, line 2105: @mihaibudiu this means that the `ON` clause is mandatory for `ASOF` joins as well, is that intentional? Snowflake's `ASOF JOIN` only requires the `MATCH_CONDITION`, with `ON` / `USING` being optional - https://docs.snowflake.com/en/sql-reference/constructs/asof-join#syntax.

mihaibudiu: I guess that it is mandatory with this implementation
And you can't just use 'true' either, since it requires an equijoin

yashmayya: > And you can't just use 'true' either, since it requires an equijoin

Actually, this does seem to work because the `ConjunctionOfEqualities` class in `SqlValidatorImpl` doesn't override the literal visit method of `SqlShuttle` so `ConjunctionOfEqualities.illegal` retains its default value of false for a join condition like `ON true`.

PR #3885, opened 2024-07-25
[CALCITE-6498] Elasticsearch multi-field mappings do not work

timgrein:
A field identified by `a.b` can be either a [nested](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html) or a [multi-mapped field](https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html).

Multi-field mapped fields weren't handled in the current implementation in two places:
- When an existing mapping is visited
- When you want to create a mapping inside `EmbeddedElasticsearchPolicy`

This PR adds support for multi-mapped fields. This is important to work correctly for [[CALCITE-3027] Support LIKE operator in Elasticsearch](https://issues.apache.org/jira/browse/CALCITE-3027).`text` mapped fields are usually multi-mapped fields with a second `keyword` mapping, which you'll need for a wildcard query to have semantics like the SQL `LIKE` operator. Without this change we wouldn't be able to write tests we'll need for the `LIKE` operator.

I've added an explicit `EmbeddedElasticsearchPolicyTest` as the create mapping functionality was broken inside `EmbeddedElasticsearchPolicy`, which should work correctly as all ES tests depend on this.

I've created two separate commits to make reviewing easier:
- Visitor: [Handle visiting of multi-field and nested mappings correctly in Elast…](https://github.com/apache/calcite/pull/3885/commits/8420d9ff19f6539038aca8992f5dae3fd693cccf)
- Create mapping in tests: [Handle creation of multi-field and nested mappings correctly in Embed…](https://github.com/apache/calcite/pull/3885/commits/cb5e222c863a7bf320f4e3a4ff83b5bf257f8c79)

(force-pushed to remove `.` from the commit messages, which broke CI)

### DIFF
--- a/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchJson.java
+++ b/elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/ElasticsearchJson.java
@@ -129,6 +129,7 @@ final class ElasticsearchJson {
     // "fields" is used for multi-mapped fields
     if (mapping.path("fields").isObject()
         && !isLeaf.test(mapping.path("fields"))) {
+      // recurse on multi-mapped field
       visitMappingProperties(path, (ObjectNode) mapping.get("fields"), consumer);
       return;
     }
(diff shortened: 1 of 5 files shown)


caicancai:
@timgrein I may need to spend some time to review this PR because I am not familiar with the module.

rubenada:
@timgrein sorry for the late review here.
I'm not very familiar with this part of Calcite, but it seems to me that your patch makes sense (and you have provided many tests), so perhaps we will be able to include it in the next release (1.38, coming soon). Could you please squash commits into a single one (keeping as commit message the same as the PR's title)?

caicancai (approved):

mihaibudiu (approved):

PR #3886, opened 2024-07-26
[CALCITE-6502] Parser loses position information for Expression3

mihaibudiu:
There was another bug in the parsing of array ITEM operators, which was uncovered by this fix.

### DIFF
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -3810,8 +3810,12 @@ List<Object> Expression2(ExprContext exprContext) :
                 AddExpression2b(list, ExprContext.ACCEPT_SUB_QUERY)
             |
                 <LBRACKET>
-                itemOp = getItemOp()
-                e = Expression(ExprContext.ACCEPT_SUB_QUERY)
+                ( <OFFSET> { itemOp = SqlLibraryOperators.OFFSET; } <LPAREN> { e = Expression(ExprContext.ACCEPT_SUB_QUERY); } <RPAREN>
+                | <ORDINAL> { itemOp = SqlLibraryOperators.ORDINAL; } <LPAREN> { e = Expression(ExprContext.ACCEPT_SUB_QUERY); } <RPAREN>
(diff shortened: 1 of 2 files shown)


core/src/main/codegen/templates/Parser.jj
```diff
@@ -3980,7 +3968,7 @@ SqlNode Expression3(ExprContext exprContext) :
     {
         if (list1.size() == 1) {
             // interpret as single value or query
-            return list1.get(0);
+            return list1.get(0).clone(list1.getParserPosition());
```
caicancai, line 3971: Is clone necessary here?

mihaibudiu: this is actually the line that fixes the JIRA bug

caicancai: I understand what you mean

caicancai (approved):

core/src/main/codegen/templates/Parser.jj
```diff
@@ -3810,8 +3810,12 @@ List<Object> Expression2(ExprContext exprContext) :
                 AddExpression2b(list, ExprContext.ACCEPT_SUB_QUERY)
             |
                 <LBRACKET>
-                itemOp = getItemOp()
-                e = Expression(ExprContext.ACCEPT_SUB_QUERY)
+                ( <OFFSET> { itemOp = SqlLibraryOperators.OFFSET; } <LPAREN> { e = Expression(ExprContext.ACCEPT_SUB_QUERY); } <RPAREN>
```
NobiGo, line 3813: Is this change necessary？ Can we add code in the `getItemOp()` method?

mihaibudiu: I didn't know how to do it that way: this way there is context that a simple item does not need ().

mihaibudiu: this change is necessary, without this the parser will accept an expression like `a[SAFE_OFFSET 2]`

caicancai: I tested this change and I think it is more appropriate.

NobiGo (approved):

PR #3889, opened 2024-07-29
[CALCITE-6481] Optimize 'VALUES...UNION ALL...VALUES' to a single 'VA…

NobiGo:
…LUES' the IN-list contains NULL and it is converted to VALUES

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/UnionToValuesRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/UnionToValuesRule.java
@@ -1,0 +1,145 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.plan.hep.HepRelVertex;
+import org.apache.calcite.plan.volcano.RelSubset;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Union;
+import org.apache.calcite.rel.core.Values;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rex.RexLiteral;
+
+import com.google.common.collect.ImmutableCollection;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+
+import org.immutables.value.Value;
+
+import java.util.List;
+
+/**
+ * Planner rule that converts
+ * {@link org.apache.calcite.rel.core.Union} with
+ * inputs only containing {@link org.apache.calcite.rel.core.Values}
+ * into an {@link org.apache.calcite.rel.core.Values}.
+ *
+ * <p>For example,
+ *
+ * <blockquote><pre>{@code
+ * Union(all=[true])
+ *     Values(tuples=[[{ 3, null }]])
+ *     Values(tuples=[[{ 7369, null }]])
+ *     Values(tuples=[[{ 1, 2 }]])
+ * }</pre></blockquote>
+ *
+ * <p>becomes
+ *
+ * <blockquote><pre>{@code
+ * Values(tuples=[[{ 3, null }, { 7369, null }, { 1, 2 }]])
+ * }</pre></blockquote>
+ *
+ * <p>If (<code>all</code> = <code>false</code>),
+ * {@link org.apache.calcite.rel.core.Values} value will be set list.
+ *
+ */
+@Value.Enclosing
+public class UnionToValuesRule extends RelRule<UnionToValuesRule.Config>
+    implements TransformationRule {
+
+  /**
+   * Creates a UnionToValuesRule.
+   */
+  protected UnionToValuesRule(UnionToValuesRule.Config config) {
+    super(config);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    Union union = call.rel(0);
+    List<RelNode> relNodeList = union.getInputs();
+    final RelDataType relDataType = union.getRowType();
+    if (!relNodeList.stream().allMatch(relNode ->
+        isValuesWithSpecifiedDataType(relNode, relDataType))) {
+      return;
+    }
+    ImmutableCollection.Builder<ImmutableList<RexLiteral>> tupleList =
+        getImmutableListBuilder(union.all, relNodeList);
+    Values values =
+        (Values) call.builder().values(tupleList.build(), union.getRowType()).build();
+    call.transformTo(values);
+  }
+
+  private static ImmutableCollection.Builder<ImmutableList<RexLiteral>> getImmutableListBuilder(
+      Boolean all, List<RelNode> relNodeList) {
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
```diff
@@ -258,7 +258,7 @@ protected boolean needToCast(SqlValidatorScope scope, SqlNode node,
     RelDataType fromType = validator.deriveType(scope, node);
     // This depends on the fact that type validate happens before coercion.
     // We do not have inferred type for some node, i.e. LOCALTIME.
```
caicancai, line 260: It seems that judging the type of node, null and SqlTypeName.NUL are not equivalent in my opinion. Do we need to improve this part of the comment?

core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
```diff
@@ -258,7 +258,7 @@ protected boolean needToCast(SqlValidatorScope scope, SqlNode node,
...
-    if (fromType == null) {
+    if (fromType == null || fromType.getSqlTypeName() == SqlTypeName.NULL) {
```
mihaibudiu, line 261: I don't see why this is right, and I don't see why it's needed for this optimization.
In general many tests will fail if you supply a NULL, but will pass if you cast the NULL to the correct type. So I suspect this is actually needed.

NobiGo: @mihaibudiu Yes, I will check again. After this PR, some unit test behavior is unexpected.

core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
NobiGo: This repair method is not comprehensive enough. I will check again.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -491,6 +491,28 @@ private static String toSql(RelNode root, SqlDialect dialect,
     relFn(relFn).ok(expected);
   }
 
+  @Test void testSelectWhereIn2() {
```
mihaibudiu, line 494: how are these tests related to the issue solved?

mihaibudiu: I am guessing that the IN list is converted to a union and it's converted back to a values
Perhaps you can add a comment to say that

core/src/main/java/org/apache/calcite/rel/rules/UnionToValuesRule.java
```diff
@@ -0,0 +1,149 @@
...
+import com.google.common.collect.ImmutableCollection;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+
+import org.immutables.value.Value;
+
+import java.util.List;
+
+/**
+ * Planner rule that convert
```
mihaibudiu, line 37: converts

core/src/main/java/org/apache/calcite/rel/rules/UnionToValuesRule.java
```diff
@@ -0,0 +1,149 @@
...
+ * {@link org.apache.calcite.rel.core.Union} with
+ * input only containing {@link org.apache.calcite.rel.core.Values}
```
mihaibudiu, line 39: inputs

core/src/main/java/org/apache/calcite/rel/rules/UnionToValuesRule.java
```diff
@@ -0,0 +1,149 @@
...
+ * <p>becomes
+ *
+ * <blockquote><pre>{@code
+ * Values(tuples=[[{ 3, null }, { 7369, null }, { 1, 2 }]])
+ * }</pre></blockquote>
+ *
+ * <p>If (<code>all</code> = <code>false</code>),
+ * {@link org.apache.calcite.rel.core.Values} value will be set list.
+ *
+ * <p>If {@link org.apache.calcite.rel.core.Union} only one input, will be handled by
```
mihaibudiu, line 60: I am not sure this documents the current rule. It is rather some advice to use the other rule to handle this case.

core/src/main/java/org/apache/calcite/rel/rules/UnionToValuesRule.java
```diff
@@ -0,0 +1,149 @@
...
+    call.transformTo(values);
+  }
+
+  private static ImmutableCollection.Builder<ImmutableList<RexLiteral>> getImmutableListBuilder(
+      Boolean all, List<RelNode> relNodeList) {
+    ImmutableCollection.Builder<ImmutableList<RexLiteral>> tupleList;
+    if (all) {
+      tupleList = ImmutableList.builder();
+    } else {
+      tupleList = ImmutableSet.builder();
```
mihaibudiu, line 101: What is concerning here is that two VALUES may have the same "value" but not be "equal".
In particular, I am not sure about implicit casts.
Perhaps you can add some tests like `VALUE(5.0) UNION ALL VALUES(5)` or
`VALUES(CAST 5.0 AS INT) UNION ALL VALUES(5)`

core/src/main/java/org/apache/calcite/rel/rules/UnionToValuesRule.java
NobiGo: I will remove this document. Because UNION_REMOVE only can handle `all = false`. But this rule can handle both when its input is `Values`.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
NobiGo: I had the same idea before. I added this test case because I want to know if RelBuilder generates the UNION when IN-List has CAST. But It doesn't.

core/src/main/java/org/apache/calcite/rel/rules/UnionToValuesRule.java
NobiGo: Maybe we will find another issue with `VALUE(5.0) UNION ALL VALUES(5)`. 
Here is a test case.
```
!use post
select *
from (values (5.0)) as t(x)
union
select *
from (values (5)) as t(y);
X
5
5.0
!ok

EnumerableUnion(all=[false])
  EnumerableValues(tuples=[[{ 5.0 }]])
  EnumerableValues(tuples=[[{ 5 }]])
!plan

select *
from (values (cast(5.0 as int))) as t(x)
union
select *
from (values (5)) as t(y);
X
5
!ok

EnumerableUnion(all=[false])
  EnumerableValues(tuples=[[{ 5 }]])
  EnumerableValues(tuples=[[{ 5 }]])
!plan
```
But in Postgresql. The SQL will return 5.0:
```
select *
from (values (5.0)) as t(x)
union
select *
from (values (5)) as t(y);
```

mihaibudiu: Calcite's pervasive use of implicit casts makes solving such problems much harder than expected. Other compilers have a stage which converts implicit casts into explicit casts in the program representation, which is then always unambiguously typed. I think SqlToRel should do that, but that seems like a very difficult task.

NobiGo: I agree with your opinion. In Calcite, implicit cast transformations are in different stages according to different operations. In this PR, I have kept this result `5.0 5`. We can move to another issue to resolve this. Now I'm not sure how big the impact of the issue is.

mihaibudiu: The problem is that your transformation will no longer produce an equivalent program. If the UNION removes duplicates, and after transformation you have just VALUES with duplicates, the two programs are no longer computing the same result.

mihaibudiu: you could check to see if the type of all VALUES is the same, and then you can still apply the transformation as it is, otherwise you leave the UNION in the program. The transformation will probably be correct for UNION ALL.

NobiGo: I added type validation according to your suggestion, and it seems that the result is correct.

mihaibudiu (approved):

PR #3890, opened 2024-07-29
Add Some testcast and testDecimalProject for the Arrow adapter

caicancai:
part of [6305](https://issues.apache.org/jira/browse/CALCITE-6305)

### DIFF
--- a/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
+++ b/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
@@ -518,6 +518,56 @@ class ArrowAdapterTest {
         .explainContains(plan);
   }
 
+  @Test void testDecimalProject() {
+    String sql = "select SAL from EMP";
+    String plan = "PLAN=ArrowToEnumerableConverter\n"
+        + "  ArrowProject(SAL=[$5])\n"
+        + "    ArrowTableScan(table=[[ARROW, EMP]], fields=[[0, 1, 2, 3, 4, 5, 6, 7]])\n\n";
+    String result = "SAL=800.00\n"
+        + "SAL=1600.00\n"
+        + "SAL=1250.00\n";
+    CalciteAssert.that()
+        .with(arrow)
+        .query(sql)
+        .limit(3)
+        .returns(result)
+        .explainContains(plan);
+  }
+
+  @Test void testCastDecimalToSmallInt() {
+    String sql = "select CAST(LOSAL AS SMALLINT) from SALGRADE";
+    String plan =
+        "PLAN=EnumerableCalc(expr#0..2=[{inputs}], expr#3=[CAST($t1):SMALLINT], EXPR$0=[$t3])\n"
+            + "  ArrowToEnumerableConverter\n"
+            + "    ArrowTableScan(table=[[ARROW, SALGRADE]], fields=[[0, 1, 2]])\n\n";
+    String result = "EXPR$0=700\n";
+
+    CalciteAssert.that()
+        .with(arrow)
+        .query(sql)
+        .typeIs("[EXPR$0 SMALLINT]")
+        .limit(1)
+        .returns(result)
+        .explainContains(plan);
+  }
+
+  @Test void testCastDecimalToTinyInt() {
+    String sql = "select CAST(\"intField\" AS TINYINT) from arrowdata";
+    String plan =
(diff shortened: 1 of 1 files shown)


caicancai:
Through some testing I found some bugs, I will open a jira case to discuss

caicancai:
@mihaibudiu PATL, thanks. There may be some bugs in other tests, I will open a jira case to fix them

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
```diff
@@ -518,6 +518,56 @@ static void initializeArrowState(@TempDir Path sharedTempDir) throws IOException
...
+        + "SAL=1250.00\n";
+    CalciteAssert.that()
+        .with(arrow)
+        .query(sql)
+        .limit(3)
+        .returns(result)
+        .explainContains(plan);
+  }
+
+  @Test void testCastDecimalToSmallInt() {
```
mihaibudiu, line 537: If you really want to test the functionality of the cast you should test a case where the value is too large for the target type.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
```diff
@@ -518,6 +518,56 @@ static void initializeArrowState(@TempDir Path sharedTempDir) throws IOException
...
+    CalciteAssert.that()
+        .with(arrow)
+        .query(sql)
+        .typeIs("[EXPR$0 SMALLINT]")
+        .limit(1)
+        .returns(result)
+        .explainContains(plan);
+  }
+
+  @Test void testCastDecimalToTinyInt() {
```
NobiGo, line 554: According to the generated plan, the CAST function can't push down to Arrow. So this is a test of the EnumerableCalc successfully executing the CAST function.  Is there any other reason why we need to cover these tests here？

NobiGo (changes requested):

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
caicancai: I noticed that there are similar tests, such as testCastDecimalToInt. I don't know how this test is related to pushdown. I would be grateful if you can answer my doubts.

NobiGo: Hi, @caicancai  AFAIK, `EnumerableCalc(expr#0..2=[{inputs}], expr#3=[CAST($t1):SMALLINT], EXPR$0=[$t3])` means we execute the CAST by Enumerable. If we can push down the CAST, the top plan will be `ArrowToEnumerableConverter`.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
caicancai: This will fail. I tried it. I think I should think about how to show this failure case in the test.

NobiGo (commented):
LGTM

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
caicancai:
```
  @Test void testCastDecimalToTinyInt() {
    String sql = "select CAST(LOSAL AS TINYINT) from SALGRADE";
    String plan =
        "PLAN=EnumerableCalc(expr#0..2=[{inputs}], expr#3=[CAST($t1):TINYINT], EXPR$0=[$t3])\n"
            + "  ArrowToEnumerableConverter\n"
            + "    ArrowTableScan(table=[[ARROW, SALGRADE]], fields=[[0, 1, 2]])\n\n";
    String result = "EXPR$0=700\n";

    CalciteAssert.that()
        .with(arrow)
        .query(sql)
        .typeIs("[EXPR$0 TINYINT]")
        .limit(1)
        .returns(result)
        .explainContains(plan);
  }
```
This test will fail, but I don't know how to describe it with a test
The arrow adapter still has a lot of problems, I'm not sure if adding these tests will be useful
I borrowed from: https://github.com/apache/datafusion-comet

mihaibudiu: You are testing a feature which does not really work.
Is there an issue for the bug? If there is you can add the negative test and disable it until the bug is fixed.

PR #3891, opened 2024-07-29
Fixed JSON_OBJECT issue

kenstott:
Redshift has no JSON_OBJECT function, this rewrites JSON_OBJECT using ANSI SQL.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/RedshiftSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/RedshiftSqlDialect.java
@@ -20,12 +20,15 @@ import org.apache.calcite.avatica.util.Casing;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
 import org.apache.calcite.rel.type.RelDataTypeSystemImpl;
-import org.apache.calcite.sql.*;
+import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlUserDefinedTypeNameSpec;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlDataTypeSpec;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.type.SqlTypeName;
-
-import org.apache.calcite.sql.util.SqlString;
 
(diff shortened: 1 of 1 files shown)


caicancai:
You need to create a jira case

kenstott:
Requested account. Account will be under ***@***.*** ***@***.***>”.

kenstott:
[CALCITE-6505](https://issues.apache.org/jira/browse/CALCITE-6505) has been created.

core/src/main/java/org/apache/calcite/sql/dialect/RedshiftSqlDialect.java
```diff
@@ -118,4 +117,27 @@ public RedshiftSqlDialect(Context context) {
...
+    if (call.getOperator() == SqlStdOperatorTable.JSON_OBJECT) {
+      assert call.operandCount() % 2 == 1;
+
+      SqlWriter.Frame frame = writer.startFunCall("");
+      SqlWriter.Frame listFrame = writer.startList("'{' || ", " || '}'");
+      for(int i = 1; i < call.operandCount(); i += 2) {
+        String keyName = call.operand(i).toSqlString(this).toString();
+        String value = call.operand(i + 1).toSqlString(this).toString();
+        keyName = keyName.substring(1, keyName.length() - 1);
+        String jsonFragment = String.format("'\"%s\": ' || COALESCE(CASE WHEN %s::varchar ~'^-?\\\\d*(\\\\.\\\\d+)?$' THEN %s::varchar ELSE quote_ident(%s) END, 'null')", keyName, value, value, value);
```
mihaibudiu, line 130: this deserves an explanation as a comment

core/src/main/java/org/apache/calcite/sql/dialect/RedshiftSqlDialect.java
```diff
@@ -20,14 +20,13 @@
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
 import org.apache.calcite.rel.type.RelDataTypeSystemImpl;
-import org.apache.calcite.sql.SqlDataTypeSpec;
-import org.apache.calcite.sql.SqlDialect;
-import org.apache.calcite.sql.SqlNode;
-import org.apache.calcite.sql.SqlUserDefinedTypeNameSpec;
-import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.*;
```
mihaibudiu, line 23: Calcite does not like * imports.

mihaibudiu (changes requested):
you need to write at least one test to exercise this code.
you also need to use the same title for the commit message and PR as you used in the JIRA case.

PR #3896, opened 2024-08-01
[CALCITE-6509] MongoAdapter throws 'Invalid JSON Number' exception if first character of field name is number

dssysolyatin:
… first character of field name is number

### DIFF
--- a/mongodb/src/main/java/org/apache/calcite/adapter/mongodb/MongoRules.java
+++ b/mongodb/src/main/java/org/apache/calcite/adapter/mongodb/MongoRules.java
@@ -103,25 +103,13 @@ public class MongoRules {
   }
 
   static String maybeQuote(String s) {
-    if (!needsQuote(s)) {
-      return s;
-    }
+    // MongoDB Extended JSON conforms to the JSON RFC. It is safe to use double quotes for
+    // everything.
     return quote(s);
   }
 
   static String quote(String s) {
-    return "'" + s + "'"; // TODO: handle embedded quotes
-  }
-
-  private static boolean needsQuote(String s) {
-    for (int i = 0, n = s.length(); i < n; i++) {
-      char c = s.charAt(i);
-      if (!Character.isJavaIdentifierPart(c)
-          || c == '$') {
-        return true;
-      }
-    }
-    return false;
+    return "\"" + s.replace("\"", "\\\"") + "\"";
   }
 
   /** Translator from {@link RexNode} to strings in MongoDB's expression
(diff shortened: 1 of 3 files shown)


mongodb/src/main/java/org/apache/calcite/adapter/mongodb/MongoRules.java
```diff
@@ -114,10 +114,14 @@ static String quote(String s) {
   }
 
   private static boolean needsQuote(String s) {
-    for (int i = 0, n = s.length(); i < n; i++) {
+    if (!s.isEmpty()
```
NobiGo, line 117: The field name can contain '$' ? The new code doesn't cover when '$' is in the middle of the field like 'te$st'.

NobiGo (changes requested):

mongodb/src/main/java/org/apache/calcite/adapter/mongodb/MongoRules.java
dssysolyatin: It is valid to use $ in the middle of an identifier. It is invalid only when it is at the beginning of the string.
```
db.zip.aggregate({ $project: { te$st: "$zip_code" } })
```

NobiGo: Yes. Looks good. Do you think we should add this case in a unit test?

dssysolyatin: I don't think so, because $ in the middle will work in any case, whether quoted or unquoted. Instead of writing all possible tests for quoting, it would be better to rewrite the Mongo adapter to use MongoDB builders. This way, we can avoid such mistakes.

See: https://issues.apache.org/jira/browse/CALCITE-6510

NobiGo: The newly modified code and test cases do not cover the original code logic. Testing coverage is not enough for this PR

NobiGo: Based on your reply, I know it works for me. But only focus on the code, I don't know.

dssysolyatin: Probably the best option would be to find a specification for fieldName and add it as a comment. I will try to find it tomorrow. Usually, people don't write tests to check if letters 'e', 'a', or 'b' are allowed in the middle of a string because the specification states that they are allowed.

dssysolyatin: @NobiGo, could you please recheck the PR? Instead of fixing individual quoting cases, I decided to quote every identifier because MongoDB Extended JSON conforms to the JSON RFC. Therefore, it is safe to use double quotes.

dssysolyatin: @NobiGo What are your thoughts, should we merge this PR, or just close it since other folks don't seem too interested ?

PR #3897, opened 2024-08-01
[CALCITE-6480] OracleDialect does not support CASE WHEN returning boo…

suibianwanwank:


core/src/main/java/org/apache/calcite/sql/SqlDialect.java
```diff
@@ -784,6 +784,11 @@ public boolean supportsWindowFunctions() {
     return true;
   }
 
+  /** Returns whether this dialect supports case when return boolean type. */
+  public boolean supportBooleanCaseWhen() {
```
mihaibudiu, line 788: does the old Oracle dialect support Booleans in other expressions?

suibianwanwank: What I'm trying to convey is that CASE WHEN statements are used as predicates in filters or concatenation conditions and return a value of bool.  could you make some suggestions for better naming of the method? 
edit: ... sry, I repeated the edit multiple times

mihaibudiu: Is this about CASE expressions or about the BOOLEAN data type not being supported?
Or is it about CASE expressions that evaluate to BOOLEAN results not being supported - which is what you imply?

suibianwanwank: It is the latter，about CASE expressions that evaluate to BOOLEAN results not being supported

mihaibudiu: In that case the name of the function is adequate.
Just to make sure: you say that an expression such as "TRUE" would work in a WHERE expression?

suibianwanwank: Did you say "true" literal?， I have supported in [CALCITE-6482](https://github.com/apache/calcite/pull/3863)

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -726,6 +726,31 @@ public SqlNode toSql(@Nullable RexProgram program, RexNode rex) {
...
         elseNode = caseNodeList.get(caseNodeList.size() - 1);
+
+        if (caseCall.getType().getSqlTypeName() == SqlTypeName.BOOLEAN
+            && !dialect.supportBooleanCaseWhen()) {
+          // Transformed when expressions of boolean type in SqlCase
+          // For example, given
+          //     CASE WHEN x > 1 THEN y > 1 ELSE y < 10 END
+          // Transformed:
+          //     CASE WHEN x > 1 THEN CASE WHEN y > 1 THEN 1 ELSE 0 END
+          // ELSE CASE WHEN y < 10 THEN 1 ELSE 0 END END = 1
```
mihaibudiu, line 737: the comment looks wrong, is the last comparison supposed to be there?

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
mihaibudiu: can you add a test that combines the two, e.g., `CASE WHEN x > 1 THEN TRUE ...`?

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
suibianwanwank: What's wrong with this, This guarantees the semantics of the bool expression
Regarding the tests expected in the comments below, Perhaps query1 in the test fulfills these, and I used true literal in the else expression

mihaibudiu: Can you use some parentheses to make the order of evaluation clearer?
It's not obvious how the expression is parsed. 
I guess it's parsed as (CASE ... END) = 1
Now I am worried about a CASE expression that mixes boolean and non-boolean values.
How about something like `(CASE WHEN ... x > 1 ELSE 2 END)`?
Is the translation of this also preserving its semantics?

suibianwanwank: good idea, i have added this part about parentheses.
In Oracle, would there be a difference between then and else expression statements, also in the default implementation of calcite this would throw the error "Illegal mixing of types in CASE or COALESCE statement"

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -726,6 +726,31 @@ public SqlNode toSql(@Nullable RexProgram program, RexNode rex) {
           }
         }
         elseNode = caseNodeList.get(caseNodeList.size() - 1);
+
+        if (caseCall.getType().getSqlTypeName() == SqlTypeName.BOOLEAN
```
NobiGo, line 730: For other databases, this is a special implementation. Can we rewrite CASE WHEN in OracleSqlDialect? Maybe `OracleSqlDialect.unparseCall` is a good place.

NobiGo (changes requested):

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
suibianwanwank: I wanted to design it this way at first, but I couldn't get the return type of SqlCase in unparseCall, and this conversion has to ensure that the return type is a boolean.

NobiGo: +1

NobiGo (approved):

mihaibudiu:
You can squash the commits for merging.

suibianwanwank:
Done ✅

PR #3899, opened 2024-08-02
[CALCITE-6506] Type inference for IN list is incorrect

NobiGo:


core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
```diff
@@ -330,6 +330,10 @@ public static boolean containsNullable(RelDataType type) {
   public static RelDataType keepSourceTypeAndTargetNullability(RelDataType sourceRelDataType,
                                              RelDataType targetRelDataType,
                                              RelDataTypeFactory typeFactory) {
+    checkArgument(
+        (targetRelDataType.isStruct() && sourceRelDataType.isStruct())
+            || (!targetRelDataType.isStruct() && !sourceRelDataType.isStruct()),
+        "Input sourceRelDataType and targetRelDataType must have same struct type");
```
mihaibudiu, line 336: I think that a better message is "one is a struct, while the other one is not".
"same struct type" implies that the type must be exactly the same.

core/src/test/resources/sql/sub-query.iq
```diff
@@ -3812,4 +3812,242 @@ WHERE s1.total > (SELECT avg(total) FROM agg_sal s2 WHERE s1.deptno = s2.deptno)
 
 !ok
 
+# [CALCITE-6506] Type inference for IN list is incorrect
+
+# Test LHS is no nullable and RHS is no nullable
```
mihaibudiu, line 3817: "not nullable" sounds better

core/src/test/resources/sql/sub-query.iq
```diff
@@ -3812,4 +3812,242 @@ WHERE s1.total > (SELECT avg(total) FROM agg_sal s2 WHERE s1.deptno = s2.deptno)
...
+|  7900 | false  |
+|  7902 | false  |
+|  7934 | false  |
++-------+--------+
+(14 rows)
+
+!ok
+
+
+EnumerableCalc(expr#0..5=[{inputs}], expr#6=[IS NOT NULL($t5)], expr#7=[0:BIGINT], expr#8=[<>($t1, $t7)], expr#9=[AND($t6, $t8)], expr#10=[<($t2, $t1)], expr#11=[null:BOOLEAN], expr#12=[IS NULL($t5)], expr#13=[AND($t10, $t11, $t8, $t12)], expr#14=[OR($t9, $t13)], expr#15=[CAST($t14):BOOLEAN NOT NULL], EMPNO=[$t0], EXPR$1=[$t15])
```
mihaibudiu, line 3842: this is a very complicated plan!
you would think a simpler one could be generated using SEARCH

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql/type/SqlTypeUtil.java
NobiGo: It has been modified as recommended.

core/src/test/resources/sql/sub-query.iq
NobiGo: It has been modified as recommended.

core/src/test/resources/sql/sub-query.iq
NobiGo: Yes. The RHS is (nullable, nullable) which makes this plan complicated. If we can optimize this EnumerableCalc using SEARCH, I will check this.

mihaibudiu:
I had already approved, since the changes suggested were small

mihaibudiu (approved):

PR #3900, opened 2024-08-02
[CALCITE-6513] FilterProjectTransposeRule may cause OOM when Project expressions are complex

kasakrisz:
Add a similar bloat config to `FilterProjectTransposeRule` that `ProjectMergeRule` has. Do not push down the filter If the resulting predicate expression complexity exceeds the bloat value during pushdown.


core/src/main/java/org/apache/calcite/rel/rules/FilterProjectTransposeRule.java
```diff
@@ -258,5 +261,14 @@ default Config withOperandFor(Class<? extends Filter> filterClass,
                   b2.operand(relClass).anyInputs())))
           .as(Config.class);
     }
+
+    /** Limit how much complexity can increase during merging.
+     * Default is {@link RelOptUtil#DEFAULT_BLOAT} (100). */
```
asolimando, line 266: Nit: Probably better to omit the value itself as one can easily look it up with the IDE. The risk I see is that, in case the value gets changed in the code, we end up with an inconsistency.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -2699,6 +2699,40 @@ private void checkProjectCorrelateTransposeRuleSemiOrAntiCorrelate(JoinRelType t
...
+        CoreRules.FILTER_PROJECT_TRANSPOSE.config
+            .withOperandSupplier(b0 ->
+                b0.operand(Filter.class).predicate(filter -> true)
+                    .oneInput(b1 ->
+                        b1.operand(Project.class).predicate(project -> true)
+                            .anyInputs()))
+            .as(FilterProjectTransposeRule.Config.class)
+            .withCopyFilter(true)
+            .withCopyProject(true)
+            .withBloat(10)
```
asolimando, line 2727: Can we use a smaller value and make the unit test code smaller? It's clear what you want to show but it's a bit hard to follow in the current form IMO.

asolimando (commented):
@kasakrisz, LGTM, left few (minor) comments.

core/src/main/java/org/apache/calcite/rel/rules/FilterProjectTransposeRule.java
kasakrisz: Removed

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
kasakrisz: It is not a simple change: whether to merge `Projects` or not is controlled by `RelBuilder.Config.bloat()`, and the tests in `RelOptRulesTest` using `FnRelSupplier` create `RelBuilder` instances using a [constant config](https://github.com/apache/calcite/blob/979981481f8effde5e2fe5bb605d5c3424554e4b/testkit/src/main/java/org/apache/calcite/test/RelSupplier.java#L131).

I tried to improve `RelSupplier` and `RelOptFixture.withRelBuilderConfig` to make `RelBuilder` configurable when `FnRelSupplier` is used. See: https://github.com/apache/calcite/pull/3900/commits/b8d1dce88672746e4f10925ef3298921c39c38c6

Please let me know If I'm on the right track to tackle this.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -8177,7 +8195,8 @@ LogicalProject(DEPTNO=[$0])
   LogicalJoin(condition=[true], joinType=[inner])
     LogicalTableScan(table=[[scott, DEPT]])
     LogicalAggregate(group=[{}], DUMMY_COUNT=[COUNT()])
-      LogicalTableScan(table=[[scott, EMP]])
+      LogicalProject(DUMMY=[0])
```
asolimando, line 8198: Do you know why we have this extra `Project` now?

asolimando (commented):
@kasakrisz, thanks for tackling the issue, it looks like an improvement to have more flexibility in the `RelBuilder` configuration in tests, I have stumbled on this too in the past and it has been extremely cumbersome to alter the bloat value.

The unit-test is also way simpler and easier to understand.

I have just one question around an extra `Project`, the rest LGTM.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
kasakrisz: Yes. Different `RelBuilder` configuration instances are used at different phase of the test. 
[RelBuilder.Config.DEFAULT](https://github.com/apache/calcite/blob/979981481f8effde5e2fe5bb605d5c3424554e4b/core/src/main/java/org/apache/calcite/tools/RelBuilder.java#L4925-L4927) vs [RelOptFixture.DEFAULT](https://github.com/apache/calcite/blob/979981481f8effde5e2fe5bb605d5c3424554e4b/testkit/src/main/java/org/apache/calcite/test/RelOptFixture.java#L88)
`pruneInputOfAggregate` has different values in the two instances. It would be good to use only one instance in the whole test but I haven't been able to achieve this so far.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
kasakrisz: I managed to write a test case using Sql, so no need to change `RelSupplier` and `RelOptFixture`.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
kasakrisz: Reverted this change.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -2699,6 +2699,35 @@ private void checkProjectCorrelateTransposeRuleSemiOrAntiCorrelate(JoinRelType t
         .check();
   }
 
+  /** Tests a variant of {@link FilterProjectTransposeRule}
+   * that pushes a Filter that contains a correlating variable. */
```
asolimando, line 2703: Is `x1` a correlating variable?

asolimando (approved):
LGTM, I think the comment should be fixed but the rest seems good now

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
julianhyde: I think not. @kasakrisz, can you correct the javadoc.

kasakrisz: I copy-pasted an existing test hence the misleading comment. Fixed.

asolimando:
@kasakrisz, there seems to be no further comments, please squash the commits and I will merge the PR, thanks!

asolimando (approved):

asolimando (approved):

PR #3901, opened 2024-08-02
CALCITE-6465: Rework code generator to use Flink code splitter

jduo:


jduo:
This builds but introduces a dependency on flink-core. Looking for a way around this.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/MethodDeclaration.java
```diff
@@ -59,20 +61,38 @@ public MethodDeclaration(int modifier, String name, Type resultType,
...
+    // From Flink TableConfigOptions.MAX_LENGTH_GENERATED_CODE
+    // Note that Flink uses 4000 rather than 64KB explicitly:
+    // "Specifies a threshold where generated code will be split into sub-function calls.
+    //  Java has a maximum method length of 64 KB. This setting allows for finer granularity if
+    //  necessary.
+    //  Default value is 4000 instead of 64KB as by default JIT refuses to work on methods with
+    //  more than 8K byte code."
+    final int flinkDefaultMaxGeneratedCodeLength = 4000;
+
+    // From Flink TableConfigOptions.MAX_MEMBERS_GENERATED_CODE
```
davidradl, line 89: these Flink considerations seem in appropriate in Calcite, unless we can make them generic so we do not reference Flink at all.

jduo: I've removed references to Flink here.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/AbstractNode.java
```diff
@@ -51,7 +51,7 @@ public Type getType() {
   }
 
   @Override public String toString() {
-    ExpressionWriter writer = new ExpressionWriter(true);
+    ExpressionWriter writer = new ExpressionWriter(true, true);
```
jduo, line 54: Ideally this would read from the CalciteSystemProperty but calcite-core depends on linq4j so it can't be looked up here.

rubenada: Will this have any practical impact on the actual code generation? (like splitting in some circumstances where the CalciteSystemProperty has not been set to true). If not, I'd say this is acceptable.
It's true that it is a bit unfortunate that we cannot check the CalciteSystemProperty here, but I can't think of a better solution at this point.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
```diff
@@ -58,8 +58,8 @@ private Expressions() {}
    * extra type information in generics.
    */
   public static String toString(List<? extends Node> expressions, String sep,
-      boolean generics) {
-    final ExpressionWriter writer = new ExpressionWriter(generics);
+      boolean generics, boolean methodSplit) {
```
rubenada, line 61: Probably the javadoc of this method should be updated to mention the new "optional split"

linq4j/src/main/java/org/apache/calcite/linq4j/tree/MethodDeclaration.java
```diff
@@ -57,20 +60,39 @@ public MethodDeclaration(int modifier, String name, Type resultType,
   }
 
   @Override public void accept(ExpressionWriter writer) {
-    String modifiers = Modifier.toString(modifier);
-    writer.append(modifiers);
+    final ExpressionWriter writerForUnsplitMethod = writer.usesMethodSplitting()
```
rubenada, line 63: I have the impression the intent of this line is not very obvious, perhaps a comment would help to clarify it.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
```diff
@@ -70,8 +70,8 @@ public static String toString(List<? extends Node> expressions, String sep,
   /**
    * Converts an expression to Java source code.
    */
-  public static String toString(Node expression) {
-    return toString(Collections.singletonList(expression), "", true);
+  public static String toString(Node expression, boolean methodSplit) {
```
rubenada, line 73: Since we are touching public methods of a public class, to facilitate backwards compatibility I wonder if it would make sense to overload these `toString` methods (to have a "default false methodSplit", in case any downstream project was using them directly):
```
public static String toString(List<? extends Node> expressions, String sep,
      boolean generics) {
  return toString(expressions, sep, generics, false);
}

public static String toString(Node expression) {
  return toString(expression, false);
}

```

linq4j/src/main/java/org/apache/calcite/linq4j/tree/AbstractNode.java
davidradl: > Ideally this would read from the CalciteSystemProperty but calcite-core depends on linq4j so it can't be looked up here.

I do not know this code, but wonder if the split configuration values be passed by value into the linq4j layer. If possible , this would avoid the dependancy recursion.

jduo: I don't think this toString() method is used in code generation, just logging/debugging.
This option is passed in to Expressions static functions, which is really where ExpressionWriters are instantiated most of the time.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
jduo: Done in [dc1c66d](https://github.com/apache/calcite/pull/3901/commits/dc1c66d14b38ad122312b99861fca20794701504)

linq4j/src/main/java/org/apache/calcite/linq4j/tree/MethodDeclaration.java
jduo: Done in [dc1c66d](https://github.com/apache/calcite/pull/3901/commits/dc1c66d14b38ad122312b99861fca20794701504)

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
jduo: Done in [dc1c66d](https://github.com/apache/calcite/pull/3901/commits/dc1c66d14b38ad122312b99861fca20794701504)

linq4j/src/main/java/org/apache/calcite/linq4j/tree/MethodDeclaration.java
rubenada: Thanks @jduo , just a small comment, because I find the comment a bit confusing (maybe I'm getting it wrong), where it says:
```
... serialize the method declaration directly to the supplied writer if method splitting is enabled or ...
```
shouldn't it say "disabled" ?

jduo: Yes that's correct. I've updated this comment, but note that it is now in ClassDeclaration instead of MethodDeclaration. The method splitter requires an enclosing class to write some of its output too (it can extract variable names to instance variables and move blocks of code to generated private methods to reduce the size of a method).

core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
```diff
@@ -420,6 +420,18 @@ public final class CalciteSystemProperty<T> {
...
+  /**
+   * The length of code before automatic method splitting is enabled
+   *
+   * <p>A value of <code>0</code>, the default, disables automatic method splitting.
+   *
+   * <p>Some queries can generate methods exceeding the JVM limit of 4000 characters per method.
+   * Use this setting to automatically detect and split methods larger than the specified
+   * limit.
+   */
+  public static final CalciteSystemProperty<Integer> METHOD_SPLITTING_THRESHOLD =
```
rubenada, line 432: nit: proposal (to make the new property robust in case of negative values)
```
public static final CalciteSystemProperty<Integer> METHOD_SPLITTING_THRESHOLD =
      intProperty("calcite.linq.method_splitting_threshold", 0, v -> v >= 0);
```

linq4j/src/test/java/org/apache/calcite/linq4j/test/ExpressionTest.java
```diff
@@ -1719,6 +1724,60 @@ public void checkBlockBuilder(boolean optimizing, String expected) {
...
+        null,
+        ImmutableList.of(),
+        ImmutableList.of(hugeMethod));
+
+    final String originalClass = Expressions.toString(classDecl);
+    final String classWithMethodSplitting = Expressions.toString(classDecl, 4000);
+
+    // Some local variables in hugeMethod have been extracted to be fields named local$<number> in
+    // the enclosing class definition.
+    Assertions.assertFalse(originalClass.contains("local$"));
```
julianhyde, line 1777: Use `assertThat` rather than `assertFalse` and `assertTrue`.

linq4j/src/test/java/org/apache/calcite/linq4j/test/ExpressionTest.java
```diff
@@ -1719,6 +1724,60 @@ public void checkBlockBuilder(boolean optimizing, String expected) {
         Expressions.toString(Expressions.constant(set)));
   }
 
+  @Test void testLargeMethod() {
```
julianhyde, line 1727: I believe that there is only one test in the whole of calcite that sues method-splitting.

Need a test where splitting is set to a low positive value, e.g. 3. Just to make sure that the splitter handles gracefully the case where it's impossible to split.

Is it possible to also use method-splitting in one or two of Calcite's system tests?

core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
```diff
@@ -420,6 +420,18 @@ public final class CalciteSystemProperty<T> {
   public static final CalciteSystemProperty<Integer> FUNCTION_LEVEL_CACHE_MAX_SIZE =
       intProperty("calcite.function.cache.maxSize", 0, v -> v >= 0);
 
+  /**
+   * The length of code before automatic method splitting is enabled
+   *
+   * <p>A value of <code>0</code>, the default, disables automatic method splitting.
+   *
+   * <p>Some queries can generate methods exceeding the JVM limit of 4000 characters per method.
```
julianhyde, line 428: Can you include a citation for your statement "JVM limit of 4000 characters per method"?

Java 7 had a limit [65534 bytes](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.3) (bytecode size). When did that change?

bom/build.gradle.kts
```diff
@@ -111,6 +111,7 @@ dependencies {
         apiv("org.apache.commons:commons-pool2")
         apiv("org.apache.commons:commons-collections4")
         apiv("org.apache.commons:commons-text")
+        apiv("org.apache.flink:flink-table-code-splitter")
```
julianhyde, line 114: Add a note above this list of declarations that we import flink-table-code-splitter but we must not import core flink modules.

innodb/src/main/java/org/apache/calcite/adapter/innodb/InnodbToEnumerableConverter.java
```diff
@@ -150,7 +150,9 @@ static List<String> innodbFieldNames(final RelDataType rowType) {
                 InnodbMethod.INNODB_QUERYABLE_QUERY.method, fields,
                 selectFields, cond, ascOrder));
     if (CalciteSystemProperty.DEBUG.value()) {
-      System.out.println("Innodb: " + Expressions.toString(enumerable));
+      System.out.println(
+          "Innodb: " + Expressions.toString(enumerable,
+          CalciteSystemProperty.METHOD_SPLITTING_THRESHOLD.value()));
```
julianhyde, line 155: indentation

linq4j/src/main/java/org/apache/calcite/linq4j/tree/ExpressionWriter.java
```diff
@@ -32,15 +32,25 @@ class ExpressionWriter {
...
+  private final int methodSplittingThreshold;
+  private boolean indentPending;
 
   ExpressionWriter() {
-    this(true);
+    this(true, 0);
   }
 
-  ExpressionWriter(boolean generics) {
+  ExpressionWriter(boolean generics, int methodSplittingThreshold) {
```
julianhyde, line 43: Need javadoc for this constructor.

You don't say the units of the threshold (bytes, lines), that it applies to source code length, what values disable splitting, and whether it's `>` or `>=`.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/ExpressionWriter.java
```diff
@@ -32,15 +32,25 @@ class ExpressionWriter {
...
     this.generics = generics;
+    this.methodSplittingThreshold = methodSplittingThreshold;
+  }
+
+  public ExpressionWriter duplicateState() {
```
julianhyde, line 48: public method needs javadoc

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
```diff
@@ -71,7 +81,15 @@ public static String toString(List<? extends Node> expressions, String sep,
    * Converts an expression to Java source code.
    */
   public static String toString(Node expression) {
-    return toString(Collections.singletonList(expression), "", true);
+    return toString(expression, 0);
+  }
+
+  /**
+   * Converts an expression to Java source code.
+   * Optionally splits the generated method if the method is too large.
```
julianhyde, line 89: "too large" is vague. can do better.

julianhyde (changes requested):

linq4j/src/main/java/org/apache/calcite/linq4j/tree/AbstractNode.java
```diff
@@ -51,7 +51,8 @@ public Type getType() {
   }
 
   @Override public String toString() {
-    ExpressionWriter writer = new ExpressionWriter(true);
+    // Use the JVM limit for method size (4K) as the method splitting threshold.
+    ExpressionWriter writer = new ExpressionWriter(true, 4000);
```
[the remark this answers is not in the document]
jduo: Done.

core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
```diff
@@ -420,6 +420,18 @@ public final class CalciteSystemProperty<T> {
   public static final CalciteSystemProperty<Integer> FUNCTION_LEVEL_CACHE_MAX_SIZE =
       intProperty("calcite.function.cache.maxSize", 0, v -> v >= 0);
 
+  /**
+   * The length of code before automatic method splitting is enabled
```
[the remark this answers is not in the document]
jduo: I've renamed this to be MAX_METHOD_LENGTH_IN_CHARS_BEFORE_SPLITTING and clarified that splitting starts happening if the code is longer than the limit and not when equal to the limit. This is a long name though so I'm open to better names.

innodb/src/main/java/org/apache/calcite/adapter/innodb/InnodbToEnumerableConverter.java
jduo: Done.

core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
jduo: I got this 4K from Flink's documentation on the configuration option relating to this:
https://github.com/apache/flink/blob/4101316ef37d3a7082bff57cb732f4e9e0349d09/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/TableConfigOptions.java#L225

So it's not exactly that the JVM has a limit of 4K, it's that the default JIT allows 8K only. Let me find a source for this.

core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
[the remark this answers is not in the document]
jduo: On the topic of the constant to indicate no limit, I've changed this to be -1 instead of 0. I don't see a good place to hold the magic number though -- CalciteSystemProperty would be a reasonable user-facing location, but ExpressionWriter can't reference it.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableInterpretable.java
```diff
@@ -111,7 +111,10 @@ public static Bindable toBindable(Map<String, Object> parameters,
             parameters);
 
     final ClassDeclaration expr = relImplementor.implementRoot(rel, prefer);
-    String s = Expressions.toString(expr.memberDeclarations, "\n", false);
+    String s =
+        Expressions.toString(expr.memberDeclarations,
```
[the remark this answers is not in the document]
jduo: Done.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Expressions.java
jduo: Changed this to be more specific about relating to the limit.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/ExpressionWriter.java
jduo: I made this package-private rather than adding javadoc -- it really seems like most of the methods in ExpressionWriter can be package-private.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/ExpressionWriter.java
jduo: Expanded comments and renamed this parameter.

bom/build.gradle.kts
jduo: Done.

core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
jduo: It's specified here in OpenJDK (other vendors use the same limit): https://github.com/openjdk/jdk11u-dev/blob/a754a3d8972abc67aee768d31a2dc2e8214274cf/src/hotspot/share/runtime/globals.hpp#L2391

It's considered a developer option though, rather than an end-user option, so documentation is limited.

linq4j/src/test/java/org/apache/calcite/linq4j/test/ExpressionTest.java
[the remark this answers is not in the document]
jduo: Clarified the meaning of the MyClass comment block and added to the header public javadoc about the test + JIRA.

linq4j/src/test/java/org/apache/calcite/linq4j/test/ExpressionTest.java
jduo: Done

core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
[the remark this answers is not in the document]
jduo: I made the -1 magic number a constant in Expressions

linq4j/src/test/java/org/apache/calcite/linq4j/test/ExpressionTest.java
jduo: I've added the test described above and also added a test to ensure method splitting does _not_ happen when methods remain under the limit.

I'm not familiar with Calcite's system tests. Would they be under the core module?

jduo:
@rubenada @julianhyde , just wondering if there was some more feedback on this and would like some clarification on where system tests reside within Calcite. Thanks.

rubenada:
@jduo , not sure what you mean by "system tests". If you want to add a test with a large dynamic code that fails when this feature is turn off / passes when it is turned on; maybe you could check and adapt the test that was done for the slightly related [CALCITE-3094](https://issues.apache.org/jira/browse/CALCITE-3094) (and then refined in https://issues.apache.org/jira/browse/CALCITE-6593): `LargeGeneratedJoinTest`. The query in that test, with the `JOIN_SELECTOR_COMPACT_CODE_THRESHOLD` disabled, should throw the "code grows beyond 64 KB InternalCompilerException". At that point, turning on the code splitter from this PR, should make the test pass.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableInterpretable.java
```diff
@@ -111,7 +111,9 @@ public static Bindable toBindable(Map<String, Object> parameters,
             parameters);
 
     final ClassDeclaration expr = relImplementor.implementRoot(rel, prefer);
-    String s = Expressions.toString(expr.memberDeclarations, "\n", false);
+    String s =
+        Expressions.toString(expr.memberDeclarations, "\n", false,
```
asdfgh19, line 115: It seems that code split is not effective here. This is because the code splitting logic in this pr is implemented in ClassDeclaration, and the expr.memberDeclarations here is usually MethodDeclaration.

core/src/main/java/org/apache/calcite/interpreter/JaninoRexCompiler.java
```diff
@@ -186,7 +186,9 @@ static Scalar.Producer baz(ParameterExpression context_,
     final ClassDeclaration classDeclaration =
         Expressions.classDecl(Modifier.PUBLIC, "Buzz", null,
             ImmutableList.of(Scalar.Producer.class), declarations);
-    String s = Expressions.toString(declarations, "\n", false);
+    String s =
+        Expressions.toString(declarations, "\n", false,
```
asdfgh19, line 190: The declarations here usually do not look like ClassDeclarations either.

core/src/main/java/org/apache/calcite/rex/RexExecutorImpl.java
```diff
@@ -100,7 +100,9 @@ private static String compile(RexBuilder rexBuilder, List<RexNode> constExps,
         Expressions.methodDecl(Modifier.PUBLIC, Object[].class,
             BuiltInMethod.FUNCTION1_APPLY.method.getName(),
             ImmutableList.of(root0_), blockBuilder.toBlock());
-    String code = Expressions.toString(methodDecl);
+    String code =
+        Expressions.toString(methodDecl,
```
asdfgh19, line 104: The methodDecl here is a MethodDeclaration, not a ClassDeclaration, so it seems that the code splitting logic implemented in the ClassDeclaration does not work here.

core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableInterpretable.java
jduo: @asdfgh19 , thanks for your comments. Would you be able to add tests for your scenarios to the PR?

For this one, it seems like we a can pass expr to Expressions.toString(), but I'm not sure what effect that has on the calls that depend on s (`` Hook.JAVA_PLAN.run(s);``)

core/src/main/java/org/apache/calcite/rex/RexExecutorImpl.java
jduo: It seems that this code relates to RexExecutable. The method body is generated here and written to String, then gets attached as the body of a class in RexExecutable that's run directly in Janino's IClassBodyEvaluator. This will be quite complicated to rework.

PR #3904, opened 2024-08-06
CALCITE-6433 SUBSTRING function can return incorrect empty result

zstan:


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1054,6 +1042,18 @@ static int clamp(long s) {
     return (int) s;
   }
 
+  /** SQL SUBSTRING(string FROM ...) function. */
```
mihaibudiu, line 1045: why move this function which seems unchanged?

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
```diff
@@ -172,6 +173,58 @@ static <E> List<E> list() {
...
+    assertThat(substring("string", Long.MIN_VALUE), is("string"));
+    assertThat(substring("string", Integer.MIN_VALUE + 10), is("string"));
+    assertThat(substring("string", Integer.MAX_VALUE), is(""));
+    assertThat(substring("string", Long.MAX_VALUE), is(""));
+    assertThat(substring("string", Integer.MAX_VALUE - 10), is(""));
+    assertThat(substring("string", Integer.MIN_VALUE - 10L), is("string"));
+    assertThat(substring("string", Integer.MAX_VALUE + 10L), is(""));
+
+    // str vs multi params
+    assertThat(substring("string", -1, 1), is(""));
```
mihaibudiu, line 192: is this behavior documented?
it's not obvious whether clamping happens first or the index computation happens first.

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
```diff
@@ -172,6 +173,58 @@ static <E> List<E> list() {
...
+
+    assertThat(substring("string", -1, Integer.MAX_VALUE), is("string"));
+    assertThat(substring("string", -1L, Integer.MAX_VALUE), is("string"));
+    assertThat(substring("string", -1, Long.MAX_VALUE), is("string"));
+
+    assertThat(substring("string", Integer.MIN_VALUE, Integer.MAX_VALUE), is(""));
+    assertThat(substring("string", Integer.MIN_VALUE, Integer.MAX_VALUE + 10L), is("string"));
+    assertThat(substring("string", Long.MIN_VALUE, Integer.MAX_VALUE), is(""));
+    assertThat(substring("string", Integer.MIN_VALUE, Long.MAX_VALUE), is("string"));
+    assertThat(substring("string", Integer.MIN_VALUE - 10L, Long.MAX_VALUE), is("string"));
```
mihaibudiu, line 225: how about a few tests with other numeric types, to make sure that conversions work.
I would add a tinyint, a decimal, a double. The decimal is particularly interesting.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1076,15 +1076,39 @@ public static String substring(String c, int s, int l) {
...
+      }
+    }
+    int l0 = clamp(l);
+    return SqlFunctions.substring(c, s, l0);
   }
 
   public static String substring(String c, long s, int l) {
-    return substring(c, clamp(s), l);
+    if (s < 0) {
+      long s0 = s + l;
```
mihaibudiu, line 1091: it makes it easier to review if you are consistent in your implementations
elsewhere you used l += s

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
zstan: this test is deal with concrete function implementations, you mean checkScalar like tests, isn`t it ?

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
zstan: this equals to PG, what behavior are you talking about ?

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
mihaibudiu: indeed.

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
mihaibudiu: As I said, one can imagine that negative s is first clamped to 0 and then the function is evaluated with 0 and an unchanged l.
Does the reference.md file document this behavior?
It's also not obvious to me that substring's behavior is defined by Postgres.
According to https://calcite.apache.org/docs/reference.html#character-string-operators-and-functions it is a standard function.
We have seen many cases where various SQL dialects treat corner cases differently for function implementations (but this may be out of scope for the current PR).

mihaibudiu:
In general I would recommend not squashing commits until the end.
Having additional commits makes it easier to review the changes in response to review comments.

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
zstan: done

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
zstan: done

zstan:
> In general I would recommend not squashing commits until the end. Having additional commits makes it easier to review the changes in response to review comments.

I read it a bit later ( ok i understand in future

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
zstan: i understand what you are talking about but i check sql standard, there is nothing about concrete start position definition specified... if i need to fix smth here - plz let me know.

mihaibudiu: The documentation is a bit sparse, but maybe you could improve it a bit describing what happens for negative values

zstan: Fixed, if it`s not enough or clear can you suggest or fix it yourself, i`m not a native english speaking guy, thanks

mihaibudiu:
The style checker is upset that your git commit message starts with a lowercase letter.
I guess you can squash the commits, and hopefully this will make the style checker happy too.

mihaibudiu (approved):

zstan:
> The style checker is upset that your git commit message starts with a lowercase letter. I guess you can squash the commits, and hopefully this will make the style checker happy too.

yes, i did it because you request not to use force push, seems i need to comment each consequence changes with the same comment as in initial.

mihaibudiu:
it doesn't have to be the same comment, just to obey some style rules

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1076,15 +1076,39 @@ public static String substring(String c, int s, int l) {
...
+        return substring(c, 0, l);
+      }
+    }
+    int l0 = clamp(l);
+    return SqlFunctions.substring(c, s, l0);
   }
 
   public static String substring(String c, long s, int l) {
-    return substring(c, clamp(s), l);
+    if (s < 0) {
```
caicancai, line 1090: `return s < 0 ? (s += l) > 0 ? substring(c, 0, s) : "" : /* your else case here */; `
Better？

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
```diff
@@ -172,6 +173,58 @@ static <E> List<E> list() {
     assertThat(concat(null, "b"), is("nullb"));
   }
 
+  @Test void testSubString() {
```
caicancai, line 176: Add jira link

zstan: done

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
zstan: no, it become less readable, i would prefer my current approach

caicancai (approved):

PR #3905, opened 2024-08-06
[CALCITE-6518] ClassCastException during validation when loading multiple libraries

mihaibudiu:
The PR is much more involved than expected, because it also solves another bug where functions declared in multiple libraries cannot be resolved by the validator. There were no tests using multiple libraries at the same time.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlFunction.java
@@ -310,7 +310,9 @@ public class SqlFunction extends SqlOperator {
           // if we succeed, the arguments would be wrapped with CAST operator.
           if (function != null) {
             TypeCoercion typeCoercion = validator.getTypeCoercion();
-            if (typeCoercion.userDefinedFunctionCoercion(scope, call, function)) {
+            if ((function.category == SqlFunctionCategory.USER_DEFINED_FUNCTION
+                || function.category == SqlFunctionCategory.USER_DEFINED_TABLE_FUNCTION)
+                && typeCoercion.userDefinedFunctionCoercion(scope, call, function)) {
               break validCoercionType;
             }
           }
--- a/core/src/main/java/org/apache/calcite/sql/SqlKind.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlKind.java
@@ -29,7 +29,7 @@ import java.util.Set;
  * Enumerates the possible types of {@link SqlNode}.
  *
  * <p>The values are immutable, canonical constants, so you can use Kinds to
- * find particular types of expressions quickly. To identity a call to a common
+ * find particular types of expressions quickly. To identify a call to a common
  * operator such as '=', use {@link org.apache.calcite.sql.SqlNode#isA}:
  *
  * <blockquote>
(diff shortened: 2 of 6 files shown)


caicancai:
This reminds me of the loading problem of SqlStdOperatorTable and SqlLibraryOperators. When an Operator exists in SqlStdOperatorTable, the same operator of SqlLibraryOperators will not take effect.

core/src/main/java/org/apache/calcite/sql/SqlFunction.java
```diff
@@ -310,7 +310,9 @@ argTypes, argNames, getFunctionType(), SqlSyntax.FUNCTION,
           // if we succeed, the arguments would be wrapped with CAST operator.
           if (function != null) {
             TypeCoercion typeCoercion = validator.getTypeCoercion();
-            if (typeCoercion.userDefinedFunctionCoercion(scope, call, function)) {
+            if ((function.category == SqlFunctionCategory.USER_DEFINED_FUNCTION
```
caicancai, line 313: I'm not really sure why these judgments are made. Maybe I can guess a little, but it may not be very accurate. Should JavaDoc be added?

mihaibudiu: JavaDoc comments can only be added to methods and classes. I suspect you are asking for regular Java comments.
Also, I am not sure what you mean by "judgments". I didn't make any judgments, I only attempted to fix a bug.

This is where the bug is located, in the absence of these checks the program fails with a class cast exception.
The implementation of the `userDefinedFunctionCoercion` already has a JavaDoc comment indicating that it is designed for user-defined functions.
And indeed, if I restrict the function to be invoked only on user-defined functions, as the inserted "if" does, the crash goes away.
I don't really know what additional comments to write which would clarify what this change does. I think the code is simple enough and the names are clear enough to describe what is going on.

caicancai: Thanks for your answer, I checked the code and it is indeed

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -29,7 +29,7 @@
  * Enumerates the possible types of {@link SqlNode}.
  *
  * <p>The values are immutable, canonical constants, so you can use Kinds to
- * find particular types of expressions quickly. To identity a call to a common
+ * find particular types of expressions quickly. To identify a call to a common
```
caicancai, line 32: good catch

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -2424,6 +2424,14 @@ private static void checkConcatFunc(SqlOperatorFixture f) {
     f.checkFails("^concat()^", INVALID_ARGUMENTS_NUMBER, false);
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6518">
+   * ClassCastException during validation when loading multiple libraries</a>. */
+  @Test void issue6518() {
```
caicancai, line 2429: I'm not sure issue6518 is appropriate, as there doesn't seem to be a similar test starting with that issue yet

caicancai (approved):
Overall LGTM

core/src/main/java/org/apache/calcite/sql/SqlKind.java
caicancai: @mihaibudiu I noticed a few more similar typos, would you mind fixing them?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: I will fix these and squash the commits at the same time, since this is a small PR. Hopefully we can merge it.

caicancai (approved):
LGTM

NobiGo:
After this PR, what is Calcite's behavior when loading multiple libraries if there are identical function names?

mihaibudiu:
This PR should not affect the behavior of Calcite in this respect, other than fixing the bug. I believe that the first function that typechecks (where libraries are probed in the order they are specified) will be the one used, but this is not documented as far as I can tell.

This may actually be a problem for projects that use multiple libraries: as Calcite evolves and adds overloads to these functions, and there are lots of them, with different behaviors, software using Calcite that used to work may stop working. But this is not a problem that is new.

NobiGo:
> This PR should not affect the behavior of Calcite in this respect, other than fixing the bug. I believe that the first function that typechecks (where libraries are probed in the order they are specified) will be the one used, but this is not documented as far as I can tell.
> 
> This may actually be a problem for projects that use multiple libraries: as Calcite evolves and adds overloads to these functions, and there are lots of them, with different behaviors, software using Calcite that used to work may stop working. But this is not a problem that is new.

Thanks for your reply. When loading multiple libraries with identical function names if we don't throw exceptions and can clearly determine which library's function is effective, then I think it's not a problem for us. I'm worried we won't be able to decide on which library's function is effective. I will create another issue to document it.

PR #3906, opened 2024-08-07
[CALCITE-6464] Type inference for DECIMAL division seems incorrect

mihaibudiu:
Alternative PR to #3855. I have taken some tests from that PR, but adjusted the algorithm.
See the JIRA and the comments for a justification.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
+++ b/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
@@ -259,7 +259,7 @@ public interface RelDataTypeSystem {
    * <li>Then the result type is a decimal with:
    *   <ul>
    *   <li>d = p1 - s1 + s2</li>
-   *   <li>s &lt; max(6, s1 + p2 + 1)</li>
+   *   <li>s = max(6, s1 + p2 + 1)</li>
    *   <li>p = d + s</li>
    *   </ul>
    * </li>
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
```diff
@@ -294,21 +294,21 @@ default boolean shouldUseDoubleMultiplication(RelDataTypeFactory typeFactory,
...
+        int precision = d + scale;
+
+  // Rules from
+  // https://learn.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql
+        int bound = getMaxNumericPrecision() - 6;  // 32 in the MS documentation
+        if (precision <= bound) {
+          scale = Math.min(scale, getMaxNumericScale() - (precision - scale));
+        } else {
+          // precision > bound
+          scale = Math.min(6, scale);
```
suibianwanwank, line 308: If MaxNumericScale < 6, is it possible here to exceed

mihaibudiu: I will fix this case too, but, frankly, if a type system for DECIMAL does not have 6 digits of scale it's probably not very useful

suibianwanwank: +1

mihaibudiu:
Can someone review this PR? It fixes a genuine bug.

suibianwanwank (approved):
LGTM, I'm not sure if this is the most applicable for calcite, but it looks like sqlserver's is better than PG's

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
```diff
@@ -294,22 +294,41 @@ default boolean shouldUseDoubleMultiplication(RelDataTypeFactory typeFactory,
...
+
+  // Rules from
+  // https://learn.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql
+        // Rules reproduced here in case the web page goes away.
+        // Note that in SQL server getMaxNumericPrecision() == 38 and
+        // getMaxNumericScale() > 6.
+        // In multiplication and division operations, we need precision - scale places to store
+        // the integral part of the result. The scale might be reduced using the following rules:
+        //
+        // - The resulting scale is reduced to min(scale, 38 - (precision-scale))
```
NobiGo, line 310: Do we need to add another tests for these rules?

NobiGo (approved):

mihaibudiu:
> LGTM, I'm not sure if this is the most applicable for calcite, but it looks like sqlserver's is better than PG's

The type system of SqlServer (at least the version described in that article) is very different from the type system of Postgres. Postgres supports essentially arbitrary precision decimals, and negative scales. But computations with very high precision are very costly. So it's not easy to say which one is "better".

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
mihaibudiu: I think that the existing tests cover these changes - as a proof I had to change many types in the division result types in the validator and operator tests. I also added a test for a type system with lower precision, which motivated filing this bug in the first place.

caicancai (approved):
LGTM

PR #3907, opened 2024-08-07
[CALCITE-6519] Non-aggregate query that uses measure in ORDER BY

julianhyde:
As of [CALCITE-4496], a non-aggregate query can use measures in its `SELECT` clause; this change further allows a non-aggregate query to use measures in its `ORDER BY` clause. Example:
```
  WITH empm AS
    (SELECT *, avg(sal) AS MEASURE avgSal FROM emp)
  SELECT avgSal, deptno
  FROM empm
  ORDER BY avgSal DESC LIMIT 3;
```

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -5841,6 +5841,53 @@ class RelOptRulesTest extends RelOptTestBase {
         .check();
   }
 
+  /** Tests use of a measure in a non-aggregate query. Calls
+   * {@link RelOptFixture#checkUnchanged()} because Sql-to-rel has done the
+   * necessary work already. */
+  @Test void testMeasureWithoutGroupBy() {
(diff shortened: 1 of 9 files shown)


mihaibudiu (approved):

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -5841,6 +5841,53 @@ private HepProgram getTransitiveProgram() {
         .check();
   }
 
+  /** Tests use of a measure in a non-aggregate query. Calls
+   * {@link RelOptFixture#checkUnchanged()} because Sql-to-rel has done the
+   * necessary work already. */
```
caicancai, line 5846: Would it be better to add jira link?

julianhyde: No, because this query has no `ORDER BY` such queries were working before this change.

PR #3908, opened 2024-08-07
Keep the findBestExp method exception consistent between HepPlanner and VolcanoPlanner

ChengJie1053:


mihaibudiu (commented):
I personally prefer to quote the identifiers in question in such messages:
'root' must not be null

ChengJie1053:
> I personally prefer to quote the identifiers in question in such messages: 'root' must not be null

Ok, thank you. I'll change the code

mihaibudiu (approved):

caicancai:
@ChengJie1053 Please git squash

ChengJie1053:
> @ChengJie1053 Please git squash

Ok, thank you

caicancai (approved):

core/src/main/java/org/apache/calcite/plan/hep/HepPlanner.java
```diff
@@ -186,14 +186,14 @@ public HepPlanner(
   }
 
   @Override public RelNode findBestExp() {
-    requireNonNull(root, "root");
+    requireNonNull(root, "'root' must not be null");
```
NobiGo, line 189: What is the error message before this PR? I want to know `must not be null` it is repetitive with `requireNonNull`? The other places in this project all use it like `requireNonNull(root, 'root');`

ChengJie1053: Ok, thanks for reviewing the code for me. I was initially trying to get the same error message from findBestExp for HepPlanner and VolcanoPlanner
image

ChengJie1053: Ok, thanks for reviewing the code for me

NobiGo:
@ChengJie1053  Hi, How about we change the commit info to `Keep the findBestExp method exception consistent between HepPlanner and VolcanoPlanner` ?

ChengJie1053:
> @ChengJie1053 Hi, How about we change the commit info to `Keep the findBestExp method exception consistent between HepPlanner and VolcanoPlanner` ?

Ok, thanks for reviewing the code for me

NobiGo:
@ChengJie1053 Please squash the commits. If the CI pass, I will merge it.

NobiGo:
@ChengJie1053 Thanks for the contribution.

PR #3909, opened 2024-08-08
[CALCITE-6522] MAP_KEYS and MAP_VALUES   function should throw if a key value is null

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6522


caicancai:
```
cala> val df = spark.sql("select map_values(map(cast(null as int),0, 'foo', 1))");
df: org.apache.spark.sql.DataFrame = [map_values(map(CAST(NULL AS INT), 0, foo, 1)): array<int>]

scala> df.show()
org.apache.spark.SparkRuntimeException: [NULL_MAP_KEY] Cannot use null as map key.
at org.apache.spark.sql.errors.QueryExecutionErrors$.nullAsMapKeyNotAllowedError(QueryExecutionErrors.scala:445)
```
From the spark log, the error is reported at runtime, not parameter verification.
Similar to the previous 6400

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8367,23 +8367,15 @@ void checkArrayReverseFunc(SqlOperatorFixture f0, SqlFunction function,
     final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
     f.checkScalar("map_keys(map['foo', 1, 'bar', 2])", "[foo, bar]",
         "CHAR(3) NOT NULL ARRAY NOT NULL");
-    f.checkScalar("map_keys(map['foo', 1, null, 2])", "[foo, null]",
-        "CHAR(3) ARRAY NOT NULL");
+
```
mihaibudiu, line 8370: can you see who added these tests in the first place and check why they added them? maybe the behavior has changed, or maybe they are not testing spark?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -5682,11 +5682,21 @@ public static List mapEntries(Map<Object, Object> map) {
 
   /** Support the MAP_KEYS function. */
   public static List mapKeys(Map map) {
+    for (Object key : map.keySet()) {
```
julianhyde, line 5685: It would be more efficient to create the list and then check whether it contains nulls. (`ArrayList.contains` is fast.)

julianhyde: On reflection, I think you should use `ImmutableList.copyOf(map.keySet())`. Immutable lists are better, and this will be a quick operation if map is an `ImmutableMap`.

Before doing this, use `map.keySet().contains(null)`. It's more efficient than your `for` loop.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -5682,11 +5682,21 @@ public static List mapEntries(Map<Object, Object> map) {
...
+      if (key == null) {
+        throw new IllegalArgumentException("Cannot use null as map key");
```
julianhyde, line 5687: You should `throw RESOURCE.something`. We have resources for error messages that the user might see during normal operation.

julianhyde:
There's a tab or multiple spaces in the jira case summary, commit message, PR summary. Please fix.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: `map.keySet().contains(null)` and
```
    ImmutableList<Object> arrayList =  ImmutableList.copyOf(map.keySet());
    if(arrayList.contains(null)){
      throw new IllegalArgumentException("Cannot use null as map key");
    }
```
Is there a difference? I'm sorry I didn't understand what you meant.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: @chucheng92 If you have time, could you please review this PR?

caicancai:
> There's a tab or multiple spaces in the jira case summary, commit message, PR summary. Please fix.

done

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: @mihaibudiu I noticed that the changes here and the changes in the previous [PR](https://github.com/apache/calcite/pull/3787) were made by @chucheng92 in the same jira case. I think it is better to refer to his answer last time.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -5682,11 +5682,17 @@ public static List mapEntries(Map<Object, Object> map) {
 
   /** Support the MAP_KEYS function. */
   public static List mapKeys(Map map) {
+    if (map.keySet().contains(null)) {
```
mihaibudiu, line 5685: what is interesting is that you detect that a map key is null only when you call the function map_keys. But shouldn't this be detected when inserting something into the map, if it is indeed illegal?

If it's not illegal to insert entries with null keys, maybe the error message should say that the function map_keys does not support null keys in maps.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
julianhyde: `ImmutableList.copyOf` will throw if the list contains null. You don't want that.

`ImmutableList.copyOf(map.keySet())` will be very efficient if `map` happens to be an `ImmutableMap`. You do want that.

caicancai: No, Spark will throw an exception if a null value appears

caicancai: done，thanks

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
mihaibudiu: I don't know what parameters you are talking about.
I am saying that the error message does not seem right. The error is not inserting an entry with a null key in a map, it's using a map with a null key in the function mapKeys

mihaibudiu:
There's error-prone error which has to be fixed.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
caicancai: Thank you for your patience, I misunderstood you, I will fix it.

caicancai:
@mihaibudiu @julianhyde Sorry for the late fix, I'm on vacation and have time to fix this PR.
Can you review this PR for me if you have time?

caicancai:
@mihaibudiu I squashed the commit and this problem should be solved in 1.38

mihaibudiu (approved):

PR #3911, opened 2024-08-10
[CALCITE-6396] Add ADD_MONTHS function (enabled in Oracle, Spark library)

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6396

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
@@ -207,6 +207,8 @@ public class StandardConvertletTable extends ReflectiveConvertletTable {
 
     registerOp(SqlLibraryOperators.DATE_ADD,
         new TimestampAddConvertlet());
+    registerOp(SqlLibraryOperators.ADD_MONTHS,
+        new TimestampAddConvertlet());
     registerOp(SqlLibraryOperators.DATE_DIFF,
         new TimestampDiffConvertlet());
     registerOp(SqlLibraryOperators.DATE_SUB,
@@ -2135,11 +2137,41 @@ public class StandardConvertletTable extends ReflectiveConvertletTable {
       final RexNode op2;
       switch (call.operandCount()) {
       case 2:
-        // BigQuery-style 'TIMESTAMP_ADD(timestamp, interval)'
-        final SqlBasicCall operandCall = call.operand(1);
-        qualifier  = operandCall.operand(1);
-        op1 = cx.convertExpression(operandCall.operand(0));
-        op2 = cx.convertExpression(call.operand(0));
+        // Oracle-style 'ADD_MONTHS(date, integer months)'
+        if (call.getOperator() == SqlLibraryOperators.ADD_MONTHS) {
+          qualifier = new SqlIntervalQualifier(TimeUnit.MONTH, null, SqlParserPos.ZERO);
+          RexNode opfirstparameter = cx.convertExpression(call.operand(0));
+          if (opfirstparameter.getType().getSqlTypeName() == SqlTypeName.TIMESTAMP) {
+            RelDataType timestampType = cx.getTypeFactory().createSqlType(SqlTypeName.DATE);
+            op2 = rexBuilder.makeCast(timestampType, opfirstparameter);
+          } else {
+            op2 = cx.convertExpression(call.operand(0));
+          }
+
+          RexNode opsecondparameter = cx.convertExpression(call.operand(1));
+          RelDataTypeFactory typeFactory = cx.getTypeFactory();
+
+          // In Calcite, cast('1.2' as integer) is invalid.
+          // For details, see https://issues.apache.org/jira/browse/CALCITE-1439
+          // When trying to cast a string value to an integer type, Calcite may
+          // encounter errors if the string value cannot be successfully converted.
+          // To handle this, the string needs to be first converted to a double type,
+          // and then the double value can be converted to an integer type.
+          // Since the final target type is integer, converting the string to double first
+          // will not lose precision for the add_months operation.
+          if (opsecondparameter.getType().getSqlTypeName() == SqlTypeName.CHAR) {
+            RelDataType doubleType = typeFactory.createSqlType(SqlTypeName.DOUBLE);
+            opsecondparameter = rexBuilder.makeCast(doubleType, opsecondparameter);
+          }
+          RelDataType intType = typeFactory.createSqlType(SqlTypeName.INTEGER);
+          op1 = rexBuilder.makeCast(intType, opsecondparameter);
+        } else {
+          // BigQuery-style 'TIMESTAMP_ADD(timestamp, interval)'
+          final SqlBasicCall operandCall = call.operand(1);
+          qualifier = operandCall.operand(1);
+          op1 = cx.convertExpression(operandCall.operand(0));
+          op2 = cx.convertExpression(call.operand(0));
+        }
         break;
       default:
         // JDBC-style 'TIMESTAMPADD(unit, count, timestamp)'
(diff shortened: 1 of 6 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13519,6 +13519,32 @@ void testTimestampDiff(boolean coercionEnabled) {
...
+   */
+  @Test void testAddMonths() {
+    final SqlOperatorFixture f0 = fixture()
+        .setFor(SqlLibraryOperators.ADD_MONTHS);
+    f0.checkFails("^add_months(date '2008-12-25', "
+            + "5)^",
+        "No match found for function signature "
+            + "ADD_MONTHS\\(<DATE>, <NUMERIC>\\)", false);
+    final SqlOperatorFixture f = f0.withLibrary(SqlLibrary.SPARK);
+    f.checkScalar("add_months(date '2016-02-22', 2)",
```
mihaibudiu, line 13534: how about a few tests with other data types than integers?
strings, longs, decimals

mihaibudiu (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: This will fail during parameter parsing. Calcite currently does not support string, double, and other types for date function calculations.

mihaibudiu: if that's the correct behavior you should leave these tests in.
otherwise, it means the implementation is not complete

caicancai: I read the official document of DataRisks, and the description in the document is only int type
image
https://docs.databricks.com/en/sql/language-manual/functions/add_months.html

caicancai:
It looks like this PR is not yet completed

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13519,6 +13519,46 @@ void testTimestampDiff(boolean coercionEnabled) {
     f.checkNull("date_add(CAST(NULL AS DATE), interval 5 day)");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6396">[CALCITE-6396]
+   * Add ADD_MONTHS function enabled in Oracle, Spark library)</a>.
```
NobiGo, line 13524: Missing `(` in Jira summary.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13519,6 +13519,46 @@ void testTimestampDiff(boolean coercionEnabled) {
...
+   */
+  @Test void testAddMonths() {
+    final SqlOperatorFixture f0 = fixture()
+        .setFor(SqlLibraryOperators.ADD_MONTHS);
+    f0.checkFails("^add_months(date '2008-12-25', "
+            + "5)^",
+        "No match found for function signature "
+            + "ADD_MONTHS\\(<DATE>, <NUMERIC>\\)", false);
+    final Consumer<SqlOperatorFixture> consumer = f -> {
+      f.checkScalar("add_months(date '2016-02-22', 2)",
```
NobiGo, line 13534: Please add tests like `select ADD_MONTHS('2016-08-31',1);`  
In Spark, this SQL will return
```
spark-sql (default)> select  ADD_MONTHS('2016-08-31',1);
2016-09-30
Time taken: 0.118 seconds, Fetched 1 row(s)
```

NobiGo (changes requested):
Need boundary test.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: done

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -2135,11 +2137,21 @@ private static class TimestampAddConvertlet implements SqlRexConvertlet {
       final RexNode op2;
       switch (call.operandCount()) {
       case 2:
-        // BigQuery-style 'TIMESTAMP_ADD(timestamp, interval)'
-        final SqlBasicCall operandCall = call.operand(1);
-        qualifier  = operandCall.operand(1);
-        op1 = cx.convertExpression(operandCall.operand(0));
-        op2 = cx.convertExpression(call.operand(0));
+        if (call.getOperator() == SqlLibraryOperators.ADD_MONTHS) {
+          qualifier = new SqlIntervalQualifier(TimeUnit.MONTH, null, SqlParserPos.ZERO);
```
suibianwanwank, line 2141: Add Comment like _BigQuery-style 'TIMESTAMP_ADD(timestamp, interval)_ here , would be easier to understand. WDYT

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13519,6 +13519,52 @@ void testTimestampDiff(boolean coercionEnabled) {
...
+          "2016-04-22",
+          "DATE NOT NULL");
+      f.checkScalar("add_months(date '2016-02-22', -2)",
```
suibianwanwank, line 13537: I think it would be a good idea to include a timestamp type test? In spark, you can use timestamp and return a date type

caicancai: The Oracle and Spark official documents describe the date type, which I think should not be changed to timestamp.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -2135,11 +2137,34 @@ private static class TimestampAddConvertlet implements SqlRexConvertlet {
...
+
+          // In Calcite, cast('1.2' as integer) is invalid.
+          // For details, see https://issues.apache.org/jira/browse/CALCITE-1439
+          // When trying to cast a string value to an integer type, Calcite may
+          // encounter errors if the string value cannot be successfully converted.
+          // To handle this, the string needs to be first converted to a double type,
+          // and then the double value can be converted to an integer type.
+          // Since the final target type is integer, converting the string to double first
+          // will not lose precision for the add_months operation.
+          if (op.getType().getSqlTypeName() == SqlTypeName.CHAR) {
```
caicancai, line 2154: I have tested all the boundary tests I think and it is compatible. This is my method and argument. If you have any questions, you are welcome to discuss.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
caicancai: done

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
suibianwanwank: The Oracle Functions documentation describes it as: "The date argument can be a datetime value or any value that can be implicitly converted to DATE". Although the Spark documentation doesn't explicitly mention this, such implicit conversions exist. I assume we can use makeCast to let the user determine compatibility between types? WDYT

caicancai: Yes, this can be done, is there an Oracle environment where I can verify this?

caicancai: > The Oracle Functions documentation describes it as: "The date argument can be a datetime value or any value that can be implicitly converted to DATE". Although the Spark documentation doesn't explicitly mention this, such implicit conversions exist. I assume we can use makeCast to let the user determine compatibility between types? WDYT

thanks, you are right

NobiGo (approved):
LGTM

suibianwanwank (approved):
LGTM

caicancai:
no other comment, squashed.
Thanks for your review @mihaibudiu @NobiGo @suibianwanwank

PR #3912, opened 2024-08-12
[CALCITE-6501] Assertion Error in JoinUnifyRule Due to Type Mismatch

suibianwanwank:
Assertion errors occur in some joinUnifyRules because of the change in nullable type produced by the join. Intercepts these cases and replaces the assertion error with a return of null. For a specific example, see the test.

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/MaterializedViewSubstitutionVisitorTest.java
+++ b/core/src/test/java/org/apache/calcite/test/MaterializedViewSubstitutionVisitorTest.java
@@ -1136,6 +1136,111 @@ public class MaterializedViewSubstitutionVisitorTest {
     sql(mv, query).ok();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6501">[CALCITE-6501]
+   * Assertion Error in JoinUnifyRule Due to Type Mismatch</a>. */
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):

core/src/test/java/org/apache/calcite/test/MaterializedViewSubstitutionVisitorTest.java
```diff
@@ -1136,6 +1136,90 @@ protected final MaterializedViewFixture sql(String materialize,
     sql(mv, query).ok();
   }
 
+  @Test void testLeftProjectOnRightJoinToJoinFailed() {
```
caicancai, line 1139: Add jira link

caicancai (commented):
Overall LGTM

core/src/test/java/org/apache/calcite/test/MaterializedViewSubstitutionVisitorTest.java
suibianwanwank: Done✅

caicancai (approved):

suibianwanwank:
I've squash the commits and if there are no more reviews, I think it's ready to merge.

suibianwanwank:
Thanks for your review @mihaibudiu @caicancai

PR #3913, opened 2024-08-12
Enable more tests in RelBuilderExample

yangli-stu:
The original for loop was set to iterate from 0 to 3, which excluded the case 4 from being processed in the doExample method. This commit updates the loop condition to i <= 4 to ensure all cases are covered.
image

### DIFF
--- a/core/src/test/java/org/apache/calcite/examples/RelBuilderExample.java
+++ b/core/src/test/java/org/apache/calcite/examples/RelBuilderExample.java
@@ -44,7 +44,7 @@ public class RelBuilderExample {
     // to the SCOTT database, with tables EMP and DEPT.
     final FrameworkConfig config = RelBuilderTest.config().build();
     final RelBuilder builder = RelBuilder.create(config);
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):

yangli-stu:
After fixing the loop condition, it was found that the table in Example 4 no longer exists in the latest schema, so another fix was carried out.

Before fix:
image

After fix:
image

NobiGo:
@yangli-stu The PR looks good. In Calcite, the PR commit info does not start with Fix. we can follow the style of previous. Like 'Enable more tests in RelBuilderExample'.

NobiGo (approved):

caicancai (approved):

NobiGo:
@yangli-stu One last thing, please squash the commits.

mihaibudiu:
don't worry about that, it fails intermittently

NobiGo:
Thanks for your contribution @yangli-stu .

PR #3915, opened 2024-08-13
[CALCITE-6527] Add DATE_ADD function (enabled in Spark library)

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6527

### DIFF
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2772,6 +2772,7 @@ In the following:
 | b s | DATE_FROM_UNIX_DATE(integer)                 | Returns the DATE that is *integer* days after 1970-01-01
 | p r | DATE_PART(timeUnit, datetime)                | Equivalent to `EXTRACT(timeUnit FROM  datetime)`
 | b | DATE_ADD(date, interval)                       | Returns the DATE value that occurs *interval* after *date*
+| s | DATE_ADD(start_date, num_days)                 | Returns the DATE that is *num_days* after *start_date*
 | b | DATE_DIFF(date, date2, timeUnit)               | Returns the whole number of *timeUnit* between *date* and *date2*
 | b | DATE_SUB(date, interval)                       | Returns the DATE value that occurs *interval* before *date*
 | b | DATE_TRUNC(date, timeUnit)                     | Truncates *date* to the granularity of *timeUnit*, rounding to the beginning of the unit
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2696,7 +2696,7 @@ In the following:
 | m | expr1 <=> expr2                                | Whether two values are equal, treating null values as the same, and it's similar to `IS NOT DISTINCT FROM`
 | p | ACOSD(numeric)                                 | Returns the inverse cosine of *numeric* in degrees as a double. Returns NaN if *numeric* is NaN. Fails if *numeric* is less than -1.0 or greater than 1.0.
 | * | ACOSH(numeric)                                 | Returns the inverse hyperbolic cosine of *numeric*
-| o s | ADD_MONTHS(start_date, num_months)           | Returns the date that is *num_months* after *start_date*
+| o s | ADD_MONTHS(date, num_months)                 | Returns the date that is *num_months* after *date*
 | s | ARRAY([expr [, expr ]*])                       | Construct an array in Apache Spark. The function allows users to use `ARRAY()` to create an empty array
 | s | ARRAY_APPEND(array, element)                   | Appends an *element* to the end of the *array* and returns the result. Type of *element* should be similar to type of the elements of the *array*. If the *array* is null, the function will return null. If an *element* that is null, the null *element* will be added to the end of the *array*
 | s | ARRAY_COMPACT(array)                           | Removes null values from the *array*
@@ -2772,7 +2772,7 @@ In the following:
 | b s | DATE_FROM_UNIX_DATE(integer)                 | Returns the DATE that is *integer* days after 1970-01-01
 | p r | DATE_PART(timeUnit, datetime)                | Equivalent to `EXTRACT(timeUnit FROM  datetime)`
 | b | DATE_ADD(date, interval)                       | Returns the DATE value that occurs *interval* after *date*
-| s | DATE_ADD(start_date, num_days)                 | Returns the DATE that is *num_days* after *start_date*
+| s | DATE_ADD(date, num_days)                       | Returns the DATE that is *num_days* after *date*
 | b | DATE_DIFF(date, date2, timeUnit)               | Returns the whole number of *timeUnit* between *date* and *date2*
 | b | DATE_SUB(date, interval)                       | Returns the DATE value that occurs *interval* before *date*
 | b | DATE_TRUNC(date, timeUnit)                     | Truncates *date* to the granularity of *timeUnit*, rounding to the beginning of the unit
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -13825,6 +13825,54 @@ public class SqlOperatorTest {
   }
 
   /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6527">[CALCITE-6527]
+   * Add DATE_ADD function (enabled in Spark library)</a>.
+   */
+  @Test void testDateAddSpark() {
+    final SqlOperatorFixture f0 = fixture()
+        .setFor(SqlLibraryOperators.DATE_ADD_SPARK);
+    f0.checkFails("^date_add(date '2008-12-25', "
+            + "5)^",
+        "No match found for function signature "
+            + "DATE_ADD\\(<DATE>, <NUMERIC>\\)", false);
+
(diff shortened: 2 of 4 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13824,6 +13824,54 @@ void testTimestampDiff(boolean coercionEnabled) {
...
+        "2016-02-20",
+        "DATE NOT NULL");
+    f.checkScalar("date_add(timestamp '2016-02-22 13:00:01', -2)",
+        "2016-02-20",
+        "DATE NOT NULL");
+    f.checkScalar("date_add(timestamp '2016-02-22 13:00:01', -2.0)",
+        "2016-02-20",
+        "DATE NOT NULL");
+    f.checkNull("date_add(CAST(NULL AS DATE), 5)");
+    f.checkNull("date_add(date '2016-02-22', CAST(NULL AS INTEGER))");
```
mihaibudiu, line 13871: how about a few tests around year 0 and the Gregorian calendar cutoff date?
For example, date_add('0001-01-01', -2). Notice that there is no year 0, so this probably should throw.

caicancai: Good suggestion. I tested it in spark and it seems that spark is not aware of this error. Should we fix this error in calcite or keep this behavior?
scala> val df = spark.sql("SELECT date_add('0001-01-01', -2)")
df: org.apache.spark.sql.DataFrame = [date_add(0001-01-01, -2): date]

```
scala> df.show()
+------------------------+
|date_add(0001-01-01, -2)|
+------------------------+
|              0000-12-30|
+------------------------+


scala> val df = spark.sql("SELECT date_add('0000-01-01', -2)")
df: org.apache.spark.sql.DataFrame = [date_add(0000-01-01, -2): date]

scala> df.show()
+------------------------+
|date_add(0000-01-01, -2)|
+------------------------+
|             -0001-12-30|
+------------------------+

```

mihaibudiu: You should check the issues open for spark, this looks like a bug to me. There is no year 0.
If there is no bug maybe you can open one.

caicancai: OK, I will try again in the latest spark code. If there is no corresponding jira record, I will file a jira case to describe this bug.

I feel that 000-00-01 is reasonable, but negative numbers are unreasonable. Calcite currently follows this principle. The following are my test results in Calcite.
```
    f.checkScalar("date_add('0001-01-01', -2)",
        "0000-12-30",
        "DATE NOT NULL");
    f.checkFails("date_add('0000-01-01', -2)",
        "Value of YEAR field is out of range in '0000-01-01'",
        true);

```

mihaibudiu: on the contrary, year 0 is not legal, it never existed.
and people lived before Christ, so years BC are completely legal in many databases. whether they show up as negative numbers, that's a different question. Calcite doesn't support them, though

site/_docs/reference.md
```diff
@@ -2772,6 +2772,7 @@ In the following:
 | b s | DATE_FROM_UNIX_DATE(integer)                 | Returns the DATE that is *integer* days after 1970-01-01
 | p r | DATE_PART(timeUnit, datetime)                | Equivalent to `EXTRACT(timeUnit FROM  datetime)`
 | b | DATE_ADD(date, interval)                       | Returns the DATE value that occurs *interval* after *date*
+| s | DATE_ADD(start_date, num_days)                 | Returns the DATE that is *num_days* after *start_date*
```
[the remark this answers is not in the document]
caicancai: Thanks for the reminder, I'll fix it later.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: I have already raised a jira case in spark
https://issues.apache.org/jira/browse/SPARK-49247

site/_docs/reference.md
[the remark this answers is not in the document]

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: I think that in my previous review I asked for a few more tests around boundary conditions, I don't see any here.

julianhyde: _Au contraire_. `0000-00-00` is a valid 8601 date.

See https://en.wikipedia.org/wiki/ISO_8601#Calendar_dates

mihaibudiu: Will the Calcite DATE datatype accept such a value?

caicancai: I think I will do some tests and describe them later. It seems that calcite has done some processing but it is not perfect. I will fix it in this PR

julianhyde: @mihaibudiu, I don't recall. I just happened to reading about ISO-8601 so I thought I'd correct misinformation.

caicancai: Sorry for the late reply.
After testing, the date_add function will format the date. For example, invalid dates such as `0000-01-01` will fail the date format.

caicancai:
@mihaibudiu I added some tests that I thought were borderline and found some issues, I'd like to hear your thoughts

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13824,6 +13824,63 @@ void testTimestampDiff(boolean coercionEnabled) {
...
+    f.checkScalar("date_add(timestamp '2016-02-22 13:00:01', -2)",
+        "2016-02-20",
+        "DATE NOT NULL");
+    f.checkScalar("date_add(timestamp '2016-02-22 13:00:01', -2.0)",
+        "2016-02-20",
+        "DATE NOT NULL");
+    f.checkScalar("date_add(date '0001-01-01', '-2.0')",
+        "0000-12-30",
+        "DATE NOT NULL");
+    f.checkScalar("date_add(date '0001-01-01', '-367')",
```
mihaibudiu, line 13873: what is going on here? Why is `000/` a valid year?

caicancai: @mihaibudiu I raised a jira to discuss this issue. Should I delete this test?
https://issues.apache.org/jira/browse/CALCITE-6536

mihaibudiu: I would still like to know why this test passes.

mihaibudiu: If a test passes that shouldn't it means that there's somewhere another bug that is masking a problem.

caicancai: > If a test passes that shouldn't it means that there's somewhere another bug that is masking a problem.

agree. 
I will fix it, the date function does validation when it handles 000/or 0000 as a parameter, but when 000/or 0000 is returned as a result, no such validation is done, I think this is the bug

caicancai: @mihaibudiu I'm not sure how to solve this bug, it looks a bit complicated to me 😭

mihaibudiu: I don't think Calcite can offer exact compatibility to Spark in this case.
The best solution for Calcite is to give a runtime error.
But I would still like to know why the date with a slash didn't report a different error.

caicancai: > I don't think Calcite can offer exact compatibility to Spark in this case.
> The best solution for Calcite is to give a runtime error.
> But I would still like to know why the date with a slash didn't report a different error.

@mihaibudiu Calcite will perform parser validation on the input parameters.
```
    final Calendar cal =
        DateTimeUtils.parseDateFormat(s, Format.get().date,
            DateTimeUtils.UTC_ZONE);
```
However, there is no restriction on the output result set. I've been researching for a long time and I'm not sure where to limit the result set

mihaibudiu: I don't understand what you are saying.
If you add two very large numbers in Calcite you will get a runtime error.
How is this different?

caicancai: I mean, I don't know why I get a runtime error.
Sorry, I didn't express myself clearly.

caicancai:
@mihaibudiu Almost all date functions have this problem, I consider it another bug, I will fix it. Before fixing it, this PR is marked as draft.
Thank you for your review

PR #3917, opened 2024-08-16
[CALCITE-4838] Add RoundingMode in RelDataTypeSystem to specify the rounding behavior

NobiGo:
…nding mode when cast an approximate numeric to int

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
@@ -456,45 +456,51 @@ public class RexToLixTranslator implements RexVisitor<RexToLixTranslator.Result>
           && scale != RelDataType.SCALE_NOT_SPECIFIED) {
         if (sourceType.getFamily() == SqlTypeFamily.CHARACTER) {
           return Expressions.call(
-              BuiltInMethod.CHAR_DECIMAL_CAST.method,
+              BuiltInMethod.CHAR_DECIMAL_CAST_ROUNDING_MODE.method,
               operand,
               Expressions.constant(precision),
-              Expressions.constant(scale));
+              Expressions.constant(scale),
+              Expressions.constant(typeFactory.getTypeSystem().roundingMode()));
         } else if (sourceType.getFamily() == SqlTypeFamily.INTERVAL_DAY_TIME) {
           return Expressions.call(
-              BuiltInMethod.SHORT_INTERVAL_DECIMAL_CAST.method,
+              BuiltInMethod.SHORT_INTERVAL_DECIMAL_CAST_ROUNDING_MODE.method,
               operand,
               Expressions.constant(precision),
               Expressions.constant(scale),
-              Expressions.constant(sourceType.getSqlTypeName().getEndUnit().multiplier));
+              Expressions.constant(sourceType.getSqlTypeName().getEndUnit().multiplier),
+              Expressions.constant(typeFactory.getTypeSystem().roundingMode()));
         } else if (sourceType.getFamily() == SqlTypeFamily.INTERVAL_YEAR_MONTH) {
           return Expressions.call(
-              BuiltInMethod.LONG_INTERVAL_DECIMAL_CAST.method,
+              BuiltInMethod.LONG_INTERVAL_DECIMAL_CAST_ROUNDING_MODE.method,
               operand,
               Expressions.constant(precision),
               Expressions.constant(scale),
-              Expressions.constant(sourceType.getSqlTypeName().getEndUnit().multiplier));
+              Expressions.constant(sourceType.getSqlTypeName().getEndUnit().multiplier),
+              Expressions.constant(typeFactory.getTypeSystem().roundingMode()));
         } else if (sourceType.getSqlTypeName() == SqlTypeName.DECIMAL) {
           // Cast from DECIMAL to DECIMAL, may adjust scale and precision.
           return Expressions.call(
-              BuiltInMethod.DECIMAL_DECIMAL_CAST.method,
+              BuiltInMethod.DECIMAL_DECIMAL_CAST_ROUNDING_MODE.method,
               operand,
               Expressions.constant(precision),
-              Expressions.constant(scale));
+              Expressions.constant(scale),
+              Expressions.constant(typeFactory.getTypeSystem().roundingMode()));
         } else if (SqlTypeName.INT_TYPES.contains(sourceType.getSqlTypeName())) {
           // Cast from INTEGER to DECIMAL, check for overflow
           return Expressions.call(
-              BuiltInMethod.INTEGER_DECIMAL_CAST.method,
+              BuiltInMethod.INTEGER_DECIMAL_CAST_ROUNDING_MODE.method,
               operand,
               Expressions.constant(precision),
-              Expressions.constant(scale));
+              Expressions.constant(scale),
+              Expressions.constant(typeFactory.getTypeSystem().roundingMode()));
         }  else if (SqlTypeName.APPROX_TYPES.contains(sourceType.getSqlTypeName())) {
           // Cast from FLOAT/DOUBLE to DECIMAL
           return Expressions.call(
-              BuiltInMethod.FP_DECIMAL_CAST.method,
+              BuiltInMethod.FP_DECIMAL_CAST_ROUNDING_MODE.method,
               operand,
               Expressions.constant(precision),
-              Expressions.constant(scale));
+              Expressions.constant(scale),
+              Expressions.constant(typeFactory.getTypeSystem().roundingMode()));
         }
       }
       return defaultExpression.get();
@@ -505,9 +511,9 @@ public class RexToLixTranslator implements RexVisitor<RexToLixTranslator.Result>
     case SMALLINT: {
       if (SqlTypeName.NUMERIC_TYPES.contains(sourceType.getSqlTypeName())) {
         return Expressions.call(
-            BuiltInMethod.INTEGER_CAST.method,
+            BuiltInMethod.INTEGER_CAST_ROUNDING_MODE.method,
             Expressions.constant(Primitive.of(typeFactory.getJavaClass(targetType))),
-            operand);
+            operand, Expressions.constant(typeFactory.getTypeSystem().roundingMode()));
       }
       return defaultExpression.get();
     }
--- a/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
+++ b/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
@@ -22,6 +22,8 @@ import org.apache.calcite.util.Glossary;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
+import java.math.RoundingMode;
+
 /**
  * Type system.
  *
@@ -59,6 +61,9 @@ public interface RelDataTypeSystem {
 
   /** Returns the maximum precision of a NUMERIC or DECIMAL type. */
   int getMaxNumericPrecision();
+
+  /** Returns the rounding behavior for numerical operations capable of discarding precision. */
+  RoundingMode roundingMode();
 
   /** Returns the LITERAL string for the type, either PREFIX/SUFFIX. */
   @Nullable String getLiteral(SqlTypeName typeName, boolean isPrefix);
--- a/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystemImpl.java
+++ b/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystemImpl.java
@@ -22,6 +22,8 @@ import org.apache.calcite.sql.type.SqlTypeName;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
+import java.math.RoundingMode;
+
 /** Default implementation of
  * {@link org.apache.calcite.rel.type.RelDataTypeSystem},
  * providing parameters from the SQL standard.
@@ -158,6 +160,10 @@ public abstract class RelDataTypeSystemImpl implements RelDataTypeSystem {
 
   @Override public int getMaxNumericPrecision() {
     return 19;
+  }
+
+  @Override public RoundingMode roundingMode() {
+    return RoundingMode.DOWN;
   }
 
   @Override public @Nullable String getLiteral(SqlTypeName typeName, boolean isPrefix) {
--- a/core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
+++ b/core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
@@ -125,6 +125,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.nio.charset.Charset;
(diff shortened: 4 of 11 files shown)


core/src/test/java/org/apache/calcite/test/CoreQuidemTest.java
```diff
@@ -88,6 +88,33 @@ public static void main(String[] args) throws Exception {
...
+          // Same as "scott", but uses Calcite's default conformance.
+          return CalciteAssert.that()
+              .with(CalciteConnectionProperty.PARSER_FACTORY,
+                  ExtensionDdlExecutor.class.getName() + "#PARSER_FACTORY")
+              .with(CalciteConnectionProperty.CONFORMANCE,
+                  SqlConformanceEnum.DEFAULT)
+              .with(CalciteAssert.Config.SCOTT)
+              .connect();
+        case "scott-sqlserver2008":
+          // Same as "scott", but uses Calcite's default conformance.
```
julianhyde, line 101: comment is copy-pasted and identical to other comments

core/src/test/resources/sql/misc.iq
```diff
@@ -2658,4 +2658,151 @@ FROM "hr"."emps";
...
+| EXPR$0 | EXPR$1 | EXPR$2 | EXPR$3 | EXPR$4 | EXPR$5 | EXPR$6 | EXPR$7 | EXPR$8 | EXPR$9 |
++--------+--------+--------+--------+--------+--------+--------+--------+--------+--------+
+|      6 |      4 |      2 |      1 |      1 |     -1 |     -1 |     -2 |     -3 |     16 |
++--------+--------+--------+--------+--------+--------+--------+--------+--------+--------+
+(1 row)
+
+!ok
+
+
+select cast(5.5 as smallint),
```
julianhyde, line 2705: for each statement, add comments explaining what rounding mode is being used

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
```diff
@@ -384,9 +384,15 @@ static void checkRoundedRange(Number value, double min, double max) {
     }
   }
 
+  /** Called from BuiltInMethod.INTEGER_CAST */
```
julianhyde, line 387: javadoc is same as next method

if methods have the same description, are both required?

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
```diff
@@ -477,28 +483,41 @@ static BigDecimal checkOverflow(BigDecimal value, int precision, int scale) {
     return checkOverflow(decimal, precision, scale);
   }
 
+  public @Nullable Object numberValue(Number value) {
+    return numberValue(value, RoundingMode.DOWN);
+  }
   /**
```
julianhyde, line 489: need space between methods

core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
```diff
@@ -583,4 +585,20 @@ public interface SqlConformance {
...
+   * <p>Among the built-in conformance levels,
+   * {@link RoundingMode#HALF_UP} in
+   * {@link SqlConformanceEnum#MYSQL_5},
+   * {@link SqlConformanceEnum#ORACLE_10},
+   * {@link SqlConformanceEnum#ORACLE_12};
+   * {@link RoundingMode#DOWN} otherwise.
+   *
+   * <p> {@link org.apache.calcite.sql.SqlDialect.DatabaseProduct#CALCITE}'s
+   * default is {@link RoundingMode#DOWN} and {@link org.apache.calcite.sql.SqlDialect.DatabaseProduct#POSTGRESQL},
+   * {@link org.apache.calcite.sql.SqlDialect.DatabaseProduct#MYSQL} and {@link org.apache.calcite.sql.SqlDialect.DatabaseProduct#ORACLE} is {@link RoundingMode#HALF_UP}.
```
julianhyde, line 601: line is too long and can be wrapped

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -1435,6 +1435,14 @@ public Expression handle(Expression x) {
...
 
+  static Expression getDefaultValue(Type type, RoundingMode roundingMode) {
+    Primitive p = Primitive.of(type);
+    if (p != null) {
+      return Expressions.constant(p.defaultValue, type, roundingMode);
+    }
+    return Expressions.constant(null, type);
+  }
+
   static Expression getDefaultValue(Type type) {
```
julianhyde, line 1446: should the method without `roundingMode` be deprecated?

core/src/main/java/org/apache/calcite/rex/RexExecutorImpl.java
```diff
@@ -56,11 +56,19 @@
 public class RexExecutorImpl implements RexExecutor {
 
   private final DataContext dataContext;
+  private final SqlConformance sqlConformance;
 
   public RexExecutorImpl(DataContext dataContext) {
     this.dataContext = dataContext;
+    this.sqlConformance = SqlConformanceEnum.DEFAULT;
```
julianhyde, line 63: I don't think conformance belongs here. Conformance governs validation behavior, not runtime behavior.

core/src/main/java/org/apache/calcite/rex/RexExecutorImpl.java
```diff
@@ -133,7 +152,7 @@ public static RexExecutable getExecutable(RexBuilder rexBuilder, List<RexNode> e
     try {
       code = compile(rexBuilder, constExps, (list, index, storageType) -> {
         throw new UnsupportedOperationException();
-      });
+      }, sqlConformance);
```
julianhyde, line 155: code is more readable if the lambda is the last parameter

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -505,9 +505,9 @@ private Expression getConvertExpression(
     case SMALLINT: {
       if (SqlTypeName.NUMERIC_TYPES.contains(sourceType.getSqlTypeName())) {
         return Expressions.call(
-            BuiltInMethod.INTEGER_CAST.method,
+            BuiltInMethod.INTEGER_CAST_ROUNDING_MODE.method,
```
mihaibudiu, line 508: this should apply to casts from decimal to decimal above

core/src/main/java/org/apache/calcite/sql/validate/SqlAbstractConformance.java
```diff
@@ -144,4 +146,8 @@ public abstract class SqlAbstractConformance implements SqlConformance {
   @Override public boolean allowLenientCoercion() {
     return SqlConformanceEnum.DEFAULT.allowLenientCoercion();
   }
+
+  @Override public RoundingMode roundingMode() {
+    return SqlConformanceEnum.DEFAULT.roundingMode();
```
mihaibudiu, line 151: I think rounding should be part of the type system

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
```diff
@@ -315,6 +316,8 @@ public enum BuiltInMethod {
   FP_DECIMAL_CAST(Primitive.class, "fpDecimalCast",
       Number.class, int.class, int.class),
   INTEGER_CAST(Primitive.class, "integerCast", Primitive.class, Object.class),
+  INTEGER_CAST_ROUNDING_MODE(Primitive.class, "integerCast",
```
mihaibudiu, line 319: The other cast method should probably be removed.
but many other cast methods have to be considered, including DECIMAL to DECIMAL, FP to integer, FP to decimal, DECIMAL to FP and perhaps even string to decimal

mihaibudiu (commented):
This is sorely needed work, thank you for doing it.

core/src/main/java/org/apache/calcite/sql/validate/SqlAbstractConformance.java
julianhyde: I agree.

I commented elsewhere that including `SqlConformance` in `RexExecutor` made me feel uneasy. `SqlConformance` is about what is allowed in valid SQL.

core/src/main/java/org/apache/calcite/util/BuiltInMethod.java
NobiGo: Yes. According to your advice, I found some places that use the default `RoudingMode.DOWN`. I'll see what I can do.

core/src/main/java/org/apache/calcite/sql/validate/SqlAbstractConformance.java
NobiGo: @julianhyde @mihaibudiu I will move the `roundingMode` to `RelDataTypeSystem`. Thanks for the advice.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
```diff
@@ -385,22 +385,20 @@ static void checkRoundedRange(Number value, double min, double max) {
   }
 
   /** Called from BuiltInMethod.INTEGER_CAST */
-  public static @Nullable Object integerCast(Primitive primitive, final Object value) {
-    return requireNonNull(primitive, "primitive").numberValue((Number) value);
+  public static @Nullable Object integerCast(Primitive primitive, final Object value,
```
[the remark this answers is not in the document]
mihaibudiu: an alternative is to leave the old API around and just add the new one.
I prefer this solution, but it may cause problems for other projects that try to upgrade to a new version of Calcite.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
```diff
@@ -412,69 +410,75 @@ static BigDecimal checkOverflow(BigDecimal value, int precision, int scale) {
...
+      @Nullable Number value, int precision, int scale, RoundingMode roundingMode) {
     if (value == null) {
       return null;
     }
     final BigDecimal decimal = BigDecimal.valueOf(value.doubleValue());
-    return checkOverflow(decimal, precision, scale);
+    return checkOverflow(decimal, precision, scale, roundingMode);
+  }
+
+  public @Nullable Object numberValue(Number value) {
```
mihaibudiu, line 480: This looks like a dangerous API. 
If you insist on having this function, I would call it numberValueRoundDown to make it clear what it does.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -1435,6 +1435,14 @@ public Expression handle(Expression x) {
     }
   }
 
+  static Expression getDefaultValue(Type type, RoundingMode roundingMode) {
```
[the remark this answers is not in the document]
NobiGo: I will revert the changes and check the unit test.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
```diff
@@ -35,6 +37,13 @@ public interface RelDataTypeSystem {
   /** Default type system. */
   RelDataTypeSystem DEFAULT = new RelDataTypeSystemImpl() { };
 
+  /** Same as {@link #DEFAULT}. But the rounding mode is {@link RoundingMode#HALF_UP}. */
+  RelDataTypeSystem DEFAULT_ROUNDING_MODE_HALF_UP = new RelDataTypeSystemImpl() {
```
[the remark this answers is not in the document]
NobiGo: Yes, Calclite‘s default rounding behavior is DOWN. I added this for the unit test.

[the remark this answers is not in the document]
mihaibudiu: then I think this should be in a test class

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystemImpl.java
```diff
@@ -160,6 +162,10 @@ public abstract class RelDataTypeSystemImpl implements RelDataTypeSystem {
     return 19;
   }
 
+  @Override public RoundingMode roundingMode() {
```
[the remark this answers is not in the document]
NobiGo: Do you mean using the default RoundingMode? There are already APIs in Calcite that use DOWN, so I set the default rounding is DOWN. `Primitive#numberValue` uses the DOWN mode to convert the decimal、float and double to long. WDYT？ And I check the `BigDecimal#intValue` uses the DOWN to convert number to int. So the default RoundingMode is DOWN.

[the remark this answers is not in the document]
mihaibudiu: Yes, that's why I expected the DEFAULT to be DOWN.
But if DEFAULT was used only for testing and it will move to a testing class, this is fine.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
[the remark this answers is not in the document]
NobiGo: Ok. I will add a new one.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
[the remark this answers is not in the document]
NobiGo: I will move it to a testing class.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
NobiGo: Revert code.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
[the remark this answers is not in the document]
mihaibudiu: I really don't see why it has to be in this class for quidem, in fact, the way you setup the type system I think you could move it to the CoreQuidemTest class.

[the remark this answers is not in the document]
NobiGo: Done.

mihaibudiu (approved):
There may be other places where rounding happens silently, but I guess we won't discover these easily. (For example I think BigDecimal.longValue does rounding silently). 
One way to discover them would be to write a lot of tests that cast every type possible to every type possible that requires rounding, using multiple values around the corner cases. But that could be left for a separate PR.

NobiGo:
> There may be other places where rounding happens silently, but I guess we won't discover these easily. (For example I think BigDecimal.longValue does rounding silently). One way to discover them would be to write a lot of tests that cast every type possible to every type possible that requires rounding, using multiple values around the corner cases. But that could be left for a separate PR.

@mihaibudiu Yes. I know. Many places still use default APIs, which are not completely consistent with this configuration. We are currently unable to assess the boundaries of this issue. Once we discover the problem, we will submit a new PR to fix it.

mihaibudiu:
I have tagged this with "merge soon", let's see if anyone else has comments.
Perhaps you can squash the commits too, since the PR isn't too large and no other people have commented yet.

PR #3918, opened 2024-08-16
Remove FENNEL VM; enable some disabled tests

mihaibudiu:
The VmName.FENNEL does not seem to be used anywhere, so I deleted it.
I also re-enabled a few SqlOperatorTest tests that work given recent fixes.
And fixed a tiny bug in the regex quoting code.

There are many more disabled tests that should be audited. I audited some of them, and they still fail (or I would have enabled them).

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/sql/test/SqlOperatorFixture.java
+++ b/testkit/src/main/java/org/apache/calcite/sql/test/SqlOperatorFixture.java
@@ -101,7 +101,7 @@ public interface SqlOperatorFixture extends AutoCloseable {
    * Name of a virtual machine that can potentially implement an operator.
    */
   enum VmName {
-    FENNEL, JAVA, EXPAND
+    JAVA, EXPAND
   }
 
   //~ Methods ----------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


caicancai:
I think this PR title is weird, maybe Remove FENNEL VM (enable some disabled tests) is better.

NobiGo:
Hi @mihaibudiu, I want to know Why we can remove this and What FENNEL means here. Some test cases used this, while others used JAVA and EXPAND.

mihaibudiu:
Based on the appearances fennel is some alternative to the enumerable based executor which is long gone. @julianhyde ?

mihaibudiu:
Can I merge this PR?

NobiGo:
> Can I merge this PR?

Hi, @mihaibudiu. I found that VolcanoPlanner has some documents that use the FENNEL, like `FennelAggRel.FENNEL_EXEC`. This should refer to the same execution engine. Do we need to remove or modify that?  I guess FENNEL is a programming language like java. Perhaps in the early stages of design, we plan to use FENNEL to implement it. some info:
1) https://forum.defold.com/t/using-fennel-language-with-defold/69083
2) https://fennel-lang.org/

mihaibudiu:
I don't think it's the same Fennel, the line I removed was added to Calcite in 2012, but the Fennel language seems to have started in 2016.

NobiGo:
> I don't think it's the same Fennel, the line I removed was added to Calcite in 2012, but the Fennel language seems to have started in 2016.

Ok. Then I am ok to merge this.

NobiGo (approved):

PR #3919, opened 2024-08-17
[CALCITE-6533] Division between integer and decimal produces incorrect result

mihaibudiu:
Turns out it's a bug I introduced during the fix for https://issues.apache.org/jira/browse/CALCITE-6464

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
+++ b/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
@@ -297,6 +297,7 @@ public interface RelDataTypeSystem {
         int six = Math.min(6, getMaxNumericScale());
         int d = p1 - s1 + s2;
         int scale = Math.max(six, s1 + p2 + 1);
+        scale = Math.min(scale, getMaxNumericScale());
         int precision = d + scale;
 
   // Rules from
@@ -320,7 +321,7 @@ public interface RelDataTypeSystem {
         //   can't fit into 32 digits.
         int bound = getMaxNumericPrecision() - six;  // This was '32' in the MS documentation
         if (precision <= bound) {
-          scale = Math.min(scale, getMaxNumericScale() - (precision - scale));
+          scale = Math.min(scale, getMaxNumericPrecision() - (precision - scale));
         } else {
           // precision > bound
           scale = Math.min(six, scale);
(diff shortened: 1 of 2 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -2661,6 +2663,19 @@ void checkModOperatorDivByZero(SqlOperatorFixture f) {
...
+              @Override public int getMaxNumericPrecision() {
+                return 28;
+              }
+
+              @Override public int getMaxNumericScale() {
+                return 10;
+              }
+            }));
+    f0.checkScalarExact("95.0 / 100", "DECIMAL(12, 10) NOT NULL", "0.95");
+    f0.checkScalarExact("95 / 100.0", "DECIMAL(17, 6) NOT NULL", "0.95");
```
caicancai, line 2678: Maybe adding another test would be better such as` 95/100`

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
```diff
@@ -297,6 +297,7 @@ default boolean shouldUseDoubleMultiplication(RelDataTypeFactory typeFactory,
         int six = Math.min(6, getMaxNumericScale());
         int d = p1 - s1 + s2;
         int scale = Math.max(six, s1 + p2 + 1);
+        scale = Math.min(scale, getMaxNumericScale());
```
caicancai, line 300: I don't quite understand the rules, but I see some paradoxes.
Assume the following scenario.
```
six = Math.min(6, getMaxNumericScale()) = getMaxNumericScale()
scale = Math.max(six, s1 + p2 + 1) = six = getMaxNumericScale()
scale = Math.min(scale, getMaxNumericScale()) = getMaxNumericScale()
```
Is this a repeated judgment?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: We are only fixing a bug for decimals, not for integers.

mihaibudiu:
Since this is a very small bugfix for a bug I have introduced, I will merge it without review.

PR #3921, opened 2024-08-18
[CALCITE-6534] Adjust type when pulling up Calc in JoinUnifyRule

suibianwanwank:
This PR addresses both [CALCITE-6534] and [CALCITE-6528] and is based on [CALCITE-6501]. For the problem of nullablility due to Join null-side, if RexNode satisfies null-if-null, we will adjust the nullability, if not, we should reject this rewrite regardless of the nullablility.

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
+++ b/core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
@@ -77,6 +77,7 @@ import com.google.common.collect.Sets;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -127,6 +128,14 @@ import static java.util.Objects.requireNonNull;
  */
 public class SubstitutionVisitor {
   private static final boolean DEBUG = CalciteSystemProperty.DEBUG.value();
+  private static final Strong STRONG = new Strong() {
+    @Override public boolean isNull(RexInputRef ref) {
+      // Here strong is used to determine if the nullable side calc can be pulled up to the join,
+      // and we will adjust nullability if RexInputRef is not null,
+      // so here RexInputRef is always satisfies null-if-null
+      return true;
+    }
+  };
 
   public static final ImmutableList<UnifyRule> DEFAULT_RULES =
       ImmutableList.of(
@@ -1209,6 +1218,7 @@ public class SubstitutionVisitor {
       final MutableJoin query = (MutableJoin) call.query;
       final MutableCalc qInput0 = (MutableCalc) query.getLeft();
       final MutableRel qInput1 = query.getRight();
+      MutableRel qInput0Input = qInput0.getInput();
       final Pair<RexNode, List<RexNode>> qInput0Explained = explainCalc(qInput0);
       final RexNode qInput0Cond = qInput0Explained.left;
       final List<RexNode> qInput0Projs = qInput0Explained.right;
@@ -1222,13 +1232,8 @@ public class SubstitutionVisitor {
       if (joinRelType == null) {
         return null;
       }
-      // Check if filter under join can be pulled up.
-      if (!canPullUpFilterUnderJoin(joinRelType, qInput0Cond, null)) {
-        return null;
-      }
-
-      // Check if join project nullable can be pull up
-      if (!canPullUpProjectUnderJoin(joinRelType, qInput0Projs, null)) {
+      // Check if calc under join can be pulled up.
+      if (!canPullUpCalcUnderJoin(joinRelType, qInput0Explained, null)) {
         return null;
       }
 
@@ -1246,7 +1251,7 @@ public class SubstitutionVisitor {
             final int newIdx = ((RexInputRef) qInput0Projs.get(idx)).getIndex();
             return new RexInputRef(newIdx, inputRef.getType());
           } else {
-            int newIdx = idx - fieldCnt(qInput0) + fieldCnt(qInput0.getInput());
+            int newIdx = idx - fieldCnt(qInput0) + fieldCnt(qInput0Input);
             return new RexInputRef(newIdx, inputRef.getType());
           }
         }
@@ -1257,16 +1262,10 @@ public class SubstitutionVisitor {
       // MutableJoin matches only when the conditions are analyzed to be same.
       if (splitted != null && splitted.isAlwaysTrue()) {
         final RexNode compenCond = qInput0Cond;
-        final List<RexNode> compenProjs = new ArrayList<>();
-        for (int i = 0; i < fieldCnt(query); i++) {
-          if (i < fieldCnt(qInput0)) {
-            compenProjs.add(qInput0Projs.get(i));
-          } else {
-            final int newIdx = i - fieldCnt(qInput0) + fieldCnt(qInput0.getInput());
-            compenProjs.add(
-                new RexInputRef(newIdx, query.rowType.getFieldList().get(i).getType()));
-          }
-        }
+        final List<RexNode> compenProjs =
+            shiftAndAdjustProjectExpr(query, target, rexBuilder,
+                qInput0Projs, qInput0Input.rowType.getFieldList(), null, null);
+
         final RexProgram compenRexProgram =
             RexProgram.create(target.rowType, compenProjs, compenCond,
                 query.rowType, rexBuilder);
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -127,6 +128,11 @@
  */
 public class SubstitutionVisitor {
   private static final boolean DEBUG = CalciteSystemProperty.DEBUG.value();
+  private static final Strong STRONG = new Strong() {
```
mihaibudiu, line 131: This deserves a small comment.  why are input refs assumed to be always null?
I guess it's because you will apply it to some specific types of expressions.

core/src/test/java/org/apache/calcite/test/MaterializedViewSubstitutionVisitorTest.java
```diff
@@ -1138,8 +1138,13 @@ protected final MaterializedViewFixture sql(String materialize,
 
   /** Test case for
    * <a href="https://issues.apache.org/jira/browse/CALCITE-6501">[CALCITE-6501]
-   * Assertion Error in JoinUnifyRule Due to Type Mismatch</a>. */
-  @Test void testLeftProjectOnRightJoinToJoinFailed() {
+   * Assertion Error in JoinUnifyRule Due to Type Mismatch</a>.
+   *
+   * <p>Originally, this test was noMat due to a type mismatch.
```
mihaibudiu, line 1143: I don't think this kind of comment is useful long term. It is useful for the reviewers of this PR, but not for maintaining the code. This comment would be better as a github comment rather than a code comment. Same for the subsequent comments.

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -2169,57 +2142,106 @@ public static boolean equalType(String desc0, MutableRel rel0, String desc1,
...
     }
-    return true;
+    return compenProjs;
   }
 
-  /** Returns whether all projects are nullable. */
-  private static boolean allProjectsNullable(List<RexNode> projects) {
-    return projects.stream()
-        .allMatch(project -> project.getType().isNullable());
+  /** Cast RexNode to the given type if only nullability differs. */
```
mihaibudiu, line 2233: I would also add "Otherwise throw". The current comment implies that in all other cases the type is left unchanged.

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -2169,57 +2142,106 @@ public static boolean equalType(String desc0, MutableRel rel0, String desc1,
...
+    if (qInput1Explained != null
+        && joinType.generatesNullsOn(1)
+        && !isCalcStrong(qInput1Explained)) {
+      return false;
     }
-    return false;
+    return true;
+  }
+
+  /** Determines if all projects are strong and the condition is always true. */
```
mihaibudiu, line 2167: not clear what the "projects" are and what the condition is given single input parameter. Maybe you can add `@param` to describe where inputExplained comes from.

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -2169,57 +2142,106 @@ public static boolean equalType(String desc0, MutableRel rel0, String desc1,
...
-  private static boolean canPullUpProjectUnderJoin(JoinRelType joinType,
-      @Nullable List<RexNode> leftProjects, @Nullable List<RexNode> rightProjects) {
-    if (leftProjects != null && joinType.generatesNullsOn(0)
-        && !allProjectsNullable(leftProjects)) {
-      return false;
-    }
-    if (rightProjects != null && joinType.generatesNullsOn(1)
-        && !allProjectsNullable(rightProjects)) {
-      return false;
+  private static List<RexNode> shiftAndAdjustProjectExpr(MutableJoin query, MutableJoin target,
```
mihaibudiu, line 2184: I think JavaDoc at least for some parameters could also help.
In particular, what is the target?

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -2169,57 +2142,106 @@ public static boolean equalType(String desc0, MutableRel rel0, String desc1,
...
+
+    int[] adjustments0 = new int[target.rowType.getFieldCount()];
+    int[] adjustments1 = new int[target.rowType.getFieldCount()];
+
+    if (qInput1Projs != null) {
+      Arrays.fill(adjustments1, fieldCnt(target.getLeft()));
+    }
+
+    // In cases such as JoinOnLeftCalcToJoinUnifyRule and JoinOnCalcsToJoinUnifyRule rules,
+    // the left calc need to be pulled up the target, initialize the converter.
```
mihaibudiu, line 2197: should this be "where the left calc needs to be pulled up"?

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -2169,57 +2142,106 @@ public static boolean equalType(String desc0, MutableRel rel0, String desc1,
...
+        new RelOptUtil.RexInputConverter(rexBuilder, qInput0InputFields,
+            target.rowType.getFieldList(), adjustments0);
+
+    // In cases such as JoinOnRightCalcToJoinUnifyRule and JoinOnCalcsToJoinUnifyRule rules,
+    // the right calc need to be pulled up the target, initialize the converter.
+    RelOptUtil.RexInputConverter converter1 =
+        new RelOptUtil.RexInputConverter(rexBuilder, qInput1InputFields,
+            target.rowType.getFieldList(), adjustments1);
+
+    final List<RexNode> compenProjs = new ArrayList<>();
```
mihaibudiu, line 2208: what does "compen" mean?

mihaibudiu (commented):
I personally I am not very familiar with this part of the codebase so I am not confident enough to approve. Based on the tests, this seems to work very well, though.
The code looks generally fine, I left most comments that could improve readability.

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
suibianwanwank: This retains the naming of the previous code, I'm guessing this means compensation.

suibianwanwank:
@mihaibudiu Thanks for the review, this function both shift and type adjustments, and I think it's really not easy to understand, so I've optimized those comments and added more of them.

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -2169,57 +2145,155 @@ public static boolean equalType(String desc0, MutableRel rel0, String desc1,
...
+   * based on the provided join targets and inputs.
+   *
+   * <p>Used in the Join rewrite to pull up the calc in query
+   * to the join in mv to ensure operator equivalence. (Already make sure that pull up is valid).
    * Working in rules: {@link JoinOnLeftCalcToJoinUnifyRule} <br/>
    * {@link JoinOnRightCalcToJoinUnifyRule} <br/>
    * {@link JoinOnCalcsToJoinUnifyRule} <br/>
+   *
+   * @param query MutableRel of query
+   * @param target MutableRel of target
```
mihaibudiu, line 2192: This information can be already obtained from the parameter declaration, the question I had is "what is the target?" Is the target the place where you are building the modified query?

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
suibianwanwank: The naming here comes from UnifyRuleCall, UnifyRule is bottom-down matching, target means the RelNode of the mv that is trying to match. 
If you're interested in this part, [Refinement for Substitution-Based Materialized View Matching](https://docs.google.com/document/d/1JpwGNFE3hw3yXb7W3-95-jXKClZC5UFPKbuhgYDuEu4) is a great resource

mihaibudiu: Ideally that google doc would be posted somewhere publicly and linked from the code here.

mihaibudiu:
You can squash the commits for merging

core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
suibianwanwank: Sounds good, it can be done in other issue

suibianwanwank:
> You can squash the commits for merging

Done✅

PR #3923, opened 2024-08-19
AbstractNamespace unwrap add isInstance judgments

ChengJie1053:
Some other classes add isInstance judgments before calling clazz.cast

image

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/AbstractNamespace.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/AbstractNamespace.java
@@ -189,8 +189,11 @@ abstract class AbstractNamespace implements SqlValidatorNamespace {
     return true;
   }
 
-  @Override public <T> T unwrap(Class<T> clazz) {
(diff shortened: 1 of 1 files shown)


ChengJie1053:
> However, the checker framework is upset because this function is not allowed to return null.

Ok, thanks for reviewing the code for me

caicancai:
@ChengJie1053 Can you explain this problem through a counter-example test? Otherwise, I think it will be difficult for many users or contributors to understand.

ChengJie1053:
> Ok, thanks for reviewing the code for me

Ok, thanks for reviewing the code for me

mihaibudiu:
You should squash the commits so we can merge this PR.

ChengJie1053:
> Ok, thanks for reviewing the code for me

Ok, thanks for reviewing the code for me

ChengJie1053:
> You should squash the commits so we can merge this PR.

Ok, thanks for reviewing the code for me

NobiGo (approved):

PR #3924, opened 2024-08-20
[CALCITE-6540] RelOptUtil.pushDownJoinConditions does not correctly...

mihaibudiu:
…adjust ASOF joins match conditions


caicancai (approved):
LGTM

PR #3925, opened 2024-08-20
[CALCITE-5883] ROWS window aggregates ignore frames when there is no ORDER BY clause

julianhyde:


core/src/test/resources/sql/winagg.iq
```diff
@@ -78,6 +78,87 @@ from emp;
...
+|     2 |     4 |
+|     2 |     4 |
++-------+-------+
+(14 rows)
+
+!ok
+
+!if (false) {
+# The result below is from Postgres.
+# Calcite currently gives the wrong results for s1p (first 2 rows),
```
mihaibudiu, line 129: do you know if there is an issue for these cases?

mihaibudiu (approved):

caicancai (approved):

core/src/test/resources/sql/winagg.iq
julianhyde: There is a problem, but there is no separate jira case. In my first commit I wanted to avoid the issue. In my 2nd/3rd commits I addressed it and enabled the test.

So I consider that CALCITE-5883 (and this fix) covers all of these related problems.

mihaibudiu: I asked a similar question on the mailing list a while ago, I wonder whether that test will work now too:
https://lists.apache.org/thread/jhzxf7k7wxgwv7szp5lj9mkocjq3n5pf

PR #3926, opened 2024-08-21
Typos in the document

NobiGo:


mihaibudiu (approved):
I think that @julianhyde suggested not to use the [MINOR] annotation on a commit message at some point. Other than that this is fine.

mihaibudiu:
I think you can probably merge such fixes even without review.

PR #1746, opened 2020-01-11
[CALCITE-3697] Implement BITCOUNT function

XuQianJin-Stars:
As illustrated in [CALCITE-3697](https://issues.apache.org/jira/browse/CALCITE-3697)
select bit_count(8);
+--------+
| EXPR$0 |
+--------+
|      1        |
+--------+
(1 row)

!ok

PR #3927, opened 2024-08-21
[CALCITE-3697] Implement BITCOUNT scalar function

normanj-bitquill:
* All libraries now have a BITCOUNT() function
* The MySQL library also has BIT_COUNT() which is an alias for BITCOUNT()
* BITCOUNT() counts all of the bits set in an integer or bytestring
* Return NULL if the argument is NULL

This is based on a previous PR that appears dormant.
https://github.com/apache/calcite/pull/1746

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -2834,6 +2834,43 @@ public class SqlFunctions {
     return binaryOperator(b0, b1, (x, y) -> (byte) (x & y));
   }
 
+  /** Helper function for implementing <code>BITCOUNT</code>. Counts the number
+   * of bits set in an integer value. */
+  public static long bitCount(long b) {
+    return Long.bitCount(b);
+  }
+
+  private static final BigDecimal BITCOUNT_MAX = new BigDecimal(2).pow(64)
+      .subtract(new BigDecimal(1));
+  private static final BigDecimal BITCOUNT_MIN = new BigDecimal(2).pow(63).negate();
+
+  /** Helper function for implementing <code>BITCOUNT</code>. Counts the number
+   * of bits set in the integer portion of a decimal value. */
+  public static long bitCount(BigDecimal b) {
+    final int comparison = b.compareTo(BITCOUNT_MAX);
+    if (comparison < 0) {
+      if (b.compareTo(BITCOUNT_MIN) <= 0) {
+        return 1;
+      } else {
+        return bitCount(b.setScale(0, RoundingMode.DOWN).longValue());
+      }
+    } else if (comparison == 0) {
+      return 64;
+    } else {
+      return 63;
+    }
+  }
+
+  /** Helper function for implementing <code>BITCOUNT</code>. Counts the number
+   * of bits set in a ByteString value. */
+  public static long bitCount(ByteString b) {
+    long bitsSet = 0;
+    for (int i = 0; i < b.length(); i++) {
+      bitsSet += Integer.bitCount(0xff & b.byteAt(i));
+    }
+    return bitsSet;
+  }
+
   /** Bitwise function <code>BIT_OR</code> applied to integer values. */
   public static long bitOr(long b0, long b1) {
     return b0 | b1;
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -71,6 +71,7 @@ import static org.apache.calcite.sql.fun.SqlLibrary.POSTGRESQL;
 import static org.apache.calcite.sql.fun.SqlLibrary.REDSHIFT;
 import static org.apache.calcite.sql.fun.SqlLibrary.SNOWFLAKE;
 import static org.apache.calcite.sql.fun.SqlLibrary.SPARK;
+import static org.apache.calcite.sql.fun.SqlStdOperatorTable.BITCOUNT;
 import static org.apache.calcite.sql.type.OperandTypes.STRING_FIRST_OBJECT_REPEAT;
 import static org.apache.calcite.sql.type.OperandTypes.STRING_FIRST_STRING_ARRAY_REPEAT;
 import static org.apache.calcite.util.Static.RESOURCE;
@@ -1122,6 +1123,21 @@ public abstract class SqlLibraryOperators {
   @LibraryOperator(libraries = {SPARK, HIVE, MYSQL})
   public static final SqlSpecialOperator NOT_RLIKE =
       new SqlLikeOperator("NOT RLIKE", SqlKind.RLIKE, true, true);
+
+  /** Alias for {@link SqlStdOperatorTable#BITCOUNT}. */
+  @LibraryOperator(libraries = {BIG_QUERY, SPARK})
+  public static final SqlFunction BIT_COUNT_BIG_QUERY =
+      BITCOUNT.withName("BIT_COUNT");
+
+  @LibraryOperator(libraries = {MYSQL})
+  public static final SqlFunction BIT_COUNT_MYSQL =
+      new SqlFunction(
+          "BIT_COUNT",
+          SqlKind.OTHER_FUNCTION,
+          ReturnTypes.BIGINT_NULLABLE,
+          null,
+          OperandTypes.NUMERIC.or(OperandTypes.BINARY),
+          SqlFunctionCategory.NUMERIC);
 
   /** The "CONCAT(arg, ...)" function that concatenates strings.
    * For example, "CONCAT('a', 'bc', 'd')" returns "abcd".
--- a/core/src/test/resources/sql/functions.iq
+++ b/core/src/test/resources/sql/functions.iq
@@ -18,6 +18,19 @@
 !use mysqlfunc
 !set outputformat mysql
 
+# BIT Functions
+
+# BIT_COUNT
+select bit_count(8);
++--------+
+| EXPR$0 |
++--------+
+|      1 |
++--------+
+(1 row)
+
+!ok
+
 # MATH Functions
 
 # CBRT
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2752,6 +2752,10 @@ In the following:
 | * | ATANH(numeric)                                 | Returns the inverse hyperbolic tangent of *numeric*
 | f | BITAND_AGG(value)                              | Equivalent to `BIT_AND(value)`
 | f | BITOR_AGG(value)                               | Equivalent to `BIT_OR(value)`
+| * | BITCOUNT(value)                                | Returns the bitwise COUNT of *value* or NULL if *value* is NULL. *value* must be and integer or binary value.
+| b s | BIT_COUNT(integer)                           | Returns the bitwise COUNT of *integer* or NULL if *integer* is NULL
+| m | BIT_COUNT(numeric)                             | Returns the bitwise COUNT of the integer portion of *numeric* or NULL if *numeric* is NULL
+| b m s | BIT_COUNT(binary)                          | Returns the bitwise COUNT of *binary* or NULL if *binary* is NULL
 | s | BIT_LENGTH(binary)                             | Returns the bit length of *binary*
 | s | BIT_LENGTH(string)                             | Returns the bit length of *string*
 | s | BIT_GET(value, position)                       | Returns the bit (0 or 1) value at the specified *position* of numeric *value*. The positions are numbered from right to left, starting at zero. The *position* argument cannot be negative
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -15523,6 +15523,71 @@ public class SqlOperatorTest {
     f0.forEachLibrary(list(functionAlias.libraries), consumer);
   }
 
+  @Test void testBitCountFunc() {
+    checkBitCount(SqlStdOperatorTable.BITCOUNT, null, false);
+  }
+
+  @Test void testBitCountBigQueryFunc() {
+    checkBitCount(SqlLibraryOperators.BIT_COUNT_BIG_QUERY,
+        list(SqlLibrary.BIG_QUERY, SqlLibrary.SPARK), false);
+  }
+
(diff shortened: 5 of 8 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1123,6 +1123,16 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
   public static final SqlSpecialOperator NOT_RLIKE =
       new SqlLikeOperator("NOT RLIKE", SqlKind.RLIKE, true, true);
 
+  @LibraryOperator(libraries = {MYSQL})
+  public static final SqlFunction BIT_COUNT_MYSQL =
```
NobiGo, line 1127: If only the name is different, you can try API use a copy of this function with a given name, like:
```
  /** Alias for {@link #BIT_GET}. */
  public static final SqlFunction GETBIT =
      BIT_GET.withName("GETBIT");
```

core/src/test/resources/sql/functions.iq
```diff
@@ -18,6 +18,17 @@
 !use mysqlfunc
 !set outputformat mysql
 
+# BIT_COUNT
```
NobiGo, line 21:
```
# BIT Functions

# BIT_COUNT
```

site/_docs/reference.md
```diff
@@ -2752,6 +2752,8 @@ In the following:
 | * | ATANH(numeric)                                 | Returns the inverse hyperbolic tangent of *numeric*
 | f | BITAND_AGG(value)                              | Equivalent to `BIT_AND(value)`
 | f | BITOR_AGG(value)                               | Equivalent to `BIT_OR(value)`
+| * | BITCOUNT(integer)                             | Returns the bitwise COUNT of non-null *integer*
```
NobiGo, line 2755: Then the test case maybe should include a negative number and zero.

NobiGo: In MYSQL:
```
select bit_count(3.5);  return 1
select bit_count(3); return 2
select bit_count(5.5); return 2
```
So support float number?

NobiGo (changes requested):

site/_docs/reference.md
mihaibudiu: In calcite these are decimal numbers

core/src/test/resources/sql/functions.iq
normanj-bitquill: Updated.

site/_docs/reference.md
mihaibudiu: Since the decimal representation of a number is not specified, how can these functions be implemented?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
normanj-bitquill: Thanks. There is now a difference in parameters now as well.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -15523,6 +15523,58 @@ void checkBitAnd(SqlOperatorFixture f0, FunctionAlias functionAlias) {
     f0.forEachLibrary(list(functionAlias.libraries), consumer);
   }
 
+  @Test public void testBitCountFunc() {
```
julianhyde, line 15526: add javadoc. explain that one tests BITCOUNT and the other tests BIT_COUNT.

are these methods identical? do you intend them to stay identical? If so, share code, don't copy-paste. We have done that for other operators.

site/_docs/reference.md
```diff
@@ -2752,6 +2752,8 @@ In the following:
 | * | ATANH(numeric)                                 | Returns the inverse hyperbolic tangent of *numeric*
 | f | BITAND_AGG(value)                              | Equivalent to `BIT_AND(value)`
 | f | BITOR_AGG(value)                               | Equivalent to `BIT_OR(value)`
+| * | BITCOUNT(integer)                              | Returns the bitwise COUNT of non-null *integer*
+| m | BIT_COUNT(numeric)                             | Returns the bitwise COUNT of the integer portion of non-null *numeric*
```
julianhyde, line 2756: do these functions apply to binary, varbinary , decimal, float? if so, say so

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -15523,6 +15523,58 @@ void checkBitAnd(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+          "^bit_count()^",
+          "Invalid number of arguments to function 'BIT_COUNT'. Was expecting 1 arguments",
+          false);
+      f.checkFails(
+          "^bit_count(1, 2)^",
+          "Invalid number of arguments to function 'BIT_COUNT'. Was expecting 1 arguments",
+          false);
+      f.checkScalar("bit_count(8)", "1", "BIGINT NOT NULL");
+      f.checkScalar("bit_count(CAST(x'ad' AS BINARY(1)))", "5", "BIGINT NOT NULL");
+      f.checkScalar("bit_count(5.23)", "2", "BIGINT NOT NULL");
```
julianhyde, line 15571: so it ignores bits in the fractional part? that's rather subtle. worth calling out in the test, and also in the `bitCount(BigDecimal)` function

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -15523,6 +15523,58 @@ void checkBitAnd(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+        "^bitcount(1, 2)^",
+        "Invalid number of arguments to function 'BITCOUNT'. Was expecting 1 arguments",
+        false);
+    f.checkScalar("bitcount(8)", "1", "BIGINT NOT NULL");
+    f.checkScalar("bitcount(CAST(x'ad' AS BINARY(1)))", "5", "BIGINT NOT NULL");
+    f.checkNull("bitcount(cast(NULL as TINYINT))");
+    f.checkNull("bitcount(cast(NULL as BINARY))");
+  }
+
+  @Test public void testBitCountMySQLFunc() {
```
julianhyde, line 15552: what are the results when applied to negative numbers? need tests for that.

do you assume two's complement? do you assume that a 32 bit integer becomes a 32 bit negative integer? or does it become a 64 bit negative integer?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
normanj-bitquill: They are not identical. `BIT_COUNT()` supports decimal values. I viewed this as a unique feature in MySQL. If it makes more sense, I can use the MySQL behaviour for both `BITCOUNT()` and `BIT_COUNT()`.

site/_docs/reference.md
normanj-bitquill: The documentation has been updated. If it makes more sense, both `BITCOUNT()` and `BIT_COUNT()` can use the MySQL behaviour and become identical.

site/_docs/reference.md
normanj-bitquill: Updated the documentation to mention that it also supports binary and numeric (in the case of `BIT_COUNT()`).

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
normanj-bitquill: Yeah, it ignores the fractional part. I have updated the documentation as well as the Javadoc for `bitCount(BigDecimal)`.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
normanj-bitquill: The input number is converted to a 64-bit integer using two's complement. It will count the bits set in this, so `bitcount(-1)` will return `64`.

Added a test.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2834,6 +2834,28 @@ public static ByteString bitAnd(ByteString b0, ByteString b1) {
...
+  /** Helper function for implementing <code>BITCOUNT</code>. Counts the number
+   * of bits set in an integer value. */
+  public static long bitCount(long b) {
+    return Long.bitCount(b);
+  }
+
+  /** Helper function for implementing <code>BITCOUNT</code>. Counts the number
+   * of bits set in the integer portion of a decimal value. */
+  public static long bitCount(BigDecimal b) {
+    return Long.bitCount(b.longValue());
```
mihaibudiu, line 2846: This also rounds down the decimal number, which is not obvious.
Please also note that a decimal may not fit into a long since we merged https://github.com/apache/calcite/pull/3589
So I am not sure this implementation is correct

mihaibudiu: Personally I think this behavior is very unexpected, and I think decimals should not be supported at all.
Moreover, this is not the number of 1 bits in the actual internal representation, but in the conversion of the decimal to a long.

normanj-bitquill: I did some more testing with MySQL.
* bitcount(18446744073709551615) = 64
* bitcount(18446744073709551616 or greater) = 63
* bitcount(-9223372036854775808 or less) = 1

I have updating the implementation of `bitcount(BigDecimal)` to match this behaviour. This is only enabled in the MySQL library.

It should match MySQL behaviour at the moment, but since this is not really documented
https://dev.mysql.com/doc/refman/8.4/en/bit-functions.html#function_bit-count
perhaps this is not worth keeping and could change at any time.

In general is it better to match MySQL behaviour (for a MySQL specific function) or is it better to take the simpler approach?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -15523,6 +15523,64 @@ void checkBitAnd(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+          false);
+      f.checkFails(
+          "^" + functionName + "(1, 2)^",
+          "Invalid number of arguments to function '" + functionName
+              + "'. Was expecting 1 arguments",
+          false);
+      f.checkScalar(functionName + "(8)", "1", "BIGINT NOT NULL");
+      f.checkScalar(functionName + "(CAST(x'ad' AS BINARY(1)))", "5", "BIGINT NOT NULL");
+      f.checkScalar(functionName + "(-1)", "64", "BIGINT NOT NULL");
+      f.checkNull(functionName + "(cast(NULL as TINYINT))");
```
mihaibudiu, line 15557: how about a test with a null without a cast?
this trips some function implementations

mihaibudiu (approved):

site/_docs/reference.md
caicancai: Are you sure Spark also has a bigcount function? I didn't find the corresponding record in the documentation

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
normanj-bitquill: Added a test for `NULL` without a cast.

site/_docs/reference.md
normanj-bitquill: `BIT_COUNT` is now enabled for Apache Spark.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2834,6 +2834,42 @@ public static ByteString bitAnd(ByteString b0, ByteString b1) {
     return binaryOperator(b0, b1, (x, y) -> (byte) (x & y));
   }
 
+  /** Helper function for implementing <code>BITCOUNT</code>. Counts the number
+   * of bits set in an integer value. */
+  public static long bitCount(long b) {
+    return Long.bitCount(b);
+  }
+
+  private static final BigDecimal BITCOUNT_MAX = new BigDecimal("18446744073709551615");
```
NobiGo, line 2843: Why this number？

site/_docs/reference.md
```diff
@@ -2752,6 +2752,11 @@ In the following:
 | * | ATANH(numeric)                                 | Returns the inverse hyperbolic tangent of *numeric*
 | f | BITAND_AGG(value)                              | Equivalent to `BIT_AND(value)`
 | f | BITOR_AGG(value)                               | Equivalent to `BIT_OR(value)`
+| * | BITCOUNT(integer)                              | Returns the bitwise COUNT of *integer* or NULL if *integer* is NULL
+| * | BITCOUNT(binary)                               | Returns the bitwise COUNT of *binary* of NULL if *binary* is NULL
```
NobiGo, line 2756: Merge these two functions into one?

NobiGo (approved):

site/_docs/reference.md
normanj-bitquill: Merged.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
normanj-bitquill: This is `2^64 - 1`. I found it through testing in MySQL. I added a comment to say that it is `2^64 - 1`.

normanj-bitquill:
Squashed the commits.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
NobiGo: Maybe we can use 
```
Math.pow(2,64) -1
```
generate it?

mihaibudiu: BigDecimal has s shiftleft method

caicancai (approved):
LGTM

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
normanj-bitquill: BigInteger has a shiftleft, BigDecimal does not. I have updated the constants to use BigDecimal power, subtract and negate.

NobiGo (approved):

normanj-bitquill:
@mihaibudiu Is there anything further needed on this PR?

PR #3928, opened 2024-08-22
[CALCITE-4924] REGR_SXX and similar aggregate functions return the wr…

NobiGo:
…ong data type


core/src/test/resources/sql/dummy.iq
```diff
@@ -15,9 +15,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-!use post
-values 1;
+!use scott
```
NobiGo, line 18: This example is for review.

core/src/test/resources/sql/agg.iq
```diff
@@ -2981,16 +2981,41 @@ group by MONTH(HIREDATE);
...
+|     9 |    -175000.0000000000 |   490000.0000 |    31250.0000 |
 +-------+-----------------------+---------------+---------------+
 (8 rows)
 
 !ok
 
+# [CALCITE-4924] REGR_SXX and similar aggregate functions return the wrong data type
+SELECT
+  MONTH(HIREDATE) as "MONTH",
+  covar_samp(SAL, COMM) as "COVAR_SAMP(SAL, COMM)",
```
NobiGo, line 2997: Compared to the previous one, this test can remove the cast.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystemImpl.java
```diff
@@ -255,8 +268,72 @@ && getDefaultPrecision(typeName) != -1) {
     return argumentType;
   }
 
-  @Override public RelDataType deriveCovarType(RelDataTypeFactory typeFactory,
+  @Override public RelDataType deriveCovarType(RelDataTypeFactory typeFactory, SqlKind sqlKind,
       RelDataType arg0Type, RelDataType arg1Type) {
+    switch (sqlKind) {
```
NobiGo, line 273: I'm not sure if this is a reasonable implementation, but from the test cases, it seems feasible.

mihaibudiu:
Today *all* Calcite aggregation functions (except COUNT) return the exact same data type as the data that is being aggregated. This behavior is permitted by the standard, and I think it's a reasonable one. From the Oracle documentation I think that Oracle behaves in the same way. Before I can review this PR we have to agree what you think the "correct" behavior is.

Maybe this should also be a property of the type system if you really want to emulate multiple databases.

NobiGo:
> Today _all_ Calcite aggregation functions (except COUNT) return the exact same data type as the data that is being aggregated. This behavior is permitted by the standard, and I think it's a reasonable one. From the Oracle documentation I think that Oracle behaves in the same way. Before I can review this PR we have to agree what you think the "correct" behavior is.
> 
> Maybe this should also be a property of the type system if you really want to emulate multiple databases.

This is my idea, and I am glad to receive your feedback on this issue. I will also check if other databases adopt a similar approach.

mihaibudiu:
Maybe we should discuss the design in the JIRA ticket, I think that's the expected way.
You may post on the dev list to attract more attention, this is an important issue I think.

PR #3929, opened 2024-08-22
[CALCITE-4512] GROUP BY expression with argument name same with SELECT field and alias causes validation error

hannerwang:
### **What changes were proposed in this pull request?**
Stop identifiers in a SQL call from being treated as aliases and expanded into select expressions.
For example:
`select replace(name, 'a', 'b') as name from users group by replace(name, 'a', 'b')`
The **name** in GROUP BY clause should not be replaced by `replace(name, 'a', 'b')`, it will be resolved as users.name,
i.e. 
`select replace(users.name, 'a', 'b') as name from users group by replace(users.name, 'a', 'b')`

### **Why are the changes needed?**
Make the sql semantics unchanged when treating identifiers within Group By clause as aliases.
current calcite behavior:
`select replace(**name**, 'a', 'b') as **name** from users group by replace(**name**, 'a', 'b')`
will be expanded into
`select replace(**name**, 'a', 'b') as **name** from users group by replace(**replace(name, 'a', 'b')**, 'a', 'b')`
the semantics has been changed.

### **Does this PR introduce any user-facing change?**
No for existing users

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
@@ -127,10 +127,12 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 
 import org.apiguardian.api.API;
+import org.checkerframework.checker.initialization.qual.UnknownInitialization;
 import org.checkerframework.checker.nullness.qual.KeyFor;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.checker.nullness.qual.PolyNull;
+import org.checkerframework.checker.nullness.qual.RequiresNonNull;
 import org.checkerframework.dataflow.qual.Pure;
 import org.slf4j.Logger;
 
@@ -7269,6 +7271,8 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
     final SqlSelect select;
     final SqlNode root;
     final Clause clause;
+    // Retain only expandable aliases or ordinals to prevent their expansion in a SQL call expr.
+    final Set<SqlNode> aliasOrdinalExpandSet = Sets.newIdentityHashSet();
 
     ExtendedExpander(SqlValidatorImpl validator, SqlValidatorScope scope,
         SqlSelect select, SqlNode root, Clause clause) {
@@ -7276,6 +7280,9 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
       this.select = select;
       this.root = root;
       this.clause = clause;
+      if (clause == Clause.GROUP_BY) {
+        addExpandableExpressions();
+      }
     }
 
     @Override public @Nullable SqlNode visit(SqlIdentifier id) {
@@ -7284,7 +7291,7 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
       }
 
       final boolean replaceAliases = clause.shouldReplaceAliases(validator.config);
-      if (!replaceAliases) {
+      if (!replaceAliases || (clause == Clause.GROUP_BY && !aliasOrdinalExpandSet.contains(id))) {
         final SelectScope scope = validator.getRawSelectScopeNonNull(select);
         SqlNode node = expandCommonColumn(select, id, scope, validator);
         if (node != id) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -7378,6 +7368,37 @@ static class ExtendedExpander extends Expander {
...
+    private void addExpandableExpressions(@UnknownInitialization ExtendedExpander this) {
+      switch (root.getKind()) {
+      case IDENTIFIER:
+      case LITERAL:
+        aliasOrdinalExpandSet.add(root);
+        break;
+      case GROUPING_SETS:
+      case ROLLUP:
+      case CUBE:
+        if (root instanceof SqlBasicCall) {
```
suibianwanwank, line 7385: Should SqlBasicCall be replaced with SqlCall here. e.g. SqlCase does not inherit SqlBasicCall.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -7378,6 +7368,37 @@ static class ExtendedExpander extends Expander {
...
+          List<SqlNode> operandList = ((SqlBasicCall) root).getOperandList();
```
suibianwanwank, line 7386: Same as above.

mihaibudiu:
@hannerwang I don't know if you have seen the comments in the JIRA.
This PR is very close to being done.
Do you think you can address the comments in time for the next release of Calcite?

I reproduce the comments here:

> Which DB are we trying to emulate the behavior of? If so, I would like to see a Quidem test that returns the same in Calcite and that DB, and would return the wrong result if the bug were not fixed. Perhaps something like select mod(empno, 3) as empno from emp group by mod(empno, 2) and make sure that we get the right number of rows.
>
> Also I would like to see a SqlValidatorTest. Something like select date_add('2024-01-01', empno) as empno from emp group by year(empno), where there will be a type error (int vs date) if the wrong column is referenced.
> 
> In short, more tests - that are broken before the fix, and pass after the fix, and are demonstrably the same behavior as a given DB - will give me the confidence to approve this PR.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
hannerwang: @suibianwanwank Thanks your remind, the switch check ensures the root is a basic call here.

hannerwang:
@mihaibudiu thanks, I have done more tests and add new comment. I'm available to push the pr merged this week.

mihaibudiu:
For now maybe you can make the CI pass

hannerwang:
> For now maybe you can make the CI pass

Hello @mihaibudiu , do you know how to rerun the CI/macos manually? the CI fails unexpectedly.

mihaibudiu:
This is how you can rerun failed jobs:
image
if you have enough privileges

hannerwang:
> This is how you can rerun failed jobs: image if you have enough privileges

Thanks, I think I have no the privilege to rerun.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -7378,6 +7368,37 @@ static class ExtendedExpander extends Expander {
...
+          for (SqlNode node : operandList) {
+            if (node.getKind() == SqlKind.ROW) {
+              List<SqlNode> rowOperandList = ((SqlCall) node).getOperandList();
+              aliasOrdinalExpandSet.addAll(rowOperandList);
```
mihaibudiu, line 7390: Can't ROWS contain other ROWS as fields?
In that case do you need a recursive call?

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -6289,6 +6296,35 @@ public boolean isBangEqualAllowed() {
...
+   * @see SqlConformance#isGroupByAlias()
+   */
+  @Test void testAliasInFunctionWithinGroupBy() {
+    final SqlConformanceEnum lenient = SqlConformanceEnum.LENIENT;
+    final SqlFunction date_add =
+        SqlBasicFunction.create(SqlKind.DATE_ADD, ReturnTypes.DATE,
+                OperandTypes.STRING_INTEGER)
+            .withFunctionType(SqlFunctionCategory.TIMEDATE);
+
+    sql("select date_add('2024-01-01', empno) as empno from emp group by ^year(empno)^")
```
mihaibudiu, line 6317: I don't really understand this test.

hannerwang: now date_add function receives empno:integer as a parameter, returns a DATE type.
if empno in year(empno) is substituted by 'date_add('2024-01-01', empno)', the year function will not raise type inconsistency error.
but we expect it throws here, because the empno in year(empno) should not be substituted. the identifier in a function should not be substituted.
I add this test as @julianhyde has mentioned this case. I paste his comment here.

> Also I would like to see a SqlValidatorTest. Something like select date_add('2024-01-01', empno) as empno from emp group by year(empno), where there will be a type error (int vs date) if the wrong column is referenced.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
hannerwang: Thanks @mihaibudiu , I test on bigquery and spark, they behave as you mentioned, we should consider the recursive rows, I added a method to check the nested rows case, you can review it when you are available.

mihaibudiu (approved):

mihaibudiu:
This PR is waiting for @julianhyde's approval, based on the discussion in JIRA.

PR #3931, opened 2024-08-26
[CALCITE-4806] Lossy CAST is incorrectly simplified

NobiGo:


NobiGo:
@mihaibudiu According to your comment in Jira, this issue has been fixed. My PR mainly involves adding test cases to verify it, so that we can close this ISSUE.

caicancai (approved):

core/src/test/resources/sql/misc.iq
```diff
@@ -3084,4 +3084,102 @@ from (values 0) as t(x);
 
 !ok
 
+# [CALCITE-4806] Lossy CAST is incorrectly simplified
```
mihaibudiu, line 3087: This test would be easier to read if it didn't check for =
But it works fine

mihaibudiu (approved):

PR #3934, opened 2024-08-27
[CALCITE-6549] Add LOG1P function (enabled in Spark library)

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6549

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -220,6 +220,7 @@ import static org.apache.calcite.sql.fun.SqlLibraryOperators.JSON_TYPE;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LEFT;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LEVENSHTEIN;
 import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG;
+import static org.apache.calcite.sql.fun.SqlLibraryOperators.LOG1P;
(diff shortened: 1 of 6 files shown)


caicancai:
@NobiGo PTAL, thanks

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4292,18 +4294,27 @@ private static List<Expression> args(RexCall call,
       operand0 = operands.left;
       operand1 = operands.right;
       boolean nonPositiveIsNull = library == SqlLibrary.MYSQL ? true : false;
```
NobiGo, line 4296: We should pass the library as the Parameter. Then the code logic will be clear. WDYT?

NobiGo: Maybe we should refactor the code in another issue. The current code here has poor extensibilit.

caicancai: Yes, I have to admit this. https://issues.apache.org/jira/browse/CALCITE-6550 is already working on this. I'll close it first.

caicancai: @NobiGo Could you please post your opinion at https://issues.apache.org/jira/browse/CALCITE-6550? I think it is a very complicated issue.

julianhyde: I don't think you should pass in a library at run time. The runtime should not be worrying about high-level concepts like llibrary.

PR #3935, opened 2024-08-28
[CALCITE-6548] Return type of COVAR_SAMP, COVAR_POP and VAR_SAMP shou…

NobiGo:
…ld be nullable

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
@@ -1402,7 +1402,8 @@ public abstract class ReturnTypes {
         typeFactory.getTypeSystem().deriveAvgAggType(typeFactory,
             opBinding.getOperandType(0));
     if (opBinding.getGroupCount() == 0 || opBinding.hasFilter()
-        || opBinding.getOperator().kind == SqlKind.STDDEV_SAMP) {
+        || opBinding.getOperator().getKind() == SqlKind.STDDEV_SAMP
+        || opBinding.getOperator().getKind() == SqlKind.VAR_SAMP) {
       return typeFactory.createTypeWithNullability(relDataType, true);
     } else {
       return relDataType;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -1414,7 +1415,9 @@ private static RelDataType multivalentStringWithSepSumPrecision(
     final RelDataType relDataType =
         typeFactory.getTypeSystem().deriveCovarType(typeFactory,
             opBinding.getOperandType(0), opBinding.getOperandType(1));
-    if (opBinding.getGroupCount() == 0 || opBinding.hasFilter()) {
+    if (opBinding.getGroupCount() == 0 || opBinding.hasFilter()
+        || opBinding.getOperator().kind == SqlKind.COVAR_POP
```
mihaibudiu, line 1419: are you sure about COVAR_POP?
Usually SAMP functions have the property that they return NULL for a non-empty set of size 1.

NobiGo: According to https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/COVAR_POP.html. This function return type is nullable.
```
The function returns a value of type NUMBER. If the function is applied to an empty set, then it returns null.
```

mihaibudiu: all aggregates return null when applied to an empty set.
but if getGroupCount() != 0 the set can never be empty (because then there would be no aggregate at all)

mihaibudiu: this is about aggregates that are null even when the set is NOT empty

mihaibudiu: empty set = set with all elements null

NobiGo: I will do some test for this.

PR #3805, opened 2024-05-30
[CALCITE-6392] Support all PostgreSQL 14 date/time patterns for to_da…

normanj-bitquill:
…te/to_timestamp

* First phase, mostly reorganizing classes
* Does not yet implement to_date or to_timestamp
* Fixed up fill mode handling
* Isolated the unit test class so that it can make Locale changes safely
* Introduced some new classes split out handling of some formats strings

PR #3936, opened 2024-08-28
[CALCITE-6551] Add DATE_FORMAT function (enabled in MySQL library)

suibianwanwank:
Document: https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html#function_date-format

### DIFF
--- a/core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
+++ b/core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
@@ -245,7 +245,7 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%02d", calendar.get(Calendar.WEEK_OF_YEAR)));
     }
   },
-  V("", "The number of the year (Monday as the first day of the week),"
+  V("", "The number of the week (Monday as the first day of the week),"
       + "If the first day of the week belongs to the previous year, "
       + "calculate the week of the previous year"
       + "as a decimal number (01-53)") {
@@ -260,7 +260,7 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%02d", calendar.get(Calendar.WEEK_OF_YEAR)));
     }
   },
-  v("", "The number of the year (Sunday as the first day of the week),"
+  v("", "The number of the week (Sunday as the first day of the week),"
       + "If the first day of the week belongs to the previous year, "
       + "calculate the week of the previous year"
       + "as a decimal number (01-53)") {
@@ -377,7 +377,7 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%03d", calendar.get(Calendar.MILLISECOND)));
     }
   },
-  MCS("", "The millisecond as a decimal number (000000-999999)") {
+  MCS("", "The microsecond as a decimal number (000000-999999)") {
     @Override public void format(StringBuilder sb, Date date) {
       final Calendar calendar = Work.get().calendar;
       calendar.setTime(date);
--- a/core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
+++ b/core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
@@ -24,8 +24,10 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.time.DayOfWeek;
 import java.time.LocalDate;
 import java.time.format.TextStyle;
+import java.time.temporal.ChronoField;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.Locale;
@@ -59,6 +61,13 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%d", calendar.get(Calendar.DAY_OF_WEEK)));
     }
   },
+  D0("", "The weekday (Monday as the first day of the week) as a decimal number (0-6)") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      sb.append(String.format(Locale.ROOT, "%d", calendar.get(Calendar.DAY_OF_WEEK) - 1));
+    }
+  },
   DAY("EEEE", "The full weekday name, in uppercase") {
     @Override public void format(StringBuilder sb, Date date) {
       final Work work = Work.get();
@@ -75,6 +84,21 @@ public enum FormatElementEnum implements FormatElement {
     @Override public void format(StringBuilder sb, Date date) {
       final Work work = Work.get();
       sb.append(work.getDayFromDate(date, TextStyle.FULL).toLowerCase(Locale.ROOT));
+    }
+  },
+  DS("", "The month as a decimal number with suffix. (1st-31th)") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      int d = calendar.get(Calendar.DAY_OF_MONTH);
+      sb.append(String.format(Locale.ROOT, "%d%s", d, getNumericSuffix(d)));
+    }
+  },
+  D1("dd", "The day of the month as a decimal number (1-31)") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      sb.append(String.format(Locale.ROOT, "%d", calendar.get(Calendar.DAY_OF_MONTH)));
     }
   },
   DD("dd", "The day of the month as a decimal number (01-31)") {
@@ -221,6 +245,33 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%02d", calendar.get(Calendar.WEEK_OF_YEAR)));
     }
   },
+  V("", "The number of the year (Monday as the first day of the week),"
+      + "If the first day of the week belongs to the previous year, "
+      + "calculate the week of the previous year"
+      + "as a decimal number (01-53)") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      calendar.setFirstDayOfWeek(Calendar.SUNDAY);
+      // Day of year of the first Sunday of the year
+      int minimalDaysInFirstWeek =
+          getFirstWeekdayOfYear(DayOfWeek.MONDAY, calendar.get(Calendar.YEAR));
+      calendar.setMinimalDaysInFirstWeek(minimalDaysInFirstWeek);
+      sb.append(String.format(Locale.ROOT, "%02d", calendar.get(Calendar.WEEK_OF_YEAR)));
+    }
+  },
+  v("", "The number of the year (Sunday as the first day of the week),"
+      + "If the first day of the week belongs to the previous year, "
+      + "calculate the week of the previous year"
+      + "as a decimal number (01-53)") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      calendar.setFirstDayOfWeek(Calendar.MONDAY);
+      calendar.setMinimalDaysInFirstWeek(4);
+      sb.append(String.format(Locale.ROOT, "%02d", calendar.get(Calendar.WEEK_OF_YEAR)));
+    }
+  },
   MI("m", "The minute as a decimal number (00-59)") {
     @Override public void format(StringBuilder sb, Date date) {
       final Calendar calendar = Work.get().calendar;
@@ -326,7 +377,16 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%03d", calendar.get(Calendar.MILLISECOND)));
     }
   },
-  SS("s", "The second as a decimal number (00-60)") {
+  MCS("", "The millisecond as a decimal number (000000-999999)") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      // It exceeds the precision of Calendar, we fill it with 0.
+      // So the actual range is (000000-999000)
+      sb.append(String.format(Locale.ROOT, "%06d", calendar.get(Calendar.MILLISECOND) * 1000));
+    }
+  },
+  SS("s", "The second as a decimal number (00-59)") {
     @Override public void format(StringBuilder sb, Date date) {
       final Calendar calendar = Work.get().calendar;
       calendar.setTime(date);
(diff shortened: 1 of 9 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5185,6 +5187,61 @@ void testBitGetFunc(SqlOperatorFixture f, String functionName) {
...
+      f.checkString("date_format(timestamp '1997-10-04 22:23:00', '%H %k %I %r %T %S %w')",
+          "22 22 10 10:23:00 PM 22:23:00 00 6",
+          "VARCHAR NOT NULL");
+      f.checkString("date_format(timestamp '1999-01-01', '%X %V')",
+          "1998 52",
+          "VARCHAR NOT NULL");
+      f.checkString("date_format(timestamp '1997-10-04 22:23:00.000000', '%c %D %e %f %h')",
+          "10 4th 4 000000 10",
+          "VARCHAR NOT NULL");
+      f.checkString("date_format(timestamp '1997-01-04 22:23:00', '%w %U %u %V %v %X %x')",
```
suibianwanwank, line 5219: Here are 4 cases testing mysql's weekForYear in two modes.
https://stackoverflow.com/questions/30364141/mysql-convert-yearweek-to-date

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5185,6 +5187,61 @@ void testBitGetFunc(SqlOperatorFixture f, String functionName) {
...
+          "39 40 1997 1997 %",
+          "VARCHAR NOT NULL");
+      f.checkString("date_format(timestamp '1997-10-04 22:23:00', '%l %p %U %u')",
+          "10 PM 39 40",
+          "VARCHAR NOT NULL");
+      f.checkQueryFails(StringAndPos.of("date_format(^timestamp '2006-06-00'^, '%d')"),
+          "Non-query expression encountered in illegal context");
+      f.checkNull("to_char(timestamp '2022-06-03 12:15:48.678', NULL)");
+//      f.checkString("date_format('1999-13-01', '%X %V')",
+//          null,
```
suibianwanwank, line 5238: MySQL supports this writing and will return null if the timestamp type is not declared and the parameter does not satisfy the timestamp format, which seems to be difficult to do in calcite

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4919,7 +4920,8 @@ void testBitGetFunc(SqlOperatorFixture f, String functionName) {
   }
 
   @Test void testToCharPg() {
-    final SqlOperatorFixture f = fixture().withLibrary(SqlLibrary.POSTGRESQL);
```
suibianwanwank, line 4922: I noticed that the default tester doesn't check the results, so I added this.

mihaibudiu:
If you want deterministic tests I think you should not use the local timezone.

asolimando:
> If you want deterministic tests I think you should not use the local timezone.

@mihaibudiu is right, tests should be timezone agnostic, see [Cassandra tests](https://github.com/apache/calcite/blob/main/cassandra/src/test/java/org/apache/calcite/test/CassandraAdapterDataTypesTest.java#L196) for an example.

suibianwanwank:
@mihaibudiu @asolimando I found this problem with all of this test classes, if the time zone is changed arbitrarily. For Calender's test, we need to set the time zone at this time like SqlOperatorTest. I'm not sure if this should be fixed in this PR .

mihaibudiu:
You can see a solution in e.g., https://github.com/apache/calcite/pull/3805, where the timezone is set before the tests and reset afterwards. It's sufficient to make your own tests pass.

suibianwanwank:
> You can see a solution in e.g., #3805, where the timezone is set before the tests and reset afterwards. It's sufficient to make your own tests pass.

I've fixed my test by keeping the data consistent with other tests. But I think in some time zones, some tests may not pass if we don't specify the time zone.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -326,7 +377,16 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%03d", calendar.get(Calendar.MILLISECOND)));
     }
   },
-  SS("s", "The second as a decimal number (00-60)") {
+  MCS("", "The millisecond as a decimal number (000000-999999)") {
+    @Override public void format(StringBuilder sb, Date date) {
+      final Calendar calendar = Work.get().calendar;
+      calendar.setTime(date);
+      // It exceeds the precision of Calendar, we fill it with 0.
```
mihaibudiu, line 384: Milliseconds are usually between 0 and 999.

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -221,6 +245,33 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(String.format(Locale.ROOT, "%02d", calendar.get(Calendar.WEEK_OF_YEAR)));
     }
   },
+  V("", "The number of the year (Monday as the first day of the week),"
```
mihaibudiu, line 248: the comment seems wrong, it is about the week number

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -399,6 +480,27 @@ public enum FormatElementEnum implements FormatElement {
       sb.append(work.yyyyFormat.format(date));
     }
   },
+  WFY("", "The year for the week where Sunday is the first day of the week") {
```
mihaibudiu, line 483: I think this should say "the week of the year"

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
```diff
@@ -399,6 +480,27 @@ public enum FormatElementEnum implements FormatElement {
...
+      calendar.setTime(date);
+      calendar.setFirstDayOfWeek(Calendar.SUNDAY);
+      // Day of year of the first Sunday of the year
+      int minimalDaysInFirstWeek =
+          getFirstWeekdayOfYear(DayOfWeek.MONDAY, calendar.get(Calendar.YEAR));
+      calendar.setMinimalDaysInFirstWeek(minimalDaysInFirstWeek);
+      sb.append(String.format(Locale.ROOT, "%04d", calendar.getWeekYear()));
+    }
+  },
+  WFY0("", "The year for the week where Monday is the first day of the week") {
```
mihaibudiu, line 495: same problem

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5185,6 +5187,61 @@ void testBitGetFunc(SqlOperatorFixture f, String functionName) {
...
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6551">[CALCITE-6551]
+   * Add DATE_FORMAT function (enabled in MySQL library)</a>. */
+  @Test void testDateFormat() {
+    final SqlOperatorFixture f = fixture()
+        .withTester(t -> TESTER).withLibrary(SqlLibrary.MYSQL);
+    f.setFor(SqlLibraryOperators.DATE_FORMAT);
+    final Locale originalLocale = Locale.getDefault();
+
+    try {
+      Locale.setDefault(Locale.US);
```
mihaibudiu, line 5200: You should not need to set the locale in the test if these functions are supposed to be deterministic.
Perhaps the function implementation should set the locale.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: what you need is to use an implicit cast to timestamp if the input is a CHAR data type.
you should be able to support that.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5185,6 +5187,61 @@ void testBitGetFunc(SqlOperatorFixture f, String functionName) {
...
+          "VARCHAR NOT NULL");
+      f.checkString("date_format(timestamp '1997-01-05 22:23:00', '%w %U %u %V %v %X %x')",
+          "0 01 01 01 01 1997 1997",
+          "VARCHAR NOT NULL");
+      f.checkString("date_format(timestamp '1997-01-06 22:23:00', '%w %U %u %V %v %X %x')",
+          "1 01 02 01 02 1997 1997",
+          "VARCHAR NOT NULL");
+      f.checkString("date_format(timestamp '1997-10-04 22:23:00', '%V %v %X %x %%')",
+          "39 40 1997 1997 %",
+          "VARCHAR NOT NULL");
```
mihaibudiu, line 5230: can you add some tests with dates before the Gregorian calendar?
I found that some of these fail.
Also, can you say how you validated the outputs?
Are these tested in MySQL?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5185,6 +5187,61 @@ void testBitGetFunc(SqlOperatorFixture f, String functionName) {
...
+   * Add DATE_FORMAT function (enabled in MySQL library)</a>. */
+  @Test void testDateFormat() {
+    final SqlOperatorFixture f = fixture()
+        .withTester(t -> TESTER).withLibrary(SqlLibrary.MYSQL);
+    f.setFor(SqlLibraryOperators.DATE_FORMAT);
+    final Locale originalLocale = Locale.getDefault();
+
+    try {
+      Locale.setDefault(Locale.US);
+      f.checkString("date_format(timestamp '2009-10-04 22:23:00', '%W %M %Y')",
```
caicancai, line 5201: Does 0000-01-01 or BC date support format?

suibianwanwank: I tested this example in mysql.
`SELECT DATE_FORMAT('0000-01-01', '%X %V');`
return 
0000-01
but
`SELECT DATE_FORMAT(timestamp '0000-01-01', '%X %V');`
return error: Incorrect DATETIME value: '0000-01-01'

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
suibianwanwank: For datetime functions, the implementation should be based on the user's time zone, just as other arbitrary databases support setting time zones. And other test do it.
And other tests have done the same

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
suibianwanwank: It throws an exception:Illegal TIMESTAMP literal, In the timestamp scenario, this is consistent with mysql's result.
Yes, I do these tests in mysql . In fact, I write these test cases in mysql and get the results, and then add them to test to validate

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
suibianwanwank: +1， i will fix it

core/src/main/java/org/apache/calcite/util/format/FormatElementEnum.java
suibianwanwank: This is an incorrect function description, it means microsecond here, I'll fix it

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
suibianwanwank: Yes, implicit cast to timestamp this can be done in calcite, but if this cast fails, calcite will throw an exception, but for mysql, it will return null. 
This is something that I don't think is easy to implement. unless we add a Config that adds a Config in the checksum phase and returns NULL if the implicit cast fails. WDYT

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
suibianwanwank: > It throws an exception:Illegal TIMESTAMP literal, In the timestamp scenario, this is consistent with mysql's result.

In the Calcite test, this throws an exception because it can't be converted to a timestamp

caicancai: Yes, Calcite has its own date standard.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: There are three distinct data types in Calcite:
TIMESTAMP
TIMESTAMP WITH LOCAL TIME ZONE
TIMESTAMP WITH TIME ZONE

You have to specify the type these functions operate on. I think you probably want functions for each data type. The ones for TIMESTAMP and TIMESTAMP WITH TIMEZONE should be deterministic. And the ones for LOCAL TIME ZONE should be implemented in terms of the other ones by setting the timezone first. But you should not rely on the Java system timezone for this.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: The Gregorian Calendar starts in 1582: https://en.wikipedia.org/wiki/Gregorian_calendar
I don't expect that mysql fails in this case.
Perhaps it fails for dates BC.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: I see, the date is incorrect.
In this case indeed you can probably not match exactly the behavior of mysql.
But I think you should still make it a negative test and leave the comment in.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
NobiGo: > SELECT DATE_FORMAT(timestamp '0000-01-01', '%X %V');

This failed because '0000-01-01' is not a timestamp. You can try:
```
SELECT DATE_FORMAT(data '0000-01-01', '%X %V');
```

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -5185,6 +5187,60 @@ void testBitGetFunc(SqlOperatorFixture f, String functionName) {
...
+      f.checkString("date_format(timestamp '2009-10-04 22:23:00', '%H:%i:%s')",
+          "22:23:00",
+          "VARCHAR NOT NULL");
+      f.checkString("date_format(timestamp '1900-10-04 22:23:00', '%D %y %a %d %m %b %j')",
+          "4th 00 Thu 04 10 Oct 277",
+          "VARCHAR NOT NULL");
+      f.checkString("date_format(timestamp '1997-10-04 22:23:00', '%H %k %I %r %T %S %w')",
+          "22 22 10 10:23:00 PM 22:23:00 00 6",
+          "VARCHAR NOT NULL");
+      f.checkString("date_format(timestamp '1999-01-01', '%X %V')",
```
NobiGo, line 5213: In MySQL,This will throw exception `ERROR 1525 (HY000) at line 1: Incorrect DATETIME value: '1994-01-01'`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: > > SELECT DATE_FORMAT(timestamp '0000-01-01', '%X %V');
> 
> This failed because '0000-01-01' is not a timestamp. You can try:
> 
> ```
> SELECT DATE_FORMAT(data '0000-01-01', '%X %V');
> ```

0000-01-01 is not even a valid date according to the SQL standard.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
suibianwanwank: Yes, I removed time to simplify testing Case, and in fact, Mysql doesn't support doing that. I'll add time to this test, but for this current test, I'm not sure we can be consistent with Mysql in Calcite

suibianwanwank: > We check the results by CalciteSqlOperatorTest. In SqlOperatorTest we check the return type.
It looks like CalciteSqlOperatorTest is better suited for this test, do I need to modify this test into CalciteSqlOperatorTest as this looks like it needs result checking too

site/_docs/reference.md
```diff
@@ -2944,6 +2944,7 @@ In the following:
 | b | TIME_SUB(time, interval)                       | Returns the TIME value that is *interval* before *time*
 | b | TIME_TRUNC(time, timeUnit)                     | Truncates *time* to the granularity of *timeUnit*, rounding to the beginning of the unit
 | m o p r | TO_CHAR(timestamp, format)               | Converts *timestamp* to a string using the format *format*
+| m | DATE_FORMAT(timestamp, format)                 | Converts *timestamp* to a string using the format *format*
```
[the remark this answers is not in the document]
suibianwanwank: Yeah, I think they're the same. In fact, Mysql does not have a to_char function.
> {@code TO_CHAR} is not supported in MySQL, but it is supported in * MariaDB, a variant of MySQL covered by {@link SqlLibrary#MYSQL}.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
suibianwanwank: Yes, this require more testing, including for BC dates.

PR #3938, opened 2024-08-28
[CALCITE-6554] Check for correlation variables in project during SqlToRelConverter.createAggImpl

ian-bertolacci:
Check for correlation variables in intermediate project when converting aggregation.

Its probably wise for us to check other places where this may be an issue.

I'm open to alternative solutions which have a wider effect, such as having the RelBuilder or even the constructor collect the correlation variables when the variable set is otherwise undefined.

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
@@ -3770,6 +3770,46 @@ class SqlToRelConverterTest extends SqlToRelTestBase {
     sql(sql).withExpand(false).withDecorrelate(false).ok();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6554">[CALCITE-6554]
+   * Nested correlated sub-query in aggregation does not have inner correlation variable bound
+   * to inner projection</a>. */
+  @Test void testCorrelationInProjectionWith1xNestedCorrelatedProjection() {
+    final String sql = "select e1.empno,\n"
+          + "  (select sum(e2.sal +\n"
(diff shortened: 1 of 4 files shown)


mihaibudiu (approved):
I am not familiar with the builder, so I cannot make a confident statement, but the tests seem fine.

core/src/test/resources/sql/sub-query.iq
```diff
@@ -668,6 +668,53 @@ FROM (SELECT 1 AS a) AS t;
 
 !ok
 
+# [CALCITE-6554] Nested correlated subqueries
```
NobiGo, line 671: It should be the same as the Jira Summary.

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -3770,6 +3770,34 @@ void checkCorrelatedMapSubQuery(boolean expand) {
     sql(sql).withExpand(false).withDecorrelate(false).ok();
   }
 
+  @Test void testCorrelationInProjectionWith1xNestedCorrelatedProjection() {
```
NobiGo, line 3773: Maybe the first test case can add the Jira link and description.

mihaibudiu:
@ian-bertolacci, can you please address the comments from @NobiGo?
Then we can probably merge this

caicancai:
@ian-bertolacci Please squash commit so that this PR can be merged

ian-bertolacci:
I've gone ahead and done it, but I'm pretty sure Github has squash-and-merge functionality, which would be nice if we could use.
I personally hate manually altering git history, and doing a squash on merge means you're not waiting on me (or anyone else in the future) to do it.

PR #3939, opened 2024-08-28
[CALCITE-6552] Enable CheckerFramework in 'server' module

julianhyde:
In [CALCITE-4199](https://issues.apache.org/jira/browse/CALCITE-4199) we enabled CheckerFramework (via annotations and CI jobs) in the 'core' and 'linq4j' modules; this change further enables CheckerFramework in the 'server' module, and fixes all violations.

There are also a large number of 'cosmetic' modifications to improve code quality without changing behavior, including:
 * Replace `this.x = x; assert x != null;` with `this.x = requireNonNull(x);` in constructors
 * Replace `assert` in other code locations where it implements an invariant. We don't use `requireNonNull` because it throws `NullPointerException`; we would prefer to throw `AssertionError` or `IllegalStateException`
 * Replace `x.equals("")` and `x.size() == 0` with `x.isEmpty()`
 * Make fields `final` where possible
 * Make private methods and inner classes `static` where possible
 * In class `Pair<K, V>` make the type variables `K` and `V` no longer nullable by default (you can make each of them nullable if you need)
 * Always import `Objects.requireNonNull`, `Integer.parseInt`, `Float.parseFloat`, `Double.parseDouble`, `Byte.parseByte`, `Short.parseShort`, `Boolean.parseBoolean`, `Long.parseLong` via static import

This PR supersedes #3937


mihaibudiu:
I imagine many of these conversions are automatically performed.
Would you like some of this code reviewed?
This may cause lots of conflicts if people have work under way.

NobiGo (approved):
LGTM.

julianhyde:
@mihaibudiu 

> Would you like some of this code reviewed?

I see @NobiGo has reviewed and approved. I think we're good.

The vast majority of changes don't require review - there was only one choice to make. In one or two places I had to make a decision; e.g. in MongoEnumerator I added the 3 `if` ... `throw` lines, because the alternative,  to change the return to `@Nullable Object`, was inferior:

```
  @Override public Object current() {
    if (current == null) {
      throw new IllegalStateException();
    }
    return current;
  }
```

> This may cause lots of conflicts if people have work under way.

Agreed. I will minimize conflicts by merging this quickly, and avoiding changes that require discretion. I will probably merge today.

@NobiGo, Thanks for your review.

PR #3940, opened 2024-08-29
[CALCITE-6516] Invalid cast inserted at AVG to SUM/COUNT transformation

LeonidChistov:


mihaibudiu (approved):
Similar code exists for all non-trivial aggregate functions

LeonidChistov:
> Similar code exists for all non-trivial aggregate functions

True, but I am little bit afraid to touch it since I am not well aware of `covar` , `stddev` and `regr` semantics.

julianhyde:
Sorry, I think you need to touch that other code. Otherwise it will be inconsistent with AVG and no one will know why.

I don't think your PR is in a fit state to commit. There are no comments, the query is a single line 120 characters long.

Also, you don't need to 'imagine' why `ensureType` was added. (Since you tagged me in the case, are you asking me to do that work?) Please do some research based on the code history, and add your findings to the jira case.

PR #3942, opened 2024-08-30
[CALCITE-3779] Implement BITAND, BITOR, BITXOR scalar functions

normanj-bitquill:
* All three functions accept either a pair of Integer or Binary values as arguments
* Binary arguments must be of the same length
* Returns NULL if any argument is NULL
* Returns a value of the same type as the first argument if both arguments are not null

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -2835,6 +2835,20 @@ public class SqlFunctions {
     return b0 & b1;
   }
 
+  /** Bitwise function <code>BITAND</code> applied to a Long and int value.
+   *  Needed for handling NULL for the first argument.
+   */
+  public static long bitAnd(Long b0, int b1) {
+    return b0 & b1;
+  }
+
+  /** Bitwise function <code>BITAND</code> applied to a Long and int value.
+   *  Needed for handling NULL for the second argument.
+   */
+  public static long bitAnd(int b0, Long b1) {
+    return b0 & b1;
+  }
+
   /** Bitwise function <code>BIT_AND</code> applied to binary values. */
   public static ByteString bitAnd(ByteString b0, ByteString b1) {
     return binaryOperator(b0, b1, (x, y) -> (byte) (x & y));
@@ -2882,6 +2896,20 @@ public class SqlFunctions {
     return b0 | b1;
   }
 
+  /** Bitwise function <code>BITOR</code> applied to a Long and int value.
+   *  Needed for handling NULL for the first argument.
+   */
+  public static long bitOr(Long b0, int b1) {
+    return b0 | b1;
+  }
+
+  /** Bitwise function <code>BITOR</code> applied to a Long and int value.
+   *  Needed for handling NULL for the second argument.
+   */
+  public static long bitOr(int b0, Long b1) {
+    return b0 | b1;
+  }
+
   /** Bitwise function <code>BIT_OR</code> applied to binary values. */
   public static ByteString bitOr(ByteString b0, ByteString b1) {
     return binaryOperator(b0, b1, (x, y) -> (byte) (x | y));
@@ -2889,6 +2917,20 @@ public class SqlFunctions {
 
   /** Bitwise function <code>BIT_XOR</code> applied to integer values. */
   public static long bitXor(long b0, long b1) {
+    return b0 ^ b1;
+  }
+
+  /** Bitwise function <code>BITXOR</code> applied to a Long and int value.
+   *  Needed for handling NULL for the first argument.
+   */
+  public static long bitXor(Long b0, int b1) {
+    return b0 ^ b1;
+  }
+
+  /** Bitwise function <code>BITXOR</code> applied to a Long and int value.
+   *  Needed for handling NULL for the second argument.
+   */
+  public static long bitXor(int b0, Long b1) {
     return b0 ^ b1;
   }
 
--- a/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
@@ -614,6 +614,44 @@ public abstract class ReturnTypes {
       ARG0_EXCEPT_INTEGER.andThen(SqlTypeTransforms.TO_NULLABLE);
 
   /**
+   * Chooses a type to return.
+   * If all arguments are null, return nullable integer type.
+   * If all arguments are integer types, choose the largest integer type. Nullable
+   * if any argument is nullable.
+   * As a fallback, choose the type of the first argument that is not of the NULL type.
+   * Nullable if at least one argument is nullable.
+   */
+  public static final SqlReturnTypeInference LARGEST_INT_OR_FIRST_NON_NULL =
+      opBinding -> {
+        final RelDataTypeFactory typeFactory = opBinding.getTypeFactory();
+        RelDataType largestIntegerType = null;
+        RelDataType firstNonNullType = null;
+        boolean allArgsInteger = true;
+        boolean nullable = false;
+        for (RelDataType opType : opBinding.collectOperandTypes()) {
+          if (firstNonNullType == null && SqlTypeName.NULL != opType.getSqlTypeName()) {
+            firstNonNullType = opType;
+          }
+          if (SqlTypeName.INT_TYPES.contains(opType.getSqlTypeName())
+              && (largestIntegerType == null
+              || largestIntegerType.getPrecision() < opType.getPrecision())) {
+            largestIntegerType = opType;
+          } else {
+            allArgsInteger = false;
+          }
+          nullable |= opType.isNullable();
+        }
+        if (allArgsInteger && largestIntegerType != null) {
+          return typeFactory.createTypeWithNullability(largestIntegerType, nullable);
+        } else if (firstNonNullType != null) {
+          return typeFactory.createTypeWithNullability(firstNonNullType, nullable);
+        }
+        throw opBinding.newError(
+            RESOURCE.atLeastOneArgumentMustNotBeNull(
+                opBinding.getOperator().getName()));
+      };
+
+  /**
    * Returns the same type as the multiset carries. The multiset type returned
    * is the least restrictive of the call's multiset operands
    */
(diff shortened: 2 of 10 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -15470,6 +15470,111 @@ private static void checkLogicalOrFunc(SqlOperatorFixture f) {
...
+        "Different length for bitwise operands: the first: 2, the second: 1",
+        true);
+  }
+
+  @Test void testBitOrScalarFunc() {
+    final SqlOperatorFixture f = fixture();
+    f.setFor(SqlStdOperatorTable.BITOR, VmName.EXPAND);
+    f.checkFails("bitor(^*^)", "Unknown identifier '\\*'", false);
+    f.checkScalar("bitor(2, 4)", "6", "INTEGER");
+    f.checkType("bitor(CAST(2 AS TINYINT), CAST(6 AS TINYINT))", "TINYINT");
```
mihaibudiu, line 15513: Can we have a few tests with negative numbers, and a few with negative shorter integers?

mihaibudiu (approved):
I don't really understand why the result for TINYINT will be correct, since today Calcite incorrectly does not truncate the results for short arithmetic.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2829,6 +2829,20 @@ public static long bitAnd(long b0, long b1) {
...
   }
 
+  /** Bitwise function <code>BITAND</code> applied to a Long and int value.
+   *  Needed for handling NULL for the first argument.
+   */
+  public static long bitAnd(Long b0, int b1) {
+    return b0 & b1;
+  }
+
+  /** Bitwise function <code>BITAND</code> applied to a Long and int value.
```
NobiGo, line 2839: Bitwise function <code>BITAND</code> applied to a int and Long value.

NobiGo (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
normanj-bitquill: Added a few more tests with negative numbers.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
normanj-bitquill: Added some more tests with different integer types for the arguments (integer and bigint).

Needed to rework the return type detection. It will now use the largest integer type if both arguments are integer types. There may be ways of cleaning the code up further for the return type inference.

normanj-bitquill:
> I don't really understand why the result for TINYINT will be correct, since today Calcite incorrectly does not truncate the results for short arithmetic.

@mihaibudiu I have created a custom `SqlReturnTypeInference` class. It will look at the arguments and choose the largest integer type (if both are integer types), otherwise use the type of the first argument. If both arguments are `NULL`, then just assume integer nullable return type.

Perhaps you mean when the result of the BITAND is converted from a long back to a TINYINT. If this is the case, do you have any examples where it fails to truncate the value correctly? No truncation should be needed, since the result cannot exceed the largest argument type.

mihaibudiu:
I think this pattern is common in all arithmetic operations for integer types. Hopefully you could have reused some code.
Your runtime only supported int and long, so I suspected the results would be wrong for short.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -613,6 +613,59 @@ public static SqlCall stripSeparator(SqlCall call) {
...
+            } else if (isIntegerTypeLarger(integerReturnType, opType)) {
+              integerReturnType = opType;
+            }
+          } else {
+            allArgsInteger = false;
+            break;
+          }
+        }
+        if (integerReturnType != null && allArgsInteger) {
+          return typeFactory.createTypeWithNullability(integerReturnType, true);
```
mihaibudiu, line 638: why are all these types nullable?

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -613,6 +613,59 @@ public static SqlCall stripSeparator(SqlCall call) {
...
+        }
+        for (RelDataType opType : opBinding.collectOperandTypes()) {
+          if (!opType.isNullable()) {
+            return typeFactory.createTypeWithNullability(opType, true);
+          }
+        }
+        return typeFactory.createTypeWithNullability(
```
mihaibudiu, line 645: When can this happen?

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -613,6 +613,59 @@ public static SqlCall stripSeparator(SqlCall call) {
...
+
+  /**
+   * Returns true if type2 is of a larger integer type than type1.
+   *
+   * @param type1 first type
+   * @param type2 second type
+   * @return true if type2 is of a larger integer type than type1
+   */
+  private static boolean isIntegerTypeLarger(RelDataType type1, RelDataType type2) {
+    if (SqlTypeName.TINYINT == type1.getSqlTypeName()) {
```
mihaibudiu, line 657: you can use the precision to make this simpler, and perhaps safer too.
In some type systems maybe some of these types could be the same?

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
normanj-bitquill: Since `NULL` is a possible result.
```
BITAND(1, NULL) = NULL
BITAND(CAST(x'01' AS BINARY(1)), NULL) = NULL
```

If nullability is set to `false` here, then `NULL` results are coerced back to integer types with the value `0`.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
normanj-bitquill: For line 642 above, it would get there when at least one argument is a non integer type and not null. This is probably not quite correct. I only want to exclude the `NULL` type. A nullable value is fine as long as it is an integer or binary type.

I'll see about correcting this.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
normanj-bitquill: Thanks, that helps a lot. I have cleaned up the code to make use of precision.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
normanj-bitquill: I see what you mean. I have reworked the return type detection. It is now only nullable if at least one argument type is nullable.

core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
normanj-bitquill: I have reworked the logic here.

If all arguments are of the NULL type, then the return type is INTEGER nullable.

If all arguments are integer types, then the return type is the largest type found. It is nullable if at least one argument is nullable.

As a fallback, the first argument that is not of the NULL type is the return type. It is nullable if at least one argument is nullable.

mihaibudiu: The fallback could be coded more explicitly as a check for the NULL type and then an assertion failure.
I like to be as explicit as possible, it's more future-proof.

mihaibudiu: In fact, if any argument has the NULL type the result type can also have the NULL type, since there is only one value with this type, NULL.

normanj-bitquill: I agree with you in theory, but I don't know how to do this in practice. When I make the change so that the NULL type is returned if any argument is of the NULL type, then the tests fail. They fail with this error occurring when trying to convert the result.
```
java.lang.NoSuchMethodException: java.lang.Void.valueOf(void)
```

It looks like it is trying to the result to a Void type in Java. It never went to an implementation of the `BITAND()` function. I also tried use a return type of `java.lang.Long` for the relevant implementation of the `BITAND()` function. This gives the a similar error but for the Long type error:
```
java.lang.NoSuchMethodException: java.lang.Long.voidValue()
```

Another way to look at this is that `BITAND(1, NULL)` should return a `NULL` value of the type of the first argument. I am treating `BITAND(1, NULL)` as the same as `BITAND(NULL, 1)`.

mihaibudiu: I was hoping that the nullpolicy would take care of that, but I guess I am wrong (the STRONG policy ensures that a function with null arguments is never called). But I guess you still need a return type, it can't be NULL. So probably the right solution is to return a compile-time error when all arguments have NULL type, and otherwise the type of the first non-NULL typed argument will do fine if any argument has type NULL.

normanj-bitquill: Updated. Now it gives an error when all arguments are NULL.

core/src/main/resources/org/apache/calcite/runtime/CalciteResource.properties
```diff
@@ -207,6 +207,7 @@ ArgumentMustBeNumericLiteralInRange=Argument to function ''{0}'' must be a numer
 ValidationError=Validation Error: {0}
 IllegalLocaleFormat=Locale ''{0}'' in an illegal format
 ArgumentMustNotBeNull=Argument to function ''{0}'' must not be NULL
+AtLeastOneArgumentMustNotBeNull=At least one argument to function ''{0}'' must not be NULL
```
mihaibudiu, line 210: there are other functions which give an error in this case, but this error message is frankly better. maybe we should update the other ones.

mihaibudiu (approved):

mihaibudiu:
Sorry, but it looks like other merges have introduced conflicts, please fix them so we can merge this one too

normanj-bitquill:
Commits have been squashed.

core/src/main/resources/org/apache/calcite/runtime/CalciteResource.properties
normanj-bitquill: Do you know what I can look for? I could update them.

mihaibudiu: I have seen this: "Illegal use of ''NULL''"

mihaibudiu: let's leave this for another PR, this has been strenuous enough

PR #3943, opened 2024-08-30
[CALCITE-4871] CAST a literal to DECIMAL type return wrong result

NobiGo:


core/src/main/java/org/apache/calcite/sql/type/SqlTypeFactoryImpl.java
```diff
@@ -88,6 +90,14 @@ public SqlTypeFactoryImpl(RelDataTypeSystem typeSystem) {
     if (maxPrecision >= 0 && precision > maxPrecision) {
       precision = maxPrecision;
     }
+    if (typeName == SqlTypeName.DECIMAL) {
+      if (precision != RelDataType.PRECISION_NOT_SPECIFIED && precision <= 0) {
+        throw RESOURCE.invalidPrecisionForDecimalType(precision, maxPrecision).ex();
+      }
+      if (scale != RelDataType.SCALE_NOT_SPECIFIED && (scale < 0 || scale > precision)) {
```
mihaibudiu, line 97: it turns our that there are SQL variants which allow both negative scales and scale > precision. There is no reason for Calcite not to support these.
See https://issues.apache.org/jira/browse/CALCITE-6407
https://issues.apache.org/jira/browse/CALCITE-6406

mihaibudiu: Java supports negative scales. It essentially means that the number has lots of zeros not represented. E.g., 1000 is 1 with scale -3.

NobiGo: I appreciate your reminder. How about we add additional configurations to control this behavior about negative scales and scale>precision？ We may need two different configurations.

mihaibudiu: yes, these should be in the type system probably, same as rounding
I personally haven't thought too much about the rules or arithmetic on negative scales
several algorithms used by calcite (e.g., infer types for arithmetic results) may need to be revisited if this is possible. I think that the existing ones were designed assuming that scale cannot be negative.

NobiGo: Yes. At first, I tested it in different databases, but it still missed some features that may have been in newer versions. I am preparing to revert the unreasonable test cases in this PR(if so, this PR only left some unit tests about CAST a literal to SECAIMAL type). Do you think this PR is still necessary? If not, I am prepared to close this PR and Jira issue.

mihaibudiu: This PR is still useful for a type system that does not support negative scales.

core/src/main/java/org/apache/calcite/rel/type/DelegatingTypeSystem.java
```diff
@@ -52,6 +52,14 @@ protected DelegatingTypeSystem(RelDataTypeSystem typeSystem) {
     return typeSystem.getMaxNumericPrecision();
   }
 
+  @Override public boolean isSupportNegativeScale() {
```
mihaibudiu, line 55: I think you should open a new issue for this change.

mihaibudiu: And probably you don't need two verbs, you can call it supportsNegativeScale.

NobiGo: This PR only includes some tests so we can close the Jira issue.

mihaibudiu (approved):

PR #3944, opened 2024-09-02
[CALCITE-6423] Invalid unparse for CHAR without precision in MssqlSqlDialect

ChengJie1053:
…Dialect


core/src/main/java/org/apache/calcite/sql/dialect/MssqlSqlDialect.java
```diff
@@ -47,9 +50,22 @@
...
  */
 public class MssqlSqlDialect extends SqlDialect {
+    /**
+     * Mssql type system.
+     */
+  public static final RelDataTypeSystem MSSQL_TYPE_SYSTEM =
+            new RelDataTypeSystemImpl() {
+      @Override public int getDefaultPrecision(SqlTypeName typeName) {
+          if (typeName == SqlTypeName.CHAR) {
+          return RelDataType.PRECISION_NOT_SPECIFIED;
```
mihaibudiu, line 60: indentation is off

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystemImpl.java
```diff
@@ -57,7 +57,7 @@ public abstract class RelDataTypeSystemImpl implements RelDataTypeSystem {
     case INTERVAL_SECOND:
       return SqlTypeName.MAX_INTERVAL_FRACTIONAL_SECOND_PRECISION;
     default:
-      return -1;
+      return RelDataType.PRECISION_NOT_SPECIFIED;
```
NobiGo, line 60: SCALE_NOT_SPECIFIED ?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3684,12 +3684,15 @@ private SqlDialect nonOrdinalDialect() {
     String query = "select cast(\"product_id\" as char) from \"product\"";
     final String expectedMysql = "SELECT CAST(`product_id` AS CHAR)\n"
         + "FROM `foodmart`.`product`";
+    final String expectedMssql = "SELECT CAST([product_id] AS CHAR)\n"
```
NobiGo, line 3687: Add test about if Char with precision.

NobiGo (changes requested):

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
ChengJie1053: Ok, thank you for helping me review the code, I will modify it

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystemImpl.java
ChengJie1053: Ok, thank you for helping me review the code, I will modify it

core/src/main/java/org/apache/calcite/sql/dialect/MssqlSqlDialect.java
ChengJie1053: Ok, thank you for helping me review the code, I will modify it

NobiGo (approved):

Akanksha-kedia:
@ChengJie1053  @mihaibudiu @NobiGo @dynamix @mikehinchey I tried reaching the community but could not so really sorry for commenting in pr.

 Description:

When translating Hive view definitions to a format that Trino can interpret and process, the Coral library is converting DECIMAL data types to DECIMAL(12,0) in its convertQuery() method. This is causing a type mismatch when the actual data in the view is of type DECIMAL(19,0), leading to a VIEW_IS_STALE exception in Trino.

Scenario:

A Hive view is created with a column of type DECIMAL(19,0) that holds values such as 12345623423423.
This view is read by Trino, which uses the Coral library for translation.
Coral incorrectly converts the DECIMAL(19,0) type to DECIMAL(12,0).
When the view is queried, Trino checks if the actual data type (DECIMAL(19,0)) can be coerced to the defined data type (DECIMAL(12,0)).
This check fails, leading to a VIEW_IS_STALE exception.

Suspected Fix:

The suspected fix involves updating the convertQuery() method in Coral to correctly translate DECIMAL data types without reducing precision. The method should be able to handle and preserve larger DECIMAL types such as DECIMAL(19,0). This change will ensure that the translated data type in Trino matches the actual data type in Hive, avoiding the VIEW_IS_STALE exception.

Proposed Solution:

Update the convertQuery() method in the Coral library to correctly handle and preserve the original precision of DECIMAL data types during translation. This would involve investigating how the current conversion to DECIMAL(12,0) is happening and updating the logic to maintain the original precision of the data type. We also need to ensure that this change does not adversely affect other parts of the Coral library or its functionality.

mihaibudiu:
This should be an issue in JIRA. Do you have a JIRA account? You can ask for one, it's free.
https://issues.apache.org

I was thinking to file your text as an issue, but I couldn't find any Coral library in the Calcite source tree.
Is this an issue with Calcite?

caicancai:
> This should be an issue in JIRA. Do you have a JIRA account? You can ask for one, it's free. https://issues.apache.org
> 
> I was thinking to file your text as an issue, but I couldn't find any Coral library in the Calcite source tree. Is this an issue with Calcite?

@Akanksha-kedia @mihaibudiu Is this the cCoral library you are referring to https://github.com/linkedin/coral

mihaibudiu:
I don't know what the Coral library is.

PR #3945, opened 2024-09-02
[CALCITE-6560] Add supportsNegativeScale in RelDataTypeSystem

NobiGo:


core/src/main/java/org/apache/calcite/sql/type/SqlTypeFactoryImpl.java
```diff
@@ -88,6 +90,15 @@ public SqlTypeFactoryImpl(RelDataTypeSystem typeSystem) {
     if (maxPrecision >= 0 && precision > maxPrecision) {
       precision = maxPrecision;
     }
+    if (typeName == SqlTypeName.DECIMAL) {
+      if (precision != RelDataType.PRECISION_NOT_SPECIFIED && precision <= 0) {
+        throw RESOURCE.invalidPrecisionForDecimalType(precision, maxPrecision).ex();
+      }
+      if (scale != RelDataType.SCALE_NOT_SPECIFIED && scale < 0
```
mihaibudiu, line 97: This is now a problem, since SCALE_NOT_SPECIFIED is -1, which may be a legal scale value. Maybe this constant should be changed to Integer.MIN_VALUE.

core/src/test/resources/sql/misc.iq
```diff
@@ -3182,4 +3182,41 @@ from (values 0) as t(x);
...
+!error
+
+# precision is 0
+values cast(15.3 as decimal(0,4));
+DECIMAL precision 0 must be between 1 and max precision 19
+!error
+
+# scale is negative number
+values cast(15.3 as decimal(1,-1));
+Error while executing SQL "values cast(15.3 as decimal(1,-1))": parse failed: Encountered "-" at line 1, column 31.
```
mihaibudiu, line 3203: this suggests that the parser has to be changed to allow negative values for scale

core/src/main/java/org/apache/calcite/sql/type/SqlTypeFactoryImpl.java
NobiGo: SCALE_NOT_SPECIFIED is Integer.MIN_VALUE.  PRECISION_NOT_SPECIFIED is -1.

core/src/test/resources/sql/misc.iq
NobiGo: Yes. I have tried to do that. But this PR will become complicated. In Calcite， it has the same specific processing about Scala. So I left it to another PR when set supportsNegativeScale is true.

mihaibudiu: I don't think this is a big change. It's probably enough to change this line in the Parser.jj file:
```
scale = UnsignedIntLiteral()
```
to
```
scale = IntLiteral()
```

NobiGo: Yes, But when we execute SQL :
```
select cast(13.4 as decima(3,-1)) 
```
will be converted to
```
select cast(13.4 as decima(3,0)) 
```
So, we can't throw exceptions about the don't support negative scale.

mihaibudiu: it means that there is at least one more bug somewhere - this shouldn't happen

NobiGo: If we support a negative scale, then we can start by Parser.jj.

NobiGo: I will do it in another new issue.

mihaibudiu: I hope it's not a lot of code, if you add it to this issue you can exercise these tests properly

mihaibudiu: it's a one line change to Parser.jj, and you have to find out where the scale is lost. that may be everything you need to change.

NobiGo: Ok. I will check it.

mihaibudiu (approved):

core/src/test/resources/sql/misc.iq
```diff
@@ -3182,4 +3182,41 @@ from (values 0) as t(x);
 EnumerableValues(tuples=[[{ true, false, true, false, true, false, true, false }]])
 !plan
 
+# [CALCITE-6560] Add supportsNegativeScale in RelDataTypeSystem
```
julianhyde, line 3185: I don't think this should be in 'misc'

julianhyde (commented):
Can you add to SqlParserTest?

Also add to either SqlValidatorTest or SqlOperatorTest. The datatype occurs in a call to CAST. Make sure that a CAST is valid if and only if the type system allows negative types.

Do you expect to be able to evaluate queries with negative scales? Cast to a negative scale will result in a shifted decimal, e.g. 12,300 represented as the int 123 but printed as 12300.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
```diff
@@ -62,6 +62,9 @@ public interface RelDataTypeSystem {
   /** Returns the maximum precision of a NUMERIC or DECIMAL type. */
   int getMaxNumericPrecision();
 
+  /** Returns whether the scale of a NUMERIC or DECIMAL type can be negative. */
```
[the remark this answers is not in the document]
NobiGo: The default value is 'false'. 
Because before this PR, when we execute like 
```
values cast(153 as decimal(3, -1));
```
will throw a parse exception. After we support the parse, the default behavior  will convert it to 
```
values cast(153 as decimal(3, 0));
```
So I set 'false' as the default value to keep the existing semantics unchanged.

NobiGo:
> Can you add to SqlParserTest?
> 
> Also add to either SqlValidatorTest or SqlOperatorTest. The datatype occurs in a call to CAST. Make sure that a CAST is valid if and only if the type system allows negative types.
> 
> Do you expect to be able to evaluate queries with negative scales? Cast to a negative scale will result in a shifted decimal, e.g. 12,300 represented as the int 123 but printed as 12300.

We have created an ISSUE to support this  https://issues.apache.org/jira/browse/CALCITE-6406. I plan to do this in CALCITE-6406.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -5944,7 +5944,7 @@ SqlTypeNameSpec SqlTypeName3(Span s) :
         precision = UnsignedIntLiteral()
         [
             <COMMA>
-            scale = UnsignedIntLiteral()
+            scale = IntLiteral()
```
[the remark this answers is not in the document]
NobiGo: Done!

core/src/main/java/org/apache/calcite/rel/type/DelegatingTypeSystem.java
```diff
@@ -52,6 +52,14 @@ protected DelegatingTypeSystem(RelDataTypeSystem typeSystem) {
     return typeSystem.getMaxNumericPrecision();
   }
 
+  @Override public boolean supportsNegativeScale() {
+    boolean isSupportNegativeScale =  typeSystem.supportsNegativeScale();
+    if (isSupportNegativeScale) {
```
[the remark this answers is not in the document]
NobiGo: Because this PR only solves the issue when supportsNegativeScale is false. Will support negative scale  in CALCITE-6406.

core/src/test/java/org/apache/calcite/sql/type/RelDataTypeSystemTest.java
```diff
@@ -143,6 +145,31 @@ static class Fixture extends SqlTypeFixture {
...
+            return true;
+          }
+        });
+    RelDataType dataType = customTypeFactory.createSqlType(SqlTypeName.DECIMAL, 10, -5);
+    assertEquals(SqlTypeName.DECIMAL, dataType.getSqlTypeName());
+    assertEquals(10, dataType.getPrecision());
+    assertEquals(-5, dataType.getScale());
+  }
+
+  @Test void testNotSupportNegativeScale() {
```
mihaibudiu, line 161: if you support a negative scale, is there a lower limit?
For example, in Oracle the scale has maximum and minimum values: https://docs.oracle.com/cd/B19306_01/server.102/b14200.pdf, page 44 in the pdf

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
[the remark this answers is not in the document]
mihaibudiu: Based on my question below and other systems, maybe the right way to do this is to have a minimumNumericScale(), and this function to return 'true' if the minimum is less than 0.
You still have to have a rule to reject some types when the negative scale is too large.

NobiGo:
@mihaibudiu If we have a minimum scale value, do we still need supportsNegativeScale? If the minimum scale value <0 means support the negative scale.

mihaibudiu:
I think it's still useful as a method, it's nicer than comparing the minimum scale with 0.

julianhyde:
Maybe one method can delegate to the other. To define a type system that allows negative scale, people should only override the `minimumScale` method. `supportsNegativeScale` would delegate to it:

```
default boolean supportsNegativeScale() {
  return minimumScale() < 0;
}
```

core/src/main/java/org/apache/calcite/sql/type/SqlTypeFactoryImpl.java
```diff
@@ -88,6 +90,17 @@ public SqlTypeFactoryImpl(RelDataTypeSystem typeSystem) {
     if (maxPrecision >= 0 && precision > maxPrecision) {
       precision = maxPrecision;
     }
+    if (typeName == SqlTypeName.DECIMAL) {
+      if (precision != RelDataType.PRECISION_NOT_SPECIFIED && precision <= 0) {
+        throw RESOURCE.invalidPrecisionForDecimalType(precision, maxPrecision).ex();
+      }
+      if (scale != RelDataType.SCALE_NOT_SPECIFIED && scale < 0) {
+        if (!typeSystem.supportsNegativeScale() || scale < typeSystem.getMinNumericScale()) {
```
mihaibudiu, line 98: is the first test necessary?
comparing the scale with the minimum one may be enough?

core/src/test/java/org/apache/calcite/test/CustomRelDataTypeSystem.java
```diff
@@ -23,18 +23,42 @@
...
+    @Override public RoundingMode roundingMode() {
+      return RoundingMode.HALF_UP;
+    }
+  };
+
+  /**
+   *  Supports negative scale and rounding mode is {@link RoundingMode#DOWN}.
+   */
+  public static final RelDataTypeSystem NEGATIVE_SCALE = new RelDataTypeSystemImpl() {
+    @Override public int getMinNumericScale() {
```
mihaibudiu, line 47: this is a very large number... but I guess it works for testing

core/src/test/resources/sql/cast.iq
```diff
@@ -1234,4 +1234,385 @@ EnumerableCalc(expr#0..7=[{inputs}], expr#8=[CAST($t6):INTEGER], expr#9=[IS NOT
...
+select cast(12340 as decimal(3, -2));
++--------+
+| EXPR$0 |
++--------+
+|  12300 |
++--------+
+(1 row)
+
+!ok
+
```
mihaibudiu, line 1353: frankly, I am surprised these work so well.
There are certainly some operations that won't work with negative scales (some in code I wrote), but these are not in the scope of this PR.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1702,6 +1703,51 @@ void testCastToBoolean(CastType castType, SqlOperatorFixture f) {
...
+    negativeScaleFixture.checkScalar("cast(12.3 as decimal(3, -1))",
+        "10", "DECIMAL(3, -1) NOT NULL");
+    // cast decimal to decimal
+    negativeScaleFixture.checkScalar("cast(cast(12.3 as decimal(3, 1)) as decimal(3, -1))",
+        "10", "DECIMAL(3, -1) NOT NULL");
+    // cast string to decimal
+    negativeScaleFixture.checkScalar("cast('12.3' as decimal(3, -1))",
+        "10", "DECIMAL(3, -1) NOT NULL");
+    // cast interval to decimal
+    negativeScaleFixture.checkScalar("cast(INTERVAL '5' hour as decimal(3, -1))",
```
mihaibudiu, line 1747: is 0 the right result?

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
```diff
@@ -62,6 +64,14 @@ public interface RelDataTypeSystem {
   /** Returns the maximum precision of a NUMERIC or DECIMAL type. */
   int getMaxNumericPrecision();
 
+  /** Returns the minimum scale of a NUMERIC or DECIMAL type. */
+  int getMinNumericScale();
```
julianhyde, line 68: You should say that the default type system returns 0,
and that there are no 'special values' (like SCALE_NOT_SPECIFIED).

julianhyde, line 68: We currently have both `getMaxScale(SqlTypeName)` and `getMaxNumericScale()`. (The latter delegates to the former.)

So should we instead be creating `getMinScale(SqlTypeName)`? It would allow a type system to have different min scales for (say) a decimal and a binary floating point.

core/src/test/java/org/apache/calcite/test/CustomRelDataTypeSystem.java
NobiGo: Yes. Because In PostgreSQL this value is -1000. So I set -1000.

[1] https://www.postgresql.org/docs/current/datatype-numeric.html

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
NobiGo: @mihaibudiu According to test in testCastIntervalToNumeric
```
    f.checkScalarExact("cast(INTERVAL '5' minute as decimal(2,1))",
        "DECIMAL(2, 1) NOT NULL",
        isDecimal("5.0"));
    f.checkScalarExact("cast(INTERVAL '5' hour as decimal(2,1))",
        "DECIMAL(2, 1) NOT NULL",
        isDecimal("5.0"));
```

So I think it is right. WDYT?

mihaibudiu: frankly, I don't think that this particular test is that useful.
it depends on the semantics of converting intervals to decimals, which I am not sure is defined in the same way in various SQL dialects.
so you are essentially testing casting 5.0 to DECIMAL(3,-1), right?

NobiGo: Yes. I want to cover the INTERVAL type. But I can't find a database to reference.

core/src/main/java/org/apache/calcite/sql/type/SqlTypeFactoryImpl.java
NobiGo: I will remove it.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
NobiGo: RelDataTypeSystem is an interface It's strange to add a Java document here. So I added it in RelDataTypeSystemImpl. We have a SCALE_NOT_SPECIFIED in RelDataType for the types where the scale is not allowed.

core/src/test/resources/sql/cast.iq
NobiGo: Because this is quite complex, I made code changes based on the test cases, Can you give me a unit test?

mihaibudiu: I can't give you a unit test, but at least in Primitive.checkOverflow(BigDecimal, ...) there is an implicit assumption that negative scales are not legal. I think we can leave this for another PR, here you introduce support, but not all operations that need negative scales need to work yet.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
NobiGo: I checked the SQL TypeName#getMinScale and it processed the Interval type, which returns -1 for all other types. Do we need to add this method?

NobiGo: @julianhyde

core/src/test/resources/sql/cast.iq
NobiGo: Ok. Thanks @mihaibudiu. Your suggestion is very helpful.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystem.java
julianhyde: We should be explicit in the javadoc.

`getMaxScale(SqlTypeName)` seems to return `RelDataType.SCALE_NOT_SPECIFIED` (Integer.MIN_VALUE) for any data type that does not support scale. Can you amend its javadoc to state this?

The javadoc for `getMinScale(SqlTypeName)` should state its behavior for types that do not support scale. I propose `Integer.MAX_VALUE`.

The javadoc for `getMinScale(SqlTypeName)` should also say that the default type system returns 0 for `DECIMAL` and interval types. I know it's an interface, but interfaces can and should say what the default or typical behavior is.

I guess you should change `supportsNegativeScale()` to `supportsNegativeScale(SqlTypeName)`, and remove `getMinNumericScale()`.

NobiGo:
@julianhyde I have made some improvements based on your suggestions.
1. Refactor SqlTypeName#getDefaultScale to RelDataTypeSystem#getDefaultScale(SqlTypeName) and mark SqlTypeName#getDefaultScale as deprecated.
2. Refactor SqlTypeName#getMinPrecision to RelDataTypeSystem#getMinPrecision(SqlTypeName) and mark SqlTypeName#getMinPrecision as deprecated.
3. Refactor SqlTypeName#getMinScale to RelDataTypeSystem#getMinScale(SqlTypeName) and mark SqlTypeName#getMinScale as deprecated.
4. Add Javadoc in RelDataTypeSystem#getMaxScale, RelDataTypeSystem#getMinScale, RelDataTypeSystem#getDefaultScale, and RelDataTypeSystem#getMinPrecision.
5. Add getMinNumericScale in RelDataTypeSystem. getMinScale will use it when extracting decimal min scale（Same as getMaxPrecision(SqlTypeName) and getMaxNumericPrecision).
6. I use RelDataType.SCALE_NOT_SPECIFIED as the default minimum scale for the type that doesn’t support scale. Not the Integer.MAX_VALUE. Because I think 
RelDataType.SCALE_NOT_SPECIFIED is enough and makes the code logically simple.

PR #3950, opened 2024-09-05
[CALCITE-6559] Query with measure that applies AVG to SMALLINT throws AssertionError "Cannot add expression of different type to set"

julianhyde:


core/src/test/resources/sql/measure.iq
```diff
@@ -179,6 +179,53 @@ from empm;
...
+| CLERK     |      7769 |
+| MANAGER   |      7682 |
+| PRESIDENT |      7839 |
+| SALESMAN  |      7629 |
++-----------+-----------+
+(5 rows)
+
+!ok
+
+!if (false) {
```
mihaibudiu, line 205: Is there an issue filed for this other problem?

mihaibudiu (approved):

julianhyde:
@amarkowitz, PTAL

core/src/test/resources/sql/measure.iq
```diff
@@ -179,6 +179,53 @@ from empm;
 
 !ok
 
+# [CALCITE-6559] LogicalCorrelate Type mismatch
+# Occurs when AVG measure is applied to SMALLINT or TINYINT column.
+# A CAST is added internally to the result of SUM / COUNT, but this cast
```
amarkowitz, line 184: It's difficult to tell from this comment whether this was the cause of the bug or whether this is the new expectation after fixing.  Have you considered moving this comment to CALCITE-6559 instead?

core/src/test/resources/sql/measure.iq
julianhyde: I've logged https://issues.apache.org/jira/browse/CALCITE-6564

PR #3953, opened 2024-09-05
Add cast.iq, a Quidem test for CAST

NobiGo:


core/src/test/resources/sql/cast.iq
```diff
@@ -0,0 +1,1237 @@
...
+| true   |
++--------+
+(14 rows)
+
+!ok
+EnumerableCalc(expr#0..7=[{inputs}], expr#8=[CAST($t6):INTEGER], expr#9=[IS NOT NULL($t8)], EXPR$0=[$t9])
+  EnumerableTableScan(table=[[scott, EMP]])
+!plan
+
+# End cast.iq
```
mihaibudiu, line 1237: there seem to be about 100 new lines in this file, if you can point out what they are it will make reviewing easier

NobiGo: Hi, I set the output format to `mysql` in the file head. Some tests are `CSV ` before. That's why it includes about 100 new lines in this file.

mihaibudiu (approved):

PR #3954, opened 2024-09-09
[CALCITE-6550] Improve SQL function overloading

normanj-bitquill:
* RexImpTable can better handle collisions in the scalar function map
* Return the implementor if only one implementor is found for an operator key
* If there are multiple implementors for an operator key, look for one with the exact same operator
* An operator key is the operator name and kind

The main idea here is to allow collisions in the scalar function map. The collisions are resolved as described above. Much of the change here is managing the scalar function map in `RexImpTable`.

Alternate solutions:
* Fix up the hashcode and equals methods for `SqlBasicFunction` options. There will still be some collisions in `RexImpTable`, since a function could be defined multiple times (ie it is in multiple function libraries with different implementations).
* Reconsider putting all functions in `RexImpTable`. We should only need the functions for selected libraries at any given time.

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -53,6 +53,7 @@ import org.apache.calcite.schema.ImplementableFunction;
 import org.apache.calcite.schema.impl.AggregateFunctionImpl;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlBinaryOperator;
+import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlIntervalQualifier;
 import org.apache.calcite.sql.SqlJsonConstructorNullClause;
 import org.apache.calcite.sql.SqlJsonEmptyOrError;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -528,24 +531,94 @@ public class RexImpTable {
...
     this.tvfImplementorMap = ImmutableMap.copyOf(builder.tvfImplementorMap);
   }
 
+  /**
+   * A pairing of an SqlOperator and a RexCallImplementor. OpImplementor pairs
+   * will be values in the scalar operators map. This allows an "==" check on
+   * the SqlOperator to select the correct operator since the hash of operators
+   * may not be unique.
+   */
+  private static class OpImplementatorPair {
```
julianhyde, line 555: Could you use `Pair<SqlOperator, RexCallImplementor>` instead? I don't think `hashCode` or `equals` are ever called on the values in a map or array-list-multimap.

julianhyde: Or, maybe better, change `scalarMap` to type `Map<SqlOperator, ImmutablePairList<SqlOperator, RexCallImplementor>>`.

normanj-bitquill: I have rebased this on top of your change to refactor `RexImpTable`. I am now using `ImmutablePairList`.

caicancai (commented):
I will test it locally in the next few days to see if there are any other problems. Overall, it is good.

@normanj-bitquill If I test next week this week, I think we can merge this PR next week without introducing more conflicts

caicancai (approved):
LGTM

mihaibudiu:
@julianhyde you asked for changes, please approve if you are satisfied.

PR #3956, opened 2024-09-10
[CALCITE-6566] JDBC adapter should generate PI function with parenthe…

NobiGo:
…ses When unparse PI function without parentheses

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
@@ -1933,7 +1933,7 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
   public static final SqlFunction PI =
       SqlBasicFunction.create("PI", ReturnTypes.DOUBLE, OperandTypes.NILADIC,
           SqlFunctionCategory.NUMERIC)
-          .withSyntax(SqlSyntax.FUNCTION_ID);
+          .withSyntax(SqlSyntax.FUNCTION_ID_CONSTANT);
 
   /** {@code FIRST} function to be used within {@code MATCH_RECOGNIZE}. */
   public static final SqlFunction FIRST =
--- a/core/src/main/java/org/apache/calcite/sql/SqlSyntax.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlSyntax.java
@@ -148,6 +148,23 @@ public enum SqlSyntax {
   },
 
   /**
+   * Function syntax which takes no parentheses and return specific constant value, for
+   * example "PI".
+   *
+   * @see SqlConformance#allowNiladicConstantWithoutParentheses()
+   */
+  FUNCTION_ID_CONSTANT(FUNCTION) {
+    @Override public void unparse(
+        SqlWriter writer,
+        SqlOperator operator,
+        SqlCall call,
+        int leftPrec,
+        int rightPrec) {
+      SqlUtil.unparseFunctionSyntax(operator, writer, call, false);
+    }
+  },
+
+  /**
    * Syntax of an internal operator, which does not appear in the SQL.
    */
   INTERNAL {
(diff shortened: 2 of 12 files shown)


caicancai (approved):

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -1815,7 +1815,7 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
 
   /** The {@code PI} function. */
   public static final SqlFunction PI =
-      SqlBasicFunction.create("PI", ReturnTypes.DOUBLE, OperandTypes.NILADIC,
+      SqlBasicFunction.create(SqlKind.PI, ReturnTypes.DOUBLE, OperandTypes.NILADIC,
```
caicancai, line 1818: Does this have to be changed to SqlKind.PI?
The others look good

YiwenWu:
Based on the discussion in jira, is there a more general way to implement this rather than being specific to `PI` functions?

NobiGo:
> Based on the discussion in jira, is there a more general way to implement this rather than being specific to `PI` functions?

@YiwenWu @caicancai Thanks for the review. We can discuss this in Jira.

caicancai (approved):
LGTM

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -1854,7 +1854,7 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
   public static final SqlFunction PI =
       SqlBasicFunction.create("PI", ReturnTypes.DOUBLE, OperandTypes.NILADIC,
           SqlFunctionCategory.NUMERIC)
-          .withSyntax(SqlSyntax.FUNCTION_ID);
+          .withSyntax(SqlSyntax.FUNCTION_ID_CONSTANT);
```
mihaibudiu, line 1857: aren't there more functions that need this syntax?
I expect many system functions would

NobiGo: This type of function is special because of the current implementation of Calcite. In fact, this type of function requires parentheses to be called in most databases (I haven't found any other databases yet).

mihaibudiu: what about all the system functions like USER, CURRENT_TIME, or functions like RANDOM?

NobiGo: The Function like USER、CURRENT_TIME、CURRENT_USER、SESSION_USER、SYSTEM_USER、CURRENT_PATH、CURRENT_ROLE、CURRENT_CATALOG、CURRENT_SCHEMA、LOCALTIME、LOCALTIMESTAMP、CURRENT_TIME and CURRENT_TIMESTAMP behaviour controlled by `allowNiladicParentheses`. Before this PR, The PI is also controlled by `allowNiladicParentheses`, but this will result in differences in behavior compared to other databases. In other databases the PI always needs parentheses. I add `allowNiladicConstantWithoutParentheses`.On the one hand, it is to be compatible with the current implementation of Calcite, and on the other hand, to make its calls more in line with the usage of most databases.

NobiGo: Yes. The PI behavior is different from the User. The PI always needs to use PI(). Before this PR, Calcite treats PI and USER as a class of functions.

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
mihaibudiu: I think this is an important point, and it would be nice to have this in the code somewhere as a comment.

NobiGo:

PR #3959, opened 2024-09-11
[CALCITE-6576] SqlParser: allow using table alias with column identifiers in SET section of UPDATE statement

nielspardon:
Changing the column identifier from `SimpleIdentifier` to `CompoundIdentifier` allows using the table alias with the column identifiers in the SET section of an UPDATE statement.

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
+++ b/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
@@ -5102,6 +5102,17 @@ public class SqlParserTest {
             + "WHERE (`EMPNO` = 12)");
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6576">[CALCITE-6576]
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -5102,6 +5102,12 @@ void checkPeriodPredicate(Checker checker) {
             + "WHERE (`EMPNO` = 12)");
   }
 
+  @Test void testUpdateTableAlias() {
```
caicancai, line 5105: Add jira link

caicancai (approved):
Overall LGTM

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
nielspardon: I added a comment with the link to the Jira issue

PR #3962, opened 2024-09-11
[CALCITE-6572] Add more tests for NULL result of TO_CHAR PostgreSQL f…

normanj-bitquill:
…unction

* Added a test for first argument NULL without a cast
* Added a test for both arguments NULL without a cast


NobiGo (approved):

caicancai (approved):
Maybe this doesn't require creating a jira case

julianhyde (commented):
+1 when you add a newline in line 4897

mihaibudiu (approved):

PR #3966, opened 2024-09-13
[CALCITE-6570] Add SCALAR_QUERY to sourceExpressionList of SqlUpdate

xtern:
Validation currently fails with an AssertionError on an UPDATE query with a sub-query that requires an implicit type cast. (See CALCITE-6570 for details).

Example
```sql
update emp set empno = (
    select cast(empno as BIGINT) from emp
)
```

During validation, when registering subqueries, Calcite wraps the sub-query in a SCALAR_QUERY call in `selectList` (`sqlUpdate.getSourceSelect().getSelectList()`).

But after that, `TypeCoersion` tries to coerce types for expressions in `sourceExpressionList` (see `TypeCoersionImpl.coerceSourceRowType()`), but it contains `SqlSelect` instead of `SCALAR_QUERY` call.

As a solution I added `SCALAR_QUERY` to `sourceExpressionList` as well.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
@@ -2941,6 +2941,19 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
           SqlSelect.HAVING_OPERAND);
       registerSubQueries(selectScope2,
           SqlNonNullableAccessors.getSelectList(select));
+
+      if (enclosingNode.getKind() == SqlKind.UPDATE) {
+        registerSubQueries(selectScope2,
+            ((SqlUpdate) enclosingNode).getSourceExpressionList());
+      } else if (enclosingNode.getKind() == SqlKind.MERGE) {
+        SqlUpdate updateCall = ((SqlMerge) enclosingNode).getUpdateCall();
+
+        if (updateCall != null) {
+          registerSubQueries(selectScope2,
+              updateCall.getSourceExpressionList());
+        }
+      }
+
       final SqlNodeList orderList = select.getOrderList();
       if (orderList != null) {
         // If the query is 'SELECT DISTINCT', restrict the columns
--- a/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
@@ -3121,6 +3121,18 @@ class SqlToRelConverterTest extends SqlToRelTestBase {
 
   /**
    * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6570">[CALCITE-6570]
+   * UPDATE with sub-query that requires type cast gives AssertionError</a>.
+   */
+  @Test void testUpdateSubQueryWithCast() {
+    final String sql = "update emp\n"
+        + "set empno = (\n"
+        + "  select cast(min(empno) as BIGINT) from emp as e where e.deptno = emp.deptno)";
+    sql(sql).ok();
+  }
+
+  /**
+   * Test case for
    * <a href="https://issues.apache.org/jira/browse/CALCITE-3229">[CALCITE-3229]
    * UnsupportedOperationException for UPDATE with IN query</a>.
    */
@@ -3210,6 +3222,20 @@ class SqlToRelConverterTest extends SqlToRelTestBase {
         + "set ename = t.ename, deptno = t.deptno, sal = t.sal * .1\n"
         + "when not matched then insert (empno, ename, deptno, sal)\n"
         + "values(t.empno, t.ename, 10, t.sal * .15)";
+    sql(sql).ok();
+  }
+
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6570">[CALCITE-6570]
+   * UPDATE with sub-query that requires type cast gives AssertionError</a>.
+   */
+  @Test void testMergeSubQueryWithCast() {
+    final String sql = "merge into emp t0\n"
+        + "using emp t1 ON t0.empno = t1.empno\n"
+        + "when matched then\n"
+        + "update set deptno = (select cast(deptno as BIGINT) from emp where deptno > 1)";
+
     sql(sql).ok();
   }
 
--- a/core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
@@ -4711,6 +4711,28 @@ LogicalTableModify(table=[[CATALOG, SALES, EMPNULLABLES]], operation=[MERGE], up
 ]]>
     </Resource>
   </TestCase>
+  <TestCase name="testMergeSubQueryWithCast">
+    <Resource name="sql">
+      <![CDATA[merge into emp t0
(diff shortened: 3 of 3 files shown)


core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -3119,6 +3119,13 @@ void checkCorrelatedMapSubQuery(boolean expand) {
     sql(sql).ok();
   }
 
+  @Test void testUpdateSubQueryWithCast() {
```
caicancai, line 3122: Add jira link

caicancai (approved):

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
xtern: Done, thanks

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -2941,6 +2941,12 @@ private void registerQuery(
           SqlSelect.HAVING_OPERAND);
       registerSubQueries(selectScope2,
           SqlNonNullableAccessors.getSelectList(select));
+
+      if (enclosingNode.getKind() == SqlKind.UPDATE) {
```
mihaibudiu, line 2945: I wonder whether there are other statements that would need this.

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
xtern: Thanks, at least MERGE suffers from the same issue.
Added the corresponding test and fix.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -2941,6 +2941,19 @@ private void registerQuery(
...
+        registerSubQueries(selectScope2,
+            ((SqlUpdate) enclosingNode).getSourceExpressionList());
+      } else if (enclosingNode.getKind() == SqlKind.MERGE) {
+        SqlUpdate updateCall = ((SqlMerge) enclosingNode).getUpdateCall();
```
mihaibudiu, line 2949: this looks wrong, you mean SqlMerge?

xtern: :thinking: what exactly looks wrong?
Here we register subqueries that may be present in updateCall (SqlUpdate), which may be present in SqlMerge statement.

mihaibudiu: Sorry, my mistake.

xtern:
@mihaibudiu, @caicancai, 
we found one issue with this patch that needs attention.
After this fix, SCALAR_QUERY is present in sourceExpressionList of original `SqlUpdate` AST node.
The `SqlToRelConverter` test uses the deprecated `withExpand(false)` config setting, and as a result, SCALAR_QUERY is replaced to a constant.
And that's ok, but when using ``withExpand(true)`  this SCALAR_QUERY will not be replaced with a constant even at planning phase (may be the problem in `SubQueryRemoveRule` or somewhere else).
As a result, we can observe in the "physical" plan the original SCALAR_SUBQUERY (with LOGICAL nodes) in the `sourceExpressionList` of `TableModify` :thinking: 

If the patch needs to be improved, then feel free to remove it from the current scope (1.38.0), so that it doesn't delay the release.

caicancai:
> @mihaibudiu, @caicancai, we found one issue with this patch that needs attention. After this fix, SCALAR_QUERY is present in sourceExpressionList of original `SqlUpdate` AST node. The `SqlToRelConverter` test uses the deprecated `withExpand(false)` config setting, and as a result, SCALAR_QUERY is replaced to a constant. And that's ok, but when using `withExpand(true)` this SCALAR_QUERY will not be replaced with a constant even at planning phase (may be the problem in `SubQueryRemoveRule` or somewhere else). As a result, we can observe in the "physical" plan the original SCALAR_SUBQUERY (with LOGICAL nodes) in the `sourceExpressionList` of `TableModify` 🤔
> 
> If the patch needs to be improved, then feel free to remove it from the current scope (1.38.0), so that it doesn't delay the release.

You mean that it can only be recognized as a constant in the physical stage? It feels strange.

xtern:
> > @mihaibudiu, @caicancai, we found one issue with this patch that needs attention. After this fix, SCALAR_QUERY is present in sourceExpressionList of original `SqlUpdate` AST node. The `SqlToRelConverter` test uses the deprecated `withExpand(false)` config setting, and as a result, SCALAR_QUERY is replaced to a constant. And that's ok, but when using `withExpand(true)` this SCALAR_QUERY will not be replaced with a constant even at planning phase (may be the problem in `SubQueryRemoveRule` or somewhere else). As a result, we can observe in the "physical" plan the original SCALAR_SUBQUERY (with LOGICAL nodes) in the `sourceExpressionList` of `TableModify` 🤔
> > If the patch needs to be improved, then feel free to remove it from the current scope (1.38.0), so that it doesn't delay the release.
> 
> @xtern Hi. You mean that it can only be recognized as a constant in the physical stage? It feels strange.

In my previous message I mixed up the `withExpand` flag values .I fixed the message. 
By default the `isExpand() = false`, but in `SqlToRelConverterTest uses` `isExpand() = true`

Let's turn it off and see the difference in plan tree in existing test `testUpdateSubQuery` (it's not new test) 

Without this fix the plan is:
```
LogicalTableModify(table=[[CATALOG, SALES, EMP]], operation=[UPDATE], updateColumnList=[[EMPNO]], sourceExpressionList=[[$0]], flattened=[true])
  LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], EXPR$0=[$SCALAR_QUERY({
LogicalAggregate(group=[{}], EXPR$0=[MIN($0)])
  LogicalProject(EMPNO=[$0])
    LogicalFilter(condition=[=($7, $cor1.DEPTNO)])
})])
```

Note that `sourceExpressionList=[[$0]]`, and SCALAR_QUERY resides only under additional logical project.

But after applying this patch the plan is: 
```
LogicalTableModify(table=[[CATALOG, SALES, EMP]], operation=[UPDATE], updateColumnList=[[EMPNO]], sourceExpressionList=[[$SCALAR_QUERY({
LogicalAggregate(group=[{}], EXPR$0=[MIN($0)])
  LogicalProject(EMPNO=[$0])
    LogicalFilter(condition=[=($7, $cor1.DEPTNO)])
})]], flattened=[true])
  LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], EXPR$0=[$SCALAR_QUERY({
LogicalAggregate(group=[{}], EXPR$0=[MIN($0)])
  LogicalProject(EMPNO=[$0])
    LogicalFilter(condition=[=($7, $cor0.DEPTNO)])
})])
```
`sourceExpressionList` contains the original SCALAR_QUERY, and this looks like an issue that needs to be fixed before merging this PR :pensive:

PR #3967, opened 2024-09-13
[CALCITE-6580] Remove Locale.setDefault

normanj-bitquill:
* Remove its usage in tests where possible
* Needed for a few tests that are testing locale translations
* If a test uses Locale.setDefault, then it is isolated to avoid affecting other tests


normanj-bitquill:
Looks this was merged in with another change. Closing this PR.

PR #3968, opened 2024-09-15
[CALCITE-6581] Incorrect INTERVAL math for WEEK and QUARTER

gianm:
WEEK and QUARTER intervals were incorrectly treated like HOUR and MONTH respectively. This patch fixes the handling, and adds unit tests.

This patch also renames the two "fillIntervalValueArray" methods to "fillDayIntervalValueArray" (for INTERVAL_DAY_TIME type intervals) and "fillYearMonthIntervalValueArray" (for INTERVAL_YEAR_MONTH type intervals) to make it more clear which one is which.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlIntervalQualifier.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlIntervalQualifier.java
@@ -575,7 +575,7 @@ public class SqlIntervalQualifier extends SqlNode {
     return new BigDecimal("0." + secondFracStr).multiply(THOUSAND);
   }
 
-  private static int[] fillIntervalValueArray(
+  private static int[] fillYearMonthIntervalValueArray(
       int sign,
       BigDecimal year,
       BigDecimal month) {
@@ -588,7 +588,7 @@ public class SqlIntervalQualifier extends SqlNode {
     return ret;
   }
 
-  private static int[] fillIntervalValueArray(
+  private static int[] fillDayTimeIntervalValueArray(
       int sign,
       BigDecimal day,
       BigDecimal hour,
@@ -636,7 +636,7 @@ public class SqlIntervalQualifier extends SqlNode {
       checkLeadFieldInRange(typeSystem, sign, year, TimeUnit.YEAR, pos);
 
       // package values up for return
-      return fillIntervalValueArray(sign, year, ZERO);
+      return fillYearMonthIntervalValueArray(sign, year, ZERO);
     } else {
       throw invalidValueException(pos, originalValue);
     }
@@ -676,7 +676,7 @@ public class SqlIntervalQualifier extends SqlNode {
       }
 
       // package values up for return
-      return fillIntervalValueArray(sign, year, month);
+      return fillYearMonthIntervalValueArray(sign, year, month);
     } else {
       throw invalidValueException(pos, originalValue);
     }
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/sql/SqlIntervalQualifier.java
```diff
@@ -746,7 +746,8 @@ private int[] evaluateIntervalLiteralAsQuarter(
       checkLeadFieldInRange(typeSystem, sign, quarter, TimeUnit.QUARTER, pos);
 
       // package values up for return
-      return fillIntervalValueArray(sign, ZERO, quarter);
+      final BigDecimal months = quarter.multiply(BigDecimal.valueOf(3));
```
mihaibudiu, line 749: I don't think quarters should be convertible to seconds, they do not have a fixed length

gianm: They are being converted to 3 months here, since QUARTER is treated as a kind of YEAR-MONTH interval.

core/src/test/java/org/apache/calcite/sql/parser/SqlParserUtilTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * Tests {@link SqlParserUtil}.
+ */
+public class SqlParserUtilTest {
```
caicancai, line 30: SqlParserIntervalQualifierTest?

core/src/main/java/org/apache/calcite/sql/SqlIntervalQualifier.java
mihaibudiu: sorry, I misread the JIRA issue

mihaibudiu (approved):

core/src/test/java/org/apache/calcite/sql/parser/SqlParserUtilTest.java
gianm: I figured that at some point, this test file might be expanded to include more tests for `SqlParserUtil`, so I named it after the main class itself. However, please let me know if it's considered better to scope the test files more narrowly.

mihaibudiu:
Can you add a couple of end-to-end tests, either to SqlOperatorTest, or to one of the *.iq files?

core/src/test/java/org/apache/calcite/sql/parser/SqlParserUtilTest.java
caicancai: It seems that the impact is not great.
I think we can add JavaDoc to let subsequent contributors know what this test file is for.

caicancai:
> Can you add a couple of end-to-end tests, either to SqlOperatorTest, or to one of the *.iq files?

+1

caicancai (approved):

gianm:
> Can you add a couple of end-to-end tests, either to SqlOperatorTest, or to one of the *.iq files?

Sure, I've added a test to `sql/misc.iq`.

core/src/test/java/org/apache/calcite/sql/parser/SqlParserUtilTest.java
gianm: I've added some additional javadoc.

core/src/main/java/org/apache/calcite/sql/SqlIntervalQualifier.java
gianm: Ah, ok!

gianm:
Committing; thank you for the reviews.

mihaibudiu:
Don't forget to mark the jira case as closed and to record the actual commit

gianm:
> Don't forget to mark the jira case as closed and to record the actual commit

Thank you for reminding me!

PR #3969, opened 2024-09-15
[CALCITE-6527] Add DATE_ADD function (enabled in Spark library)

caicancai:
The following is my recent review of why the BC date is equal to 0000.
In calcite, the date function will go through a series of conversions, such as the bigquery date_add function will be converted to DATETIME_PLUS later,and the current adapted DATE_ADD is no exception.

After being converted to DATETIME_PLUS, it seems that it will be converted to PLUS operator operation later, and finally enter `RexlmpTable.DatetimeArithmeticImplementor `to perform parameter type conversion and partial calculation, `return normalize(typeName, Expressions.add(trop0, trop1))`. Of course, there may be higher levels of abstraction and conversion later, but unfortunately I haven't found it yet.

In this process, I found that calcite did not restrict the result set of the date function to the date BC date and 0000. I raised [jira](https://issues.apache.org/jira/browse/CALCITE-6544) and emailed to seek answers, there is Julian's explanation in jira, and calcite has its own DATE data rules.

Why 0000 and BC dates are not allowed as parameters? This is because calcite has already made restrictions on dates when parsing.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -132,6 +132,16 @@ public abstract class SqlLibraryOperators {
   public static final SqlFunction DATEADD =
       new SqlTimestampAddFunction("DATEADD");
 
+  /** The "DATE_ADD(start_date, num_days)" function
+   * (Spark) Returns the date that is num_days after start_date. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction DATE_ADD_SPARK =
+      SqlBasicFunction.create(SqlKind.DATE_ADD, ReturnTypes.DATE_NULLABLE,
+              OperandTypes.DATE_ANY)
+          .withFunctionType(SqlFunctionCategory.TIMEDATE);
+
+  /** The "ADD_MONTHS(start_date, num_months)" function
+   * (SPARK) Returns the date that is num_months after start_date. */
   @LibraryOperator(libraries = {ORACLE, SPARK})
   public static final SqlFunction ADD_MONTHS =
       SqlBasicFunction.create(SqlKind.ADD_MONTHS, ReturnTypes.ARG0_NULLABLE,
--- a/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
@@ -209,6 +209,8 @@ public class StandardConvertletTable extends ReflectiveConvertletTable {
         new TimestampAddConvertlet());
     registerOp(SqlLibraryOperators.ADD_MONTHS,
         new TimestampAddConvertlet());
+    registerOp(SqlLibraryOperators.DATE_ADD_SPARK,
+        new TimestampAddConvertlet());
     registerOp(SqlLibraryOperators.DATE_DIFF,
         new TimestampDiffConvertlet());
     registerOp(SqlLibraryOperators.DATE_SUB,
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -133,6 +133,16 @@ private SqlLibraryOperators() {
   public static final SqlFunction DATEADD =
       new SqlTimestampAddFunction("DATEADD");
 
+  /** The "DATE_ADD(start_date, num_days)" function
+   * (Spark) Returns the date that is num_days after start_date. */
+  @LibraryOperator(libraries = {SPARK})
+  public static final SqlFunction DATE_ADD_SPARK =
```
NobiGo, line 139: DATE_ADD is enough. The LiberaryOperator will let us know if it follows the Spark style.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -2150,8 +2152,11 @@ private static class TimestampAddConvertlet implements SqlRexConvertlet {
       switch (call.operandCount()) {
       case 2:
         // Oracle-style 'ADD_MONTHS(date, integer months)'
-        if (call.getOperator() == SqlLibraryOperators.ADD_MONTHS) {
-          qualifier = new SqlIntervalQualifier(TimeUnit.MONTH, null, SqlParserPos.ZERO);
+        if (call.getOperator() == SqlLibraryOperators.ADD_MONTHS
```
NobiGo, line 2155: Use another condition to do this. Don't need to merge it.

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
caicancai: @NobiGo Bigquery's DATE_ADD already exists. Duplicate naming will cause conflicts.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
caicancai: I optimized the code to make it more readable.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13910,6 +13910,63 @@ void testTimestampDiff(boolean coercionEnabled) {
...
+    f.checkScalar("date_add(date '2016-02-28', 2.0)",
+        "2016-03-01",
+        "DATE NOT NULL");
+    f.checkScalar("date_add(date '2016-02-28', '2.0')",
+        "2016-03-01",
+        "DATE NOT NULL");
+    f.checkScalar("date_add(timestamp '2016-02-22 13:00:01', '-2.0')",
+        "2016-02-20",
+        "DATE NOT NULL");
+    f.checkScalar("date_add(timestamp '2016-02-22 13:00:01', -2)",
```
NobiGo, line 13950: Add test like:
```
spark-sql> select date_add('2016-02-22 13:00:01', -2);
2016-02-20
spark-sql> select date_add('13:00:01', -2);
NULL
Time taken: 0.195 seconds, Fetched 1 row(s)
spark-sql> select date_add('1', -2);
NULL
spark-sql> select date_add(1, -2);
Error in query: cannot resolve 'date_add(1, -2)' due to data type mismatch: argument 1 requires date type, however, '1' is of int type.; line 1 pos 7;
'Project [unresolvedalias(date_add(1, -2), None)]
+- OneRowRelation
```

caicancai: @NobiGo Calcite has its own date type rules, which are different from Spark. I am wondering whether it is necessary to return null in select date_add('13:00:01', -2)🤔.

The dataricks documentation states that only date expressions are accepted，https://docs.databricks.com/en/sql/language-manual/functions/date_add.html.

I think you can refer to the NVL2 function I adapted before, https://issues.apache.org/jira/browse/CALCITE-6397

NobiGo (approved):

caicancai:
@NobiGo no other comment. squashed

site/_docs/reference.md
```diff
@@ -2804,6 +2804,7 @@ In the following:
 | b s | DATE_FROM_UNIX_DATE(integer)                 | Returns the DATE that is *integer* days after 1970-01-01
 | p r | DATE_PART(timeUnit, datetime)                | Equivalent to `EXTRACT(timeUnit FROM  datetime)`
 | b | DATE_ADD(date, interval)                       | Returns the DATE value that occurs *interval* after *date*
+| s | DATE_ADD(date, num_days)                       | Returns the DATE that is *num_days* after *date*
```
julianhyde, line 2807: use camelCase for arguments, not snake_case.

caicancai:
@julianhyde @NobiGo Thanks for your review

caicancai:
@NobiGo If there are no other comments we can merge this PR this week

PR #3970, opened 2024-09-18

caicancai:

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorUtil.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorUtil.java
@@ -1323,6 +1323,69 @@ public class SqlValidatorUtil {
   }
 
   /**
+   * Adjusts the types of specified operands in a map operation to match a given target type.
+   * This is particularly useful in the context of SQL operations involving array functions,
+   * where it's necessary to ensure that all operands have consistent types for the operation
+   * to be valid.
+   *
+   * <p>This method operates on the assumption that the operands to be adjusted are part of a
+   * {@link SqlCall}, which is bound within a {@link SqlOperatorBinding}. The operands to be
+   * cast are identified by their indexes within the {@code operands} list of the {@link SqlCall}.
+   * The method performs a dynamic check to determine if an operand is a basic call to a map.
+   * If so, it casts each element within the map to the target type.
+   * Otherwise, it casts the operand itself to the target type.
+   *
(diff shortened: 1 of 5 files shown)


caicancai:
@rubenada Looks like my local git crashed for me, I will open a new PR. Sorry.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorUtil.java
```diff
@@ -1322,6 +1322,69 @@ public static boolean isMeasure(SqlNode selectItem) {
...
+      }
+      requireNonNull(operandTypes, "operandType of" + operandTypes);
+
+      if (!operandTypes.equalsSansFieldNames(elementType)) {
+        operandsmap.add(i, castTo(operands.get(i), elementType));
+      } else {
+        operandsmap.add(i, operands.get(i));
+      }
+    }
+    call.setOperand(0, castMapTo(operandsmap));
```
mihaibudiu, line 1370: the question we haven't resolved properly is whether the type checker is allowed to mutate the expression that is being checked.

rubenada: Uhmm, it seems there is already some existing code that calls `call.setOperand` on this file, e.g. inside `adjustTypeForArrayFunctions` or `adjustTypeForMultisetConstructor`; however, these methods seem to be called from a `SqlReturnTypeInference` (not from a `SqlOperandTypeInference`). Am I correct? 
@caicancai  would it be possible to adapt this patch so that the new code is contained in a `SqlReturnTypeInference` (where there is precedence on Calcite about such modifications)?

caicancai: @rubenada ok, My first version of the solution is actually to modify SqlOperandTypeInference. I will fix this pr later.

mihaibudiu: Is there a reason to think these changes were not wrong as well?

mihaibudiu: If the ReturnTypeInference is allowed to change the arguments, hopefully this fact is documented somewhere

caicancai: I will send an email to discuss later. If it is not suitable, you can remove 1.38.0 from this jira.

PR #3971, opened 2024-09-18
[CALCITE-6587] Support Java 23 and Guava 33.3.0

julianhyde:
@libenchao, Can you review this? Part of it relates to deprecated methods in Avatica, and [CALCITE-5136].

I'd like to get this change into Calcite 1.38, but the Avatica fixes can wait a few weeks.


libenchao (approved):
LGTM. (I finally recovered my Github account via disabling two-factor verification by Github staff)

elasticsearch/src/main/java/org/apache/calcite/adapter/elasticsearch/PredicateAnalyzer.java
```diff
@@ -117,7 +117,7 @@ static QueryBuilder analyze(RexNode expression) throws ExpressionNotAnalyzableEx
       }
       return e != null ? e.builder() : null;
     } catch (Throwable e) {
-      Throwables.propagateIfPossible(e, UnsupportedOperationException.class);
+      Throwables.throwIfInstanceOf(e, UnsupportedOperationException.class);
```
stoty, line 120: According to the Javadoc, throwIfInstanceOf is not a 1:1 replacement for propagateIfPossible.

The old code would throw other RunTime exceptions directly, while the new one wraps them.

This may be a good thing, but the commit message wording does not suggest behavioural changes.

julianhyde: I wasn't aware that the methods have different behavior. Thanks for bringing it to my attention.

However, there is no obligation to document behavioral changes that occur in 'behavior is unspecified' territory.

This block - like pretty much every `catch (Throwable)` block - is in that territory, and I didn't think too much about which use cases would be affected because - by definition - there are no documented, tested use cases.

PR #3972, opened 2024-09-19
[CALCITE-6584] Validate prefixed column identifiers in SET clause of UPDATE statement

nielspardon:
This PR adds validation of prefixed column identifiers in SET clause of UPDATE statement (and in the UPDATE statement within a MERGE statement)

- This is a follow-up PR to https://github.com/apache/calcite/pull/3959
- I first had to change `SqlValidatorUtil.addFields()` to use the `Util.last()` method for getting the last element of a compound identifier instead of relying on simple identifiers
- I then implemented a validation logic in `SqlValidatorImpl.createTargetRowType()` which is being used for validating INSERT, UPDATE and MERGE
- in presence of the new `targetTableAlias` parameter the validation logic checks the prefix of any non-simple identifiers or in other words it checks that any prefixed column identifiers are prefixed with the target table alias
- I tried to check for any existing validation logic that I could reuse and it's possible I might have missed a utility method that I could have used.
- The validation logic supports a mix of prefixed and unprefixed columns in the SET clause. Not sure if I should only allow prefixed columns here.

Let me know your thoughts.

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
@@ -4531,6 +4531,23 @@ public class SqlValidatorTest extends SqlValidatorTestCase {
         .fails("Table 'SALES.BAD' not found");
   }
 
+    /** Test case for
+     * <a href="https://issues.apache.org/jira/browse/CALCITE-6584">[CALCITE-6584]
+     * Validate prefixed column identifiers in SET clause of UPDATE statement</a>. */
+  @Test void testAliasInSetClauseOfUpdate() {
+    // good examples
+    sql("UPDATE sales.emp AS e SET e.deptno = 10").ok();
+    sql("UPDATE emp AS e SET e.deptno = 10").ok();
+
+    // bad examples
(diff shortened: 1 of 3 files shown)


mihaibudiu (approved):

caicancai (approved):
LGTM

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -4531,6 +4531,23 @@ private void checkNegWindow(String s, String msg) {
         .fails("Table 'SALES.BAD' not found");
   }
 
+    /** Test case for
+     * <a href="https://issues.apache.org/jira/browse/CALCITE-6584">[CALCITE-6584]
+     * Validate prefixed column identifiers in SET clause of UPDATE statement</a>. */
+  @Test void testAliasInSetClauseOfUpdate() {
+    // good examples
+    sql("UPDATE sales.emp AS e SET e.deptno = 10").ok();
+    sql("UPDATE emp AS e SET e.deptno = 10").ok();
```
mihaibudiu, line 4540: postgres actually rejects this: `Query Error: error: column "e" of relation "emp" does not exist`

PR #3976, opened 2024-09-24
[CALCITE-6586] Some Rules not firing due to RelMdPredicates returning…

suibianwanwank:
This PR supports getPredicates(RelSubset) returning getPredicate(stripped) in place of EMPTY as a temporary solution.
But it exceeds the expected range because some rules are re-fire in VolcanoPlanner.
Major Changes: 
In SortRemoveConstantKey the constant key in the sort is removed. before this PR the traitset of the sort needs to be consistent with the collation. It would result in the new sort having a different traitset than the original one, it would not be in the same RelSubSet and would not be selected.
So I tried to modify the traitSet assert in sort so that it can be set to RelComposetrait.
For example : order by a,b,c 
If we can make sure that b is constant, then the traitSet of the sort can be [[a,b,c],[a,c]]. Also, when collation is empty and there are only offset and fetch, the original input collation traits can be kept in the sort.
There are also a few more changes that I will add comments on. For the Sort changes, we can probably discuss them in Jira if needed. Thanks for the suggestion and review.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
@@ -456,49 +456,21 @@ public class RelMdColumnUniqueness
       ImmutableBitSet columns, boolean ignoreNulls) {
     columns = decorateWithConstantColumnsFromPredicates(columns, rel, mq);
     for (RelNode rel2 : rel.getRels()) {
-      if (rel2 instanceof Aggregate
-          || rel2 instanceof Filter
-          || rel2 instanceof Values
-          || rel2 instanceof Sort
-          || rel2 instanceof TableScan
-          || simplyProjects(rel2, columns)) {
-        try {
-          final Boolean unique = mq.areColumnsUnique(rel2, columns, ignoreNulls);
-          if (unique != null) {
-            if (unique) {
-              return true;
-            }
-          } else {
-            return null;
+      try {
+        final Boolean unique = mq.areColumnsUnique(rel2, columns, ignoreNulls);
+        if (unique != null) {
+          if (unique) {
+            return true;
           }
-        } catch (CyclicMetadataException e) {
-          // Ignore this relational expression; there will be non-cyclic ones
-          // in this set.
+        } else {
+          return null;
         }
+      } catch (CyclicMetadataException e) {
+        // Ignore this relational expression; there will be non-cyclic ones
+        // in this set.
       }
     }
     return false;
-  }
-
-  private static boolean simplyProjects(RelNode rel, ImmutableBitSet columns) {
-    if (!(rel instanceof Project)) {
-      return false;
-    }
-    Project project = (Project) rel;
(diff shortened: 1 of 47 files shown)


geode/src/main/java/org/apache/calcite/adapter/geode/rel/GeodeSort.java
```diff
@@ -66,7 +66,7 @@ public class GeodeSort extends Sort implements GeodeRel {
 
   @Override public Sort copy(RelTraitSet traitSet, RelNode input,
       RelCollation newCollation, RexNode offset, RexNode fetch) {
-    return new GeodeSort(getCluster(), traitSet, input, collation, fetch);
```
suibianwanwank, line 69: Use the new traitset in the copy of the sort, and the other sort copies as well.

innodb/src/main/java/org/apache/calcite/adapter/innodb/InnodbFilter.java
```diff
@@ -96,4 +98,9 @@ public static InnodbFilter create(RelOptCluster cluster, RelTraitSet traitSet,
   public RelCollation getImplicitCollation() {
     return indexCondition.getImplicitCollation();
   }
+
+  @Override public RexNode getCondition() {
```
suibianwanwank, line 102: InnodbFIlter's condition contains both indexed and non-indexed conditions, but only the indexed filter will be executed. This will cause the FilterReduceExpression to produce incorrect results.

core/src/test/resources/sql/sub-query.iq
```diff
@@ -922,10 +922,9 @@ EnumerableCalc(expr#0..3=[{inputs}], expr#4=[null:BOOLEAN], expr#5=[IS NOT NULL(
     EnumerableCalc(expr#0..7=[{inputs}], EMPNO=[$t0], SAL=[$t5])
       EnumerableTableScan(table=[[scott, EMP]])
     EnumerableLimit(fetch=[1])
-      EnumerableSort(sort0=[$0], dir0=[DESC])
```
suibianwanwank, line 925: The changes to sub-query.iq are due to JoinOnUniqueToSemiJoinRule and SortRemoveConstantKeysRule taking effect.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -456,51 +456,23 @@ public Boolean areColumnsUnique(Values rel, RelMetadataQuery mq,
       ImmutableBitSet columns, boolean ignoreNulls) {
     columns = decorateWithConstantColumnsFromPredicates(columns, rel, mq);
     for (RelNode rel2 : rel.getRels()) {
-      if (rel2 instanceof Aggregate
```
suibianwanwank, line 459: I'm not sure why this restriction was made, but after pullUpPredicate supported RelSubset,  columns would union the constant field, And return false due to simplyProject.  So I remove the restriction here.

PR #3978, opened 2024-09-24
[CALCITE-6585] In the Postgres TO_CHAR function, improve caching

normanj-bitquill:
* Refactored code to separate parsing a date/time format and applying it
* A parsed date/time format can now be reused
* Fixed up some minor issues with when to space pad day or month names


normanj-bitquill:
Currently the caching is static. Trying to figure out how to get the `DateFormatFunction` instance created with the `DataContext`.

normanj-bitquill:
I have updated to match @julianhyde suggestions. It will now create a new instance of `DateFormatFunctionPg` and pass the `DataContext` in to the constructor.

rorueda:
I've been meaning to ask this for some time. While investigating performance regarding datetime formatting, I found out that the cache is disabled by default, which contradicts the javadoc. Would you know which one is correct? Or am I missing something?

```java
   * <p>The value of this parameter limits the size of every function-level
   * cache in Calcite. The default value is 1,000.
   */
  public static final CalciteSystemProperty<Integer> FUNCTION_LEVEL_CACHE_MAX_SIZE =
      intProperty("calcite.function.cache.maxSize", 0, v -> v >= 0);
```

julianhyde:
@rorueda Wow, good catch. It was my fault too. I've logged https://issues.apache.org/jira/browse/CALCITE-6596 and will fix shortly.

PR #3980, opened 2024-09-25
[CALCITE-6595] Preserve collation on non-distinct aggCalls in AggregateExpandWithinDistinctRule

tjbanghart:


core/src/main/java/org/apache/calcite/rel/rules/AggregateExpandWithinDistinctRule.java
```diff
@@ -411,7 +414,7 @@ int getCount(int filterArg) {
       RelBuilder.AggCall aggCall;
       if (c.distinctKeys == null) {
         aggCall =
-            b.aggregateCall(c.getParserPosition(), SqlStdOperatorTable.MIN,
+            b.aggregateCall(SqlStdOperatorTable.MIN,
```
mihaibudiu: Can you also preserve the parser position?

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -17378,6 +17378,33 @@ LogicalProject(DEPTNO=[$0], EXPR$1=[CASE(=($2, 0), null:INTEGER, $1)], EXPR$2=[$
         LogicalAggregate(group=[{0, 3}], groups=[[{0, 3}, {0}]], agg#0=[$SUM0($1) FILTER $2], agg#1=[COUNT() FILTER $2], agg#2=[MIN($1)], agg#3=[MAX($1)], agg#4=[GROUPING($0, $3)])
           LogicalProject(DEPTNO=[$7], SAL=[$5], $f2=[>($5, 1000)], JOB=[$2])
             LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testWithinDistinctPreservesNonDistinctCollation">
+    <Resource name="sql">
+      <![CDATA[SELECT SUM(sal) WITHIN DISTINCT (job),
+LISTAGG(ename, '; ') WITHIN GROUP (ORDER BY sal DESC)
```
mihaibudiu, line 17387: how is the collation visible in the plan?

tjbanghart: `agg#2=[LISTAGG($3, $4) WITHIN GROUP ([1 DESC])]` is present in the `planAfter`

Before this change the `planAfter` stripped the `WITHIN GROUP ([1 DESC])` i.e. just `agg#2=[LISTAGG($3, $4)]`

mihaibudiu (approved):
These days the custom seems to be to add .iq test. These are much easier to review than reading plans.

tjbanghart:
Sure, i'll try making an iq test as well. `RelOptRulesTest.java` had plenty of `AggregateExpandWithinDistinctRule` tests so was easy to add there.

PR #3983, opened 2024-09-26
[CALCITE-3592] Implement BITNOT scalar function

normanj-bitquill:
* Implemented the BITNOT scalar function as a standard function
* Supports integer and binary arguments
* Returns NULL if the argument is NULL

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -15738,6 +15738,39 @@ public class SqlOperatorTest {
         true);
   }
 
+  /**
+   * Test cases for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-3592">[CALCITE-3592]
+   * Implement BITNOT scalar function</a>. */
+  @Test void testBitNotScalarFunc() {
+    final SqlOperatorFixture f = fixture();
+    f.setFor(SqlStdOperatorTable.BITNOT, VmName.EXPAND);
+    f.checkFails("bitnot(^*^)", "Unknown identifier '\\*'", false);
+    f.checkScalar("bitnot(2)", Integer.toString(~2), "INTEGER NOT NULL");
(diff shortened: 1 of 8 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -15738,6 +15738,29 @@ private static void checkLogicalOrFunc(SqlOperatorFixture f) {
         true);
   }
 
+  @Test void testBitNotScalarFunc() {
+    final SqlOperatorFixture f = fixture();
+    f.setFor(SqlStdOperatorTable.BITXOR, VmName.EXPAND);
```
mihaibudiu, line 15743: XOR?
How about a failing test with DECIMAL?

mihaibudiu (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
normanj-bitquill: Interesting that using XOR had not impact on the tests. Fixed now.

Also added a test for `BITNOT(DECIMAL)`.

mihaibudiu: I think that setFor is just decorative

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -15738,6 +15738,35 @@ private static void checkLogicalOrFunc(SqlOperatorFixture f) {
         true);
   }
 
+  @Test void testBitNotScalarFunc() {
```
caicancai, line 15741: Add jira link

normanj-bitquill: Updated.

caicancai (approved):

normanj-bitquill:

PR #3984, opened 2024-09-27
[CALCITE-6301] Extend ‘Must-filter’ columns to support a conditional bypass list

olivrlee:


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4185,6 +4217,48 @@ private static void forEachQualified(SqlNode node, SqlValidatorScope scope,
     });
   }
 
+  /* If the supplied SqlNode when fully qualified is in the set of bypassQualifieds, then we
+  remove all entries in the qualifieds set as well as remnantMustFilterFields
```
tanclary, line 4221: is this formatted correctly?

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -12235,10 +12235,10 @@ private void checkCustomColumnResolving(String table) {
             + "  where empno = 1)\n"
             + "where j = 'doctor'")
         .ok();
-      // Deceitful alias #2. Filter on 'job' is a filter on the underlying
-      // 'ename', so the underlying 'job' is missing a filter.
+    // Deceitful alias #2. Filter on 'job' is a filter on the underlying
+    // 'slacker', so the underlying 'job' is missing a filter.
```
tanclary, line 12239: nice  (no action req)

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -12379,6 +12379,370 @@ private void checkCustomColumnResolving(String table) {
...
+    fixture.withSql("^select *\n"
+            + "from emp\n"
+            + "where concat(emp.empno, ' ') = 'abc'^")
+        .fails(missingFilters("JOB"));
+
+    fixture.withSql("select *\n"
+            + "from emp\n"
+            + "where concat(emp.ename, ' ') = 'abc'^")
+        .ok();
+    // ENAME is a bypass field
```
tanclary, line 12446: would this belong better above the test case ?

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -12379,6 +12379,370 @@ private void checkCustomColumnResolving(String table) {
...
+    // are still required.
+
+    fixture.withSql("^select *\n"
+            + "from emp a1\n"
+            + "join emp a2 on a1.empno = a2.empno\n"
+            + "where a2.empno = 1\n"
+            + "and a1.empno = 1\n"
+            + "and a2.job = 'doctor'^")
+        .fails(missingFilters("JOB"));
+    // There are two JOB columns but only one is filtered
```
tanclary, line 12564: having the comments below the cases is a little confusing

core/src/main/java/org/apache/calcite/sql/validate/AbstractNamespace.java
```diff
@@ -62,6 +64,10 @@ abstract class AbstractNamespace implements SqlValidatorNamespace {
    * should typically be re-assigned on validate. */
   protected ImmutableBitSet mustFilterFields = ImmutableBitSet.of();
 
+  protected ImmutableBitSet mustFilterBypassFields = ImmutableBitSet.of();
```
julianhyde, line 67: fields need javadoc, like `mustFilterFields`

core/src/main/java/org/apache/calcite/sql/validate/SemanticTable.java
```diff
@@ -44,4 +50,21 @@ public interface SemanticTable {
   default boolean mustFilter(int column) {
     return getFilter(column) != null;
   }
+
+  /**
+   * Returns a list of column ordinals (0-based) of fields that defuse
+   * must-filter columns when filtered on.
+   */
+  default List<Integer> bypassFieldList() {
+    return emptyList();
```
julianhyde, line 59: Use `ImmutableList.of()` not `Collections.emptyList()`.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -1217,12 +1219,21 @@ protected void validateNamespace(final SqlValidatorNamespace namespace,
...
         final ImmutableBitSet mustFilterFields =
             namespace.getMustFilterFields();
-        if (!mustFilterFields.isEmpty()) {
+        // Remnant must filter fields are fields that are not selected and cannot
+        // be defused unlessa bypass field defuses it.
+        // A top-level namespace must not have any remnant-must-filter fields.
+        final Set<SqlQualified> remnantMustFilterFields = namespace.getRemnantMustFilterFields();
+        if (!mustFilterFields.isEmpty() || !remnantMustFilterFields.isEmpty()) {
+          Stream<String> mustFilterStream =
+                  StreamSupport.stream(mustFilterFields.spliterator(), false)
```
julianhyde, line 1228: should be indent 4 not indent 8

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -1217,12 +1219,21 @@ protected void validateNamespace(final SqlValidatorNamespace namespace,
...
+          Stream<String> remnantStream =
+                  StreamSupport.stream(remnantMustFilterFields.spliterator(), false)
+                      .map(q -> q.suffix().get(0));
+
           // Set of field names, sorted alphabetically for determinism.
-          Set<String> fieldNameSet =
-              StreamSupport.stream(mustFilterFields.spliterator(), false)
-                  .map(namespace.getRowType().getFieldNames()::get)
-                  .collect(Collectors.toCollection(TreeSet::new));
+          Set<String> fieldNameSet = Stream.concat(mustFilterStream, remnantStream)
```
julianhyde, line 1235: add linefeed after `=`

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -1217,12 +1219,21 @@ protected void validateNamespace(final SqlValidatorNamespace namespace,
         // fields; these are neutralized if the consuming query filters on them.
         final ImmutableBitSet mustFilterFields =
             namespace.getMustFilterFields();
-        if (!mustFilterFields.isEmpty()) {
+        // Remnant must filter fields are fields that are not selected and cannot
```
julianhyde, line 1222: Not sure whether the 'remnant' concept is necessary. Instead could we just remove must-filter fields that have been satisfied?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4185,6 +4217,48 @@ private static void forEachQualified(SqlNode node, SqlValidatorScope scope,
...
+      SelectScope fromScope, ScopeChild child, List<String> fieldNames) {
+    fields.forEachInt(i ->
+        qualifiedSet.add(
+            SqlQualified.create(fromScope, 1, child.namespace,
+                new SqlIdentifier(
+                    ImmutableList.of(child.name, fieldNames.get(i)),
+                    SqlParserPos.ZERO))));
+
+  }
+  private static boolean qualifiedMatchesIdentifier(SqlQualified q1, SqlQualified q2) {
```
julianhyde, line 4258: add blank line before

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
julianhyde: No, it should be a javadoc comment. Also it should be written like a method javadoc, e.g. "Removes all entries..."

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -12379,6 +12379,370 @@ private void checkCustomColumnResolving(String table) {
...
+    // theoretically be filtered by an enclosing query.
+    fixture.withSql("^select empno\n"
+            + "from emp^")
+        .fails(missingFilters("JOB"));
+    fixture.withSql("^select empno,\n"
+            + "  sum(sal) over (order by mgr)\n"
+            + "from emp^")
+        .fails(missingFilters("JOB"));
+  }
+
```
julianhyde, line 12414: remove blank line

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -12379,6 +12379,370 @@ private void checkCustomColumnResolving(String table) {
...
+        .ok();
+
+    // Deceitful alias #1. Filter on 'j' is a filter on the underlying 'job'.
+    fixture.withSql("select * from (\n"
+            + "  select job as j, ename as job\n"
+            + "  from emp\n"
+            + "  where empno = 1)\n"
+            + "where j = 'doctor'")
+        .ok();
+      // Deceitful alias #2. Filter on 'job' is a filter on the underlying
```
julianhyde, line 12474: indent

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
julianhyde: > having the comments below the cases is a little confusing

Agree. Comments after cases is a neat idea (because it puts the comment close to the error message), but I think it will confuse people, and the comments will become detached from the test cases as the code is maintained. I think you should move the comments above the test case (i.e. 7 lines earlier in this case).

Expand 'There are two JOB columns but only one is filtered' to 'Query is invalid because there are two JOB columns but only one is filtered'. It is a little redundant but it ties the 'what' to the 'why'.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
olivrlee: I think it is necessary to handle the case where the bypass field is selected, but the must-filter fields are not selected. 

Imagine this setup
```
Table: EMP (empno, name, id)
must_filters: empno
bypass_filter: name
```
Query:
`select * from (select name from emp) where name = 'bob' `

The inner query `select name from emp` can't error on `empno` not being filtered at this point because `name` was selected, and could be filtered on later. The must-filter for `empno` is not satisfied yet at this point. 

When we get to the `select * from _ where name = 'bob'`, that's where we defuse the error for `empno` based on seeing `name` being filtered on. 


If alternatively the query was 
`select * from (select name, id from emp) where id = '12' `
When we get to `select * from _ where id = '12'`, we're at the top-level, and remnantMustFilter's set is not empty, hence we throw an error.

julianhyde: I see. If `t` has a must-filter field `f` and bypass fields `b0` and `b1`, then the query `select b0, b1 from t` has no must-filter fields, no bypass fields, but two remnant fields `[b0, b1]`.

If you filter on any of those remnant fields then they all go away.

core/src/main/java/org/apache/calcite/sql/validate/MustFilterRequirements.java
```diff
@@ -37,9 +35,9 @@ public class MustFilterRequirements {
 
   ImmutableBitSet mustFilterFields;
   ImmutableBitSet mustFilterBypassFields;
-  protected Set<SqlQualified> remnantMustFilterFields;
+  protected ImmutableSet<SqlQualified> remnantMustFilterFields;
   public MustFilterRequirements(ImmutableBitSet mustFilterFields,
-      ImmutableBitSet mustFilterBypassFields, Set<SqlQualified> remnantMustFilterFields) {
+      ImmutableBitSet mustFilterBypassFields, ImmutableSet<SqlQualified> remnantMustFilterFields) {
```
julianhyde, line 42: Parameters and fields should be declared as Set not ImmutableSet

(I don't like the rule, but it minimizes binary compatibility problems for people who have shaded Guava.)

You will often see the following pattern in constructors:
```
  this.remnantMustFilterFields = ImmutableSet.copyOf(remnantMustFilterFields);
```
which ensures immutability and non-null.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4710,8 +4710,9 @@ protected void validateMustFilterRequirements(SqlSelect select, SelectNamespace
         }
         // Remaining must-filter fields can be defused by a bypass-field,
         // so we pass this to the consumer.
-        Set<SqlQualified> remnantMustFilterFields = Stream.of(remnantQualifieds, qualifieds)
-            .flatMap(Set::stream).collect(Collectors.toSet());
+        ImmutableSet<SqlQualified> remnantMustFilterFields =
+            Stream.of(remnantQualifieds, qualifieds)
+            .flatMap(Set::stream).collect(ImmutableSet.toImmutableSet());
```
julianhyde, line 4715: indent + 4

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4710,8 +4710,9 @@ protected void validateMustFilterRequirements(SqlSelect select, SelectNamespace
...
         ns.mustFilterRequirements =
             new MustFilterRequirements(ImmutableBitSet.fromBitSet(mustFilterFields),
             ImmutableBitSet.fromBitSet(mustFilterBypassFields),
```
julianhyde, line 4718: indent + 4

core/src/main/java/org/apache/calcite/sql/validate/MustFilterRequirements.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * <p>mustFilterFields: the ordinals (in the row type) of the "must-filter" fields,
+ * fields that must be filtered in a query.
+ * <p> mustFilterBypassFields: the ordinals (in the row type) of the "bypass" fields,
+ * fields that can defuse validation errors on mustFilterFields if filtered on.
+ * <p> remnantMustFilterFields: Set of mustFilterField SqlQualifieds that have not been defused
+ * in the current query, but can still be defused by filtering on a bypass field in the
+ * enclosing query.
+ */
+public class MustFilterRequirements {
```
julianhyde, line 36: Must the class be `public`?

Class names should not be plural (except utility classes). It becomes problematic to name collections of such classes.

'Must' and 'Requirement' seem redundant. So I suggest 'FilterRequirement'.

Maybe `mustFilterFields` &rarr; `filterFields`, `mustFilterBypassFields` &rarr; `bypassFields`; `remnantMustFilterFields` &rarr; `filterAnyFields`.

Could the fields be `final`? (Public final fields are fine, but if you must have non-final fields they should probably be `private`.)

Immutable classes are much easier to reason about. Consider adding 'withXxx' methods.

core/src/main/java/org/apache/calcite/sql/validate/TableNamespace.java
```diff
@@ -61,22 +61,19 @@ private TableNamespace(SqlValidatorImpl validator, SqlValidatorTable table,
...
-            this.remnantMustFilterFields = new HashSet<>(semanticTable.remnantMustFilterFields()));
-
+        .ifPresent(semanticTable -> {
+          ImmutableBitSet mustFilterFields = table.getRowType().getFieldList().stream()
+              .map(RelDataTypeField::getIndex)
+              .filter(semanticTable::mustFilter)
+              .collect(toImmutableBitSet());
+          ImmutableBitSet bypassFieldList = semanticTable.bypassFieldList().stream()
+              .collect(toImmutableBitSet());
+          // We pass in an empty set for remnantMustFilterFields here because it isn't exposed to
```
julianhyde, line 72:
```
semanticTable.bypassFieldList().stream()
              .collect(toImmutableBitSet())
```
could be `semanticTable.bypassFieldList()` or `ImmutableBitSet.of(semanticTable.bypassFieldList())`

julianhyde:
Latest commit looks fairly close.

There needs to be an cogent explanation in the code for the concepts of bypass fields etc. The javadoc for `class FilterRequirement` seems like a good place. Add a couple of worked examples, like my 'table t has must-filter field f and bypass field b and the query "select ... from t" has ...'.

Move the current descriptions of the 3 fields onto the fields themselves.

'Remnant' is not a good name because it forces you to talk about how we got here - the winnowing algorithm - rather than there we are. I prefer 'anyFitlerFields' because for the query to be valid you need to filter any of them.

I think you can make `FilterRequirement` immutable if you have a subclass of `SqlVisitor` that has 3 mutable fields.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
olivrlee: Actually that is not how the sets are populated.
The purpose of remnant fields is to capture the fields that were marked as required, but no longer accessible to be filtered on and defused in the select clause. In other words, you cannot filter on any of the remnant fields. The presence of values being in the remant-filter set serves the purpose of being an indicator for erroring. 

They can only be defused if any of the bypass fields are filtered on.

In your example this would be the correct set values:
`select b0, b1 from t`

must-filter field: []
bypass-fields: [`b0`, `b1`]
remnant-fields: [f]

core/src/main/java/org/apache/calcite/sql/validate/MustFilterRequirements.java
olivrlee: Does this apply to the first two parameters as well?
I see I could use `ImmutableBitSet.fromBitSet()`

core/src/main/java/org/apache/calcite/sql/validate/MustFilterRequirements.java
olivrlee: It could be package-private.
I'll rename it to `FilterRequirement` and `filterFields`, `bypassFields`. 

Still thinking about the rename of `remnantMustFilterFields` ... I still think the name needs to capture the connotation that they're valid filter requirements left-over from a previous query, but also distinct from the set that can actively be defused

core/src/main/java/org/apache/calcite/sql/validate/MustFilterRequirements.java
olivrlee: answer: No because it is not guava

olivrlee:
Still requires a review before final merge, but last commit was fairly close.

PR #3986, opened 2024-09-29
[CALCITE-6604] Add support for SqlWindowTableFunction in RelToSql Con…

ehds:
see https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6604

Add WindowTableFunctionScanContext to convert RexCall(SqlWindowFunction) to SqlNdoe.

1. First convert SqlWindowFunction to SqlCall without PARAM_DATA. 
2. Finally reconstruct a new SqlCall with PARAM_DATA as the first operand.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
@@ -81,6 +81,7 @@ import org.apache.calcite.sql.SqlSelect;
 import org.apache.calcite.sql.SqlTableRef;
 import org.apache.calcite.sql.SqlUpdate;
 import org.apache.calcite.sql.SqlUtil;
+import org.apache.calcite.sql.SqlWindowTableFunction;
 import org.apache.calcite.sql.fun.SqlInternalOperators;
 import org.apache.calcite.sql.fun.SqlMinMaxAggFunction;
 import org.apache.calcite.sql.fun.SqlSingleValueAggFunction;
@@ -116,6 +117,8 @@ import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+
+import static com.google.common.base.Preconditions.checkArgument;
 
 import static org.apache.calcite.rex.RexLiteral.stringValue;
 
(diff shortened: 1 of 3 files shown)


mihaibudiu (approved):

mihaibudiu:
I assume we are in a merging freeze until we release 1.38, so I will postpone merging this

mihaibudiu:
The CI is not passing anyway

ehds:
> The CI is not passing anyway

Fixed.

mihaibudiu:
You need to provide at least one test that shows that the bug is fixed.
The example you gave in the issue would be fine.
You can probably add that in the file RelToSqlCovnerterTest

mihaibudiu:
Ignore my comment, I think I was looking at the last commit by mistake.
There is a test.

mihaibudiu (approved):

mihaibudiu:
Will merge this after 1.38 is released.

mihaibudiu:
Can you please squash the commits so we can merge?

ehds:
> Can you please squash the commits so we can merge?

done.

core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java
```diff
@@ -1286,13 +1289,31 @@ public Result visit(Uncollect e) {
 
   public Result visit(TableFunctionScan e) {
     final List<SqlNode> inputSqlNodes = new ArrayList<>();
+    final List<SqlNode> fieldNodes = new ArrayList<>();
+
+    final int filedCount = e.getRowType().getFieldCount();
```
mihaibudiu, line 1294: sorry, didn't notice this before, but I think this should be "fieldCount" instead of "filedCount"

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1719,6 +1725,46 @@ class TableFunctionScanContext extends BaseContext {
...
+
+    @Override public SqlNode toSql(@Nullable RexProgram program, RexNode rex) {
+      if (isWindowTableFunctionRex(rex)) {
+        // Convert SqlWindowTableFunction operator without the PARAM_DATA to sqlNode.
+        SqlNode callNode = super.toSql(null, rex);
+        // Reconstruct the callNode with inputTableNode as the PARAM_DATA.
+        List<SqlNode> operandList = new ArrayList<>();
+        operandList.add(inputTableNode);
+        operandList.addAll(((SqlBasicCall) callNode).getOperandList());
+        callNode = new SqlBasicCall(((SqlBasicCall) callNode).getOperator(), operandList, POS);
```
mihaibudiu, line 1761: Recently we added source position information to RexCall nodes, you should take the source position information from the call.

mihaibudiu (changes requested):

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
ehds: Thanks for the suggestion, done.

mihaibudiu (approved):

PR #3987, opened 2024-09-30
[CALCITE-6605] Lattice SQL supports complex column expressions

YiwenWu:
https://issues.apache.org/jira/browse/CALCITE-6605

### DIFF
--- a/core/src/main/java/org/apache/calcite/materialize/Lattice.java
+++ b/core/src/main/java/org/apache/calcite/materialize/Lattice.java
@@ -39,6 +39,7 @@ import org.apache.calcite.schema.impl.MaterializedViewTable;
 import org.apache.calcite.schema.impl.StarTable;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.SqlIdentifier;
 import org.apache.calcite.sql.SqlJoin;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlNode;
@@ -88,6 +89,7 @@ import java.util.stream.Collectors;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import static org.apache.calcite.linq4j.Nullness.castNonNull;
+import static org.apache.calcite.rel.rel2sql.SqlImplementor.POS;
 
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/materialize/LatticeSuggesterTest.java
```diff
@@ -752,6 +756,93 @@ private void checkDerivedColumn(Lattice lattice, List<String> tables,
     checkDerivedColumn(lattice, tables, derivedColumns, 3, "n11", false);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6605">[CALCITE-6605]
+   * Lattice SQL supports complex column expressions </a>. */
+  @Test void testExpressionLatticeSql() throws Exception {
```
caicancai, line 762: Maybe you can add some blank lines or a method to make the test sample code more readable.🤔

core/src/main/java/org/apache/calcite/materialize/Lattice.java
```diff
@@ -370,6 +370,23 @@ public String sql(ImmutableBitSet groupSet, boolean group,
...
+    fieldFuncRef[0] = f -> {
+      Column column = columns.get(f);
+      if (column instanceof BaseColumn) {
+        return new SqlIdentifier(ImmutableList.of(((BaseColumn) column).column), POS);
+      }
+      if (column instanceof DerivedColumn) {
+        return new SqlImplementor.SimpleContext(dialect, fieldFuncRef[0])
+            .toSql(null, ((DerivedColumn) column).e);
+      }
+      throw new UnsupportedOperationException();
```
mihaibudiu, line 385: This probably should have some more descriptive error

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/materialize/Lattice.java
caicancai: +1

core/src/test/java/org/apache/calcite/materialize/LatticeSuggesterTest.java
YiwenWu: Added description for tests.

core/src/main/java/org/apache/calcite/materialize/Lattice.java
YiwenWu: It will be called when lattice column generates sql.
```
final Column column = columns.get(i);
column.toSql(w);
```

Originally, for `DerivedColumn`, the `UnsupportedOperationException` will be thrown directly. This PR will obtain the corresponding column based on index and return the field object. 

Currently there are only two types of columns: `BaseColumn`, `DerivedColumn`, theoretically no exception will be thrown. But in order to avoid extreme cases that are not covered, the logic of throwing exceptions before is also retained.

caicancai (approved):

PR #3991, opened 2024-10-03
[CALCITE-6608] RexBuilder#makeIn should create EQUALS instead of SEARCH for single point values

zabetak:


NobiGo (approved):

rubenada (approved):

core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
```diff
@@ -957,6 +957,16 @@ private void checkDate(RexLiteral literal) {
     assertThat(inCall.getKind(), is(SqlKind.SEARCH));
   }
 
+  @Test void testMakeInSingleValue() {
```
caicancai, line 960: Add jira link

caicancai (approved):

core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
zabetak: The JIRA link doesn't add much info. It is clear why the test is added from the git history.

julianhyde: Git history can get lost if the code is significantly altered.

The point of this test is to check that if `makeIn` has a single argument the result is not a sarg, and that is not explicit.

It's a seven line test but which is the key line that demonstrates the bug? Is it that we used `SqlTypeName.INTEGER` rather than `SqlTypeName.SMALLINT`? As test author you know that six of those lines are boilerplate, and that before the bug (actually feature request) was fixed the string would have been "SEARCH($0, Sarg([100, 100]))" and that string is still in some sense correct but "=($0, 100)" is in some sense better.

We don't necessarily need a jira link but we do need more on the page (probably javadoc, or possibly a message in the `assertThat`). 

I am a huge fan of using javadoc to state the almost obvious - what is obvious to the author but not obvious to the reader.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1750,6 +1750,9 @@ public RexNode makeNullLiteral(SqlTypeName typeName) {
    * literal, "SEARCH(arg, SARG([point0..point0], [point1..point1], ...)";
    * otherwise creates a disjunction, "arg = point0 OR arg = point1 OR ...". */
   public RexNode makeIn(RexNode arg, List<? extends RexNode> ranges) {
```
julianhyde, line 1752: This javadoc needs to be fixed to match the behavior of the code.

core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
```diff
@@ -957,6 +957,16 @@ private void checkDate(RexLiteral literal) {
...
+    RelDataTypeFactory typeFactory = new SqlTypeFactoryImpl(RelDataTypeSystem.DEFAULT);
+    RexBuilder rexBuilder = new RexBuilder(typeFactory);
+    RelDataType intType = typeFactory.createSqlType(SqlTypeName.INTEGER);
+    RexNode column = rexBuilder.makeInputRef(intType, 0);
+    RexLiteral literal = rexBuilder.makeLiteral(100, intType);
+    RexNode inCall = rexBuilder.makeIn(column, ImmutableList.of(literal));
```
julianhyde, line 966: Note that `makeIn(column, ImmutableList.of(literal, literal))` currently returns a singleton Sarg but could be optimized.

core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
zabetak: I tried to make the purpose of the test more explicit in https://github.com/apache/calcite/pull/3991/commits/d6ade5afaf85254e14ff1e60121a87c8e6feb224. Please check.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
zabetak: Added javadoc in https://github.com/apache/calcite/pull/3991/commits/32f0485ff2225f48d312566930bb12df384ec7f1.

core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
zabetak: Modified the current approach in https://github.com/apache/calcite/pull/3991/commits/32f0485ff2225f48d312566930bb12df384ec7f1 to account for duplicates. 

I also added more tests for this purpose in https://github.com/apache/calcite/pull/3991/commits/a4dbffdd04ae802ae53687de54a9e01e762af640.

zabetak:
If there are no more comments/objections I will merge this PR in the next 48 hours.

mihaibudiu:
don't forget to squash the commits

PR #3994, opened 2024-10-06
[CALCITE-6586] Some Rules not firing due to RelMdPredicates returning null in VolcanoPlanner

suibianwanwank:
The fundamental problem this PR solves is to make RelMetadataQuery#getPulledUpPredicates(RelSubset) not always return EMPTY, allowing some rules to be applied in VolcanoPlanner.
However, after some rules were applied, there were some other issues that prevented the test from passing completely. So I split it into three commits for easy review

### DIFF
--- a/core/src/main/java/org/apache/calcite/interpreter/Bindables.java
+++ b/core/src/main/java/org/apache/calcite/interpreter/Bindables.java
@@ -269,6 +269,11 @@ public class Bindables {
           .itemIf("projects", projects, !projects.equals(identity()));
     }
 
+    @Override public double estimateRowCount(RelMetadataQuery mq) {
+      double f = filters.isEmpty() ? 1d : 0.5d;
+      return super.estimateRowCount(mq) * f;
+    }
+
     @Override public @Nullable RelOptCost computeSelfCost(RelOptPlanner planner,
         RelMetadataQuery mq) {
       boolean noPushing = filters.isEmpty()
--- a/core/src/main/java/org/apache/calcite/util/Bug.java
+++ b/core/src/main/java/org/apache/calcite/util/Bug.java
@@ -220,6 +220,12 @@ public abstract class Bug {
    * is fixed. */
   public static final boolean CALCITE_6328_FIXED = false;
 
+  /** Whether
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6611">[CALCITE-6611]
+   * Rules that modify the sort collation cannot be applied in VolcanoPlanner</a>
+   * is fixed. */
+  public static final boolean CALCITE_6611_FIXED = false;
+
   /**
    * Use this to flag temporary code.
    */
--- a/core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
@@ -160,13 +160,18 @@ class JdbcAdapterTest {
             + "where \"product_id\" = 1")
         .runs()
         .enable(CalciteAssert.DB == CalciteAssert.DatabaseInstance.HSQLDB)
-        .planHasSql("SELECT *\n"
+        .planHasSql("SELECT 1 AS \"product_id\", \"time_id\", \"customer_id\", "
+            + "\"promotion_id\", \"store_id\", \"store_sales\", "
+            + "\"store_cost\", \"unit_sales\"\n"
+            + "FROM (SELECT \"time_id\", \"customer_id\", \"promotion_id\", \"store_id\", "
+            + "\"store_sales\", \"store_cost\", \"unit_sales\"\n"
             + "FROM \"foodmart\".\"sales_fact_1997\"\n"
             + "WHERE \"product_id\" = 1\n"
             + "UNION ALL\n"
-            + "SELECT *\n"
+            + "SELECT \"time_id\", \"customer_id\", \"promotion_id\", \"store_id\", "
+            + "\"store_sales\", \"store_cost\", \"unit_sales\"\n"
             + "FROM \"foodmart\".\"sales_fact_1998\"\n"
-            + "WHERE \"product_id\" = 1");
+            + "WHERE \"product_id\" = 1) AS \"t3\"");
   }
 
   @Test void testInPlan() {
--- a/core/src/test/java/org/apache/calcite/test/ScannableTableTest.java
+++ b/core/src/test/java/org/apache/calcite/test/ScannableTableTest.java
@@ -278,9 +278,10 @@ public class ScannableTableTest {
     final Table table = new BeatlesProjectableFilterableTable(buf, false);
     final String explain = "PLAN="
         + "EnumerableAggregate(group=[{0}], C=[COUNT()])\n"
-        + "  EnumerableAggregate(group=[{0, 1}])\n"
-        + "    EnumerableInterpreter\n"
-        + "      BindableTableScan(table=[[s, beatles]], filters=[[=($2, 1940)]], projects=[[2, 0]])";
+        + "  EnumerableCalc(expr#0=[{inputs}], expr#1=[1940], k=[$t1], i=[$t0])\n"
+        + "    EnumerableAggregate(group=[{1}])\n"
+        + "      EnumerableInterpreter\n"
+        + "        BindableTableScan(table=[[s, beatles]], filters=[[=($2, 1940)]], projects=[[2, 0]])";
     CalciteAssert.that()
         .with(newSchema("s", PairList.of("beatles", table)))
         .query(sql)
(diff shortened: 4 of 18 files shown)


core/src/main/java/org/apache/calcite/interpreter/Bindables.java
```diff
@@ -269,6 +269,12 @@ public static BindableTableScan create(RelOptCluster cluster,
           .itemIf("projects", projects, !projects.equals(identity()));
     }
 
+    @Override public double estimateRowCount(RelMetadataQuery mq) {
```
suibianwanwank, line 272: The push-down fliter is not in the row count calculation, resulting in inaccurate costs

innodb/src/main/java/org/apache/calcite/adapter/innodb/InnodbRules.java
```diff
@@ -184,10 +184,13 @@ RelNode convert(LogicalFilter filter, InnodbTableScan scan) {
       final IndexCondition indexCondition =
           translator.translateMatch(filter.getCondition());
 
+      RexNode condition =
+          RexUtil.composeConjunction(cluster.getRexBuilder(),
+              indexCondition.getPushDownConditions());
```
suibianwanwank, line 189: In InnodbFilter, condition contains both indexed and non-indexed conditions, but the actual execution only executes the indexed condition, which causes some rules to incorrectly optimize

suibianwanwank:
Some Druid tests have failed, and I will work on fixing them. It may take some time to run these tests locally, as I haven't done that before.

mihaibudiu:
For druid see git@github.com:zabetak/calcite-druid-dataset.git

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -160,13 +160,18 @@ class JdbcAdapterTest {
             + "where \"product_id\" = 1")
         .runs()
         .enable(CalciteAssert.DB == CalciteAssert.DatabaseInstance.HSQLDB)
-        .planHasSql("SELECT *\n"
+        .planHasSql("SELECT 1 AS \"product_id\", \"time_id\", \"customer_id\", "
```
mihaibudiu, line 163: this plan seems worse, is this expected?

mihaibudiu (approved):
I can't really tell whether the plans correspond to your expectations after these changes.
Otherwise the changes look fine.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
suibianwanwank: Before Plan:
```
JdbcToEnumerableConverter
  JdbcUnion(all=[true])
    JdbcFilter(condition=[=($0, 1)])
      JdbcTableScan(table=[[foodmart, sales_fact_1997]])
    JdbcFilter(condition=[=($0, 1)])
      JdbcTableScan(table=[[foodmart, sales_fact_1998]])
```
After Plan:
```
JdbcToEnumerableConverter
  JdbcProject(product_id=[1], time_id=[$0], customer_id=[$1], promotion_id=[$2], store_id=[$3], store_sales=[$4], store_cost=[$5], unit_sales=[$6])
    JdbcUnion(all=[true])
      JdbcProject(time_id=[$1], customer_id=[$2], promotion_id=[$3], store_id=[$4], store_sales=[$5], store_cost=[$6], unit_sales=[$7])
        JdbcFilter(condition=[=($0, 1)])
          JdbcTableScan(table=[[foodmart, sales_fact_1997]])
      JdbcProject(time_id=[$1], customer_id=[$2], promotion_id=[$3], store_id=[$4], store_sales=[$5], store_cost=[$6], unit_sales=[$7])
        JdbcFilter(condition=[=($0, 1)])
          JdbcTableScan(table=[[foodmart, sales_fact_1998]])
```
In terms of SQL length the plan did get worse. This is because UnionPullUpConstantRule optimizes a RexInputRef to be a constant(Because of the predicate _where product_id = 1_), so what was originally 8 fields for Union's input is now only 7.

mihaibudiu:
can you please fix the conflicts and squash the commits?

suibianwanwank:
> can you please fix the conflicts and squash the commits?

I will resolve these conflicts today. Also, do you think it would be easier to understand if we keep them as separate commit? It might make it clearer.

mihaibudiu:
Having separate commits if they represent independent logical changes to the code.
I think a good rule which tells you whether something could be broken into separate commits is: when you bisect the commits, the code should build and run fine between the commits. Otherwise they should not be separate.

mihaibudiu:
Are you trying to close 4 issues with this PR?
Ideally it should have been 4 separate PRs, especially if they are self-contained.

mihaibudiu:
If no one objects, I will merge this as 4 commits that close 4 issues, but it is easy to miss the fact that the commits address separate issues - it's not the typical workflow. Keeping JIRA and github in sync is a non-trivial amount of work for maintainers, so it's good if we can make the task simpler.

suibianwanwank:
> If no one objects, I will merge this as 4 commits that close 4 issues, but it is easy to miss the fact that the commits address separate issues - it's not the typical workflow. Keeping JIRA and github in sync is a non-trivial amount of work for maintainers, so it's good if we can make the task simpler.

I apologize for the inconvenience caused. I will follow the correct workflow afterward. And I have compiled and tested each commit individually. 
Also, [CALCITE-6611] is only a temporary solution, so it shouldn't be closed yet.

mihaibudiu:
So what should we do about the corresponding commit for 6611? Someone who reads the commit log may infer that the commits solves the problem. Maybe you can remove the issue number from the commit?

mihaibudiu:
To make sure: this PR is closing the following 3 issues: 6615, 6614, and 6586.
Is that correct?

suibianwanwank:
> To make sure: this PR is closing the following 3 issues: 6615, 6614, and 6586. Is that correct?

Yes, that's correct! Also, I tried to resolve the conflict, but this caused some problems regarding pullUpPredicate being slow with too many values. I'll try to look into this.

mihaibudiu:
you want to do this in the current PR? maybe I should just merge it

suibianwanwank:
> you want to do this in the current PR? maybe I should just merge it

Thank you for your interest in this PR. But I don't think it could MERGE yet, as I don't have a good way of passing this test. I'll create a JIRA ISSUE to report this issue and then we'll evaluate what to do. WDYT

mihaibudiu:
certainly, CI has to pass first

mihaibudiu:
I have converted this to draft, please restore it when it's ready for review

core/src/main/java/org/apache/calcite/util/Bug.java
```diff
@@ -220,6 +220,12 @@ public abstract class Bug {
    * is fixed. */
   public static final boolean CALCITE_6328_FIXED = false;
 
+  /** Whether
```
mihaibudiu, line 223: The title of the commit that introduces this change is misleading: you are using the convention used for a commit which *solves* an issue, but this commit only works around the issue. So I think before merging the title of the commit should be changed.

core/src/test/java/org/apache/calcite/test/ScannableTableTest.java
```diff
@@ -278,9 +278,10 @@ public class ScannableTableTest {
     final Table table = new BeatlesProjectableFilterableTable(buf, false);
     final String explain = "PLAN="
         + "EnumerableAggregate(group=[{0}], C=[COUNT()])\n"
-        + "  EnumerableAggregate(group=[{0, 1}])\n"
-        + "    EnumerableInterpreter\n"
-        + "      BindableTableScan(table=[[s, beatles]], filters=[[=($2, 1940)]], projects=[[2, 0]])";
+        + "  EnumerableCalc(expr#0=[{inputs}], expr#1=[1940], k=[$t1], i=[$t0])\n"
```
mihaibudiu, line 281: I personally find it very difficult to evaluate whether these plans are correct by inspection (including the ones in XML files). Or even if they are better in any way.
On the other hand this PR is all about plan quality.
So I will have to trust you that the plans are better than they used to be before.

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/util/Bug.java
suibianwanwank: Do you have any suggestions for a commit message, I think it might be good to include this in [CALCITE-6586] commit.

mihaibudiu: I have no problem with 6586, just with 6328

mihaibudiu: The commit message can be "Introduced bug guard for CALCITE-6328."

suibianwanwank: Done.

mihaibudiu:
If no one else has comments I plan to merge this as 4 commits, which close 3 issues.

PR #3997, opened 2024-10-07
Ddl extend

zhaoxiangyublzcat:
Expand the DDL executor, implement MongoDdlExecutor, and add the auto-service dependency. 

We have added the MongoDdlExecutor class, which is registered as an DdlExecutor service via the AutoService annotation. The service loading logic is implemented in the CalcitePrepareImpl class, and the auto-service dependency is added to the Gradle configuration.


caicancai:
@zhaoxiangyublzcat Can you provide a jira case to describe it?

PR #3998, opened 2024-10-07
[CALCITE-6617] Regression after CALCITE-2067 which gives a ClassCastException

mihaibudiu:
This fixes the bug I could reproduce from https://issues.apache.org/jira/browse/CALCITE-6617
I have created a test case for the bug I could not reproduce, to show that it works


mihaibudiu:
@arkanovicz can you please check if this solves the issue you have reported?

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -292,16 +292,17 @@ protected boolean binaryArithmeticWithStrings(
     }
 
     RelDataType commonType;
-    if (SqlTypeUtil.sameNamedType(type1, type2)) {
+    // First check when one of the types is a string (see JavaDoc on cmmonTypeForBinaryComparison)
+    commonType = commonTypeForBinaryComparison(type1, type2);
```
mihaibudiu, line 296: previous to this change, coercion would not insert casts when comparing say int to double, because they didn't have the same type name. that was a bug.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -292,16 +292,17 @@ protected boolean binaryArithmeticWithStrings(
...
+    if (commonType == null && SqlTypeUtil.isComparable(type1, type2)) {
```
caicancai, line 297: Why do we need to judge this `SqlTypeUtil.isComparable(type1, type2)`?

caicancai (approved):
overall LGTM

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
caicancai: agree

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
mihaibudiu: This is about comparison operators, you'd think that if 2 values with types are comparable, then the values can be compared by being cast to a common type.

NobiGo (commented):
Strangely, we need to change the numbers written in SQL, which may cause compatibility issues with previous versions.

rubenada (approved):

core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
```diff
@@ -253,17 +254,17 @@ private static ImmutableList<RelDataType> combine(
 
     // intersect
     sql("select t1_int, t1_decimal, t1_smallint, t1_double from t1 "
-        + "intersect select t2_varchar20, t2_decimal, t2_float, t2_bigint from t2 ")
```
NobiGo, line 256: Why did we change all `float` type to `real` type?

mihaibudiu: Because float and double are actually the same type, so the tests have better coverage now.
https://issues.apache.org/jira/browse/CALCITE-6074

NobiGo (approved):

mihaibudiu:
> Strangely, we need to change the numbers written in SQL, which may cause compatibility issues with previous versions.

You don't need to modify any SQL programs, but the unparsed programs may look different.

PR #4000, opened 2024-10-08
[CALCITE-6620] VALUES created by RelBuilder do not have a homogeneous type

mihaibudiu:
The issue fixed here was first reported as part of https://issues.apache.org/jira/browse/CALCITE-6617

### DIFF
--- a/core/src/main/java/org/apache/calcite/tools/RelBuilder.java
+++ b/core/src/main/java/org/apache/calcite/tools/RelBuilder.java
@@ -2886,6 +2886,31 @@ public class RelBuilder {
     return ((AggCallPlus) c).op().kind == SqlKind.GROUP_ID;
   }
 
+  /** Given a list of literals and a target row type, make the literals
+   * respectively match the fields types of the row.
+   *
+   * @param rowType  Type expected for values
+   * @param values   A list of literals that should match the rowType */
+  private List<RexLiteral> convertLiteralTypes(
+      RelDataType rowType, List<RexLiteral> values) {
+    assert values.size() == rowType.getFieldCount()
+        : "List of literals of size " + values.size() + " does not match expected type " + rowType;
+
+    List<RelDataTypeField> fields = rowType.getFieldList();
+    List<RexLiteral> constants = new ArrayList<>();
+
+    for (int i = 0; i < values.size(); i++) {
+      RexLiteral vi = values.get(i);
+      RelDataType type = fields.get(i).getType();
+      RexNode e = cluster.getRexBuilder().makeAbstractCast(type, vi, false);
+      RexNode simplified = simplifier.simplify(e);
+      assert simplified instanceof RexLiteral
+          : "Could not simplify expression to literal" + simplified;
+      constants.add((RexLiteral) simplified);
+    }
+    return ImmutableList.copyOf(constants);
+  }
+
   private RelBuilder setOp(boolean all, SqlKind kind, int n) {
     List<RelNode> inputs = new ArrayList<>();
     for (int i = 0; i < n; i++) {
@@ -2917,7 +2942,11 @@ public class RelBuilder {
       requireNonNull(rowType, () -> "leastRestrictive(" + inputTypes + ")");
       final List<List<RexLiteral>> tuples = new ArrayList<>();
       for (RelNode input : inputs) {
-        tuples.addAll(((Values) input).tuples);
+        ImmutableList<ImmutableList<RexLiteral>> literals = ((Values) input).tuples;
+        for (ImmutableList<RexLiteral> l : literals) {
+          List<RexLiteral> converted = convertLiteralTypes(rowType, l);
+          tuples.add(converted);
+        }
       }
       final List<List<RexLiteral>> tuples2 =
           all ? tuples : Util.distinctList(tuples);
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -7765,7 +7765,7 @@ class RelToSqlConverterTest {
     final String expectedDefault = "INSERT INTO \"SCOTT\".\"DEPT\""
         + " (\"DEPTNO\", \"DNAME\", \"LOC\")\n"
         + "VALUES (1, 'Fred', 'San Francisco'),\n"
-        + "(2, 'Eric', 'Washington')";
+        + "(2, 'Eric', 'Washington   ')";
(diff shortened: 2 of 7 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -10754,6 +10757,17 @@ private void testCurrentDateFunc(Pair<String, Hook.Closeable> pair) {
     f0.forEachLibrary(libraries, consumer);
   }
 
+  @Test void differentTypeValues() {
```
julianhyde, line 10760: Does this test belong here? It isn't testing an operator, or use the fixture.

Could it be a Quidem test?

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -2886,6 +2886,36 @@ private static boolean isGroupId(AggCall c) {
...
+   * respectively match the fields types of the row.
+   *
+   * @param cluster  Cluster
+   * @param rowType  Type expected for values
+   * @param values   A list of literals that should match the rowType */
+  private static List<RexLiteral> convertLiteralTypes(RelOptCluster cluster,
+      RelDataType rowType, List<RexLiteral> values) {
+    assert values.size() == rowType.getFieldCount();
+    RexExecutor executor = cluster.getPlanner().getExecutor();
+    if (executor == null) {
```
NobiGo, line 2899: Do we have a test for the `executor == null`？I'm not sure why this happened.

mihaibudiu: Who knows who is calling the builder and with what setup?
This is just to be safe.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: I will move it.
I don't know if I can create the same environment in Quidem. A simple attempt to reproduce this bug with standard settings failed.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
suibianwanwank: +1, getExecutor is a nullable method.

mihaibudiu:
Several quidem tests now have extra spaces in strings because they use a type system without raggedUnionTypesToVarying. The extra spaces are introduced when unifying the types of the VALUES. The new results seem the right ones.

suibianwanwank (approved):

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
NobiGo: Maybe we can improve it by `RexExecutor rexImpl =
        Util.first(cluster.getPlanner().getExecutor(), RexUtil.EXECUTOR);`

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -9174,6 +9174,17 @@ LogicalProject(DEPTNO=[$0], EXPR$1=[$1])
   LogicalSort(sort0=[$2], dir0=[ASC-nulls-first])
     LogicalProject(DEPTNO=[$7], EXPR$1=[RANK() OVER (PARTITION BY $0 ORDER BY $6 DESC NULLS LAST)], EXPR$2=[ROW_NUMBER() OVER (PARTITION BY $0 ORDER BY $6 NULLS FIRST)])
       LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+]]>
+    </Resource>
```
rubenada, line 9178: Do we need to add this to `SqlToRelConverterTest.xml` if we re not adding anything to `SqlToRelConverterTest.java` ?

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -2896,6 +2896,9 @@ private static List<RexLiteral> convertLiteralTypes(RelOptCluster cluster,
       RelDataType rowType, List<RexLiteral> values) {
     assert values.size() == rowType.getFieldCount();
     RexExecutor executor = cluster.getPlanner().getExecutor();
+    if (executor == null) {
+      executor = RexUtil.EXECUTOR;
+    }
     if (executor == null) {
```
NobiGo, line 2902: Need to delete.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -2886,6 +2886,39 @@ private static boolean isGroupId(AggCall c) {
...
+      executor = RexUtil.EXECUTOR;
+    }
+    if (executor == null) {
+      return values;
+    }
+
+    List<RelDataTypeField> fields = rowType.getFieldList();
+    List<RexLiteral> constants = new ArrayList<>();
+    RexBuilder builder = cluster.getRexBuilder();
+    RexSimplify simplify = new RexSimplify(builder, RelOptPredicateList.EMPTY, executor);
```
rubenada, line 2909: Nit: Could the RexSimplify be passed as a parameter of this method? That would make the code simpler & cleaner, and we would avoid (re)creating a RexSimplify on each iteration that calls `convertLiteralTypes`.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
mihaibudiu: I don't understand this comment.

rubenada: I think @NobiGo means that the currently the patch code is:
```
if (executor == null) {
  executor = RexUtil.EXECUTOR;
}
if (executor == null) {
  return values;
}
```
The second `if` could be removed.

suibianwanwank: I'm not sure RexSimplify#simlify is necessarily needed here. if not, the user should be allowed to set Executor to NULL. Instead, I think
`RexExecutor rexImpl = Util.first(cluster.getPlanner().getExecutor(), RexUtil.EXECUTOR);`
is good

mihaibudiu: If you don't simplify the expression won't be a RexLiteral

mihaibudiu: But turns out that the RelBuilder already contains a simplifier

rubenada (approved):

NobiGo (approved):

PR #4002, opened 2024-10-11
[CALCITE-6624] SqlParser should parse MySQL DATETIME type

dssysolyatin:


core/src/main/codegen/templates/Parser.jj
```diff
@@ -6182,6 +6182,12 @@ SqlTypeNameSpec DateTimeTypeName() :
     {
         return new SqlBasicTypeNameSpec(typeName, precision, s.end(this));
     }
+|
```
mihaibudiu, line 6185: is this supposed to be in Babel?

dssysolyatin: Regarding Babel, my initial plan was to implement DATETIME datatype parsing at the Babel parser level, but I noticed that we have DATETIME literal parsing for BigQuery in the core, which is also valid for MySQL. Therefore, I decided to put DATETIME type parsing in Parser.jj, as it would be better to have both DATETIME datatype parsing and DATETIME literal parsing in the same place.

babel/src/test/resources/sql/mysql.iq
```diff
@@ -0,0 +1,39 @@
...
++---------------------+
+| EXPR$0              |
++---------------------+
+| 1000-01-01 00:00:00 |
++---------------------+
+(1 row)
+
+!ok
+
+SELECT cast(cast(TIMESTAMP'9999-12-31 23:59:59' as DATETIME) as DATETIME);
```
mihaibudiu, line 31: Can you also write DATETIME '1000-01-01 00:00:00'?
Are you supposed to?

dssysolyatin: MySQL syntax is kind of messy

1. Time literals can only be defined using DATE 'str', TIME 'str', or TIMESTAMP 'str'. There is no syntax like DATETIME 'str'.

2. > The [TIMESTAMP](https://dev.mysql.com/doc/refman/8.4/en/datetime.html) syntax produces a [DATETIME](https://dev.mysql.com/doc/refman/8.4/en/datetime.html) value in MySQL because [DATETIME](https://dev.mysql.com/doc/refman/8.4/en/datetime.html) has a range that more closely corresponds to the standard SQL [TIMESTAMP](https://dev.mysql.com/doc/refman/8.4/en/datetime.html) type, which has a year range from 0001 to 9999. (The MySQL [TIMESTAMP](https://dev.mysql.com/doc/refman/8.4/en/datetime.html) year range is 1970 to 2038.)

Currently, `SELECT DATETIME '1000-01-01 00:00:00'` works in Calcite when using the MySQL dialect. Ideally, it should throw an exception. However, I’m not sure if anyone actually needs this behavior or if it’s worth spending time on.

mihaibudiu:
If we decide that this is a useful feature, the PR is fine.
I was very confused by the PR title (and the commit message). I changed the PR title to match the issue title, that's better.
I am worried that `DATETIME` means different things in different contexts.

mihaibudiu:
Anyone else wants to comment on this issue?
I think the PR is fine if we accept the fact that the issue is a problem.

PR #4003, opened 2024-10-14
Chore: Fix STDDEV inline comments

tisonkun:
TRIVIAL AS IS

Let me know if any process should be followed.


core/src/main/java/org/apache/calcite/rel/rules/AggregateReduceFunctionsRule.java
```diff
@@ -331,7 +331,7 @@ private RexNode reduceAgg(
         return reduceStddev(oldAggRel, oldCall, true, true, newCalls,
```
tisonkun, line 331: I wonder why the boolean flag is called `biased` ...

mihaibudiu (approved):

caicancai (approved):

mihaibudiu:
I think that we prefer not to have "chore" in the commit message, but it's fine

tisonkun:
> I think that we prefer not to have "chore" in the commit message, but it's fine

I'm glad to follow the pattern in this project. Just don't know what it is >_<

PR #4005, opened 2024-10-15
[CALCITE-6293] Support OR condition in Arrow adapter

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6293
https://issues.apache.org/jira/browse/CALCITE-6294

### DIFF
--- a/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
+++ b/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
@@ -120,23 +120,15 @@ public class ArrowTable extends AbstractTable
       projector = null;
 
       final List<TreeNode> conditionNodes = new ArrayList<>(conditions.size());
-      String op = conditions.get(conditions.size() - 1);
+      String op = "";
       for (String condition : conditions) {
-        if (condition.equals("or")) {
+        String[] data = condition.split(" ");
+        if (data[0].equals("or")) {
+          op = data[0];
           continue;
         }
-        String[] data = condition.split(" ");
-        List<TreeNode> treeNodes = new ArrayList<>(2);
-        treeNodes.add(
-            TreeBuilder.makeField(schema.getFields()
-                .get(schema.getFields().indexOf(schema.findField(data[0])))));
 
-        // if the split condition has more than two parts it's a binary operator
-        // with an additional literal node
-        if (data.length > 2) {
-          treeNodes.add(makeLiteralNode(data[2], data[3]));
-        }
-
+        List<TreeNode> treeNodes = createTreeNodes(data, schema);
         String operator = data[1];
         conditionNodes.add(
             TreeBuilder.makeFunction(operator, treeNodes, new ArrowType.Bool()));
@@ -145,7 +137,7 @@ public class ArrowTable extends AbstractTable
       if (conditionNodes.size() == 1) {
         filterCondition = TreeBuilder.makeCondition(conditionNodes.get(0));
       } else {
-        if (op .equals("or")) {
+        if (op.equals("or")) {
           TreeNode treeNode = TreeBuilder.makeOr(conditionNodes);
           filterCondition = TreeBuilder.makeCondition(treeNode);
         } else {
@@ -162,6 +154,27 @@ public class ArrowTable extends AbstractTable
     }
 
     return new ArrowEnumerable(arrowFileReader, fields, projector, filter);
+  }
+
+  /**
+   * Creates a list of TreeNode objects based on the provided data and schema.
+   *
+   * @param data   An array of strings representing the condition parts.
+   * @param schema The schema used to find the fields.
+   * @return A list of TreeNode objects representing the condition.
+   */
+  private List<TreeNode> createTreeNodes(String[] data, Schema schema) {
+    List<TreeNode> treeNodes = new ArrayList<>(2);
+    treeNodes.add(
+        TreeBuilder.makeField(schema.getFields()
+            .get(schema.getFields().indexOf(schema.findField(data[0])))));
+
+    // if the split condition has more than two parts it's a binary operator
+    // with an additional literal node
+    if (data.length > 2) {
+      treeNodes.add(makeLiteralNode(data[2], data[3]));
+    }
+    return treeNodes;
   }
 
   @Override public <T> Queryable<T> asQueryable(QueryProvider queryProvider,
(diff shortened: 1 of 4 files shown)


arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
```diff
@@ -278,13 +273,12 @@ static void initializeArrowState(@TempDir Path sharedTempDir)
     if (Bug.CALCITE_6294_FIXED) {
       plan = "PLAN=ArrowToEnumerableConverter\n"
           + "  ArrowProject(intField=[$0], stringField=[$1])\n"
```
mihaibudiu, line 275: how can both sides be changed?
The above bug should ensure that one side is never executed.

core/src/main/java/org/apache/calcite/util/Bug.java
```diff
@@ -204,11 +204,6 @@ public abstract class Bug {
    * is fixed. */
   public static final boolean CALCITE_6391_FIXED = false;
 
-  /** Whether
-   * <a href="https://issues.apache.org/jira/browse/CALCITE/issues/CALCITE-6293">
-   * [CALCITE-6293] Support OR condition in Arrow adapter</a> is fixed. */
-  public static final boolean CALCITE_6293_FIXED = false;
```
mihaibudiu, line 210: is this not used anywhere?

caicancai: yes

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
caicancai: @mihaibudiu I have implemented the in operator, but the implementation of the in operator will be based on the implementation of the or operator. Since these are two separate jira cases, I have to submit two separate PRs.

in addition，`SEARCH($0, Sarg[0, 1, 2]) `will be converted to `OR(=($0, 0), =($0, 1), =($0, 12))` in translateAnd.
In a way, this is not a bug.

mihaibudiu: this is not my question.
My question is "why do both cases of the if statement need to change?"

caicancai: @mihaibudiu Sorry, I misunderstood.
Here is my explanation.

`where in (0, 1, 2)` will be translated into `ArrowFilter(condition=[SEARCH($0, Sarg[0, 1, 2])]) `in the filter, and then translated into `OR(=($0, 0), =($0, 1), =($0, 2))`.
The same goes for `where "intField"=12 or "intField"= 11`, which will be translated into `ArrowFilter(condition=[SEARCH($0, Sarg[11, 12])])`, and then translated into `OR(=($0, 11), =($0, 12))`

Since my changes support `OR(=($0, 0), =($0, 1), =($0, 2) `and similar OR operators, it seems that they are all effective.

caicancai:
I added some tests related to the or operator

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
caicancai: I added tests related to the or operator to verify

mihaibudiu:
My main question is: is there a spec of Arrow so we can validate these strings that are emitted by the compiler?
How do we know these are correct?

caicancai:
> My main question is: is there a spec of Arrow so we can validate these strings that are emitted by the compiler? How do we know these are correct?

I understand. I am not very familiar with Arrow either. Maybe we need to seek help from someone who is more familiar with Arrow.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
```diff
@@ -120,7 +120,11 @@ public Enumerable<Object> query(DataContext root, ImmutableIntList fields,
       projector = null;
 
       final List<TreeNode> conditionNodes = new ArrayList<>(conditions.size());
+      String op = conditions.get(conditions.size() - 1);
```
NobiGo, line 123: Can we add a complex condition that includes different supported operators？ like
`a = ? and (b = ? or (a is null) ) `

caicancai:
Currently my approach does not support CNF, I will reopen this PR after completing https://issues.apache.org/jira/browse/CALCITE-6293

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
caicancai: I refactored this part of the code, and the current code has almost no impact on the functionality of the original code.

cc @mihaibudiu

caicancai: done

caicancai:
> My main question is: is there a spec of Arrow so we can validate these strings that are emitted by the compiler? How do we know these are correct?

@mihaibudiu Hello, I have read some arrow codes, and my current observation is this
In the makeOr api, arrow.ornode will be generated, and serialization will be performed inside. I understand that the serialization process will require data verification to determine whether the data is legal
```
class OrNode implements TreeNode {
  private final List<TreeNode> children;

  OrNode(List<TreeNode> children) {
    this.children = children;
  }

  @Override
  public GandivaTypes.TreeNode toProtobuf() throws GandivaException {
    GandivaTypes.OrNode.Builder orNode = GandivaTypes.OrNode.newBuilder();

    for (TreeNode arg : children) {
      orNode.addArgs(arg.toProtobuf());
    }

    GandivaTypes.TreeNode.Builder builder = GandivaTypes.TreeNode.newBuilder();
    builder.setOrNode(orNode.build());
    return builder.build();
  }
}
```
The same is true for other makeAnd and make Condition interfaces

mihaibudiu:
What is this code that you are quoting here with GandivaTypes?
My question is different: if I understand correctly, a predicate in Arrow cannot have an arbitrary shape, it *must* be in CNF form. Does this code ensure the fact that the predicate is in CNF form, or could it emit predicates that Arrow may interpret as incorrect?

Note that there are Boolean expression that are exponentially larger in CNF form, so this is not a simple problem.

caicancai:
> What is this code that you are quoting here with GandivaTypes? My question is different: if I understand correctly, a predicate in Arrow cannot have an arbitrary shape, it _must_ be in CNF form. Does this code ensure the fact that the predicate is in CNF form, or could it emit predicates that Arrow may interpret as incorrect?
> 
> Note that there are Boolean expression that are exponentially larger in CNF form, so this is not a simple problem.

Yes, my solution does not support CNF. If it does, it will be very complicated, but I will try it.

PR #4006, opened 2024-10-17
[CALCITE-6631] The common type for a comparison operator returns the …

NobiGo:
…wrong type when comparing a Java type long with a SQL type INTEGER

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
@@ -502,6 +502,13 @@ public abstract class AbstractTypeCoercion implements TypeCoercion {
       return null;
     }
 
+    // this prevents the conversion between JavaType and normal RelDataType,
+    // as well as between JavaType and JavaType.
+    if (type1 instanceof RelDataTypeFactoryImpl.JavaType
+        || type2 instanceof RelDataTypeFactoryImpl.JavaType) {
+      return null;
+    }
+
     SqlTypeName typeName1 = type1.getSqlTypeName();
     SqlTypeName typeName2 = type2.getSqlTypeName();
 
--- a/core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
+++ b/core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
@@ -407,6 +407,17 @@ class TypeCoercionTest {
         f.recordType("a", f.arrayType(f.intType)));
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6631">[CALCITE-6631]
+   * The common type for a comparison operator returns the wrong type
+   * when comparing a Java type long with a SQL type INTEGER</a>. */
+  @Test void testComparisonCoercionWithJavaType() {
+    final Fixture f = fixture();
+    f.comparisonCommonType(f.intJavaType, f.bigintJavaType, null);
+    f.comparisonCommonType(f.intJavaType, f.bigintType, null);
+  }
+
 
   /** Test case for case when expression and COALESCE operator. */
   @Test void testCaseWhen() {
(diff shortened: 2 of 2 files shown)


core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
```diff
@@ -747,8 +755,10 @@ protected Fixture(RelDataTypeFactory typeFactory,
       smallintType = this.typeFactory.createSqlType(SqlTypeName.SMALLINT);
       nullableSmallintType = this.typeFactory.createTypeWithNullability(smallintType, true);
       intType = this.typeFactory.createSqlType(SqlTypeName.INTEGER);
+      intJavaType = this.typeFactory.createJavaType(Integer.class);
       nullableIntType = this.typeFactory.createTypeWithNullability(intType, true);
       bigintType = this.typeFactory.createSqlType(SqlTypeName.BIGINT);
+      bigintJavaType = this.typeFactory.createJavaType(Long.class);
```
caicancai, line 761: Why not use BigInteger.class?

caicancai: The long type may be a timestamp

NobiGo: According to JavaToSqlTypeConversionRules#rules, the Java type corresponding to SqlTypeName.BIGINT  is long. So I use `Long`.

core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
```diff
@@ -407,6 +407,12 @@ private static ImmutableList<RelDataType> combine(
         f.recordType("a", f.arrayType(f.intType)));
   }
 
+  @Test void testComparisonCoercionWithJavaType() {
```
caicancai, line 410: Add jira link

core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
```diff
@@ -502,6 +502,11 @@ private RelDataType getTightestCommonTypeOrThrow(
       return null;
     }
 
+    if (type1 instanceof RelDataTypeFactoryImpl.JavaType
```
mihaibudiu, line 505: this means that no casts will be inserted. is this the desired behavior?

NobiGo: @mihaibudiu 
Hi, I found this problem in the ReflectiveSchemaTest#testCompareJavaAndSqlTypes.
Now in the condition `primitiveLong > 0`, the primitiveLong will be cast to RelDataType.INTEGER. This will cause value overflow(Now don't, because we don't add CAST in primitiveLong).
According to AbstractTypeCoercion,#needToCast prevents casting a JavaType to a normal RelDataType, so we don't need support to convert it.
Before CALCITE-6617, the ReflectiveSchemaTest#testCompareJavaAndSqlTypes returned `null` when comparing a Java type with an SQL type INTEGER.
Now the ReflectiveSchemaTest#testCompareJavaAndSqlTypes can't handle it when the input is Java type and Java type. Because no exception was thrown in the unit test, So I don't support it now.
﻿

mihaibudiu: This discussion is very useful, can you please add a little bit of this as a comment to this method, so other people who may wonder later have an answer ready?

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
NobiGo: Done!

core/src/test/java/org/apache/calcite/test/TypeCoercionTest.java
NobiGo: Done!

PR #4011, opened 2024-10-20
[CALCITE-6638] Optimization that simplifies expressions such as '1 > a or 1 <= a or a is null' to TRUE is incorrect when it requires casts that are not lossless

NobiGo:
… has lossless casts


mihaibudiu (approved):
Approved, but see the comment in JIRA about the issue/commit title.

PR #4012, opened 2024-10-23
[CALCITE-6639] Optimization that pulls up predicates causes ASOF JOIN validation failure

mihaibudiu:


mihaibudiu:
The macOS CI failed with a SIGBUS. I swear it's not the fault of this PR.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -774,8 +774,6 @@ public RelOptPredicateList inferPredicates(
       case INNER:
       case LEFT:
       case ANTI:
-      case ASOF:
-      case LEFT_ASOF:
```
caicancai, line 778: So in these two situations, should an exception be thrown, or do we need to take other measures?

caicancai (approved):
LGTM, just one question

suibianwanwank (approved):

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
suibianwanwank: It will return EMPTY in the end, looks good.

caicancai: > It will return EMPTY in the end, looks good.

thanks

PR #4013, opened 2024-10-23
[CALCITE-6640] RelMdUniqueKeys grows exponentially when key columns are repeated in projections

zabetak:
Avoid generating non-minimal/redundant key sets when computing the unique keys for columns that are repeated in the output.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdUniqueKeys.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdUniqueKeys.java
@@ -46,7 +46,7 @@ import org.apache.calcite.util.Util;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Maps;
+import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
@@ -54,7 +54,6 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -62,8 +61,6 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import static org.apache.calcite.rel.metadata.RelMdColumnUniqueness.PASSTHROUGH_AGGREGATIONS;
 import static org.apache.calcite.rel.metadata.RelMdColumnUniqueness.getConstantColumnSet;
-
-import static java.util.Objects.requireNonNull;
 
 /**
  * RelMdUniqueKeys supplies a default implementation of
@@ -172,8 +169,7 @@ public class RelMdUniqueKeys
       return ImmutableSet.of();
     }
 
-    Map<Integer, ImmutableBitSet> mapInToOutPos =
-        Maps.transformValues(inToOutPosBuilder.build().asMap(), ImmutableBitSet::of);
+    Multimap<Integer, Integer> mapInToOutPos = inToOutPosBuilder.build();
 
     ImmutableSet.Builder<ImmutableBitSet> resultBuilder = ImmutableSet.builder();
     // Now add to the projUniqueKeySet the child keys that are fully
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/calcite/rel/metadata/RelMdUniqueKeys.java
```diff
@@ -184,19 +180,12 @@ private static Set<ImmutableBitSet> getProjectUniqueKeys(SingleRel rel, RelMetad
...
-              Util.transform(colMask, in ->
-                  Util.filter(
-                      requireNonNull(mapInToOutPos.get(in),
-                          () -> "no entry for column " + in
-                              + " in mapInToOutPos: " + mapInToOutPos).powerSet(),
-                      bs -> !bs.isEmpty())));
-
-      resultBuilder.addAll(Util.transform(product, ImmutableBitSet::union));
+      // select key1, key1, val1, val2, key2 from ...
+      // the resulting unique keys would be {{0},{4}}, {{1},{4}}
```
caicancai, line 184: Maybe this part of the javadoc can be improved a little bit. When I first read this part of the javadoc, I didn’t understand what {0},{4} meant.

caicancai (approved):
Overall LGTM

core/src/main/java/org/apache/calcite/rel/metadata/RelMdUniqueKeys.java
caicancai: for example,
```
// Select fields key1, key1, val1, val2, and key2
// The query results will return records with unique key combinations
// Example of unique key combinations:
// {{0}, {4}} indicates that the key1 value of the first record is 0, and the key2 value is 4
// {{1}, {4}} indicates that the key1 value of the second record is 1, and the key2 value is 4
```

soumyakanti3578:
+1 on the request to improve documentation. Otherwise, this LGTM!

NobiGo (approved):

core/src/main/java/org/apache/calcite/rel/metadata/RelMdUniqueKeys.java
caicancai: Thank you very much for your answer, I agree with you

soumyakanti3578:
@zabetak It felt to me that it is a bit difficult to understand the documentation explaining the unique key combinations. But I agree with your comment above that this is not the right place to add more detailed documentation. So please ignore my comments regarding doc above. Thanks!

core/src/main/java/org/apache/calcite/util/ImmutableBitSet.java
```diff
@@ -978,6 +978,18 @@ public static boolean allContain(Collection<ImmutableBitSet> bitSets, int bit) {
     return true;
   }
 
+  /**
+   * Returns whether this is a minimal set with respect to the specified collection of bitSets.
+   */
```
julianhyde, line 983: Can you give an example in the javadoc?

julianhyde (changes requested):
You've added a test for minimality but I would go further - make sure every value returned by any RelMdUniqueKeys provider is minimal. I think the execution cost will be (pardon the pun) minimal.

Consider adding a method `static ImmutableBitSet areMinimal(Iterable<ImmutableBitSet>)` (or `List<ImmutableBitSet>` if it's more efficient). I have a feeling that the current implementation makes `N^2` tests but an implementation could make `N * (N - 1) / 2` tests (less than half as many).

zabetak:
@julianhyde Putting minimality logic on every return of the `RelMdUniqueKeys` handler doesn't feel right to me. Even if the overhead is minimal why adding seemingly redundant code? 

Moreover, if the handler goes rogue and starts to generate not minimal keys at some place then chances are that we are going to fail before even arriving to the minimality check/filter.

I don't mind adding the checks/fiters if you feel strongly about it but I see more cons than pros in this approach.

For the record, we already have [RelMdUniqueKeys#filterSupersets](https://github.com/apache/calcite/blob/ed6c33eb1d27882b26fc16b995dde78de1709aff/core/src/main/java/org/apache/calcite/rel/metadata/RelMdUniqueKeys.java#L369) that is currently used by the `Aggregate` handler to ensure that keys are minimal. If decide to apply the filter in every other handler then I guess we don't need another method in `ImmutableBitSet` and probably don't need the minimality check in the tests either.

zabetak:
The codebase has already "minimization" code in various other places (e.g., `RelMdUniqueKeys#filterSupersets`) and there is overlapping functionality with the new `ImmutableBitSet#isMinimal` API.
Based on these findings, I don't think a new API is necessary thus I [reverted](https://github.com/apache/calcite/pull/4013/commits/1e3b09a02bf465320b327de9714ec30c12e29e65) the respective code changes from.

If we want to add a new public API its better to refactor/use the existing "minimization" logic.
However, the refactoring will require some discussion on where to move the code and how we want the API to look like so it is out of the scope of this PR.

If there are no objections I plan to merge this PR as is in ~24hrs. If people feel that we need more work on the minimization checks/logic let me know and we can discuss this under a dedicated JIRA.

PR #4015, opened 2024-10-24
[CALCITE-6641] Compiling programs with ASOF joins can report obscure errors

mihaibudiu:


caicancai (approved):
I don't really understand this part, but the code looks good

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3911,14 +3918,21 @@ void checkAnd(SqlCall call) {
       }
     }
 
-    @Override public @Nullable SqlNode visit(final org.apache.calcite.sql.SqlCall call) {
+    @Override public @Nullable SqlNode visit(final SqlCall call) {
       SqlKind kind = call.getKind();
-      if (kind != SqlKind.AND && kind != SqlKind.EQUALS) {
+      if (kind != SqlKind.AND && kind != SqlKind.EQUALS && kind != SqlKind.CAST) {
```
caicancai, line 3923: It would be nice if you could explain why SqlKind cannot be equal to these types

mihaibudiu: The JavaDoc says "shuttle which determines whether an expression is a simple conjunction of equalities". AND is conjunction, and EQUALS is equalities. CAST is only allowed when applied to field names.

caicancai: thanks

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -3911,14 +3918,21 @@ void checkAnd(SqlCall call) {
...
         illegal = true;
```
NobiGo, line 3924: I check the `checkAnd(call)` and this method, Maybe we can exit directly when we find the illegal value is true.

NobiGo: Some minor advice:
1) SqlValidatorImpl#validateJoin(Line#3784). There is an extra `{}` behind the ASOF.
2) ConjunctionOfEqualities#checkAnd method's java comment. Before `Children`  There is an extra space.

mihaibudiu: the extra {} are actually useful, they create a local scope, and the variables in the local scope do not interfere with variables from the rest of the function. It also prevents accidental updates to other variables that may be in scope. If also helps with the debugger too: such variables go out of scope when the inner scope {} terminates, so in large functions with lots of variables it makes the debugging experience better. I use this trick a lot when I have to write very large functions.

NobiGo (approved):

PR #4016, opened 2024-10-25
[CALCITE-6642] AggregateUnionTransposeRule should account for changes in nullability of pushed down aggregates

arkanovicz:
When pushing down aggregates through child branches of a union in `AggregateUnionTransposeRule`, the nullability of aggregated columns may differ from the nullability of the corresponding union'ed columns.

Hence, the declared types of the aggregates created in each branch must be re-evaluated when needed.

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -7434,6 +7434,21 @@ class RelOptRulesTest extends RelOptTestBase {
         .check();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6642">[CALCITE-6642]
+   * AggregateUnionTransposeRule throws an assertion error when creating children aggregates
+   * when one input only has a non-nullable column</a>. */
+  @Test void testAggregateUnionTransposeWithOneInputNonNullable() {
+    final String sql = "select deptno, SUM(t) from (\n"
+        + "select deptno, 1 as t from sales.emp e1\n"
+        + "union all\n"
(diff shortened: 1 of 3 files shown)


rubenada (approved):
LGTM

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -7434,6 +7434,21 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
         .check();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-XXXX">[CALCITE-XXXX]
```
NobiGo, line 7438: The Java doc needs to be updated.

arkanovicz: Fixed.

suibianwanwank (approved):
LGTM.
Some minor suggestions: 
1. Building AggCall with RelBuilder
2. It seems that no multiple arguments AggCall is converted in this rule except for _Count_. Therefore, `origCall.getArgList().size() ! = 1` can be used as an assertion that alerts people to the need to redesign this part if they extend other AggFunctions in the future

arkanovicz:
> 1. Building AggCall with RelBuilder

IIUC, RelBuilder uses an intermediate AggCall construct, I'm not sure that this is needed here.

> 2. It seems that no multiple arguments AggCall is converted in this rule except for _Count_. Therefore, `origCall.getArgList().size() ! = 1` can be used as an assertion that alerts people to the need to redesign this part if they extend other AggFunctions in the future

Done.

PR #4017, opened 2024-10-25
[CALCITE-6643] Char_Length Function is not recognized in PrestoSql. Add LENGTH function in PrestoSqlDialect

vikramahuja1001:
PrestoSqlDialect by default will change any length function in a query to char_length function but char_length function is not available/supported in PrestoSql.
Add length function support in PrestoSqlDialect

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -7264,6 +7264,18 @@ class RelToSqlConverterTest {
     sql.withSnowflake().ok(expectedSnowflake);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6643">[CALCITE-6643]
+   *  Char_Length Function is not recognized in PrestoSql.
+   *  Add LENGTH function in PrestoSqlDialect</a>. */
+  @Test void testPrestoSqlLength() {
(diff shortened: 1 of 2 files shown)


mihaibudiu:
the commit message should match the JIRA issue and PR title
usually you start with the JIRA issue

vikramahuja1001:
@mihaibudiu i have updated the same. Thanks or letting me know.

mihaibudiu (approved):
Haven't checked the presto spec, but assuming this is right, the PR looks fine.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7264,6 +7264,14 @@ private void checkLiteral2(String expression, String expected) {
     sql.withSnowflake().ok(expectedSnowflake);
   }
 
+  @Test void testPrestoSqlLength() {
```
caicancai, line 7267: Add jira link

caicancai (approved):
overall LGTM

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
vikramahuja1001: Added the same

vikramahuja1001:
Thanks for the review @mihaibudiu , @caicancai

PR #4019, opened 2024-10-27
[CALCITE-6645] Support user-defined function without parentheses when db dialect's allowNiladicParentheses property is false

strongduanmu:
* pass SqlSyntax.FUNCTION_ID to SqlUserDefinedFunction when function has no parameters, and allowNiladicParentheses is false
* add udf test case for function without parentheses

### DIFF
--- a/core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
+++ b/core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
@@ -275,7 +275,7 @@ public class CalciteCatalogReader implements Prepare.CatalogReader {
     getFunctionsFrom(opName.names)
         .stream()
         .filter(predicate)
-        .map(function -> toOp(opName, function))
+        .map(function -> toOp(opName, function, config))
         .forEachOrdered(operatorList::add);
   }
 
@@ -296,7 +296,7 @@ public class CalciteCatalogReader implements Prepare.CatalogReader {
     for (String name : schema.getFunctionNames()) {
       schema.getFunctions(name, true).forEach(function -> {
         final SqlIdentifier id = new SqlIdentifier(name, SqlParserPos.ZERO);
-        list.add(toOp(id, function));
+        list.add(toOp(id, function, CalciteConnectionConfig.DEFAULT));
       });
     }
     return SqlOperatorTables.of(list);
@@ -304,7 +304,7 @@ public class CalciteCatalogReader implements Prepare.CatalogReader {
 
   /** Converts a function to a {@link org.apache.calcite.sql.SqlOperator}. */
   private static SqlOperator toOp(SqlIdentifier name,
-      final org.apache.calcite.schema.Function function) {
+      final org.apache.calcite.schema.Function function, CalciteConnectionConfig config) {
     final Function<RelDataTypeFactory, List<RelDataType>> argTypesFactory =
         typeFactory -> function.getParameters()
             .stream()
@@ -344,8 +344,12 @@ public class CalciteCatalogReader implements Prepare.CatalogReader {
     if (function instanceof ScalarFunction) {
       final SqlReturnTypeInference returnTypeInference =
           infer((ScalarFunction) function);
+      SqlSyntax syntax = function.getParameters().isEmpty()
+          && !config.conformance().allowNiladicParentheses()
+          ? SqlSyntax.FUNCTION_ID
+          : SqlSyntax.FUNCTION;
       return new SqlUserDefinedFunction(name, kind, returnTypeInference,
-          operandTypeInference, operandMetadata, function);
+          operandTypeInference, operandMetadata, function, syntax);
     } else if (function instanceof AggregateFunction) {
       final SqlReturnTypeInference returnTypeInference =
           infer((AggregateFunction) function);
@@ -423,7 +427,7 @@ public class CalciteCatalogReader implements Prepare.CatalogReader {
       if (schema != null) {
         for (String name : schema.getFunctionNames()) {
           schema.getFunctions(name, true).forEach(f ->
-              builder.add(toOp(new SqlIdentifier(name, SqlParserPos.ZERO), f)));
+              builder.add(toOp(new SqlIdentifier(name, SqlParserPos.ZERO), f, config)));
         }
       }
     }
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlUserDefinedFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlUserDefinedFunction.java
@@ -23,6 +23,7 @@ import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlFunctionCategory;
 import org.apache.calcite.sql.SqlIdentifier;
 import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlSyntax;
 import org.apache.calcite.sql.type.SqlOperandMetadata;
 import org.apache.calcite.sql.type.SqlOperandTypeChecker;
 import org.apache.calcite.sql.type.SqlOperandTypeInference;
@@ -41,6 +42,8 @@ import java.util.List;
 */
 public class SqlUserDefinedFunction extends SqlFunction {
   public final Function function;
+
+  public final SqlSyntax syntax;
 
   @Deprecated // to be removed before 2.0
   public SqlUserDefinedFunction(SqlIdentifier opName,
@@ -63,7 +66,17 @@ public class SqlUserDefinedFunction extends SqlFunction {
       @Nullable SqlOperandMetadata operandMetadata,
       Function function) {
     this(opName, kind, returnTypeInference, operandTypeInference,
-        operandMetadata, function, SqlFunctionCategory.USER_DEFINED_FUNCTION);
+        operandMetadata, function, SqlFunctionCategory.USER_DEFINED_FUNCTION, SqlSyntax.FUNCTION);
+  }
+
+  /** Creates a {@link SqlUserDefinedFunction} with sql syntax. */
+  public SqlUserDefinedFunction(SqlIdentifier opName, SqlKind kind,
+      SqlReturnTypeInference returnTypeInference,
+      SqlOperandTypeInference operandTypeInference,
+      @Nullable SqlOperandMetadata operandMetadata,
+      Function function, SqlSyntax syntax) {
+    this(opName, kind, returnTypeInference, operandTypeInference,
+        operandMetadata, function, SqlFunctionCategory.USER_DEFINED_FUNCTION, syntax);
   }
 
   /** Constructor used internally and by derived classes. */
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -283,7 +283,7 @@ private static SqlMonikerImpl moniker(CalciteSchema schema, @Nullable String nam
    * or classes.
    *
    * @see ModelHandler#addFunctions */
-  public static SqlOperatorTable operatorTable(String... classNames) {
+  public SqlOperatorTable operatorTable(String... classNames) {
```
mihaibudiu, line 286: why remove the static?

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -303,7 +303,7 @@ public static SqlOperatorTable operatorTable(String... classNames) {
   }
 
   /** Converts a function to a {@link org.apache.calcite.sql.SqlOperator}. */
-  private static SqlOperator toOp(SqlIdentifier name,
+  private SqlOperator toOp(SqlIdentifier name,
```
mihaibudiu, line 306: here as well

core/src/main/java/org/apache/calcite/sql/validate/SqlUserDefinedFunction.java
```diff
@@ -61,9 +64,9 @@ public SqlUserDefinedFunction(SqlIdentifier opName, SqlKind kind,
       SqlReturnTypeInference returnTypeInference,
       SqlOperandTypeInference operandTypeInference,
       @Nullable SqlOperandMetadata operandMetadata,
-      Function function) {
+      Function function, SqlSyntax syntax) {
```
mihaibudiu, line 67: you cannot change public constructors - this is a breaking change for clients.
you have to keep this constructor, and add another one.

core/src/test/java/org/apache/calcite/test/UdfTest.java
```diff
@@ -423,7 +447,8 @@ private CalciteAssert.AssertThat withUdf() {
 
   @Test void testUdfOverloadedNullable() {
     final CalciteAssert.AssertThat with = withUdf();
-    with.query("values (\"adhoc\".count_args(),\n"
+    with.with(SqlConformanceEnum.MYSQL_5)
```
mihaibudiu, line 450: why is this change needed?

plus/src/main/java/org/apache/calcite/chinook/CalciteConnectionProvider.java
```diff
@@ -44,6 +44,7 @@ public Properties provideConnectionInfo() throws IOException {
     Properties info = new Properties();
     info.setProperty("lex", "MYSQL");
     info.setProperty("model", "inline:" + provideSchema());
+    info.setProperty("conformance", "MYSQL_5");
```
mihaibudiu, line 47: why is this change needed?

testkit/src/main/java/org/apache/calcite/test/CalciteAssert.java
```diff
@@ -205,6 +205,10 @@ private CalciteAssert() {}
           return this;
         }
 
+        @Override public AssertThat with(SqlConformanceEnum conformance) {
```
mihaibudiu, line 208: what happens here?

testkit/src/main/java/org/apache/calcite/util/Smalls.java
```diff
@@ -683,6 +683,13 @@ public static int eval(int x) throws IllegalArgumentException, IOException {
     }
   }
 
+  /** User-defined function with niladic parentheses. */
+  public static class MyNiladicParenthesesFunction {
+    public String eval() {
```
mihaibudiu, line 688: please add a comment explaining that this function returns the constant string "foo".

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
strongduanmu: Since we need to call the non-static `config` field inside the toOp method to determine `config.conformance().allowNiladicParentheses()`, the toOp method cannot be a static method.

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
strongduanmu: The operatorTable method will call the toOp method. If the toOp method cannot be a static method, then the operatorTable method also needs to remove static.

strongduanmu: Besides, I found operatorTable is not called anywhere, should we delete this method?

core/src/main/java/org/apache/calcite/sql/validate/SqlUserDefinedFunction.java
strongduanmu: Hi @mihaibudiu thank you for your suggestion. I found this public constructor, which is only called in the `CalciteCatalogReader#toOp` method. When we add SqlSyntax to the toOp method, the original constructor that does not contain SqlSyntax will not be called.

In this case, does it make sense to keep a constructor that is not called? In addition, I don't know whether this constructor will be called by the user. If the user can call it, then we should indeed keep the original constructor.

core/src/test/java/org/apache/calcite/test/UdfTest.java
strongduanmu: When we use the default conformance, its allowNiladicParentheses method returns false, and functions without parameters should not use parentheses. In order to ensure that the existing test cases can run normally, we need to explicitly set the conformance and ensure that the allowNiladicParentheses of the conformance returns true.

testkit/src/main/java/org/apache/calcite/test/CalciteAssert.java
strongduanmu: Same as the previous one, we need to set the conformance props to control whether the parameterless function can use parentheses.

testkit/src/main/java/org/apache/calcite/util/Smalls.java
strongduanmu: I'm very sorry that this naming is confusing for you. This function is to return a constant, and this constant has no meaning. It is just for testing niladic function. It has no meaning just like 0 or -1 returned in other functions.

strongduanmu: Do you have any suggestions for better naming for constants used like this?

core/src/main/java/org/apache/calcite/sql/validate/SqlUserDefinedFunction.java
mihaibudiu: Since Calcite is an open-source library you cannot assume that the only code that calls these functions is in the Calcite tree. There are lots of third party projects using Calcite, and any of them may be calling any of the public functions. You cannot break compatibility unless you change the major version of the project.

core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
mihaibudiu: See my comment below about public methods; you cannot change public method signature without changing the major version of the project. That is called "semantic versioning" https://semver.org/

core/src/test/java/org/apache/calcite/test/UdfTest.java
mihaibudiu: perhaps a comment explaining that MYSQL_5 allows nilladic functions without parentheses will help people understand the point. That is not obvious.

plus/src/main/java/org/apache/calcite/chinook/CalciteConnectionProvider.java
mihaibudiu: The problem I see is that this is affecting all tests that use this connection provider. Do you really want all of these tests to use MYSQL_5 conformance?

testkit/src/main/java/org/apache/calcite/util/Smalls.java
mihaibudiu: The point I am making is that this is a test for a function with no arguments, and there are two kinds of such functions: non-deterministic, and constant. This is actually a constant function, and it is used for testing the syntax. That's what the comment should say.

core/src/main/java/org/apache/calcite/sql/validate/SqlUserDefinedFunction.java
strongduanmu: Thanks for your suggestion, I will keep the original constructor and add a new one for passing SqlSyntax.

core/src/test/java/org/apache/calcite/test/UdfTest.java
strongduanmu: Thank you @mihaibudiu, I will add comment for this change.

plus/src/main/java/org/apache/calcite/chinook/CalciteConnectionProvider.java
strongduanmu: Considering the lex used by the connection provider is MySQL, I think it should use MYSQL_5 conformance here, which makes the config consistent. What do you think?

testkit/src/main/java/org/apache/calcite/util/Smalls.java
strongduanmu: Thanks for your guidance, I will add comments to improve readability.

plus/src/main/java/org/apache/calcite/chinook/CalciteConnectionProvider.java
mihaibudiu: Perhaps this would work. I guess the fact that the tests passed shows it's not really a problem.

testkit/src/main/java/org/apache/calcite/util/Smalls.java
mihaibudiu: no need to reply on github for every suggestion - if you plan to do it anyway in your next commit.
Please add a new commit to the PR if you make changes, to make reviewing easier.
You will have to squash the commits before they get merged, though, but after the reviews

NobiGo:
I will review it within a week.

core/src/test/java/org/apache/calcite/test/UdfTest.java
```diff
@@ -407,14 +414,33 @@ private CalciteAssert.AssertThat withUdf() {
         .returns("P=null\n");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6645">[CALCITE-6645]
+   * User-defined function with niladic parentheses</a>. */
+  @Test void testUserDefinedFunctionWithNiladicParentheses() {
```
mihaibudiu, line 420: This looks pretty good.
Are there already negative tests that show errors when the conformance is wrong?

mihaibudiu (approved):

core/src/test/java/org/apache/calcite/test/UdfTest.java
strongduanmu: Thanks for your suggestion, I have added some negative tests.

mihaibudiu: In general you should wait to squash the commits until the PR has been approved.
This makes reviewing changes easier.

mihaibudiu (approved):

core/src/test/java/org/apache/calcite/test/UdfTest.java
strongduanmu: Thank you for your guidance. I will pay attention to this point in my future PRs.

mihaibudiu:
@NobiGo should we wait for your review?

caicancai (approved):
So cool

strongduanmu:
> @NobiGo should we wait for your review?

Hi @NobiGo, do you have time to review this PR? Thank you very much.

strongduanmu:
@NobiGo Thank you for your feedback, @mihaibudiu Could you please merge this PR? Thanks a lot.

PR #4021, opened 2024-10-29
[CALCITE-6649] Enhance RelMdPredicates pull up predicate from PROJECT

NobiGo:


mihaibudiu (approved):
I hope this is not too expensive; some of these objects are reallocated potentially many times.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -200,15 +204,17 @@ public RelOptPredicateList getPredicates(Project project,
             project.getRowType().getFieldCount());
 
     for (Ord<RexNode> expr : Ord.zip(project.getProjects())) {
-      if (expr.e instanceof RexInputRef) {
-        int sIdx = ((RexInputRef) expr.e).getIndex();
-        m.set(sIdx, expr.i);
+      RexNode rexNode = simplify.simplify(expr.e);
+      Ord<RexNode>  newExpr = new Ord<>(expr.i, rexNode);
+      if (newExpr.e instanceof RexInputRef) {
```
caicancai, line 209: Maybe adding some javadoc would be better

caicancai (approved):
LGTM

NobiGo:
This PR has been fixed by CALCITE-6586. I only added some tests to confirm we don't need to optimize this in RelMdPredicates.

NobiGo:

PR #4022, opened 2024-10-29
[CALCITE-6650] Optimize the IN sub-query and SOME sub-query by Metada…

NobiGo:
…ta RowCount

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
@@ -170,6 +170,13 @@ public class SubQueryRemoveRule
    */
   private static RexNode rewriteSome(RexSubQuery e, Set<CorrelationId> variablesSet,
       RelBuilder builder, int subQueryIndex) {
+    // If the sub-query is guaranteed to return 0 row, just return
+    // FALSE.
+    final RelMetadataQuery mq = e.rel.getCluster().getMetadataQuery();
+    final Double maxRowCount = mq.getMaxRowCount(e.rel);
+    if (maxRowCount != null && maxRowCount <= 0D) {
+      return builder.getRexBuilder().makeLiteral(Boolean.FALSE, e.getType(), true);
+    }
     // Most general case, where the left and right keys might have nulls, and
     // caller requires 3-valued logic return.
     //
@@ -459,11 +466,11 @@ public class SubQueryRemoveRule
     // TRUE.
     final RelMetadataQuery mq = e.rel.getCluster().getMetadataQuery();
     final Double minRowCount = mq.getMinRowCount(e.rel);
-    if (minRowCount != null && minRowCount >= 1D) {
+    if (minRowCount != null && minRowCount > 0D) {
       return builder.literal(true);
     }
     final Double maxRowCount = mq.getMaxRowCount(e.rel);
-    if (maxRowCount != null && maxRowCount < 1D) {
+    if (maxRowCount != null && maxRowCount <= 0D) {
       return builder.literal(false);
     }
     builder.push(e.rel);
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -170,6 +170,13 @@ private static RexNode rewriteCollection(RexSubQuery e,
    */
   private static RexNode rewriteSome(RexSubQuery e, Set<CorrelationId> variablesSet,
       RelBuilder builder, int subQueryIndex) {
+    // If the sub-query is guaranteed to return 0 row, just return
+    // FALSE.
+    final RelMetadataQuery mq = e.rel.getCluster().getMetadataQuery();
+    final Double maxRowCount = mq.getMaxRowCount(e.rel);
+    if (maxRowCount != null && maxRowCount < 1D) {
```
dssysolyatin, line 177: Are you sure that maxRowCount < 1D guarantees that the subquery will return zero records?

Since maxRowCount is a Double, I think if maxRowCount = 0.7, there is a possibility that it could return more than 0 records.

julianhyde: I have the same concern as @dssysolyatin. Could we instead check whether it is an instance of `Values` and then call `Values.isEmpty`? (Which is how `PruneEmptyRules` work.)

zabetak: We have [rule in PruneEmptyRules ](https://github.com/apache/calcite/blob/ed6c33eb1d27882b26fc16b995dde78de1709aff/core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java#L649) that does:
```java
          return maxRowCount != null && maxRowCount == 0.0;
```
Given that comparisons with doubles are inherently imprecise we may need to revisit that rule as well. Should we employ a threshold comparison?

zabetak: In addition, it seems that we employ the `maxRowCount < 1D` check in other places as well so whatever we decide we may need to propagate the changes everywhere.

mihaibudiu: The semantics of this value should be documented. Is it always guaranteed to be less than the actual row count of the table?

zabetak:
```
    /**
     * Estimates the max number of rows which will be returned by a relational
     * expression.
     *
     * <p>The default implementation for this query returns
     * {@link Double#POSITIVE_INFINITY},
     * but metadata providers can override this with their own cost models.
     *
     * @return upper bound on the number of rows returned
     */
```
I would say that the Javadoc is already explicit.

NobiGo: > I have the same concern as @dssysolyatin. Could we instead check whether it is an instance of `Values` and then call `Values.isEmpty`? (Which is how `PruneEmptyRules` works.)

I think we can't. Now we apply `SubQueryRemoveRule` before `PruneEmptyRules`. so the sub-query is not a VALUES.

NobiGo: Now we apply SubQueryRemoveRule before PruneEmptyRules. So PrunEmptyRules can no longer handle some plans converted by SubQueryRemoveRule.

NobiGo: @zabetak Perhaps we should clarify more clearly that when 'maxRowCount==0D' means no data is returned?

julianhyde: Sure. If you can find a better place to put the logic at https://github.com/apache/calcite/blob/3db4dd1ebf04abaf74d29464ae311a598ae929f7/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java#L456 then do it. Maybe a method `@Nullable Boolean RelMetadataQuery.isEmpty(RelNode)` (where true means always empty, false means always non-empty, null for everything else).

I cannot imagine a case where a provider would return a maxRowCount value less than one where the relation was not guaranteed empty, so I wouldn't worry too much whether the condition is `< 1D` or `== 0D`.

NobiGo: I think add `@Nullable Boolean RelMetadataQuery.isEmpty(RelNode)` is a good way for users to customize their implementation. The Default behavior will use `maxRowCount == 0.0`.

julianhyde: Note that we don't allow people to override methods in `RelMetadataQuery`. If we want people to be able to customize we should create a new kind of metadata.

NobiGo: I have created an issue CALCITE-6665 to log it.

mihaibudiu (approved):
This looks fine to me. Unless there are objections in the next few days I plan to merge this.

caicancai (approved):

mihaibudiu:
Feel free to squash the commits and merge if there are no other suggestions.

PR #4024, opened 2024-10-29
[CALCITE-6350] Unexpected result from UNION with literals expression

mihaibudiu:
The bug fixed is the following: 

In a set operation (UNION, INTERSECTION, EXCEPT) the participating relations are supposed to have the same columns with the same types. The type of the result is inferred to be the wider of the participating types. This is correct.

However, the types of the operands participating in the set operations need to be coerced to the output type. This part was missing. 

The fix has two parts: 
- in SetopOperandsTypeChecker the missing coercion has been introduced (the code was there, but wasn't called in most cases)
- in AbstractTypeCoercion a check for useless casts has been strengthened. The comments on the check were right, but the implementation was wrong.

The complete fix is only a few lines of code, but... unfortunately this change leads to many test failures, which had to be patched, 63 at my count. Some of the tests are obviously wrong, while the other are more subtle. I had to disable a couple which I couldn't actually fix (not because the results are wrong, but because the condition they were testing is no longer present in the test).

Many tests that contain plans are very hard for humans to evaluate for correctness.

I also had to change the typeSystem of BigQuery to use raggedUnionTypesToVarying, from empirical testing this seems to be right. (In fact, I suspect other dialects also needs this change, the default in Calcite is the opposite, which is very weird).

This may also seemingly degrade the quality of some query plans, because they will contain casts that were not there previously. Unfortunately, these plans were wrong.


core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
```diff
@@ -274,9 +274,10 @@ protected boolean needToCast(SqlValidatorScope scope, SqlNode node,
       return false;
     }
 
-    // No need to cast between char and varchar.
+    // No need to cast between char and unlimited varchar.
     if (SqlTypeUtil.isCharacter(toType)
-        && SqlTypeUtil.isCharacter(fromType)) {
+        && SqlTypeUtil.isCharacter(fromType)
+        && toType.getPrecision() == RelDataType.PRECISION_NOT_SPECIFIED) {
```
caicancai, line 280: Why do we need to judge precision?

caicancai: @caicancai `Character`includes char, varchar, longvarchar. Are these what you want?

mihaibudiu: clearly, the original code is wrong. The comment says "varchar", but the code does not check for varchar. Here I have made the code match the comment, which seems correct. I haven't investigated what happens to binary objects. It's possible that this check is still not strong enough, but it is certainly better than it was, and it fixes the problem reported.

In general, when fixing a bug, I am trying to make the minimal changes necessary to solve the problem reported, without breaking anything else. Calcite is very complex, as you can see, even a small change may affect lots of tests.

caicancai (approved):
LGTM

NobiGo (approved):
The PR looks good. Can this PR cover CALCITE-4590? If can, please add another test to cover it. If can't, it does not affect the integration of this PR.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -405,7 +405,8 @@ final RelMetadataFixture sql(String sql) {
         .assertColumnOriginIsEmpty();
   }
 
-  @Test void testColumnOriginsUnion() {
+  @Test @Disabled("Plan contains casts, which inhibit metadata propagation")
```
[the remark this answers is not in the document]
mihaibudiu: I could interpret your question in two ways:
- should metadata be propagated when a column has a cast?
- should we create a different test for metadata propagation?
The second one is a bit easier to do.
For the first one, I don't know the answer. If the answer is "yes", we should create an issue.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -4395,7 +4395,7 @@ private SqlDialect nonOrdinalDialect() {
         + "FROM (SELECT \"product_id\", \"net_weight\"\n"
         + "FROM \"foodmart\".\"product\"\n"
         + "UNION ALL\n"
-        + "SELECT \"product_id\", 0 AS \"net_weight\"\n"
+        + "SELECT \"product_id\", 0E0 AS \"net_weight\"\n"
```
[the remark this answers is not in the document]
mihaibudiu: Yes, this is necessary, since now the 0 constant is cast to the type of net_weight, which is double.

mihaibudiu:
@NobiGo I have filed https://github.com/apache/calcite/pull/4028 with a fix for https://issues.apache.org/jira/browse/CALCITE-4590. It is a completely different bug.

mihaibudiu:
Unfortunately this interferes negatively with the fix for https://issues.apache.org/jira/browse/CALCITE-4590, so more work will be needed.

mihaibudiu:
I have found one more bug in the implementation of casts, where casts from VARCHAR(N) to CHAR(N) for the same N were essentially ignored. This is wrong, because VARCHAR(N) does not store the trailing spaces, whereas CHAR(N) needs to store them.
With this change, which I added in an extra commit, hopefully the tests will pass again.

server/src/test/resources/sql/materialized_view.iq
```diff
@@ -39,12 +39,12 @@ select * from dept where deptno > 10;
 
 # Check contents
 select * from v;
-+--------+-------------+
-| DEPTNO | NAME        |
-+--------+-------------+
-|     20 | Marketing   |
-|     30 | Engineering |
```
NobiGo, line 46: The changes here are a bit strange. Maybe we can change the table definition statement.
```
# Create a source table
create table dept (deptno int not null, name varchar(10));
(0 rows modified)

!update

insert into dept
values (10, 'Sales'), (20, 'Marketing'), (30, 'Engineering');
(3 rows modified)
```
According to the output content, was cropping done during output or input？

mihaibudiu: I actually don't know how insert is implemented in Calcite.
But no matter what, the result is now correct, and it wasn't before. 
So I think this is actually a good test.

NobiGo: Ok. I get it.

mihaibudiu:
I am thinking to merge this with 2 commits, since the second commit solves an independent bug....

caicancai:
> I am thinking to merge this with 2 commits, since the second commit solves an independent bug....

Agreed, but it's not clear to me whether to create a jira case for the second commit

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7777,7 +7777,7 @@ private void checkLiteral2(String expression, String expected) {
     final String expectedDefault = "INSERT INTO \"SCOTT\".\"DEPT\""
         + " (\"DEPTNO\", \"DNAME\", \"LOC\")\n"
         + "VALUES (1, 'Fred', 'San Francisco'),\n"
-        + "(2, 'Eric', 'Washington   ')";
```
suibianwanwank, line 7780: Is this the expected result, char type value in calcite seem to require aligned lengths.

mihaibudiu: The types of these fields in the Scott database is actually VARCHAR(20), and VARCHAR types are not supposed to store the padding spaces, unlike CHAR(N) types. So the result is correct. I don't know exactly why the result was wrong before.

suibianwanwank (approved):

PR #4025, opened 2024-10-31
[CALCITE-6655] Aggregation of deeply nested window not detected when unparsing

linorosa:


core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -2122,6 +2123,28 @@ private boolean hasNested(
       return false;
     }
 
+    /** Recursively looks for a node matching the operandPredicate. */
```
mihaibudiu, line 2126: this is usually done using a visitor patter, see SqlShuttle.

linorosa: ahh thanks for the feedback @mihaibudiu. I just pushed a refactor to use an SqlShuttle at #
226b7d3

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -8741,6 +8741,48 @@ private void checkLiteral2(String expression, String expected) {
         .dialect(MssqlSqlDialect.DEFAULT).ok(mssqlExpected);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6655">[CALCITE-6655]
+   * Aggregation of deeply nested aggregations </a>.
```
mihaibudiu, line 8746: extra space before closing tag
You need to use the exact same message as in the JIRA case.
Please use this for your PR as well.
These conventions help us keep track of the correspondence between commits, PRs, and JIRA cases.

mihaibudiu (approved):
The squashing is done usually after the changes have accepted.
Code review may be easier if the reviewer can see separate commits addressing the suggestions.
So, yes, please go ahead and squash the commits in preparation for merging.

PR #4026, opened 2024-11-01
[CALCITE-6662] RelJson.rangePointFromJson cannot create Double numerics

olivrlee:


core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
```diff
@@ -880,6 +880,12 @@ private static RexNode translateInput(RelJson relJson, int input,
       assertThat(result, isLinux(expected));
     }
 
+    RexNode betweenDoubles =
```
mihaibudiu, line 883: I suspect this will test Doubles, but are there any tests for decimal at this point?

core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -939,10 +944,21 @@ public static <C extends Comparable<C>> Range<C> rangeFromJson(
   @SuppressWarnings({"rawtypes", "unchecked"})
   private static <C extends Comparable<C>> C rangeEndPointFromJson(Object o) {
     Exception e = null;
+    for (Class clsType : NUMERIC_VALUE_CLASSES) {
```
mihaibudiu, line 947: doesn't the JSON include type information so you don't have to guess?
at least for literal, it seems to include it.

olivrlee: The JSON includes the type information at the `RangeSet` level, but not down into the Range and rangeEndpoints levels. 
The rangeEndpoint only has `lower` and `upper`

core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
olivrlee: Hmm you're right, there's a missing case. 
The existing test that has values (`14`, `20`) that test Integers will use BigDecimal and not Double, however 

a test case like 


```
    RexNode betweenDecimal =
        rexBuilder.makeBetween(b.literal(45),
            b.literal(BigDecimal.valueOf(20.0)),
            b.literal(BigDecimal.valueOf(30.0)));
```
results in it using Doubles ... 
Let me know if you have any ideas -- we might have to serialize the typing when creating the JSON to avoid this?

mihaibudiu:
The "right" way to do this is to use the type information, passing it from the higher levels of the JSON document which know it.
The type information should be passed down from toRex() through sargFromJson. Today it's not, and these methods are public, so probably a new version of these methods (sargFromJson, rangeSetFromJson, rangeFromJson, rangeEndpointFromJson) should be created which takes a type as an extra argument.

core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -939,10 +944,21 @@ public static <C extends Comparable<C>> Range<C> rangeFromJson(
...
+        }
+      } catch (JsonProcessingException|InvalidAttributesException  ex) {
+        e = ex;
+      }
+    }
     for (Class clsType : VALUE_CLASSES) {
       try {
-        return (C) OBJECT_MAPPER.readValue((String) o, clsType);
-      } catch (JsonProcessingException ex) {
+          return (C) OBJECT_MAPPER.readValue((String) o, clsType);
```
NobiGo, line 960: It seems that you are refactoring this PR. Please note that there are some unnecessary changes here.

olivrlee:
Thanks for the reviews yesterday. I've updated with the suggested changes to passing in types and have updated tests.

core/src/main/java/org/apache/calcite/rel/externalize/RelJson.java
```diff
@@ -951,6 +1014,42 @@ private static <C extends Comparable<C>> C rangeEndPointFromJson(Object o) {
         e);
   }
 
+  private static <C extends Comparable<C>> C rangeEndPointFromJson(Object o, RelDataType type) {
```
mihaibudiu, line 1017: I think this is much better

mihaibudiu (approved):

PR #4028, opened 2024-11-01
[CALCITE-4590] Incorrect query result with fixed-length string

mihaibudiu:


core/src/main/java/org/apache/calcite/util/NlsString.java
```diff
@@ -227,6 +227,27 @@ public NlsString rtrim() {
...
   }
 
+  /**
+   * Returns a string the same as this but with spaces trimmed from the
+   * right such that the result has the specified size.
+   *
+   * @param resultSize Expected size for result string.  If negative, it indicates
+   *                   that no trimming should be done.
+   */
+  public NlsString rtrim(int resultSize) {
```
dssysolyatin, line 237: @mihaibudiu The implementation is a bit confusing: For example:
```
NlsString string = new NlsString("hello world ", null, null);
string.rtrim(5);
```
produces "helloworld". It removes spaces from the middle, even though the documentation claims it should remove them only from the right.

mihaibudiu: Good catch, should call rtrim on the fragment, not trim.

mihaibudiu: Sorry, pushed one more correction, hopefully now it makes more sense.

dssysolyatin (approved):

PR #4029, opened 2024-11-02
[CALCITE-6666] Support Oracle SYSDATE and SYSTIMESTAMP functions

strongduanmu:
- Support Oracle SYSDATE and SYSTIMESTAMP functions

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -5364,6 +5364,30 @@ public class SqlFunctions {
     return timestampToTime(localTimestamp(root));
   }
 
+  /** SQL {@code SYSTIMESTAMP} function. */
+  @NonDeterministic
+  public static long sysTimestamp(DataContext root) {
+    return DataContext.Variable.SYS_TIMESTAMP.get(root);
+  }
+
+  /** SQL {@code SYSDATE} function.
+   *
+   * <p> When the date is before 1970-01-01 00:00:00, for example: 1969-12-31 23:59:59,
+   * the timestamp will return a negative value, such as -1000. The date(days since epoch)
+   * returned by timestampToDate(-1000) is 0, so we need to additionally judge the result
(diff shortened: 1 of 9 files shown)


caicancai (commented):
It is better to add the test in SqlOperatorTest

strongduanmu:
> It is better to add the test in SqlOperatorTest

@caicancai Thanks for your suggestion, I have added the SqlOperatorTest.

caicancai (approved):
LGTM

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -5364,6 +5364,24 @@ public static int localTime(DataContext root) {
...
+  public static long sysTimestamp(DataContext root) {
+    return DataContext.Variable.SYS_TIMESTAMP.get(root);
+  }
+
+  /** SQL {@code SYSDATE} function. */
+  @NonDeterministic
+  public static int sysDate(DataContext root) {
+    final long timestamp = sysTimestamp(root);
+    int date = timestampToDate(timestamp);
+    final int time = timestampToTime(timestamp);
```
mihaibudiu, line 5378: why would this happen?
it it for dates before 1970?

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
strongduanmu: Yes, when the date is before `1970-1-1`, for example: `1969-12-31 23:59:59`, timestamp will return `-1000`, and the date(days since epoch) calculated by `timestampToDate(-1000)` is 0, so we need to additionally judge the result of `timestampToTime(-1000)`. If its value is less than 0, we need to reduce date by 1 to ensure the accuracy of date.

mihaibudiu: this would be a nice comment to add to this code

PR #4031, opened 2024-11-03
[CALCITE-6667] Arrow adapter should push down arithmetic operations

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6667

### DIFF
--- a/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
+++ b/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
@@ -170,9 +170,47 @@ class ArrowTranslator {
     case CAST:
       // FIXME This will not work in all cases (for example, we ignore string encoding)
       return translateBinary2(op, ((RexCall) left).operands.get(0), right);
+    case PLUS:
+      right = translateArithmeticOp(left, right);
+      return translateBinary2(op, ((RexCall) left).operands.get(0), right);
+    case MINUS:
+      right = translateArithmeticOp(left, right);
+      return translateBinary2(op, ((RexCall) left).operands.get(0), right);
     default:
       return null;
     }
+  }
+
+  /**
+   * Translates a call to an arithmetic operator.
+   * This method handles arithmetic operations like addition and subtraction
+   * between literals and adjusts the right operand accordingly.
+   *
+   * @param left The left operand, which is an arithmetic operation
+   * @param right The right operand, which is a literal
+   * @return A new RexNode representing the adjusted right operand
+   */
+  private RexNode translateArithmeticOp(RexNode left, RexNode right) {
+    RexNode leftOperand = ((RexCall) left).operands.get(1);
+    switch (left.getKind()) {
+    case PLUS:
+      if (leftOperand.getKind() == SqlKind.LITERAL && right.getKind() == SqlKind.LITERAL) {
+        BigDecimal leftValue = ((RexLiteral) leftOperand).getValueAs(BigDecimal.class);
+        BigDecimal rightValue = ((RexLiteral) right).getValueAs(BigDecimal.class);
+        return rexBuilder.makeLiteral(rightValue.subtract(leftValue), left.getType(), false);
+      }
+      break;
+    case MINUS:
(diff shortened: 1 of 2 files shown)


caicancai:
Currently, only simple constant addition and subtraction operations are supported. I did not perform complex type tests because there are some other problems with the current adaptation of the arrow type.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
```diff
@@ -170,11 +170,49 @@ private String translateBinary(String op, String rop, RexCall call) {
     case CAST:
       // FIXME This will not work in all cases (for example, we ignore string encoding)
       return translateBinary2(op, ((RexCall) left).operands.get(0), right);
+    case PLUS:
+      right = translateArithmeticOp(left, right);
+      return translateBinary2(op, ((RexCall) left).operands.get(0), right);
+    case MINUS:
```
mihaibudiu, line 176: these two cases could be merged

caicancai: done

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
```diff
@@ -170,11 +170,47 @@ private String translateBinary(String op, String rop, RexCall call) {
...
+   * Translates a call to an arithmetic operator.
+   * This method handles arithmetic operations like addition and subtraction
+   * between literals and adjusts the right operand accordingly.
+   *
+   * @param left The left operand, which is an arithmetic operation
+   * @param right The right operand, which is a literal
+   * @return A new RexNode representing the adjusted right operand
+   */
+  private RexNode translateArithmeticOp(RexNode left, RexNode right) {
+    RexNode leftOperand = ((RexCall) left).operands.get(1);
```
mihaibudiu, line 192: This optimization is already part of PROJECT_REDUCE_EXPRESSIONS. I don't think it should be replicated in any specific backend.

caicancai: I'm not sure if it works, and it seems no one has tried it yet.

mihaibudiu: I don't know what you mean, in RelOptRulesTest there are at least 59 tests that use this optimization.

caicancai: I understand what you mean, but it seems that no other adapters are specified for testing. I understand that RelOptRulesTest is for those Jdbc protocol dialects, such as mysql, postgres and so on. I am not sure whether adapters such as arrow, es, etc. are effective.

caicancai: If possible, I will write a test to verify it over the weekend

mihaibudiu: That is incorrect, the RelOptRules are designed to work for all dialects.
A Calcite-based compiler will have a parser, a converter, an optimizer, and a back-end.
The job of doing optimizations belongs to the optimizer, not to the back-end.
If you put optimizations in the back-end, you have to reimplement them for every target.

caicancai: Thank you for your answer, so I just need to translate the plan without any optimization, I will improve this pr later

caicancai: @mihaibudiu It seems that Arrow's API does not support operations like `"intField\" + 1 = 12`. I had to convert it to `"intField\" = 11` in Calcite.

Should I call the PROJECT_REDUCE_EXPRESSIONS optimization rule in the arrow adapter?

mihaibudiu: Adapters should never call optimizations.
This particular optimization does constant folding, and the expression you show does not have any constant subexpressions, so this optimization will not help.

PR #4032, opened 2024-11-03
[CALCITE-6668] Arrow adapter should support IS FALSE Operator and IS TRUE Operator

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6668

### DIFF
--- a/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
+++ b/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
@@ -114,7 +114,13 @@ class ArrowTranslator {
     return predicates;
   }
 
-  /** Translate a binary or unary relation. */
+  /**
+   * Translates a binary or unary relation.
+   *
+   * @param node A RexNode that always evaluates to a boolean expression.
+   *             Currently, this method is only called from translateAnd.
+   * @return The translated SQL string for the relation.
+   */
   private String translateMatch2(RexNode node) {
     switch (node.getKind()) {
     case EQUALS:
@@ -133,6 +139,15 @@ class ArrowTranslator {
       return translateUnary("isnull", (RexCall) node);
     case IS_NOT_NULL:
       return translateUnary("isnotnull", (RexCall) node);
+    case IS_NOT_TRUE:
+      return translateUnary("isnottrue", (RexCall) node);
+    case IS_NOT_FALSE:
+      return translateUnary("isnotfalse", (RexCall) node);
+    case INPUT_REF:
+      final RexInputRef inputRef = (RexInputRef) node;
+      return fieldNames.get(inputRef.getIndex()) + " istrue";
+    case NOT:
+      return translateUnary("isfalse", (RexCall) node);
     default:
       throw new UnsupportedOperationException("Unsupported operator " + node);
     }
--- a/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterDataTypesTest.java
+++ b/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterDataTypesTest.java
@@ -176,4 +176,19 @@ public class ArrowAdapterDataTypesTest {
         .returns(result)
         .explainContains(plan);
   }
+
+  @Test void testBooleanProject() {
+    String sql = "select \"booleanField\" from arrowdatatype";
+    String plan = "PLAN=ArrowToEnumerableConverter\n"
+        + "  ArrowProject(booleanField=[$7])\n"
+        + "    ArrowTableScan(table=[[ARROW, ARROWDATATYPE]], fields=[[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]])\n\n";
+    String result = "booleanField=null\nbooleanField=true\nbooleanField=false\n";
+    CalciteAssert.that()
+        .with(arrow)
+        .query(sql)
+        .limit(3)
+        .returns(result)
+        .explainContains(plan);
+  }
+
 }
--- a/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
+++ b/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
@@ -77,6 +77,10 @@ class ArrowAdapterTest {
     arrowDataGenerator.writeArrowData(dataLocationFile);
     arrowDataGenerator.writeScottEmpData(arrowFilesDirectory);
 
+    File datatypeLocationFile = arrowFilesDirectory.resolve("arrowdatatype.arrow").toFile();
+    ArrowData arrowtypeDataGenerator = new ArrowData();
+    arrowtypeDataGenerator.writeArrowDataType(datatypeLocationFile);
+
     arrow = ImmutableMap.of("model", modelFileTarget.toAbsolutePath().toString());
   }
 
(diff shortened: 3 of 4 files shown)


arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
```diff
@@ -133,6 +133,15 @@ private String translateMatch2(RexNode node) {
       return translateUnary("isnull", (RexCall) node);
     case IS_NOT_NULL:
       return translateUnary("isnotnull", (RexCall) node);
+    case IS_NOT_TRUE:
+      return translateUnary("isnottrue", (RexCall) node);
+    case IS_NOT_FALSE:
+      return translateUnary("isnotfalse", (RexCall) node);
+    case INPUT_REF:
```
mihaibudiu, line 140: I don't understand why this assumes the type of the result is boolean.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterDataTypesTest.java
```diff
@@ -176,4 +176,94 @@ static void initializeArrowState(@TempDir Path sharedTempDir)
...
+
+  @Test void testArrowProjectFieldsWithNotFilter() {
+    String sql = "select \"booleanField\"\n"
+        + "from arrowdatatype\n"
+        + "where \"booleanField\" is false";
+    String plan = "PLAN=ArrowToEnumerableConverter\n"
+        + "  ArrowProject(booleanField=[$7])\n"
+        + "    ArrowFilter(condition=[NOT($7)])\n"
+        + "      ArrowTableScan(table=[[ARROW, ARROWDATATYPE]], fields=[[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]])\n\n";
+    String result = "booleanField=false\nbooleanField=false\n";
```
mihaibudiu, line 223: why is this result correct?
There aren't two rows where the field is false

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
caicancai: This is because of the `TreeBuilder.makeCondition `interface feature of arrow, which requires a function to generate the corresponding filter condition, and `istrue` is such a function for it.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterDataTypesTest.java
caicancai: Because the filter condition is `where "booleanField" is false`, the returned result is two false, which seems to be no problem.

mihaibudiu: what is the content of the input table?

caicancai: At present, there are three main types of content: `true, false, null`

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
mihaibudiu: why is this function translating a filter?

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterDataTypesTest.java
caicancai: No, there are 20 rows of data in total
```
  private void booleanField(FieldVector fieldVector, int rowCount) {
    BitVector bitVector = (BitVector) fieldVector;
    bitVector.setInitialCapacity(rowCount);
    bitVector.allocateNew();
    for (int i = 0; i < rowCount; i++) {
      if (i % 3 == 0) {
        bitVector.setNull(i);
      } else {
        bitVector.set(i, this.booleanValue ? 1 : 0);
      }
      this.booleanValue = !this.booleanValue;
    }
    fieldVector.setValueCount(rowCount);
  }
```
It's just that the field content is distributed regularly according to the three types of content: null, true, and false.
I set limit(2) on the final result, so it only shows two rows, but there are actually more than two rows.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
caicancai: I'm not sure, because when the filter structure is passed in, a node node is passed in, which forces me to translate is true

caicancai:
```
    String plan = "PLAN=EnumerableCalc(expr#0=[{inputs}], booleanField=[$t0], $condition=[$t0])\n"
        + "  ArrowToEnumerableConverter\n"
        + "    ArrowProject(booleanField=[$7])\n"
        + "      ArrowTableScan(table=[[ARROW, ARROWDATATYPE]], fields=[[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]])\n\n";
```
This is the untranslated plan. I hope it helps you. Since I don’t use Calcite in my work, I am not very clear about this part for now.

caicancai: I thought about it and it is OK to translate is true. Calcite will trim simple is true, such as `where 1 is true`, because they are all constants. But `booleanField is true `is obviously not.
@mihaibudiu What do you think?

mihaibudiu: This is never a constant, it's a reference to a field within the row of the input relation, which can have any type.
Why do you think the type is boolean at this point?
Can you add an assertion that the type is always Boolean?

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterDataTypesTest.java
```diff
@@ -176,4 +176,94 @@ static void initializeArrowState(@TempDir Path sharedTempDir)
...
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6638">[CALCITE-6638]
+   * Arrow adapter should support IS FALSE Operator and IS TRUE Operator</a>. */
+  @Test void testArrowProjectWithIsTrueFilter() {
+    String sql = "select \"booleanField\"\n"
+        + "from arrowdatatype\n"
+        + "where \"booleanField\" is true";
+    String plan = "PLAN=ArrowToEnumerableConverter\n"
```
suibianwanwank, line 201: Why is this test in ArrowAdapterDataTypesTest, it seems more appropriate in ArrowAdapterTest.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterDataTypesTest.java
```diff
@@ -176,4 +176,94 @@ static void initializeArrowState(@TempDir Path sharedTempDir)
...
+        .limit(2)
+        .returns(result)
+        .explainContains(plan);
+  }
+
+
+  @Test void testArrowProjectFieldsWithIsNotTrueFilter() {
+    String sql = "select \"booleanField\"\n"
+        + "from arrowdatatype\n"
+        + "where \"booleanField\" is not true";
```
suibianwanwank, line 237: Can more complex expressions be supported at this time? e.g. `(intField > 10) IS TRUE`

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterDataTypesTest.java
caicancai: done

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterDataTypesTest.java
caicancai: done

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
mihaibudiu: The JavaDoc on this function is actually wrong, which leads to the confusion.
The JavaDoc should say that node is always an expression that evaluates to a Boolean value. This is only called from translateAnd. With this information the implementation makes sense.

caicancai: @mihaibudiu Oh, I see, thanks for reminding me. I have improved the JavaDoc

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
```diff
@@ -114,7 +114,13 @@ private List<String> translateAnd(RexNode condition) {
     return predicates;
   }
 
-  /** Translate a binary or unary relation. */
+  /**
+   * Translates a binary or unary relation.
+   *
+   * @param node A RexNode that always evaluates to a boolean expression.
+   *             Currently, this method is only called from translateAnd.
```
caicancai, line 121: Since the current Arrow adapter does not support CNF, this method is currently only called from translateAnd

mihaibudiu: A Boolean formula does not have to be in CNF or DNF form.
Currently Arrow supports only DNF forms for Boolean formulas.
There is an algorithm to convert any Boolean formula to DNF, but it can be very expensive in the worst case. There is no workaround, since the worst-case DNF formula can be exponential in the size of the original formula.

caicancai: Yes, I'm trying, but it seems very complicated

mihaibudiu (approved):

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
caicancai: Thank you very much for your advice, I am supporting CNF, but it seems that my thinking is wrong.

mihaibudiu:
you can squash the commits for merging

caicancai:
> you can squash the commits for merging

If there is no other comment after 24 hours, I will squash

suibianwanwank (approved):

caicancai:
No other comments, squash commits.
@mihaibudiu @suibianwanwank thank you for your review

PR #4033, opened 2024-11-05
[CALCITE-6651] Use RelBuilder in SqlToRelConverter to construct Union

suibianwanwank:


babel/src/test/resources/sql/big-query.iq
```diff
@@ -1594,13 +1594,13 @@ WITH letters AS
...
 SELECT SPLIT(letter_group, ' ') as example
 FROM letters;
-+-----------+
-| example   |
-+-----------+
-| []        |
-| [a]       |
-| [b, c, d] |
-+-----------+
++--------------+
```
suibianwanwank, line 1597: The test results here look a little strange, but this seems to be as expected. [PR-4024] Changed raggedUnionTypesToVarying, which will fix the problem

babel/src/test/resources/sql/big-query.iq
```diff
@@ -1647,8 +1647,8 @@ FROM letters;
 +-----------+
 | example   |
 +-----------+
-| []        |
-| [A]       |
+| [     ]   |
```
suibianwanwank, line 1650: I'm not familiar with the binray type, but from the plan it makes sense.But the execution result is not what I expected and I don't know where BabelTest is executed.
```
LogicalProject(example=[SPLIT($0, X'ff':BINARY(1))])
  LogicalValues(tuples=[[{ X'0000000000' }, { X'4100000000' }, { X'42ff43ff44' }]])
```

mihaibudiu: Let's merge https://github.com/apache/calcite/pull/4024 first, it actually takes care of some of these problems.
It addresses some very related issues.

mihaibudiu:
Please rebase on main, I expect there will be some conflicts.
But the tests you mention should probably work out fine.

suibianwanwank:
> Please rebase on main, I expect there will be some conflicts. But the tests you mention should probably work out fine.

Done, the issues have been resolved.

caicancai (approved):
Overall LGTM

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -2605,7 +2606,13 @@ private void checkPredicates(RelOptCluster cluster, RelOptTable empTable,
     final String sql = "select cast(null as integer) as a\n"
         + "union all\n"
         + "select 5 as a";
-    final Union rel = (Union) sql(sql).toRel();
+    final Union rel = (Union) sql(sql)
+        .withFactory(a ->
+            a.withSqlToRelConfig(
```
mihaibudiu, line 2611: can you factor out this configuration into a single function to avoid repeating this pattern?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -3148,7 +3148,7 @@ private void checkPushJoinThroughUnionOnRightDoesNotMatchSemiOrAntiJoin(JoinRelT
     sql(sql)
         .withRule(CoreRules.UNION_TO_VALUES)
         .withInSubQueryThreshold(0)
-        .check();
+        .checkUnchanged();
```
mihaibudiu, line 3151: it really looks like these tests are no longer testing this particular rule.
I am not sure what should be done about that.

core/src/test/java/org/apache/calcite/tools/PlannerTest.java
```diff
@@ -452,11 +452,6 @@ private void emptyUnions32(UnionMergeRule... extraRules)
         + "  EnumerableValues(tuples=[[{ 1 }]])\n"
         + "  EnumerableValues(tuples=[[{ 2 }]])\n";
 
-    checkUnionPruning("values(1)"
```
mihaibudiu, line 455: what happened with this test?

core/src/test/resources/sql/sub-query.iq
```diff
@@ -3452,9 +3452,8 @@ EnumerableCalc(expr#0..12=[{inputs}], expr#13=[0:BIGINT], expr#14=[=($t8, $t13)]
           EnumerableAggregate(group=[{}], agg#0=[COUNT()], agg#1=[COUNT($0)])
             EnumerableValues(tuples=[[{ null }, { 7782 }]])
     EnumerableSort(sort0=[$0], dir0=[ASC])
-      EnumerableAggregate(group=[{0}], agg#0=[MIN($1)])
-        EnumerableCalc(expr#0=[{inputs}], expr#1=[true], proj#0..1=[{exprs}])
-          EnumerableValues(tuples=[[{ null }, { 7782 }]])
+      EnumerableCalc(expr#0=[{inputs}], expr#1=[true], proj#0..1=[{exprs}])
```
mihaibudiu, line 3455: do you know what happened to the aggregate?

suibianwanwank: AggregateRemoveRule is applied, The difference is in the metadata of the extracted UNION
```
if (!SqlFunctions.isTrue(mq.areColumnsUnique(input, aggregate.getGroupSet()))) {
      return;
}
```

core/src/test/java/org/apache/calcite/tools/PlannerTest.java
suibianwanwank: Originally this test would return:
`EnumerableValues(tuples=[[{ 1 }, { 2 }]])`
This is a little different from the plan above, it has to do with the order in which the Values are created, Values with Filters can't be merged, they are converted to unions. 
I'm not sure I need to add back this test.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
suibianwanwank: I would optimize this part, for setting  simplifyValues to false we can do this optimization, and also for that reason I keep this rule

suibianwanwank:
> Would you mind adding another test and adding a jira link? This may help new developers understand it later.

The case I mentioned in JIRA has been fixed from the root cause in [Commit](https://github.com/apache/calcite/commit/c7cace6cca8916875c6b5dc5b51089d608d97e47).
So this PR mainly optimizes union with values plan, there are already enough test in SqlToRelConverterTest, so I didn't add new test cases.

testkit/src/main/java/org/apache/calcite/test/RelMetadataFixture.java
```diff
@@ -121,6 +122,17 @@ public RelMetadataFixture withSql(String sql) {
         convertAsCalc, relTransform);
   }
 
+  public RelMetadataFixture withConfig(
+          UnaryOperator<SqlToRelConverter.Config> transform) {
+    return withFactory(f -> f.withSqlToRelConfig(transform));
+  }
+
+  /** Creates a copy of this fixture that uses a given SQL query. */
```
mihaibudiu, line 130: I don't think this JavaDoc explains properly what this function does.

mihaibudiu (approved):
I think you can squash the commits for merging.

suibianwanwank:
> I think you can squash the commits for merging.

Done

PR #4034, opened 2024-11-06
[CALCITE-6674] Make RelDecorrelator rules configurable

rubenada:
See https://issues.apache.org/jira/browse/CALCITE-6674

### DIFF
--- a/core/src/test/java/org/apache/calcite/sql2rel/RelDecorrelatorTest.java
+++ b/core/src/test/java/org/apache/calcite/sql2rel/RelDecorrelatorTest.java
@@ -17,6 +17,7 @@
 package org.apache.calcite.sql2rel;
 
 import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.plan.RelOptRule;
 import org.apache.calcite.plan.RelTraitDef;
 import org.apache.calcite.plan.hep.HepProgram;
 import org.apache.calcite.rel.RelNode;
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):
Frankly I never understood why the decorrelator is special and different from the other optimization passes. You could claim it's a pre-packaged set of optimizations.

core/src/test/java/org/apache/calcite/sql2rel/RelDecorrelatorTest.java
```diff
@@ -175,4 +180,86 @@ public static Frameworks.ConfigBuilder config() {
         + "                LogicalTableScan(table=[[scott, EMP]])\n";
     assertThat(after, hasTree(planAfter));
   }
+
+  @Test void testDecorrelatorCustomizeRules() {
```
caicancai, line 184: Add jira link

caicancai (approved):

core/src/test/java/org/apache/calcite/sql2rel/RelDecorrelatorTest.java
rubenada: Done

mihaibudiu:
Removed the "merge" label until the discussion in JIRA is settled.

rubenada:
Replaced by https://github.com/apache/calcite/pull/4042

PR #4036, opened 2024-11-07
[CALCITE-6677] HAVING clauses fail validation when type coercion is applied to GROUP BY clause

mihaibudiu:
The fix is to slightly change the validation order: 
- validate the SELECT list before the HAVING 
- perform type inference in HAVING before checking the aggregations


caicancai (approved):
LGTM

ogtracy (approved):

PR #4037, opened 2024-11-07
[CALCITE-6679] Reduced the number of type checks in VolcanoPlanner

xiaochen-zhou:
Reduced the number of type checks by moving the instanceof check outside the loop, improving performance and code clarity.

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
+++ b/core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java
@@ -421,14 +421,14 @@ public class VolcanoPlanner extends AbstractRelOptPlanner {
       return false;
     }
 
+    final boolean isTransFormRule = rule instanceof TransformationRule;
     // Each of this rule's operands is an 'entry point' for a rule call.
     // Register each operand against all concrete sub-classes that could match
     // it.
     for (RelOptRuleOperand operand : rule.getOperands()) {
       for (Class<? extends RelNode> subClass
           : subClasses(operand.getMatchedClass())) {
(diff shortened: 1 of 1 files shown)


mihaibudiu:
no need to file an issue for such small changes

NobiGo (approved):

caicancai (approved):

caicancai:
> no need to file an issue for such small changes

+1

zabetak:
> no need to file an issue for such small changes

@mihaibudiu I don't fully agree with this suggestion. It maybe be OK to skip JIRA for typos, whitespaces etc., but for actual code changes I find it useful to have JIRA ids for the reasons I expressed in https://lists.apache.org/thread/3xx1rdzgnh44ykdm8xlvgn630b60m8vb

zabetak (commented):
I guess its fine to merge this change but the benefits are not clear. The code was already readable before and the "perf gain" from moving the check will hardly make any real perf difference.

caicancai:
> I guess its fine to merge this change but the benefits are not clear. The code was already readable before and the "perf gain" from moving the check will hardly make any real perf difference.

Unfortunately, Calcite does not have a benchmark similar to benchmark

caicancai:
> > no need to file an issue for such small changes
> 
> @mihaibudiu I don't fully agree with this suggestion. It maybe be OK to skip JIRA for typos, whitespaces etc., but for actual code changes I find it useful to have JIRA ids for the reasons I expressed in https://lists.apache.org/thread/3xx1rdzgnh44ykdm8xlvgn630b60m8vb

@zabetak I think this part can be updated to the Calcite documentation to facilitate the review of subsequent PRs

PR #4039, opened 2024-11-08
[CALCITE-6680] RexImpTable erroneously declares NullPolicy.NONE for IS_EMPTY

mihaibudiu:
Based on a patch submitted in Jira by Chris Dennis
https://issues.apache.org/jira/secure/attachment/13072690/is-empty-nullable.patch

Had to tweak some other parts of Calcite, since the Java type system claimed arrays and maps cannot be null.
I inspected some other array operations as well, hopefully I didn't miss any.

### DIFF
--- a/babel/src/test/resources/sql/spark.iq
+++ b/babel/src/test/resources/sql/spark.iq
@@ -448,4 +448,223 @@ EXPR$0
 [2.111111111111112, 2.1]
 !ok
 
+# Tests for CALCITE-6680
+# RexImpTable erroneously declares NullPolicy.NONE for IS_EMPTY
+# Added as Babel tests since they need the ARRAY() and MAP() functions
+
+CREATE OR REPLACE TABLE COMPLEX(
+   KEY INT, "list" INT ARRAY, "long" BIGINT, "map" MAP<VARCHAR, INT>, "set" INT MULTISET);
+(0 rows modified)
+
+!update
+
+INSERT INTO COMPLEX (VALUES
+ (0, NULL, 5, NULL, NULL),
+ -- Apparently there is no function of constructor to create an empty multiset
+ (1, array(), 4, NULL, multiset[1] MULTISET EXCEPT multiset[1]),
+ (2, ARRAY[0], 3, MAP(), multiset[0]),
+ (3, ARRAY[0, 1], 2, MAP['zero', 0], multiset[0, 1]),
+ (4, ARRAY[0, 1, 2], 1, MAP['zero', 0, 'one', 1], multiset[0, 1, 2]),
+ (5, ARRAY[0, 1, 2, 3], 0, MAP['zero', 0, 'one', 1, 'two', 2], multiset[0, 1, 2, 3])
+);
+(6 rows modified)
+
+!update
+!set outputformat mysql
+
+SELECT * FROM COMPLEX WHERE "list" IS EMPTY;
++-----+------+------+-----+-----+
+| KEY | list | long | map | set |
++-----+------+------+-----+-----+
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -1019,7 +1019,7 @@ void populate2() {
       defineMethod(ARRAY_EXCEPT, BuiltInMethod.ARRAY_EXCEPT.method, NullPolicy.ANY);
       defineMethod(ARRAY_JOIN, BuiltInMethod.ARRAY_TO_STRING.method,
           NullPolicy.STRICT);
-      defineMethod(ARRAY_INSERT, BuiltInMethod.ARRAY_INSERT.method, NullPolicy.NONE);
+      defineMethod(ARRAY_INSERT, BuiltInMethod.ARRAY_INSERT.method, NullPolicy.ARG0);
```
caicancai, line 1022: Why is this ARG0 instead of STRICT?

mihaibudiu: Because you can insert a NULL value into an array and the result is not NULL

caicancai (approved):

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3877,13 +3883,41 @@ public static double lesser(double b0, double b1) {
...
+    if (b0 == null) {
+      return b1;
+    }
+    if (b1 == null) {
+      return b0;
+    }
+    return lt(b0, b1) ? b0 : b1;
+  }
+
+  public static @Nullable <T extends Comparable<T>> List<T> greater(
```
caicancai, line 3897: Add JavaDoc

mihaibudiu: there are at least 10 versions of this function in this file, which share one common comment.

babel/src/test/resources/sql/spark.iq
```diff
@@ -448,4 +448,177 @@ EXPR$0
 [2.111111111111112, 2.1]
 !ok
 
+# Tests for CALCITE-6680
```
caicancai, line 451: Is there any verification in Spark?

mihaibudiu: No, this is based on the spec of these functions in Calcite. I don't have access to a Spark system.
We should build a proper way to test queries on other systems, not do it manually, this is not sustainable.

caicancai: Sure

babel/src/test/resources/sql/spark.iq
```diff
@@ -448,4 +448,177 @@ EXPR$0
...
+# Added as Babel tests since they need the ARRAY() and MAP() functions
+
+CREATE OR REPLACE TABLE COMPLEX(
+   KEY INT, "list" INT ARRAY, "long" BIGINT, "map" MAP<VARCHAR, INT>, "set" INT MULTISET);
+(0 rows modified)
+
+!update
+
+INSERT INTO COMPLEX (VALUES
+ (0, NULL, 5, NULL, NULL),
```
NobiGo, line 462: Maybe we can add unit tests like 'list is null' or 'map is null'?

mihaibudiu: Added in a new commit, please check.

NobiGo (approved):
LGTM

PR #4040, opened 2024-11-08
[CALCITE-6682] Framing unsupported error message lacks LAG/LEAD functions

arkanovicz:
The query

```
select lead(n_nationkey) over (partition by n_regionkey order by n_nationkey rows between current row and 1 following) from nation
```

fails with the error message

```
 ROW/RANGE not allowed with RANK, DENSE_RANK, ROW_NUMBER or PERCENTILE_CONT/DISC function
```

whereas it should also reference the LAG/LEAD functions.


rubenada:
LGTM

caicancai:
@rubenada I think this PR was merged too quickly and many people haven’t had time to review it.

PR #4042, opened 2024-11-08
[CALCITE-6674] Make RelDecorrelator rules configurable

rubenada:
See https://issues.apache.org/jira/browse/CALCITE-6674

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
@@ -83,6 +83,7 @@ import org.apache.calcite.sql.fun.SqlSingleValueAggFunction;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.tools.RelBuilder;
 import org.apache.calcite.tools.RelBuilderFactory;
+import org.apache.calcite.tools.RuleSet;
 import org.apache.calcite.util.Holder;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -354,26 +376,54 @@ protected RelNode decorrelate(RelNode root) {
...
+            context,
+            true,
+            createCopyHook(),
+            RelOptCostImpl.FACTORY);
+    planner.setDecorrelator(this);
+    return planner;
   }
 
+  /**
+   * Tries to remove the {@link org.apache.calcite.rel.core.Correlate} instances of a query
```
mihaibudiu, line 391: I think "removes some instances of Correlate from a query plan" is a bit more precise.

mihaibudiu (approved):
I haven't checked that the decorrelator is always obtained from the planner (i.e., the replacement has been done everywhere in the codebase), but these changes look fine.

julianhyde (approved):
+1

The one change I would make is for `Planner.getDecorrelator()` to return not-null, and throw a message like "No decorrelator" if it has not been set.

caicancai (approved):

rubenada:
Thanks for the feedback. I have addressed the suggestions.

mihaibudiu:
You can probably squash the commits and merge this PR.

PR #4043, opened 2024-11-10
[CALCITE-6684] Support Arrow filter pushdown conditions that have subexpressions of a decimal type

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6684

### DIFF
--- a/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
+++ b/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
@@ -185,18 +185,23 @@ public class ArrowTable extends AbstractTable
   }
 
   private static TreeNode makeLiteralNode(String literal, String type) {
-    switch (type) {
-    case "integer":
+    if (type.startsWith("decimal")) {
+      String[] typeParts =
+          type.substring(type.indexOf('(') + 1, type.indexOf(')')).split(",");
+      int precision = parseInt(typeParts[0]);
+      int scale = parseInt(typeParts[1]);
+      return TreeBuilder.makeDecimalLiteral(literal, precision, scale);
+    } else if (type.equals("integer")) {
       return TreeBuilder.makeLiteral(parseInt(literal));
-    case "long":
+    } else if (type.equals("long")) {
       return TreeBuilder.makeLiteral(parseLong(literal));
-    case "float":
+    } else if (type.equals("float")) {
       return TreeBuilder.makeLiteral(parseFloat(literal));
-    case "double":
+    } else if (type.equals("double")) {
       return TreeBuilder.makeLiteral(parseDouble(literal));
-    case "string":
+    } else if (type.equals("string")) {
       return TreeBuilder.makeStringLiteral(literal.substring(1, literal.length() - 1));
-    default:
+    } else {
       throw new IllegalArgumentException("Invalid literal " + literal
           + ", type " + type);
     }
--- a/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
+++ b/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
@@ -31,7 +31,6 @@ import org.apache.calcite.util.DateString;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
-import java.math.BigDecimal;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.List;
@@ -194,7 +193,7 @@ class ArrowTranslator {
   private String translateOp2(String op, String name, RexLiteral right) {
     Object value = literalValue(right);
     String valueString = value.toString();
-    String valueType = getLiteralType(value);
+    String valueType = getLiteralType(right.getType());
 
     if (value instanceof String) {
       final RelDataTypeField field = requireNonNull(rowType.getField(name, true, false), "field");
@@ -234,20 +233,20 @@ class ArrowTranslator {
     return name + " " + op;
   }
 
-  private static String getLiteralType(Object literal) {
-    if (literal instanceof BigDecimal) {
-      BigDecimal bigDecimalLiteral = (BigDecimal) literal;
-      int scale = bigDecimalLiteral.scale();
-      if (scale == 0) {
-        return "integer";
-      } else if (scale > 0) {
-        return "float";
-      }
-    } else if (String.class.equals(literal.getClass())) {
-      return "string";
-    } else if (literal instanceof Double) {
+  private static String getLiteralType(RelDataType  type) {
+    if (type.getSqlTypeName() == SqlTypeName.DECIMAL) {
+      return "decimal" + "(" + type.getPrecision() + "," + type.getScale() + ")";
+    } else if (type.getSqlTypeName() == SqlTypeName.REAL) {
       return "float";
+    } else if (type.getSqlTypeName() == SqlTypeName.DOUBLE) {
+      return "double";
+    } else if (type.getSqlTypeName() == SqlTypeName.INTEGER) {
+      return "integer";
+    } else if (type.getSqlTypeName() == SqlTypeName.VARCHAR
+        || type.getSqlTypeName() == SqlTypeName.CHAR) {
+      return "string";
+    } else {
+      throw new UnsupportedOperationException("Unsupported type " + type);
     }
-    throw new UnsupportedOperationException("Unsupported literal " + literal);
   }
 }
--- a/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
+++ b/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
@@ -961,4 +961,58 @@ class ArrowAdapterTest {
         .returns(result)
         .explainContains(plan);
   }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6684">[CALCITE-6684]
+   * Arrow adapter should supports filter conditions of Decimal type</a>. */
+  @Test void testArrowProjectFieldsWithDecimalFilter() {
+    String sql = "select \"decimalField\"\n"
+        + "from arrowdatatype\n"
+        + "where \"decimalField\" = 1.00";
+    String plan = "PLAN=ArrowToEnumerableConverter\n"
+        + "  ArrowProject(decimalField=[$8])\n"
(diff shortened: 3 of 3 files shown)


arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
```diff
@@ -194,6 +194,8 @@ private static TreeNode makeLiteralNode(String literal, String type) {
       return TreeBuilder.makeLiteral(parseFloat(literal));
     case "double":
       return TreeBuilder.makeLiteral(parseDouble(literal));
+    case "decimal":
+      return TreeBuilder.makeDecimalLiteral(literal, 19, 0);
```
caicancai, line 198: Comply with Calcite's default Decimal type

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
```diff
@@ -234,8 +234,10 @@ private String translateUnaryOp(String op, String name) {
     return name + " " + op;
   }
 
-  private static String getLiteralType(Object literal) {
-    if (literal instanceof BigDecimal) {
+  private static String getLiteralType(Object literal, RelDataType  type) {
+    if (type.getSqlTypeName() == SqlTypeName.DECIMAL) {
+      return "decimal";
+    } else if (literal instanceof BigDecimal) {
```
mihaibudiu, line 240: since now you are passing type information to this function you should use it everywhere.  There is no reason to guess the type based on the Java type of the literal.

caicancai: done

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
mihaibudiu: You should use the type system to retrieve the default type, assuming it is available here.
Maybe there is a default type for Arrow, but the "default" type for Calcite is irrelevant in this context.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java
```diff
@@ -234,20 +233,20 @@ private String translateUnaryOp(String op, String name) {
...
+    } else if (type.getSqlTypeName() == SqlTypeName.REAL) {
       return "float";
+    } else if (type.getSqlTypeName() == SqlTypeName.DOUBLE) {
+      return "double";
+    } else if (type.getSqlTypeName() == SqlTypeName.INTEGER) {
+      return "integer";
+    } else if (type.getSqlTypeName() == SqlTypeName.VARCHAR
+        || type.getSqlTypeName() == SqlTypeName.CHAR) {
+      return "string";
+    } else {
```
mihaibudiu, line 248: there are many unhandled cases, but perhaps this is a good start.

caicancai: @mihaibudiu 
Unfortunately, Arrow does not have a default decimal type.
The Arrow adapter currently converts Decimal types to DECIMAL(19, 0) by default. I don't know why the previous person implemented it this way. It seems that he simply called the calcite native API without considering any precision issues, including timestamp, time and other types.

I am working on a new type system for the Arrow adapter, which seems to be a big project

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
mihaibudiu: There is no way this can be right, this drops all digits after the decimal point.
At least make 19 and 0 some constants like ARROW_DEFAULT_PRECISION and ARROW_DEFAULT_SCALE.

caicancai: done

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
```diff
@@ -130,7 +130,16 @@ public Enumerable<Object> query(DataContext root, ImmutableIntList fields,
         // if the split condition has more than two parts it's a binary operator
         // with an additional literal node
         if (data.length > 2) {
-          treeNodes.add(makeLiteralNode(data[2], data[3]));
+          if (data[3].startsWith("decimal")) {
```
caicancai, line 133: It seems that string is not a good choice at the moment. I plan to restructure it into Calcite.Sqltype type

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
```diff
@@ -197,6 +200,26 @@ private static TreeNode makeLiteralNode(String literal, String type) {
...
+  }
+
+  /**
+   * Creates a literal node for a data type with precision.
+   *
+   * @param literal the string representation of the value
+   * @param type the data type, expected to be in the format "type(precision,scale)"
+   * @return the corresponding TreeNode for the literal
+   */
+  private static TreeNode makeLiteralNode2(
```
mihaibudiu, line 214: Why pass the type as a string and then parse it rather than using RelDataType?

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowDataTypeConstants.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.adapter.arrow;
+
+/**
+ * Constants for Arrow data types.
+ */
+public interface ArrowDataTypeConstants {
+  int ARROW_DEFAULT_DECIMAL_PRECISION = 19;
```
NobiGo, line 23: Is the decimal precision an ARROW default value or a user-defined variable? As far as I know, when the RelNode is in the Adapter, we should already have a precision and Scale in Decimal Type.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
```diff
@@ -185,18 +188,24 @@ private static RelDataType deduceRowType(Schema schema,
   }
 
   private static TreeNode makeLiteralNode(String literal, String type) {
-    switch (type) {
-    case "integer":
+    if (type.startsWith("decimal")) {
```
NobiGo, line 191: If we have RelDataType, then we can switch by SqlTypeName and get the precision and scale by RelDataType.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowDataTypeConstants.java
caicancai: You are right, the decimal of arrow must specify the precision

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
caicancai: It looks a bit cumbersome at the moment, and it's almost like refactoring the filter structure. Another reason is that the parameter of arrow's makeLiteralNode api must be a string, which means I need another layer of conversion, but you can try

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowDataTypeConstans.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.adapter.arrow;
+
+/**
+ * Constants for Arrow data types.
+ */
+public interface ArrowDataTypeConstans {
+  int ARROW_DEFAULT_PRECISION = 19;
```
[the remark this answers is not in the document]
caicancai: remove. According to the decimal type rules of arrow, scale and precision must be specified

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
caicancai: It may be a problem of personal coding habits. I have unified it in one function.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTable.java
caicancai: @NobiGo I have created a Jira case, but I am very busy with work recently and I think I will have to wait until later to start working on this.
https://issues.apache.org/jira/browse/CALCITE-6701

mihaibudiu: this is a private function, why does the parameter have to be a string?

caicancai: @mihaibudiu @NobiGo Hi !
In the `ArrowTranslator `class, we will translate the corresponding filter conditions according to the logical plan and return a string list, such as `decimalField equal 1 decimal(19,0)`. Then I need to parse this string array and execute the corresponding arrow API in arrowtable.
Since the parameters I received in ArrowTable are already string arrays, I can only process strings, which is why makeLiteralNode is all strings.

mihaibudiu: This is a strange design. Turns out that the table holds a program represented as a text that contains the pushdown filters and projections.

caicancai: Yes, I don’t know why I wrote it like this in the beginning. I am thinking about how to optimize this part. It seems to be a very big change.

caicancai:
@mihaibudiu @NobiGo Hello, do you have any other comments?

mihaibudiu (approved):

caicancai:
@mihaibudiu @NobiGo Thanks for your review, squashed

mihaibudiu:
Sorry, one more request: the issue as phrased is incorrect. Conditions do not have decimal type.
What happens here is that you want to support Arrow filter pushdown conditions that have subexpressions of a decimal type. I would change the JIRA title, the commit and PR title to say this.

caicancai:
> Sorry, one more request: the issue as phrased is incorrect. Conditions do not have decimal type. What happens here is that you want to support Arrow filter pushdown conditions that have subexpressions of a decimal type. I would change the JIRA title, the commit and PR title to say this.

done, thanks

PR #4044, opened 2024-11-12
[CALCITE-6685] Support checked arithmetic

mihaibudiu:
This implementation adds support for checked arithmetic in Calcite.
This solves some bugs which are at least 12 years old in Calcite, such as FNL_25.
This implementation only handles pure arithmetic. Aggregates are not handled currently, but using the existing framework it should become very easy.

The core idea is as follows:
- add a set of checked arithmetic operators to StdSqlOperatorTable. There are five: plus, minus, divide, times, and unary minus.
- add a new boolean property to the SqlConformance API: checkedArithmetic().
- if this property is true, the Prepare class will use a RelShuttle/RexShuttle to replace all arithmetic operations with their checked counterparts.

The handling of Decimals is a bit different: there is not checked arithmetic for decimals, but casts to decimals perform overflow checking. As such, casts to decimals may not be elided from the program even if they seem noops (the input type matches the output type).


mihaibudiu:
This design is based on a discussion on the mailing list. Many thanks to all the people who helped drive this design.

caicancai:
> This design is based on a discussion on the mailing list. Many thanks to all the people who helped drive this design.

Looks really cool. Could you please post the link to the discussion on the mailing list so people can know the whole story?

mihaibudiu:
This is the email thread: https://lists.apache.org/list?dev@calcite.apache.org:2024-10:checked

caicancai:
> This is the email thread: https://lists.apache.org/list?dev@calcite.apache.org:2024-10:checked

Thanks. I'm busy with work recently, I will review this PR on the weekend

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -3419,8 +3423,10 @@ static void checkOverlaps(OverlapChecker c) {
   }
 
   @Test void testMultiplyOperator() {
-    final SqlOperatorFixture f = fixture();
-    f.setFor(SqlStdOperatorTable.MULTIPLY, VmName.EXPAND);
+    final SqlOperatorFixture f = fixture()
+        .setFor(SqlStdOperatorTable.MULTIPLY, VmName.EXPAND)
+        // BigQuery requires arithmetic overflows
```
caicancai, line 3428: just bigquery?

caicancai:
I have a question, will this pr check for numerical overflow?

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -294,6 +294,19 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
           InferTypes.FIRST_KNOWN,
           OperandTypes.DIVISION_OPERATOR);
 
+  /**
+   * Checked version of arithmetic division operator, '<code>/</code>'.
+   */
+  public static final SqlBinaryOperator CHECKED_DIVIDE =
```
caicancai, line 300: I have some concerns about division. Does Calcite have a fixed strategy for division with a certain number of decimal places?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -3258,8 +3264,10 @@ static void checkOverlaps(OverlapChecker c) {
   }
 
   @Test void testMinusOperator() {
-    final SqlOperatorFixture f = fixture();
-    f.setFor(SqlStdOperatorTable.MINUS, VmName.EXPAND);
+    SqlOperatorFixture f = fixture()
```
caicancai, line 3267: I'm not sure if it only affects these operators, but it looks like it can be expanded later and the problem is not big

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
caicancai: Does integer division mean that the result is of integer type by default?

caicancai (approved):
LGTM

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2459,6 +2493,40 @@ public static int minus(int b0, int b1) {
...
+  public static long checkedMinus(long b0, long b1) {
+    return Math.subtractExact(b0, b1);
+  }
+
+  public static byte checkedUnaryMinus(byte b) {
+    return intToByte(-b);
+  }
+
+  public static short checkedUnaryMinus(short b) {
+    return intToShort(b);
```
suibianwanwank, line 2519: Why not `intToShort(-b)` here?

suibianwanwank: I mean here it should be `intToShort(-b)` not `intToShort(b)`

mihaibudiu: I have pushed a new commit with a fix and a test.

suibianwanwank (approved):

mihaibudiu:
I marked this as ready to merge, but I will wait a couple of days for more comments, since it's an important change.

mihaibudiu:
I have added the change in Conformance as a breaking change to the history.md file.
I have squashed the commits in preparation for merging, but will wait a bit more for comments.

PR #4045, opened 2024-11-13
Modify chinese in DAYNAME and MONTHNAME doc to english to keep doc consistent

strongduanmu:
- Currently, the chinese characters `星期日` and `二月` are used in the DAYNAME and MONTHNAME document examples. Although these two results are legal, considering that the document is written in english, it will provide a more consistent reading experience if `Sunday` and `February` can be used.


mihaibudiu (commented):
Frankly, I think it's useful to have non-English locale examples and tests.
Perhaps you can keep the Chinese one and add the English one?

strongduanmu:
> Frankly, I think it's useful to have non-English locale examples and tests. Perhaps you can keep the Chinese one and add the English one?

This is a good idea, I'll add English instructions and describe how to get different results with the local parameter.

site/_docs/reference.md
```diff
@@ -2785,7 +2785,7 @@ In the following:
 | * | CSC(numeric)                                   | Returns the cosecant of *numeric* in radians
 | * | CSCH(numeric)                                  | Returns the hyperbolic cosecant of *numeric*
 | b | CURRENT_DATETIME([ timeZone ])                 | Returns the current time as a TIMESTAMP from *timezone*
-| m | DAYNAME(datetime)                              | Returns the name, in the connection's locale, of the weekday in *datetime*; for example, it returns '星期日' for both DATE '2020-02-10' and TIMESTAMP '2020-02-10 10:10:10'
+| m | DAYNAME(datetime)                              | Returns the name, in the connection's locale, of the weekday in *datetime*; for example, if you config locale to en, it will return 'Sunday' for both DATE '2020-02-10' and TIMESTAMP '2020-02-10 10:10:10', and if you config locale to zh, it will return '星期日'
```
mihaibudiu, line 2788: I think an impersonal sentence is preferrable, i.e., without "you"
"For a locale of en it will return 'Sunday'" etc.

strongduanmu: Thanks for your suggestion, I have modified the description.

PR #4050, opened 2024-11-16
[CALCITE-6693] Add Source SQL Dialect to RelToSqlConverterTest

hannerwang:
### What changes were proposed in this pull request?
Add Source SQL Dialect to RelToSqlConverterTest to make dialect translation more clear.

### Why are the changes needed?
Currently, RelToSqlConverterTest converts the original SQL to RelNode using the default validator config and the target dialect's type system, which is confusing. We should ensure the conversion conforms to the source dialect's specifications, and then convert from the source dialect's RelNode to the target dialect's SQL.

### Does this PR introduce any user-facing change?
Yes, the null collation the PlannerImpl use will be validator config's null collation when calcite connection property unset.

### DIFF
--- a/core/src/main/java/org/apache/calcite/prepare/PlannerImpl.java
+++ b/core/src/main/java/org/apache/calcite/prepare/PlannerImpl.java
@@ -46,6 +46,7 @@ import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlOperatorTable;
 import org.apache.calcite.sql.parser.SqlParseException;
 import org.apache.calcite.sql.parser.SqlParser;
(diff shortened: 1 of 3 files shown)


caicancai:
Please fix CLI

hannerwang:
> Please fix CLI

Thanks for the reminder, it's fixed.

mihaibudiu (approved):
I am approving this, let's see if anyone has any other comments.

caicancai (approved):

julianhyde:
-1 let's discuss in jira

mihaibudiu:
removed label until the discussion in JIRA is settled.

PR #4052, opened 2024-11-17
[CALCITE-6690] Refactor the Arrow adapter type system

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6690

### DIFF
--- a/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
+++ b/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
@@ -25,7 +25,6 @@ import org.apache.arrow.vector.types.pojo.ArrowType;
 
 import java.math.BigDecimal;
 import java.sql.Date;
-import java.util.List;
 
 import static java.util.Objects.requireNonNull;
 
@@ -39,13 +38,14 @@ enum ArrowFieldType {
   FLOAT(Primitive.FLOAT),
   DOUBLE(Primitive.DOUBLE),
   DATE(Date.class),
-  LIST(List.class),
   DECIMAL(BigDecimal.class),
   LONG(Primitive.LONG),
   BYTE(Primitive.BYTE),
   SHORT(Primitive.SHORT);
 
   private final Class<?> clazz;
+  private int precision;
+  private int scale;
 
   ArrowFieldType(Primitive primitive) {
     this(requireNonNull(primitive.boxClass, "boxClass"));
@@ -55,9 +55,25 @@ enum ArrowFieldType {
     this.clazz = clazz;
   }
 
+  ArrowFieldType(Class<?> clazz, int precision, int scale) {
+    this.clazz = clazz;
+    this.precision = precision;
+    this.scale = scale;
+  }
+
+  public void setPrecisionAndScale(int precision, int scale) {
+    this.precision = precision;
+    this.scale = scale;
+  }
+
   public RelDataType toType(JavaTypeFactory typeFactory) {
     RelDataType javaType = typeFactory.createJavaType(clazz);
-    RelDataType sqlType = typeFactory.createSqlType(javaType.getSqlTypeName());
+    RelDataType sqlType = null;
+    if (javaType.getSqlTypeName().getName().equals("DECIMAL")) {
+      sqlType = typeFactory.createSqlType(javaType.getSqlTypeName(), precision, scale);
+    } else {
+      sqlType = typeFactory.createSqlType(javaType.getSqlTypeName());
+    }
     return typeFactory.createTypeWithNullability(sqlType, true);
   }
 
@@ -94,6 +110,9 @@ enum ArrowFieldType {
     case Date:
       return DATE;
     case Decimal:
+      DECIMAL.setPrecisionAndScale(
+          ((ArrowType.Decimal) arrowType).getPrecision(),
+          ((ArrowType.Decimal) arrowType).getScale());
       return DECIMAL;
     default:
       throw new IllegalArgumentException("Unsupported type: " + arrowType);
--- a/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
+++ b/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
@@ -55,6 +55,12 @@ enum ArrowFieldType {
     this.clazz = clazz;
   }
 
+  ArrowFieldType(Class<?> clazz, int precision, int scale) {
+    this.clazz = clazz;
+    this.precision = precision;
+    this.scale = scale;
+  }
+
   public void setPrecisionAndScale(int precision, int scale) {
     this.precision = precision;
     this.scale = scale;
@@ -104,8 +110,7 @@ enum ArrowFieldType {
     case Date:
       return DATE;
     case Decimal:
-      DECIMAL.setPrecisionAndScale(
-          ((ArrowType.Decimal) arrowType).getPrecision(),
+      DECIMAL.setPrecisionAndScale(((ArrowType.Decimal) arrowType).getPrecision(),
           ((ArrowType.Decimal) arrowType).getScale());
       return DECIMAL;
     default:
--- a/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
+++ b/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
@@ -55,12 +55,6 @@ enum ArrowFieldType {
     this.clazz = clazz;
   }
 
-  ArrowFieldType(Class<?> clazz, int precision, int scale) {
-    this.clazz = clazz;
-    this.precision = precision;
-    this.scale = scale;
-  }
-
   public void setPrecisionAndScale(int precision, int scale) {
     this.precision = precision;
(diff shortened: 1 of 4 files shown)


caicancai:
After this pr, for some data types, we can easily identify their precision

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
```diff
@@ -25,27 +25,17 @@
...
-  STRING(String.class),
-  FLOAT(Primitive.FLOAT),
-  DOUBLE(Primitive.DOUBLE),
-  DATE(Date.class),
-  LIST(List.class),
-  DECIMAL(BigDecimal.class),
-  LONG(Primitive.LONG),
-  BYTE(Primitive.BYTE),
-  SHORT(Primitive.SHORT);
+class ArrowFieldType {
```
mihaibudiu, line 34: I think enums are preferable, since they are typechecked by the compiler.
A Class can be anything.

caicancai: Later, I may dynamically pass some parameters according to the data type of arrow, which is difficult to achieve through enumeration.

mihaibudiu: Why later if you are refactoring the code now?
Do it correctly the first time.
I am suggesting to have 3 arguments for the enum constructor: Class, precision, scale.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowData.java
```diff
@@ -237,35 +239,38 @@ public void writeArrowDataType(File file) throws IOException {
       vectorSchemaRoot.setRowCount(numRows);
       for (Field field : vectorSchemaRoot.getSchema().getFields()) {
         FieldVector vector = vectorSchemaRoot.getVector(field.getName());
-        switch (vector.getMinorType()) {
-        case TINYINT:
+        switch (field.getName()) {
+        case "tinyIntField":
```
NobiGo, line 243: This is a necessary change? What is the difference between decimalField2 and decimalField?

caicancai: Required, the scale of decimalfield and decimalfiled2 are different, one is 2, the other is 3, the main purpose is to capture the accuracy of whether it is effective

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
caicancai: @mihaibudiu Sorry for the late reply, you are right. I looked at the arrow types and it seems that Class, precision, and scale are sufficient.

mihaibudiu: I am not sure I understand the question.
There are lots of enum classes with parameters in the codebase.
See https://github.com/apache/calcite/blob/main/linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java for example

caicancai: @mihaibudiu thanks, done

caicancai:
Sorry for confusing the commit message. I'm not sure if it was my changes that caused the cli to fail. I haven't reproduced the problem locally.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
caicancai: The strange thing is that when I don't use an enumeration to pass in the precision, the CLI will not report an error

mihaibudiu: What is the CLI?

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
```diff
@@ -55,9 +45,20 @@ enum ArrowFieldType {
     this.clazz = clazz;
   }
 
+  ArrowFieldType(Class<?> clazz, int precision, int scale) {
```
mihaibudiu, line 48: I think the other constructors should call this one.
There are some values in the type system for "unspecified", maybe these could be reused.
Using 0 may not be always safe.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
```diff
@@ -732,8 +732,8 @@ static void initializeArrowState(@TempDir Path sharedTempDir)
   @Test void testFilteredAgg() {
     String sql = "select SUM(SAL) FILTER (WHERE COMM > 400) as SALESSUM from EMP";
     String plan = "PLAN=EnumerableAggregate(group=[{}], SALESSUM=[SUM($0) FILTER $1])\n"
-        + "  EnumerableCalc(expr#0..7=[{inputs}], expr#8=[400:DECIMAL(19, 0)], expr#9=[>($t6, $t8)], "
-        + "expr#10=[IS TRUE($t9)], SAL=[$t5], $f1=[$t10])\n"
+        + "  EnumerableCalc(expr#0..7=[{inputs}], expr#8=[CAST($t6):DECIMAL(10, 2)], expr#9=[400.00:DECIMAL(10, 2)], "
```
mihaibudiu, line 735: can you also add or modify a test that uses a decimal number with digits after the decimal point? This would exercise the scale.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
caicancai: > What is the CLI?

Sorry. it is CI

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowAdapterTest.java
caicancai: This is a table that comes with calcite and is not filled in the arrow adapter. If you modify it, it will affect many tests.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
caicancai: I'm using an enum class to represent this, but it doesn't seem ideal, CI always fails.
https://github.com/apache/calcite/actions/runs/12014265742/job/33489690898

mihaibudiu: Have you tried to apply the recommendations made in the output of the failed CI run?
```
/home/runner/work/calcite/calcite/core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java:2654: Note: [MethodCanBeStatic] A private method that does not reference the enclosing instance can be static
    private void onMatch2(
    ^
    (see https://errorprone.info/bugpattern/MethodCanBeStatic)
  Did you mean 'private static void onMatch2('?
/home/runner/work/calcite/calcite/core/src/main/java/org/apache/calcite/util/format/postgresql/format/compiled/NumberCompiledPattern.java:82: Note: [MethodCanBeStatic] A private method that does not reference the enclosing instance can be static
  private String getOrdinalSuffix(String stringValue) {
  ^
    (see https://errorprone.info/bugpattern/MethodCanBeStatic)
  Did you mean 'private static String getOrdinalSuffix(String stringValue) {'?
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :arrow:compileJava
/home/runner/work/calcite/calcite/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowTranslator.java:233: Note: [MethodCanBeStatic] A private method that does not reference the enclosing instance can be static
  private String translateUnaryOp(String op, String name) {
  ^
    (see https://errorprone.info/bugpattern/MethodCanBeStatic)
  Did you mean 'private static String translateUnaryOp(String op, String name) {'?
/home/runner/work/calcite/calcite/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java:47: warning: [ImmutableEnumChecker] enums should be immutable: 'ArrowFieldType' has non-final field 'precision'
  private int precision;
              ^
    (see https://errorprone.info/bugpattern/ImmutableEnumChecker)
  Did you mean 'private final int precision;'?
/home/runner/work/calcite/calcite/arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java:48: warning: [ImmutableEnumChecker] enums should be immutable: 'ArrowFieldType' has non-final field 'scale'
  private int scale;
              ^
    (see https://errorprone.info/bugpattern/ImmutableEnumChecker)
  Did you mean 'private final int scale;'?
```

caicancai:
I increasingly feel that the ArrowFieldType design is a bit complicated. In fact, it is unnecessary. We can directly convert Arrowtype to RelDataType. I don't understand why there is an extra layer of conversion.

I will refactor this class next weekend to make ArrowFieldType align directly with RelDataType.

cc @mihaibudiu

PR #4054, opened 2024-11-18
[CALCITE-6698] Add Javadoc to PartiallyOrderedSet#getNonChildren and getNonParents to clarify their behavior

zabetak:
1. Add Javadoc to clarify the results of each method and avoid potential misinterpretation of their name.
2. Add some test cases with explicit tests for the aforementioned methods.

### DIFF
--- a/core/src/test/java/org/apache/calcite/util/PartiallyOrderedSetTest.java
+++ b/core/src/test/java/org/apache/calcite/util/PartiallyOrderedSetTest.java
@@ -20,6 +20,7 @@ import org.junit.jupiter.api.Test;
 
 import java.util.AbstractList;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/util/PartiallyOrderedSetTest.java
```diff
@@ -180,6 +181,32 @@ private static boolean isBitSuperset(Integer e1, Integer e2) {
...
+    PartiallyOrderedSet<Integer> poset =
+        new PartiallyOrderedSet<>((i, j) -> i >= j, Arrays.asList(20, 30, 40));
+    assertThat(poset.getNonParents(), hasToString("[20]"));
+  }
+
+  @Test void testGetNonChildrenOnGteIntPosetReturnsMaxValue() {
+    PartiallyOrderedSet<Integer> poset =
+        new PartiallyOrderedSet<>((i, j) -> i >= j, Arrays.asList(20, 30, 40));
+    printValidate(poset);
+    System.out.println(poset.getChildren(20));
```
asolimando, line 206: This has slipped through the cracks I guess

julianhyde:
-1 There is no bug here. You are just mistaken what 'non-parents' means.

core/src/test/java/org/apache/calcite/util/PartiallyOrderedSetTest.java
zabetak: Thanks for catching this. I removed those in the last commit.

zabetak:
As discussed, under the ticket there is no reason to change the behavior of the methods. The scope of the PR changed to simply clarify the current behavior by adding Javadoc comments so I updated the summary and description.

PR #4055, opened 2024-11-19
[CALCITE-6699] Invalid unparse for Varchar in StarRocksDialect

xiaochen-zhou:
https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6699

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
@@ -121,6 +121,10 @@ public class StarRocksSqlDialect extends MysqlSqlDialect {
       return new SqlDataTypeSpec(
           new SqlBasicTypeNameSpec(SqlTypeName.BIGINT, SqlParserPos.ZERO),
           SqlParserPos.ZERO);
+    case VARCHAR:
+      return new SqlDataTypeSpec(
+          new SqlBasicTypeNameSpec(SqlTypeName.VARCHAR, SqlParserPos.ZERO),
+          SqlParserPos.ZERO);
     case TIMESTAMP:
       return new SqlDataTypeSpec(
           new SqlAlienSystemTypeNameSpec(


core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -121,6 +121,10 @@ public StarRocksSqlDialect(Context context) {
       return new SqlDataTypeSpec(
           new SqlBasicTypeNameSpec(SqlTypeName.BIGINT, SqlParserPos.ZERO),
           SqlParserPos.ZERO);
+    case VARCHAR:
```
NobiGo, line 124: We need to add a unit test.

NobiGo (changes requested):

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
mihaibudiu: what happens to the precision if it is specified?

caicancai: +1. The accuracy issue is not considered here.

xiaochen-zhou: > test

Sorry for the late reply. I will add it later today.

xiaochen-zhou: > what happens to the precision if it is specified?

I initially thought that if precision is not set here, the default precision for varchar might be used. However, I just looked at the code, and it seems that if precision is not set, the default is PRECISION_NOT_SPECIFIED. Could this lead to a loss of precision, or did I not look at the code carefully?

@mihaibudiu @caicancai

NobiGo: You can try to follow the startRocks behavior. Does Starrocks support Varchar without setting precision? If so, add test cases to override it. If not, set the default varchar precision to Starrocks' default value. Also, add unit test to cover when Varchar setting precision.

suibianwanwank: I tried cast varchar with precision in Starrocks, and it seems that the precision doesn't affect the result and type(If the precision does not exceed 65533).

caicancai: > I tried cast varchar with precision in Starrocks, and it seems that the precision doesn't affect the result and type(If the precision does not exceed 65533).

Although this is not a common situation, we should still consider it.

mihaibudiu: VARCHAR(N) is supposed to truncate to N characters, and then trim the whitespaces from the end of the string.
If the dialect does not support VARCHAR(N) perhaps it should report an error when an attempt is made to use it.
Then you should also add an assertion that the precision is "not specified".

PR #4056, opened 2024-11-20
[CALCITE-6678] Calcite should support dual table query when db provides this feature

strongduanmu:
* Use view table to implement dual table query for better optimize
* Move dual table query test case to new `dual-table-query.iq` file

### DIFF
--- a/core/src/test/resources/sql/dummy.iq
+++ b/core/src/test/resources/sql/dummy.iq
@@ -20,58 +20,5 @@ values 1;
 EXPR$0
 1
 !ok
-
-# [CALCITE-6678] Support dual table query (enabled in MySQL, Oracle libraries)
-!set outputformat mysql
-!use scott-mysql
-
-# MySQL supports users to specify the dual table, and also supports users not to specify the dual table.
-SELECT 1 + 1 FROM DUAL;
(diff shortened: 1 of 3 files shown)


NobiGo (approved):

core/src/test/resources/sql/dummy.iq
```diff
@@ -20,58 +20,4 @@ values 1;
 EXPR$0
 1
 !ok
-
```
mihaibudiu, line 23: Can you please add a comment here saying "do not modify this file, this is intended only for temporary debugging code"?

strongduanmu: Good suggestion, I've added a comment.

caicancai (approved):

strongduanmu:
Hi @mihaibudiu @julianhyde, could you please review this PR? I have made the changes as suggested in Jira. Thank you very much.

mihaibudiu:
You already have an approval, the main question was whether @julianhyde is happy, he raised objections to the previous implementation. If he does not object, I will merge this in the next few days.

mihaibudiu:
you should also squash the commits for merging.

PR #4057, opened 2024-11-20
[CALCITE-4758] When SOME sub-query is SqlNodeList, Calcite returns wr…

NobiGo:
…ong result


NobiGo:
Before this PR, When the subquery is SqlNodeList, this subquery will be a special case for `config.isExpand`, After this PR, we convert the SqlNodeList to a RexSubquery, then can handled by SubqueryRemoveRule.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -6007,6 +6012,26 @@ public List<SqlMonotonicity> getColumnMonotonicities() {
     }
   }
 
+  private RexNode getSubqueryRexNode(SqlKind kind,
```
mihaibudiu, line 6015: I find the name of this function not very easy to understand, perhaps adding some JavaDoc will help.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -3148,9 +3148,7 @@ private void checkPushJoinThroughUnionOnRightDoesNotMatchSemiOrAntiJoin(JoinRelT
         + "from t1\n"
         + "where (t1.a, t1.y) in ((1, 2), (3, null), (7369, null), (7499, 30), (null, 20), (null, 5))";
     sql(sql)
-        .withRelBuilderConfig(b -> b.withSimplifyValues(false))
-        .withRule(CoreRules.PROJECT_VALUES_MERGE,
-            CoreRules.UNION_TO_VALUES)
+        .withRule(CoreRules.FILTER_SUB_QUERY_TO_CORRELATE)
```
mihaibudiu, line 3151: are these tests testing the same thing as the original tests?
if not, perhaps you should add new tests rather than modifying the original ones

core/src/test/resources/sql/some.iq
```diff
@@ -847,5 +847,92 @@ where sal > all (select comm from "scott".emp where comm is not null);
 
 !ok
 
+# [CALCITE-4758] When SOME sub-query is SqlNodeList, Calcite returns wrong result
+
+# make sure the SqlNodeList converted to VALUES not OR condition
+
```
mihaibudiu, line 853: this empty line makes it less clear that the comment applies to the threshold setting

core/src/test/resources/sql/sub-query.iq
```diff
@@ -3448,13 +3448,13 @@ select * from "scott".emp where empno not in (null, 7782);
 
 !ok
 
-EnumerableCalc(expr#0..12=[{inputs}], expr#13=[0:BIGINT], expr#14=[=($t8, $t13)], expr#15=[IS NULL($t12)], expr#16=[>=($t9, $t8)], expr#17=[AND($t15, $t16)], expr#18=[OR($t14, $t17)], proj#0..7=[{exprs}], $condition=[$t18])
+EnumerableCalc(expr#0..12=[{inputs}], expr#13=[0], expr#14=[=($t8, $t13)], expr#15=[IS NULL($t12)], expr#16=[>=($t9, $t8)], expr#17=[AND($t15, $t16)], expr#18=[OR($t14, $t17)], proj#0..7=[{exprs}], $condition=[$t18])
```
mihaibudiu, line 3451: the disappearance of the cast of 0 may be a hint that the typing is wrong someplace - either the original plan or the new plan must be wrong.

core/src/test/resources/sql/sub-query.iq
```diff
@@ -3936,15 +3934,15 @@ select comm, comm in (500, 300, 0) from emp;
 
 !ok
 
-EnumerableCalc(expr#0..5=[{inputs}], expr#6=[IS NOT NULL($t5)], expr#7=[0:BIGINT], expr#8=[<>($t1, $t7)], expr#9=[AND($t6, $t8)], expr#10=[IS NULL($t3)], expr#11=[<($t2, $t1)], expr#12=[OR($t10, $t11)], expr#13=[null:BOOLEAN], expr#14=[IS NULL($t5)], expr#15=[AND($t12, $t13, $t8, $t14)], expr#16=[OR($t9, $t15)], COMM=[$t0], EXPR$1=[$t16])
-  EnumerableMergeJoin(condition=[=($3, $4)], joinType=[left])
-    EnumerableSort(sort0=[$3], dir0=[ASC])
-      EnumerableCalc(expr#0..3=[{inputs}], expr#4=[CAST($t1):DECIMAL(10, 2)], COMM=[$t1], $f0=[$t2], $f1=[$t3], COMM0=[$t4])
+EnumerableCalc(expr#0..6=[{inputs}], expr#7=[0], expr#8=[=($t2, $t7)], expr#9=[false], expr#10=[CAST($t1):DECIMAL(10, 2)], expr#11=[IS NULL($t10)], expr#12=[null:BOOLEAN], expr#13=[IS NOT NULL($t6)], expr#14=[true], expr#15=[<($t3, $t2)], expr#16=[CASE($t8, $t9, $t11, $t12, $t13, $t14, $t15, $t12, $t9)], COMM=[$t1], EXPR$1=[$t16])
```
mihaibudiu, line 3937: I cannot review these changes, they are too complex.

mihaibudiu (commented):
Has the behavior of isExpand changed with this PR (other than producing a correct result for queries that were wrong)?

NobiGo:
> Has the behavior of isExpand changed with this PR (other than producing a correct result for queries that were wrong)

`isExpand` only changed the behavior when the subquery is a SqlNodeList. Others stay the same.

mihaibudiu:
This does not explain what kinds of SQL program this affects.
SqlNodeList is a very general data structure.
Can this only happen when there is a list of expressions?
Can the expressions refer to correlated variables, for example?

NobiGo:
I think this is a reasonable approach and there is no need to maintain the logic when expand is false. But it is incompatible with previous versions.

mihaibudiu:
At the very least I believe breaking changes have to be documented in history.md.
But first I would like to understand what is actually affected. Could you make a short summary of programs that will compile differently after this change? Seems like the JIRA issue is the right place.
If the only change is to compile correctly programs that were previously causing a compiler bug or an incorrect output, then this is not a breaking change.

NobiGo:
> At the very least I believe breaking changes have to be documented in history.md. But first I would like to understand what is actually affected. Could you make a short summary of programs that will compile differently after this change? Seems like the JIRA issue is the right place. If the only change is to compile correctly programs that were previously causing a compiler bug or an incorrect output, then this is not a breaking change.

Add description in Jira.

mihaibudiu:
So the JIRA issue says that some programs were incorrectly compiled: with SOME and ALL subqueries.
Does this PR fix these cases?

mihaibudiu:
More precisely, you seem to convert the SOME query into a different query, that has a VALUES subquery.
Why not just reject it?

NobiGo:
> So the JIRA issue says that some programs were incorrectly compiled: with SOME and ALL subqueries.
> Does this PR fix these cases?

This PR can solve all the problems described in the description.

mihaibudiu:
Does it solve them by rejecting the illegal queries or by rewriting them into a different query that is legal?

NobiGo:
> Does it solve them by rejecting the illegal queries or by rewriting them into a different query that is legal?

We can't reject this query because we have supported it. https://github.com/apache/calcite/blob/main/core/src/test/resources/sql/some.iq#L515
It returns the right result because It converted to OR condition.` (the SqlNodeList's size is less than config.getInSubQueryThreshold())`.
But if we set `config.getInSubQueryThreshold()` less than the SqlNodeList's size. The Sql will return the wrong answer.

This PR resolved it by converting SqlNodeList to Values is legal? I think it is legal. The results of these two SQL statements are actually consistent, or in other words, the expression of Values is a more standard way of writing SQL NodeList subqueries. WDYT？

core/src/test/resources/sql/sub-query.iq
mihaibudiu: I think this indicates that there is a bug somewhere, but it's probably not directly related to these changes, it's just unmasked by these changes.

mihaibudiu (approved):

core/src/test/resources/sql/sub-query.iq
NobiGo: > I think this indicates that there is a bug somewhere, but it's probably not directly related to these changes, it's just unmasked by these changes.

Thanks for the review. I will check this again to make sure we get the right plan.

NobiGo: Hi @mihaibudiu, I check the code. These two places build the literal 0 with different methods,
In SqlToRel:
```
final RelDataType longType =typeFactory.createSqlType(SqlTypeName.BIGINT);
final RexLiteral zero =rexBuilder.makeExactLiteral(BigDecimal.ZERO, longType);
```
In SubqueryRemoveRule:
```
builder.literal(0)
```
Because need to compare with the value of COUNT, the first one is more reasonable.

mihaibudiu: yes, the second one seems wrong, since it does not set a type for the zero.

NobiGo: I will try to find an SQL test case to test if this is a bug.

NobiGo:

PR #4058, opened 2024-11-21
[CALCITE-5590] NullPointerException when converting 'in' expression that is used inside select list and group by

dssysolyatin:
Note that almost all deleted lines are from my incorrect previous fix for task [CALCITE-5209](https://issues.apache.org/jira/browse/CALCITE-5209)

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
@@ -523,6 +523,16 @@ class SqlToRelConverterTest extends SqlToRelTestBase {
     sql(sql).ok();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-4549">[CALCITE-4549]
+   * IndexOutOfBoundsException when group view by a sub query</a>. */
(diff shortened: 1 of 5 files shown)


mihaibudiu (approved):
This looks a bit like magic, but it seems to work based on the tests.

caicancai (approved):
LGTM

dssysolyatin:
@mihaibudiu @caicancai Thanks for reviewing. I made one more small change—I added a test case from the task https://issues.apache.org/jira/browse/CALCITE-4549 to ensure that it also works.

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -523,6 +523,16 @@ public static void checkActualAndReferenceFiles() {
     sql(sql).ok();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-4549">[CALCITE-4549]
+   * IndexOutOfBoundsException when group view by a sub query</a>. */
+  @Test void testGroupView() {
```
caicancai, line 529: Maybe the test name could be improved, but it doesn't seem to matter.

PR #4059, opened 2024-11-22
[CALCITE-6702] Strong Policy for the POWER Function is wrong

ninidhiaeddine:
CALCITE-6702: Strong Policy for the `SqlStdOperatorTable.POWER` Function is wrongly assigned `AS_IS` when it should be `ANY`.

**Context:**
In various SQL db implementations, the `POWER(a, b)` function returns `NULL` when either or both operands are null.
```
SELECT POWER(NULL, 2) AS "ALIAS1"; -- => Returns NULL
SELECT POWER(2, NULL) AS "ALIAS2"; -- => Returns NULL
SELECT POWER(NULL ,NULL) AS "ALIAS3"; -- => Returns NULL
```

These are all the cases I could think of:
image

This means that the correct null policy for the `POWER` function is `ANY` instead of `AS_IS`, since this expression is `NULL` **if and only if** at least one of its arguments is `NULL`.

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/Strong.java
+++ b/core/src/main/java/org/apache/calcite/plan/Strong.java
@@ -336,6 +336,7 @@ public class Strong {
     map.put(SqlKind.TIMESTAMP_ADD, Policy.ANY);
     map.put(SqlKind.TIMESTAMP_DIFF, Policy.ANY);
     map.put(SqlKind.ITEM, Policy.ANY);
+    map.put(SqlKind.POWER, Policy.ANY);
 
     // Assume that any other expressions cannot be simplified.
     for (SqlKind k
--- a/core/src/main/java/org/apache/calcite/sql/SqlBasicFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlBasicFunction.java
@@ -118,6 +118,17 @@ public final class SqlBasicFunction extends SqlFunction {
         SqlFunctionCategory.SYSTEM, call -> SqlMonotonicity.NOT_MONOTONIC, false);
   }
 
+  /** Creates a {@code SqlBasicFunction} whose name is the same as its kind. */
+  public static SqlBasicFunction create(SqlKind kind,
+      SqlReturnTypeInference returnTypeInference,
+      SqlOperandTypeChecker operandTypeChecker,
+      SqlFunctionCategory category) {
+    return new SqlBasicFunction(kind.name(), kind,
+        SqlSyntax.FUNCTION, true, returnTypeInference, null,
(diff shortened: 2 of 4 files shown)


NobiGo:
@ninidhiaeddine We need a unit test for this change.

ninidhiaeddine:
> @ninidhiaeddine We need a unit test for this change.

I was thinking to add a unit test for this change, but I didn't find any existing unit tests that are checking the returned `Strong.Policy` for a given `SqlOperator`.

The only unit tests that I found to be remotely related are: `SqlOperatorTest#testPowerFun()` and `SqlOperatorTest#testPowerDecimalFunc()` which check that `POWER(2, NULL)` and `POWER(NULL, 2)` both return `NULL`. However, none of these methods make any call to `Strong.policy(arg)`.

Any ideas on what to test here exactly? Please let me know.

caicancai:
> > @ninidhiaeddine We need a unit test for this change.
> 
> I was thinking to add a unit test for this change, but I didn't find any existing unit tests that are checking the returned `Strong.Policy` for a given `SqlOperator`.
> 
> The only unit tests that I found to be remotely related are: `SqlOperatorTest#testPowerFun()` and `SqlOperatorTest#testPowerDecimalFunc()` which check that `POWER(2, NULL)` and `POWER(NULL, 2)` both return `NULL`. However, none of these methods make any call to `Strong.policy(arg)`.
> 
> Any ideas on what to test here exactly? Please let me know.

You can add tests in sqloperatortest

mihaibudiu:
The Strong class is used during optimizations.
The JavaDoc for Strong says "Utilities for strong predicates."
However, POWER is not a predicate.
So I don't know whether this change has any effect.
It is not easy to audit all the places where Strong is used in the codebase to determine whether it will have any effect on any program. Most places I could find only simplify predicates.
I don't know if this has any effect on the *_REDUCE_EXPRESSIONS set of optimization rules, where perhaps POWER with a constant null argument could be simplified to NULL at compile-time.
Do you have an example which is affected by this issue?

ninidhiaeddine:
> The Strong class is used during optimizations. The JavaDoc for Strong says "Utilities for strong predicates." However, POWER is not a predicate. So I don't know whether this change has any effect. It is not easy to audit all the places where Strong is used in the codebase to determine whether it will have any effect on any program. Most places I could find only simplify predicates. I don't know if this has any effect on the *_REDUCE_EXPRESSIONS set of optimization rules, where perhaps POWER with a constant null argument could be simplified to NULL at compile-time. Do you have an example which is affected by this issue?

Thank you for your reply @mihaibudiu . Let's continue the discussion on the Jira issue's description if you don't mind, since @julianhyde is probably not seeing the discussion here. Once we agree on the changes, we will continue here

PR #4060, opened 2024-11-22
[CALCITE-6703] RelJson cannot handle timestamps prior to 1970-01-25 2…

tanclary:
…0:31:23.648

### DIFF
--- a/core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
+++ b/core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
@@ -740,6 +740,61 @@ class RelWriterTest {
         .assertThatPlan(isLinux(expected));
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6703">[CALCITE-6703]
+   * RelJson cannot handle timestamps prior to 1970-01-25 20:31:23.648</a>. */
+  @Test void testJsonToRexForTimestamp() {
+    // Below Integer.MAX_VALUE
+    final String timestampRepresentedAsInt = "{\n"
+          + "  \"literal\": 2129400000,\n"
+          + "  \"type\": {\n"
+          + "    \"type\": \"TIMESTAMP\",\n"
+          + "    \"nullable\": false\n"
+          + "  }\n"
+          + "}\n";
+    // Above Integer.MAX_VALUE
+    final String timestampRepresentedAsLong = "{\n"
+          + "  \"literal\": 3129400000,\n"
+          + "  \"type\": {\n"
+          + "    \"type\": \"TIMESTAMP\",\n"
+          + "    \"nullable\": false\n"
+          + "  }\n"
+          + "}\n";
+
+    // These timestamps were verified using BigQuery's UNIX_MILLIS function.
+    assertThatReadExpressionResult(timestampRepresentedAsInt, is("1970-01-25 15:30:00"));
+    assertThatReadExpressionResult(timestampRepresentedAsLong, is("1970-02-06 05:16:40"));
+  }
+
+  /** Test case for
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
```diff
@@ -740,6 +740,53 @@ private static Fixture relFn(Function<RelBuilder, RelNode> relFn) {
...
+        + "}\n";
+      final String timestampRepresentedAsLong = "{\n"
+        + "  \"literal\": 3129400000,\n"
+        + "  \"type\": {\n"
+        + "    \"type\": \"TIMESTAMP\",\n"
+        + "    \"nullable\": false\n"
+        + "  }\n"
+        + "}\n";
+
+    assertThatReadExpressionResult(timestampRepresentedAsInt, is("1970-01-25 15:30:00"));
```
mihaibudiu, line 762: is this supposed to work for dates before the epoch or the Gregorian calendar cutoff?
maybe you can add tests for these too - if they are not supposed to work, the errors can be tested.

core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
```diff
@@ -740,6 +740,53 @@ private static Fixture relFn(Function<RelBuilder, RelNode> relFn) {
...
+  @Test void testJsonToRexForTimestamp() {
+      final String timestampRepresentedAsInt = "{\n"
+        + "  \"literal\": 2129400000,\n"
+        + "  \"type\": {\n"
+        + "    \"type\": \"TIMESTAMP\",\n"
+        + "    \"nullable\": false\n"
+        + "  }\n"
+        + "}\n";
+      final String timestampRepresentedAsLong = "{\n"
+        + "  \"literal\": 3129400000,\n"
```
mihaibudiu, line 755: not easy to check that this value corresponds to the date below.

mihaibudiu (approved):

core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
tanclary: Yeah I also wasn't sure about this.. the point I was trying to illustrate is that one case is below Integer.MAX_VALUE and the other is above, any suggestions?

mihaibudiu: putting this in a comment will help a bit - the purpose of this choice is clearer
but if you have validated this using other external means, like another DB, you can also mention it in a comment.

core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
tanclary: So the mechanism going from millis->timestamp String actually lives in DateTimeUtils in Avatica and is pretty well-tested in DateTimeUtilsTest (let me know if you need links.) The real bug here is that JSON doesn't distinguish the numeric types so we have to ensure it's a Long (or convert) it. So that's the scope I'm trying to test here.

mihaibudiu: I have approved, I leave to you this decision.

core/src/test/java/org/apache/calcite/plan/RelWriterTest.java
tanclary: Ok i've added some comments

tanclary:
@mihaibudiu have you seen the Windows failures like the ones on this PR before? They look unrelated but thought I'd ask?

mihaibudiu:
I have seen similar errors in windows builds, but not this many... they usually go away if you rerun the failed jobs.
Must be some problem with the build machines.

PR #4061, opened 2024-11-22
[CALCITE-6691] Query with QUALIFY on CTE generates a plan that references wrong columns

julianhyde:


core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -3475,53 +3464,52 @@ private static JoinRelType convertJoinType(JoinType joinType) {
    * @param bb            Scope within which to resolve identifiers
    * @param select        Query
    * @param orderExprList Additional expressions needed to implement ORDER BY
+   * @param extraList     Additional expressions needed to implement QUALIFY
    */
   protected void convertAgg(Blackboard bb, SqlSelect select,
-      List<SqlNode> orderExprList) {
+      List<SqlNode> orderExprList, List<SqlNode> extraList) {
```
bchong95, line 3470: nit: maybe we call these `orderByExpressions` and `quailfyExpressions`?

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -6513,12 +6552,11 @@ LIMIT 5]]>
 LogicalSort(sort0=[$2], dir0=[ASC], fetch=[5])
   LogicalAggregate(group=[{0, 1, 2, 3}])
     LogicalProject(EMPNO=[$0], ENAME=[$1], DEPTNO=[$2], RANK_VAL=[$3])
-      LogicalFilter(condition=[$5])
-        LogicalProject(EMPNO=[$0], ENAME=[$1], DEPTNO=[$2], RANK_VAL=[$3], EXPR$0=[$4], QualifyExpression=[=($3, $4)])
+      LogicalFilter(condition=[$4])
+        LogicalProject(EMPNO=[$0], ENAME=[$1], DEPTNO=[$7], RANK_VAL=[RANK() OVER (PARTITION BY $1 ORDER BY $7 DESC)], QualifyExpression=[=(RANK() OVER (PARTITION BY $1 ORDER BY $7 DESC), $9)])
```
bchong95, line 6556: I think this is a performance regression, since we are computing the window function multiple times? I believe this is the scenario that the comment "This is a very specific application of Common Subexpression Elimination" was trying to handle.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -4696,19 +4693,49 @@ private RelNode convertMultisets(final List<SqlNode> operands,
...
+    if (qualify != null) {
+      // Add the QUALIFY expression to the select-list,
+      // convert the extended list,
+      // filter by the last field,
+      // and remove the last field.
+      convertSelectList(bb, measureBb, select, orderExprList,
+          ImmutableList.of(addAlias(qualify, "QualifyExpression")), null);
+      bb.setRoot(
+          relBuilder.push(bb.root())
+              .filter(last(relBuilder.fields()))// filter on last column
```
bchong95, line 4710: That's a neat way of doing this :)

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -6568,11 +6606,10 @@ QUALIFY rank_val = (SELECT COUNT(*) FROM emp)]]>
     <Resource name="plan">
       <![CDATA[
 LogicalProject(EMPNO=[$0], ENAME=[$1], DEPTNO=[$2], RANK_VAL=[$3])
-  LogicalFilter(condition=[$5])
-    LogicalProject(EMPNO=[$0], ENAME=[$1], DEPTNO=[$2], RANK_VAL=[$3], EXPR$0=[$4], QualifyExpression=[=($3, $4)])
+  LogicalFilter(condition=[$4])
+    LogicalProject(EMPNO=[$0], ENAME=[$1], DEPTNO=[$7], RANK_VAL=[RANK() OVER (PARTITION BY $1 ORDER BY $7 DESC)], QualifyExpression=[=(RANK() OVER (PARTITION BY $1 ORDER BY $7 DESC), $9)])
```
bchong95, line 6610: I think the JIRA mentioned something about the ProjectMerge logic in RelBuilder not handling the indexes in the window function. Did that ever get resolved?

julianhyde: I don't know. The cause of this bug seems simple - the qualify expression needs to be translated relative to the inputs to the SELECT clause, not relative to the outputs - and when I did that, the query seems to work.

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
julianhyde: I made note of this regression, and I'd rather it had not happened, but I don't view it as very serious. `Project` (unlike `Calc`) is not able to represent common subexpressions, so the appearance of duplicate expressions in `Project` does not mean that they are going to be executed separately.

A reasonable planner configuration will do subexpression elimination. Sql-to-rel shouldn't try too hard to do that.

bchong95: Good point!

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
bchong95: Yeah I that makes sense to me

bchong95 (approved):
Thank you for making this fix!

mihaibudiu:
@julianhyde do you want to merge this PR?

julianhyde:
Yes, will do, in next 24 hours.

PR #4065, opened 2024-11-26
Remove IN_FENNEL operator

mihaibudiu:
This operator does not seem to ever be instantiated.


caicancai:
> This operator does not seem to ever be instantiated.

why？

NobiGo (approved):

NobiGo:
We have already removed the relative code in https://github.com/apache/calcite/commit/30304bb489bb9eea84e556bc059010354a49c46d. So I think it is OK.

caicancai (approved):

PR #4066, opened 2024-11-27
[CALCITE-6706] Checked arithmetic does not take effect in subqueries

mihaibudiu:


mihaibudiu:
@NobiGo, I invited you to review since you commented on the issue. I have added the tests you suggested.

core/src/main/java/org/apache/calcite/sql2rel/ConvertToChecked.java
```diff
@@ -92,8 +98,7 @@ static class ConvertRexToChecked extends RexShuttle {
           result = call;
         }
         return builder.makeCast(call.getType(), result);
-      } else if (!SqlTypeName.EXACT_TYPES.contains(resultType)
-          || (resultType == SqlTypeName.DECIMAL)) {
```
NobiGo, line 96: Is this a related change?

core/src/main/java/org/apache/calcite/sql2rel/ConvertToChecked.java
```diff
@@ -50,13 +51,18 @@ public ConvertToChecked(RexBuilder builder) {
...
-  static class ConvertRexToChecked extends RexShuttle {
+  class ConvertRexToChecked extends RexShuttle {
     private final RexBuilder builder;
 
     ConvertRexToChecked(RexBuilder builder) {
       this.builder = builder;
     }
 
+    @Override public RexNode visitSubQuery(RexSubQuery subQuery) {
+      RelNode result = subQuery.rel.accept(ConvertToChecked.this);
```
NobiGo, line 62: SubQuery has operands that need to handle? maybe SQL like:
```
select count(*) as c
from "scott".emp as e
where  -CAST(-32768 AS SMALLINT) not in (
  select deptno
  from dept
  where dname = e.ename)
```
`-CAST(-32768 AS SMALLINT)` will be the Subquery's operands.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -3297,6 +3297,27 @@ static void checkOverlaps(OverlapChecker c) {
...
+        ".*Value 32768 does not fit in a SMALLINT", true);
+    f.checkFails("SELECT CAST(32768 AS SMALLINT)",
+        "Value 32768 out of range", true);
+    f.checkFails("SELECT -CAST(32768 AS SMALLINT)",
+        "Value 32768 out of range", true);
+    final SqlOperatorFixture f0 = fixture();
+    // This query does not fail if checked arithmetic is not used
+    f0.checkScalar("SELECT -CAST(-32768 AS SMALLINT)",
+        "-32768", "SMALLINT");
+    // The last two queries should fail in any conformance level
```
NobiGo, line 3314: Because `32768 AS SMALLINT` is an illegal expression? Can we add a statement to explain that the out-of-bounds are caused by the parameter type, not the final result type?

core/src/main/java/org/apache/calcite/sql2rel/ConvertToChecked.java
caicancai: +1

mihaibudiu: turns out that the second condition in the OR can never be true since it is covered by the previous if.

core/src/main/java/org/apache/calcite/sql2rel/ConvertToChecked.java
mihaibudiu: yes, this calling the visitor recursively on the subquery. This was the bug fixed in this PR: the recursive call wasn't made.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: I have pushed another commit which enhances this comment explaining why this fails in all conformance levels.

core/src/main/java/org/apache/calcite/sql2rel/ConvertToChecked.java
NobiGo: Can we add an SQL with a subquery to test this change?

mihaibudiu: The test in the issue fails without this change: `SELECT -CAST(-32768 AS SMALLINT)`
Turns out that because of the way Prepare works, this is a RexSubquery at the point where this visitor is invoked.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -3297,6 +3297,28 @@ static void checkOverlaps(OverlapChecker c) {
         DECIMAL_OVERFLOW, true);
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6706">[CALCITE-6706]
+   * Checked arithmetic does not take effect in subqueries</a>. */
+  @Test void testCastOverflow() {
+    final SqlOperatorFixture f = fixture().withConformance(SqlConformanceEnum.BIG_QUERY);
+    f.checkFails("SELECT -CAST(-32768 AS SMALLINT)",
+        ".*Value 32768 does not fit in a SMALLINT", true);
```
NobiGo, line 3305: `.*` is necessary for this test?

core/src/main/java/org/apache/calcite/sql2rel/ConvertToChecked.java
NobiGo: Ok, I test it in local mode, This SQL generates an SCALAR_QUERY in RelNode(It's strange for this SQL to generate SCALAR_QUERY, but it has nothing to do with this PR). So the test is valid.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: That takes the place of the substring "integer overflow:" I have replaced the * with this string in the test.

NobiGo (approved):

core/src/main/java/org/apache/calcite/sql2rel/ConvertToChecked.java
```diff
@@ -50,13 +51,18 @@ public ConvertToChecked(RexBuilder builder) {
...
+      return subQuery.clone(result);
```
caicancai, line 63: why `clone`?

mihaibudiu: This is not the Java Object.clone() method, that one has no arguments.

You can see the JavaDoc for the clone methods for RexCall in RexCall.clone().
```
  /**
   * Creates a new call to the same operator with different operands.
   *
   * @param type     Return type
   * @param operands Operands to call
   * @return New call
   */
  public RexCall clone(RelDataType type, List<RexNode> operands) {
```

caicancai (approved):

core/src/main/java/org/apache/calcite/sql2rel/ConvertToChecked.java
caicancai: thanks

NobiGo: If `result==subQuery.rel`, do we still need to clone it?

mihaibudiu: Sorry, didn't see this until too late, if this is redundant I will fix it in a subsequent PR.
Have to find out whether clone makes the check or not.

PR #4070, opened 2024-11-30
[CALCITE-6712] FROM_BASE64 and PARSE_URL's return data type should al…

NobiGo:
…ways be nullable

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/sql/test/SqlOperatorFixture.java
+++ b/testkit/src/main/java/org/apache/calcite/sql/test/SqlOperatorFixture.java
@@ -345,6 +345,18 @@ public interface SqlOperatorFixture extends AutoCloseable {
   void checkNull(String expression);
 
   /**
+   * Tests that a SQL expression returns the SQL NULL value and
+   * the data type of the returned result is nullable. For example,
+   *
+   * <blockquote>
+   * <pre>checkNullValueWithNullableType("CHAR_LENGTH(CAST(NULL AS VARCHAR(3))");</pre>
+   * </blockquote>
+   *
+   * @param expression Scalar expression
+   */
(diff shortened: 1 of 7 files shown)


mihaibudiu (commented):
Can you also update the spec in the site/_docs/reference.md file?

NobiGo:
> Can you also update the spec in the site/_docs/reference.md file?

Reference.md file mainly describes the function functionality, without adding any explanation about the return type of the function, I think we should not add it.

mihaibudiu:
Not the type, but the fact that on illegal inputs the result is null

NobiGo:
> Not the type, but the fact that on illegal inputs the result is null

Get it.

testkit/src/main/java/org/apache/calcite/sql/test/SqlTests.java
```diff
@@ -65,6 +65,15 @@ public abstract class SqlTests {
   public static final TypeChecker ANY_TYPE_CHECKER = (sql, type) -> {
   };
 
+  /**
+   * Checker which data type should be nullable.
```
mihaibudiu, line 69: I would rephrase this as "checker which enforces that a data type is nullable"

testkit/src/main/java/org/apache/calcite/sql/test/SqlOperatorFixture.java
```diff
@@ -344,6 +344,18 @@ void checkString(
    */
   void checkNull(String expression);
 
+  /**
+   * Tests that a SQL expression returns the SQL NULL value and
+   * returns the data type is NULLABLE. For example,
```
mihaibudiu, line 349: and the data type of the returned result is nullable

mihaibudiu (approved):

PR #4071, opened 2024-12-01
[CALCITE-6713] NVL2's return data type should be nullable if and only…

NobiGo:
… if at least one of the second and third parameters are nullable

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -11692,9 +11692,9 @@ public class SqlOperatorTest {
       f12.checkScalar("NVL2(NULL, 3.0, 2.111)", "2.111", "DECIMAL(4, 3) NOT NULL");
       f12.checkScalar("NVL2(3.111, 3.1415926, 2.111)", "3.1415926", "DECIMAL(8, 7) NOT NULL");
 
-      f12.checkNull("nvl2('ab', CAST(NULL AS VARCHAR(6)), 'def')");
-      f12.checkNull("nvl2(NULL, 'abc', NULL)");
-      f12.checkNull("nvl2(NULL, NULL, NULL)");
+      f12.checkNullValueWithNullableType("nvl2('ab', CAST(NULL AS VARCHAR(6)), 'def')");
+      f12.checkNullValueWithNullableType("nvl2(NULL, 'abc', NULL)");
+      f12.checkNullValueWithNullableType("nvl2(NULL, NULL, NULL)");
 
       f12.checkFails("^NVL2(2.0, 2.0, true)^", "Parameters must be of the same type", false);
       f12.checkFails("^NVL2(NULL, 2.0, true)^", "Parameters must be of the same type", false);
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -303,8 +303,7 @@ public abstract class SqlLibraryOperators {
   @LibraryOperator(libraries = {ORACLE, REDSHIFT, SPARK})
   public static final SqlBasicFunction NVL2 =
       SqlBasicFunction.create(SqlKind.NVL2,
-          ReturnTypes.NVL2_RESTRICTIVE
-              .andThen(SqlTypeTransforms.TO_NULLABLE_ALL),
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java
```diff
@@ -603,13 +603,21 @@ public static SqlCall stripSeparator(SqlCall call) {
...
-      opBinding.getTypeFactory().leastRestrictive(
-          Arrays.asList(opBinding.getOperandType(1),
-              opBinding.getOperandType(2)));
+   * Type-inference strategy for NVL2 function.
+   * If the first operand is nullable,
+   * it returns the least restrictive type between the second and third operands.
+   * If the first operand is non-nullable,
+   * it return the second operand type.
+   */
+  public static final SqlReturnTypeInference NVL2_RESTRICTIVE = opBinding -> {
```
caicancai, line 612: I remember that this function was adapted by me and it seems to comply with Oracle's specifications.

mihaibudiu: does this mean you think these changes are wrong?

NobiGo: `select nvl2(1, 2.0, 3.12);`
In Spark, return 2.00;
In Oracle, return 2;

`select nvl2(1, 2.1, 3.12);`
In Spark, return 2.10;
In Oracle, return 2.1;

caicancai: @NobiGo I think this issue has been discussed in jira
https://issues.apache.org/jira/browse/CALCITE-6397

NobiGo: Get it. In Oracle:
`select nvl2(null, 2.1, 3.120);`
return 3.12.
Oracle's return result will automatically omit 0, while Spark does not.

NobiGo: Therefore, the current strategy is correct, this PR only modifies whether the value should be nullable.

caicancai: I agree that some exceptions to this are understandable.

caicancai (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -11692,9 +11692,9 @@ void assertSubFunReturns(boolean binary, String s, int start,
       f12.checkScalar("NVL2(NULL, 3.0, 2.111)", "2.111", "DECIMAL(4, 3) NOT NULL");
       f12.checkScalar("NVL2(3.111, 3.1415926, 2.111)", "3.1415926", "DECIMAL(8, 7) NOT NULL");
 
-      f12.checkNull("nvl2('ab', CAST(NULL AS VARCHAR(6)), 'def')");
-      f12.checkNull("nvl2(NULL, 'abc', NULL)");
-      f12.checkNull("nvl2(NULL, NULL, NULL)");
+      f12.checkNullValueWithNullableType("nvl2('ab', CAST(NULL AS VARCHAR(6)), 'def')");
```
mihaibudiu, line 11695: so you will fix this in another PR?

mihaibudiu (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
NobiGo: There are two ways to fix this issue. 
One way is to first fix all function issues with `checkNullValueWithNullableType`, and then strengthen the 'checkNull' method to detect return data type is nullable, then delete  `checkNullValueWithNullableType`.
The other way is to directly delete the 'checkNullValueWithNullableType' method, and then use different methods to retest `checkNull`, such as `check`, which will retest the value and return type. The `checkNull` method is meaningless.
Do you have any suggestions?

mihaibudiu: Maybe it's simplest to just add a type check to checkNull.

NobiGo: Yes. I will add a type check to `checkNull` after deleting `checkNullValueWithNullableType`.Not now, because if we add type check in `checkNull` now, The unit test will not pass.

PR #4074, opened 2024-12-02
[CALCITE-6700] MySQL BIT_COUNT function should return result when parameter is Boolean, String, Date, Time and Timestamp types

strongduanmu:
* Support MySQL BIT_COUNT with boolean and String types parameter

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -868,7 +868,7 @@ public class RexImpTable {
       // bitwise
       defineMethod(BITCOUNT, BuiltInMethod.BITCOUNT.method, NullPolicy.STRICT);
       defineMethod(BIT_COUNT_BIG_QUERY, BuiltInMethod.BITCOUNT.method, NullPolicy.STRICT);
-      defineMethod(BIT_COUNT_MYSQL, BuiltInMethod.BITCOUNT.method, NullPolicy.STRICT);
+      define(BIT_COUNT_MYSQL, new BitCountMySQLImplementor());
 
       // datetime
       define(DATETIME_PLUS, new DatetimeArithmeticImplementor());
@@ -2689,6 +2689,44 @@ public class RexImpTable {
           argValueList.get(1));
     }
   }
+
+  /** Implementor for MYSQL {@code BIT_COUNT} function. */
+  private static class BitCountMySQLImplementor extends AbstractRexCallImplementor {
+    BitCountMySQLImplementor() {
+      super("bitCount", NullPolicy.STRICT, false);
+    }
+
+    @Override Expression implementSafe(final RexToLixTranslator translator,
+        final RexCall call, final List<Expression> argValueList) {
+      Expression expr = argValueList.get(0);
+      RelDataType relDataType = call.getOperands().get(0).getType();
+      if (SqlTypeUtil.isNull(relDataType)) {
+        return argValueList.get(0);
+      }
+      // In MySQL, BIT_COUNT(TIMESTAMP '1996-08-03 16:22:34') is converted to
+      // BIT_COUNT('19960803162234') for calculation, so the internal int value
+      // needs to be converted to DATE/TIME and TIMESTAMP.
+      SqlTypeName type = relDataType.getSqlTypeName();
+      switch (type) {
+      case VARBINARY:
+      case BINARY:
+        return Expressions.call(SqlFunctions.class, "bitCount", expr);
+      case DATE:
+        expr = Expressions.call(BuiltInMethod.INTERNAL_TO_DATE.method, expr);
+        break;
+      case TIME:
+        expr = Expressions.call(BuiltInMethod.INTERNAL_TO_TIME.method, expr);
+        break;
+      case TIMESTAMP:
+        expr = Expressions.call(BuiltInMethod.INTERNAL_TO_TIMESTAMP.method, expr);
+        break;
+      default:
+        break;
+      }
+      return Expressions.call(SqlFunctions.class, "bitCountMySQL", Expressions.box(expr));
+    }
+  }
+
 
   /** Implementor for the {@code MONTHNAME} and {@code DAYNAME} functions.
    * Each takes a {@link java.util.Locale} argument. */
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -2916,6 +2916,46 @@ public class SqlFunctions {
     return bitsSet;
   }
 
+  /** Helper function for implementing MySQL <code>BIT_COUNT</code>. Counts the number
+   * of bits set in a boolean value. */
+  public static long bitCountMySQL(Boolean b) {
+    return Long.bitCount(b ? 1L : 0L);
+  }
+
+  /** Helper function for implementing MySQL <code>BIT_COUNT</code>. Counts the number
+   * of bits set in a string value. */
+  public static long bitCountMySQL(String b) {
+    try {
+      return bitCount(new BigDecimal(b));
+    } catch (Exception ignore) {
+      return 0;
+    }
+  }
+
+  /** Helper function for implementing MySQL <code>BIT_COUNT</code>. Counts the number
+   * of bits set in a number value. */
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1166,7 +1166,14 @@ static RelDataType deriveTypeSplit(SqlOperatorBinding operatorBinding,
           SqlKind.OTHER_FUNCTION,
           ReturnTypes.BIGINT_NULLABLE,
           null,
-          OperandTypes.NUMERIC.or(OperandTypes.BINARY),
+          OperandTypes.or(OperandTypes.NUMERIC,
+              OperandTypes.BINARY,
+              OperandTypes.BOOLEAN,
+              OperandTypes.CHARACTER,
+              OperandTypes.DATETIME,
```
mihaibudiu, line 1173: you have here more types than the JIRA title promises.
Also, it seems from the JIRA comments that you are not really supporting these other types yet.

The solution for the question you asked in the JIRA is to pass some type information to the runtime. If the input is a Date, you need to pass a "DATE" flag to the runtime so that it interprets the integer as a date.

strongduanmu: Thanks for your feedback @mihaibudiu, is there any example I can refer to? I am not familiar with the Calcite runtime and I will take some time to learn it.

mihaibudiu: This function you wrote `bitCountMySQL` is part of the runtime.
You just need to pass it an extra argument that describes the type.

strongduanmu: Thanks for your guidance, I will try this solution.

NobiGo (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -16024,6 +16024,25 @@ void checkBitAnd(SqlOperatorFixture f0, FunctionAlias functionAlias) {
 
   @Test void testBitCountMySQLFunc() {
     checkBitCount(SqlLibraryOperators.BIT_COUNT_MYSQL, list(SqlLibrary.MYSQL), true);
+    checkBitCountMySQL();
+  }
+
+  private void checkBitCountMySQL() {
```
caicancai, line 16030: Add jira link

strongduanmu: Done, thank you for your suggeation.

core/src/test/resources/sql/functions.iq
```diff
@@ -31,6 +31,80 @@ select bit_count(8);
...
++--------+
+| EXPR$0 |
++--------+
+|      1 |
++--------+
+(1 row)
+
+!ok
+
+select bit_count(joinedat), bit_count(joinetime), bit_count(joinetimestamp) from emps_date_time ORDER BY empno;
```
mihaibudiu, line 94: have these been validated with MySQL?
if yes, please add a comment about this in the iq file.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -2916,6 +2916,46 @@ public static long bitCount(ByteString b) {
...
   }
 
+  /** Helper function for implementing MySQL <code>BIT_COUNT</code>. Counts the number
+   * of bits set in a boolean value. */
+  public static long bitCountMySQL(Boolean b) {
+    return Long.bitCount(b ? 1L : 0L);
+  }
+
+  /** Helper function for implementing MySQL <code>BIT_COUNT</code>. Counts the number
+   * of bits set in a string value. */
```
mihaibudiu, line 2926: if you can point to the MySQL documentation where this is specified, please add a link to the comments.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -2690,6 +2690,48 @@ private static class ContainsSubstrImplementor extends AbstractRexCallImplemento
...
+      case DATE:
+        expr = Expressions.call(BuiltInMethod.INTERNAL_TO_DATE.method, expr);
+        break;
+      case TIME:
+      case TIME_WITH_LOCAL_TIME_ZONE:
+      case TIME_TZ:
+        expr = Expressions.call(BuiltInMethod.INTERNAL_TO_TIME.method, expr);
+        break;
+      case TIMESTAMP:
+      case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
```
mihaibudiu, line 2723: Perhaps tests with the local timezone are difficult, but I don't see tests with timezone either.
It would be great to cover all these cases with tests.

core/src/test/resources/sql/functions.iq
strongduanmu: Yes, I have verified the results in MySQL. Here are the results on MySQL.

```
mysql> SELECT BIT_COUNT(DATE '1996-08-03'), BIT_COUNT(TIME '16:22:34'), BIT_COUNT(TIMESTAMP '1996-08-03 16:22:34') UNION ALL
    -> SELECT BIT_COUNT(DATE '2001-01-01'), BIT_COUNT(TIME '12:20:00'), BIT_COUNT(TIMESTAMP '2001-01-01 12:20:00') UNION ALL
    -> SELECT BIT_COUNT(DATE '2002-05-03'), BIT_COUNT(TIME '13:12:14'), BIT_COUNT(TIMESTAMP '2002-05-03 13:12:14') UNION ALL
    -> SELECT BIT_COUNT(DATE '2005-09-07'), BIT_COUNT(TIME '06:02:04'), BIT_COUNT(TIMESTAMP '2005-09-07 06:02:04') UNION ALL
    -> SELECT BIT_COUNT(DATE '2007-01-01'), BIT_COUNT(TIME '23:09:59'), BIT_COUNT(TIMESTAMP '2007-01-01 23:09:59');
+------------------------------+----------------------------+--------------------------------------------+
| BIT_COUNT(DATE '1996-08-03') | BIT_COUNT(TIME '16:22:34') | BIT_COUNT(TIMESTAMP '1996-08-03 16:22:34') |
+------------------------------+----------------------------+--------------------------------------------+
|                           12 |                         11 |                                         24 |
|                           12 |                          8 |                                         22 |
|                           14 |                          5 |                                         22 |
|                           16 |                          9 |                                         25 |
|                           14 |                         10 |                                         21 |
+------------------------------+----------------------------+--------------------------------------------+
5 rows in set (0.01 sec)
```

mihaibudiu: you can paste this as a comment in the iq file, together with the version of mysql used.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
strongduanmu: This is MySQL BIT_COUNT function document - https://dev.mysql.com/doc/refman/8.4/en/bit-functions.html#bit-operations-bit-count, but unfortunately, it does not describe the details about the BIT_COUNT function. These special behaviors of BIT_COUNT come from actual tests with MySQL.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
strongduanmu: Hi @mihaibudiu, I checked the documentation of MySQL TIMESTAMP - https://dev.mysql.com/doc/refman/8.4/en/datetime.html, and found that it does not provide syntax for `with time zone` and `with local time zone`. Does this mean I should remove `TIMESTAMP_WITH_LOCAL_TIME_ZONE` and `TIMESTAMP_TZ` here?

mihaibudiu: It's certainly simpler not to support them

strongduanmu:
Hi @mihaibudiu, I have revised the code according to your suggestions. If you have time, please help review it again. Thank you very much for your guidance.

mihaibudiu (approved):
Please squash the commits into a single commit so we can merge.
Use the same message as the JIRA title.

PR #4076, opened 2024-12-04
[CALCITE-6714] Cast literal to interval gives the wrong result if literal is casted

alex-plekhanov:


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1000,6 +1000,10 @@ void testCastToInterval(CastType castType, SqlOperatorFixture f) {
         "cast(-5723 as interval minute(4))",
         "-5723",
         "INTERVAL MINUTE(4) NOT NULL");
+    f.checkScalar(
```
caicancai, line 1003: just integer?

alex-plekhanov: What else? I can add DECIMAL, like this:
```
    f.checkScalar(
        "cast(cast(1 as decimal) as interval minute)",
        "+1",
        "INTERVAL MINUTE NOT NULL");
```
Is it ok?
Approximate numerics (FLOAT, DOUBLE) are not supported for casting to interval.

caicancai: @alex-plekhanov Thanks, it would be helpful if you could add some more different types of tests you can think of

alex-plekhanov: Ok, thank you, I've added the test with DECIMAL.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -1000,6 +1000,14 @@ void testCastToInterval(CastType castType, SqlOperatorFixture f) {
...
+        "cast(cast(1 as integer) as interval minute)",
+        "+1",
+        "INTERVAL MINUTE NOT NULL");
+    f.checkScalar(
```
mihaibudiu, line 1007: how about other interval types, such as `INTERVAL HOUR`?

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1336,8 +1337,15 @@ protected RexLiteral makeLiteral(
       o = ((TimestampWithTimeZoneString) o).round(p);
       break;
     case DECIMAL:
-      if (o != null && type.getScale() != RelDataType.SCALE_NOT_SPECIFIED) {
-        assert o instanceof BigDecimal;
+      if (o == null) {
```
mihaibudiu, line 1340: is this a problem with literals only, or does it happen to non-constant expressions too?
Can you check if there is a Quidem test (in an .iq file) for this?
If not, can you please add some?

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1336,8 +1337,15 @@ protected RexLiteral makeLiteral(
...
+        break;
+      }
+      assert o instanceof BigDecimal;
+      if (type instanceof IntervalSqlType) {
+        SqlIntervalQualifier qualifier = ((IntervalSqlType) type).getIntervalQualifier();
+        o = ((BigDecimal) o).multiply(qualifier.getUnit().multiplier);
```
mihaibudiu, line 1346: multiplier can also be fractional, can you also add a test for some of these units, like MILLISECONDS?
It may be easier to add tests to a quide file than to SqlOperatorTest.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
alex-plekhanov: For non-constant expressions actual casting performed by `CastImplementor` (see `RexToLixTranslator#checkExpressionPadTruncate`, there is multiplication to time unit multiplier too). But if cast contains trivial expressions it can be simplified to just literal, without call of `CastImplementor`.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
alex-plekhanov: There are no intervals less than SECONDS. Multiplier for seconds is 1000.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
alex-plekhanov: I've added test for HOUR, MONTH and SECOND intervals.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
julianhyde: An analogous problem might happen with MONTH-YEAR intervals. (MONTH is the smallest unit, so would be analogous to SECONDS , and YEAR is larger, so would be analogous to MINUTE.) Can you add a test for that case?

alex-plekhanov: @julianhyde, test for MONTH has already been added with the last commit.

mihaibudiu (approved):

mihaibudiu:
Thank you.

caicancai (approved):
LGTM

PR #4077, opened 2024-12-04
[CALCITE-6715] Enhance RelFieldTrimmer to trim LogicalCorrelate nodes

kgyrtkirk:


core/src/test/java/org/apache/calcite/sql2rel/RelFieldTrimmerTest.java
```diff
@@ -544,4 +548,36 @@ public static Frameworks.ConfigBuilder config() {
...
+                    SqlStdOperatorTable.PLUS,
+                    builder.literal(10),
+                    builder.field(v.get(), "DEPTNO"))))
+        .correlate(JoinRelType.LEFT, v.get().id, builder.field(2, 0, "DEPTNO"))
+        .aggregate(builder.groupKey("ENAME"), builder.max(builder.field("EMPNO")))
+        .build();
+
+    final RelFieldTrimmer fieldTrimmer = new RelFieldTrimmer(null, builder);
+    final RelNode trimmed = fieldTrimmer.trim(root);
+    final String expected = ""
```
mihaibudiu, line 571: a bit difficult to see what the trimmer has done.
perhaps if you have the expected tree before the trim it will be easier?
then you can also assert that they are different

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -1336,6 +1338,132 @@ public TrimResult trimFields(
...
+  /**
+   * Updates correlate references in {@link RexNode} expressions.
+   */
+  static class RexCorrelVariableMapShuttle extends RexShuttle {
+    private final CorrelationId correlationId;
+    private final Mapping mapping;
+    private final RelDataType newCorrelRowType;
+    private final RexBuilder rexBuilder;
+
+    RexCorrelVariableMapShuttle(final CorrelationId correlationId,
```
mihaibudiu, line 1410: Even though the class is not public, I think JavaDoc for the parameters would make this easier to understand.

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -1336,6 +1338,132 @@ public TrimResult trimFields(
...
+   * Concatenates multiple mappings.
+   *
+   * <pre>
+   * [ 1:0, 2:1] // sourceCount:100
+   * [ 1:0, 2:1] // sourceCount:100
+   * output:
+   * [ 1:0, 2:1, 101:2, 102:3 ] ; sourceCount:200
+   * </pre>
+   */
+  private Mapping makeMapping(List<Mapping> inputMappings) {
```
mihaibudiu, line 1444: concatenateMappings is a more suggesting name.

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -1336,6 +1338,132 @@ public TrimResult trimFields(
...
+    RexBuilder rexBuilder = relBuilder.getRexBuilder();
+
+    RexCorrelVariableMapShuttle rexVisitor =
+        new RexCorrelVariableMapShuttle(correlate.getCorrelationId(),
+            newInputs.get(0).getRowType(), mapping, rexBuilder);
+    final LogicalCorrelate newCorrelate =
+        correlate
+            .copy(correlate.getTraitSet(),
+                newInputs.get(0),
+                newInputs.get(1).accept(new RexRewritingRelShuttle(rexVisitor)),
```
mihaibudiu, line 1393: I find it easier to debug if I avoid complex expressions - I would use a temporary for the result of the visitor.

mihaibudiu: this also assumes implicitly that newInputs.size() == 2; perhaps this is obvious, but then it would not harm to assert that this holds for the correlate

kgyrtkirk: extracted them to be `newLeft` and `newRight` (I was thinking the same earlier ; but I've only extracted the `rexVistior`)

`LogicalCorrelate`'s constructor has only `left` and `right` - but that may change...I've added a check to ensure that this only applies when `correlate.getInputs().size()` is 2

core/src/test/java/org/apache/calcite/sql2rel/RelFieldTrimmerTest.java
kgyrtkirk: I've updated the testcase to use an `Uncollect` instead - and also added the before-trim plan to the test.

kgyrtkirk:
Thank you @mihaibudiu  for taking a look!
I've updated the PR according to your suggestions - kept the new changes as a separate commit for now.
Could you please take another look?

core/src/test/resources/sql/sub-query.iq
```diff
@@ -481,7 +481,7 @@ EnumerableCalc(expr#0..2=[{inputs}], proj#0..1=[{exprs}])
   EnumerableCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{0}])
     EnumerableValues(tuples=[[{ 1, 2 }]])
     EnumerableAggregate(group=[{0}])
-      EnumerableCalc(expr#0..7=[{inputs}], expr#8=[true], expr#9=[CAST($t7):INTEGER], expr#10=[$cor0], expr#11=[$t10.A], expr#12=[=($t9, $t11)], i=[$t8], $condition=[$t12])
+      EnumerableCalc(expr#0..7=[{inputs}], expr#8=[true], expr#9=[CAST($t7):INTEGER], expr#10=[$cor0], expr#11=[$t10.EXPR$0], expr#12=[=($t9, $t11)], i=[$t8], $condition=[$t12])
```
suibianwanwank, line 484: Why has it changed here, and why hasn't it changed below when the plan below is the same as here?

core/src/test/java/org/apache/calcite/sql2rel/RelFieldTrimmerTest.java
suibianwanwank: Can you add a more generalized example, where both left and right are SCAN and are correlate by some field, and both left and right be trimmed.

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -1336,6 +1338,144 @@ public TrimResult trimFields(
...
+   * Concatenates multiple mappings.
+   *
+   * <pre>
+   * [ 1:0, 2:1] // sourceCount:100
+   * [ 1:0, 2:1] // sourceCount:100
+   * output:
+   * [ 1:0, 2:1, 101:2, 102:3 ] ; sourceCount:200
+   * </pre>
+   */
+  private Mapping concatenateMappings(List<Mapping> inputMappings) {
```
rubenada, line 1456: IMO this method might be re-used on other contexts, so it could be moved as a `public static` utility method inside `Mappings` class.

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -1336,6 +1338,144 @@ public TrimResult trimFields(
...
+    for (Mapping mapping : inputMappings) {
+      fieldCount += mapping.getSourceCount();
+      newFieldCount += mapping.getTargetCount();
+    }
+
+    Mapping mapping =
+        Mappings.create(MappingType.INVERSE_SURJECTION, fieldCount, newFieldCount);
+    int offset = 0;
+    int newOffset = 0;
+    for (int i = 0; i < inputMappings.size(); i++) {
```
rubenada, line 1468: nit: we could have this for loop simplified as `for (Mapping inputMapping : inputMappings) {...}`

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -1336,6 +1338,144 @@ public TrimResult trimFields(
     return result(newSnapshot, inputMapping, snapshot);
   }
 
+  /**
+   * Trims {@link LogicalCorrelate} nodes.
+   */
+  public TrimResult trimFields(LogicalCorrelate correlate,
+      ImmutableBitSet fieldsUsed, Set<RelDataTypeField> extraFields) {
+    if (!extraFields.isEmpty() || correlate.getInputs().size() != 2) {
```
rubenada, line 1346: Correlate is a BiRel, so the condition `correlate.getInputs().size() != 2` seems unnecessary.

core/src/test/resources/sql/sub-query.iq
kgyrtkirk: Actually both of them are misleading; there is no `A` in either plans...due to the fact that `RexInputRef` accesses column by idx not by name - it has no real effect

### for the 1st case the following happens:

<details>
<summary>RelFieldTrimmer match</summary>

matches on the plan:
```
LogicalCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{0}])
  LogicalProject(EXPR$0=[$0], EXPR$1=[$1])
    LogicalValues(tuples=[[{ 1, 2 }]])
  LogicalAggregate(group=[{0}])
    LogicalProject(i=[true])
      LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7])
        LogicalFilter(condition=[=($8, $cor0.A)])
          LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], DEPTNO0=[CAST($7):INTEGER])
```
* left is left intact
* the right is trimmed from:
```
LogicalAggregate(group=[{0}])
  LogicalProject(i=[true])
    LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7])
      LogicalFilter(condition=[=($8, $cor0.A)])
        LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], DEPTNO0=[CAST($7):INTEGER])

```
to:
```
LogicalAggregate(group=[{0}])
  LogicalProject(i=[true])
    LogicalFilter(condition=[=($1, $cor0.A)])
      LogicalProject(EMPNO=[$0], DEPTNO0=[CAST($7):INTEGER])
```
* trim is executed
* the rowType used for left is `RecordType(INTEGER EXPR$0, INTEGER EXPR$1)` (as it has the Project at this point)
* during the rewrite: a `RexFieldAccess` is created to reference column `0`
</details>

<details>
<summary>ProjectRemoveRule fires</summary>

the project of:
```
LogicalProject(EXPR$0=[$0], EXPR$1=[$1])
  LogicalValues(subset=[rel#99:RelSubset#0.NONE.[]], tuples=[[{ 1, 2 }]])
```
Is identified as [isTrivial](https://github.com/apache/calcite/blob/0eb83b1dc6316cba73c2c49e8fbccfa390333fbe/core/src/main/java/org/apache/calcite/rel/rules/ProjectRemoveRule.java#L82)

The project is removed.
</details>


### in the other case:

<details>
<summary>RelFieldTrimmer matches on a different plan</summary>

```
LogicalCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{0}])
  LogicalProject(A=[$0], B=[$1])
    LogicalValues(tuples=[[{ 1, 2 }]])
  LogicalAggregate(group=[{0}])
    LogicalProject(i=[true])
      LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7])
        LogicalFilter(condition=[=($8, $cor0.A)])
          LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], DEPTNO0=[CAST($7):INTEGER])
```

* which has `LogicalProject(A=[$0], B=[$1])` ; so the rowtype for left is `RecordType(INTEGER A, INTEGER B)`
* similar transform happens ; but rowType has `A` in it.

</details>
<details>
<summary>ProjectRemoveRule fires</summary>

the project of:
```
LogicalProject(A=[$0], B=[$1])
  LogicalValues(subset=[rel#100:RelSubset#0.NONE.[]], tuples=[[{ 1, 2 }]])
```
Is identified as [isTrivial](https://github.com/apache/calcite/blob/0eb83b1dc6316cba73c2c49e8fbccfa390333fbe/core/src/main/java/org/apache/calcite/rel/rules/ProjectRemoveRule.java#L82)

The project is removed.
</details>

core/src/test/java/org/apache/calcite/sql2rel/RelFieldTrimmerTest.java
kgyrtkirk: I've added another test which removes an unecessarily projected expression from both sides

mihaibudiu:
Lint didn't like your commit message - too bad, since it will go away anyway

mihaibudiu:
This is better, but @rubenada made some very good suggestions, so I will review once you have addressed these too.

kgyrtkirk:
I think I've addressed all comments so-far in [this commit](https://github.com/apache/calcite/pull/4077/commits/18d03e36d904441b0c91931c3f47f2d66c786a93)

Lint is kinda angry because I've multiple commits on the branch and not squashed the changes to be a single commit...would you prefer to squash them? 

...maybe I should use conforming format for all commits next time; that could keep the `LintTest` happy :)

kgyrtkirk:
I've force pushed the branch to get a proper CI run; I'll add `LintTest` compliant commits from now on :)

mihaibudiu (approved):

mihaibudiu:
If there are no other comments this looks good to merge. Thank you.

suibianwanwank (approved):

rubenada (approved):

kgyrtkirk:
Thank you @mihaibudiu , @rubenada and @suibianwanwank for reviewing the changes!

PR #4078, opened 2024-12-06
[CALCITE-6718] Optimize SubstitutionVisitor's splitFilter with early return and uniform simplification for equivalence checking

hannerwang:
### What changes were proposed in this pull request?
1. Implement early return for materialized view range checking.
2. Apply uniform simplification for expression equivalence checking.

### Why are the changes needed?
Early Return: Many materialized views do not have filters, so implementing early return can avoid unnecessary checks and improve performance.
Uniform Simplification: In user-customized applications, we cannot guarantee that both sides of the equivalence check have been simplified using simplifyUnknownAsFalse. Adding this ensures consistent behavior.

### Does this PR introduce any user-facing change?
No.

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
+++ b/core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
@@ -297,8 +297,11 @@ public class SubstitutionVisitor {
       RexNode condition, RexNode target) {
     final RexBuilder rexBuilder = simplify.rexBuilder;
     condition = simplify.simplify(condition);
-    target = simplify.simplify(target);
     RexNode condition2 = canonizeNode(rexBuilder, condition);
+    if (target.isAlwaysTrue()) {
+      return condition2;
+    }
+    target = simplify.simplify(target);
     RexNode target2 = canonizeNode(rexBuilder, target);
 
     // First, try splitting into ORs.
@@ -321,7 +324,7 @@ public class SubstitutionVisitor {
           RexUtil.composeConjunction(rexBuilder,
               ImmutableList.of(condition2, target2));
       RexNode r =
-          canonizeNode(rexBuilder, simplify.simplifyUnknownAsFalse(x2));
+          canonizeNode(rexBuilder, simplify.simplify(x2));
       if (!r.isAlwaysFalse() && isEquivalent(condition2, r)) {
         List<RexNode> conjs = RelOptUtil.conjunctions(r);
         for (RexNode e : RelOptUtil.conjunctions(target2)) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/plan/SubstitutionVisitor.java
```diff
@@ -322,7 +325,9 @@ void register(MutableRel result, MutableRel query) {
               ImmutableList.of(condition2, target2));
       RexNode r =
           canonizeNode(rexBuilder, simplify.simplifyUnknownAsFalse(x2));
-      if (!r.isAlwaysFalse() && isEquivalent(condition2, r)) {
+      RexNode simplifiedCond2 =
+          canonizeNode(rexBuilder, simplify.simplifyUnknownAsFalse(condition2));
+      if (!r.isAlwaysFalse() && isEquivalent(simplifiedCond2, r)) {
```
suibianwanwank, line 330: I agree with Early Return, it cuts out some unnecessary steps. But doing another simplify for condition2 seems unnecessary, condition2 only canonizeNode after simplify.
How about using simplify instead of simplifyUnknownAsFalse for _x2_ .

hannerwang: Actually I'm not sure why this function uses both simplify and simplifyUnknownAsFalse, so I made a conservative change to minimize impact. If possible, we could replace simplify with simplifyUnknownAsFalse, but that seems too radical.

asolimando: I agree with @suibianwanwank, we shouldn't need another simplification on `condition2` while early return seems good.

Here we are dealing with filtering predicates, so I guess that `simplifyUnknownAsFalse` has the right semantics, rather than just `simplify`, but we need tests covering these changes and showing that we are doing the right thing on corner cases involving `unknown` and the early return condition.

@hannerwang, can you add them?

hannerwang: Hello @asolimando ,
I attempted to change simplifyUnknownAsFalse to simplify, but I encountered some issues. 
For instance, with the expression `x >= 5 and x > 5`, the simplify method does not reduce it, whereas simplifyUnknownAsFalse simplifies it to` x > 5`. Additionally, expressions like` x > 5 and x < 3` are simplified to `x <> x` when UNKNOWN is treated as UNKNOWN, and to `false` when UNKNOWN is treated as FALSE. 
Therefore, we cannot simply replace simplifyUnknownAsFalse with simplify, as it would cause many tests to fail.

PR #4080, opened 2024-12-06
[CALCITE-6711] Functions whose output value can be null should return…

NobiGo:
… a nullable type

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -575,7 +575,7 @@ public abstract class SqlLibraryOperators {
    * Returns the substring in value that matches the regexp. Returns NULL if there is no match. */
   @LibraryOperator(libraries = {BIG_QUERY})
   public static final SqlBasicFunction REGEXP_EXTRACT =
-      SqlBasicFunction.create("REGEXP_EXTRACT", ReturnTypes.VARCHAR_NULLABLE,
+      SqlBasicFunction.create("REGEXP_EXTRACT", ReturnTypes.VARCHAR_FORCE_NULLABLE,
           OperandTypes.STRING_STRING_OPTIONAL_INTEGER_OPTIONAL_INTEGER,
           SqlFunctionCategory.STRING);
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1524,23 +1524,24 @@ private static RelDataType arrayCompactReturnType(SqlOperatorBinding opBinding)
   private static RelDataType arrayInsertReturnType(SqlOperatorBinding opBinding) {
     final RelDataType arrayType = opBinding.collectOperandTypes().get(0);
     final RelDataType componentType = arrayType.getComponentType();
-    final RelDataType elementType = opBinding.collectOperandTypes().get(2);
+    final RelDataType elementType1 = opBinding.collectOperandTypes().get(1);
```
mihaibudiu, line 1527: This function collectOperandTypes() is called 3 times. If it allocates a new list, the list should probably be reused.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -11692,9 +11683,9 @@ void assertSubFunReturns(boolean binary, String s, int start,
       f12.checkScalar("NVL2(NULL, 3.0, 2.111)", "2.111", "DECIMAL(4, 3) NOT NULL");
       f12.checkScalar("NVL2(3.111, 3.1415926, 2.111)", "3.1415926", "DECIMAL(8, 7) NOT NULL");
 
-      f12.checkNullValueWithNullableType("nvl2('ab', CAST(NULL AS VARCHAR(6)), 'def')");
-      f12.checkNullValueWithNullableType("nvl2(NULL, 'abc', NULL)");
-      f12.checkNullValueWithNullableType("nvl2(NULL, NULL, NULL)");
+      f12.checkNull("nvl2('ab', CAST(NULL AS VARCHAR(6)), 'def')");
```
mihaibudiu, line 11686: I thought you also meant to change checkNull to assert that the type is nullable.
Or is that already happening?

mihaibudiu (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
NobiGo: This PR makes `checknull` to support the check nullable type.

PR #4082, opened 2024-12-10
[CALCITE-6723] Type inference for ARRAY_INSERT function is incorrect

mihaibudiu:


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8221,36 +8221,36 @@ void checkArrayReverseFunc(SqlOperatorFixture f0, SqlFunction function,
...
-        "[1, 2, 3, 4]", "INTEGER NOT NULL ARRAY NOT NULL");
+        "[1, 2, 3, 4]", "INTEGER ARRAY NOT NULL");
     f1.checkScalar("array_insert(array[1, 3, 4], -1, 2)",
-        "[1, 3, 4, 2]", "INTEGER NOT NULL ARRAY NOT NULL");
+        "[1, 3, 4, 2]", "INTEGER ARRAY NOT NULL");
     f1.checkScalar("array_insert(array[1, 3, 4], -3, 2)",
-        "[1, 2, 3, 4]", "INTEGER NOT NULL ARRAY NOT NULL");
+        "[1, 2, 3, 4]", "INTEGER ARRAY NOT NULL");
     f1.checkScalar("array_insert(array[2, 3, null, 4], -6, 1)",
         "[1, null, 2, 3, null, 4]", "INTEGER ARRAY NOT NULL");
```
NobiGo, line 8234: Yes, I overlooked this and only covered the correctness of the Array data type outside.

NobiGo: Can we strengthen this `checkScalar` method? Can detect if other functions have similar issues. This may be an overlooked testing point. WDYT?

NobiGo: The current issue under discussion does not affect the merge of this PR.

NobiGo (approved):

PR #4083, opened 2024-12-10
[CALCITE-6724] MockTable support for multiple (individual & composite) keys

zabetak:
MockTable currently allows only one key to be specified; the key can be either simple or composite. In real scenarios though a table can have multiple keys.

1. Add `MockTable#addKey` API to explicitly support multiple (individual & composite) keys.
2. Modify `MockTable#addColumn` to create individual keys when `isKey` is `true`. There is a slight change in behavior since when the method is called multiple times for the same table it will now create multiple simple keys instead of one composite.

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReader.java
+++ b/testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReader.java
@@ -316,7 +316,7 @@ public abstract class MockCatalogReader extends CalciteCatalogReader {
     protected final double rowCount;
     protected final List<Map.Entry<String, RelDataType>> columnList =
         new ArrayList<>();
-    protected final List<Integer> keyList = new ArrayList<>();
+    protected final List<ImmutableBitSet> keyList = new ArrayList<>();
     protected final List<RelReferentialConstraint> referentialConstraints =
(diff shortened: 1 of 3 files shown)


testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReader.java
```diff
@@ -651,11 +647,43 @@ public void addColumn(String name, RelDataType type) {
 
     public void addColumn(String name, RelDataType type, boolean isKey) {
       if (isKey) {
-        keyList.add(columnList.size());
+        keyList.add(ImmutableBitSet.of(columnList.size()));
       }
       columnList.add(Pair.of(name, type));
     }
 
+    public void addKey(String... columns) {
```
ILuffZhe, line 655: nitpick: what about `addKeys`?

ILuffZhe (approved):
LGTM!

testkit/src/main/java/org/apache/calcite/test/catalog/MockCatalogReader.java
zabetak: The method adds one key (over possibly multiple columns) so using singular noun seems more appropriate.

PR #4088, opened 2024-12-12
[CALCITE-6727] Column uniqueness constrain should only apply to inner join

joeyutong:
In the existing test
```
// RelMetadataTest.java
@Test void testColumnUniquenessForJoinOnLimit1() {
  final String sql = ""
      + "select *\n"
      + "from emp A\n"
      + "join (\n"
      + "  select * from emp\n"
      + "  limit 1) B\n"
      + "on A.empno = B.empno";
  sql(sql)
      .assertThatAreColumnsUnique(bitSetOf(0), is(true))
      .assertThatAreColumnsUnique(bitSetOf(1), is(true))
      .assertThatAreColumnsUnique(bitSetOf(9), is(true))
      .assertThatAreColumnsUnique(bitSetOf(10), is(true))
      .assertThatAreColumnsUnique(bitSetOf(), is(true))
      .assertThatUniqueKeysAre(bitSetOf());
} 
```
the join result column `A.ENAME` is considered to be unique because `A.EMPNO` is the unique key. The test still passes when the query changes to left join, in which case `A.ENAME` is not guaranteed to be unique.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
@@ -321,15 +321,16 @@ public class RelMdColumnUniqueness
     }
 
     final JoinInfo joinInfo = rel.analyzeCondition();
-
-    // Joining with a singleton constrains the keys on the other table
-    final Double rightMaxRowCount = mq.getMaxRowCount(right);
-    if (rightMaxRowCount != null && rightMaxRowCount <= 1.0) {
-      leftColumns = leftColumns.union(joinInfo.leftSet());
-    }
-    final Double leftMaxRowCount = mq.getMaxRowCount(left);
-    if (leftMaxRowCount != null && leftMaxRowCount <= 1.0) {
-      rightColumns = rightColumns.union(joinInfo.rightSet());
+    if (rel.getJoinType() == JoinRelType.INNER) {
+      // Joining with a singleton constrains the keys on the other table
+      final Double rightMaxRowCount = mq.getMaxRowCount(right);
+      if (rightMaxRowCount != null && rightMaxRowCount <= 1.0) {
+        leftColumns = leftColumns.union(joinInfo.leftSet());
+      }
+      final Double leftMaxRowCount = mq.getMaxRowCount(left);
+      if (leftMaxRowCount != null && leftMaxRowCount <= 1.0) {
+        rightColumns = rightColumns.union(joinInfo.rightSet());
+      }
     }
 
     // If the original column mask contains columns from both the left and
--- a/core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
@@ -1193,6 +1193,48 @@ public class RelMetadataTest {
         .assertThatUniqueKeysAre(bitSetOf());
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6727">[CALCITE-6727]
+   * Column uniqueness constrain should only apply to inner join</a>. */
+  @Test void testColumnUniquenessForLeftJoinOnLimit1() {
+    final String sql = ""
+        + "select A.empno as a_empno,\n"
+        + " A.ename as a_ename,\n"
+        + " B.empno as b_empno,\n"
+        + " B.ename as b_ename\n"
+        + "from emp A\n"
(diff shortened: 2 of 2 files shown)


mihaibudiu (commented):
Why not add the test?

suibianwanwank (approved):
LGTM

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -1193,6 +1193,24 @@ private void checkColumnUniquenessForFilterWithConstantColumns(String sql) {
         .assertThatUniqueKeysAre(bitSetOf());
   }
 
+  @Test void testColumnUniquenessForLeftJoinOnLimit1() {
```
mihaibudiu, line 1196: it is customary to add a JavaDoc comment to the test indicating the JIRA issue that is being addressed. This makes it easier for maintainers to understand the rationale for a test. You can find lots of examples in the code about how that is supposed to be structured, please follow the pattern.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
```diff
@@ -321,15 +321,16 @@ public Boolean areColumnsUnique(Intersect rel, RelMetadataQuery mq,
...
-      leftColumns = leftColumns.union(joinInfo.leftSet());
-    }
-    final Double leftMaxRowCount = mq.getMaxRowCount(left);
-    if (leftMaxRowCount != null && leftMaxRowCount <= 1.0) {
-      rightColumns = rightColumns.union(joinInfo.rightSet());
+    if (rel.getJoinType() == JoinRelType.INNER) {
+      // Joining with a singleton constrains the keys on the other table
+      final Double rightMaxRowCount = mq.getMaxRowCount(right);
+      if (rightMaxRowCount != null && rightMaxRowCount <= 1.0) {
+        leftColumns = leftColumns.union(joinInfo.leftSet());
```
mihaibudiu, line 328: is this correct even if maxRowCount is 0?

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
joeyutong: Added the comment. Thanks for your reminder

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnUniqueness.java
joeyutong: It seems to me that the constrain still works even when the build side is empty

suibianwanwank: +1

joeyutong: @mihaibudiu Any more thoughts on this? Please take another look when you have a moment

mihaibudiu (approved):

ILuffZhe:
Hi @joeyutong , just a little effort and I think this PR can be merged.
We can `git commit --allow-empty` and then rebase it. I'll merge it as long as CI passes.

ILuffZhe:
It seems Jenkins failed again. I check the logs and it's not code relevant.
Do you have any idea? @mihaibudiu

mihaibudiu:
I think you can ignore Jenkins. It fails non deterministically

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -1193,6 +1193,27 @@ private void checkColumnUniquenessForFilterWithConstantColumns(String sql) {
...
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6727">[CALCITE-6727]
+   * Column uniqueness constrain should only apply to inner join</a>. */
+  @Test void testColumnUniquenessForLeftJoinOnLimit1() {
+    final String sql = ""
+        + "select A.empno as a_empno,\n"
+        + " A.ename as a_ename,\n"
+        + " B.empno as b_empno,\n"
+        + " B.ename as b_ename\n"
```
ILuffZhe, line 1204: Sorry for asking this, should `b_empno` be unique if this case is `right join`?
If yes, do you mind adding a test?

joeyutong: I have added the corresponding test. Thanks for you feedback.

PR #4090, opened 2024-12-13
[CALCITE-6664] Replace GREATEST, LEAST functions in Spark library wit…

joeyutong:
`GREATEST` or `LEAST` function in Spark library returns null if any of its parameters are null, it is not consistent with the description of Spark [docs](https://spark.apache.org/docs/latest/api/sql/index.html#greatest)

> Returns the greatest value of all parameters, skipping null values.

This PR tries to replacement the implementation of these functions with the implementation in PostgreSQL Library

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlKind.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlKind.java
@@ -475,10 +475,10 @@ public enum SqlKind {
   /** {@code NVL2} function (Oracle, Spark). */
   NVL2,
 
-  /** {@code GREATEST} function (Oracle, Spark). */
+  /** {@code GREATEST} function (Oracle). */
   GREATEST,
 
-  /** {@code GREATEST} function (PostgreSQL). */
+  /** {@code GREATEST} function (PostgreSQL, Spark). */
   GREATEST_PG,
 
   /** The two-argument {@code CONCAT} function (Oracle). */
@@ -503,7 +503,7 @@ public enum SqlKind {
   /** {@code LEAST} function (Oracle). */
   LEAST,
 
-  /** {@code LEAST} function (PostgreSQL). */
+  /** {@code LEAST} function (PostgreSQL, Spark). */
   LEAST_PG,
 
   /** {@code LOG} function. (Mysql, Spark). */
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -11664,7 +11664,7 @@ public class SqlOperatorTest {
           "VARCHAR(5) NOT NULL");
     };
     final List<SqlLibrary> libraries =
-        list(SqlLibrary.BIG_QUERY, SqlLibrary.ORACLE, SqlLibrary.SPARK);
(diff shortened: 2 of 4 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -11646,7 +11646,8 @@ void assertSubFunReturns(boolean binary, String s, int start,
 
   @Test void testGreatestFunc() {
     final SqlOperatorFixture f0 =
-        fixture().setFor(SqlLibraryOperators.GREATEST, VmName.EXPAND);
+        Fixtures.forOperators(true)
```
NobiGo, line 11649: Why do we need to change this?

NobiGo (commented):
If we don't change the SqlKind, this issue will become simple.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
joeyutong: To actually execute the call and check the result, not just validate the call. It makes more sense to me to test the function like `greatest`.

mihaibudiu: the operators execute in CalciteSqlOperatorTest

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -475,11 +475,11 @@ public enum SqlKind {
...
   NVL2,
 
-  /** {@code GREATEST} function (Oracle, Spark). */
+  /** {@code GREATEST} function (Oracle). */
   GREATEST,
 
-  /** {@code GREATEST} function (PostgreSQL). */
-  GREATEST_PG,
+  /** {@code GREATEST} function (PostgreSQL, Spark). */
+  GREATEST_SKIP_NULL,
```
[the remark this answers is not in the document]
joeyutong: Revert the changes. Please take another look

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -11732,7 +11733,8 @@ void assertSubFunReturns(boolean binary, String s, int start,
       f.checkScalar("least(CAST(NULL AS INTEGER), CAST(NULL AS INTEGER))", isNullValue(),
           "INTEGER");
     };
-    final List<SqlLibrary> libraries = list(SqlLibrary.POSTGRESQL, SqlLibrary.REDSHIFT);
+    final List<SqlLibrary> libraries =
+        list(SqlLibrary.POSTGRESQL, SqlLibrary.REDSHIFT, SqlLibrary.SPARK);
```
NobiGo, line 11737: Why `SqlLibrary.REDSHIFT` works? I noticed before this PR, it already has it.

joeyutong: It inherits functions from PG
```
// SqlLibrary.java
REDSHIFT("r", "redshift", POSTGRESQL)
```

NobiGo (approved):
LGTM

mihaibudiu:
please squash the commits so we can merge this

NobiGo:
Commit Info needs to be consistent with Jira Summary and I think Jira summary is good.

PR #4092, opened 2024-12-14
[CALCITE-6690] Refactor the Arrow adapter type system

caicancai:


arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
```diff
@@ -17,84 +17,69 @@
...
-  private final Class<?> clazz;
-
-  ArrowFieldType(Primitive primitive) {
-    this(requireNonNull(primitive.boxClass, "boxClass"));
-  }
+public class ArrowFieldType {
 
-  ArrowFieldType(Class<?> clazz) {
-    this.clazz = clazz;
+  private ArrowFieldType() {
```
mihaibudiu, line 31: this looks like a factory, so the name of the class is probably not appropriate.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowData.java
```diff
@@ -237,35 +239,38 @@ public void writeArrowDataType(File file) throws IOException {
       vectorSchemaRoot.setRowCount(numRows);
       for (Field field : vectorSchemaRoot.getSchema().getFields()) {
         FieldVector vector = vectorSchemaRoot.getVector(field.getName());
-        switch (vector.getMinorType()) {
-        case TINYINT:
+        switch (field.getName()) {
```
mihaibudiu, line 242: why change these?

caicancai: Sorry for taking so long to reply. 
This is because I set two fields for decimal type, one with scale 2 and one with scale 3. Since there are two decimal types, it is no longer appropriate to distinguish based on data type.

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
caicancai: Do you need me to optimize it?

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowData.java
mihaibudiu: So, this is just test code. Can you change the file name to ArrowTestData?

arrow/src/main/java/org/apache/calcite/adapter/arrow/ArrowFieldType.java
mihaibudiu: you don't need to optimize anything, just use a name that is appropriate. This is not a type, it is a class which creates types, so it is a factory.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowData.java
caicancai: @mihaibudiu thanks, done

caicancai:
I'm sorry I messed up the branches, I'll make a new PR

PR #4093, opened 2024-12-16
[CALCITE-6733] Type inferred by coercion for comparisons with decimal is too narrow

mihaibudiu:
The current implementation is based on the rules for inferring the result for addition (but does not require an extra digit).

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
@@ -612,17 +612,14 @@ public abstract class AbstractTypeCoercion implements TypeCoercion {
     }
 
     if (SqlTypeUtil.isExactNumeric(type1) && SqlTypeUtil.isExactNumeric(type2)) {
-      if (SqlTypeUtil.isDecimal(type1)) {
-        // Use max precision
+      if (SqlTypeUtil.isDecimal(type1) || SqlTypeUtil.isDecimal(type2)) {
+        // Precision used must be large enough to fit either of the types
+        int maxScale = Math.max(type1.getScale(), type2.getScale());
         RelDataType result =
-            factory.createSqlType(type1.getSqlTypeName(),
-                Math.max(type1.getPrecision(), type2.getPrecision()), type1.getScale());
-        return factory.createTypeWithNullability(result, type1.isNullable() || type2.isNullable());
-      } else if (SqlTypeUtil.isDecimal(type2)) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
```diff
@@ -612,17 +612,14 @@ private RelDataType getTightestCommonTypeOrThrow(
...
-        RelDataType result =
-            factory.createSqlType(type2.getSqlTypeName(),
-                Math.max(type1.getPrecision(), type2.getPrecision()), type2.getScale());
+            factory.createSqlType(SqlTypeName.DECIMAL,
+                Math.max(type1.getPrecision() - type1.getScale(),
+                         type2.getPrecision() - type2.getScale()) + maxScale,
+                maxScale);
         return factory.createTypeWithNullability(result, type1.isNullable() || type2.isNullable());
       }
       if (type1.getPrecision() > type2.getPrecision()) {
```
NobiGo, line 625: Can this handle types without precision and proportion? For example:
TINYINT、SMALLINT、INTEGER and BIGINT? Or the default behavior is right.
If type1 is TINYINT type2 is SMALLINT, then the common type becomes SMALLINT. This is right.
But when type1 is SMALLINT and  type2 is TINYINT. then the common type becomes TINYINT. This should be unreasonable.

mihaibudiu: this should be completely symmetric in type1 and type2.

mihaibudiu: BTW: all integer types have a positive precision and a scale or zero.

NobiGo: According to SqlTypeName, TINYINT、SMALLINT、INTEGER, and BIGINT don't have Precision and Scale.

mihaibudiu: See RelDataTypeSystemImpl.getDefaultPrecision

NobiGo: But the `RelDataTypeSystemImpl#getDefaultPrecision` has a default precision for this. It is right. I misunderstood.

NobiGo (approved):
LGTM

PR #4095, opened 2024-12-17
[CALCITE-6734] RelFieldTrimmer should trim Aggregate's input fields which are arguments of unused aggregate functions

ViggoC:
Before
```
LogicalAggregate(group=[{3}], SAL=[SUM($2)])
  LogicalFilter(condition=[>($3, 100)])
    LogicalProject(EMPNO=[$0], ENAME=[$1], SAL=[$5], DEPTNO=[$7])
```

After
```
LogicalAggregate(group=[{2}], SAL=[SUM($1)])
  LogicalFilter(condition=[>($2, 100)])
    LogicalProject(EMPNO=[$0], SAL=[$5], DEPTNO=[$7])
```

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
@@ -1054,16 +1054,20 @@ public class RelFieldTrimmer implements ReflectiveVisitor {
     // 1. group fields are always used
     final ImmutableBitSet.Builder inputFieldsUsed =
         aggregate.getGroupSet().rebuild();
-    // 2. agg functions
+    // 2. agg functions: consider only the ones that are needed according to fieldsUsed
+    int aggCallIndex = aggregate.getGroupCount();
     for (AggregateCall aggCall : aggregate.getAggCallList()) {
-      inputFieldsUsed.addAll(aggCall.getArgList());
-      if (aggCall.filterArg >= 0) {
-        inputFieldsUsed.set(aggCall.filterArg);
+      if (fieldsUsed.get(aggCallIndex)) {
+        inputFieldsUsed.addAll(aggCall.getArgList());
+        if (aggCall.filterArg >= 0) {
+          inputFieldsUsed.set(aggCall.filterArg);
+        }
+        if (aggCall.distinctKeys != null) {
+          inputFieldsUsed.addAll(aggCall.distinctKeys);
+        }
+        inputFieldsUsed.addAll(RelCollations.ordinals(aggCall.collation));
       }
-      if (aggCall.distinctKeys != null) {
-        inputFieldsUsed.addAll(aggCall.distinctKeys);
-      }
-      inputFieldsUsed.addAll(RelCollations.ordinals(aggCall.collation));
+      aggCallIndex++;
     }
 
     // Create input with trimmed columns.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -1055,15 +1055,19 @@ public TrimResult trimFields(
...
     // 2. agg functions
+    int index = aggregate.getGroupCount();
     for (AggregateCall aggCall : aggregate.getAggCallList()) {
-      inputFieldsUsed.addAll(aggCall.getArgList());
-      if (aggCall.filterArg >= 0) {
-        inputFieldsUsed.set(aggCall.filterArg);
-      }
-      if (aggCall.distinctKeys != null) {
-        inputFieldsUsed.addAll(aggCall.distinctKeys);
+      if (fieldsUsed.get(index)) {
```
ILuffZhe, line 1060: I'm not familiar here, just one small question: Why do the index start with aggregate's groupCount? In this case, if the top Project contains ENAME, can it be covered?

ILuffZhe (approved):
Overall looks good.

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
ViggoC: Because all group by keys are necessary in aggregate no matter if they are used in the consumer.  If the top Project contains ENAME, `fieldsUsed.get(indexOfEname)` will be true.

ILuffZhe: Thanks for the clarify

rubenada: I agree with @ILuffZhe that the new code, while correct, is not straightforward to understand, so perhaps a comment will be useful here. I'd suggest maybe updating the comment line 1057 with something like:
`// 2. agg functions: consider only the ones that are needed according to fieldsUsed`

core/src/main/java/org/apache/calcite/sql2rel/RelFieldTrimmer.java
```diff
@@ -1055,15 +1055,19 @@ public TrimResult trimFields(
     final ImmutableBitSet.Builder inputFieldsUsed =
         aggregate.getGroupSet().rebuild();
     // 2. agg functions
+    int index = aggregate.getGroupCount();
```
rubenada, line 1058: nit: I think it would be more clear if this variable was named something like `int aggCallIndex`

ViggoC:
@rubenada Actually, I think **aggregate functions that are not used by the consumer** is already trimmed. What this MR do is trimming the unused input fields of aggregate function.

rubenada:
@ViggoC I see, what about "[CALCITE-6734] RelFieldTrimmer should trim Aggregate's input fields which are arguments of unused aggregate functions" ?

ViggoC:
@rubenada Make sense, I'll revise it. Thx

ILuffZhe:
@ViggoC  you don't need to squash your commits every time before the final merging. And it helps other reviewers to see the differences, either.

rubenada:
@ViggoC LGTM. 
There is a checkStyle error (line too long) to be fixed.

ViggoC:
@ILuffZhe Thank you for the reminder.

rubenada (approved):

rubenada:
LGTM. @ViggoC  could you please squash commits into a single one to prepare the merge (I plan to proceed on the next 24-48h if no further remark arrives)

PR #4096, opened 2024-12-18
[CALCITE-6735] Type coercion for comparisons does not coerce ROW types

mihaibudiu:
I also had to implement in the RexToLixTranslator casts between ROW types.

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
@@ -32,6 +32,7 @@ import org.apache.calcite.linq4j.tree.ParameterExpression;
 import org.apache.calcite.linq4j.tree.Primitive;
 import org.apache.calcite.linq4j.tree.Statement;
 import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexCallBinding;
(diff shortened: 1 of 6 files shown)


mihaibudiu:
The CI seems finished, just not detected properly by the UI.

mihaibudiu:
In practical terms, what this PR does is to convert a comparison of ROW values with different types to a common type using casts.

core/src/main/java/org/apache/calcite/sql/validate/implicit/AbstractTypeCoercion.java
```diff
@@ -629,6 +630,28 @@ private RelDataType getTightestCommonTypeOrThrow(
...
+      if (typeName2 != typeName1) {
+        return null;
+      }
+      if (type1.getFieldCount() != type2.getFieldCount()) {
+        return null;
+      }
+      List<RelDataTypeField> leftFields = type1.getFieldList();
+      List<RelDataTypeField> rightFields = type2.getFieldList();
+      List<RelDataTypeField> resultFields = new ArrayList<>();
+      for (int i = 0; i < leftFields.size(); i++) {
```
NobiGo, line 643: Need to confirm the list size is equal?

mihaibudiu: That is done a few lines above

mihaibudiu: @NobiGo that's line 675, comparing getFieldCount().

NobiGo (approved):

core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -331,6 +332,35 @@ private Expression getConvertExpression(
       return Expressions.convert_(cast, typeFactory.getJavaClass(nullableTarget));
     }
 
+    if (targetType.getSqlTypeName() == SqlTypeName.ROW) {
+      assert sourceType.getSqlTypeName() == SqlTypeName.ROW;
+      List<RelDataTypeField> targetTypes = targetType.getFieldList();
+      List<RelDataTypeField> sourceTypes = sourceType.getFieldList();
+      assert targetTypes.size() == sourceTypes.size();
```
caicancai, line 339: Why do we need to judge this? The sizes of targettype and sourcetype may be different?

mihaibudiu: The validator will reject such programs.

PR #4098, opened 2024-12-18
[CALCITE-6736] Validator accepts comparisons between arrays, multiset…

mihaibudiu:
…s, maps without regard to element types


core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -501,6 +501,19 @@ static SqlOperatorTable operatorTableFor(SqlLibrary library) {
     expr("^x'abcd'<>1^")
         .fails("(?s).*Cannot apply '<>' to arguments of type "
             + "'<BINARY.2.> <> <INTEGER>'.*");
+    // Test cases for [CALCITE-6736] Validator accepts comparisons between arrays, multisets, maps
+    // without regard to element types
+    expr("^array[x'a4'] = array[1]^")
```
ILuffZhe, line 506: Do we need to add some positive tests for those types(like array[2]= array[1])?

ILuffZhe (approved):
Overall looks good, just one minor question.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
mihaibudiu: I assumed such tests exists, but didn't look too much. This change only rejects some programs that used to be accepted, so I really focused on negative tests.

NobiGo (approved):

PR #4099, opened 2024-12-19
[CALCITE-6665] Add isEmpty metadata to check if a relational expressi…

NobiGo:
…on returns no rows

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMdUtil.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMdUtil.java
@@ -210,6 +210,18 @@ public class RelMdUtil {
     return b != null && b;
   }
 
+  public static boolean isRelDefinitelyEmpty(RelMetadataQuery mq,
+      RelNode rel) {
+    Boolean b = mq.isEmpty(rel);
+    return b != null && b;
+  }
+
+  public static boolean isRelDefinitelyNotEmpty(RelMetadataQuery mq,
+      RelNode rel) {
+    Boolean b = mq.isEmpty(rel);
+    return b != null && !b;
+  }
+
   public static @Nullable Boolean areColumnsUnique(RelMetadataQuery mq, RelNode rel,
       List<RexInputRef> columnRefs) {
     ImmutableBitSet.Builder colMask = ImmutableBitSet.builder();
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelMetadataQuery.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelMetadataQuery.java
@@ -293,7 +293,7 @@ public class RelMetadataQuery extends RelMetadataQueryBase {
    * statistic.
    *
    * @param rel the relational expression
-   * @return max row count
+   * @return min row count
    */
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/apache/calcite/rel/metadata/RelMdUtil.java
```diff
@@ -210,6 +210,18 @@ public static boolean areColumnsDefinitelyUnique(RelMetadataQuery mq,
     return b != null && b;
   }
 
+  public static boolean isRelDefinitelyEmpty(RelMetadataQuery mq,
+      RelNode rel) {
+    Boolean b = mq.isEmpty(rel);
+    return b != null && b;
+  }
+
+  public static boolean isRelDefinitelyNotEmpty(RelMetadataQuery mq,
```
ILuffZhe, line 219: Can this method be replaced by "!isRelDefinitelyEmpty"?

core/src/main/java/org/apache/calcite/rel/metadata/RelMetadataQuery.java
```diff
@@ -305,6 +305,26 @@ public static RelMetadataQuery instance() {
     }
   }
 
+  /**
+   * Returns whether the return rows of a given relational expression are empty.
```
ILuffZhe, line 309: nit: "a given relational expression is empty"

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
```diff
@@ -2678,10 +2679,7 @@ private boolean alreadyUnique(List<AggCallPlus> aggCallList,
       List<RexNode> extraNodes) {
     final RelMetadataQuery mq = peek().getCluster().getMetadataQuery();
     if (aggCallList.isEmpty() && groupSet.isEmpty()) {
-      final Double minRowCount = mq.getMinRowCount(peek());
-      if (minRowCount == null || minRowCount < 1d) {
-        // We can't remove "GROUP BY ()" if there's a chance the rel could be
```
ILuffZhe, line 2683: Why delete the doc?

core/src/main/java/org/apache/calcite/rel/metadata/RelMdUtil.java
NobiGo: It can't because the isEmpty value can be null.

core/src/main/java/org/apache/calcite/tools/RelBuilder.java
NobiGo: Added it.

core/src/main/java/org/apache/calcite/rel/metadata/RelMetadataQuery.java
NobiGo: Let's see if there are any other opinions

ILuffZhe: My bad, the doc is OK.

ILuffZhe (approved):
Looks good! Thanks for the clarify.

caicancai (approved):

PR #4100, opened 2024-12-19
[CALCITE-6728] Introduce new methods to lookup tables and schemas inside schemas

kramerul:
### Motivation

For databases with a huge set of schemas and tables it takes quite long to prepare queries. Currently all tables/schemas are loaded into memory.

Caching all these schemas and tables is not an option
1. It will require a lot of memory
2. The eviction of the cache must happen quite often since it's likely that every second one of these table is changed.

Therefore, we tried to find a way to load only those tables/schemas, which are required to prepare a query.

### API Changes

This PR introduces a new mechanism to lookup tables and schemas within a schema. For this purpose a new interface is introduced

```java
public interface Lookup<T> {
  @Nullable T get(String name);
  @Nullable Named<T> getIgnoreCase(String name);
  Set<String> getNames(LikePattern pattern);
}
```

The `LikePattern` was extracted from `CalciteMetaImpl` to hold a pattern, which can be used to query tables and schemas inside a JDBC database using the `LIKE` operator. Additionally, it also supports the conversion to a `Predicate1<String>` which can be used to implement filters in plain java.

The `Schema` is now using this `Lookup` interface to find schemas and tables. It could be also extended to functions and types.

```java
public interface Schema {
  default Lookup<Table> tables() {
    ...
  }
  default Lookup<? extends Schema> subSchemas() {
    ...
  }
  ...
}
```

### Implementation

The case insensitive search is now directly implemented in the specific `Schema` using matching implementation of the `Lookup` interface. Formerly, it was done in the `CalciteSchema`.

`JdbcSchema` and `JdbcCatalogSchema` are using a special implementation of `Lookup`: `LoadingCacheLookup`. This implementation is using a `LoadingCache` inside to speed up things. If only case sensitive schema/table lookup is required, this can be done quite fast since `DatabaseMetaData#getTables` can be used to query a single table. The result is cached inside the `LoadingCache` for one minute.

Unfortunately `DatabaseMetaData#getTables` doesn't support case insensitive queries. In this case, it's still required to load all database tables to perform case insensitive lookups.

The performance gain for huge sets of tables/schemas in database schemas can only be achieved if caching is turned off in Calcite (`SimpleCalciteSchema` is used instead of `CachingCalciteSchema`). 

I tried to keep the behavior of `CachingCalciteSchema` exactly the same. This behavior includes that all tables/schemas are loaded into memory.  `CachedLookup` is used to achieve this.

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcBaseSchema.java
+++ b/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcBaseSchema.java
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.adapter.jdbc;
+
+import org.apache.calcite.linq4j.tree.Expression;
+import org.apache.calcite.rel.type.RelProtoDataType;
+import org.apache.calcite.schema.Function;
+import org.apache.calcite.schema.Schema;
+import org.apache.calcite.schema.SchemaPlus;
+import org.apache.calcite.schema.SchemaVersion;
+import org.apache.calcite.schema.Schemas;
+import org.apache.calcite.schema.Table;
+import org.apache.calcite.schema.lookup.LikePattern;
+import org.apache.calcite.schema.lookup.Lookup;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Base class for JDBC schemas.
+ */
+public abstract class JdbcBaseSchema implements Schema {
+
+  @Override public abstract Lookup<Table> tables();
+
+
+  @Deprecated @Override public @Nullable Table getTable(String name) {
+    return tables().get(name);
+  }
+
+  @Deprecated @Override public Set<String> getTableNames() {
+    return tables().getNames(LikePattern.any());
+  }
+
+  @Override public abstract Lookup<? extends Schema> subSchemas();
+
+  @Deprecated @Override public @Nullable Schema getSubSchema(String name) {
+    return subSchemas().get(name);
+  }
+
+  @Deprecated @Override public Set<String> getSubSchemaNames() {
+    return subSchemas().getNames(LikePattern.any());
+  }
+
+
+  @Override public @Nullable RelProtoDataType getType(String name) {
+    return null;
+  }
+
+  @Override public Set<String> getTypeNames() {
+    return Collections.emptySet();
+  }
+
+  @Override public final Collection<Function> getFunctions(String name) {
+    return Collections.emptyList();
+  }
+
+  @Override public final Set<String> getFunctionNames() {
+    return Collections.emptySet();
+  }
+
+  @Override public Expression getExpression(final @Nullable SchemaPlus parentSchema,
+      final String name) {
+    requireNonNull(parentSchema, "parentSchema");
+    return Schemas.subSchemaExpression(parentSchema, name, getClass());
+  }
+
+  @Override public boolean isMutable() {
+    return false;
+  }
+
+  @Override public Schema snapshot(final SchemaVersion version) {
+    return this;
+  }
+}
--- a/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCatalogSchema.java
+++ b/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCatalogSchema.java
@@ -22,22 +22,27 @@ import org.apache.calcite.linq4j.tree.Expressions;
 import org.apache.calcite.schema.Schema;
 import org.apache.calcite.schema.SchemaPlus;
 import org.apache.calcite.schema.Schemas;
+import org.apache.calcite.schema.Table;
 import org.apache.calcite.schema.Wrapper;
-import org.apache.calcite.schema.impl.AbstractSchema;
+import org.apache.calcite.schema.lookup.IgnoreCaseLookup;
+import org.apache.calcite.schema.lookup.LikePattern;
+import org.apache.calcite.schema.lookup.LoadingCacheLookup;
+import org.apache.calcite.schema.lookup.Lookup;
 import org.apache.calcite.sql.SqlDialect;
 import org.apache.calcite.sql.SqlDialectFactory;
 import org.apache.calcite.sql.SqlDialectFactoryImpl;
 import org.apache.calcite.util.BuiltInMethod;
 
 import com.google.common.base.Suppliers;
-import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
 import java.util.function.Supplier;
 import javax.sql.DataSource;
 
@@ -51,20 +56,21 @@ import static java.util.Objects.requireNonNull;
  * an instance of {@link JdbcSchema}.
  *
  * <p>This schema is lazy: it does not compute the list of schema names until
- * the first call to {@link #getSubSchemaMap()}. Then it creates a
- * {@link JdbcSchema} for each schema name. Each JdbcSchema will populate its
+ * the first call to {@link #subSchemas()} and {@link Lookup#get(String)}. Then it creates a
+ * {@link JdbcSchema} for this schema name. Each JdbcSchema will populate its
  * tables on demand.
  */
-public class JdbcCatalogSchema extends AbstractSchema implements Wrapper {
+public class JdbcCatalogSchema extends JdbcBaseSchema implements Wrapper {
   final DataSource dataSource;
   public final SqlDialect dialect;
   final JdbcConvention convention;
   final String catalog;
+  private final Lookup<JdbcSchema> subSchemas;
 
-  /** Sub-schemas by name, lazily initialized. */
+  /** default schema name, lazily initialized. */
   @SuppressWarnings({"method.invocation.invalid", "Convert2MethodRef"})
-  final Supplier<SubSchemaMap> subSchemaMapSupplier =
-      Suppliers.memoize(() -> computeSubSchemaMap());
+  private final Supplier<Optional<String>> defaultSchemaName =
+      Suppliers.memoize(() -> Optional.ofNullable(computeDefaultSchemaName()));
 
   /** Creates a JdbcCatalogSchema. */
   public JdbcCatalogSchema(DataSource dataSource, SqlDialect dialect,
@@ -73,6 +79,40 @@ public class JdbcCatalogSchema extends AbstractSchema implements Wrapper {
     this.dialect = requireNonNull(dialect, "dialect");
     this.convention = requireNonNull(convention, "convention");
     this.catalog = catalog;
+    this.subSchemas = new LoadingCacheLookup<>(new IgnoreCaseLookup<JdbcSchema>() {
+      @Override public @Nullable JdbcSchema get(String name) {
+        try (Connection connection = dataSource.getConnection();
+            ResultSet resultSet =
+                connection.getMetaData().getSchemas(catalog, name)) {
+          while (resultSet.next()) {
+            final String schemaName =
+                requireNonNull(resultSet.getString(1),
+                    "got null schemaName from the database");
+            return new JdbcSchema(dataSource, dialect, convention, catalog, schemaName);
+          }
+        } catch (SQLException e) {
+          throw new RuntimeException(e);
+        }
+        return null;
+      }
+
+      @Override public Set<String> getNames(LikePattern pattern) {
+        final ImmutableSet.Builder<String> builder =
+            ImmutableSet.builder();
+        try (Connection connection = dataSource.getConnection();
+            ResultSet resultSet =
+                connection.getMetaData().getSchemas(catalog, pattern.pattern)) {
+          while (resultSet.next()) {
+            builder.add(
+                requireNonNull(resultSet.getString(1),
+                    "got null schemaName from the database"));
+          }
+        } catch (SQLException e) {
+          throw new RuntimeException(e);
+        }
+        return builder.build();
+      }
+    });
   }
 
   public static JdbcCatalogSchema create(
@@ -103,34 +143,25 @@ public class JdbcCatalogSchema extends AbstractSchema implements Wrapper {
     return new JdbcCatalogSchema(dataSource, dialect, convention, catalog);
   }
 
-  private SubSchemaMap computeSubSchemaMap() {
-    final ImmutableMap.Builder<String, Schema> builder =
-        ImmutableMap.builder();
-    @Nullable String defaultSchemaName;
-    try (Connection connection = dataSource.getConnection();
-         ResultSet resultSet =
-             connection.getMetaData().getSchemas(catalog, null)) {
-      defaultSchemaName = connection.getSchema();
-      while (resultSet.next()) {
-        final String schemaName =
-            requireNonNull(resultSet.getString(1),
-                "got null schemaName from the database");
-        builder.put(schemaName,
-            new JdbcSchema(dataSource, dialect, convention, catalog, schemaName));
-      }
+  @Override public Lookup<Table> tables() {
+    return Lookup.empty();
+  }
+
+  @Override public Lookup<? extends Schema> subSchemas() {
+    return subSchemas;
+  }
+
+  private @Nullable String computeDefaultSchemaName() {
+    try (Connection connection = dataSource.getConnection()) {
+      return connection.getSchema();
     } catch (SQLException e) {
       throw new RuntimeException(e);
     }
-    return new SubSchemaMap(defaultSchemaName, builder.build());
-  }
-
-  @Override protected Map<String, Schema> getSubSchemaMap() {
-    return subSchemaMapSupplier.get().map;
   }
 
   /** Returns the name of the default sub-schema. */
   public @Nullable String getDefaultSubSchemaName() {
-    return subSchemaMapSupplier.get().defaultSchemaName;
+    return defaultSchemaName.get().orElse(null);
   }
 
   /** Returns the data source. */
@@ -147,17 +178,5 @@ public class JdbcCatalogSchema extends AbstractSchema implements Wrapper {
       return clazz.cast(getDataSource());
     }
     return null;
-  }
-
-  /** Contains sub-schemas by name, and the name of the default schema. */
-  private static class SubSchemaMap {
-    final @Nullable String defaultSchemaName;
-    final ImmutableMap<String, Schema> map;
-
-    private SubSchemaMap(@Nullable String defaultSchemaName,
-        ImmutableMap<String, Schema> map) {
-      this.defaultSchemaName = defaultSchemaName;
-      this.map = map;
-    }
   }
 }
--- a/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
+++ b/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
@@ -26,7 +26,6 @@ import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeImpl;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
 import org.apache.calcite.rel.type.RelProtoDataType;
-import org.apache.calcite.schema.Function;
 import org.apache.calcite.schema.Schema;
 import org.apache.calcite.schema.SchemaFactory;
 import org.apache.calcite.schema.SchemaPlus;
@@ -34,19 +33,19 @@ import org.apache.calcite.schema.SchemaVersion;
 import org.apache.calcite.schema.Schemas;
 import org.apache.calcite.schema.Table;
 import org.apache.calcite.schema.Wrapper;
+import org.apache.calcite.schema.lookup.IgnoreCaseLookup;
+import org.apache.calcite.schema.lookup.LikePattern;
+import org.apache.calcite.schema.lookup.Lookup;
 import org.apache.calcite.sql.SqlDialect;
 import org.apache.calcite.sql.SqlDialectFactory;
 import org.apache.calcite.sql.SqlDialectFactoryImpl;
 import org.apache.calcite.sql.type.SqlTypeFactoryImpl;
 import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.util.LazyReference;
 import org.apache.calcite.util.Pair;
 import org.apache.calcite.util.Util;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableMultimap;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Multimap;
 import com.google.common.collect.Ordering;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
@@ -58,13 +57,17 @@ import java.sql.DatabaseMetaData;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.ArrayList;
-import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.Spliterator;
+import java.util.Spliterators;
 import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 import javax.sql.DataSource;
 
 import static java.lang.Integer.parseInt;
@@ -77,7 +80,7 @@ import static java.util.Objects.requireNonNull;
  * queries against this schema are executed against those tables, pushing down
  * as much as possible of the query logic to SQL.
  */
-public class JdbcSchema implements Schema, Wrapper {
+public class JdbcSchema extends JdbcBaseSchema implements Schema, Wrapper {
   private static final Logger LOGGER = LoggerFactory.getLogger(JdbcSchema.class);
 
   final DataSource dataSource;
@@ -85,8 +88,8 @@ public class JdbcSchema implements Schema, Wrapper {
   final @Nullable String schema;
   public final SqlDialect dialect;
   final JdbcConvention convention;
-  private @Nullable ImmutableMap<String, JdbcTable> tableMap;
-  private final boolean snapshot;
+  private final LazyReference<Lookup<Table>> tables = new LazyReference<>();
+  private final Lookup<JdbcSchema> subSchemas = Lookup.empty();
 
   @Experimental
   public static final ThreadLocal<@Nullable Foo> THREAD_METADATA = new ThreadLocal<>();
@@ -105,19 +108,11 @@ public class JdbcSchema implements Schema, Wrapper {
    */
   public JdbcSchema(DataSource dataSource, SqlDialect dialect,
       JdbcConvention convention, @Nullable String catalog, @Nullable String schema) {
-    this(dataSource, dialect, convention, catalog, schema, null);
-  }
-
-  private JdbcSchema(DataSource dataSource, SqlDialect dialect,
-      JdbcConvention convention, @Nullable String catalog, @Nullable String schema,
-      @Nullable ImmutableMap<String, JdbcTable> tableMap) {
     this.dataSource = requireNonNull(dataSource, "dataSource");
     this.dialect = requireNonNull(dialect, "dialect");
     this.convention = convention;
     this.catalog = catalog;
     this.schema = schema;
-    this.tableMap = tableMap;
-    this.snapshot = tableMap != null;
   }
 
   public static JdbcSchema create(
@@ -218,13 +213,33 @@ public class JdbcSchema implements Schema, Wrapper {
         password);
   }
 
+  @Override public Lookup<Table> tables() {
+    return tables.getOrCompute(() -> new IgnoreCaseLookup<Table>() {
+      @Override public @Nullable Table get(String name) {
+        try (Stream<MetaImpl.MetaTable> s = getMetaTableStream(name)) {
+          return s.findFirst().map(it -> jdbcTableMapper(it)).orElse(null);
+        }
+      }
+
+      @Override public Set<String> getNames(LikePattern pattern) {
+        try (Stream<MetaImpl.MetaTable> s = getMetaTableStream(pattern.pattern)) {
+          return s.map(it -> it.tableName).collect(Collectors.toSet());
+        }
+      }
+    });
+  }
+
+  @Override public Lookup<? extends Schema> subSchemas() {
+    return subSchemas;
+  }
+
+
   @Override public boolean isMutable() {
     return false;
   }
 
   @Override public Schema snapshot(SchemaVersion version) {
-    return new JdbcSchema(dataSource, dialect, convention, catalog, schema,
-        tableMap);
+    return this;
   }
 
   // Used by generated code.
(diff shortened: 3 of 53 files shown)


kramerul:
We are still interested in getting this PR merged.

mihaibudiu:
I will try to review this, although it's in an area where I don't know much about the codebase.

kramerul:
I know that this PR is quite huge. I discussed it with Julian Hyde, if it makes sense to open such a PR. For details see https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6728

core/src/main/java/org/apache/calcite/util/LazyReference.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * This class can be used to lazily initialize an object.
+ *
+ * @param <T> Element Type
+ */
+public class LazyReference<T> {
+
+  private final AtomicReference<T> value = new AtomicReference<>();
+
+  public T getOrCompute(Supplier<T> supplier) {
```
mihaibudiu, line 31: maybe the code is obvious, but I would still document the public API.
Since this is about concurrency, you should add in the comments a description of what this promises with respect to concurrent accesses.

core/src/main/java/org/apache/calcite/schema/lookup/Named.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+package org.apache.calcite.schema.lookup;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * This class is used to hold an object including its name.
```
mihaibudiu, line 25: I deduce that names are strings - case sensitive.
I would have done it the other way, having an INamed interface which can be implemented by various objects. But perhaps this is less disruptive.

core/src/main/java/org/apache/calcite/schema/lookup/MappedLookup.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.apache.calcite.schema.lookup;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Set;
+import java.util.function.BiFunction;
+
+/**
+ * A Lookup class which can be used to map different element types.
```
mihaibudiu, line 25: map -> transform
using a supplied transform function.
I would rename "mapper" to "transform"
I think "map" is way too overloaded

core/src/main/java/org/apache/calcite/schema/lookup/Lookup.java
```diff
@@ -0,0 +1,96 @@
...
+
+import org.apache.calcite.util.NameMap;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Set;
+import java.util.function.BiFunction;
+
+/**
+ * A casesensitive/insensitive lookup for tables, schems, functions ...
```
mihaibudiu, line 27: typo.
Probably you want to add types to the list too.

core/src/main/java/org/apache/calcite/schema/lookup/Lookup.java
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Returns a named entity with a given name ignoring the case, or null if not found.
+   *
+   * @param name Name
+   * @return Entity, or null
+   */
+  @Nullable Named<T> getIgnoreCase(String name);
+
+  /**
+   * Returns the names of the entities in matching pattern.
```
mihaibudiu, line 49: I assume the pattern can specify case-insensitive matching too (ILIKE).
please specify

core/src/main/java/org/apache/calcite/schema/Schema.java
```diff
@@ -56,9 +59,30 @@
...
+   * @return Lookup
+   */
+  default Lookup<? extends Schema> subSchemas() {
+    return new CompatibilityLookup<>(this::getSubSchema, this::getSubSchemaNames);
+  }
+
   /**
    * Returns a table with a given name, or null if not found.
    *
+   * <p>Please use {@link Schema#tables()} and {@link Lookup#get(String)} instead.
```
mihaibudiu, line 84: people will not find this comment.
maybe the API should be marked as deprecated.
on the other hand, this API makes sense for some cases - and it should perhaps have a default implementation doing exactly what you suggest.
There must be some downside to doing that - what is it?

core/src/main/java/org/apache/calcite/schema/lookup/CompatibilityLookup.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * This class can be used to wrap existing schemas with a pair of {@code get...}
+ * and {@code get...Names} into a Lookup object.
+ *
+ * @param <T> Element type
+ */
+public class CompatibilityLookup<T> extends IgnoreCaseLookup<T> {
```
mihaibudiu, line 34: Frankly, this class does not seem to have anything to do with schemas.
So I would first document what the class does, and then say that it can be used to wrap schemas.

core/src/main/java/org/apache/calcite/schema/lookup/CompatibilityLookup.java
```diff
@@ -0,0 +1,54 @@
...
+
+  private final Function<String, @Nullable T> get;
+  private final Supplier<Set<String>> getNames;
+
+  public CompatibilityLookup(Function<String, @Nullable T> get, Supplier<Set<String>> getNames) {
```
mihaibudiu, line 39: Please document the public constructor.

core/src/main/java/org/apache/calcite/schema/lookup/ConcatLookup.java
```diff
@@ -0,0 +1,62 @@
...
+package org.apache.calcite.schema.lookup;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * This class can be used to concat a list of lookups.
```
mihaibudiu, line 26: what does this mean?
please describe the semantics - especially for the case of common names.

core/src/main/java/org/apache/calcite/schema/lookup/EmptyLookup.java
```diff
@@ -0,0 +1,45 @@
...
+package org.apache.calcite.schema.lookup;
+
+import com.google.common.collect.ImmutableSet;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Set;
+
+/**
+ * This class implements an empty Lookup.
```
mihaibudiu, line 26: i.e., that contains no named object.
Lookup returns null for all names.

core/src/main/java/org/apache/calcite/schema/lookup/IgnoreCaseLookup.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.calcite.util.LazyReference;
+import org.apache.calcite.util.NameMap;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * An abstract base class for lookups. implementing case insensitive lookup
```
mihaibudiu, line 28: dot in the middle of the sentence
is this notion of "case sensitivity" dependent on the Locale?

core/src/main/java/org/apache/calcite/schema/lookup/IgnoreCaseLookup.java
```diff
@@ -0,0 +1,77 @@
...
+          .firstEntry();
+      if (entry != null) {
+        T result = get(entry.getValue());
+        return result == null ? null : new Named<>(entry.getKey(), result);
+      }
+      retryCounter++;
+      if (retryCounter > 1) {
+        return null;
+      }
+      nameMap.reset();
```
mihaibudiu, line 64: I don't get the point of this, please explain

core/src/main/java/org/apache/calcite/schema/lookup/IgnoreCaseLookup.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Set;
+
+/**
+ * An abstract base class for lookups. implementing case insensitive lookup
+ *
+ * @param <T> Element type
+ */
+public abstract class IgnoreCaseLookup<T> implements Lookup<T> {
+
+  private LazyReference<NameMap<String>> nameMap = new LazyReference<>();
```
mihaibudiu, line 34: I think this should be documented. If I am guessing right, it maps a name to it's "canonical" form. If that's true, call it canonicalNameMap.

core/src/main/java/org/apache/calcite/schema/lookup/LikePattern.java
```diff
@@ -0,0 +1,92 @@
...
+ * limitations under the License.
+ */
+package org.apache.calcite.schema.lookup;
+
+import org.apache.calcite.linq4j.function.Predicate1;
+
+import java.util.regex.Pattern;
+
+/**
+ * This class is used as parameter to Lookup.getNames
```
mihaibudiu, line 24: you should say what the class does, not what it is used for.

core/src/main/java/org/apache/calcite/schema/lookup/LikePattern.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+public class LikePattern {
+  private static final String ANY = "%";
+  public final String pattern;
+
+  public LikePattern(String pattern) {
```
mihaibudiu, line 30: please document the structure of the pattern.

core/src/main/java/org/apache/calcite/jdbc/CachingCalciteSchema.java
```diff
@@ -117,15 +100,14 @@ private CachingCalciteSchema(@Nullable CalciteSchema parent, Schema schema,
...
-    for (String schemaName2
-        : subSchemaCache.names.range(schemaName, caseSensitive)) {
-      return subSchemaCache.cache.getUnchecked(schemaName2);
-    }
-    return null;
+  @Override protected CalciteSchema createSubSchema(Schema schema, String name) {
+    return new CachingCalciteSchema(this, schema, name);
+  }
+
+  @Override protected <S> Lookup<S> decorateLookup(Lookup<S> lookup) {
```
mihaibudiu, line 107: I cannot guess what "decorate" means here

core/src/main/java/org/apache/calcite/jdbc/SimpleCalciteSchema.java
```diff
@@ -100,36 +100,8 @@ private SimpleCalciteSchema(@Nullable CalciteSchema parent,
...
-      return null;
-    }
-    final Schema s = schema.getSubSchema(schemaName2);
-    if (s == null) {
-      return null;
-    }
-    return new SimpleCalciteSchema(this, s, schemaName2);
-  }
-
-  @Override protected @Nullable TableEntry getImplicitTable(String tableName,
```
mihaibudiu, line 119: what happened to all these functions?
(I don't know what they were meant for.)

core/src/main/java/org/apache/calcite/schema/lookup/CachedLookup.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * This class can be used to make a snapshot of a lookups.
+ *
+ * @param <T> Element Type
+ */
+public class CachedLookup<T> implements Lookup<T> {
+
+  private final Lookup<T> delegate;
+  private LazyReference<Lookup<T>> cachedDelegate = new LazyReference<>();
```
mihaibudiu, line 34: I am assuming this is the "snapshot".
I would call it "snapshot" or "delegateSnapshot".
Is this class mixing two functionalities? Caching and snapshotting look to me to be separate behaviors.

core/src/main/java/org/apache/calcite/schema/lookup/LikePattern.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+    final Pattern regex = likeToRegex(likePattern);
+    return v1 -> regex.matcher(v1).matches();
+  }
+
+  /**
+   * Converts a LIKE-style pattern (where '%' represents a wild-card, escaped
+   * using '\') to a Java regex.
+   */
+  public static Pattern likeToRegex(String pattern) {
```
mihaibudiu, line 61: So this is always case-sensitive. Please document.

core/src/main/java/org/apache/calcite/schema/lookup/LoadingCacheLookup.java
```diff
@@ -0,0 +1,82 @@
...
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * This class can be used to cache lookups.
```
mihaibudiu, line 33: Please say what it does, before you can say what it can be used for.
The important aspect seems to be some kind of cache expiry mechanism.

I wonder about consistency. Can you say something about this?
I am guessing this assumes (and perhaps schemas do too) that objects are only inserted, and never deleted?

core/src/main/java/org/apache/calcite/jdbc/CalciteSchema.java
```diff
@@ -335,14 +334,14 @@ public NavigableMap<String, LatticeEntry> getLatticeMap() {
...
-    // Add explicit tables, case-sensitive.
-    builder.addAll(tableMap.map().keySet());
-    // Add implicit tables, case-sensitive.
-    addImplicitTableToBuilder(builder);
-    return builder.build();
+  public final Set<String> getTableNames() {
+    return getTableNames(LikePattern.any());
+  }
+
+  /** Returns the set of filtered table names. Includes implicit and explicit tables
```
mihaibudiu, line 341: filtered by the pattern

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCatalogSchema.java
```diff
@@ -103,34 +143,25 @@ public static JdbcCatalogSchema create(
...
-            requireNonNull(resultSet.getString(1),
-                "got null schemaName from the database");
-        builder.put(schemaName,
-            new JdbcSchema(dataSource, dialect, convention, catalog, schemaName));
-      }
+  @Override public Lookup<Table> tables() {
+    return Lookup.empty();
+  }
+
+  @Override public Lookup<? extends Schema> subSchemas() {
```
mihaibudiu, line 150: what happens if the subschemas change in the underlying connection?
is there an invalidation protocol?

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
```diff
@@ -237,80 +252,86 @@ public DataSource getDataSource() {
...
-        // This can happen if you start JdbcSchema off a "public" PG schema
-        // The tables are not designed to be queried by users, however we do
-        // not filter them as we keep all the other table types.
-        final String tableTypeName2 =
-            tableDef.tableType == null
+  private static @Nullable String intern(@Nullable String string) {
+    if (string == null) {
+      return null;
+    }
+    return string.intern();
```
mihaibudiu, line 312: I believe that the default Java string interning mechanism has a tiny set reserved, and can get very inefficient for a large number of strings. Is this controlled anywhere?

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
```diff
@@ -237,80 +252,86 @@ public DataSource getDataSource() {
...
+  }
+
+  private static TableType getTableType(String tableTypeName) {
+    // Clean up table type. In particular, this ensures that 'SYSTEM TABLE',
+    // returned by Phoenix among others, maps to TableType.SYSTEM_TABLE.
+    // We know enum constants are upper-case without spaces, so we can't
+    // make things worse.
```
mihaibudiu, line 319: worse than what?

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
```diff
@@ -237,80 +252,86 @@ public DataSource getDataSource() {
...
+    //
+    // PostgreSQL returns tableTypeName==null for pg_toast* tables
```
mihaibudiu, line 321: I am not thrilled by this code - which seems inherited.
Ideally you would be able to check whether this is a Postgres database, and only then do these "cleanups".

mihaibudiu (commented):
Has any benchmarking been done to prove the efficiency of this approach?
I am not an expert in this part of the code, but the PR looks pretty good to me.
I have only made "syntactic" comments.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcBaseSchema.java
```diff
@@ -0,0 +1,94 @@
...
+public abstract class JdbcBaseSchema implements Schema {
+
+  @Override public abstract Lookup<Table> tables();
+
+
+  @Override public @Nullable Table getTable(String name) {
+    return tables().get(name);
+  }
+
+  @Override public Set<String> getTableNames() {
```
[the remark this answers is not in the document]
kramerul: No this is not possible. The newly introduced functions `tables()` and `subSchemas()` must have a default implementation.
Moving the implementation of `getTableNames()` to `Schema` will result in a stack overflow because `getTableNames()`  -> `tables()`  -> `getTableNames()`

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -7489,7 +7488,7 @@ private void checkGetTimestamp(Connection con) throws SQLException {
     aSchema.setCacheEnabled(true);
 
     // explicit should win implicit.
-    assertThat(aSchema.getSubSchemaNames(), hasSize(1));
+    assertThat(aSchema.subSchemas().getNames(LikePattern.any()), hasSize(1));
```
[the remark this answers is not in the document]
kramerul: I first startet with a `@Deprecated` annotation on `getSubSchemaNames()`. Therefore, I needed to rename it in all tests.
Afterwards, I removed the `@Deprecated` annotation because it might cause trouble in some cases.

I could revert this change, if you prefer this.

core/src/test/java/org/apache/calcite/schema/lookup/MapLookup.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Override public @Nullable String get(final String name) {
+    return map.get(name);
+  }
+
+  @Override public @Nullable Named<String> getIgnoreCase(final String name) {
+    return ignoreCaseMap.get(name.toLowerCase(Locale.ROOT));
+  }
+
+  @Override public Set<String> getNames(final LikePattern pattern) {
```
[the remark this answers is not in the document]
kramerul: This was by mistake. I fixed it.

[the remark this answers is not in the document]
kramerul: It was not necessary since the class was only used for testing.

core/src/test/java/org/apache/calcite/schema/lookup/ConcatLookupTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @Test void testNull() {
+    assertThat(testee.get("c"), nullValue());
+  }
+
+  @Test void test() {
+    assertThat(testee.get("a"), equalTo("1"));
+  }
+
+  @Test void testIgnoreCase() {
```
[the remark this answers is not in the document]
kramerul: I added a test.

core/src/main/java/org/apache/calcite/util/LazyReference.java
kramerul: I added a description.

core/src/main/java/org/apache/calcite/schema/lookup/Lookup.java
kramerul: Fixed

core/src/main/java/org/apache/calcite/schema/lookup/Lookup.java
kramerul: No it's always casesensitive.

kramerul: The search is always casesensitive. This is caused by the fact that
`DatabaseMetaData.getTables(...)` doesn't support caseinsensitive lookups.

I added this also to the method documentation.

core/src/main/java/org/apache/calcite/schema/Schema.java
asolimando: If there is consensus that the method should be deprecated I see no problem in marking it as such, we do it on a regular basis already. Given the pervasive use of Calcite in many downstream projects, we haven't removed any deprecated method so far (there are several comments saying "to be removed before v2.0 in the codebase"), but the fact that downstream projects are warned ahead of time is exactly what we want.

core/src/main/java/org/apache/calcite/schema/lookup/CompatibilityLookup.java
kramerul: I changed the documentation accordingly

core/src/main/java/org/apache/calcite/schema/lookup/CompatibilityLookup.java
kramerul: OK

core/src/main/java/org/apache/calcite/schema/lookup/IgnoreCaseLookup.java
kramerul: "case sensitivity" does not depend on Locale. It uses `String.compareToIgnoreCase` which does not take a Locale as argument.

core/src/main/java/org/apache/calcite/schema/lookup/IgnoreCaseLookup.java
kramerul: If the name was not found in the cached list of names,
we try to reload the cache once because the table/schema
might have been created in the meantime.

core/src/main/java/org/apache/calcite/schema/lookup/IgnoreCaseLookup.java
kramerul: This member is used to lazily load the list of all names into memory

A `NameMap` is used, which is capable to lookup names in a caseinsensitive way.

core/src/main/java/org/apache/calcite/schema/lookup/LikePattern.java
kramerul: I added a description of the pattern to the class documentation.

core/src/main/java/org/apache/calcite/jdbc/CachingCalciteSchema.java
kramerul: I renamed it to `enhanceLookup`and changed the documentation in `CalciteSchema`:

The derived class is able to enhance the lookup e.g. by introducing a cache

core/src/main/java/org/apache/calcite/jdbc/SimpleCalciteSchema.java
kramerul: `SimpleCalciteSchema` and `CachingCalciteSchema` formerly had two list for tables and sub schemas each:
* implicit entries which are loaded from the associated Schema (member `schema`)
* explicit entries which are managed by calls to `add...`

The concatenation of both lists was formerly happening in `CalciteSchema.getTableNames`.  This is now happening in `CalciteSchema.tables()` with the help of `Lookup.concat()`.
There is no longer a special need to have different implementations of `getImplicitTables` for `SimpleCalciteSchema` and `CachingCalciteSchema` since the caching is done in `CachingCalciteSchema.enhanceLookup` for both tables and sub schemas.

core/src/main/java/org/apache/calcite/schema/lookup/CachedLookup.java
kramerul: I renamed it to `SnapshotLookup`

core/src/main/java/org/apache/calcite/schema/lookup/LikePattern.java
kramerul: OK

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcCatalogSchema.java
kramerul: The invalidation is happening after one minute as I described in my comment to class `LoadingCacheLookup`

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
kramerul: I removed the `intern` method. This is no longer required. We needed this to ensure that we require as less memory as possible. With the new lazy loading approach, this is nor longer needed.

core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java
kramerul: I didn't change this comment. I just moved it around. Therefore, I'm not able to tell you.

kramerul:
> Has any benchmarking been done to prove the efficiency of this approach? I am not an expert in this part of the code, but the PR looks pretty good to me. I have only made "syntactic" comments.

This PR only improves performance for huge database. We are using a database with more than 500000 schemas containing up to 500000 tables.

In such an environment, it takes more than 10 seconds to load all table names from the database.  Formerly, this was necessary during the preparation of each query. With the new approach, only the involved tables are loaded from the database. This speeds up the preparation by factors. It also takes much less memory, because it's no longer necessary to hold a list of all tables in memory (snapshot).

core/src/main/java/org/apache/calcite/jdbc/CalciteSchema.java
kramerul: I changed the documentation

core/src/main/java/org/apache/calcite/schema/lookup/LoadingCacheLookup.java
kramerul: I changed the documentation accordingly:

This class is using a `LoadingCache`to speed up lookups,
 delegated to another `Lookup` instance.
 
 This class is thread safe. All entries are evicted after one minute.
 Negative matches are never cached. If a new entry
 becomes available in the associated {@code Lookup}, it's immediately
 visible outside. If an entry is deleted in the associated {@code Lookup},
 it takes one minute until it disappears if it was cached in the last minute.
 Otherwise, it disappears immediately.

core/src/main/java/org/apache/calcite/schema/lookup/EmptyLookup.java
kramerul: I changed the documentation

core/src/main/java/org/apache/calcite/schema/Schema.java
kramerul: For me, it would be OK to add the `@Deprecated` annotation. 

Since it's my first PR that introduces an incompatible change, I wanted to wait for others' opinions first.

mihaibudiu:
Please ask for a new review when you are done

core/src/main/java/org/apache/calcite/schema/Schema.java
mihaibudiu: If these methods are not used anywhere anymore, then it may be good to mark them as deprecated, and document this in history.md.

core/src/main/java/org/apache/calcite/schema/lookup/CompatibilityLookup.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * This class can be used to implement the methods {@code Schema.tables()}
```
mihaibudiu, line 29: I still think it's a good idea to describe what a class represents or does, rather than what it can be used for. What it can be used for can be in the JavaDoc, but in the second paragraph.  The "what is" could be as simple as "helper class for ...".

core/src/main/java/org/apache/calcite/schema/lookup/CompatibilityLookup.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param <T> Element type
+ */
+public class CompatibilityLookup<T> extends IgnoreCaseLookup<T> {
+
+  private final Function<String, @Nullable T> get;
+  private final Supplier<Set<String>> getNames;
+
+  /**
+   * Constructor.
```
mihaibudiu, line 46: "Constructor" is not particularly useful - this is already clear.
But I agree that "Creates a CompabilityLookup" does not say much more...

core/src/main/java/org/apache/calcite/schema/lookup/ConcatLookup.java
```diff
@@ -0,0 +1,65 @@
...
+package org.apache.calcite.schema.lookup;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * This class can be used to concat a list of lookups into a new lookup object.
```
mihaibudiu, line 26: This creates a Lookup object from an ordered sequence of Lookup objects.
In the resulting lookup, the name lookup is done from left to right...
Same thing about "can be used".

core/src/main/java/org/apache/calcite/schema/lookup/IgnoreCaseLookup.java
```diff
@@ -0,0 +1,86 @@
...
+   *
+   * <p>A {@link NameMap} is used, which is capable to lookup names in a
+   * caseinsensitive way.
+   */
+  private LazyReference<NameMap<String>> nameMap = new LazyReference<>();
+
+  /**
+   * Returns a named entity with a given name, or null if not found.
+   *
+   * @param name Name
```
mihaibudiu, line 45: I don't insist having documentation for parameters which are "obvious" - here the documentation does not add much. Some tools do. For `@return` it's use to say "Entity with the specified name, or null when such entity is not found".

core/src/main/java/org/apache/calcite/schema/lookup/LoadingCacheLookup.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * This class is using a {@code LoadingCache} to speed up lookups,
+ * delegated to another {@link Lookup} instance.
+ *
+ * <p>This class is thread safe. All entries are evicted after one minute.
```
mihaibudiu, line 37: I don't see the "1 minute" in the code.
I would document the constructor parameters too, especially the expiration.

core/src/main/java/org/apache/calcite/schema/lookup/Lookup.java
```diff
@@ -0,0 +1,99 @@
...
+
+import org.apache.calcite.util.NameMap;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Set;
+import java.util.function.BiFunction;
+
+/**
+ * A casesensitive/insensitive lookup for tables, schemas, functions, types ...
```
mihaibudiu, line 27: I think casesensitive is two words. Maybe with a hyphen? This shows up in multiple places, not only here.

core/src/main/java/org/apache/calcite/schema/lookup/Lookup.java
```diff
@@ -0,0 +1,99 @@
...
+   * @param name Name
+   * @return Entity, or null
+   */
+  @Nullable T get(String name);
+
+  /**
+   * Returns a named entity with a given name ignoring the case, or null if not found.
+   *
+   * @param name Name
+   * @return Entity, or null
```
mihaibudiu, line 44: particularly interface members can have longer documentation - e.g., as I wrote in an example above.

core/src/main/java/org/apache/calcite/schema/lookup/SnapshotLookup.java
```diff
@@ -0,0 +1,78 @@
...
+
+import org.apache.calcite.util.LazyReference;
+import org.apache.calcite.util.NameMap;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Set;
+
+/**
+ * This class can be used to make a snapshot of a lookups.
```
mihaibudiu, line 27: Creates a snapshot of lookups at one point in time.

core/src/main/java/org/apache/calcite/schema/lookup/TransformingLookup.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.apache.calcite.schema.lookup;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Set;
+import java.util.function.BiFunction;
+
+/**
+ * A Lookup class which can be used to map different element types.
```
mihaibudiu, line 25: you really like "can be used".

core/src/main/java/org/apache/calcite/util/LazyReference.java
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.util;
+
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
+/**
+ * This class can be used to lazily initialize an object.
```
mihaibudiu, line 23: Lazy initialization of objects.

core/src/main/java/org/apache/calcite/schema/Schema.java
```diff
@@ -67,6 +99,8 @@ public interface Schema {
   /**
    * Returns the names of the tables in this schema.
    *
+   * <p>Please use {@link Schema#tables()} and {@link Lookup#getNames(LikePattern)} instead.
```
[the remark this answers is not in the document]
kramerul: OK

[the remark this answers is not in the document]
kramerul: OK

core/src/main/java/org/apache/calcite/schema/lookup/CompatibilityLookup.java
kramerul: I agree with you, but it's hard for me to find other words. I will try my best.

core/src/main/java/org/apache/calcite/schema/lookup/CompatibilityLookup.java
kramerul: OK

core/src/main/java/org/apache/calcite/schema/lookup/ConcatLookup.java
kramerul: OK

core/src/main/java/org/apache/calcite/schema/lookup/LoadingCacheLookup.java
kramerul: OK

core/src/main/java/org/apache/calcite/schema/lookup/IgnoreCaseLookup.java
kramerul: OK

core/src/main/java/org/apache/calcite/schema/lookup/Lookup.java
kramerul: I corrected all places.

core/src/main/java/org/apache/calcite/schema/lookup/Lookup.java
kramerul: OK

core/src/main/java/org/apache/calcite/util/LazyReference.java
kramerul: OK

site/_docs/history.md
```diff
@@ -78,6 +78,12 @@ large results set to a manageable value. Users that need a bigger/smaller limit
 should create a new instance of `RelMdUniqueKeys` and register it using the
 metadata provider of their choice.
 
+* [<a href="https://issues.apache.org/jira/browse/CALCITE-6728">CALCITE-6728</a>]
+introduces new methods to lookup tables and sub schemas inside schemas.
+The methods used before (`Schema:getTable(String name)`, `Schema:getTableNames()`,
+`Schema.getSubSchema(String name)` and `Schema.getSubSchemaNames(String name)`)
+have been markes as deprecated.
```
mihaibudiu, line 85: typo

mihaibudiu (approved):

mihaibudiu:
I think this is ready for merging; please fix the typo when you squash the commits.

kramerul:
I fixed the type and squashed all commits.

core/src/main/java/org/apache/calcite/jdbc/CalciteSchema.java
```diff
@@ -322,7 +322,6 @@ public final NavigableMap<String, CalciteSchema> getSubSchemaMap() {
     final ImmutableSortedMap.Builder<String, CalciteSchema> builder =
         new ImmutableSortedMap.Builder<>(NameSet.COMPARATOR);
     builder.putAll(subSchemaMap.map());
-    addImplicitSubSchemaToBuilder(builder);
```
rubenada, line 325: @kramerul Sorry, I'm late to the party here. Quick question: isn't this a breaking change? Previously `getSubSchemaMap` would return both the explicit and implicit sub-schema (as its javadoc states), but now the implicit ones would no longer be considered? (I guess those could still be accessed via the new `subSchemas()` method), but it'd seem that `getSubSchemaMap` has changed its behavior (and its javadoc is no longer accurate). Or I'm missing something?

kramerul: Hi @rubenada,

you are completely right. This is a breaking change. 

`getSubSchemaMap` needs to be changed to read all sub schemas using `subSchemas`.

kramerul: I think it should look like this

```java
  public final NavigableMap<String, CalciteSchema> getSubSchemaMap() {
    // Build a map of implicit sub-schemas first, then explicit sub-schemas.
    // If there are implicit and explicit with the same name, explicit wins.
    final ImmutableSortedMap.Builder<String, CalciteSchema> builder =
        new ImmutableSortedMap.Builder<>(NameSet.COMPARATOR);
    Lookup<CalciteSchema> schemas = subSchemas();
    for ( String name : schemas.getNames(LikePattern.any())) {
      builder.put(name, schemas.get(name));
    }
    return builder.build();
  }
```

rubenada: Thanks for the confirmation @kramerul , I'll create a PR accordingly

rubenada: No problem @kramerul ! Thanks you for suggesting the code to fix it :)
If you have a bit of time, could you please take a look at https://github.com/apache/calcite/pull/4254 to confirm it's ok?

PR #4102, opened 2024-12-20
[CALCITE-6741] The type of a comparison is nullable when either opera…

mihaibudiu:
…nd is nullable

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
@@ -4480,7 +4480,7 @@ public class SqlValidatorTest extends SqlValidatorTestCase {
     expr("(1,2) in ((1,2), (3,4))")
         .columnType("BOOLEAN NOT NULL");
     expr("'medium' in (cast(null as varchar(10)), 'bc')")
-        .columnType("BOOLEAN NOT NULL");
+        .columnType("BOOLEAN");
 
     // nullability depends on nullability of both sides
     sql("select empno in (1, 2) from emp")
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -4480,7 +4480,7 @@ private void checkNegWindow(String s, String msg) {
     expr("(1,2) in ((1,2), (3,4))")
         .columnType("BOOLEAN NOT NULL");
     expr("'medium' in (cast(null as varchar(10)), 'bc')")
-        .columnType("BOOLEAN NOT NULL");
```
NobiGo, line 4483: In my opinion, this change is reasonable, but it's strange that it only affects this test case. Do we need to add more test cases to cover the corresponding changes?

mihaibudiu: I suspect is affects many programs, but there a few reasons we don't see this:
- this affects only some subexpressions, but often the final expression in a Project has the same type
- many program representations do not display the types of subexpressions
- the Enumerable representation is in Java, where all Objects are nullable
- there aren't enough tests with complex expressions and null values
Our runtime is written in Rust, and there are separate types for non-nullable (T) and nullable (Option<T>). Many mistakes are caught at compile-time. Other mistakes that are not caught at compile-time become panics at runtime due to unwrap() calls on Option objects. In Java both of these may be silent.
It's not easy to write good tests to catch these problems. I think adding more tests like the one that has failed has value, but only limited value.

NobiGo: I get it. If there are no test cases, future PRs are likely to reverse the current changes.

mihaibudiu: @NobiGo see a second commit with more coercion tests

NobiGo (approved):
LGTM

PR #4107, opened 2024-12-21
[CALCITE-6742] StandardConvertletTable.convertCall loses casts from ROW comparisons

mihaibudiu:


core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -563,9 +563,15 @@ private Expression getConvertExpression(
     case TINYINT:
     case SMALLINT: {
       if (SqlTypeName.NUMERIC_TYPES.contains(sourceType.getSqlTypeName())) {
+        Type javaClass = typeFactory.getJavaClass(targetType);
```
mihaibudiu, line 566: This change is only indirectly related to this PR: without this change, the code generated for ROW casts crashes at runtime.

caicancai: Why does it crash? It feels a bit strange

mihaibudiu:
This PR fixes a real bug and should not be too hard to review.

caicancai (approved):

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -7654,6 +7654,28 @@ from orders]]>
...
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testStructCast1">
+    <Resource name="sql">
+      <![CDATA[select CAST(CAST(ROW('x', 1) AS ROW(l INTEGER, r DOUBLE)) AS ROW(l BIGINT, r INTEGER)) = ROW(RAND(), RAND())]]>
+    </Resource>
+    <Resource name="plan">
+      <![CDATA[
+LogicalProject(EXPR$0=[AND(=(CAST(CAST('x'):INTEGER NOT NULL):DOUBLE NOT NULL, RAND()), =(1.0E0, RAND()))])
```
ILuffZhe, line 7677: `1.0E0` comes from `cast 1 as DOUBLE`?

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -7654,6 +7654,28 @@ from orders]]>
 LogicalDelta
   LogicalProject(ROWTIME=[$0], PRODUCTID=[$1], ORDERID=[$2], C=[COUNT() OVER (PARTITION BY $1 ORDER BY $0 RANGE 1000:INTERVAL SECOND PRECEDING)])
     LogicalTableScan(table=[[CATALOG, SALES, ORDERS]])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testStructCast">
+    <Resource name="sql">
+      <![CDATA[select ROW(1, 'x') = ROW('x', 1)]]>
```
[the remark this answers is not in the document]
mihaibudiu: This becomes `CAST('x' AS INTEGER) = 1 AND 1 = CAST('x' AS INTEGER)`, which is then simplified

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
mihaibudiu: yes, there are many places where Calcite simplifies expressions, it's hard to write a test exhibits just the change I made.

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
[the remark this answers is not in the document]
mihaibudiu: I will change the test the way you suggested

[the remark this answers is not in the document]
ILuffZhe: Got it. Thanks for the illustration.

PR #4108, opened 2024-12-22
[CALCITE-6730] Add CONVERT function(enabled in Oracle library)

ILuffZhe:


ILuffZhe:
Oops, seems Windows' default charset `windows-1252`(equivalent to `latin1`)  is not supported in Calcite currently.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -740,6 +741,7 @@ void populate1() {
       defineMethod(CONCAT_WS_SPARK,
           BuiltInMethod.MULTI_TYPE_STRING_ARRAY_CONCAT_WITH_SEPARATOR.method,
           NullPolicy.ARG0);
+      defineMethod(ORACLE_CONVERT, BuiltInMethod.ORACLE_CONVERT.method, NullPolicy.ARG0);
```
strongduanmu, line 744: What do you think about changing `ORACLE_CONVERT` to `CONVERT_ORACLE`? It looks like the current naming format is functionName + dbType, for example: CONCAT_WS_SPARK, BIT_COUNT_MYSQL.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -1681,6 +1681,29 @@ public static String translateWithCharset(String s, String transcodingName) {
     }
   }
 
+  /** Oracle's {@code CONVERT(charValue, destCharsetName[, srcCharsetName])} function,
+   * return null if s is null or empty. */
+  public static String oracleConvert(String s, String... args) {
```
strongduanmu, line 1686: Similar to the previous suggestion, if `oracleConvert` could be changed to `convertOracle`, perhaps this would be consistent with the existing `regexpReplacePg`.

core/src/main/java/org/apache/calcite/sql/fun/SqlOracleConvertFunction.java
```diff
@@ -0,0 +1,123 @@
...
+
+import static org.apache.calcite.sql.type.NonNullableAccessors.getCollation;
+
+import static java.util.Objects.requireNonNull;
+
+/** Oracle's "CONVERT(charValue, destCharsetName[, srcCharsetName])" function.
+ *
+ * <p>It has a slight different semantics to standard SQL's
+ * {@link SqlStdOperatorTable#CONVERT} function on operands' order, and default
+ * charset will be used if the {@code srcCharsetName} ixs not specified.
```
strongduanmu, line 47: Is ixs a typo?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
ILuffZhe: Sure, I'll change it later, as well as `oracleConvert` in SqlFunctions.

core/src/main/java/org/apache/calcite/sql/fun/SqlOracleConvertFunction.java
ILuffZhe: Good catch!

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
```diff
@@ -853,15 +854,35 @@ protected RexNode convertFloorCeil(SqlRexContext cx, SqlCall call) {
...
-    final String destCharset = call.operand(2).toString();
-    final RexBuilder rexBuilder = cx.getRexBuilder();
-    return rexBuilder.makeCall(pos, SqlStdOperatorTable.CONVERT,
-        cx.convertExpression(expr),
-        rexBuilder.makeLiteral(srcCharset),
-        rexBuilder.makeLiteral(destCharset));
+    final SqlOperator op = call.getOperator();
+    final String srcCharset;
+    final String destCharset;
+    if (op == SqlStdOperatorTable.CONVERT) {
```
NobiGo, line 863: We could use SWITCH  and throw unsupported exceptions when we have another function.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4646,6 +4646,49 @@ static void checkRlikeFails(SqlOperatorFixture f) {
...
+
+    final Consumer<SqlOperatorFixture> consumer = f0 -> {
+      f0.checkFails("convert('a', utf8, utf10)", "UTF10", false);
+      f0.checkFails("^convert(1, utf8, gbk)^",
+          "Invalid type 'INTEGER NOT NULL' in 'CONVERT' function\\. "
+              + "Only 'CHARACTER' type is supported",
+          false);
+      f0.checkType("convert('a', utf16, gbk)", "CHAR(1) NOT NULL");
+      f0.checkType("convert('a', utf16)", "CHAR(1) NOT NULL");
+      f0.checkType("convert(null, utf16, gbk)", "NULL");
```
NobiGo, line 4665: Maybe we can add test `convert('a', null, gbk)` or `convert('a', gbk, null)`.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -4646,6 +4646,49 @@ static void checkRlikeFails(SqlOperatorFixture f) {
...
+              + " from (values(true)) where ^cast(convert('col', latin1) as char(3))=_GBK'col'^",
+          "Cannot apply operation '=' to strings with "
+              + "different charsets 'ISO-8859-1' and 'GBK'",
+          false);
+      // the result of convert('col', gbk) has GBK charset
+      // while CHAR(3) has ISO-8859-1 charset, which is not allowed to cast
+      f.checkFails("select 'a' as alia\n"
+              + " from (values(true)) where ^cast(convert('col', gbk) as char(3))^=_GBK'col'",
+          "Cast function cannot convert value of type "
+              + "CHAR\\(3\\) CHARACTER SET \"GBK\" NOT NULL to type CHAR\\(3\\) NOT NULL",
```
NobiGo, line 4683: We can add the target encoding format to the abnormal information like `ISO-8859-1 charset`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
ILuffZhe: Oracle's CONVERT doesn't support this, the 2nd or 3rd operand should always be Character set.

core/src/main/codegen/templates/Parser.jj
```diff
@@ -6319,10 +6319,17 @@ SqlNode BuiltinFunctionCall() :
...
-                }
+                (
+                    <COMMA> e = SimpleIdentifier() { args.add(e); }
+                    <RPAREN> {
+                        SqlOperator op = SqlStdOperatorTable.getConvertFuncByConformance(this.conformance);
+                        return op.createCall(s.end(this), args);
+                    }
+                |
+                    <RPAREN> {
+                        return SqlLibraryOperators.ORACLE_CONVERT.createCall(s.end(this), args);
```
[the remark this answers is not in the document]
ILuffZhe: It's necessary because we handle form like `CONVERT('a', UTF8)` here.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
ILuffZhe: You mean adding a test that throws messages contain `ISO-8859-1`?
I'm not sure we can do this since the basic char or varchar type has the `ISO-8859-1` encoding by default. Or you can see if the above test(line 4673) is satisfied.

core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java
ILuffZhe: How about adding a SqlKind `ORACLE_CONVERT`, and we use `op.getKind()` here? Now they all have `CONVERT` SqlKind.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
ILuffZhe: I also add `convert('a', utf8, ^null^)` and `convert('a', ^null^, utf8)`.

ILuffZhe:
I've updated the PR, please take a look when you guys are free. @strongduanmu @NobiGo

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -5741,6 +5741,19 @@ private static Matcher<SqlNode> isCharLiteral(String s) {
             + "FROM (VALUES (ROW(TRUE))))");
   }
 
+  @Test void testOracleConvert() {
```
strongduanmu, line 5744: Would it be better to change it to testConvertOracle? Other similar method names and comments need to be consistent.

ILuffZhe: Done.

strongduanmu (approved):
LGTM +1

NobiGo (approved):

PR #4109, opened 2024-12-23
[CALCITE-6744] Support getColumnOrigins for correlate in RelMdColumnOrigins

suibianwanwank:
This PR introduces support for the getOriginColumn method for Correlate. 
To represent the origin of the corVar field, the ColumnOrigin class has been modified accordingly. This caused breaking changes, so I'm documenting it in history.md. If there's a better way to do it, welcome any suggestions.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/metadata/RelColumnOrigin.java
+++ b/core/src/main/java/org/apache/calcite/rel/metadata/RelColumnOrigin.java
@@ -17,8 +17,11 @@
 package org.apache.calcite.rel.metadata;
 
 import org.apache.calcite.plan.RelOptTable;
+import org.apache.calcite.rel.core.CorrelationId;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static java.util.Objects.requireNonNull;
 
 /**
  * RelColumnOrigin is a data structure describing one of the origins of an
@@ -27,9 +30,13 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 public class RelColumnOrigin {
   //~ Instance fields --------------------------------------------------------
 
-  private final RelOptTable originTable;
+  private final @Nullable RelOptTable originTable;
+
+  private final @Nullable CorrelationId correlationId;
 
   private final int iOriginColumn;
+
+  private final boolean isCorVar;
 
   private final boolean isDerived;
 
@@ -39,16 +46,38 @@ public class RelColumnOrigin {
       RelOptTable originTable,
       int iOriginColumn,
       boolean isDerived) {
+    this(originTable, null, iOriginColumn, isDerived, false);
+  }
+
+  public RelColumnOrigin(
+      CorrelationId correlationId,
+      int iOriginColumn,
+      boolean isDerived) {
+    this(null, correlationId, iOriginColumn, isDerived, true);
+  }
+
+  private RelColumnOrigin(@Nullable RelOptTable originTable,
+      @Nullable CorrelationId correlationId,
+      int iOriginColumn,
+      boolean isDerived,
+      boolean isCorVar) {
     this.originTable = originTable;
+    this.correlationId = correlationId;
     this.iOriginColumn = iOriginColumn;
     this.isDerived = isDerived;
+    this.isCorVar = isCorVar;
   }
 
   //~ Methods ----------------------------------------------------------------
 
-  /** Returns table of origin. */
-  public RelOptTable getOriginTable() {
+  /** Returns table of origin and null only if isCorVar is true. */
+  public @Nullable RelOptTable getOriginTable() {
     return originTable;
+  }
+
+  /** Returns correlateId of origin and null only if isCorVar is true. */
+  public @Nullable CorrelationId getCorrelationId() {
+    return correlationId;
   }
 
   /** Returns the 0-based index of column in origin table; whether this ordinal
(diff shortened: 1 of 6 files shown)


suibianwanwank:
There are some CI errors, I'll fix it tomorrow.

core/src/main/java/org/apache/calcite/rel/metadata/RelColumnOrigin.java
```diff
@@ -17,20 +17,27 @@
...
+
 /**
  * RelColumnOrigin is a data structure describing one of the origins of an
  * output column produced by a relational expression.
  */
 public class RelColumnOrigin {
   //~ Instance fields --------------------------------------------------------
 
-  private final RelOptTable originTable;
+  private final @Nullable RelOptTable originTable;
```
mihaibudiu, line 33: Ideally this would be implemented using an interface and two subclasses, one for each kind of origin.
Not sure how easy this can be accomplished while keeping compatibility.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnOrigins.java
```diff
@@ -303,6 +328,17 @@ private static Set<RelColumnOrigin> getMultipleColumns(RexNode rexNode, RelNode
             }
             return null;
           }
+
+          @Override public Void visitFieldAccess(RexFieldAccess fieldAccess) {
```
mihaibudiu, line 332: shouldn't this actually use as origin the table that correlated variable comes from (recursively)?

suibianwanwank: I'm not sure I fully understand what you mean. There is a case where the RelNode that calls getColumnOrigin is not the top-level RelNode, so it may not get the original table.

core/src/main/java/org/apache/calcite/rel/metadata/RelColumnOrigin.java
suibianwanwank: I think this would also be difficult to keep compatible. Another compromise would be to keep the RelColumnOrigin unchanged and return null if the input field's CorField comes from outside the input RelNode.
This way we just need to add a Map to maintain the mapping of CorrelationId, Index to ColumnOrigin, and go top-down to getColumnOrigin.
` public @Nullable Set<RelColumnOrigin> getColumnOrigins(Aggregate rel,
      RelMetadataQuery mq, int iOutputColumn, Map<Pair<CorrelationId, Integer>, ColumnOrigin>) corVarOriginMap) {
...
}`

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -458,6 +460,56 @@ void testColumnOriginsUnion() {
...
+        mq.getColumnOrigins(relNode.getInput(0).getInput(1), 0);
+
+    assertThat(origins, hasSize(2));
+    for (RelColumnOrigin origin : origins) {
+      if (origin.isCorVar()) {
+        CorrelationId correlationId = origin.getCorrelationId();
+        assertThat(correlationId, notNullValue());
+        assertThat(correlationId.getName(), equalTo("$cor1"));
+        assertThat(origin.getOriginColumnOrdinal(), equalTo(1));
+        continue;
```
mihaibudiu, line 505: in this case I think that an `else` is more appropriate.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -458,6 +460,56 @@ void testColumnOriginsUnion() {
...
+        .assertColumnOriginDouble("EMP", "ENAME",
+            "DEPT", "NAME", true);
+
+    // check correlate input column origins
+    final RelMetadataFixture.MetadataConfig metadataConfig =
+        fixture.metadataConfig;
+    final RelMetadataQuery mq =
+        new RelMetadataQuery(metadataConfig.getDefaultHandlerProvider());
+    Set<RelColumnOrigin> origins =
+        mq.getColumnOrigins(relNode.getInput(0).getInput(1), 0);
```
mihaibudiu, line 496: is this LogicalAggregate? Maybe you can say this in a comment.

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -458,6 +460,56 @@ void testColumnOriginsUnion() {
...
+      if (origin.isCorVar()) {
+        CorrelationId correlationId = origin.getCorrelationId();
+        assertThat(correlationId, notNullValue());
+        assertThat(correlationId.getName(), equalTo("$cor1"));
+        assertThat(origin.getOriginColumnOrdinal(), equalTo(1));
+        continue;
+      }
+      assertThat(origin.getOriginTable(), notNullValue());
+      assertThat(origin.getOriginTable().getQualifiedName().get(2), equalTo("DEPT"));
+      assertThat(origin.getOriginColumnOrdinal(), equalTo(1));
```
mihaibudiu, line 509: is this `dept.name`?

core/src/test/java/org/apache/calcite/test/RelMetadataTest.java
```diff
@@ -458,6 +460,56 @@ void testColumnOriginsUnion() {
...
+    Set<RelColumnOrigin> origins =
+        mq.getColumnOrigins(relNode.getInput(0).getInput(1), 0);
+
+    assertThat(origins, hasSize(2));
+    for (RelColumnOrigin origin : origins) {
+      if (origin.isCorVar()) {
+        CorrelationId correlationId = origin.getCorrelationId();
+        assertThat(correlationId, notNullValue());
+        assertThat(correlationId.getName(), equalTo("$cor1"));
+        assertThat(origin.getOriginColumnOrdinal(), equalTo(1));
```
mihaibudiu, line 504: want to check that originTable is null too?
Is this `ENAME`?

suibianwanwank:
@mihaibudiu, Thanks for the review! In my opinion, such a change breaks too much. A good way would be to return null on correlation variable. For correlated origin, maybe add other metadataQuery. So I've changed the issue and will be adding a new PR.

PR #4110, opened 2024-12-23
[CALCITE-6738] Suport a UUID type natively

mihaibudiu:


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -11262,11 +11264,12 @@ private static void checkSubstringFunction(SqlOperatorFixture f) {
           String.format(Locale.ROOT, "^substring('string', CAST(%d AS DOUBLE), "
               + "CAST(%d AS DOUBLE))^", Byte.MIN_VALUE, Byte.MAX_VALUE + 10),
           "Cannot apply 'SUBSTRING' to arguments of type "
-            + ".*\\n.*\\n.*\\n.*\\n.*\\n.*\\n.*\\n.*", false);
+            + ".*\\n.*\\n.*\\n.*\\n.*\\n.*\\n.*\\n.*\\n.*\\n.*", false);
```
ILuffZhe, line 11267: What changed here?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -277,6 +279,20 @@ public static boolean throwUnless(boolean condition, String message) {
     return condition;
   }
 
+  public static String uuidToString(UUID uuid) {
```
ILuffZhe, line 282: Shall we add some tests in `SqlFunctionsTest`?

core/src/test/resources/sql/misc.iq
```diff
@@ -18,7 +18,75 @@
...
++--------+
+| EXPR$0 |
++--------+
+| true   |
++--------+
+(1 row)
+
+!ok
+
+# Compare varchar to number
```
ILuffZhe, line 89: The former doc points out there is implicit cast for it.

core/src/test/resources/sql/misc.iq
```diff
@@ -18,7 +18,75 @@
 !use post
 !set outputformat mysql
 
-# Compared by casting varchar to number
+SELECT UUID '123e4567-e89b-12d3-a456-426655440000';
```
ILuffZhe, line 21: What are the results of `cast(null as UUID)` and `UUID null`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
mihaibudiu: Because of implicit casts SUBSTRING can now also allow UUID as arguments, which are implicitly cast to VARCHAR.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
mihaibudiu: These functions are so small that I didn't think there is much to test about them. They essentially delegate all the work to the library functions.

core/src/test/resources/sql/misc.iq
mihaibudiu: Yes, I wrote that in a recent PR as well. Clearly, this must be an implicit cast. I can restore that part, but the phrasing wasn't very good.

core/src/test/resources/sql/misc.iq
mihaibudiu: I will add tests for these. But they should be NULL.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1374,6 +1377,10 @@ public RexLiteral makeLiteral(boolean b) {
     return b ? booleanTrue : booleanFalse;
   }
 
+  public RexLiteral makeUuidLiteral(@Nullable UUID uuid) {
+    return new RexLiteral(uuid, typeFactory.createSqlType(SqlTypeName.UUID), SqlTypeName.UUID);
```
NobiGo, line 1381: Can the UUID be null? According to SqlUuidLiteral‘s implementation, it can't.
```
  public UUID getUuid() {
    return (UUID) requireNonNull(value, "value");
  }
```

mihaibudiu: SqlUuidLiteral cannot be null, but the RexLiteral can be null, due to simplifications of expressions such as `CAST(NULL AS UUID)`.

core/src/test/resources/sql/misc.iq
mihaibudiu: the latter is in fact illegal, and won't parse; `UUID` must be followed by a string literal. This is similar to `DATE`, `TIMESTAMP`, and other literals.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
mihaibudiu: BTW: this pattern holds true for many other literals, such as `SqlTimestampLiteral`

ILuffZhe (approved):
Overall looks good to me. Let's see if there are any more comments in the next few days.

NobiGo (approved):

PR #4112, opened 2024-12-24
[CALCITE-6754] Remove deprecated method calling in Driver

ILuffZhe:


ILuffZhe:
> Now calcite does not support the MINOR keyword. You can try to modify the commit message

Thanks for your reminder.

caicancai (approved):

caicancai:
Just a comment, I think this PR is worth creating a jira

ILuffZhe:
> Just a comment, I think this PR is worth creating a jira

Sure, I'll do it later this day.

ILuffZhe:
@caicancai Hi, I've changed the commit info and filed a JIRA ticket, please take a look when you have time.

core/src/main/java/org/apache/calcite/jdbc/Driver.java
```diff
@@ -101,13 +101,12 @@ public CalcitePrepare createPrepare() {
...
     }
-    return createPrepareFactory().apply();
+    return CalcitePrepare.DEFAULT_FACTORY.apply();
   }
 
   /** Returns a factory with which to create a {@link CalcitePrepare}.
    *
-   * <p>Now deprecated; if you wish to use a custom prepare, overrides of this
-   * method will still work, but we prefer that you call
+   * <p>Now deprecated; if you wish to use a custom prepare, please call
```
NobiGo, line 109: Can't we override this method now?

ILuffZhe: Yes, we can override `createPrepare()` to achieve the same goal.
Since `createPrepareFactory()` is deprecated, maybe we shouldn't call or override it any more.

caicancai (approved):
LGTM

ILuffZhe:
I think removing this deprecated method calling is just a matter of time. Let's see if there comes another reviewer.

PR #4114, opened 2024-12-27
[CALCITE-6747] Multiple SLF4J bindings in Spark unit test

AstraHoo:


spark/build.gradle.kts
```diff
@@ -31,5 +31,5 @@ dependencies {
     runtimeOnly("xerces:xercesImpl")
 
     testImplementation(project(":testkit"))
-    testRuntimeOnly("org.apache.logging.log4j:log4j-slf4j-impl")
+    testRuntimeOnly("org.slf4j:slf4j-reload4j")
```
ILuffZhe, line 34: Can you add the environment reports in JIRA after this change？ I think that will be more intuitive for others.

AstraHoo: Before:
SparkAdapterTest > testFilterTrue() STANDARD_ERROR
    SLF4J: Class path contains multiple SLF4J bindings.
    SLF4J: Found binding in [jar:file:/Users/hubinhua/.gradle/caches/modules-2/files-2.1/org.apache.logging.log4j/log4j-slf4j-impl/2.17.1/84692d456bcce689355d33d68167875e486954dd/log4j-slf4j-impl-2.17.1.jar!/org/slf4j/impl/StaticLoggerBinder.class]
    SLF4J: Found binding in [jar:file:/Users/hubinhua/.gradle/caches/modules-2/files-2.1/org.slf4j/slf4j-reload4j/1.7.36/db708f7d959dee1857ac524636e85ecf2e1781c1/slf4j-reload4j-1.7.36.jar!/org/slf4j/impl/StaticLoggerBinder.class]
    SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
    SLF4J: Actual binding is of type [org.apache.logging.slf4j.Log4jLoggerFactory] 


After:
SparkAdapterTest > testOrderByLimit() STANDARD_OUT
    2024-12-27 08:31:25,429 [ForkJoinPool-1-worker-4] INFO  - Your platform does not provide complete low-level API for accessing direct buffers reliably. Unless explicitly requested, heap buffer will always be preferred to avoid potential system unstability.
          2.7sec, org.apache.calcite.test.SparkAdapterTest > testFilterFalse()
          2.7sec, org.apache.calcite.test.SparkAdapterTest > testFilterTrue()
          2.7sec, org.apache.calcite.test.SparkAdapterTest > testOrderByLimit()
          2.8sec, org.apache.calcite.test.SparkAdapterTest > testSortXDescYDescProjectY()

AstraHoo: @ILuffZhe Hi,I have added the CI reports  before and after the modification

AstraHoo:
The test cases for SparkAdapter Test are executed concurrently in a thread pool, so these functions ,testOrderByLimit(),testFilterFalse(),testFilterTrue(),testOrderByLimit(),testSortXDescYDescProjectY(),may not necessarily come first, but the exception will disappear after modification

ILuffZhe:
There are also some `multiple SLF4J bingings` in pig adapter, can you take a look?

zabetak:
@ILuffZhe It might not be possible to drop the log4j dependency from Pig. I vaguely remember that there were strong dependencies to log4j classes from within Pig.

zabetak (approved):
LGTM. I tested and the binding is removed from spark module.It may be possible to do the same for the pig and piglet modules but it's perfectly fine to do that in separate PRs.

ILuffZhe (approved):
Separate PR also makes sense.

PR #4115, opened 2024-12-27
[CALCITE-6749] RelMdUtil#setAggChildKeys may return an incorrect result

rubenada:
see https://issues.apache.org/jira/browse/CALCITE-6749


rubenada:
As a consequence of the fix, there are a couple of plan changes on *.iq tests, but the result is still correct.

core/src/test/java/org/apache/calcite/rel/metadata/RelMdUtilTest.java
```diff
@@ -110,4 +117,39 @@ final RelMetadataFixture sql(String sql) {
     });
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6749">[CALCITE-6749]
+   * RelMdUtil#setAggChildKeys may return an incorrect result</a>. */
+  @Test void testSetAggChildKeys() {
```
mihaibudiu, line 123: Frankly, this is hard to review.
I saw the discussion in the JIRA, but it's not obvious at all that "9" is the right answer.
I wonder whether having the full plan here would actually be easier - at least in a comment.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdUtil.java
```diff
@@ -568,15 +568,15 @@ public static void setAggChildKeys(
...
       Aggregate aggRel,
       ImmutableBitSet.Builder childKey) {
-    List<AggregateCall> aggCalls = aggRel.getAggCallList();
+    final List<AggregateCall> aggCallList = aggRel.getAggCallList();
+    final List<Integer> groupList = aggRel.getGroupSet().asList();
     for (int bit : groupKey) {
       if (bit < aggRel.getGroupCount()) {
         // group by column
-        childKey.set(bit);
+        childKey.set(groupList.get(bit));
```
arkanovicz, line 576: Why is the conversion to list needed? Why not:
```
for (int i: rel.getGroupSet())
```

core/src/test/java/org/apache/calcite/rel/metadata/RelMdUtilTest.java
rubenada: Basically we need to convert: "field index X on the Aggregate corresponds to which field index Y on the Aggregate's input?"
If we have this plan
```
LogicalAggregate(group=[{0}], EXPR$1=[COUNT(DISTINCT $1)])
  LogicalProject(DEPTNO=[$9], JOB=[$2])
    LogicalJoin(condition=[=($7, $9)], joinType=[right])
```
The Aggregate rowType has two fields (0 and 1), the first one (index 0) corresponds to the group 0, and the second one (index 1) corresponds to the argCall COUNT(DISTINCT $1). In this case, if we want to convert into the input's, we'll have 0 => 0, and 1 => 1; there is no problem on the current code.

However, if we have this equivalent plan (same query plan, but without the intermediate Project):
```
LogicalAggregate(group=[{9}], EXPR$1=[COUNT(DISTINCT $2)])
  LogicalJoin(condition=[=($7, $9)], joinType=[right])
```
Now the Aggregate still has two fields (0 and 1), the first one (index 0) corresponds to the group 9, and the second one (index 1) corresponds to the argCall COUNT(DISTINCT $2). In this case, if we want to convert into the input's, we'll have 0 => 9, and 1 => 2; the current code will work fine only for the argCall conversion, but not for the group conversion.
Notice I'm just applying here the same conversion that exists already e.g. when propagating the [RelMdColumnOrigins computation past an Aggregate](https://github.com/apache/calcite/blob/901aadbe8bd55e2da791d9d827185a8c803b1b76/core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnOrigins.java#L74) (in terms of our second example: "if we want to get origin of the column 0 of the Aggregate, we need to compute the origin of the column 9 of the Aggregate's input")

core/src/main/java/org/apache/calcite/rel/metadata/RelMdUtil.java
```diff
@@ -568,15 +568,15 @@ public static void setAggChildKeys(
...
       } else {
```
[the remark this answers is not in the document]
rubenada: `aggRel.getGroupSet().get(i)` would return a boolean, what we need here is an index conversion. I'm basically following the same logic as the one already applied on [RelMdColumnOrigins for aggregates](https://github.com/apache/calcite/blob/901aadbe8bd55e2da791d9d827185a8c803b1b76/core/src/main/java/org/apache/calcite/rel/metadata/RelMdColumnOrigins.java#L74).

mihaibudiu (approved):

PR #4116, opened 2024-12-28
[CALCITE-6751] Reduction of CAST from string to interval is incorrect

mihaibudiu:
While fixing this bug I uncovered two more bugs, https://issues.apache.org/jira/browse/CALCITE-6752 and https://issues.apache.org/jira/browse/CALCITE-6753. Some of the code changes are workarounds for these other bugs.

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
@@ -304,8 +304,8 @@ public class RexToLixTranslator implements RexVisitor<RexToLixTranslator.Result>
       ConstantExpression format) {
     Expression convert = getConvertExpression(sourceType, targetType, operand, format);
     Expression convert2 = checkExpressionPadTruncate(convert, sourceType, targetType);
-    Expression convert3 = expressionHandlingSafe(convert2, safe, targetType);
-    return scaleValue(sourceType, targetType, convert3);
+    Expression convert3 = scaleValue(sourceType, targetType, convert2);
+    return expressionHandlingSafe(convert3, safe, targetType);
   }
 
   private Expression getConvertExpression(
@@ -1213,7 +1213,7 @@ public class RexToLixTranslator implements RexVisitor<RexToLixTranslator.Result>
     final SqlTypeFamily sourceFamily = sourceType.getSqlTypeName().getFamily();
     if (targetFamily == SqlTypeFamily.NUMERIC
         // multiplyDivide cannot handle DECIMALs, but for DECIMAL
-        // destination types the result is already scaled.
+        // target types the result is already scaled.
         && targetType.getSqlTypeName() != SqlTypeName.DECIMAL
         && (sourceFamily == SqlTypeFamily.INTERVAL_YEAR_MONTH
             || sourceFamily == SqlTypeFamily.INTERVAL_DAY_TIME)) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/adapter/enumerable/RexToLixTranslator.java
```diff
@@ -304,8 +304,8 @@ Expression translateCast(
       ConstantExpression format) {
     Expression convert = getConvertExpression(sourceType, targetType, operand, format);
     Expression convert2 = checkExpressionPadTruncate(convert, sourceType, targetType);
-    Expression convert3 = expressionHandlingSafe(convert2, safe, targetType);
-    return scaleValue(sourceType, targetType, convert3);
+    Expression convert3 = scaleValue(sourceType, targetType, convert2);
```
mihaibudiu, line 307: this is the right order to do the conversion: first scaling, then checking for safety.
Otherwise the safe cast implementation returns null, which cannot be an input for the scaling computation.

testkit/src/main/java/org/apache/calcite/test/CalciteAssert.java
```diff
@@ -569,6 +569,13 @@ static void assertQuery(
       try {
         if (updateChecker == null) {
           resultSet = statement.executeQuery(sql);
+          if (resultChecker == null && exceptionChecker != null) {
```
mihaibudiu, line 572: This is a separate bug, which was uncovered by the `@NonDeterministic` annotation on toInt. The exception was happening at compilation time, instead of runtime. This change is necessary to uncover exceptions that happen at runtime.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4271,6 +4271,10 @@ public static int toInt(java.sql.Time v) {
     return v == null ? castNonNull(null) : toInt(v);
   }
 
+  // Method tagged as non-deterministic because it can throw.
+   // The DeterministicCodeOptimizer may otherwise try to lift it out of try-catch blocks.
```
ILuffZhe, line 4275: Nit: the indent.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -4271,6 +4271,10 @@ public static int toInt(java.sql.Time v) {
...
+  // See https://issues.apache.org/jira/browse/CALCITE-6753
```
ILuffZhe, line 4276: Is this change should be added in CALCITE-6753?

mihaibudiu: I don't have a fix for 6753, only a workaround

core/src/test/resources/sql/misc.iq
```diff
@@ -18,6 +18,40 @@
...
+| EXPR$0   |
++----------+
+| 10:00:01 |
++----------+
+(1 row)
+
+!ok
+
+# Unfortunately, due to CALCITE-6752 the following test crashes
+# SELECT TIME '10:00:00' + CAST('1.1' AS INTERVAL SECOND);
```
NobiGo, line 33: Can the annotations here add the correct results？Even if it can't run here.

NobiGo (approved):
LGTM

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
NobiGo, line 4276: CALCITE-6753 can be resolved in an independent PR.

core/src/test/resources/sql/misc.iq
mihaibudiu: I will add this by amending the existing commit.

PR #4117, opened 2024-12-28
[CALCITE-6748] RelToSqlConverter returns the wrong result when Aggregate is on Sort

suibianwanwank:


core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -1965,6 +1965,13 @@ private boolean needNewSubQuery(
         return true;
       }
 
+      if (rel instanceof Aggregate
```
mihaibudiu, line 1968: seems you need an extra check for null here

suibianwanwank: Done, the wrong reference object was used here.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -8616,6 +8616,20 @@ private void checkLiteral2(String expression, String expected) {
         .withSpark().ok(sparkExpected);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6748">[CALCITE-6748]
+   * RelToSqlConverter returns the wrong result when Aggregate is on Sort</a>. */
+  @Test void testAggregateOnSort() {
+    final String query = "select max(\"product_class_id\") "
+        + "from (select * from \"product\" order by \"brand_name\" asc limit 10) t";
```
NobiGo, line 8624: The unit test didn't cover `offset`.

NobiGo (approved):

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
suibianwanwank: Done, I add a test.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -8620,14 +8620,22 @@ private void checkLiteral2(String expression, String expected) {
...
         + "FETCH NEXT 10 ROWS ONLY) AS \"t1\"";
-    sql(query).ok(expected);
+    sql(query0).ok(expected0);
+
+    final String query1 = "select max(\"product_class_id\") "
+        + "from (select * from \"product\" offset 10 ) t";
+    final String expected1 = "SELECT MAX(\"product_class_id\")\n"
+        + "FROM (SELECT \"product_class_id\"\n"
+        + "FROM \"foodmart\".\"product\"\n"
+        + "OFFSET 10 ROWS) AS \"t1\"";
```
ILuffZhe, line 8637: order by expression `ORDER BY \"brand_name\"` is missing.Shall we add it?

suibianwanwank: Adding an order by here would cause the problem described in [CALCITE-6579]. So I didn't add it and it doesn't conflict with this PR.

ILuffZhe: I'm not sure if it's the expected plan as you described in CALCITE-6759. The result of `COUNT` seems ok if there is no `order by`, do you try `MAX(\"product_class_id\")` when `order by product_class_id asc`? If the later loses `order by`, I think there might be some problem.

suibianwanwank: There is a problem with both `COUNT` and `MAX`. When there is an emp table in PG with 200 rows. The query `select * from emp offset 10` will returns 190 rows.

ILuffZhe: OK, I get it. I thought only `order by` is missing at first.

PR #4119, opened 2025-01-02
[CALCITE-6759] SqlToRelConverter will remove the sort with offset

suibianwanwank:


core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -777,6 +777,7 @@ protected void convertSelectImpl(
     // we would never remove the ORDER BY, because "ORDER BY ... LIMIT" is about
     // semantics. It is not a 'pure order'.
     if (RelOptUtil.isPureOrder(castNonNull(bb.root))
+        && !RelOptUtil.isOffset(castNonNull(bb.root))
```
caicancai, line 780: It seems that the JavaDoc needs to be improved

mihaibudiu: should offset be incorporated in isPureOrder?

mihaibudiu (commented):
This looks fine, but if the offset check can be incorporated in isPureOrder it may be better.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
suibianwanwank: Done.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -189,10 +189,10 @@ public static boolean isPureLimit(RelNode rel) {
   }
 
   /**
-   * Whether this node is a sort without limit specification.
+   * Whether this node is a sort without limit or offset specification.
```
ILuffZhe, line 192: Should the doc be `and`? not `or`.

ILuffZhe (approved):
Overall looks good.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -774,7 +774,7 @@ protected void convertSelectImpl(
     //
     // Semantics example. Given the view definition
     //   CREATE VIEW v2 AS SELECT * FROM t ORDER BY x LIMIT 10
```
NobiGo, line 776: Do we need to add another SQL about the change？

NobiGo (commented):
Leave some suggestions on Jira.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
caicancai: +1

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
suibianwanwank: +1, apologies for the late reply, I've fixed it in the latest commit.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
suibianwanwank: Done.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
julianhyde: I'd use 'or', as in 'without salt or pepper'. But I agree it's a little ambiguous. I'll change to 'with neither a limit nor an offset specification' when I merge.

mihaibudiu:
commits should have been squashed

julianhyde:
> commits should have been squashed

They were squashed, just not in the PR. I added a commit in my [own branch](https://github.com/julianhyde/calcite/tree/6759-order-offset), then squashed and merged to main when the tests pass.

It's a useful flow because it preserves the history of the dev branch, and allows the committer (me in this case) to copy-edit and add more tests.

PR #4120, opened 2025-01-02
[CALCITE-6760] Support time data type in Arrow adapter

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6760

### DIFF
--- a/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowDataTest.java
+++ b/arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowDataTest.java
@@ -31,12 +31,14 @@ import org.apache.arrow.vector.Float8Vector;
 import org.apache.arrow.vector.FloatingPointVector;
 import org.apache.arrow.vector.IntVector;
 import org.apache.arrow.vector.SmallIntVector;
+import org.apache.arrow.vector.TimeSecVector;
 import org.apache.arrow.vector.TinyIntVector;
 import org.apache.arrow.vector.VarCharVector;
 import org.apache.arrow.vector.VectorSchemaRoot;
 import org.apache.arrow.vector.ipc.ArrowFileWriter;
 import org.apache.arrow.vector.types.DateUnit;
 import org.apache.arrow.vector.types.FloatingPointPrecision;
+import org.apache.arrow.vector.types.TimeUnit;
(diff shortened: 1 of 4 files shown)


caicancai:
I encountered some problems adapting the timestmap type, so this pr first adapts to the time type

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowDataTest.java
```diff
@@ -411,4 +420,14 @@ private void dateField(FieldVector fieldVector, int rowCount) {
     }
     fieldVector.setValueCount(rowCount);
   }
+
+  private void timeField(FieldVector fieldVector, int rowCount) {
+    TimeSecVector timeVector = (TimeSecVector) fieldVector;
+    timeVector.setInitialCapacity(rowCount);
+    timeVector.allocateNew();
+    for (int i = 0; i < rowCount; i++) {
+      timeVector.set(i, i * 1000);
```
ILuffZhe, line 429: Why multiply `i` by 1000?

ILuffZhe (approved):
Overall looks good, just left one minor question.

arrow/src/test/java/org/apache/calcite/adapter/arrow/ArrowDataTest.java
caicancai: It depends on arrow-java's adaptation of the time type. Like most Java libraries, the basic unit is milliseconds, so 1000ms == 1s

ILuffZhe: You mean there exist unit that is not milliseconds? How do we handle this.

caicancai: I am adapting and discussing with the Arrow community. Currently, the Arrow community only has the SQL standard for timestamp, but no SQL standard for time type.

caicancai: @ILuffZhe After testing, no matter which timevector is chosen, it is in milliseconds.

ILuffZhe: Thanks for the clarify.

PR #4122, opened 2025-01-04
[CALCITE-6761] StarRocks generates incorrect SQL for certain units in the EXTRACT function

xiaochen-zhou:
StarRocks does not support the following `EXTRACT` unit syntax, and `StarRocksSqlDialect` may require some adjustments to accommodate this.

image

For example:

```
SELECT EXTRACT(DOW FROM DATE '2023-12-01');

SELECT EXTRACT(DOY FROM DATE '2023-12-01');
```

To ensure compatibility during the StarRocks dialect conversion, the above syntax can be translated as follows:

```
SELECT DAYOFWEEK(DATE '2023-12-01');

SELECT DAYOFYEAR(DATE '2023-12-01');
```

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
@@ -257,6 +257,27 @@ public class MysqlSqlDialect extends SqlDialect {
       unparseListAggCall(writer, call, null, leftPrec, rightPrec);
       break;
 
+    case EXTRACT:
+      SqlLiteral node = call.operand(0);
+      TimeUnitRange unit = node.getValueAs(TimeUnitRange.class);
+      String funName;
+      switch (unit) {
+      case DOW:
+        funName = "DAYOFWEEK";
+        break;
+      case DOY:
+        funName = "DAYOFYEAR";
+        break;
+      default:
+        super.unparseCall(writer, call, leftPrec, rightPrec);
+        return;
+      }
+      writer.print(funName);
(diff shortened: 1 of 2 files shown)


ILuffZhe (approved):
LGTM!

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -102,6 +102,26 @@ public StarRocksSqlDialect(Context context) {
               timeUnitNode.getParserPosition());
       SqlFloorFunction.unparseDatetimeFunction(writer, newCall, "DATE_TRUNC", false);
       break;
+    case EXTRACT:
```
ILuffZhe, line 105: Maybe we should do it in MySqlDialect since MySQL extract function also doesn't support DOW and DOY.

xiaochen-zhou: > Maybe we should do it in MySqlDialect since MySQL extract function also doesn't support DOW and DOY.

Done.

caicancai (approved):

ILuffZhe:
@xiaochen-zhou  Can you squash the commits?

caicancai:
@xiaochen-zhou your commit message should be` [CALCITE-6761] StarRocks generates incorrect SQL for certain units in the EXTRACT function `

xiaochen-zhou:
> @xiaochen-zhou your commit message should be`[CALCITE-6761] StarRocks generates incorrect SQL for certain units in the EXTRACT function`

Done.

PR #4123, opened 2025-01-04
[CALCITE-6699] Invalid unparse for Varchar in StarRocksDialect

xiaochen-zhou:
When performing starrocks dialect conversion for varchar, it defaults to using MySQL dialect，MySQL doesn't have a VARCHAR type, only CHAR,So during StarRocks dialect conversion, VARCHAR will also be converted to CHAR

Since StarRocks supports the VARCHAR type, we should modify the StarRocks dialect

image

https://docs.starrocks.io/docs/sql-reference/data-types/string-type/VARCHAR/

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
@@ -19,6 +19,8 @@ package org.apache.calcite.sql.dialect;
 import org.apache.calcite.avatica.util.TimeUnitRange;
 import org.apache.calcite.config.NullCollation;
 import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeSystem;
+import org.apache.calcite.rel.type.RelDataTypeSystemImpl;
 import org.apache.calcite.sql.SqlAbstractDateTimeLiteral;
 import org.apache.calcite.sql.SqlAlienSystemTypeNameSpec;
 import org.apache.calcite.sql.SqlBasicTypeNameSpec;
@@ -41,9 +43,33 @@ import static org.apache.calcite.util.RelToSqlConverterUtil.unparseHiveTrim;
  */
 public class StarRocksSqlDialect extends MysqlSqlDialect {
 
+  /** StarRocks type system. */
+  public static final RelDataTypeSystem STARROCKS_TYPE_SYSTEM =
+      new RelDataTypeSystemImpl() {
+        @Override public int getMaxPrecision(SqlTypeName typeName) {
+          switch (typeName) {
+          case CHAR:
+            return 255;
+          case VARCHAR:
+            return 65533;
+          case VARBINARY:
+            return 1048576;
+          default:
+            return super.getMaxPrecision(typeName);
+          }
+        }
+        @Override public int getDefaultPrecision(SqlTypeName typeName) {
+          if (typeName == SqlTypeName.CHAR) {
+            return RelDataType.PRECISION_NOT_SPECIFIED;
+          }
+          return super.getDefaultPrecision(typeName);
+        }
+      };
+
   public static final SqlDialect.Context DEFAULT_CONTEXT = SqlDialect.EMPTY_CONTEXT
       .withDatabaseProduct(SqlDialect.DatabaseProduct.STARROCKS)
       .withIdentifierQuoteString("`")
+      .withDataTypeSystem(STARROCKS_TYPE_SYSTEM)
       .withNullCollation(NullCollation.LOW);
 
   public static final SqlDialect DEFAULT = new StarRocksSqlDialect(DEFAULT_CONTEXT);
@@ -127,6 +153,10 @@ public class StarRocksSqlDialect extends MysqlSqlDialect {
               "DATETIME",
               type.getSqlTypeName(),
               SqlParserPos.ZERO),
+          SqlParserPos.ZERO);
+    case VARCHAR:
+      return new SqlDataTypeSpec(
+          new SqlBasicTypeNameSpec(SqlTypeName.VARCHAR, type.getPrecision(), SqlParserPos.ZERO),
           SqlParserPos.ZERO);
     default:
       return super.getCastSpec(type);
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -3748,6 +3748,39 @@ class RelToSqlConverterTest {
   }
 
   /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6699">[CALCITE-6699]
+   * Invalid unparse for Varchar in StarRocksDialect </a>. */
+  @Test void testStarRocksCastToVarcharWithLessThanMaxPrecision() {
+    final String query = "select cast(\"product_id\" as varchar(50)), \"product_id\" "
+        + "from \"product\" ";
+    final String expected = "SELECT CAST(`product_id` AS VARCHAR(50)), `product_id`\n"
+        + "FROM `foodmart`.`product`";
+    sql(query).withStarRocks().ok(expected);
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6699">[CALCITE-6699]
+   * Invalid unparse for Varchar in StarRocksDialect </a>. */
+  @Test void testStarRocksCastToVarcharWithGreaterThanMaxPrecision() {
+    final String query = "select cast(\"product_id\" as varchar(150000)), \"product_id\" "
+        + "from \"product\" ";
+    final String expected = "SELECT CAST(`product_id` AS VARCHAR(65533)), `product_id`\n"
+        + "FROM `foodmart`.`product`";
+    sql(query).withStarRocks().ok(expected);
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6699">[CALCITE-6699]
+   * Invalid unparse for Varchar in StarRocksDialect </a>. */
+  @Test void testStarRocksCastToVarcharWithDefaultPrecision() {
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -41,9 +43,32 @@
...
 
+  public static final RelDataTypeSystem STARROCKS_TYPE_SYSTEM =
+      new RelDataTypeSystemImpl() {
+        @Override public int getMaxPrecision(SqlTypeName typeName) {
+          switch (typeName) {
+          case CHAR:
+            return 255;
+          case VARCHAR:
+            return 65533;
+          case VARBINARY:
```
ILuffZhe, line 54: Is `VARBINARY` different from what in MySQL? I noticed this is the only difference, maybe MySQL and StarRocks can share this type system to get max precision.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3747,6 +3747,30 @@ private SqlDialect nonOrdinalDialect() {
     sql(query).dialect(mySqlDialect(NullCollation.LAST)).ok(expected);
   }
 
+  @Test void testStarRocksCastToVarcharWithLessThanMaxPrecision() {
```
ILuffZhe, line 3750: Do we need tests to cover `CHAR` and `VARBINARY` types?

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
xiaochen-zhou: > Is `VARBINARY` different from what in MySQL? I noticed this is the only difference, maybe MySQL and StarRocks can share this type system to get max precision.

I thought it would be the same as MySQL, but what’s strange is that the StarRocks official website lists the maximum as 65,533  
image

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xiaochen-zhou: > Do we need tests to cover `CHAR` and `VARBINARY` types?

I think CHAR might not be necessary since it is the same as in MySQL. However, VARBINARY might need tests. I will add them.

xiaochen-zhou: > Do we need tests to cover `CHAR` and `VARBINARY` types?

Done.

ILuffZhe:
Hi @xiaochen-zhou , you don't need to force-push for every commit, and it helps other reviewers to see the changes better.

xiaochen-zhou:
> https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6699



> Hi @xiaochen-zhou , you don't need to force-push for every commit, and it helps other reviewers to see the changes better.

OK. Thank you for your reminder.

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -128,6 +153,15 @@ public StarRocksSqlDialect(Context context) {
               type.getSqlTypeName(),
               SqlParserPos.ZERO),
           SqlParserPos.ZERO);
+    case VARCHAR:
+      int vcMaxPrecision = this.getTypeSystem().getMaxPrecision(SqlTypeName.VARCHAR);
+      int precision = type.getPrecision();
+      if (vcMaxPrecision > 0 && precision > vcMaxPrecision) {
```
NobiGo, line 159: The precision can be negative? Maybe we don't need `vcMaxPrecision > 0`?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3747,6 +3747,46 @@ private SqlDialect nonOrdinalDialect() {
...
+
+  @Test void testStarRocksCastToVarcharWithGreaterThanMaxPrecision() {
+    final String query = "select cast(\"product_id\" as varchar(150000)), \"product_id\" "
+        + "from \"product\" ";
+    final String expected = "SELECT CAST(`product_id` AS VARCHAR(65533)), `product_id`\n"
+        + "FROM `foodmart`.`product`";
+    sql(query).withStarRocks().ok(expected);
+  }
+
+  @Test void testStarRocksCastToVarcharWithOutPrecision() {
```
NobiGo, line 3766: testStarRocksCastToVarcharWithDefaultPrecision?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3747,6 +3747,46 @@ private SqlDialect nonOrdinalDialect() {
...
+    final String query = "select cast(\"product_id\" as varchar), \"product_id\" "
+        + "from \"product\" ";
+    final String expected = "SELECT CAST(`product_id` AS VARCHAR), `product_id`\n"
+        + "FROM `foodmart`.`product`";
+    sql(query).withStarRocks().ok(expected);
+  }
+
+  @Test void testStarRocksCastToVarbinaryWithLessThanMaxPrecision() {
+    final String query = "select CAST(x'ABCDEF12' AS VARBINARY(2)), \"product_id\" "
```
NobiGo, line 3775: Why do we need constant literal to test this?

xiaochen-zhou: > Why do we need constant literal to test this?

I initially wanted to test the default precision of the VARBINARY cast, but it seems unnecessary, so I removed it.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xiaochen-zhou: > testStarRocksCastToVarcharWithDefaultPrecision?

Done.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -3747,6 +3747,30 @@ private SqlDialect nonOrdinalDialect() {
...
+        + "from \"product\" ";
+    final String expected = "SELECT CAST(`product_id` AS VARCHAR(50)), `product_id`\n"
+        + "FROM `foodmart`.`product`";
+    sql(query).withStarRocks().ok(expected);
+  }
+
+  @Test void testStarRocksCastToVarcharWithGreaterThanMaxPrecision() {
+    final String query = "select cast(\"product_id\" as varchar(150000)), \"product_id\" "
+        + "from \"product\" ";
+    final String expected = "SELECT CAST(`product_id` AS VARCHAR(65533)), `product_id`\n"
```
mihaibudiu, line 3761: is this supposed to change the precision silently, or should this at least produce a warning?

xiaochen-zhou:
I have updated the code. Please take a look when you have time. @mihaibudiu @NobiGo @ILuffZhe

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
NobiGo: @mihaibudiu  I noticed that our default implementation silently changes the conversion in `SQLTypeUtil.convertTypeToSpec`, so I think there is no problem here.

NobiGo (approved):

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -128,6 +153,10 @@ public StarRocksSqlDialect(Context context) {
               type.getSqlTypeName(),
               SqlParserPos.ZERO),
           SqlParserPos.ZERO);
+    case VARCHAR:
```
caicancai, line 156: I remember we discussed this issue before. Does StarRocks have any restrictions on varchar types, such as varchar(65533)? If so, please explain it, at least add JavaDoc.

xiaochen-zhou: > I remember we discussed this issue before. Does StarRocks have any restrictions on varchar types, such as varchar(65533)? If so, please explain it, at least add JavaDoc.

There are restrictions. 

image

Calcite will truncate according to the maximum precision. I have added test cases
image

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -41,9 +43,32 @@
  */
 public class StarRocksSqlDialect extends MysqlSqlDialect {
 
+  public static final RelDataTypeSystem STARROCKS_TYPE_SYSTEM =
+      new RelDataTypeSystemImpl() {
+        @Override public int getMaxPrecision(SqlTypeName typeName) {
+          switch (typeName) {
```
caicancai, line 49: Add JavaDoc

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
caicancai, line 3750: Add jira link

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
caicancai: thanks

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xiaochen-zhou: > Add jira link

Done.

xiaochen-zhou:
I have updated the code. Please take a look when you have time. @mihaibudiu @NobiGo @ILuffZhe @caicancai

PR #4124, opened 2025-01-04
[CALCITE-6762] Preserving the CAST conversion for String type operands in Presto

xiaochen-zhou:
In Calcite, if the dialect's `supportsImplicitTypeCoercion` is set to `true` (the default value is `true`), Calcite will remove the cast in the `SqlImplementor#stripCastFromString()` method. For example, the following SQL:

```
FROM applydata_bigdata.kafka_topic_cluster_user_dept_month_one_day_msg_info
WHERE ymd > CAST('20240801' AS INTEGER)
ORDER BY ymd DESC
```

will be transformed into:

```
FROM "hive"."applydata_bigdata"."kafka_topic_cluster_user_dept_month_one_day_msg_info"
WHERE "ymd" > '20240801'
ORDER BY "ymd" IS NULL DESC, "ymd" DESC LIMIT 200
```

Here, the `CAST` is removed.

However, in Presto, different data types cannot be directly compared, which will result in an error. 

Similar situations also exist:

```
select * from hive.applydata_bigdata.kafka_topic_cluster_user_dept_month_one_day_msg_info 
where CURRENT_DATE > '2024-08-01'
order by ymd desc LIMIT 200
```

image

To avoid this issue, `supportsImplicitTypeCoercion` should be set to `false` in `PrestoSqlDialect`.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
@@ -20,6 +20,9 @@ import org.apache.calcite.avatica.util.Casing;
 import org.apache.calcite.config.NullCollation;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlBasicCall;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlDialect;
@@ -33,6 +36,8 @@ import org.apache.calcite.sql.fun.SqlLibraryOperators;
 import org.apache.calcite.sql.fun.SqlMapValueConstructor;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.type.SqlTypeUtil;
 import org.apache.calcite.util.RelToSqlConverterUtil;
 
 import com.google.common.collect.ImmutableList;
@@ -82,6 +87,17 @@ public class PrestoSqlDialect extends SqlDialect {
   @Override public void unparseOffsetFetch(SqlWriter writer, @Nullable SqlNode offset,
       @Nullable SqlNode fetch) {
     unparseUsingLimit(writer, offset, fetch);
+  }
+
+  @Override public boolean supportsImplicitTypeCoercion(RexCall call) {
+    RexNode rexNode = call.getOperands().get(0);
+    return super.supportsImplicitTypeCoercion(call)
+        && RexUtil.isLiteral(rexNode, false)
+        && (rexNode.getType().getSqlTypeName() == SqlTypeName.VARCHAR
+        || rexNode.getType().getSqlTypeName() == SqlTypeName.CHAR)
+        && !SqlTypeUtil.isNumeric(call.type)
+        && !SqlTypeUtil.isDate(call.type)
+        && !SqlTypeUtil.isTimestamp(call.type);
   }
 
   /** Unparses offset/fetch using "OFFSET offset LIMIT fetch " syntax. */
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7918,6 +7918,24 @@ private void checkLiteral2(String expression, String expected) {
     sql(query).withClickHouse().ok(expectedSql);
   }
 
+  @Test void testImplicitTypeCoercion() {
```
caicancai, line 7921: Add jira link

xiaochen-zhou: > Add jira link

Done.

NobiGo (commented):
Some information needs to be clearly stated in Jira. Does the CAST operation need to stay for all the types?

xiaochen-zhou:
> Some information needs to be clearly stated in Jira. Does the CAST operation need to stay for all the types?

OK.

core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
```diff
@@ -84,6 +85,10 @@ public PrestoSqlDialect(Context context) {
     unparseUsingLimit(writer, offset, fetch);
   }
 
+  @Override public boolean supportsImplicitTypeCoercion(RexCall call) {
```
mihaibudiu, line 88: do you have a reference to a document that describes this behavior of Presto?
If yes, perhaps you could include it here as a comment?

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
xiaochen-zhou: > do you have a reference to a document that describes this behavior of Presto? If yes, perhaps you could include it here as a comment?


The official documentation only contains the following related description: 
https://trino.io/docs/current/functions/conversion.html
image

However, my tests show that there is no conversion between string and date types either. 

image

Therefore, I believe that to ensure query correctness, the CAST conversion for String type operands in Presto should be preserved.

xiaochen-zhou:
> Some information needs to be clearly stated in Jira. Does the CAST operation need to stay for all the types?

The official documentation only contains the following related description: 
https://trino.io/docs/current/functions/conversion.html
image

However, my tests show that there is no conversion between string and date types either. 

image

Therefore, I believe that to ensure query correctness, the CAST conversion for String type operands in Presto should be preserved.

mihaibudiu:
there is a problem, because TypeCoercion handles all sorts of implicit casts, and here you are turning all of them off.
There is currently no way to control which implicit casts are legal in a dialect.

xiaochen-zhou:
> there is a problem, because TypeCoercion handles all sorts of implicit casts, and here you are turning all of them off.
> There is currently no way to control which implicit casts are legal in a dialect.

Sorry, I’m a bit confused. I found that this method currently seems to only be used in the SqlImplementor#stripCastFromString() method to determine whether to remove the cast.

image

mihaibudiu:
Look at the way the function is used in the BigQuery dialect: it needs to look at the RexCall and decide based on that whether the cast is supported. You cannot return 'false' always.

xiaochen-zhou:
> Look at the way the function is used in the BigQuery dialect: it needs to look at the RexCall and decide based on that whether the cast is supported. You cannot return 'false' always.

I have updated the code. Please take a look when you have time.

core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
```diff
@@ -86,7 +88,11 @@ public PrestoSqlDialect(Context context) {
   }
 
   @Override public boolean supportsImplicitTypeCoercion(RexCall call) {
-    return false;
+    return super.supportsImplicitTypeCoercion(call)
+        && RexUtil.isLiteral(call.getOperands().get(0), false)
```
mihaibudiu, line 92: this seems to do more than the JIRA issue states: it will not convert a literal except to numeric, date, or timestamp types. But you are not considering the literal type.

PR #4125, opened 2025-01-05
[CALCITE-6763] Optimize logic to select the tiles with the fewest rows

xiaochen-zhou:
Our current logic using a PriorityQueue to select the tile with the fewest rows when multiple tiles are available. However, there are several potential issues:

1. The PriorityQueue stores all satisfiable tiles. When the number of tiles is large, maintaining the heap structure during element insertion has a time complexity of O(log n), which also increases memory usage.
2. The initial size of the PriorityQueue is difficult to estimate and is currently set to 1, causing frequent resizing of the PriorityQueue.

We can optimize the code by keeping only a single bestCandidate tile to improve performance.

### DIFF
--- a/core/src/main/java/org/apache/calcite/materialize/MaterializationService.java
+++ b/core/src/main/java/org/apache/calcite/materialize/MaterializationService.java
@@ -50,7 +50,6 @@ import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.PriorityQueue;
 import java.util.Set;
 
 import static org.apache.calcite.linq4j.Nullness.castNonNull;
@@ -241,8 +240,7 @@ public class MaterializationService {
     // TODO: Use a partially-ordered set data structure, so we are not scanning
     // through all tiles.
     if (!exact) {
-      final PriorityQueue<Pair<CalciteSchema.TableEntry, TileKey>> queue =
-          new PriorityQueue<>(1, C);
+      Pair<CalciteSchema.TableEntry, TileKey> bestCandidate = null;
       for (Map.Entry<TileKey, MaterializationKey> entry
           : actor.keyByTile.entrySet()) {
         final TileKey tileKey2 = entry.getKey();
@@ -254,12 +252,12 @@ public class MaterializationService {
           final CalciteSchema.TableEntry tableEntry =
               checkValid(materializationKey);
           if (tableEntry != null) {
-            queue.add(Pair.of(tableEntry, tileKey2));
+            Pair<CalciteSchema.TableEntry, TileKey> candidate = Pair.of(tableEntry, tileKey2);
+            if (bestCandidate == null || C.compare(candidate, bestCandidate) < 0) {
+              bestCandidate = candidate;
+            }
           }
         }
-      }
-      if (!queue.isEmpty()) {
-        return queue.peek();
       }
     }
 


core/src/main/java/org/apache/calcite/materialize/MaterializationService.java
```diff
@@ -241,8 +240,7 @@ private MaterializationService() {
     // TODO: Use a partially-ordered set data structure, so we are not scanning
     // through all tiles.
     if (!exact) {
-      final PriorityQueue<Pair<CalciteSchema.TableEntry, TileKey>> queue =
-          new PriorityQueue<>(1, C);
+      Pair<CalciteSchema.TableEntry, TileKey> bestCandidate = null;
```
mihaibudiu, line 243: can the TODO comment above be deleted?

mihaibudiu (approved):
There are no tests - is there any code calling this function?
If yes, then it's perhaps fine - the behavior is not supposed to change.

ILuffZhe:
> There are no tests - is there any code calling this function? If yes, then it's perhaps fine - the behavior is not supposed to change.

+1.

core/src/main/java/org/apache/calcite/materialize/MaterializationService.java
ILuffZhe: > can the TODO comment above be deleted?

+1 with Mihai.

xiaochen-zhou:
> There are no tests - is there any code calling this function? If yes, then it's perhaps fine - the behavior is not supposed to change.

RelOptLattice#getAggregate() call this method.

core/src/main/java/org/apache/calcite/materialize/MaterializationService.java
xiaochen-zhou: > can the TODO comment above be deleted?

I don't think it can be deleted yet. The keyByTile data structure needs to be changed from a Map to a partially-ordered set data structure to avoid scanning through all tiles. Only then can it be removed.
image

xiaochen-zhou: > > can the TODO comment above be deleted?
> 
> I don't think it can be deleted yet. The keyByTile data structure needs to be changed from a Map to a partially-ordered set data structure to avoid scanning through all tiles. Only then can it be removed. @ILuffZhe 
 image

mihaibudiu:
you can probably squash the commits

xiaochen-zhou:
> you can probably squash the commits

Done.

mihaibudiu:
There are some comments in Jira from @julianhyde, so I will wait to merge this.

xiaochen-zhou:
> There are some comments in Jira from @julianhyde, so I will wait to merge this.

OK.

mihaibudiu:
What is the conclusion from the JIRA discussion? Can we merge this?

xiaochen-zhou:
> What is the conclusion from the JIRA discussion? Can we merge this?

I recently attempted to add a benchmark by referring to CALCITE-4994 and CALCITE-3827.

mihaibudiu:
So should we wait for benchmark results?

xiaochen-zhou:
> So should we wait for benchmark results?

Yes.

core/src/main/java/org/apache/calcite/materialize/MaterializationService.java
```diff
@@ -254,13 +252,13 @@ && allSatisfiable(measureList, tileKey2)) {
...
           if (tableEntry != null) {
-            queue.add(Pair.of(tableEntry, tileKey2));
+            Pair<CalciteSchema.TableEntry, TileKey> candidate = Pair.of(tableEntry, tileKey2);
+            if (bestCandidate == null || C.compare(candidate, bestCandidate) < 0) {
+              bestCandidate = candidate;
+            }
           }
         }
       }
-      if (!queue.isEmpty()) {
```
rubenada, line 261: Shouldn't we have a
```
if (bestCandidate != null)
  return bestCandidate;
```
at this place in the patch to maintain the previous logic?

xiaochen-zhou: > Shouldn't we have a
> 
> ```
> if (bestCandidate != null)
>   return bestCandidate;
> ```
> 
> at this place in the patch to maintain the previous logic?
 
I think it's needed, I'll go and modify the code.

xiedeyantu:
Will you continue to complete this PR?

xiaochen-zhou:
> Will you continue to complete this PR?

Sorry, I couldn't find a suitable method to test the performance. @mihaibudiu

mihaibudiu:
I will merge this PR even without an obvious perf improvement, since it makes the code simpler.

PR #4126, opened 2025-01-07
[CALCITE-6745] UDF without parameters cannot be validated when use default conformance

strongduanmu:
- Jira link: https://issues.apache.org/jira/browse/CALCITE-6745


core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java
```diff
@@ -368,6 +366,20 @@ private static SqlOperator toOp(SqlIdentifier name,
     }
   }
 
+  private static SqlSyntax getSqlSyntax(org.apache.calcite.schema.Function function,
+      CalciteConnectionConfig config) {
+    if (!function.getParameters().isEmpty()) {
+      return SqlSyntax.FUNCTION;
+    }
+    // Keep compatible with both Foo() and Foo function syntax for Calcite's default conformance
```
ILuffZhe, line 374: Does this change only affect UDF behavior?

strongduanmu: Yes, it only affects user-defined UDF functions, not other built-in functions.

ILuffZhe (approved):

PR #4127, opened 2025-01-08
[CALCITE-6764] Field access from a nullable ROW should be nullable

mihaibudiu:


core/src/main/java/org/apache/calcite/rex/RexChecker.java
```diff
@@ -160,15 +155,7 @@ public int getFailureCount() {
       ++failCount;
       return litmus.fail(null);
     }
-    final RelDataTypeField typeField = refType.getFieldList().get(index);
```
NobiGo, line 163: In the process of my use, the verification here is indeed effective. If only the nullability value is different, we can add 'RelOptUtil. eqIgnoreNullable'` to replace it?

core/src/main/java/org/apache/calcite/rex/RexFieldAccess.java
```diff
@@ -65,6 +67,18 @@ public class RexFieldAccess extends RexNode {
     this.expr = expr;
```
NobiGo, line 67: We can directly use ` this(expr, field, field.getType());` .

core/src/test/java/org/apache/calcite/test/TableInRootSchemaTest.java
```diff
@@ -73,4 +87,63 @@ class TableInRootSchemaTest {
...
+  }
+
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6764">[CALCITE-6764]
+   * Field access from a nullable ROW should be nullable</a>. */
+  @Test void testNullableValue() throws Exception {
+    Connection connection = DriverManager.getConnection("jdbc:calcite:");
+    CalciteConnection calciteConnection = connection.unwrap(CalciteConnection.class);
+    calciteConnection.getRootSchema().add("T", new RowTable());
+    Statement statement = calciteConnection.createStatement();
+    ResultSet resultSet = statement.executeQuery("SELECT P['a'].K FROM T");
```
NobiGo, line 144: Do we need to verify the metadata`ResultSet#getMetaData` is nullable? I'm not sure if this is possible.

core/src/main/java/org/apache/calcite/rex/RexChecker.java
mihaibudiu: Ok, I will add this check back.

core/src/test/java/org/apache/calcite/test/TableInRootSchemaTest.java
mihaibudiu: There is no type information in the result set, just concrete values as far as I can tell.

mihaibudiu: Indeed, the metadata only returns `VARCHAR` for the column type, irrespective of the nullability.

NobiGo (approved):

PR #4129, opened 2025-01-08
[CALCITE-6766] Move DATEADD and DATEDIFF from PostgreSQL to Redshift library

ILuffZhe:


babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
```diff
@@ -189,8 +189,8 @@ class BabelParserTest extends SqlParserTest {
     sql("select date(x) from t").ok(expected);
   }
 
-  /** In Redshift, PostgreSQL the DATEADD, DATEDIFF and DATE_PART functions have
-   * ordinary function syntax except that its first argument is a time unit
+  /** The DATEADD, DATEDIFF(in Redshift, Snowflake) and DATE_PART(in PostgreSQL)
```
mihaibudiu, line 192: space before opening "("?

mihaibudiu (approved):

babel/src/test/java/org/apache/calcite/test/BabelParserTest.java
ILuffZhe: Done.

mihaibudiu:
I have added a tiny commit to fix a bug I introduced in https://github.com/apache/calcite/pull/4116.
Unfortunately I could not create easily a reproduction, since the calcite simplifier by default removes the casts that are incorrectly translated. Thus no existing tests are affected.

PR #4130, opened 2025-01-08
[CALCITE-6770] Preserve column names when casts are inserted in projects

mihaibudiu:


core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -127,15 +127,15 @@ class JdbcAdapterTest {
...
+            + "    JdbcProject(ENAME=[CAST($1):VARCHAR(30)])\n"
             + "      JdbcFilter(condition=[>(CAST($0):INTEGER NOT NULL, 10)])\n"
             + "        JdbcTableScan(table=[[SCOTT, EMP]])")
         .runs()
         .enable(CalciteAssert.DB == CalciteAssert.DatabaseInstance.HSQLDB)
         .planHasSql("SELECT \"store_name\"\n"
                 + "FROM \"foodmart\".\"store\"\n"
                 + "WHERE \"store_id\" < 10")
-        .planHasSql("SELECT CAST(\"ENAME\" AS VARCHAR(30))\n"
+        .planHasSql("SELECT CAST(\"ENAME\" AS VARCHAR(30)) AS \"ENAME\"\n"
```
ILuffZhe, line 138: What if there are duplicate `ENAME` in this case? Will the 2nd `ENAME` be `ENAME0`?

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -257,6 +257,15 @@ public static void checkActualAndReferenceFiles() {
     sql(sql).ok();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6770">[CALCITE-6770]
+   * Preserve column names when casts are inserted in projects</a>. */
+  @Test void testCastNames() {
+    final String sql = "SELECT * FROM (SELECT empno, 'x' AS X FROM emp) "
+        + "UNION ALL (SELECT empno, 'xx' AS X from EMP)";
```
ILuffZhe, line 265: from `emp`?

ILuffZhe (approved):
LGTM, just left some minor questions.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
mihaibudiu: This preserves the names from the original select list that is being transformed. 
If the original list had duplicates, this one should too.
But I expect that this cannot happen at this point.

mihaibudiu: I mean, when the names are inserted, the program has been validated.

mihaibudiu:
I have added a tiny commit to fix a bug I have introduced in https://github.com/apache/calcite/pull/4116.
No tests are affected, since the calcite simplification code removes the casts in question before they reach the bug.

rongrong:
This actually breaks the following SQL:
```
SELECT
  CASE
    WHEN ID IS NOT NULL
    THEN foo
    ELSE CAST(ID AS VARCHAR)
  END AS ID
FROM testTable;
```
because it would rewrite the THEN clause to `THEN CAST(foo AS VARCHAR) AS foo` which is no longer valid SQL. Not all places accept identifiers accept alias.

mihaibudiu:
Can you please file an issue about this and assign it to me?
https://issues.apache.org/jira/projects/CALCITE/issues
(If you can't assign it I will find it anyway)

mihaibudiu:
The change is in coerceColumnType, so I was assuming your case does not apply, since foo is not a column.

rongrong:
I need to sign up for an account to create an issue. That's a lot of effort to be able to file a bug...  Generally speaking, if your goal is to make sure in SELECT the name would be preserved after adding CAST to an identifier, you should handle that in SELECT. Adding it to type coercion can cause a lot of issues, this is just one way of breaking it. Theoretically you can have situations where `a + b` is rewritten to `CAST(a AS xxx) as a + b`, which, is also invalid SQL.

mihaibudiu:
That's what I am saying, the function is called `coerceColumnType`, so I assumed it is invoked only on a `SELECT` item.
I will try to create a reproduction and if I succeed I will file an issue.

rongrong:
It's used in selectlist, case statement, in, and source row, all in `TypeCoercionImpl`.

mihaibudiu:
I see that the AS is inserted in the CASE statement, but I have trouble creating a test that fails. How did you observe this error?

rongrong:
> This actually breaks the following SQL:
> 
> ```
> 
> SELECT
> 
>   CASE
> 
>     WHEN ID IS NOT NULL
> 
>     THEN foo
> 
>     ELSE CAST(ID AS VARCHAR)
> 
>   END AS ID
> 
> FROM testTable;
> 
> ```
> 
> because it would rewrite the THEN clause to `THEN CAST(foo AS VARCHAR) AS foo` which is no longer valid SQL. Not all places accept identifiers accept alias.

This sql will fail.

mihaibudiu:
I understand that, but i could not write a test that fails. Can you give more details about the configuration of your compiler and how the error manifests itself?

mihaibudiu:
I figured out how to write a test, one way to do it is to configure the validator withIdentifierExpansion(true).
The rest should be easy.

mihaibudiu:
I have filed https://issues.apache.org/jira/browse/CALCITE-7072 and I will submit a fix shortly.
Thank you for the bug report.
But JIRA is really the place to file issues, people may not monitor closed PRs.

mihaibudiu:
@rongrong can you validate the fix in https://github.com/apache/calcite/pull/4439?

PR #3848, opened 2024-07-08
[CALCITE-6413] Add necessary casts to conditions for NATURAL/USING join

xtern:
https://issues.apache.org/jira/browse/CALCITE-6413

PR #4131, opened 2025-01-08
[CALCITE-6413] SqlValidator does not invoke TypeCoercionImpl::binaryC…

mihaibudiu:
…omparisonCoercion for both NATURAL and USING join conditions

This is a simplified PR based on the code from https://github.com/apache/calcite/pull/3848

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -3433,7 +3433,9 @@ public class SqlToRelConverter {
     final SqlNameMatcher nameMatcher = catalogReader.nameMatcher();
     final List<RexNode> list = new ArrayList<>();
     for (String name : nameList) {
-      List<RexNode> operands = new ArrayList<>();
+      // For each field joined on we compare the types from the left and right relation
+      List<RexNode> operands = new ArrayList<>(2);
+      List<RelDataType> comparedTypes = new ArrayList<>(2);
       int offset = 0;
       for (SqlValidatorNamespace n : ImmutableList.of(leftNamespace,
           rightNamespace)) {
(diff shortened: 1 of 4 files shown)


mihaibudiu:
Small PR fixing a genuine bug. Should not be too difficult to review. 
It essentially treats NATURAL and USING joins in the same way standard JOINs are already treated: if a comparison requires a cast, the cast is inserted in the program.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -3443,12 +3444,32 @@ private RexNode convertUsing(SqlValidatorNamespace leftNamespace,
           throw new AssertionError("field " + name + " is not found in "
               + rowType + " with " + nameMatcher);
         }
+
+        fieldTypes.add(field.getType());
```
ILuffZhe, line 3448: Can we just use `rowType` here?

NobiGo (approved):
LGTM

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -3443,12 +3444,32 @@ private RexNode convertUsing(SqlValidatorNamespace leftNamespace,
...
                 offset + field.getIndex()));
         offset += rowType.getFieldList().size();
       }
-      list.add(rexBuilder.makeCall(SqlStdOperatorTable.EQUALS, operands));
+
+      RelDataType resultType =
+          validator().getTypeCoercion().commonTypeForBinaryComparison(
+              fieldTypes.get(0), fieldTypes.get(1));
+      if (resultType == null) {
+        throw new IllegalArgumentException("Cannot compute compatible row type: " + fieldTypes);
```
ILuffZhe, line 3459: `row type` or `row types`?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
mihaibudiu: This list just contains two fields, one from the left operand of the join and one from the right operand.
The list is created for each pair of fields that are joined on.
It's not really row type.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
mihaibudiu: I guess this exception message is not very good, I will change it.
What happens here is that the join is done on two fields which cannot be compared.
But I am not sure this exception can ever happen, because at this point the program is validated already, so if the fields are not comparable, this point should not be reached.

ILuffZhe: Thanks for the clarify, I think you are right about whether exception can be thrown here.

mihaibudiu: I will try to add a test to see what happens.

mihaibudiu: I have checked, the validator will reject a program with a message like 
`Column 'X' matched using NATURAL keyword or USING clause has incompatible types: cannot compare`
I will add a comment that the exception is supposed to be unreachable.

mihaibudiu: I have made the code clearer - but by mistake I amended the commit instead of adding a new one.

ILuffZhe (approved):
LGTM

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
ILuffZhe: Now the exception message is more clear.

suibianwanwank (approved):
LGTM

PR #4132, opened 2025-01-08
[CALCITE-6771] Convert Type from FLOAT to DOUBLE in PrestoDialect

xiaochen-zhou:
Since FLOAT type is not supported in Presto

According to the documentation: https://prestodb.io/docs/current/language/types.html#real
image
Using CAST(xxxx AS FLOAT) will result in an error:

```

SELECT CAST(message_count AS FLOAT)

FROM applydata_bigdata.kafka_topic_cluster_user_dept_month_one_day_msg_info

WHERE ymd = 20250105

```
image
It should be converted to the DOUBLE type instead.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
@@ -24,7 +24,9 @@ import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlBasicCall;
+import org.apache.calcite.sql.SqlBasicTypeNameSpec;
 import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlDataTypeSpec;
 import org.apache.calcite.sql.SqlDialect;
 import org.apache.calcite.sql.SqlIntervalQualifier;
 import org.apache.calcite.sql.SqlKind;
@@ -146,7 +148,14 @@ public class PrestoSqlDialect extends SqlDialect {
   }
 
   @Override public @Nullable SqlNode getCastSpec(RelDataType type) {
-    return super.getCastSpec(type);
+    switch (type.getSqlTypeName()) {
+    // PRESTO only supports REAL、DOUBLE for floating point types.
+    case FLOAT:
+      return new SqlDataTypeSpec(
+          new SqlBasicTypeNameSpec(SqlTypeName.DOUBLE, SqlParserPos.ZERO), SqlParserPos.ZERO);
+    default:
+      return super.getCastSpec(type);
+    }
   }
 
   @Override public void unparseCall(SqlWriter writer, SqlCall call,
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -9093,6 +9093,20 @@ class RelToSqlConverterTest {
     sql(query).withOracle().ok(oracle);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6771">[CALCITE-6771]
+   * Convert Type from FLOAT to DOUBLE in PrestoDialect</a>. */
+  @Test void testPrestoFloatingPointTypesCast() {
+    String query = "SELECT CAST(\"department_id\" AS float), "
+        + "CAST(\"department_id\" AS double), "
+        + "CAST(\"department_id\" AS real) FROM \"employee\"";
+    String expected = "SELECT CAST(\"department_id\" AS DOUBLE), "
+        + "CAST(\"department_id\" AS DOUBLE), "
+        + "CAST(\"department_id\" AS REAL)\nFROM \"foodmart\".\"employee\"";
+    sql(query)
+        .withPresto().ok(expected);
+  }
+
   /** Fluid interface to run tests. */
   static class Sql {
     private final CalciteAssert.SchemaSpec schemaSpec;


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -9093,6 +9093,18 @@ private void checkLiteral2(String expression, String expected) {
     sql(query).withOracle().ok(oracle);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6771">[CALCITE-6771]
+   * Convert Type from FLOAT to DOUBLE in PrestoDialect </a>. */
+  @Test void testPrestoCastToFloat() {
+    String query = "SELECT CAST(\"department_id\" AS float) FROM \"employee\"";
```
NobiGo, line 9100: Maybe we can add a unit test to cover `CAST(xxxx as REAL)`  and `CAST(xxxx as DOUBLE)`.

xiaochen-zhou: > Maybe we can add a unit test to cover `CAST(xxxx as REAL)` and `CAST(xxxx as DOUBLE)`.

Done.

NobiGo: This looks redundant, can it be merged into one SQL statement？

NobiGo: I have already approved it, whether or not to make changes depends on you.

core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
```diff
@@ -146,7 +148,14 @@ private static void unparseUsingLimit(SqlWriter writer, @Nullable SqlNode offset
   }
 
   @Override public @Nullable SqlNode getCastSpec(RelDataType type) {
-    return super.getCastSpec(type);
+    switch (type.getSqlTypeName()) {
+    // PRESTO only supports REAL、DOUBLE for floating point types.
+    case FLOAT:
+      return new SqlDataTypeSpec(
+          new SqlBasicTypeNameSpec(SqlTypeName.DOUBLE, SqlParserPos.ZERO), SqlParserPos.ZERO);
```
caicancai, line 155: I think it is more reasonable to convert it to SqlTypeName.REAL

In some databases REAL is equivalent to Float

caicancai, line 155: I think it is more reasonable to convert it to SqlTypeName.REAL. In some database systems, REAL is equivalent to float, of course, this depends on the real implementation of calcite?
cc @NobiGo

xiaochen-zhou: Calcite FLOAT and DOUBLE are the same type, 
image

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xiaochen-zhou: > I have already approved it, whether or not to make changes depends on you.

OK.

core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
xiaochen-zhou: > I think it is more reasonable to convert it to SqlTypeName.REAL. In some database systems, `REAL` is equivalent to float, of course, this depends on the `REAL` implementation of calcite? cc @NobiGo

Calcite FLOAT and DOUBLE are the same type, 
image 

I think it should be converted to double in Presto.

image

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xiaochen-zhou: > This looks redundant, can it be merged into one SQL statement？

OK.

core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
caicancai: thanks

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xiaochen-zhou: > This looks redundant, can it be merged into one SQL statement？

Done.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -9093,6 +9093,20 @@ private void checkLiteral2(String expression, String expected) {
     sql(query).withOracle().ok(oracle);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6771">[CALCITE-6771]
+   * Convert Type from FLOAT to DOUBLE in PrestoDialect </a>. */
```
NobiGo, line 9098: Minor: No space in `PrestoDialect` and `</a>`;

NobiGo (changes requested):

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xiaochen-zhou: > Minor: No space in `PrestoDialect` and `</a>`;

Done.

mihaibudiu:
I will let @NobiGo merge this, since he requested changes.

NobiGo (approved):

PR #4135, opened 2025-01-09
[CALCITE-6776] Multiple expanded `IS NOT DISTINCT FROM` cannot be col…

linorosa:
…lapsed back

### DIFF
--- a/core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
+++ b/core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
@@ -1747,32 +1747,35 @@ public abstract class RelOptUtil {
    * and
    * {@link #splitJoinCondition(List, List, RexNode, List, List, List, List)}.
    *
-   * <p>If the given expr <code>call</code> is an expanded version of
+   * <p>If the given expr <code>rexCall</code> contains an expanded version of
    * {@code IS NOT DISTINCT FROM} function call, collapses it and return a
    * {@code IS NOT DISTINCT FROM} function call.
    *
    * <p>For example: {@code t1.key IS NOT DISTINCT FROM t2.key}
-   * can rewritten in expanded form as
+   * can be rewritten in expanded form as
    * {@code t1.key = t2.key OR (t1.key IS NULL AND t2.key IS NULL)}.
    *
-   * @param call       Function expression to try collapsing
+   * @param rexCall       Function expression to try collapsing
    * @param rexBuilder {@link RexBuilder} instance to create new {@link RexCall} instances.
-   * @return If the given function is an expanded IS NOT DISTINCT FROM function call,
-   *         return a IS NOT DISTINCT FROM function call. Otherwise return the input
-   *         function call as it is.
+   * @return A function where all IS NOT DISTINCT FROM are collapsed.
    */
-  public static RexCall collapseExpandedIsNotDistinctFromExpr(final RexCall call,
+  public static RexCall collapseExpandedIsNotDistinctFromExpr(final RexCall rexCall,
       final RexBuilder rexBuilder) {
-    switch (call.getKind()) {
-    case OR:
-      return doCollapseExpandedIsNotDistinctFromOrExpr(call, rexBuilder);
+    final RexShuttle shuttle = new RexShuttle() {
+      @Override public RexNode visitCall(RexCall call) {
+        RexCall recursivelyExpanded = (RexCall) super.visitCall(call);
 
-    case CASE:
-      return doCollapseExpandedIsNotDistinctFromCaseExpr(call, rexBuilder);
-
-    default:
-      return call;
-    }
+        switch (recursivelyExpanded.getKind()) {
+        case OR:
+          return doCollapseExpandedIsNotDistinctFromOrExpr(recursivelyExpanded, rexBuilder);
+        case CASE:
+          return doCollapseExpandedIsNotDistinctFromCaseExpr(recursivelyExpanded, rexBuilder);
+        default:
+          return recursivelyExpanded;
+        }
+      }
+    };
+    return (RexCall) rexCall.accept(shuttle);
   }
 
   private static RexCall doCollapseExpandedIsNotDistinctFromOrExpr(final RexCall call,
--- a/core/src/test/java/org/apache/calcite/plan/RelOptUtilTest.java
+++ b/core/src/test/java/org/apache/calcite/plan/RelOptUtilTest.java
@@ -34,6 +34,7 @@ import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
 import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.schema.SchemaPlus;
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -1755,24 +1755,30 @@ private static void splitJoinCondition(
...
-      return call;
-    }
+    final RexShuttle shuttle = new RexShuttle() {
+      @Override public RexNode visitCall(RexCall call) {
+        switch (call.getKind()) {
+        case OR:
+          return doCollapseExpandedIsNotDistinctFromOrExpr(call, rexBuilder);
+        case CASE:
+          return doCollapseExpandedIsNotDistinctFromCaseExpr(call, rexBuilder);
+        default:
```
asolimando, line 1773: Here you are changing the default from no-op to assuming it's a conjunction, I guess we might run into problems. I suggest to add an explicit case for `AND` and leave the default case as it used to be.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -1755,24 +1755,30 @@ private static void splitJoinCondition(
...
    *
-   * @param call       Function expression to try collapsing
+   * @param rexNode       Function expression to try collapsing
    * @param rexBuilder {@link RexBuilder} instance to create new {@link RexCall} instances.
    * @return If the given function is an expanded IS NOT DISTINCT FROM function call,
    *         return a IS NOT DISTINCT FROM function call. Otherwise return the input
    *         function call as it is.
    */
-  public static RexCall collapseExpandedIsNotDistinctFromExpr(final RexCall call,
+  public static RexNode collapseExpandedIsNotDistinctFromExpr(final RexNode rexNode,
```
asolimando, line 1764: We can't change the signature of public methods to avoid breaking backward compatibility, you can create a new method with this signature and call the other method with an appropriate cast, and mark the original method as deprecated (grep "@Deprecated // to be removed before 2.0" and you will find many examples).

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -1755,24 +1755,30 @@ private static void splitJoinCondition(
    * can rewritten in expanded form as
```
asolimando, line 1755:
```suggestion
   * can be rewritten in expanded form as
```

I know it's not part of your PR but since I noticed it, let's take the chance to fix it if you agree on the change

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -1755,24 +1755,30 @@ private static void splitJoinCondition(
...
    * {@code t1.key = t2.key OR (t1.key IS NULL AND t2.key IS NULL)}.
    *
-   * @param call       Function expression to try collapsing
+   * @param rexNode       Function expression to try collapsing
    * @param rexBuilder {@link RexBuilder} instance to create new {@link RexCall} instances.
    * @return If the given function is an expanded IS NOT DISTINCT FROM function call,
    *         return a IS NOT DISTINCT FROM function call. Otherwise return the input
```
asolimando, line 1761:
```suggestion
   *         return a IS NOT DISTINCT FROM function call. Otherwise, return the input
```
same as above

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
mihaibudiu, line 1764: you are changing the signature of a public function, which technically cannot be done without changing the major version of the project (it is an incompatible change).

I am afraid you will have to create a new function instead, and perhaps mark this one as deprecated.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -1755,24 +1755,30 @@ private static void splitJoinCondition(
...
-    }
+    final RexShuttle shuttle = new RexShuttle() {
+      @Override public RexNode visitCall(RexCall call) {
+        switch (call.getKind()) {
+        case OR:
+          return doCollapseExpandedIsNotDistinctFromOrExpr(call, rexBuilder);
+        case CASE:
+          return doCollapseExpandedIsNotDistinctFromCaseExpr(call, rexBuilder);
+        default:
+          final List<RexNode> collapsedOperands = call.getOperands().stream().map(x ->
```
mihaibudiu, line 1774: Is this call necessary?
won't the shuttle do this anyway if you call the super.visitCall() function?

core/src/test/java/org/apache/calcite/plan/RelOptUtilTest.java
asolimando, line 1: @linorosa can you also show a test case with an expression mixing AND/OR/CASE just to validate that we aren't missing anything?

asolimando (changes requested):
Minor issues apart from the signature change which has to be addressed

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
linorosa: I see your point. You mean something like this?

```java
@Override public RexNode visitCall(RexCall call) {
        switch (call.getKind()) {
        case OR:
          return doCollapseExpandedIsNotDistinctFromOrExpr(call, rexBuilder);
        case CASE:
          return doCollapseExpandedIsNotDistinctFromCaseExpr(call, rexBuilder);
        case AND:
          return super.visitCall(call); // simplified this to take advantage of the recursive RexShuttle calls
        default:
          return call;
        }
```

The reason I changed the default case is I was afraid to miss some logical operators if I went that route. I can think of `NOT` for example, but there might be a bunch others

asolimando: Yes, that's what I have in mind (incorporating also @mihaibudiu's suggestion which is correct).

For sure it's incomplete, but it's sound, which is preferable, IMO, as it's very hard to reason about corner cases without adding at least an extra test per construct.

Of course nothing prevents to add more "cases" for `NOT` etc., it's not strictly needed in the context of this PR as it is already an improvement over the previous version and I don't think it is a regression on what is supported.

If you feel like adding few trivial cases like `NOT`, I encourage you to write tests similar to the ones you have added, featuring more constructs (I have already suggested to write a more complex test to cover more deeply nested examples mixing what we currently support).

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
linorosa: thanks, I reverted back to the old signature

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
linorosa: good point 👍

core/src/test/java/org/apache/calcite/plan/RelOptUtilTest.java
```diff
@@ -728,6 +728,42 @@ private void splitJoinConditionHelper(RexNode joinCond, List<Integer> expLeftKey
     assertThat(castNode2.explain(), is(expectNode2.explain()));
   }
 
+  /**
+   * Test that {@link RelOptUtil#collapseExpandedIsNotDistinctFromExpr(RexNode, RexBuilder)}
+   * can handle collapsing IS NOT DISTINCT FROM nested within other expressions.
+   */
+  @Test void testCollapseExpandedIsNotDistinctFromNestedConditions() {
```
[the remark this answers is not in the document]
linorosa: I moved it up closer to some more related tests

mihaibudiu:
when this is ready please ask for another review

linorosa:
just did @mihaibudiu . A CI check failed but doesn't seem code related? 

```

* What went wrong:
--
more_vert | 75 | 02:53:00 PM | A problem occurred configuring project ':mongodb'.
more_vert | 76 | 02:53:00 PM | > Could not read workspace metadata from /home/jenkins/.gradle/caches/8.7/kotlin-dsl/accessors/e87f145b84ad0f218a607cdc62ed5d1a/metadata.bin


```

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -1752,27 +1752,33 @@ private static void splitJoinCondition(
...
+        switch (call.getKind()) {
+        case OR:
+          return doCollapseExpandedIsNotDistinctFromOrExpr(call, rexBuilder);
+        case CASE:
+          return doCollapseExpandedIsNotDistinctFromCaseExpr(call, rexBuilder);
+        case NOT:
+        case AND:
+          return super.visitCall(call);
+        default:
+          return call;
```
mihaibudiu, line 1777: are you sure about this?
what if the root of the tree is a plus?

linorosa: humm then `visitCall` would fall into the `default` case and return the unaltered "plus" no? Isn't that what we want?

asolimando: @mihaibudiu that was my idea too, I have in mind cases like `NULLIF` that would require special handling other than `super.visitCall(call)`.

This was the original behavior, covering `CASE` and `OR` and bailing out for all other cases.

If you think defaulting to `super.visitCall(call)` might work I am OK trying but then I'd like to see tests covering most constructs to gain confidence that we introducing an unsound change.

Btw, looking at how `CASE` and `OR` are handled, their methods seems to just look for syntactical variants of  `IS NOT DISTINCT FROM`, they don't look recursively in their operands, so I guess that if you try `OR(NOT(IS DISTINCT FROM), ...)` it won't already work as we expect.

If we want to support fully recursive detection we should improve our tests first IMO.

What do you guys think?

linorosa: Yeah @asolimando my first instinct was to recursively collapse any `RexCall` - and you're right I didn't recursively visit `OR`/`CASE` to collapse them. I didn't want to rock the boat too much. 

That said, I'm open to doing "the right thing" and rewrite this to collapse anything recursively (i.e. both moving `return super.visitCall(call);` to the `default` case and also taking care of collapsing `OR`/`CASE` operands.

The thing is, is it realistic to condition this change to adding tests for everything under `SqlStdOperatorTable.*`? Or do you think you'd be comfortable with testing some subset of them?

mihaibudiu: one or two tests should be enough.
it's most useful to cover the cases that arise in practice

linorosa: Do you want to suggest which operators I should focus on? @asolimando mentioned `NULLIF` and you mentioned `PLUS`. Are we good with testing those two?

mihaibudiu: if one works, they all will work

asolimando: Great @linorosa, good to hear that,I think that having a fully recursive implementation would be useful.

If you write a few more complex tests with some good mix of operators I think it's fine, as @mihaibudiu suggested.

I skimmed the changes and they LGTM, if we manage to get this last change to the PR, it will be good to go.

Feel free ping me whenever you feel it's ready to be reviewed again.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -1757,24 +1757,26 @@ private static void splitJoinCondition(
    *
    * @param rexCall       Function expression to try collapsing
    * @param rexBuilder {@link RexBuilder} instance to create new {@link RexCall} instances.
-   * @return If the given function is an expanded IS NOT DISTINCT FROM function call,
-   *         return a IS NOT DISTINCT FROM function call. Otherwise, return the input
-   *         function call as it is.
+   * @return A rexCall that's semantically equivalent to the original rexCall, with all
```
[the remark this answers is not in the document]
linorosa: @asolimando No problem, just made changed it back to "function". Also made a couple of changes:

* Reworked the tests so it's hopefully more clear which isolated concepts they're actually testing. Happy to go whichever way you suggest.
* Added support for deeply nested `IS NOT DISTINCT FROM`. Not sure this ever arrives in practice - or whether it's even valid SQL - for for completeness sake.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -1747,40 +1747,46 @@ private static void splitJoinCondition(
...
+   * can be rewritten in expanded form as
    * {@code t1.key = t2.key OR (t1.key IS NULL AND t2.key IS NULL)}.
    *
-   * @param call       Function expression to try collapsing
+   * @param rexCall       Function expression to try collapsing
    * @param rexBuilder {@link RexBuilder} instance to create new {@link RexCall} instances.
-   * @return If the given function is an expanded IS NOT DISTINCT FROM function call,
-   *         return a IS NOT DISTINCT FROM function call. Otherwise return the input
-   *         function call as it is.
+   * @return A function that's semantically equivalent to the original, with all
```
mihaibudiu, line 1760: I find this description less useful than it could be. It's not obvious what the "original" form is.
Why not just say "expression blah is converted to expression foo"?

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -1747,40 +1747,46 @@ private static void splitJoinCondition(
...
-    case OR:
-      return doCollapseExpandedIsNotDistinctFromOrExpr(call, rexBuilder);
+    final RexShuttle shuttle = new RexShuttle() {
+      @Override public RexNode visitCall(RexCall call) {
+        SqlKind callKind  = call.getKind();
+        int operandCount = call.getOperands().size();
 
-    case CASE:
-      return doCollapseExpandedIsNotDistinctFromCaseExpr(call, rexBuilder);
+        boolean isExpandedOrExpr = callKind == SqlKind.OR && operandCount == 2;
```
mihaibudiu, line 1771: these variables should be probably called maybeExpandedOrExpression. they are a quick heuristic test to filter out analyzing some cases.

core/src/test/java/org/apache/calcite/plan/RelOptUtilTest.java
```diff
@@ -707,9 +930,9 @@ private void splitJoinConditionHelper(RexNode joinCond, List<Integer> expLeftKey
     final RelDataType castRowType2 = typeFactory
         .createStructType(
             ImmutableList.of(
-            Pair.of(fieldEmpno.getName(), fieldEmpno.getType()),
-            Pair.of(fieldEname.getName(), fieldEname.getType()),
-            Pair.of("JOB_CNT2", fieldJobCnt.getType())));
+                Pair.of(fieldEmpno.getName(), fieldEmpno.getType()),
```
mihaibudiu, line 933: I find it surprising that the style checker hasn't complained about these in the past.

core/src/test/java/org/apache/calcite/plan/RelOptUtilTest.java
```diff
@@ -420,6 +421,228 @@ private void splitJoinConditionHelper(RexNode joinCond, List<Integer> expLeftKey
...
+  }
+
+  /**
+   * Test that {@link RelOptUtil#collapseExpandedIsNotDistinctFromExpr(RexCall, RexBuilder)}
+   * can handle collapsing IS NOT DISTINCT FROM composed of other RexNodes.
+   */
+  @Test void testCollapseExpandedIsNotDistinctFromOnContainingRexNodes() {
+    final RexBuilder rexBuilder = relBuilder.getRexBuilder();
+
+    final RexNode leftEmpNo =
```
mihaibudiu, line 615: these expression trees are not as easy to read, a comment describing the expression in SQL syntax would help a lot.

linorosa: sure thing, I added some comments now

core/src/test/java/org/apache/calcite/plan/RelOptUtilTest.java
linorosa: ouch, I reverted these. `./gradlew autostyleApply` was giving me trouble and I had to do some manual fixes. I guess I went too far.

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
linorosa: yeah makes sense, I moved them back there

asolimando (approved):
LGTM

mihaibudiu (approved):

mihaibudiu:
yes, I think this will be ready for merging once the commits are squashed

linorosa:
Squashed. Thanks for reviewing!

PR #4136, opened 2025-01-09
[CALCITE-6775] ToChar and ToTimestamp PG implementors should use translator's root instead of creating a new root expression

rubenada:
See CALCITE-6775


asolimando (approved):
LGTM!

PR #4137, opened 2025-01-10
[CALCITE-6777] Conversion between UUID and BINARY produces wrong results

mihaibudiu:


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -284,12 +283,20 @@ public static String uuidToString(UUID uuid) {
   }
 
   public static UUID binaryToUuid(ByteString bytes) {
-    return Conversion.byteArrayToUuid(bytes.getBytes(), 0);
+    if (bytes.length() < 16) {
```
caicancai, line 286: I see that there is no jira link added to the test. Maybe you can consider adding JavaDoc here so that developers can understand why the changes are made here.

PR #4138, opened 2025-01-11
[CALCITE-6779] Casts from UUID to DATE should be invalid

mihaibudiu:


core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java
```diff
@@ -175,7 +175,7 @@ public enum SqlTypeName {
       ImmutableList.of(BOOLEAN);
 
   public static final List<SqlTypeName> BINARY_TYPES =
-      ImmutableList.of(BINARY, VARBINARY, UUID);
+      ImmutableList.of(BINARY, VARBINARY);
```
ILuffZhe, line 178: If this makes casts from UUID to character type valid, I think it would be better to change the summary.

mihaibudiu: Cast to and from characters has not changed

ILuffZhe (approved):
LGTM!

PR #4139, opened 2025-01-11
[CALCITE-6764] Field access from a nullable ROW should be nullable (part 2)

mihaibudiu:
We have already merged a PR fixing this issue, but I have reopened it, and this is a second PR addressing a second bug related to the same problem.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
+++ b/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
@@ -179,15 +179,34 @@ public interface RelDataTypeFactory {
    * Creates a type that is the same as another type but with possibly
    * different nullability. The output type may be identical to the input
    * type. For type systems without a concept of nullability, the return value
-   * is always the same as the input.
+   * is always the same as the input. This function never returns a nullable
+   * record type; when applied to a record type, it recursively makes all fields
+   * nullable and the record itself is non-nullable.
    *
-   * @param type     input type
-   * @param nullable true to request a nullable type; false to request a NOT
-   *                 NULL type
-   * @return output type, same as input type except with specified nullability
+   * @param type input type
+   * @param nullable true to request a nullable type; false to request a NOT NULL type.
+   * @return output type, same as input type except with specified nullability (see description
+   * for record types).
    * @throws NullPointerException if type is null
    */
   RelDataType createTypeWithNullability(
+      RelDataType type,
+      boolean nullable);
+
+  /**
+   * Creates a type that is the same as another type but with possibly
+   * different nullability. The output type may be identical to the input
+   * type. For type systems without a concept of nullability, the return value
+   * is always the same as the input. This differs from {@link
+   * #createTypeWithNullability(RelDataType, boolean)} in the handling of record
+   * types.  This function returns a nullable struct.
+   *
+   * @param type input type
+   * @param nullable true to request a nullable type; false to request a NOT NULL type.
+   * @return output type, same as input type except with specified nullability
+   * @throws NullPointerException if type is null
+   */
+  RelDataType enforceTypeWithNullability(
       RelDataType type,
       boolean nullable);
 
--- a/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
+++ b/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
@@ -410,6 +410,31 @@ public abstract class RelDataTypeFactoryImpl implements RelDataTypeFactory {
     return canonize(newType);
   }
 
+  @Override public RelDataType enforceTypeWithNullability(
+      final RelDataType type,
+      final boolean nullable) {
+    requireNonNull(type, "type");
+    RelDataType newType;
+    if (type.isNullable() == nullable) {
+      newType = type;
+    } else if (type instanceof RelRecordType) {
+      return createStructType(type.getStructKind(),
+          new AbstractList<RelDataType>() {
+            @Override public RelDataType get(int index) {
+              return type.getFieldList().get(index).getType();
+            }
+
+            @Override public int size() {
+              return type.getFieldCount();
+            }
+          },
+          type.getFieldNames(), nullable);
+    } else {
+      newType = copySimpleType(type, nullable);
+    }
+    return canonize(newType);
+  }
+
   /**
    * Registers a type, or returns the existing type if it is already
    * registered.
--- a/core/src/test/java/org/apache/calcite/test/TableInRootSchemaTest.java
+++ b/core/src/test/java/org/apache/calcite/test/TableInRootSchemaTest.java
@@ -87,27 +87,11 @@ class TableInRootSchemaTest {
     connection.close();
   }
 
-  /**
-   * Helper class for the test for [CALCITE-6764] below.
-   */
-  private static class RowTable extends AbstractQueryableTable {
-    protected RowTable() {
+  /** Represents a table with no data. An abstract base class,
+   * derived classes need to define the schema. */
+  private abstract static class EmptyTable extends AbstractQueryableTable {
+    protected EmptyTable() {
       super(Object[].class);
-    }
-
-    @Override public RelDataType getRowType(RelDataTypeFactory typeFactory) {
-      final PairList<String, RelDataType> columnDesc = PairList.withCapacity(1);
-      // Schema contains a column whose type is MAP<VARCHAR, ROW(VARCHAR)>, but
-      // the ROW type can be nullable.
-      final RelDataType colType =
-          typeFactory.createMapType(typeFactory.createSqlType(SqlTypeName.VARCHAR),
-            new RelRecordType(
-                StructKind.PEEK_FIELDS,
-                  ImmutableList.of(
-                      new RelDataTypeFieldImpl("K", 0,
-                          typeFactory.createSqlType(SqlTypeName.VARCHAR))), true));
-      columnDesc.add("P", colType);
-      return typeFactory.createStructType(columnDesc);
     }
 
     @Override public <T> Queryable<T> asQueryable(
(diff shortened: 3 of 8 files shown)


core/src/test/java/org/apache/calcite/test/TableInRootSchemaTest.java
```diff
@@ -134,6 +118,51 @@ protected RowTable() {
...
+                  ImmutableList.of(
+                      new RelDataTypeFieldImpl("K", 0,
+                          typeFactory.createSqlType(SqlTypeName.VARCHAR))), true));
+      columnDesc.add("P", colType);
+      return typeFactory.createStructType(columnDesc);
+    }
+  }
+
+  /** Helper class for the test for [CALCITE-6764] below. */
+  private static class RowTable2 extends EmptyTable {
```
asolimando, line 146: Nit: the difference between `RowTable` and `RowTable2` seems to be in the nullability of the map composing the row type, can we make this explicit by calling them `NullableRowTable` and `RowTable` or something along this line?

core/src/test/java/org/apache/calcite/test/TableInRootSchemaTest.java
```diff
@@ -146,4 +175,17 @@ protected RowTable() {
     statement.close();
     connection.close();
   }
+
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6764">[CALCITE-6764]
+   * Field access from a nullable ROW should be nullable</a>. */
```
asolimando, line 180: I don't see where we are actually testing the result being `NULL`?

asolimando (commented):
Nothing major but I am confused on a couple of points @mihaibudiu, can you take a look?

core/src/test/java/org/apache/calcite/test/TableInRootSchemaTest.java
mihaibudiu: this is implicitly tested by the Validator, which crashes without this fix.
It fails with the same error as described in the JIRA issue: `java.lang.RuntimeException: java.lang.AssertionError: Conversion to relational algebra failed to preserve datatypes`.

asolimando: Let's add it as a comment for who might overlook that the way I did

mihaibudiu:
@asolimando I have pushed a commit which hopefully addresses your comments.

core/src/test/java/org/apache/calcite/test/TableInRootSchemaTest.java
julianhyde: Can you break the line after 'for', as is usual for a 'Test case for ...' javadoc comment.

mihaibudiu: there are lots of examples with either structure, but I have added a newline

mihaibudiu:
I will write a few more tests, so I will mark this PR as draft until I am sure it passes all of them. Hopefully this won't take too long.

core/src/test/java/org/apache/calcite/test/TableInRootSchemaTest.java
```diff
@@ -134,16 +118,84 @@ protected RowTable() {
...
+          true);
+      columnDesc.add("P", colType);
+      return typeFactory.createStructType(columnDesc);
+    }
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6764">[CALCITE-6764]
    * Field access from a nullable ROW should be nullable</a>. */
   @Test void testNullableValue() throws Exception {
```
asolimando, line 169:
```suggestion
  @Test void testNullableRowInMap() throws Exception {
```

core/src/test/java/org/apache/calcite/test/TableInRootSchemaTest.java
```diff
@@ -134,16 +118,84 @@ protected RowTable() {
...
     ResultSet resultSet = statement.executeQuery("SELECT P['a'].K FROM T");
     resultSet.close();
     statement.close();
     connection.close();
   }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6764">[CALCITE-6764]
+   * Field access from a nullable ROW should be nullable</a>. */
+  @Test void testNullableValue2() throws Exception {
```
asolimando, line 187:
```suggestion
  @Test void testNullableRowToplevel() throws Exception {
```

asolimando:
> @asolimando I have pushed a commit which hopefully addresses your comments.

Thanks @mihaibudiu, LGTM, just proposed two improved test names, the rest is good, feel free to ping me for an extra review round once the extra tests are ready, if needed

mihaibudiu:
The complete fix required some additions to the type factory interfaces and making a method public.
As I commented in the jira https://issues.apache.org/jira/browse/CALCITE-6764, I think that the methods from the type factory to create nullable types are broken for record types.

mihaibudiu:
@asolimando I made some changes that probably will require a new review

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
```diff
@@ -179,7 +179,9 @@ RelDataType createMultisetType(
    * Creates a type that is the same as another type but with possibly
    * different nullability. The output type may be identical to the input
    * type. For type systems without a concept of nullability, the return value
-   * is always the same as the input.
+   * is always the same as the input.  This function never returns a nullable
```
asolimando, line 182:
```suggestion
   * is always the same as the input. This function never returns a nullable
```

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
```diff
@@ -191,6 +193,24 @@ RelDataType createTypeWithNullability(
       RelDataType type,
       boolean nullable);
 
+  /**
+   * Creates a type that is the same as another type but with possibly
+   * different nullability. The output type may be identical to the input
+   * type. For type systems without a concept of nullability, the return value
+   * is always the same as the input.  This differs from {@link
```
asolimando, line 200:
```suggestion
   * is always the same as the input. This differs from {@link
```

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
```diff
@@ -191,6 +193,24 @@ RelDataType createTypeWithNullability(
...
+   * #createTypeWithNullability(RelDataType, boolean)} in the handling of struct
+   * types.  This function returns a nullable struct.
```
asolimando, line 202:
```suggestion
   * types. This function returns a nullable struct.
```

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java
```diff
@@ -191,6 +193,24 @@ RelDataType createTypeWithNullability(
...
+   *
+   * @param type     input type
```
asolimando, line 204: Nit: I see that some javadoc here has this kind of formatting with extra spaces to align columns "horizontally", AFAICT this is not what we generally use in the codebase, so I'd rather not enforce this here

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
```diff
@@ -410,6 +410,31 @@ private RelDataType copyRecordType(
...
+            @Override public RelDataType get(int index) {
+              return type.getFieldList().get(index).getType();
+            }
+
+            @Override public int size() {
+              return type.getFieldCount();
+            }
+          },
+          type.getFieldNames(), nullable);
+    } else {
```
asolimando, line 432: Aren't there other non simple types like maps (`MultisetSqlType`)/arrays (`ArraySqlType`)? Is it enough in this form?

EDIT: I have just seen you have done that for the `SqlTypeFactoryImpl`, why here it is not needed?

asolimando (commented):
Looks mostly good to me @mihaibudiu, just confused about the fact that we are not considering anything other than Records in `RelDataTypeFactoryImpl`, if you could clarify that we are good for me.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
mihaibudiu: As far as I can tell the other types are handled correctly by the factory.
Only Record types are anomalous.
This class (RelDataTypeFactoryImpl) is an abstract base class out of which SqlTypeFactoryImpl is derived. In the end, the SqlTypeFactoryImpl method gets called.

asolimando: Thanks for the clarification Mihai, I have no more questions, hence I will approve the PR, the rest is minor and I will leave it to your judgement

asolimando (approved):

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java
mihaibudiu: I will document it anyway.

asolimando: I guess so, because implementers will have to add support for it

mihaibudiu: I will add this to history.md in the section on breaking changes:
```
*RelDataTypeFactory interface*.  The fix for [<a
href="https://issues.apache.org/jira/browse/CALCITE-6764">CALCITE-6764</a>]
introduces a new method
`RelDataTypeFactory#enforceTypeWithNullability` in the existing
`RelDataTypeFactory` interface.  The behavior of the new function is
similar to the existing API `createTypeWithNullability`; however, the
existing implementations of the `createTypeWithNullability` API cannot
create nullable record (`ROW`) types.  Nullable record types are
legitimate in several SQL dialects.
```

PR #4147, opened 2025-01-15
[CALCITE-6786] ANY/SOME operator yields multiple rows in correlated queries

racevedoo:


core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -696,16 +696,15 @@ private static RexNode rewriteIn(RexSubQuery e, Set<CorrelationId> variablesSet,
...
-          builder.sortLimit(0, 1,
-              ImmutableList.of(builder.desc(builder.field("cs"))));
         } else {
           builder.distinct();
         }
+        // sorts input with desc order since we are interested
+        // only in the case when one of the values is true.
+        // When true value is absent then we are interested
+        // only in false value.
+        builder.sortLimit(0, 1,
```
racevedoo, line 706: This includes the `order by cs limit 1` clause in the correlated query case. Without it, multiple rows are returned, as described in the issue.

mihaibudiu: It looks like this was written by @vvysotskyi about 6 years ago, it would be great if he could comment.

core/src/test/resources/sql/sub-query.iq
```diff
@@ -1871,13 +1871,15 @@ select sal from "scott".emp e
...
 
 !ok
-EnumerableCalc(expr#0..4=[{inputs}], expr#5=[RAND()], expr#6=[CAST($t5):INTEGER NOT NULL], expr#7=[2], expr#8=[MOD($t6, $t7)], expr#9=[3], expr#10=[=($t8, $t9)], expr#11=[OR($t10, $t4)], SAL=[$t1], $condition=[$t11])
-  EnumerableMergeJoin(condition=[=($2, $3)], joinType=[left])
-    EnumerableSort(sort0=[$2], dir0=[ASC])
-      EnumerableCalc(expr#0..7=[{inputs}], EMPNO=[$t0], SAL=[$t5], DEPTNO=[$t7])
-        EnumerableTableScan(table=[[scott, EMP]])
-    EnumerableCalc(expr#0..2=[{inputs}], expr#3=[false], DEPTNO=[$t0], $f1=[$t3])
-      EnumerableTableScan(table=[[scott, DEPT]])
+EnumerableCalc(expr#0..2=[{inputs}], expr#3=[RAND()], expr#4=[CAST($t3):INTEGER NOT NULL], expr#5=[2], expr#6=[MOD($t4, $t5)], expr#7=[3], expr#8=[=($t6, $t7)], expr#9=[OR($t8, $t2)], SAL=[$t0], $condition=[$t9])
```
mihaibudiu, line 1874: all these now contain correlated subqueries.
I wonder whether these tests expect uncorrelated results. This would indicate that either the decorrelator fails on the new plans, or perhaps it's run too early.

racevedoo: In my understanding, the plan change is basically that the `EnumerableMergeJoin` was replaced with a `EnumerableCorrelate`. Aren't they similar in this case?

Naturally, the sort + limit is also introduced.

suibianwanwank: I agree that the plan after this PR is correct. 
For decorrelate failures, [CALCITE-6652] reported this issue. I'm trying to submit a PR next month, and after that, the plan should be better.

mihaibudiu: I believe that correlated queries in general are much less efficient than uncorrelated ones.
Moreover, the Calcite decorrelator can only decorrelate a limited number of patterns.
That's why a decorrelated plan is preferred to a plan that has correlated subqueries.
However, correctness is more important than performance, so we should take this change if it fixes a correctness bug.
The question was whether we could have both correctness and performance, but that can be part of a separate issue.

mihaibudiu (approved):

caicancai (approved):

suibianwanwank (approved):

PR #4148, opened 2025-01-15
[CALCITE-6788] LoptOptimizeJoinRule should delegate costs to optimization planner

arkanovicz:
`LoptOptimizeJoinRule` uses costs comparisons to recursively decide whether to add a join at the top of the joins tree or to push it down.

When doing so, instead of directly calling `mq.getCumulativeCost(rel)`, it should rely on `call.getPlanner.getCost(rel, mq)`, which will be used thereafter to choose the best joins tree.

This way, it becomes possible to customize the costs computation during the heuristic joins ordering phase by just altering the `getCost()` method of the optimization planner.

Without this patch, it is only possible to alter joins costs at the final phase of the algorithm, which defeats the goal of customizing them, including during the to-top/push-down phase.


xiedeyantu:
This issue has been resolved in #4231 . We will refer to the PR link and close this PR, and thank the @arkanovicz!

PR #4149, opened 2025-01-16
[CALCITE-6780] AbstractSqlTester fails to build query for expression TRIM(string)

ILuffZhe:
This patch skips extracting trimmed character for TRIM function in CheckFramework. 
Hence, after abstractSqlTester.buildQuery2(), expression `trim(' aAa ')` will be `select trim(p0) from (values (' aAa ')) as t(p0)`.
And expression `trim(both ' ' from ' aAa ')` will be `select trim(both ' ' from p0) from (values (' aAa ')) as t(p0)`.


core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
```diff
@@ -948,8 +948,8 @@ ExInst<CalciteException> illegalArgumentForTableFunctionCall(String a0,
   @BaseMessage("Illegal arguments: The length of the keys array {0,number,#} is not equal to the length of the values array {1,number,#} in MAP_FROM_ARRAYS function")
   ExInst<CalciteException> illegalArgumentsInMapFromArraysFunc(int arg0, int arg1);
 
-  @BaseMessage("Trim error: trim character must be exactly 1 character")
```
mihaibudiu, line 951: this should read "the length of the string describing the trimmed character should be 1"

mihaibudiu (commented):
How is this related to the CheckerFramework?
I suspect that the JIRA title is wrong.
Maybe there is a problem in the AbstractSqlTester, but that's unrelated.

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
ILuffZhe: I've noticed that the former doc implies changing it to `Invalid argument...`.Since it is a strict rule in SQL standard, what about "Invalid argument ''{0}'' : the length of the string describing the trimmed character must be 1".

ILuffZhe:
> How is this related to the CheckerFramework? I suspect that the JIRA title is wrong. Maybe there is a problem in the AbstractSqlTester, but that's unrelated.

You are right this is not about CheckerFramework. I'm considering changing the summary to `AbstractSqlTester failed to build query for expression TRIM(string)`. What do you think?

mihaibudiu:
that is better, but I would say "fails" instead of "failed"

NobiGo (approved):

core/src/main/java/org/apache/calcite/runtime/CalciteResource.java
ILuffZhe: > this should read "the length of the string describing the trimmed character should be 1"

Done.

mihaibudiu:
I think we can merge this after squashing

NobiGo (approved):

PR #4150, opened 2025-01-17
[CALCITE-6789] A NumberFormatException was thrown when performing an aggregate query on a csv file.

hoperunChen:
Add a check for empty strings in the charToDecimalCast method, otherwise it will cause a NumberFormatException.


mihaibudiu:
I don't think this is the correct fix. This method is also invoked by casts from strings to numeric, and there it must throw.

hoperunChen:
If the value is null, this method will return null; why can’t it return null for an empty string as well? I believe the null check at the beginning was added precisely to prevent a NumberFormatException when creating a BigDecimal, wasn’t it?

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
```diff
@@ -427,7 +427,7 @@ static BigDecimal checkOverflow(BigDecimal value, int precision, int scale,
   /** Called from BuiltInMethod.CHAR_DECIMAL_CAST_ROUNDING_MODE */
   public static @Nullable Object charToDecimalCast(
       @Nullable String value, int precision, int scale, RoundingMode roundingMode) {
-    if (value == null) {
+    if (value == null || value.isEmpty()) {
```
ILuffZhe, line 430: Can we have any tests here?

ILuffZhe (commented):
Hi @hoperunChen , please make sure your commit info consistent with JIRA ticket's summary. See https://calcite.apache.org/develop/#contributing.

linq4j/src/main/java/org/apache/calcite/linq4j/tree/Primitive.java
hoperunChen: Of course, testing can be done. My test case is attached in the comments on the [Jira ticket](https://issues.apache.org/jira/browse/CALCITE-6789); you can use my example to try it out.
If it’s truly impossible to modify the method, could you provide me with a solution to my issue? The CSV file is not controlled by the program, so I cannot change the values of the titles.

mihaibudiu:
You can run the existing tests on your machine using `./gradlew build`, this will show that this fix breaks many other tests.

hoperunChen:
Hi everyone, @ILuffZhe @mihaibudiu 
I have updated the commit message based on the previous feedback and added relevant test cases to validate the changes.
Please take a moment to review it again.  Let me know if there are any questions or concerns.  Thanks!

NobiGo:
It seems like I made a mistake. I retested the same data situation on Pg and it will threw an exception.
The test SQL in Postgresql:
```
CREATE TABLE public.newtable (
	"number" varchar NULL
);

INSERT INTO public.newtable("number") VALUES('');

INSERT INTO public.newtable("number") VALUES('1');
```

Then execute SQL:

```
SELECT sum(cast("number" as integer))
FROM public.newtable;
```
will throw exception: SQL 错误 [22P02]: ERROR: invalid input syntax for integer: "".
So maybe @hoperunChen should just need to add the data type in CSV title according to the comments in Jira.

example/csv/src/test/resources/sales/test.csv
```diff
@@ -0,0 +1,3 @@
+id,name,sex,age
```
NobiGo, line 1: According to other CSV files, there should be a data type here. If it is the default String type, it is normal for Calcite to throw an exception.

hoperunChen:
hi @NobiGo I agree with your point, but this CSV file is not prepared by developers or technical staff; it is provided by ordinary users who do not understand technology. We cannot demand that they define the type of each column, as this would lead to a poor user experience.

NobiGo:
> hi @NobiGo I agree with your point, but this CSV file is not prepared by developers or technical staff; it is provided by ordinary users who do not understand technology. We cannot demand that they define the type of each column, as this would lead to a poor user experience.

Yes. I get your point. Can we explain the default data types in the document to help you with your worries? The current PR changes have already affected all semantics of other non-CSV SQL types, which is unacceptable.

hoperunChen:
Alright, then we can only process the file again after the user submits the document.

However, I don’t quite understand why the charToDecimalCast() method only checks for null but not for an empty string. In what scenario would forcibly converting an empty string and then throwing an error be considered normal business logic? This is just a personal question, and I’d like to seek your advice.

PR #4151, opened 2025-01-18
[CALCITE-6756] Preserving CAST of STRING operand in binary comparison…

NobiGo:
… for PostgreSQL

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/PostgresqlSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/PostgresqlSqlDialect.java
@@ -21,6 +21,9 @@ import org.apache.calcite.avatica.util.TimeUnitRange;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
 import org.apache.calcite.rel.type.RelDataTypeSystemImpl;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlAlienSystemTypeNameSpec;
 import org.apache.calcite.sql.SqlBasicCall;
 import org.apache.calcite.sql.SqlCall;
@@ -40,6 +43,7 @@ import org.apache.calcite.sql.fun.SqlFloorFunction;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.type.SqlTypeUtil;
 
 import com.google.common.collect.ImmutableList;
 
@@ -153,6 +157,17 @@ public class PostgresqlSqlDialect extends SqlDialect {
     default:
       return super.supportsFunction(operator, type, paramTypes);
     }
+  }
+
+  @Override public boolean supportsImplicitTypeCoercion(RexCall call) {
+    final RexNode operand0 = call.getOperands().get(0);
+    RelDataType callType = call.getType();
+    boolean supportImplicit = super.supportsImplicitTypeCoercion(call);
+    boolean isNumericType = supportImplicit && SqlTypeUtil.isNumeric(callType);
+    if (isNumericType) {
+      return false;
+    }
+    return supportImplicit && RexUtil.isLiteral(operand0, false);
   }
 
   @Override public boolean requiresAliasForFromItems() {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/PostgresqlSqlDialect.java
```diff
@@ -155,6 +159,17 @@ public PostgresqlSqlDialect(Context context) {
...
 
+  @Override public boolean supportsImplicitTypeCoercion(RexCall call) {
+    final RexNode operand0 = call.getOperands().get(0);
+    RelDataType callType = call.getType();
+    boolean supportImplicit = super.supportsImplicitTypeCoercion(call);
+    boolean isNumericType = supportImplicit && SqlTypeUtil.isNumeric(callType);
+    if (isNumericType) {
+      return false;
+    }
+    return supportImplicit && RexUtil.isLiteral(operand0, false);
```
mihaibudiu, line 170: are you sure only literals are a problem?

NobiGo: Yes. Literals don't need the cast, others need it. I add a unit test.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7960,6 +7960,30 @@ private void checkLiteral2(String expression, String expected) {
...
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6756">[CALCITE-6756]
+   * Preserving CAST of STRING operand in binary comparison for PostgreSQL</a>. */
+  @Test void testImplicitTypeCoercionPostgreSQL() {
+    final String query = "select \"employee_id\" "
+        + "from \"foodmart\".\"employee\" "
+        + "where 10 = cast(\"full_name\" as int) and "
+        + "  \"first_name\" > cast(10 as varchar) and "
+        + "\"birth_date\" = cast('1914-02-02' as date) or "
```
ILuffZhe, line 7971: Why not add a filter like `cast('10' as varchar)  = 1`.

NobiGo: This expression is illegal in PostgreSQL. But after the dialect is converted,  it can work well.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7971,15 +7971,17 @@ private void checkLiteral2(String expression, String expected) {
         + "\"birth_date\" = cast('1914-02-02' as date) or "
         + "\"hire_date\" = cast('1996-01-01 '||'00:00:00' as timestamp) or "
         + "\"hire_date\" = '1996-01-01 00:00:00' or "
-        + "cast(\"full_name\" as timestamp) = \"hire_date\"";
+        + "cast(\"full_name\" as timestamp) = \"hire_date\" or "
+        + "cast('10' as varchar) = 1";
```
ILuffZhe, line 7975: Now I get it. The result would be `cast('10' as varchar) = 1` without this changing, and It throws exception in PostgreSQL.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7960,6 +7960,32 @@ private void checkLiteral2(String expression, String expected) {
...
+  @Test void testImplicitTypeCoercionPostgreSQL() {
+    final String query = "select \"employee_id\" "
+        + "from \"foodmart\".\"employee\" "
+        + "where 10 = cast(\"full_name\" as int) and "
+        + "  \"first_name\" > cast(10 as varchar) and "
+        + "\"birth_date\" = cast('1914-02-02' as date) or "
+        + "\"hire_date\" = cast('1996-01-01 '||'00:00:00' as timestamp) or "
+        + "\"hire_date\" = '1996-01-01 00:00:00' or "
+        + "cast(\"full_name\" as timestamp) = \"hire_date\" or "
+        + "cast('10' as varchar) = 1";
```
ILuffZhe, line 7975: One more question: what happens for `cast(null as varchar) = 1` here.

NobiGo: It will become `null = 1` works well in PG. Because NULL is literal.

ILuffZhe (approved):

PR #4155, opened 2025-01-20
[CALCITE-6790] Write LIMIT for fetch operations in Vertica

illes:
correctly unparse Vertica [LIMIT](https://docs.vertica.com/24.2.x/en/sql-reference/statements/select/limit-clause/) ... [OFFSET](https://docs.vertica.com/24.2.x/en/sql-reference/statements/select/offset-clause/)

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/VerticaSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/VerticaSqlDialect.java
@@ -19,7 +19,11 @@ package org.apache.calcite.sql.dialect;
 import org.apache.calcite.avatica.util.Casing;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.SqlWriter;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/VerticaSqlDialect.java
```diff
@@ -53,4 +57,8 @@ public VerticaSqlDialect(Context context) {
       return super.supportsFunction(operator, type, paramTypes);
     }
   }
+
+  @Override public void unparseOffsetFetch(SqlWriter writer, @Nullable SqlNode offset, @Nullable SqlNode fetch) {
```
mihaibudiu, line 61: you need to add at least one test.

illes: @mihaibudiu , I'd like to, but couldn't find existing tests for thos or other `SqlDialect` classes or `unparse*()` methods, can you advise how to start?

mihaibudiu: RelToSqlConverterTest

mihaibudiu (approved):

caicancai (approved):

PR #4156, opened 2025-01-21
[CALCITE-2295] Correlated SubQuery with Project will generate error plan

NobiGo:


NobiGo:
This PR only includes unit tests.

caicancai:
> This PR only includes unit tests.

If there is only unit testing, it seems unnecessary to create a jira case. There may be other PRs in this jira case later?

NobiGo:
> > This PR only includes unit tests.
> 
> If there is only unit testing, it seems unnecessary to create a jira case. There may be other PRs in this jira case later?

This is a Jira created a long time ago, and we can close it after adding test cases

NobiGo:
Through the test cases, it can be seen that the issues in Jira have been resolved.

PR #4158, opened 2025-01-22
[CALCITE-6303] UNION with CTE(s) results in exception during query va…

NobiGo:
…lidation

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
@@ -4431,6 +4431,15 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
       default:
         break;
       }
+    } else if (query.getKind() == SqlKind.WITH) {
+     // The modality of WITH clause depends on its body
+     // For example:
+     // SQL: WITH STREAMTABLE AS (SELECT STREAM * FROM KAFKA.MOCKTABLE) SELECT * FROM STREAMTABLE
+     // The modality should be RELATION.
+     // SQL: WITH STREAMTABLE AS (SELECT STREAM * FROM KAFKA.MOCKTABLE)
+     //      SELECT STREAM * FROM STREAMTABLE
+     // The modality should be STREAM.
+      validateModality(((SqlWith) query).body);
     } else {
       assert query.isA(SqlKind.SET_QUERY);
       final SqlCall call = (SqlCall) query;
@@ -4453,6 +4462,8 @@ public class SqlValidatorImpl implements SqlValidatorWithHints {
           : SqlModality.RELATION;
     } else if (query.getKind() == SqlKind.VALUES) {
       return SqlModality.RELATION;
+    } else if (query.getKind() == SqlKind.WITH) {
+      return deduceModality(((SqlWith) query).body);
     } else {
       assert query.isA(SqlKind.SET_QUERY);
       final SqlCall call = (SqlCall) query;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4431,6 +4431,14 @@ private void validateModality(SqlNode query) {
       default:
         break;
       }
+    } else if (query.getKind() == SqlKind.WITH) {
+    // The modality of With clause depends on its body
```
mihaibudiu, line 4435: I think this should be indented deeper

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4431,6 +4431,14 @@ private void validateModality(SqlNode query) {
...
+    // For example:
+    // SQL: WITH STREMTABLE AS SELECT STREAM * FROM KAFKA.MOCKTABLE SELET * FROM STREMTABLE
```
mihaibudiu, line 4437: is this supposed to be STREAMTABLE?
SELECT is also mistyped

mihaibudiu (approved):

ILuffZhe (approved):
LGTM

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
NobiGo: SQL1 [With List is STREAM]:
`WITH STREAMTABLE AS (SELECT STREAM * FROM KAFKA.MOCKTABLE) SELECT * FROM STREAMTABLE`
The Plan is :
`"PLAN=EnumerableTableScan(table=[[KAFKA, MOCKTABLE]])\n\n"`
SQL2: [With Body is STREAM and With List is STREAM]:
`WITH STREAMTABLE AS (SELECT STREAM * FROM KAFKA.MOCKTABLE) SELECT STREAM * FROM STREAMTABLE`
The Plan is:
```
EnumerableInterpreter
     BindableTableScan(table=[[KAFKA, MOCKTABLE, (STREAM)]])
```
SQL3:[With Body is STREAM]
`WITH STREAMTABLE AS (SELECT * FROM KAFKA.MOCKTABLE) SELECT STREAM * FROM STREAMTABLE`
The Plan is:
```
EnumerableInterpreter
     BindableTableScan(table=[[KAFKA, MOCKTABLE, (STREAM)]])
```
These SQL  are tested in KafkaAdapterTest. According to this, we only check the Body is enough. WDYT?
Maybe I can add these unit tests in KafkaAdapterTest.

NobiGo: I assumed that the logical plan of Kafka here was correct.

NobiGo: Updated. Please ignore the reply.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
NobiGo: Updated.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -4431,6 +4431,15 @@ private void validateModality(SqlNode query) {
       default:
         break;
       }
+    } else if (query.getKind() == SqlKind.WITH) {
+     // The modality of With clause depends on its body
```
caicancai, line 4435: just one comment, With clause -> WITH clause.
This may better distinguish the WITH keyword

caicancai (approved):

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
NobiGo: Update.

PR #4160, opened 2025-01-23
[CALCITE-6503] JdbcAdapter cannot push down NOT IN sub-queries

NobiGo:


core/src/main/java/org/apache/calcite/sql/SqlDialect.java
```diff
@@ -754,6 +754,7 @@ public boolean supportsGroupByLiteral() {
 
   public boolean supportsAggregateFunction(SqlKind kind) {
     switch (kind) {
+    case LITERAL_AGG:
```
mihaibudiu, line 757: this change does not seem related to the jdbc adapter
are there other aggregate functions that have to be considered for this list?
maybe this question is not related really to the bug you are solving

NobiGo: This determines which aggregation functions can be pushed down, It is used in JdbcRules#canImplement. If this aggregation function can be pushed down, SqlImplementor#toSql will make `litetal_agg(literal)` to literal.

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -234,6 +234,32 @@ class JdbcAdapterTest {
...
+        .query("select * from dept where deptno not in (select deptno from emp)")
+        .explainContains("PLAN=JdbcToEnumerableConverter\n"
+            + "  JdbcProject(DEPTNO=[$0], DNAME=[$1], LOC=[$2])\n"
+            + "    JdbcFilter(condition=[OR(=($3, 0), AND(IS NULL($6), >=($4, $3)))])\n"
+            + "      JdbcJoin(condition=[=($0, $5)], joinType=[left])\n"
+            + "        JdbcJoin(condition=[true], joinType=[inner])\n"
+            + "          JdbcTableScan(table=[[SCOTT, DEPT]])\n"
+            + "          JdbcAggregate(group=[{}], c=[COUNT()], ck=[COUNT($7)])\n"
+            + "            JdbcTableScan(table=[[SCOTT, EMP]])\n"
+            + "        JdbcAggregate(group=[{7}], i=[LITERAL_AGG(true)])\n"
```
NobiGo, line 251: @mihaibudiu Here `LITERAL_AGG(true)` will become `true`.

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
mihaibudiu: True, but this change will have other effects besides enabling push-down.
But I guess this is fine. You are solving a more general problem than the issue states, but you have discovered through this particular issue.

ILuffZhe, line 757: Do we need to add a test in `RelToSqlConverterTest`?

NobiGo: The unit test testPushDownNotIn includes the plan and HSQLDB SQL(Rel To Sql). The changes to this PR have already been covered.

PR #4163, opened 2025-01-23
[CALCITE-4561] Wrong results for plan with EnumerableHashJoin (semi) …

NobiGo:
…on nullable colunms


NobiGo:
CALCITE-5732 has fixed this issue. I added the link in Jira; this PR only includes the unit test.

mihaibudiu (approved):
I haven't checked the result, but I assume that the JIRA ticked is correct.
I guess this is just so we can close the issue.

rubenada (approved):

PR #4167, opened 2025-01-24
[CALCITE-6796] Convert Type from BINARY to VARBINARY in PrestoDialect

xiaochen-zhou:
Since BINARY  type is not supported in Presto

According to the documentation:

https://prestodb.io/docs/current/language/types.html#varbinary

Using CAST(xxxx AS BINARY) will result in an error:

```
SELECT cast(task_id as binary)
from applydata_bigdata.xflink_task_info
where reader_info like '%mongodbreader%' or writer_info like '%mongodbwriter%';
```
image

It should be converted to the VARBINARY type instead.

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -9147,6 +9147,17 @@ class RelToSqlConverterTest {
   }
 
   /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6796">[CALCITE-6796]
+   * Convert Type from BINARY to VARBINARY in PrestoDialect</a>. */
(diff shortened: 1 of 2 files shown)


caicancai (commented):
overall LGTM

caicancai (approved):

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -9160,6 +9171,8 @@ private void checkLiteral2(String expression, String expected) {
         .withPresto().ok(expected);
   }
 
+
```
caicancai, line 9174: Extra blank lines

xiaochen-zhou: > Extra blank lines

Done.

PR #4171, opened 2025-01-26
[CALCITE-6801] Linter should disallow tags such as 'Chore' in commit messages

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6801


ILuffZhe (approved):
LGTM.

PR #4173, opened 2025-01-28
[CALCITE-6806] Add CI action to ensure site remains buildable after changes

zabetak:


.github/workflows/site.yml
```diff
@@ -0,0 +1,39 @@
...
+  build-website:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Build site
+        working-directory: site
+        run: |
+          docker compose run -e JEKYLL_UID=$(id -u) -e JEKYLL_GID=$(id -g) build-site
```
F21, line 35: In the latest iteration of the site publishing workflows, we no longer need to set `JEKYLL_UID` and `JEKYLL_GID`, so I those 2 environment variables can be removed.

See: https://github.com/apache/calcite/blob/main/.github/workflows/publish-non-release-website-updates.yml#L59

F21 (changes requested):

.github/workflows/site.yml
zabetak: Thanks for the review Francis! Fixed in https://github.com/apache/calcite/pull/4173/commits/6b04e32bff588108184e5475987c394e5924fe0c

F21 (approved):

PR #4178, opened 2025-02-04
[CALCITE-6814] Support weekofyear functions for Hive

xuzifu666:
Currently calcite not support weekofyear functions for Hive Library, need to support it.
jira: https://issues.apache.org/jira/browse/CALCITE-6814


xuzifu666:
Close it first.

PR #4179, opened 2025-02-04
[CALCITE-6813] UNNEST infers incorrect nullability for the result ...

mihaibudiu:
when applied to an array that contains nullable ROW values

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
@@ -1057,6 +1057,19 @@ public class SqlValidatorTest extends SqlValidatorTestCase {
             + "Was expecting 3 arguments");
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6813">
+   * [CALCITE-6813] UNNEST infers incorrect nullability for the result when applied to
+   * an array that contains nullable ROW values</a>. */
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -1057,6 +1057,19 @@ void testDyadicCollateOperator() {
...
 
+  /** Test case for <a href="https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6813">
+   * [CALCITE-6813] UNNEST infers incorrect nullability for the result when applied to
+   * an array that contains nullable ROW values</a>. */
+  @Test void testUnnestRow() {
+    sql("with orders(data) as\n"
+        + "  (values (ARRAY[ROW(1, 'Alice'), ROW(2, NULL), ROW(NULL, 'Bob'), NULL]))\n"
+        + "select e.EXPR$0\n"
+        + "from orders, UNNEST(orders.data) as e")
+        .type(actualType -> {
```
ILuffZhe, line 1068: What if all ROW values are not nullable?

mihaibudiu: I assume you mean: "none of the values in any ROW is null."
That case was covered before. ROWs with null values were not handled correctly.

ILuffZhe (approved):

PR #4181, opened 2025-02-05
[CALCITE-6652] RelDecorrelator can't decorrelate query with limit 1

suibianwanwank:
This PR handles decorrelate since the query contains limit 1. There are two main types of conversions:
1. If sorted with no OFFSET and FETCH = 1, and only one collation field, rewrite the Sort as Aggregate using MIN/MAX function.
```
Example:
  Sort(sort0=[$0], dir0=[ASC], fetch=[1])
    input
Rewrite to:
  Aggregate(group=(corVar), agg=[min($0))
```
2. For the more general case,rewrite the sort to be
```
Aggregate(group=(corVar.. , field..))
  project(first_value(field) over (partition by corVar order by (sort collation)))
     input
```


caicancai:
If no one reviews it, I will start reviewing it this weekend, it may take some time.

suibianwanwank:
> If no one reviews it, I will start reviewing it this weekend, it may take some time.

Thank you. If you have any questions or need further clarification, feel free to ask.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -795,6 +801,122 @@ private static void shiftMapping(Map<Integer, Integer> mapping, int startIndex,
     return null;
   }
 
+  private @Nullable Frame decorrelateFetchOneSort(Sort sort, final Frame frame) {
+    final Map<Integer, Integer> mapOldToNewOutputs = new HashMap<>();
+    final NavigableMap<CorDef, Integer> corDefOutputs = new TreeMap<>();
+    //
+    // Rewrite logic:
+    //
+    // If sorted with no OFFSET and FETCH = 1, and only one collation field,
```
mihaibudiu, line 810: the check for fetch == 1 is not here (so you can mention that it's enforced by the caller).
Moreover, for symmetry, should this comment be within the if branch?

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -795,6 +801,122 @@ private static void shiftMapping(Map<Integer, Integer> mapping, int startIndex,
...
+      case DESCENDING:
+      case STRICTLY_DESCENDING:
+        aggFunction = SqlStdOperatorTable.MAX;
+        break;
+      default:
+        return null;
+      }
+
+      final int newIdx = requireNonNull(frame.oldToNewOutputs.get(collation.getFieldIndex()));
+      RelBuilder.AggCall aggCall = relBuilder.push(frame.r)
```
mihaibudiu, line 845: I am wondering: sorting has collation, but min/max do not.
For example, sorting can specify what happens to nulls.

mihaibudiu (commented):
I will try to review some more later, after I understand the relationship between order by and min/max.
Perhaps it's enough to check that the orderby does null handling in the expected way.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
rubenada: This was my exact same thought after a quick look at the PR... the proposed conversion is 100% equivalent in all cases, also when nulls are involved? What about the original Sort is nulls-first and the relevant field has null values?

suibianwanwank: Yes, as mentioned in JIRA, I missed this. This conversion is only valid if NULL LAST.

suibianwanwank:
Thanks for your feedback! I've updated the code and added null value test, thanks to review when you have time.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
rubenada: Could we have some unit tests with nulls-first, to see the expected result being the conversion *not* being applied?

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -795,6 +801,146 @@ private static void shiftMapping(Map<Integer, Integer> mapping, int startIndex,
     return null;
   }
 
+  private @Nullable Frame decorrelateFetchOneSort(Sort sort, final Frame frame) {
```
rubenada, line 804: Maybe I'm a bit too paranoid, but just to be on the safe side, I'd propose the new methods (`decorrelateFetchOneSort` and `decorrelateSortAsAggregate`) to be `protected` instead of `private`; the reason is that if, for whatever reason, a downstream project wants to "switch off" this conversion, they could simply extend the RelDecorrletor with their own child class and override these methods to simply return `null`, right?

suibianwanwank: I agree that we need a safe way to avoid errors. Usually the user uses the decorrelate by calling decorrelateQuery. So I think is it possible to add something like SqlToRelConverter#CONFIG (which can be combined with [CALCITE-6674]) to choose whether to enable it or not. WDYT

rubenada: It may be a cleaner solution, but IMO maybe a bit overkill for this particular purpose. I think protected methods would suffice to provide a way to "fallback". Let's see if there are other opinions on this subject.

suibianwanwank: Got it, I've made the changes. Thanks for the suggestion.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
mihaibudiu: I would think that MAX works for NULL LAST only, and MIN for NULL FIRST.
Am I wrong?

suibianwanwank: For MIN/MAX, null values are ignored in the databases I know (I didn't find a corresponding description in the sql standard). So I think both asc and desc should be null last, i.e.return null only if all values are null.

mihaibudiu (approved):
I am assuming that the quidem test results have been checked with another database as well.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -8577,6 +8577,111 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
...
+        .withLateDecorrelate(true)
+        .check();
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6652">[CALCITE-6652]
+   * RelDecorrelator can't decorrelate query with limit 1</a>.
+   */
+  @Test void testDecorrelateFilterWithFetchOneDesc() {
+    final String query = "SELECT name FROM dept "
```
rubenada, line 8625: Ins't this test exactly the same as the previous one?

suibianwanwank: Thanks for catching that! I've fixed it.

rubenada (approved):

PR #4182, opened 2025-02-05
[CALCITE-6816] Support DAY_OF_WEEK_IN_MONTH function format

xuzifu666:
Currently DAY_OF_WEEK_IN_MONTH format not support, need to support it.
JIRA: https://issues.apache.org/jira/browse/CALCITE-6816


xuzifu666:
> Maybe you can add a corresponding test in SqlOperatorTest

Thanks for suggestion, but currently there is no database support this kind of format, had a discuss in JIRA:https://issues.apache.org/jira/browse/CALCITE-6816, if OK would continue. @caicancai

PR #4183, opened 2025-02-06
[CALCITE-6813] UNNEST infers incorrect nullability for the result whe…

mihaibudiu:
…n applied to an array that contains nullable ROW values

This issue has already been fixed by me in https://github.com/apache/calcite/pull/4179, but the fix was incorrect.
This PR undoes some of the incorrect changes, and adds more tests.


caicancai (approved):
LGTM

PR #4186, opened 2025-02-07
[CALCITE-6819] MSSQL doesn't support TRUE/FALSE keywords in its Join predicate

sreeharshar84:
In MSSQL dialect use 1=1 instead of TRUE


mihaibudiu:
the style checker didn't like your use of reserved html characters in javadoc

suibianwanwank:
[CALCITE-6482] solves a similar issue, maybe a generic logic could be used?

suibianwanwank:
Also this problem occurs in where clause.

sreeharshar84:
Withdrawing because there are more issues to be dealt with, like HAVING TRUE clause.

PR #4187, opened 2025-02-08
[CALCITE-6820] Support Trino dialect SQL

xuzifu666:
Currently calcite not support dialect sql for trino, need to support it.
as descriptions: https://issues.apache.org/jira/browse/CALCITE-6820

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/TrinoSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/TrinoSqlDialect.java
@@ -18,32 +18,12 @@ package org.apache.calcite.sql.dialect;
 
 import org.apache.calcite.avatica.util.Casing;
 import org.apache.calcite.config.NullCollation;
-import org.apache.calcite.rel.type.RelDataType;
-import org.apache.calcite.rel.type.RelDataTypeSystem;
-import org.apache.calcite.rex.RexCall;
-import org.apache.calcite.rex.RexNode;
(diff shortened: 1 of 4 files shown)


ILuffZhe:
Your commit info should be consistent with JIRA ticket's summary.

xuzifu666:
> Your commit info should be consistent with JIRA ticket's summary.

Done，after discussing in jira，thanks~

caicancai (changes requested):
Discuss in Jira

core/src/main/java/org/apache/calcite/sql/dialect/TrinoSqlDialect.java
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.calcite.sql.dialect;
+
+import org.apache.calcite.avatica.util.Casing;
+import org.apache.calcite.config.NullCollation;
+import org.apache.calcite.sql.SqlDialect;
+
+/**
+ * A <code>SqlDialect</code> implementation for the Trino database.
+ */
+public class TrinoSqlDialect extends PrestoSqlDialect {
```
ILuffZhe, line 26: According to the changes here, it seems current PrestoSqlDialect is satisfied for querying Trino.

xuzifu666: Yes, like discussion in jira, it would close it later,Thanks~ @ILuffZhe

PR #4192, opened 2025-02-12
[CALCITE-6819] MSSQL doesn't support TRUE/FALSE keywords in its Join predicate

sreeharshar84:
In MSSQL dialect use 1=1 instead of TRUE

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/MssqlSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/MssqlSqlDialect.java
@@ -45,6 +45,8 @@ import org.apache.calcite.sql.type.SqlTypeName;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
+import static org.apache.calcite.util.RelToSqlConverterUtil.unparseBoolLiteralToCondition;
+
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -187,6 +189,15 @@ public class MssqlSqlDialect extends SqlDialect {
         super.unparseCall(writer, call, leftPrec, rightPrec);
       }
     }
+  }
+
+  @Override public void unparseBoolLiteral(SqlWriter writer,
+      SqlLiteral literal, int leftPrec, int rightPrec) {
+    Boolean value = (Boolean) literal.getValue();
+    if (value == null) {
+      return;
+    }
+    unparseBoolLiteralToCondition(writer, value);
   }
 
   @Override public boolean supportsCharSet() {
--- a/core/src/main/java/org/apache/calcite/sql/dialect/OracleSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/OracleSqlDialect.java
@@ -45,6 +45,8 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.util.List;
 
+import static org.apache.calcite.util.RelToSqlConverterUtil.unparseBoolLiteralToCondition;
+
 /**
  * A <code>SqlDialect</code> implementation for the Oracle database.
  */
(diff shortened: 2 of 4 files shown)


suibianwanwank (approved):
LGTM.

core/src/main/java/org/apache/calcite/sql/dialect/MssqlSqlDialect.java
```diff
@@ -189,6 +189,19 @@ public MssqlSqlDialect(Context context) {
     }
   }
 
+  @Override public void unparseBoolLiteral(SqlWriter writer,
+      SqlLiteral literal, int leftPrec, int rightPrec) {
+    Boolean value = (Boolean) literal.getValue();
+    if (value == null) {
+      return;
+    }
+    final SqlWriter.Frame frame = writer.startList("(", ")");
```
NobiGo, line 198: Is the code here the same as that in Oracle Dialect? If so, it can be extracted using a static method called RelToSqlConverterUtil.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -8225,6 +8225,27 @@ private void checkLiteral2(String expression, String expected) {
...
+  @Test void testJoinBoolLiteralMSSQL() {
+    final String queryTrue = "SELECT \"hire_date\", \"department_description\" FROM \"employee\" "
+        + "LEFT JOIN \"department\" ON TRUE";
+    final String mssqlExpected1 = "SELECT [employee].[hire_date],"
+        + " [department].[department_description]\nFROM [foodmart].[employee]\nLEFT JOIN"
+        + " [foodmart].[department] ON (1 = 1)";
+    sql(queryTrue)
+        .dialect(MssqlSqlDialect.DEFAULT).ok(mssqlExpected1);
+
+    final String queryFalse = "SELECT \"hire_date\", \"department_description\" FROM \"employee\" "
```
NobiGo, line 8240: 1) As discussed in Jira, this transformation not only affects the Join Predicate but also the SELECT, WHERE, and HAVING. Please cover this in the test cases.
2) Change the summary of Jira. Maybe we can remove the "Join predicate". This PR has affected other places.

core/src/main/java/org/apache/calcite/sql/dialect/MssqlSqlDialect.java
sreeharshar84: Done.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
sreeharshar84: I discussed this issue in my comment in the Jira(https://issues.apache.org/jira/browse/CALCITE-6819?focusedCommentId=17926427&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17926427). 

IMHO, fixing the issue for SELECT TRUE is a bit more complicated (CALCITE-6829). And the problem doesn't exist for HAVING AND WHERE. 

This is why I have restricted the issue to the Join related fix.

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
```diff
@@ -196,4 +197,12 @@ public static SqlSpecialOperator specialOperatorByName(String opName) {
       }
     };
   }
+
+  public static void writeAlwaysCondition(SqlWriter writer, boolean value) {
```
NobiGo, line 201: This method requires some annotations to clarify that some databases require converting True to '1=1' and false to '1=0'.

core/src/main/java/org/apache/calcite/sql/dialect/OracleSqlDialect.java
```diff
@@ -139,11 +141,7 @@ public OracleSqlDialect(Context context) {
       return;
     }
     // low version oracle not support bool literal
-    final SqlWriter.Frame frame = writer.startList("(", ")");
-    writer.literal("1");
-    writer.sep(SqlStdOperatorTable.EQUALS.getName());
-    writer.literal(value ? "1" : "0");
-    writer.endList(frame);
+    writeAlwaysCondition(writer, value);
```
NobiGo, line 144: unparseBoolLiteralToCondition?

NobiGo (approved):
Please squash the commits. If there are no more review comments, I will merge this.

sreeharshar84:
> Please squash the commits. If there are no more review comments, I will merge this.

Done.

PR #4194, opened 2025-02-13
[CALCITE-6831] Add ARRAY_SLICE function（enabled in Hive Library）

xuzifu666:
As descrption in JIRA: https://issues.apache.org/jira/browse/CALCITE-6831

### DIFF
--- a/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
+++ b/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
@@ -6524,6 +6524,15 @@ public class SqlFunctions {
     return list;
   }
 
+  /** SQL {@code ARRAY_SLICE(array, start, length)} function. */
+  public static List arraySlice(List list, int start, int length) {
+    // return empty list if start/length are out of range of the array
+    if (start + length > list.size()) {
+      return Collections.emptyList();
+    }
+    return list.subList(start, start + length);
+  }
+
   /** SQL {@code ARRAY_TO_STRING(array, delimiter)} function. */
   public static String arrayToString(List list, String delimiter) {
     return arrayToString(list, delimiter, null);
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -8204,6 +8204,49 @@ public class SqlOperatorTest {
         "DECIMAL(19, 0)");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6831">[CALCITE-6831]
+   * Add ARRAR_SLICE function (enabled in Hive library)</a>. */
+  @Test void testArraySlice() {
+    SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.ARRAY_SLICE);
+    final List<SqlLibrary> libraries =
+        ImmutableList.of(SqlLibrary.HIVE);
+    final Consumer<SqlOperatorFixture> consumer = f -> {
+      f.checkString("array_slice(array[1,2,3], 1, 2)",
+          "[2, 3]",
+          "INTEGER NOT NULL ARRAY NOT NULL");
+      f.checkString("array_slice(array[1,null,3], 1, 2)",
+          "[null, 3]",
+          "INTEGER ARRAY NOT NULL");
+      f.checkString("array_slice(array[1,2,3], 1, 10)",
+          "[]",
+          "INTEGER NOT NULL ARRAY NOT NULL");
+      f.checkString("array_slice(array['a','b','c','d'], 1, 3)",
+          "[b, c, d]",
+          "CHAR(1) NOT NULL ARRAY NOT NULL");
+      f.checkString("array_slice(array[null,null,null], 1, 2)",
+          "[null, null]",
+          "NULL ARRAY NOT NULL");
+      f.checkString("array_slice(array[1,2.2,3], 1, 2)",
+          "[2.2, 3.0]",
+          "DECIMAL(11, 1) NOT NULL ARRAY NOT NULL");
+      f.checkString("array_slice(array[1,cast(2 as double),3], 1, 2)",
+          "[2.0, 3.0]",
+          "DOUBLE NOT NULL ARRAY NOT NULL");
+      f.checkString("array_slice(array[1,2.2,3,null], 1, 3)",
+          "[2.2, 3.0, null]",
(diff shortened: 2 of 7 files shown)


core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -6524,6 +6524,15 @@ public static List reverse(List list) {
     return list;
   }
 
+  /** Support the ARRAY_SLICE function. */
+  public static List arraySlice(List list, int start, int length) {
+    // return empty list if start/length are out of range of the array
```
ILuffZhe, line 6529: How about "... if start + length exceeds the boundary of the array"?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8204,6 +8204,33 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6831">[CALCITE-6831]
+   * Add ARRAR_SLICE function (enabled in Hive library)</a>. */
+  @Test void testArraySlice() {
+    SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.ARRAY_SLICE);
+    final List<SqlLibrary> libraries =
+        ImmutableList.of(SqlLibrary.HIVE);
+    final Consumer<SqlOperatorFixture> consumer = f -> {
+      f.checkString("array_slice(array[1,2,3], 1, 2)",
```
ILuffZhe, line 8215: What happens if start is a negative number?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -6524,6 +6524,15 @@ public static List reverse(List list) {
     return list;
   }
 
+  /** Support the ARRAY_SLICE function. */
```
NobiGo, line 6527: Be consistent with the descriptions of other functions.
`/** SQL {@ xxxxxx }. */`

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8204,6 +8204,33 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
+      f.checkString("array_slice(array[1,null,3], 1, 2)",
+          "[null, 3]",
+          "INTEGER ARRAY NOT NULL");
+      f.checkString("array_slice(array[1,2,3], 1, 10)",
+          "[]",
+          "INTEGER NOT NULL ARRAY NOT NULL");
+      f.checkString("array_slice(array['a','b','c','d'], 1, 3)",
+          "[b, c, d]",
+          "CHAR(1) NOT NULL ARRAY NOT NULL");
+      f.checkString("array_slice(array[null,null,null], 1, 2)",
```
NobiGo, line 8227: What happens if start is NULL or length is NULL?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
xuzifu666: Would return empty array，related to ： https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFArraySlice.java

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
xuzifu666: Would error out and add a case with negative number.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
xuzifu666: Done

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
xuzifu666: Would return NULL when start or length is NULL， had add cases about it.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8204,6 +8204,38 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
...
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6831">[CALCITE-6831]
+   * Add ARRAR_SLICE function (enabled in Hive library)</a>. */
+  @Test void testArraySlice() {
+    SqlOperatorFixture f0 = fixture().setFor(SqlLibraryOperators.ARRAY_SLICE);
+    final List<SqlLibrary> libraries =
+        ImmutableList.of(SqlLibrary.HIVE);
+    final Consumer<SqlOperatorFixture> consumer = f -> {
+      f.checkString("array_slice(array[1,2,3], 1, 2)",
```
caicancai, line 8215: Need to add more boundary tests

caicancai: For example, other types of arrays, whether Hive runs mixed types of arrays

xuzifu666: OK, had added more test about mixed types.

NobiGo (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
caicancai: @xuzifu666 Can you add double type + int type? In Spark, double + int mixed types will undergo implicit conversion. I am not sure whether this will happen in Hive.

xuzifu666: I had add a case mixed with int and double and it would had a implicit conversion in Hive side:
      f.checkString("array_slice(array[1,2.2,3], 1, 2)",
          "[2.2, 3.0]",
          "DECIMAL(11, 1) NOT NULL ARRAY NOT NULL");

xuzifu666: Thanks @caicancai  for the remind, had added case about that.

caicancai (approved):
LGTM

ILuffZhe (approved):
LTGM! We can merge this after squashing commits.

NobiGo (approved):

PR #4196, opened 2025-02-14
[CALCITE-6832] Redundant fields/nullable entries in STRUCT serialization to JSON

zabetak:
The JSON serialization of a RelDatatypeField of type STRUCT contains redundant fields and nullable entries along with unnecessary nesting.

The redundancy leads to wasted space and convoluted representation that is hard to follow especially when there are nested STRUCTS.


ILuffZhe (approved):
LGTM.

PR #4197, opened 2025-02-15
[CALCITE-6838] RelToSqlConverter should generate double parentheses w…

NobiGo:
…hen the input to Unnest is a query statement

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlUnnestOperator.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlUnnestOperator.java
@@ -121,14 +121,7 @@ public class SqlUnnestOperator extends SqlFunctionalOperator {
 
   @Override public void unparse(SqlWriter writer, SqlCall call, int leftPrec,
       int rightPrec) {
-    if (call.operandCount() == 1
-        && call.getOperandList().get(0).getKind() == SqlKind.SELECT) {
-      // avoid double ( ) on unnesting a sub-query
-      writer.keyword(getName());
-      call.operand(0).unparse(writer, 0, 0);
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/SqlUnnestOperator.java
```diff
@@ -121,14 +121,7 @@ private static boolean allowAliasUnnestItems(SqlOperatorBinding operatorBinding)
 
   @Override public void unparse(SqlWriter writer, SqlCall call, int leftPrec,
       int rightPrec) {
-    if (call.operandCount() == 1
```
NobiGo, line 124: I tried other databases that support Unnest, and they all require two parentheses.

asolimando: Thanks for researching and fixing this @NobiGo, +1 if Calcite is the only exception, unless there is a good reason to, we generally try to comply with all major DBs.

May I only suggest to put this information in the Jira ticket description (the list of databases you tried), it might come handy later and there is less chance people will dig this information up from the PR.

NobiGo: Thanks for the review @asolimando . I have added some conclusions from the research to Jira. I hope it is helpful.

asolimando (approved):
LGTM, after reviewing the extra information in Jira I am convinced that we should align with the other DBs supporting UNNEST, as they all agree on using double parentheses, thanks @NobiGo!

ILuffZhe (approved):
LGTM!

PR #4198, opened 2025-02-16
[CALCITE-6839] The SUM function sometimes throws overflow exceptions …

NobiGo:
…due to incorrect return types

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/CoreQuidemTest.java
+++ b/core/src/test/java/org/apache/calcite/test/CoreQuidemTest.java
@@ -70,6 +70,16 @@ class CoreQuidemTest extends QuidemTest {
               .with(CalciteConnectionProperty.FUN, SqlLibrary.CALCITE.fun)
               .with(CalciteAssert.Config.SCOTT)
               .connect();
+        case "scott-spark":
+          discard(CustomTypeSystems.SPARK_TYPE_SYSTEM);
+          return CalciteAssert.that()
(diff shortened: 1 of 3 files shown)


NobiGo:
This return type does not have a unified standard. Here, the return type is referenced from [Spark's return type](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Sum.scala#L75), aiming to accommodate as many scenarios as possible.

core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystemImpl.java
```diff
@@ -357,6 +358,27 @@ && getDefaultPrecision(typeName) != RelDataType.PRECISION_NOT_SPECIFIED) {
                   argumentType.isNullable());
         }
       }
+      // For DECIMAL using default Precision and Scale,
```
mihaibudiu, line 361: as discussed in the JIRA, the right approach is to actually create a new type system that overrides this method rather than change RelDataTypeSystemImpl.

mihaibudiu:
This actually looks great.
There should be a way we can prepackage these type systems and have them available for everyone to use.

mihaibudiu (approved):

NobiGo:
> This actually looks great. There should be a way we can prepackage these type systems and have them available for everyone to use.

Create CALCITE-6849 for this.

PR #4201, opened 2025-02-17
[CALCITE-6834] In query that applies COALESCE to nullable SUM, Enumer…

NobiGo:
…ableProjectToCalcRule throws AssertionError

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/ProjectAggregateMergeRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/ProjectAggregateMergeRule.java
@@ -118,7 +118,7 @@ public class ProjectAggregateMergeRule
             final RexLiteral literal = (RexLiteral) operands.get(2);
             if (Objects.equals(literal.getValueAs(BigDecimal.class), BigDecimal.ZERO)) {
               int j = findSum0(cluster.getTypeFactory(), aggCall, aggCallList);
-              return cluster.getRexBuilder().makeInputRef(call.type, j);
+              return cluster.getRexBuilder().makeInputRef(aggCallList.get(j).getType(), j);
             }
           }
           break;
@@ -157,6 +157,7 @@ public class ProjectAggregateMergeRule
     builder.project(
         RexPermuteInputsShuttle.of(mapping).visitList(projects2),
             project.getRowType().getFieldNames());
+    builder.convert(project.getRowType(), true);
     call.transformTo(builder.build());
   }
 
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -6991,6 +6991,17 @@ class RelOptRulesTest extends RelOptTestBase {
         .check();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6834">[CALCITE-6834]
+   * In query that applies COALESCE to nullable SUM, EnumerableProjectToCalcRule
+   * throws AssertionError</a>. */
+  @Test void testProjectAggregateMergeSum01() {
+    final String sql = "select coalesce(sum(cast(mgr as tinyint)), 0) as ss0\n"
+        + "from sales.emp";
+    sql(sql).withRule(CoreRules.PROJECT_AGGREGATE_MERGE)
+        .check();
+  }
+
   /** As {@link #testProjectAggregateMergeSum0()} but there is another use of
    * {@code SUM} that cannot be converted to {@code SUM0}. */
   @Test void testProjectAggregateMergeSum0AndSum() {
--- a/core/src/test/resources/sql/agg.iq
+++ b/core/src/test/resources/sql/agg.iq
@@ -3849,4 +3849,52 @@ EnumerableAggregate(group=[{}], EXPR$0=[SUM($0)])
     EnumerableTableScan(table=[[scott, EMP]])
 !plan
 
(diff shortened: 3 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/rules/ProjectAggregateMergeRule.java
```diff
@@ -154,8 +156,22 @@ && kindCount(project.getProjects(), SqlKind.CASE) == 0) {
...
+        return builder.getRexBuilder().ensureType(
+            inputRef.getType(),
+            builder.fields().get(inputRef.getIndex()),
+            true);
+      }
+    };
+    for (RexNode rexInputRef : rexInputRefs) {
+      selectList.add(rexInputRef.accept(rexShuttle));
+    }
+    builder.project(selectList.build());
```
julianhyde, line 174: I think you can achieve the same effect with one line of code: use `RelBuilder.cast` to convert the output row type of `aggregate` to what `project` expects it to be. If `cast` adds a `CAST` call then `project` will flatten the expression on top of it.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -6925,6 +6925,13 @@ private HepProgram getTransitiveProgram() {
         .check();
   }
 
+  @Test void testProjectAggregateMergeSum01() {
```
julianhyde, line 6928: Add a hyperlink to jira case.  It will help explain what problem the test case is solving.

core/src/test/resources/sql/agg.iq
```diff
@@ -3815,4 +3815,51 @@ select distinct sum(deptno + '1') as deptsum from dept order by 1;
 
 !ok
 
+# [CALCITE-2192] In query that applies COALESCE to nullable SUM, EnumerableProjectToCalcRule throws AssertionError
```
julianhyde, line 3818: not CALCITE-2192

core/src/main/java/org/apache/calcite/rel/rules/ProjectAggregateMergeRule.java
NobiGo: What you are referring to here should be `RelBuilder.convert`, which also requires us to build a new RowType based on Mapping, so I have not made any changes here.

core/src/main/java/org/apache/calcite/rel/rules/ProjectAggregateMergeRule.java
```diff
@@ -154,9 +156,22 @@ && kindCount(project.getProjects(), SqlKind.CASE) == 0) {
     builder.push(aggregate.getInput());
     builder.aggregate(
         builder.groupKey(aggregate.getGroupSet(), aggregate.groupSets), aggCallList);
-    builder.project(
-        RexPermuteInputsShuttle.of(mapping).visitList(projects2),
-            project.getRowType().getFieldNames());
+    // If the output type of Aggregate does not match the input type required by original Project,
+    // CAST needs to be added
+    List<RexNode> rexInputRefs = RexPermuteInputsShuttle.of(mapping).visitList(projects2);
```
mihaibudiu, line 161: I imagine the project already contains the cast.
But, in general, the project could have a more complex expression of the aggregate.
Shouldn't the entire expression be preserved?

NobiGo: Here, we convert `CASE(IS NOT NULL($0), CAST($0):INTEGER NOT NULL, 0)` into `SUM0(x)` when `$0` is `SUM(x)`.
I attempted to directly define the return type of `SUM0(x)` as `INTEGER` when constructing it to be consistent with the output type of CASE. However, this was checked in the Aggregate's create method. When the parameter type is TINYINT, the result of SUM0(x) should also be TINYINT other throws exception. Therefore, this approach doesn't work.
When the returned type of the constructed SUM0(x) is TINYINT, not in line with the original plan. Therefore, in the PR, I converted the final type of `SUM0(x)` to be consistent with `CAST`. 

Why don't I directly use `CAST($0):INTEGER NOT NULL` as the parameter for constructing the Aggregate? From this AggregateCall#create API, the parameter is an ordinals, so we need to construct a Project containing `CAST($0):INTEGER NOT NULL`. This approach will make the problem more complicated.

mihaibudiu: you should not depend on a particular type checking strategy for aggregates, that can be changed as part of the type system.

mihaibudiu: in other words, what you have done is the only thing reasonable

mihaibudiu: I guess my question was whether at this point the selectList only contains RexInputRef objects.

NobiGo: This is uncertain. Maybe this naming is a bit confusing. Here, we only focus on changing RexInputRef.

NobiGo: I re-examined the root cause of this ISSUE. It lies in the fact that after adding SUM0(), the referenced type is indeed CAST type, which should be the return type of SUM0(). After performing type conversion in the final Project, it becomes much simpler. Therefore, I made corresponding changes. Please review again.

mihaibudiu (approved):

PR #4202, opened 2025-02-17
[CALCITE-6844] Add all array spark functions which also support in Hive(enable in Hive Library)

xuzifu666:
As describe in JIRA：https://issues.apache.org/jira/browse/CALCITE-6844

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
@@ -1381,7 +1381,7 @@ public abstract class SqlLibraryOperators {
 
   /** The "ARRAY(exp, ...)" function (Spark);
    * compare with the standard array value constructor, "ARRAY [exp, ...]". */
-  @LibraryOperator(libraries = {SPARK})
+  @LibraryOperator(libraries = {SPARK, HIVE})
   public static final SqlFunction ARRAY =
       SqlBasicFunction.create("ARRAY",
           SqlLibraryOperators::arrayReturnType,
@@ -1515,14 +1515,14 @@ public abstract class SqlLibraryOperators {
           OperandTypes.ARRAY_ELEMENT_NONNULL);
 
   /** The "ARRAY_DISTINCT(array)" function. */
-  @LibraryOperator(libraries = {SPARK})
+  @LibraryOperator(libraries = {SPARK, HIVE})
   public static final SqlFunction ARRAY_DISTINCT =
       SqlBasicFunction.create(SqlKind.ARRAY_DISTINCT,
           ReturnTypes.ARG0_NULLABLE,
           OperandTypes.ARRAY);
 
   /** The "ARRAY_EXCEPT(array1, array2)" function. */
-  @LibraryOperator(libraries = {SPARK})
+  @LibraryOperator(libraries = {SPARK, HIVE})
   public static final SqlFunction ARRAY_EXCEPT =
       SqlBasicFunction.create(SqlKind.ARRAY_EXCEPT,
           ReturnTypes.LEAST_RESTRICTIVE,
@@ -1572,7 +1572,7 @@ public abstract class SqlLibraryOperators {
           OperandTypes.ARRAY_INSERT);
 
   /** The "ARRAY_INTERSECT(array1, array2)" function. */
-  @LibraryOperator(libraries = {SPARK})
+  @LibraryOperator(libraries = {SPARK, HIVE})
   public static final SqlFunction ARRAY_INTERSECT =
       SqlBasicFunction.create(SqlKind.ARRAY_INTERSECT,
           ReturnTypes.LEAST_RESTRICTIVE,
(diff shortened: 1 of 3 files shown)


testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -8062,146 +8062,158 @@ void checkRegexpExtract(SqlOperatorFixture f0, FunctionAlias functionAlias) {
     f.checkNull("array_concat(cast(null as integer array), array[1])");
   }
 
-  /** Tests {@code ARRAY_CONTAINS} function from Spark. */
+  /** Tests {@code ARRAY_CONTAINS} function from Spark、Hive. */
```
rubenada, line 8065: Not the usual comma that we normally use, please change it (8 occurrences in this file) (`、` => `,` )

xuzifu666: Done

site/_docs/reference.md
```diff
@@ -2785,24 +2785,24 @@ In the following:
 | s | ARRAY_APPEND(array, element)                   | Appends an *element* to the end of the *array* and returns the result. Type of *element* should be similar to type of the elements of the *array*. If the *array* is null, the function will return null. If an *element* that is null, the null *element* will be added to the end of the *array*
 | s | ARRAY_COMPACT(array)                           | Removes null values from the *array*
 | b | ARRAY_CONCAT(array [, array ]*)                | Concatenates one or more arrays. If any input argument is `NULL` the function returns `NULL`
-| s | ARRAY_CONTAINS(array, element)                 | Returns true if the *array* contains the *element*
-| s | ARRAY_DISTINCT(array)                          | Removes duplicate values from the *array* that keeps ordering of elements
-| s | ARRAY_EXCEPT(array1, array2)                   | Returns an array of the elements in *array1* but not in *array2*, without duplicates
+| h s | ARRAY_CONTAINS(array, element)               | Returns true if the *array* contains the *element*
```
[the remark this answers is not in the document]
xuzifu666: Done, thanks for the reminder. @rubenada

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
```diff
@@ -1508,21 +1508,21 @@ private static RelDataType arrayCompactReturnType(SqlOperatorBinding opBinding)
           OperandTypes.AT_LEAST_ONE_SAME_VARIADIC);
 
   /** The "ARRAY_CONTAINS(array, element)" function. */
-  @LibraryOperator(libraries = {SPARK})
+  @LibraryOperator(libraries = {SPARK, HIVE})
```
NobiGo, line 1511:
``` 
select array_contains(array(1,null),2);
```
In Spark, returns NULL
But In Hive, it returns FALSE
This cannot be the same function.

NobiGo: Have you checked the results of executing different parameters or values in Hive?

xuzifu666: Thanks for your reminder, had change array_contains with Spark, other array functions I would check it again. @NobiGo

xuzifu666: > ```
> select array_contains(array(1,null),2);
> ```
> 
> In Spark, returns NULL But In Hive, it returns FALSE This cannot be the same function.

According discussion with Spark community，this is a bug of spark，and here array_contains(array(1,null),2) should return false not null, at the same time test array_contains(array(1, null),1) would return true in Spark，can we not consider about it currently？other functions test is well with different parameters.  @NobiGo

rubenada: Thanks for checking @xuzifu666 , has a ticket been created for that bug in Spark?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
rubenada: Sorry but the change is incomplete: you modified one ocurrence, but there are still 7 more occurrences of this character that your PR has introduced throughout this file, could you please correct them?

core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java
xuzifu666: Yes, I had submit a issue in Spark JIRA: https://issues.apache.org/jira/browse/SPARK-51319 @rubenada

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
xuzifu666: Done，Thanks @rubenada

rubenada:
LGTM
@NobiGo do you have other remarks?

NobiGo (approved):

PR #4204, opened 2025-02-19
[CALCITE-6846] Support basic dphyp join reorder algorithm

silundong:
The main enumeration process of dphyp has been implemented. There are two new rules, one for converting Joins into a HyperGraph and the other for triggering the dphyp algorithm. More details could see:https://issues.apache.org/jira/browse/CALCITE-6846

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
@@ -16,6 +16,7 @@
  */
 package org.apache.calcite.rel.rules;
 
+import org.apache.calcite.linq4j.function.Experimental;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.Aggregate;
 import org.apache.calcite.rel.core.Calc;
@@ -816,4 +817,17 @@ public class CoreRules {
       WINDOW_REDUCE_EXPRESSIONS =
       ReduceExpressionsRule.WindowReduceExpressionsRule.WindowReduceExpressionsRuleConfig
           .DEFAULT.toRule();
+
+  /** Rule that flattens a tree of {@link LogicalJoin}s
+   * into a single {@link HyperGraph} with N inputs. */
+  @Experimental
+  public static final JoinToHyperGraphRule JOIN_TO_HYPER_GRAPH =
+      JoinToHyperGraphRule.Config.DEFAULT.toRule();
+
+  /** Rule that re-orders a {@link Join} tree using dphyp algorithm.
+   *
+   * @see #JOIN_TO_HYPER_GRAPH */
+  @Experimental
+  public static final DphypJoinReorderRule HYPER_GRAPH_OPTIMIZE =
+      DphypJoinReorderRule.Config.DEFAULT.toRule();
 }
--- a/core/src/main/java/org/apache/calcite/rel/rules/DpHyp.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/DpHyp.java
@@ -1,0 +1,226 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.linq4j.function.Experimental;
+import org.apache.calcite.plan.RelOptCost;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.JoinRelType;
+import org.apache.calcite.rel.metadata.RelMetadataQuery;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.tools.RelBuilder;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * The core process of dphyp enumeration algorithm.
+ */
+@Experimental
+public class DpHyp {
+
+  private final HyperGraph hyperGraph;
+
+  private final HashMap<Long, RelNode> dpTable;
+
+  private final RelBuilder builder;
+
+  private final RelMetadataQuery mq;
+
+  public DpHyp(HyperGraph hyperGraph, RelBuilder builder, RelMetadataQuery relMetadataQuery) {
+    this.hyperGraph =
+        hyperGraph.copy(
+            hyperGraph.getTraitSet(),
+            hyperGraph.getInputs());
+    this.dpTable = new HashMap<>();
+    this.builder = builder;
+    this.mq = relMetadataQuery;
+    // make all field name unique and convert the
+    // HyperEdge condition from RexInputRef to RexInputFieldName
+    this.hyperGraph.convertHyperEdgeCond(builder);
+  }
+
+  /**
+   * The entry function of the algorithm. We use a bitmap to represent a leaf node,
+   * which indicates the position of the corresponding leaf node in {@link HyperGraph}.
+   *
+   * <p>After the enumeration is completed, the best join order will be stored
+   * in the {@link DpHyp#dpTable}.
+   */
+  public void startEnumerateJoin() {
+    int size = hyperGraph.getInputs().size();
+    for (int i = 0; i < size; i++) {
+      long singleNode = LongBitmap.newBitmap(i);
+      dpTable.put(singleNode, hyperGraph.getInput(i));
+      hyperGraph.initEdgeBitMap(singleNode);
+    }
+
+    // start enumerating from the second to last
+    for (int i = size - 2; i >= 0; i--) {
+      long csg = LongBitmap.newBitmap(i);
+      long forbidden = csg - 1;
+      emitCsg(csg);
+      enumerateCsgRec(csg, forbidden);
+    }
+  }
+
+  /**
+   * Given a connected subgraph (csg), enumerate all possible complements subgraph (cmp)
+   * that do not include anything from the exclusion subset.
+   *
+   * <p>Corresponding to EmitCsg in origin paper.
+   */
+  private void emitCsg(long csg) {
+    long forbidden = csg | LongBitmap.getBvBitmap(csg);
+    long neighbors = hyperGraph.getNeighborBitmap(csg, forbidden);
+
+    LongBitmap.ReverseIterator reverseIterator = new LongBitmap.ReverseIterator(neighbors);
+    for (long cmp : reverseIterator) {
+      List<HyperEdge> edges = hyperGraph.connectCsgCmp(csg, cmp);
+      if (!edges.isEmpty()) {
+        emitCsgCmp(csg, cmp, edges);
+      }
+      // forbidden the nodes that smaller than current cmp when extend cmp, e.g.
+      // neighbors = {t1, t2}, t1 and t2 are connected.
+      // when extented t2, we will get (t1, t2)
+      // when extented t1, we will get (t1, t2) repeated
+      long newForbidden =
+              (cmp | LongBitmap.getBvBitmap(cmp)) & neighbors;
+      newForbidden = newForbidden | forbidden;
+      enumerateCmpRec(csg, cmp, newForbidden);
+    }
+  }
+
+  /**
+   * Given a connected subgraph (csg), expands it recursively by its neighbors.
+   * If the expanded csg is connected, try to enumerate its cmp (note that for complex hyperedge,
+   * we only select a single representative node to add to the neighbors, so csg and subNeighbor
+   * are not necessarily connected. However, it still needs to be expanded to prevent missing
+   * complex hyperedge). This method is called after the enumeration of csg is completed,
+   * that is, after {@link DpHyp#emitCsg(long csg)}.
+   *
+   * <p>Corresponding to EnumerateCsgRec in origin paper.
+   */
+  private void enumerateCsgRec(long csg, long forbidden) {
+    long neighbors = hyperGraph.getNeighborBitmap(csg, forbidden);
+    LongBitmap.SubsetIterator subsetIterator = new LongBitmap.SubsetIterator(neighbors);
+    for (long subNeighbor : subsetIterator) {
+      hyperGraph.updateEdgesForUnion(csg, subNeighbor);
+      long newCsg = csg | subNeighbor;
+      if (dpTable.containsKey(newCsg)) {
+        emitCsg(newCsg);
+      }
+    }
+    long newForbidden = forbidden | neighbors;
+    subsetIterator.reset();
+    for (long subNeighbor : subsetIterator) {
+      long newCsg = csg | subNeighbor;
+      enumerateCsgRec(newCsg, newForbidden);
+    }
+  }
+
+  /**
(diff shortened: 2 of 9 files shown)


mihaibudiu:
Is this ready for review?
If not, please mark as draft.

NobiGo:
Some parameters might need to be checked whether they are NULL or undergo active type conversion.

silundong:
> Some parameters might need to be checked whether they are NULL or undergo active type conversion.

Thank you, I will update it.

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
```diff
@@ -816,4 +816,14 @@ private CoreRules() {}
       WINDOW_REDUCE_EXPRESSIONS =
       ReduceExpressionsRule.WindowReduceExpressionsRule.WindowReduceExpressionsRuleConfig
           .DEFAULT.toRule();
+  /** Rule that flattens a tree of {@link LogicalJoin}s
```
rubenada, line 819: nit: for consistency, could you please keep an empty between the new rule and the previous one?

core/src/main/java/org/apache/calcite/rel/rules/DphypJoinReorderRule.java
```diff
@@ -0,0 +1,86 @@
...
+    // make all field name unique and convert the
+    // HyperEdge condition from RexInputRef to RexInputFieldName
+    hyperGraph.convertHyperEdgeCond();
+
+    // enumerate by Dphyp
+    DpHyp dpHyp = new DpHyp(hyperGraph, relBuilder, call.getMetadataQuery());
+    dpHyp.startEnumerateJoin();
+    RelNode orderedJoin = dpHyp.getBestPlan();
+    if (orderedJoin == null) {
+      return;
```
rubenada, line 56: If we get here, it means something went wrong, doesn't it?
IIANM this would lead to a CannotPlanException (if we cannot decompose the HyperGraph into Joins, the we'll keep the HyperGraph which is a "not implementable" operator). In that case, maybe it would be a good idea to throw an exception (to fail early and with a clearer message) instead of simply returning (and failing later with a more generic and "cryptic" exception)

rubenada:
This is a great contribution @silundong.
Just out of curiosity, have you tested locally what happens if you add the new HyperGraph rules into the "default rules" (e.g. RelOptRules#BASE_RULES, or Programs#RULE_SET) and execute Calcite tests? Do you get any any suspicious error (except for the expected plan changes regarding Join-ordering that the new rules may generate)?

core/src/main/java/org/apache/calcite/rel/rules/DpHyp.java
```diff
@@ -0,0 +1,179 @@
...
+  public @Nullable RelNode getBestPlan() {
+    int size = hyperGraph.getInputs().size();
+    long wholeGraph = LongBitmap.newBitmapBetween(0, size);
+    return dpTable.get(wholeGraph);
+  }
+
+  private RelNode chooseBetterPlan(RelNode plan1, RelNode plan2) {
+    RelOptCost cost1 = mq.getCumulativeCost(plan1);
+    RelOptCost cost2 = mq.getCumulativeCost(plan2);
+    if (cost1 == null || cost2 == null) {
```
rubenada, line 172: if we have one null cost and one non-null cost, shouldn't we return the non-null one?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -9546,4 +9546,21 @@ private void checkJoinAssociateRuleWithTopAlwaysTrueCondition(boolean allowAlway
         .withRule(CoreRules.MULTI_JOIN_OPTIMIZE)
         .check();
   }
+
+  @Test void testDphypJoinReorder() {
```
rubenada, line 9550: I'd be nice to have a few more tests with more variety of scenarios.

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
silundong: Sure, I will update it.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
silundong: Sure, I will update it.

core/src/main/java/org/apache/calcite/rel/rules/DpHyp.java
silundong: I will update it.

core/src/main/java/org/apache/calcite/rel/rules/DphypJoinReorderRule.java
silundong: My original intention is that when the query is too complex, even dphyp can be very time-consuming. So, if the execution of `dpHyp.startEnumerateJoin` exceeds a certain time limit, the enumeration process should be forcibly terminated. In this case, the dp table may not have results yet. After returning null, the hypergraph simplification algorithm might be executed.
Alternatively, the rules related to dphyp could be integrated into a single method, with the original Join tree as the input parameter. If the optimization times out, the original Join tree is returned, and other Join reorder algorithms can be called subsequently.
Do you have any suggestions for this? I will update it accordingly.

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
silundong: Sure, I will update it.

silundong:
> This is a great contribution @silundong. Just out of curiosity, have you tested locally what happens if you add the new HyperGraph rules into the "default rules" (e.g. RelOptRules#BASE_RULES, or Programs#RULE_SET) and execute Calcite tests? Do you get any any suspicious error (except for the expected plan changes regarding Join-ordering that the new rules may generate)?

Thank you very much for your recognition. I did not add it to the "default rules" (e.g. RelOptRules#BASE_RULES, or Programs#RULE_SET). However, in other projects, I have integrated the Calcite parsing and optimization module and added HyperGraph rules. It has worked without exceptions under TPC-DS.

rubenada:
Thanks for addressing my initial comments @silundong , I'll take another look in the coming days.

silundong:
@rubenada Thank you very much for your review!

core/src/main/java/org/apache/calcite/rel/rules/DpHyp.java
```diff
@@ -0,0 +1,181 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * The core process of dphyp enumeration algorithm.
+ */
+public class DpHyp {
+
+  private HyperGraph hyperGraph;
```
rubenada, line 36: If the fields in the class (or any new class) can be final, please declare them as final.

core/src/main/java/org/apache/calcite/rel/rules/DpHyp.java
```diff
@@ -0,0 +1,181 @@
...
+  private RelMetadataQuery mq;
+
+  public DpHyp(HyperGraph hyperGraph, RelBuilder builder, RelMetadataQuery relMetadataQuery) {
+    this.hyperGraph = hyperGraph;
+    this.dpTable = new HashMap<>();
+    this.builder = builder;
+    this.mq = relMetadataQuery;
+  }
+
+  public void startEnumerateJoin() {
```
rubenada, line 51: IMO the methods of this class (except the "obvious" ones like `getBestPlan` or `chooseBetterPlan`) should have a javadoc describing their intent.

core/src/main/java/org/apache/calcite/rel/rules/DphypJoinReorderRule.java
```diff
@@ -0,0 +1,86 @@
...
+  protected DphypJoinReorderRule(Config config) {
+    super(config);
+  }
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    HyperGraph hyperGraph = call.rel(0);
+    RelBuilder relBuilder = call.builder();
+    // make all field name unique and convert the
+    // HyperEdge condition from RexInputRef to RexInputFieldName
+    hyperGraph.convertHyperEdgeCond();
```
rubenada, line 49: I think this call over here is a bit weird, what about calling this method inside the DpHyp constructor (to make sure we don't "forget" about this initialization)

core/src/main/java/org/apache/calcite/rel/rules/HyperEdge.java
```diff
@@ -0,0 +1,82 @@
...
+
+import org.apache.calcite.rel.core.JoinRelType;
+import org.apache.calcite.rex.RexNode;
+
+/**
+ * Edge in HyperGraph, that represents a join predicate.
+ */
+public class HyperEdge {
+
+  private long leftNodeBits;
```
rubenada, line 27: final fields?

core/src/main/java/org/apache/calcite/rel/rules/HyperEdge.java
```diff
@@ -0,0 +1,82 @@
...
+    return joinType;
+  }
+
+  public RexNode getCondition() {
+    return condition;
+  }
+
+  @Override public String toString() {
+    StringBuilder sb = new StringBuilder();
+    sb.append(LongBitmap.printBitmap(leftNodeBits))
```
rubenada, line 71: IMO the string representation should contain also the condition and the joinRelType

core/src/main/java/org/apache/calcite/rel/rules/HyperEdge.java
```diff
@@ -0,0 +1,82 @@
...
+        .append("——").append(joinType).append("——")
+        .append(LongBitmap.printBitmap(rightNodeBits));
+    return sb.toString();
+  }
+
+  // before starting dphyp, replace RexInputRef to RexInputFieldName
+  public void replaceCondition(RexNode fieldNameCond) {
```
rubenada, line 78: I'm not a fan of this type of manipulations, but I understand it's a pragmatic choice; but I'd suggest to reduce the method's visibility (package-private?)

core/src/main/java/org/apache/calcite/rel/rules/HyperGraph.java
```diff
@@ -0,0 +1,459 @@
...
+  // node bitmap overlaps edge's leftNodeBits or rightNodeBits, but does not completely cover
+  private final HashMap<Long, BitSet> overlapEdgesMap;
+
+  protected HyperGraph(RelOptCluster cluster,
+      RelTraitSet traitSet,
+      List<RelNode> inputs,
+      List<HyperEdge> edges,
+      RelDataType rowType) {
+    super(cluster, traitSet);
+    this.inputs = inputs;
```
rubenada, line 84: Probably it'll be wise to make a defensive copy of the lists of inputs and edges, just in case.

core/src/main/java/org/apache/calcite/rel/rules/HyperGraph.java
```diff
@@ -0,0 +1,459 @@
...
+
+  private final List<RelNode> inputs;
+
+  @SuppressWarnings("HidingField")
+  private final RelDataType rowType;
+
+  private final List<HyperEdge> edges;
+
+  // record the indices of complex hyper edges in the 'edges'
+  private final BitSet complexEdgesBitmap;
```
rubenada, line 62: Could this be an ImmutableBitSet (build on the constructor)?

core/src/main/java/org/apache/calcite/rel/rules/HyperGraph.java
```diff
@@ -0,0 +1,459 @@
...
+  }
+
+  @Override public RelNode copy(RelTraitSet traitSet, List<RelNode> inputs) {
+    return new HyperGraph(
+        getCluster(),
+        traitSet,
+        inputs,
+        edges,
+        rowType,
+        complexEdgesBitmap,
```
rubenada, line 127: Shouldn't we pass here copies of the maps, instead of the maps themselves?

core/src/main/java/org/apache/calcite/rel/rules/HyperGraph.java
```diff
@@ -0,0 +1,459 @@
...
+      for (int i = 0; i < input.getRowType().getFieldCount(); i++) {
+        projects.add(
+            new RexInputRef(
+            i,
+            fieldList.get(fieldIndex).getType()));
+        names.add(fieldList.get(fieldIndex).getName());
+        fieldIndex++;
+      }
+      RelNode renameProject =
+          LogicalProject.create(
```
rubenada, line 400: To make things more generic, I think this method should receive a RelBuilder, and instead of creating a new `LogicalProject`, it should do a relBuilder.push(input) + relBuilder.project(...) + relBuilder.build() to construct the new inputs; that would allow users that do not work with logical-convention to get the appropriate Project operator if the use their own RelBuilder

core/src/main/java/org/apache/calcite/rel/rules/JoinToHyperGraphRule.java
```diff
@@ -0,0 +1,192 @@
...
+            origJoin.getTraitSet(),
+            inputs,
+            edges,
+            origJoin.getRowType());
+
+    call.transformTo(result);
+  }
+
+  private static HyperEdge adjustNodeBit(HyperEdge hyperEdge, int nodeOffset, int fieldOffset) {
+    RexShuttle shuttle = new RexShuttle() {
```
rubenada, line 163: IIANM this shuttle is not needed, you can reuse `RexUtil#shift(RexNode node, final int offset)` method

core/src/main/java/org/apache/calcite/rel/rules/HyperGraph.java
```diff
@@ -0,0 +1,459 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * HyperGraph represents a join graph.
+ */
+public class HyperGraph extends AbstractRelNode {
```
rubenada, line 52: Probably not required at this point, but I think it'll be a good idea (and would avoid potential issues in the future) if this class overrides the `accept(RexShuttle shuttle)` method, which I guess should return a new HyperGraph where the RexShuttle has been applied to the edges conditions?

rubenada:
@silundong  I have left some remarks , rather generic or Calcite-related, not really dphyp-algo related, which I find I bit hard to follow, but it's good that you provided tests showing the results.
IMO we could consider adding this feature, maybe initially flagging the new classes and rule definitions as `@Experimental` (as it happened in the past with other features). Let's see what other contributors think about it...

silundong:
> @silundong I have left some remarks , rather generic or Calcite-related, not really dphyp-algo related, which I find I bit hard to follow, but it's good that you provided tests showing the results. IMO we could consider adding this feature, maybe initially flagging the new classes and rule definitions as `@Experimental` (as it happened in the past with other features). Let's see what other contributors think about it...

Thank you very much for your suggestion, I will update them.

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
```diff
@@ -816,4 +816,15 @@ private CoreRules() {}
       WINDOW_REDUCE_EXPRESSIONS =
       ReduceExpressionsRule.WindowReduceExpressionsRule.WindowReduceExpressionsRuleConfig
           .DEFAULT.toRule();
+
+  /** Rule that flattens a tree of {@link LogicalJoin}s
+   * into a single {@link HyperGraph} with N inputs. */
+  public static final JoinToHyperGraphRule JOIN_TO_HYPER_GRAPH =
```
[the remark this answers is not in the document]
silundong: Sure, I will update it

rubenada:
@silundong could you please take a look at Alessandro's minor comments in the Jira ticket?

silundong:
> @silundong could you please take a look at Alessandro's minor comments in the Jira ticket?

Sure, I will indicate in HyperGraph and HyperEdge that the code is referenced from Apache Doris. Is that ok?

asolimando:
> > @silundong could you please take a look at Alessandro's minor comments in the Jira ticket?
> 
> Sure, I will indicate in HyperGraph and HyperEdge that the code is referenced from Apache Doris. Is that ok?

Thanks @silundong, please put a link to the two corresponding files using a fixed release or commit other than master/main, so it stays stable over time.

PR #4206, opened 2025-02-19
[CALCITE-5529] Improve dialect tests, part 1: check each query against a reference dialect

julianhyde:
See CALCITE-5529.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
@@ -2882,10 +2882,9 @@ public class RelDecorrelator implements ReflectiveVisitor {
     }
 
     @Override public String toString() {
-      return "mapRefRelToCorRef=" + mapRefRelToCorRef
-          + "\nmapCorToCorRel=" + mapCorToCorRel
-          + "\nmapFieldAccessToCorRef=" + mapFieldAccessToCorRef
-          + "\n";
+      return "mapRefRelToCorRef=" + mapRefRelToCorRef + "\n"
+          + "mapCorToCorRel=" + mapCorToCorRel + "\n"
+          + "mapFieldAccessToCorRef=" + mapFieldAccessToCorRef + "\n";
     }
 
     @SuppressWarnings("UndefinedEquals")
--- a/core/src/main/java/org/apache/calcite/util/Token.java
+++ b/core/src/main/java/org/apache/calcite/util/Token.java
@@ -1,0 +1,98 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.util;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/** Hands out tokens, and throws if they are not all released.
+ *
+ * <p>Typical use:
+ *
+ * <blockquote><pre>{@code
+ * Token.Pool pool = Token.pool();
+ * Token token1 = pool.token();
+ * Token token2 = pool.token();
+ * token1.close();
+ * pool.assertEmpty(); // throws because token2 has not been closed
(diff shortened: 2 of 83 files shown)


core/src/test/java/org/apache/calcite/rel/rel2sql/DialectTestConfig.java
```diff
@@ -0,0 +1,302 @@
...
+            CalciteAssert.SchemaSpec.BOOKSTORE,
+            CalciteAssert.SchemaSpec.JDBC_FOODMART,
+            CalciteAssert.SchemaSpec.POST,
+            CalciteAssert.SchemaSpec.SCOTT,
+            CalciteAssert.SchemaSpec.SCOTT_WITH_TEMPORAL,
+            CalciteAssert.SchemaSpec.TPCH);
+        final Properties info = new Properties();
+        // Hive for RLIKE, Postgres for ILIKE, Spark for EXISTS, etc.
+        info.put(CalciteConnectionProperty.FUN.name(),
+            "standard,postgresql,bigquery,hive,spark");
```
ILuffZhe, line 270: One small question: why those libraries are chosen?

core/src/test/java/org/apache/calcite/rel/rel2sql/DialectTestConfig.java
```diff
@@ -0,0 +1,302 @@
...
+        info.put(CalciteConnectionProperty.SCHEMA.name(),
+            schemaSpec.schemaName);
+        info.put(CalciteConnectionProperty.CONFORMANCE.name(),
+            SqlConformanceEnum.LENIENT.name());
```
ILuffZhe, line 274: Why do we use LENIENT rather than DEFAULT?

core/src/test/java/org/apache/calcite/rel/rel2sql/DialectTestConfigs.java
```diff
@@ -0,0 +1,204 @@
...
+        return DialectTestConfig.of(list);
+      })::get;
+
+
+  @SuppressWarnings("SameParameterValue")
+  static HiveSqlDialect hiveDialect(int majorVersion, int minorVersion) {
+    return new HiveSqlDialect(HiveSqlDialect.DEFAULT_CONTEXT
+        .withDatabaseMajorVersion(majorVersion)
+        .withDatabaseMinorVersion(minorVersion)
+        .withNullCollation(NullCollation.LOW));
```
ILuffZhe, line 70: DEFAULT_CONTEXT in HiveSqlDialect already has the `NullCollation.LOW`.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlFixture.java
```diff
@@ -0,0 +1,591 @@
...
+
+  RelToSqlFixture withInformix() {
+    return dialect(INFORMIX);
+  }
+
+  RelToSqlFixture withJethro() {
+    return dialect(JETHRO);
+  }
+
+  RelToSqlFixture withMssql() {
```
ILuffZhe, line 304: Do we need other withMssql for different versions?

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -2882,10 +2882,9 @@ private CorelMap(Multimap<RelNode, CorRef> mapRefRelToCorRef,
     }
 
     @Override public String toString() {
-      return "mapRefRelToCorRef=" + mapRefRelToCorRef
-          + "\nmapCorToCorRel=" + mapCorToCorRel
-          + "\nmapFieldAccessToCorRef=" + mapFieldAccessToCorRef
-          + "\n";
+      return "mapRefRelToCorRef=" + mapRefRelToCorRef + "\n"
```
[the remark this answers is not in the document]
NobiGo: I noticed the code in `LintTest`.

core/src/main/java/org/apache/calcite/util/Token.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /** Releases a token id. Should be called from {@link Token#close()}. */
+    @SuppressWarnings("resource")
+    private boolean release(int id) {
+      return map.remove(id) != null;
+    }
+
+    /** Throws if not all fixtures have been released. */
+    public void assertEmpty() {
+      int size = map.size();
```
ILuffZhe, line 88: Is Pool able to handle this situation:
- when calling `map.size()` we get 0
- before calling `map.isEmpty()`, a token is added
- so the pool is not empty but with size 0

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlFixture.java
```diff
@@ -0,0 +1,591 @@
...
+    final AtomicReference<List<String>> referenceValid = new AtomicReference<>();
+    if (testConfig.refDialectCode != null) {
+      DialectTestConfig.Dialect referenceDialect =
+          testConfig.get(testConfig.refDialectCode);
+      final String referenceSql =
+          testConfig.refDialectCode == CALCITE && relFn == null
+              ? sql
+              : dialect(testConfig.refDialectCode).exec();
+      switch (phase) {
+      case PREPARE:
```
ILuffZhe, line 483: What's phase PARSE for and do we need to handle it here?

caicancai:
I will test it locally

PR #4207, opened 2025-02-21
[CALCITE-6850] ProjectRemoveRule with two Projects does not keep field names from the top one

xiedeyantu:
issue: https://issues.apache.org/jira/browse/CALCITE-6850

### DIFF
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -26,6 +26,7 @@ import org.apache.calcite.adapter.enumerable.EnumerableSort;
 import org.apache.calcite.config.CalciteConnectionConfig;
 import org.apache.calcite.plan.Contexts;
 import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.plan.RelOptCostImpl;
 import org.apache.calcite.plan.RelOptPlanner;
 import org.apache.calcite.plan.RelOptRule;
 import org.apache.calcite.plan.RelOptRuleCall;
@@ -9584,5 +9585,30 @@ class RelOptRulesTest extends RelOptTestBase {
         .withPre(program)
         .withRule(CoreRules.MULTI_JOIN_OPTIMIZE)
         .check();
+  }
+
+  /**
+   * Test case of
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6850">[CALCITE-6850]
+   * ProjectRemoveRule with two Projects
+   * does not keep field names from the top one</a>. */
+  @Test void testProjectRemoveRule() {
+    final Function<RelBuilder, RelNode> relFn = b -> {
+      RelNode r = b.scan("DEPT")
+          .project(
+              b.call(SqlStdOperatorTable.PLUS,
+              ImmutableList.of(b.field(0), b.literal(1))))
+          .build();
+      b.push(r);
+      return LogicalProject.create(r, ImmutableList.of(),
+          ImmutableList.of(b.field(0)),
+          ImmutableList.of("newAlias"), ImmutableSet.of());
+    };
+    HepProgramBuilder builder = new HepProgramBuilder();
+    builder.addRuleInstance(CoreRules.PROJECT_REMOVE);
+    HepPlanner planner =
+        new HepPlanner(builder.build(), null, true, null, RelOptCostImpl.FACTORY);
+    fixture().withRelBuilderConfig(a -> a.withBloat(-1))
+        .relFn(relFn).withPlanner(planner).check();
   }
 }
--- a/core/src/main/java/org/apache/calcite/rel/rules/ProjectRemoveRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/ProjectRemoveRule.java
@@ -58,7 +58,7 @@ public class ProjectRemoveRule
   @Override public void onMatch(RelOptRuleCall call) {
     Project project = call.rel(0);
     assert isTrivial(project);
-    RelNode stripped = project.getInput();
+    RelNode stripped = call.rel(1);
     if (stripped instanceof Project) {
       // Rename columns of child projection if desired field names are given.
       Project childProject = (Project) stripped;
@@ -97,7 +97,8 @@ public class ProjectRemoveRule
                 // Use a predicate to detect non-matches early.
                 // This keeps the rule queue short.
                 .predicate(ProjectRemoveRule::isTrivial)
-                .anyInputs());
+                .oneInput(
+                    b1 -> b1.operand(RelNode.class).anyInputs()));
 
     @Override default ProjectRemoveRule toRule() {
       return new ProjectRemoveRule(this);
--- a/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
@@ -8722,6 +8722,21 @@ LogicalProject(EXPR$0=[CAST(/(CASE(>($2, 0), $3, null:INTEGER), $2)):INTEGER])
 ]]>
     </Resource>
   </TestCase>
+  <TestCase name="testProjectRemoveRule">
+    <Resource name="planBefore">
+      <![CDATA[
+LogicalProject(newAlias=[$0])
+  LogicalProject($f0=[+($0, 1)])
+    LogicalTableScan(table=[[scott, DEPT]])
+]]>
+    </Resource>
+    <Resource name="planAfter">
+      <![CDATA[
+LogicalProject(newAlias=[+($0, 1)])
+  LogicalTableScan(table=[[scott, DEPT]])
+]]>
+    </Resource>
+  </TestCase>
   <TestCase name="testProjectSetOpTranspose">
     <Resource name="sql">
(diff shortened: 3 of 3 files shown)


mihaibudiu:
What is the problem that this is fixing?
Is there a test that can exhibit the problem?

suibianwanwank:
You should create an issue to let others understand what's going on here.

xiedeyantu:
> What is the problem that this is fixing? Is there a test that can exhibit the problem?
The stripped type always is HepRelVertex, Project type will not appear. But it's difficult to construct an example using SQL.

xiedeyantu:
> You should create an issue to let others understand what's going on here.

ok, I will do it.

mihaibudiu:
You don't necessarily have to create an example starting from SQL, there are lots of example tests that build data structures directly, e.g., using RelBuilder. But you should be able to show a bug that is being addressed. Maybe you can describe the context in which the problem happens and someone can help you craft a test.

xiedeyantu:
> 您不一定非要从 SQL 开始创建示例，有很多示例测试可以直接构建数据结构，例如使用 RelBuilder。但您应该能够展示正在解决的错误。也许您可以描述问题发生的背景，然后有人可以帮助您制定测试。

Could you tell me a test file name please? I try to add a test, thanks.

xiedeyantu:
> You don't necessarily have to create an example starting from SQL, there are lots of example tests that build data structures directly, e.g., using RelBuilder. But you should be able to show a bug that is being addressed. Maybe you can describe the context in which the problem happens and someone can help you craft a test.

Could you tell me a test file name please? I try to add a test, thanks.

mihaibudiu:
Any file named like *Test.java is a file with tests in Calcite.
How did you discover the bug?

xiedeyantu:
> Any file named like *Test.java is a file with tests in Calcite. How did you discover the bug?

It is difficult to construct using SQL, it may be a problem from a code perspective.

suibianwanwank:
I presume you should need to add a test in _RelOptRulesTest_, where you can add an example of a RelNode that was not successfully optimized before this PR, and applied the optimization after this PR.

rubenada:
Thanks for creating this PR @xiedeyantu . I'd suggest to take a look at our [contributing](https://calcite.apache.org/develop/#contributing) guidelines to know a bit more about our conventions (e.g. create a Jira ticket, do not end commit message with a period, describe the issue rather than using "Fix..." on the Jira title and commit message, etc.).

About this particular patch, I have quickly looked into it, and it seems a potentially valid (minor) issue, basically the "if" block is never executed because the `RelNode stripped` is actually not stripped, e.g. it would be a HepRelVertex (or Volcano's RelSubset) instead of the actual Project inside of it. The consequence of this is minor (the rule's result rowType will have the bottom's Project fieldnames instead of the top Project's). IMO the Jira title / commit message should describe this circumstance. As @suibianwanwank mentioned, probably RelOptRulesTest is the easiest place to produce a unit test showing the issue.

xiedeyantu:
> consequence

Sorry, I have tried many methods, including SQL and Relbuilder. Calcite can now effectively merge prejects that have same fields but different aliases in the SQLToRel stage, so I was unable to find a way to supplement a test case. Any idea for me?

rubenada:
Yes... Calcite usually is quite smart to simplify these type of things transparently on the fly, so you'd probably need to "force" a bit the situation...
What about a RelBuilder scenario with a Scan + Project (trim fields) + Filter + Project (rename); and then applying FilterProjectTranspose and then ProjectRemoveRule? I mean something like:
```
Project(x as a, y as b)
  Filter(x>0)
    Project(x, y)
      Scan (TableA, fields x, y, x)

=> FilterProjectTranspose

Project(x as a, y as b)
  Project(x, y)
    Filter(x>0)
      Scan (TableA, fields x, y, x)

=> ProjectRemoveRule! (scenario reproduced?)
```

xiedeyantu:
> Yes... Calcite usually is quite smart to simplify these type of things transparently on the fly, so you'd probably need to "force" a bit the situation... What about a RelBuilder scenario with a Scan + Project (trim fields) + Filter + Project (rename); and then applying FilterProjectTranspose and then ProjectRemoveRule? I mean something like:
> 
> ```
> Project(x as a, y as b)
>   Filter(x>0)
>     Project(x, y)
>       Scan (TableA, fields x, y, x)
> 
> => FilterProjectTranspose
> 
> Project(x as a, y as b)
>   Project(x, y)
>     Filter(x>0)
>       Scan (TableA, fields x, y, x)
> 
> => ProjectRemoveRule! (scenario reproduced?)
> ```

Many people in the issue gave me suggestions, and I tried them all. If the digest calculation of the project alias is not considered in the calculation of the equivalence relationship, I cannot find a case that can prove that the logic enters the if, because even though it enters the if, stripped also modifies the alias correctly, but stripped cannot be correctly selected in findBestExpr, so does this pr do not need to add a case? I have no better idea.

rubenada:
@xiedeyantu I understand the situation, in this case it seems we need to "force" a bit the scenario, but as a rule of thumb we really demand (generally) a unit test "showing" the issue together with the fix. IIANM using a HepPlanner with noDag=true avoids the refactoring that is "blocking" your test, could you please verify if the following test (to be added in RelOptRulesTest.java):
```
  @Test void testProjectRemoveRule() {
    final Function<RelBuilder, RelNode> relFn = b -> {
      RelNode r = b.scan("DEPT")
          .project(b.call(SqlStdOperatorTable.PLUS, ImmutableList.of(b.field(0), b.literal(1))))
          .build();
      b.push(r);
      return LogicalProject.create(r, ImmutableList.of(), ImmutableList.of(b.field(0)), ImmutableList.of("newAlias"), ImmutableSet.of());
    };
    HepProgramBuilder builder = new HepProgramBuilder();
    builder.addRuleInstance(CoreRules.PROJECT_REMOVE);
    HepPlanner planner = new HepPlanner(builder.build(), null, true, null, RelOptCostImpl.FACTORY);
    fixture().withRelBuilderConfig(a -> a.withBloat(-1))
        .relFn(relFn).withPlanner(planner).check();
  }
```

Together with this new context in RelOptRulesTest.xml (notice it must be added in alphabetical order by test name)
```
  <TestCase name="testProjectRemoveRule">
    <Resource name="planBefore">
      <![CDATA[
LogicalProject(newAlias=[$0])
  LogicalProject($f0=[+($0, 1)])
]]>
    </Resource>
    <Resource name="planAfter">
      <![CDATA[
LogicalProject(newAlias=[+($0, 1)])
]]>
    </Resource>
  </TestCase>
```

Fails on the current code, but passes with your patch? Thanks

xiedeyantu:
> @xiedeyantu I understand the situation, in this case it seems we need to "force" a bit the scenario, but as a rule of thumb we really demand (generally) a unit test "showing" the issue together with the fix. IIANM using a HepPlanner with noDag=true avoids the refactoring that is "blocking" your test, could you please verify if the following test (to be added in RelOptRulesTest.java):
> 
> ```
>   @Test void testProjectRemoveRule() {
>     final Function<RelBuilder, RelNode> relFn = b -> {
>       RelNode r = b.scan("DEPT")
>           .project(b.call(SqlStdOperatorTable.PLUS, ImmutableList.of(b.field(0), b.literal(1))))
>           .build();
>       b.push(r);
>       return LogicalProject.create(r, ImmutableList.of(), ImmutableList.of(b.field(0)), ImmutableList.of("newAlias"), ImmutableSet.of());
>     };
>     HepProgramBuilder builder = new HepProgramBuilder();
>     builder.addRuleInstance(CoreRules.PROJECT_REMOVE);
>     HepPlanner planner = new HepPlanner(builder.build(), null, true, null, RelOptCostImpl.FACTORY);
>     fixture().withRelBuilderConfig(a -> a.withBloat(-1))
>         .relFn(relFn).withPlanner(planner).check();
>   }
> ```
> 
> Together with this new context in RelOptRulesTest.xml (notice it must be added in alphabetical order by test name)
> 
> ```
>   <TestCase name="testProjectRemoveRule">
>     <Resource name="planBefore">
>       <![CDATA[
> LogicalProject(newAlias=[$0])
>   LogicalProject($f0=[+($0, 1)])
>     LogicalTableScan(table=[[scott, DEPT]])
> ]]>
>     </Resource>
>     <Resource name="planAfter">
>       <![CDATA[
> LogicalProject(newAlias=[+($0, 1)])
>   LogicalTableScan(table=[[scott, DEPT]])
> ]]>
>     </Resource>
>   </TestCase>
> ```
> 
> Fails on the current main branch, but passes with your patch? Thanks

@rubenada Thank you very much for your suggestion. I have updated the test cases, please help review them.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -9585,4 +9586,24 @@ private void checkJoinAssociateRuleWithTopAlwaysTrueCondition(boolean allowAlway
         .withRule(CoreRules.MULTI_JOIN_OPTIMIZE)
         .check();
   }
+
+  @Test void testProjectRemoveRule() {
```
rubenada, line 9590: To be consistent with other test, please add javadoc referring to the ticket:
```
  /**
   * Test case of
   * <a href="https://issues.apache.org/jira/browse/CALCITE-6850">[CALCITE-6850]
   * ...</a>. */
  @Test void testProjectRemoveRule() {
```

xiedeyantu:
@rubenada Please help me to review it, thanks.

rubenada (approved):

PR #4209, opened 2025-02-23
[CALCITE-6836] Add Rule to convert INTERSECT to EXISTS

xiedeyantu:
issue: https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6836

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/IntersectToExistsRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/IntersectToExistsRule.java
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelOptUtil.Exists;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.CorrelationId;
+import org.apache.calcite.rel.core.Intersect;
+import org.apache.calcite.rel.logical.LogicalIntersect;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexSubQuery;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.tools.RelBuilder;
+import org.apache.calcite.tools.RelBuilderFactory;
+import org.apache.calcite.util.ImmutableBitSet;
+
+import com.google.common.collect.ImmutableSet;
+
+import org.immutables.value.Value;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Planner rule that translates a {@link Intersect}
+ * (<code>all</code> = <code>false</code>)
+ * into a {@link Exists}.
+ *
+ * @see CoreRules#INTERSECT_TO_EXISTS
+ */
+@Value.Enclosing
+public class IntersectToExistsRule
+    extends RelRule<IntersectToExistsRule.Config>
+    implements TransformationRule {
+
+  /** Creates an IntersectToExistRule. */
+  protected IntersectToExistsRule(Config config) {
+    super(config);
+  }
+
+  @Deprecated // to be removed before 2.0
+  public IntersectToExistsRule(Class<? extends Intersect> intersectClass,
+      RelBuilderFactory relBuilderFactory) {
+    this(Config.DEFAULT.withRelBuilderFactory(relBuilderFactory)
+        .as(Config.class)
+        .withOperandFor(intersectClass));
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    final Intersect intersect = call.rel(0);
+    if (intersect.all) {
+      return; // nothing we can do
+    }
+
+    final RelBuilder builder = call.builder();
+    final RexBuilder rexBuilder = builder.getRexBuilder();
+
+    RelDataType rowType = intersect.getRowType();
+    List<RelNode> inputs = intersect.getInputs();
+    RelNode current = inputs.get(0);
+
+    // get all column indices of intersect
+    ImmutableBitSet fieldIndices = ImmutableBitSet.of(rowType.getFieldList()
+        .stream().map(RelDataTypeField::getIndex)
+        .collect(Collectors.toList()));
+
+    // iterate over the inputs and apply exists subquery
+    for (int i = 1; i < inputs.size(); i++) {
+      RelNode nextInput = inputs.get(i).stripped();
+
+      // create correlation
+      CorrelationId correlationId = intersect.getCluster().createCorrel();
+      RexNode correl =
+          rexBuilder.makeCorrel(rowType, correlationId);
+
+      // create condition in exists filter, and use correlation
+      List<RexNode> conditions = new ArrayList<>();
+      for (int fieldIndex : fieldIndices) {
+        RexNode outerField = rexBuilder.makeInputRef(rowType, fieldIndex);
+        RexNode innerField = rexBuilder.makeFieldAccess(correl, fieldIndex);
+        conditions.add(builder.isNotDistinctFrom(innerField, outerField));
+      }
+      RexNode condition = RexUtil.composeConjunction(rexBuilder, conditions);
+
+      // build exists subquery
+      RelNode existsSubQuery = builder.push(nextInput)
+          .filter(condition)
+          .project(builder.fields(fieldIndices))
+          .build();
+
+      // apply exists subquery to the current relation
+      current = builder.push(current)
+          .filter(ImmutableSet.of(correlationId),
+              RexSubQuery.exists(existsSubQuery))
+          .build();
+    }
+
+    builder.push(current);
+    List<RexNode> projects = new ArrayList<>();
+    for (int fieldIndex : fieldIndices) {
+      RexNode rexNode = builder.fields().get(fieldIndex);
+      RelDataType originalType =
+          rowType.getFieldList().get(projects.size()).getType();
+      RexNode expr;
+      if (!originalType.equals(rexNode.getType())) {
+        expr = rexBuilder.makeCast(originalType, rexNode, true, false);
+      } else {
+        expr = rexNode;
+      }
+      projects.add(expr);
+    }
+
+    RelNode result = builder.project(projects)
+        .distinct()
+        .build();
+
+    call.transformTo(result);
+  }
+
+  /** Rule configuration. */
+  @Value.Immutable
+  public interface Config extends RelRule.Config {
+    Config DEFAULT = ImmutableIntersectToExistsRule.Config.of()
+        .withOperandFor(LogicalIntersect.class);
+
+    @Override default IntersectToExistsRule toRule() {
+      return new IntersectToExistsRule(this);
+    }
+
+    /** Defines an operand tree for the given classes. */
+    default Config withOperandFor(Class<? extends Intersect> intersectClass) {
+      return withOperandSupplier(b -> b.operand(intersectClass).anyInputs())
+          .as(Config.class);
+    }
+  }
+}
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -9986,4 +9986,66 @@ class RelOptRulesTest extends RelOptTestBase {
         .withRule(CoreRules.HYPER_GRAPH_OPTIMIZE, CoreRules.PROJECT_REMOVE, CoreRules.PROJECT_MERGE)
         .check();
   }
+
+  /**
+   * Test case of
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6836">[CALCITE-6836]
+   * Add Rule to convert INTERSECT to EXISTS</a>. */
+  @Test void testIntersectToExistsRuleOneField() {
+    String sql = "SELECT a.ename FROM emp AS a\n"
+        + "INTERSECT\n"
+        + "SELECT b.name FROM dept AS b";
+    sql(sql).withRule(CoreRules.INTERSECT_TO_EXISTS)
+        .check();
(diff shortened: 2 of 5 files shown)


xiedeyantu:
Hi, @NobiGo please review thanks !

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -9585,4 +9585,12 @@ private void checkJoinAssociateRuleWithTopAlwaysTrueCondition(boolean allowAlway
         .withRule(CoreRules.MULTI_JOIN_OPTIMIZE)
         .check();
   }
+
+  @Test void testIntersectToExistsRule() {
```
NobiGo, line 9589: 1. Please add the Jira link
2. Add more test cases about 
-  INTERSECT ALL
-  INTERSECT on multiple fields
-  Multiple INTERSECT
3. We may not adopt this rule as the default one for now, but it can be set as the default rule to test whether the rule function is normal.

xiedeyantu: Thanks for your reply, I will add jira link and more tests.

NobiGo (commented):
According to the description provided by Jira, we need to add a Distinct at the top level of the Project, and perhaps we can also add test cases to JdbcTest to ensure that our transformation implementation is correct.
The SQL example:
```
select * from (values (1, 2), (1, 2), (3, 4), (5, 7)) as t(a, b)
intersect 
select * from (values (1, 2), (3, 4), (5, 6)) as t(a, b) ;
```
You will know why we need the Distinct.

xiedeyantu:
> According to the description provided by Jira, we need to add a Distinct at the top level of the Project, and perhaps we can also add test cases to JdbcTest to ensure that our transformation implementation is correct. The SQL example:
> 
> ```
> select * from (values (1, 2), (1, 2), (3, 4), (5, 7)) as t(a, b)
> intersect 
> select * from (values (1, 2), (3, 4), (5, 6)) as t(a, b) ;
> ```
> 
> You will know why we need the Distinct.

@NobiGo Thanks, I got it, I will update it.

xiedeyantu:
> According to the description provided by Jira, we need to add a Distinct at the top level of the Project, and perhaps we can also add test cases to JdbcTest to ensure that our transformation implementation is correct. The SQL example:
> 
> ```
> select * from (values (1, 2), (1, 2), (3, 4), (5, 7)) as t(a, b)
> intersect 
> select * from (values (1, 2), (3, 4), (5, 6)) as t(a, b) ;
> ```
> 
> You will know why we need the Distinct.

@NobiGo I know the cause of the problem. I considered distinct in the rule, but the field I chose was too special. It was the primary key of the table, so agg was automatically removed. I modified the test case again, making all the current test cases use non-primary keys, and then adding a test case using the primary key field.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -9585,4 +9585,46 @@ private void checkJoinAssociateRuleWithTopAlwaysTrueCondition(boolean allowAlway
         .withRule(CoreRules.MULTI_JOIN_OPTIMIZE)
         .check();
   }
+
+  @Test void testIntersectToExistsRuleOneField() {
```
mihaibudiu, line 9589: since plans are hard to read, I would encourage you to also add some quidem tests, these are much simpler to read and write

xiedeyantu: > since plans are hard to read, I would encourage you to also add some quidem tests, these are much simpler to read and write

@mihaibudiu Could you give me a description of quidem tests? I don't quite understand how to write them now.

mihaibudiu: Look at the files with iq suffix

xiedeyantu: > Look at the files with iq suffix

@mihaibudiu Thanks, I have seen the file with this suffix. Do I need to write a SQL test to output the final result through the optimization rules? Is there a similar example for reference?

mihaibudiu:
Actually I think I am wrong, there is no way right now to use a specific optimization in a quidem file. That ticket is still open. So you can for now forget about my request.

xiedeyantu:
> Actually I think I am wrong, there is no way right now to use a specific optimization in a quidem file. That ticket is still open. So you can for now forget about my request.

It doesn't matter, I can just learn this testing method, which may be covered later.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -5440,6 +5440,143 @@ LogicalIntersect(all=[true])
...
+LogicalIntersect(all=[false])
+  LogicalProject(ENAME=[$1], JOB=[$2])
+    LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+  LogicalProject(ENAME=[$1], JOB=[$2])
+    LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+]]>
+    </Resource>
+    <Resource name="planAfter">
+      <![CDATA[
+LogicalAggregate(group=[{0, 1}])
```
suibianwanwank, line 5463: I have not read this paper for the time being. Is it necessary to consider the case of null values here. For example, if the emp table has only one null row. in intersect it should return null, in exsits it will return 0 rows of data.

xiedeyantu: @suibianwanwank This is indeed a good question. Many database implementations are as you said, so do we need to modify the logic?
`SELECT a.CustID FROM tbl1 AS a
INTERSECT
SELECT b.CustID FROM tbl2 AS b `
The Exists SQL change to
`SELECT DISTINCT a.CustID FROM tbl1 AS a 
WHERE EXISTS (SELECT b.CustID FROM tbl2 b WHERE (a.CustID=b.CustID) or (a.CustID is NULL and b.CustID is NULL))`

@NobiGo Could you help me determine whether this logic is correct?

asolimando: Haven't read the paper either, but the specific rewrite you propose LGTM, even in the case @suibianwanwank proposed.

You can even opt for a more concise (equivalent) rewrite:
```
SELECT DISTINCT a.CustID FROM tbl1 AS a WHERE EXISTS (SELECT b.CustID FROM tbl2 b WHERE a.CustID IS NOT DISTINCT FROM b.CustID)
```

I have played a bit with it here to double check my understanding: https://www.db-fiddle.com/#&togetherjs=NSeyJDgqqy

suibianwanwank: That looks great!

xiedeyantu: > IS NOT DISTINCT FROM

@asolimando Great! I will chang my rule to fit your suggestion.

xiedeyantu:
@asolimando @suibianwanwank @NobiGo I have fixed, please review it.

core/src/main/java/org/apache/calcite/rel/rules/IntersectToExistsRule.java
```diff
@@ -0,0 +1,146 @@
...
+    final Intersect intersect = call.rel(0);
+    if (intersect.all) {
+      return; // nothing we can do
+    }
+
+    final RelBuilder builder = call.builder();
+    final RexBuilder rexBuilder = builder.getRexBuilder();
+
+    // get all column indices of intersect
+    List<Integer> fieldIndices = intersect.getRowType().getFieldList()
```
suibianwanwank, line 81: Nit: Here you can use an ImmutableBitSet to store.

core/src/main/java/org/apache/calcite/rel/rules/IntersectToExistsRule.java
```diff
@@ -0,0 +1,146 @@
...
+          .project(builder.fields(fieldIndices))
+          .build();
+
+      // apply exists subquery to the current relation
+      current = builder.push(current)
+          .filter(ImmutableSet.of(correlationId),
+              RexSubQuery.exists(existsSubQuery))
+          .build();
+    }
+    System.out.println(RelOptUtil.toString(current));
```
suibianwanwank, line 121: The output during debugging should be deleted, as should the following.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -9585,4 +9585,46 @@ private void checkJoinAssociateRuleWithTopAlwaysTrueCondition(boolean allowAlway
...
+    String sql = "SELECT a.ename FROM emp AS a\n"
+            + "INTERSECT\n"
+            + "SELECT b.name FROM dept AS b";
+    sql(sql).withRule(CoreRules.INTERSECT_TO_EXISTS)
+            .check();
+  }
+
+  @Test void testIntersectToExistsRulePrimaryKey() {
```
suibianwanwank, line 9597: The test should have a jira link added to it, so that people who see this test in the future can more easily understand the purpose of this test.

suibianwanwank (changes requested):

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
NobiGo: Just as @mihaibudiu  said, currently this quidem file cannot specify specific optimizations. Before Quidem's test, we wrote such test cases in JdbcTest. Please feel free to check it.

NobiGo (commented):
The issue raised by @suibianwanwank  is a very good test case. We should test the accuracy of this logic in the JdbcTest.

xiedeyantu:
> The issue raised by @suibianwanwank is a very good test case. We should test the accuracy of this logic in the JdbcTest.

@NobiGo Sorry, I tried many ways and couldn't find a good jdbc test method. I found that when executing the test, the plan should not include subquery, which will cause this error
```
  public RexLocalRef registerInput(RexNode expr) {
    final RexShuttle shuttle = new RegisterInputShuttle(true);
    final RexNode ref = expr.accept(shuttle);
    return (RexLocalRef) ref;  <------ ref is Subquery, can not cast to RexLocalRef
  }
```

NobiGo (commented):
```
 @Test void testIntersect() {
    final String sql = ""
        + "select \"empid\", \"name\" from \"hr\".\"emps\" where \"deptno\"=10\n"
        + "intersect\n"
        + "select \"empid\", \"name\" from \"hr\".\"emps\" where \"empid\">=150";
    CalciteAssert.hr()
        .query(sql)
        .withHook(Hook.PLANNER, (Consumer<RelOptPlanner>)
            planner -> {
          planner.removeRule(CoreRules.INTERSECT_TO_DISTINCT);
          planner.removeRule(EnumerableRules.ENUMERABLE_INTERSECT_RULE);
          planner.addRule(CoreRules.INTERSECT_TO_EXISTS);
            })
        .explainContains("")
  }
```
You can try this unit test in JdbcTest.
Now it throws exception.

xiedeyantu:
@NobiGo I have added a test case for SQL to Rel and Rel to SQL. Can this verify the conclusion? As for the execution result of this SQL, it depends on various execution engines.

xiedeyantu:
@NobiGo @suibianwanwank @mihaibudiu Hi all, Do we still need this PR？

NobiGo:
@xiedeyantu This PR cannot be added with test cases in JdbcTest due to hidden issues. I have been dealing with them and we can merge this PR in the next version.

xiedeyantu:
> @xiedeyantu This PR cannot be added with test cases in JdbcTest due to hidden issues. I have been dealing with them and we can merge this PR in the next version.

@NobiGo Got it, thanks!

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -9716,4 +9716,50 @@ private void checkJoinAssociateRuleWithTopAlwaysTrueCondition(boolean allowAlway
...
+    String sql = "SELECT a.empno FROM emp AS a\n"
+        + "INTERSECT\n"
+        + "SELECT b.empno FROM emp AS b";
+    sql(sql).withRule(CoreRules.INTERSECT_TO_EXISTS)
+        .check();
+  }
+
```
suibianwanwank, line 9739: For all remaining tests, also add the jira case

core/src/main/java/org/apache/calcite/rel/rules/IntersectToExistsRule.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    builder.push(current);
+    List<RexNode> projects = new ArrayList<>();
+    for (int fieldIndex : fieldIndices) {
+      RexNode rexNode = builder.fields().get(fieldIndex);
+      RelDataType originalType =
+          rowType.getFieldList().get(projects.size()).getType();
+      RexNode expr;
+      if (!originalType.equals(rexNode.getType())) {
```
suibianwanwank, line 131: What situation is the type adjustment for here?

core/src/main/java/org/apache/calcite/rel/rules/IntersectToExistsRule.java
```diff
@@ -0,0 +1,179 @@
...
+        .build();
+
+    call.transformTo(result);
+  }
+
+  /**
+   * Will not eliminate HepRelVertex in subqueries
+   * during the optimization process.
+   * We need to eliminate it here. */
+  private RelNode removeHepRelVertex(RelNode input) {
```
suibianwanwank, line 150: You can learn about stripped().

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -244,6 +245,33 @@ private static String toSql(RelNode root, SqlDialect dialect,
     sql(query).withMysql().ok(expected);
   }
 
+  /**
+   * Test case of
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6836">[CALCITE-6836]
+   * Add Rule to convert INTERSECT to EXISTS</a>. */
+  @Test void testIntersectToExistsRule() {
+    String query = "SELECT \"product_name\"\n"
```
suibianwanwank, line 253: In fact, I don't think it's necessary to test in RelToSqlConverter because the semantics of relational algebra is already expressed in RelNode. What is tested here is the correctness of RelNode -> sqlnode -> sqlText.

suibianwanwank (commented):
Thanks for the revisions. There were a few minor issues I left comments on. 
I'll try to test this pr locally this weekend, it would be great to find a way to test the results.
Also, you don't have to overwrite previous commits every time, and keeping a record of changes will allow other reviewers to better understand pr changes.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xiedeyantu: > In fact, I don't think it's necessary to test in RelToSqlConverter because the semantics of relational algebra is already expressed in RelNode. What is tested here is the correctness of RelNode -> sqlnode -> sqlText.

@suibianwanwank Thanks, I agree with you, but currently Calcite may not have the ability to execute this plan, and I have no other good options.

core/src/main/java/org/apache/calcite/rel/rules/IntersectToExistsRule.java
xiedeyantu: > What situation is the type adjustment for here?
Here we can get java type and sql type, when we using setop.

core/src/main/java/org/apache/calcite/rel/rules/IntersectToExistsRule.java
xiedeyantu: > You may need to stipped().

Yes, thanks.

suibianwanwank (approved):

NobiGo (approved):

PR #4210, opened 2025-02-23
[CALCITE-6853] Nested window aggregate throws UnsupportedOperationExc…

fafacao86:
JIRA details: https://issues.apache.org/jira/browse/CALCITE-6853

I encountered this bug when running TPC-DS queries with the default validatorConfig.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -2246,6 +2246,7 @@ public class SqlToRelConverter {
 
   private RexNode convertOver(Blackboard bb, SqlNode node) {
     SqlCall call = (SqlCall) node;
+    bb.getValidator().deriveType(bb.scope, call);
     SqlCall aggCall = call.operand(0);
     boolean ignoreNulls = false;
     switch (aggCall.getKind()) {
--- a/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
@@ -45,6 +45,7 @@ import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.validate.SqlConformance;
 import org.apache.calcite.sql.validate.SqlConformanceEnum;
 import org.apache.calcite.sql.validate.SqlDelegatingConformance;
+import org.apache.calcite.sql.validate.SqlValidatorUtil;
 import org.apache.calcite.test.catalog.MockCatalogReaderExtended;
 import org.apache.calcite.util.Bug;
 import org.apache.calcite.util.TestUtil;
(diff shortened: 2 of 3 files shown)


core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -5388,6 +5390,23 @@ void checkUserDefinedOrderByOver(NullCollation nullCollation) {
         .ok();
   }
 
+  @Test void testNestedWindowAggWithIdentifierExpansionDisabled() {
+    String sql = "select sum(sum(sal)) over() from emp";
+    sql(sql)
+        .withFactory(f ->
+            f.withValidator((opTab, catalogReader, typeFactory, config) -> {
+              if (config.conformance().allowGeometry()) {
```
mihaibudiu, line 5398: what is the connection with the geometry?

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -2328,6 +2328,7 @@ private RexNode convertOver(Blackboard bb, SqlNode node) {
           "already in window agg mode");
       bb.window = window;
       RexNode rexAgg = exprConverter.convertCall(bb, aggCall);
+      bb.getValidator().deriveType(bb.scope, call);
```
mihaibudiu, line 2331: a little higher up the code uses `validator().deriveType()`

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
fafacao86: Well, I copy this _withFactory_ from org.apache.calcite.test.SqlToRelFixture#DEFAULT, line 49, to make sure that only the identifierExpansion default option gets over-written to **false**, and the _geometry_ remains the same with the default.  

When I was trying to write a test for this bug, I couldn't find a relatively elegant way to only overwrite the identifierExpansion option. I was trying to use the way in org.apache.calcite.test#testFunctionExprInOver() function to overwrite the identifierExpansion option to false, but you can check out the code, it's actually **incorrect to do it that way**, the identifierExpansion will get set to true again by the _withFactory atorg.apache.calcite.test.SqlToRelFixture#DEFAULT line 49_. So I have to overwrite the _withFactory_ transformation function.

mihaibudiu: still, why does this test need the geometry to be set up?

fafacao86: This test has nothing to do with the geometry settings, I just want to keep it the same with the default settings in calcite test fixture. I will delete this geometry setting in this test case.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
fafacao86: I will move it up to the beginning of the convertOver() function

mihaibudiu (approved):

PR #4212, opened 2025-02-25
[CALCITE-6837] Invalid code generated for ROW_NUMBER function in Enum…

NobiGo:
…erable convention


core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcToEnumerableConverter.java
```diff
@@ -155,22 +162,53 @@ protected JdbcToEnumerableConverter(
     }
     final ParameterExpression e_ =
         Expressions.parameter(SQLException.class, builder.newName("e"));
+    BlockStatement valueBlock;
+    if (fieldCount == 0) {
+      // For some queries, we don't need to rely on a specific column value in the data source,
+      // we only need to know the number of rows in the table.
+      // For example: "select row_number() over () from dept"
+      // we can't push down the "row_number() over ()"
+      // So the generate code should be:
```
mihaibudiu, line 171: generated

mihaibudiu (approved):

PR #4213, opened 2025-02-25
Add JOIN test for MongoDB adapter

xuzifu666:
Currently MongoDB Adapter not have a test for join, this is nessary because other adapter may had bug in join such as elastic search( https://issues.apache.org/jira/browse/CALCITE-6274 )

### DIFF
--- a/mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
+++ b/mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
@@ -217,6 +217,27 @@ public class MongoAdapterTest implements SchemaFactory {
                 "{$project: {STATE: '$state', ID: '$_id'}}"));
   }
 
+  @Test void testJoin() {
+    assertModel(MODEL)
+        .query("select b.state, a.id from zips as a join zips as b on a.id=b.id where a.id='02401' "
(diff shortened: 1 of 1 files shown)


caicancai:
Adding tests does not require creating a Jira case

caicancai (approved):

xuzifu666:
> Adding tests does not require creating a Jira case

OK, I close jira first.

mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
```diff
@@ -217,6 +217,17 @@ private CalciteAssert.AssertThat assertModel(URL url) {
                 "{$project: {STATE: '$state', ID: '$_id'}}"));
   }
 
+  @Test void testJoin() {
+    assertModel(MODEL)
+        .query("select b.state, a.id from zips as a join zips as b on a.id=b.id where a.id='02401' "
```
NobiGo, line 222: We should add another test case so that we can make a comparison. For instance:
`select b.state, a.id from zips as a join zips as b on a.id=b.id`

xuzifu666: Done，add the case. @NobiGo

PR #4214, opened 2025-02-26
[CALCITE-6835] Invalid unparse for IS TRUE,IS FALSE,IS NOT TRUE and I…

NobiGo:
…S NOT FALSE in StarRocksDialect

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -166,8 +166,12 @@ public abstract class SqlImplementor {
 
   protected final Map<CorrelationId, Context> correlTableMap = new HashMap<>();
 
-  /** Private RexBuilder for short-lived expressions. It has its own
-   * dedicated type factory, so don't trust the types to be canonized. */
+  /**
+   * Private RexBuilder for short-lived expressions. It has its own
+   * dedicated type factory, so don't trust the types to be canonized.
+   *
+   * @deprecated Replaced by {@link RexBuilder#DEFAULT}. */
+  @Deprecated // to be removed before 2.0
   final RexBuilder rexBuilder =
       new RexBuilder(new SqlTypeFactoryImpl(RelDataTypeSystemImpl.DEFAULT));
 
@@ -652,6 +656,7 @@ public abstract class SqlImplementor {
      * @param rex Expression to convert
      */
     public SqlNode toSql(@Nullable RexProgram program, RexNode rex) {
+      rex = dialect.prepareUnparse(rex);
       final RexSubQuery subQuery;
       final SqlNode sqlSubQuery;
       final RexLiteral literal;
@@ -798,7 +803,7 @@ public abstract class SqlImplementor {
           //noinspection unchecked
           return toSql(program, search.operands.get(0), literal.getType(), sarg);
         }
-        return toSql(program, RexUtil.expandSearch(implementor().rexBuilder, program, search));
+        return toSql(program, RexUtil.expandSearch(RexBuilder.DEFAULT, program, search));
 
       case EXISTS:
       case UNIQUE:
@@ -953,7 +958,7 @@ public abstract class SqlImplementor {
         final RangeSets.Consumer<C> consumer =
             new RangeToSql<>(operandSql, orList, v ->
                 toSql(program,
-                    implementor().rexBuilder.makeLiteral(v, type)));
+                    RexBuilder.DEFAULT.makeLiteral(v, type)));
         RangeSets.forEach(sarg.rangeSet, consumer);
       }
       return SqlUtil.createCall(SqlStdOperatorTable.OR, POS, orList);
@@ -965,7 +970,7 @@ public abstract class SqlImplementor {
       final SqlNodeList list = rangeSet.asRanges().stream()
           .map(range ->
               toSql(program,
-                  implementor().rexBuilder.makeLiteral(range.lowerEndpoint(),
+                  RexBuilder.DEFAULT.makeLiteral(range.lowerEndpoint(),
                       type, true, true)))
           .collect(SqlNode.toList());
       switch (list.size()) {
@@ -1718,7 +1723,7 @@ public abstract class SqlImplementor {
             && ((RexInputRef) op0).getIndex() >= leftContext.fieldCount) {
           // Arguments were of form 'op1 = op0'
           final SqlOperator op2 = requireNonNull(call.getOperator().reverse());
-          return (RexCall) rexBuilder.makeCall(call.getParserPosition(), op2, op1, op0);
+          return (RexCall) RexBuilder.DEFAULT.makeCall(call.getParserPosition(), op2, op1, op0);
         }
         // fall through
       default:
--- a/core/src/main/java/org/apache/calcite/sql/SqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlDialect.java
@@ -795,6 +795,11 @@ public class SqlDialect {
     return true;
   }
 
+  /** Converts {@link RexNode} expression to {@link RexNode} expression before unparse. */
+  public RexNode prepareUnparse(RexNode rexNode) {
+    return rexNode;
+  }
+
   /** Returns whether this dialect supports a given function or operator.
    * It only applies to built-in scalar functions and operators, since
    * user-defined functions and procedures should be read by JdbcSchema. */
--- a/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
@@ -21,6 +21,10 @@ import org.apache.calcite.config.NullCollation;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
 import org.apache.calcite.rel.type.RelDataTypeSystemImpl;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlAbstractDateTimeLiteral;
 import org.apache.calcite.sql.SqlAlienSystemTypeNameSpec;
 import org.apache.calcite.sql.SqlBasicTypeNameSpec;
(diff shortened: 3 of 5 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -128,6 +132,52 @@ public StarRocksSqlDialect(Context context) {
               timeUnitNode.getParserPosition());
       SqlFloorFunction.unparseDatetimeFunction(writer, newCall, "DATE_TRUNC", false);
       break;
+    case IS_TRUE:
```
mihaibudiu, line 135: this won't work if the expression has side effects, e.g., the function is RAND().

NobiGo: Good catch. I overlooked this non-deterministic expression. If this expression is non-deterministic, I will throw an exception. Or do you have any other more general approaches for handling this situation?

mihaibudiu: The only solution I can think of is more complicated: you need to add a new project saving the value of the expression, and then do the expansion for that value.

NobiGo: Ok, I understand your approach. For this PR, I will throw an exception for non-deterministic expressions.

NobiGo: Perhaps we'll consider your suggestion later.

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
```diff
@@ -795,6 +795,14 @@ public boolean supportBooleanCaseWhen() {
     return true;
   }
 
+  /**
+   * Returns whether to enable unparse of "macro-like" calls such as IS TRUE.
+   * "A IS TRUE" can be unparse to "A IS NOT NULL AND A" When A is deterministic.
+   * Otherwise, inconsistent results may occur. */
+  public boolean supportMacroLikeUnparse() {
```
mihaibudiu, line 802: I don't understand how this fixes the problem of non-deterministic functions.

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -128,6 +132,72 @@ public StarRocksSqlDialect(Context context) {
...
+      if (this.supportMacroLikeUnparse()) {
```
mihaibudiu, line 136: how about having a conservative approximation?
for example, you can have a shuttle which checks whether operand(0) invokes a non-deterministic function.
You can approximate that by "calls any function" for now.
Or the dialect could have a list of non-deterministic function names, there aren't that many.

mihaibudiu: BTW: I have the impression that there are other places in Calcite where expressions are copied without concern for deterministic behavior.

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
NobiGo: This has not been fixed. It would be more accurate to say that it is a kind of compatible approach, which refers to SqlValidator#withCallRewrite.

mihaibudiu: I think this is not really necessary then.

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
NobiGo: My current approach is indeed a bit strange. Since this might not be within the purview of SqlDialect, in StarRocksSqlDialect, it is not certain whether this function is deterministic(SqlCall). If we make this judgment in SqlImplementor, we need a list of unsupported functions to determine whether to rewrite it. This list is used to write functions that Dialect does not support and which might be non-deterministic. WDYT?

core/src/main/java/org/apache/calcite/sql/SqlDialect.java
NobiGo: Ok. This approach is somewhat inappropriate.

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
mihaibudiu: This is probably not the only place where a rewrite could be wrong if a call is not deterministic. I think it would be a useful feature to be able to detect such calls.

NobiGo: We already have RexUtil#isDeterministic for detecting whether a function is deterministic. Currently, what I'm thinking is that only when there is a transformation similar to "IS TRUE", where a reference becomes multiple times, will it lead to inconsistent results. Functions like COALESCE and NULLIF fall into this category.

mihaibudiu: Right

suibianwanwank: I agree, any scenario requiring splitting a rexNode into multiple processing paths would encounter this problem. But it seems difficult to resolve at the root cause - the only solution I can think of is to specially handle it by adding an additional Project whenever this situation occurs.

NobiGo: I've revised the PR according to the feedback. When you have time, please review the updates.

core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
```diff
@@ -93,6 +94,18 @@ public StarRocksSqlDialect(Context context) {
     return true;
   }
 
+  @Override public boolean supportsUnparseOnlyDeterministic(SqlOperator sqlOperator) {
```
mihaibudiu, line 97: I am not sure this function is useful.
If the function returns 'false' and you cannot unparse, you will need to throw.
If you remove the function, you still throw in the same places.

NobiGo: If we do not support rewriting this function, we should throw exceptions at the point of rewriting. We directly convert it because only Starrocks has been identified to have this issue. If it is discovered later that operations like A IS TRUE and A IS NOT NULL are also unsupported, exceptions must be thrown during the conversion process. This conversion logic can be implemented as part of the Dialect behavior.
If I remove this function, the A IS FALSE will remain unchanged and not be converted. 
What I'm thinking here is that not all dialects have undergone such transformation. If the database supports A IS TRUE, then we won't make any changes.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -855,6 +855,76 @@ private SqlNode callToSql(@Nullable RexProgram program, RexCall call0,
       final RexCall call = (RexCall) stripCastFromString(call1, dialect);
       SqlOperator op = call.getOperator();
       switch (op.getKind()) {
+      case IS_FALSE:
```
mihaibudiu, line 858: Can this be done in unparseCall instead?
Then you don't need the new policy function.

NobiGo: If we use unparseCall, we need to parse the SqlCall. We can't extract whether it is deterministic from SqlCall. So, I put it here.

mihaibudiu: true.
I don't really like this function "supportsUnparseOnlyDeterministic()", because it really has nothing to do with IS_TRUE, etc. But I don't have a better proposal on how to structure this.

NobiGo: I have also been pondering over this for a long time. How can I accurately express it so that the developers can quickly figure out which functions should be added here and which ones should use unparseCall instead?
How about `unparseCallPartial` The meaning is that this function "unparse" is supported in certain scenarios, not in all cases.

mihaibudiu: The most generic way to handle this would be to add a hook in the dialect to rewrite calls before they are converted to sql. This hook would be called from callToSql prior to converting to SqlNode.

NobiGo: I understand. I will try to do it this way.

mihaibudiu: the only misgivings I have about that is that the Dialect class is mostly about SqlNode.
But it does have a method that takes a RexNode

NobiGo: Adding RexNode parsing to Dialect might render the current unparseCall ineffective. After all, RexNode offers greater flexibility.

NobiGo: Or we could change "supportsUnparseOnlyDeterministic" in the current PR to "supportsUnparsePartial" and mark it as @Experimental. Then we'll see if it's necessary to add the parsing of RexNode to Dialect later.

mihaibudiu: I am thinking of a function in Dialect like this: `RexCall prepareUnparse(RexCall arg)` which can rewrite a RexCall before it is unparsed; the default implementation returns arg. The argument could be RexNode too.

mihaibudiu (approved):
This is actually pretty clean.
I hope other people will like it too.
This hook is very powerful.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -166,8 +166,12 @@ public abstract class SqlImplementor {
...
   protected final Map<CorrelationId, Context> correlTableMap = new HashMap<>();
 
-  /** Private RexBuilder for short-lived expressions. It has its own
-   * dedicated type factory, so don't trust the types to be canonized. */
+  /**
+   * Private RexBuilder for short-lived expressions. It has its own
+   * dedicated type factory, so don't trust the types to be canonized.
+   *
+   * @deprecated Replaced by {@link RexBuilder#DEFAULT}. */
+  @Deprecated // to be removed before 2.0
```
mihaibudiu, line 174: I am not sure about this change, though

NobiGo: This change is necessary. I tried to obtain `RexBuilder` through `implementer().RexBuilder`, but SimpleContext may throw UnsupportedOperaException in some cases.

mihaibudiu: Well, I approved the PR.
But it would be good to know why this method needs to be essentially deprecated (because it can't be used).

NobiGo: I am concerned that other projects have implemented this interface and obtained this RexBuilder instance through `implementer(). RexBuilder`.We don't need it anymore. To maintain compatibility, I added @ Deprecated.

PR #4219, opened 2025-02-27
[CALCITE-1221] Implement DESCRIBE DATABASE, CATALOG, STATEMENT

maxzhuravkov:
Adds `SqlNode`s for DescribeDatabase, DescribeCatalog, and DescribeStatement.

JIRA: https://issues.apache.org/jira/browse/CALCITE-1221

### DIFF
--- a/core/src/main/codegen/templates/Parser.jj
+++ b/core/src/main/codegen/templates/Parser.jj
@@ -49,7 +49,10 @@ import org.apache.calcite.sql.SqlCollation;
 import org.apache.calcite.sql.SqlCollectionTypeNameSpec;
 import org.apache.calcite.sql.SqlDataTypeSpec;
 import org.apache.calcite.sql.SqlDelete;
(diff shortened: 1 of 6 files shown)


mihaibudiu (commented):
See the discussion in JIRA, two people suggested that this belongs to the server component

core/src/main/codegen/templates/Parser.jj
```diff
@@ -1518,11 +1521,18 @@ SqlNode SqlDescribe() :
 {
     <DESCRIBE> { s = span(); }
     (
-        LOOKAHEAD(2) (<DATABASE> | <CATALOG> | <SCHEMA>)
+        <DATABASE>
```
xuzifu666, line 1524: What is the plan next step?  Can resolve the issue by add Describe DABASE/CATALOG/STATEMENT to server module parserImpls.ftl which would generate code in SqlDdlParserImpl.

PR #4226, opened 2025-03-05
[CALCITE-6873] FilterProjectTransposeRule should not push the Filter …

NobiGo:
…past the Project when the Filter contains a Subquery with correlation

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexUtil.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexUtil.java
@@ -2826,6 +2826,21 @@ public class RexUtil {
     @Override public Void visitCorrelVariable(RexCorrelVariable var) {
       throw Util.FoundOne.NULL;
     }
+
+    @Override public Void visitSubQuery(RexSubQuery subQuery) {
+      if (!deep) {
+        return null;
+      }
+
+      for (RexNode operand : subQuery.operands) {
+        operand.accept(this);
+      }
+
+      if (!RelOptUtil.getVariablesUsed(subQuery.rel).isEmpty()) {
+        throw Util.FoundOne.NULL;
+      }
+      return null;
+    }
   }
 
   /** Shuttle that fixes up an expression to match changes in nullability of
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -522,6 +522,28 @@ class RelOptRulesTest extends RelOptTestBase {
   }
 
   /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6873">[CALCITE-6873]
+   * FilterProjectTransposeRule should not push the Filter past the Project
+   * when the Filter contains a Subquery with correlation</a>. */
+  @Test void testFilterProjectTransposeRule2() {
+    final String sql = "select * from (select deptno from emp) as d\n"
+        + "where NOT EXISTS (\n"
+        + "  select count(*) from emp e where e.deptno = d.deptno)";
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/apache/calcite/rel/rules/FilterProjectTransposeRule.java
```diff
@@ -213,7 +213,8 @@ protected FilterProjectTransposeRule(
   public interface Config extends RelRule.Config {
     Config DEFAULT = ImmutableFilterProjectTransposeRule.Config.of()
         .withOperandFor(Filter.class,
-            f -> !RexUtil.containsCorrelation(f.getCondition()),
+            f -> !RexUtil.containsCorrelation(f.getCondition())
+                && !RexUtil.SubQueryFinder.containsSubQuery(f),
```
zabetak, line 217: What about uncorrelated sub-query?

NobiGo: Currently, The rule can't push down uncorrelated sub-queries either, because this can be achieved by removing the sub-query through other plans. Even if the sub-query is not removed, the subsequent plans can't be pushed down and will still be fine. 
Correlated sub-queries definitely cannot be pushed down. If pushed down, problems will occur in the subsequent plans. Therefore, here I did not differentiate whether the specific sub-queries contain correlation or not.

asolimando: Shouldn't we be providing a comprehensive function that handles both and use it here then?

NobiGo: @asolimando @zabetak Based on the above discussion, I have enhanced RexUtil.containsCorrelation. If the test goes well, I will update the PR again. Maybe we can only push down the non-correlated sub-queries.

asolimando: Thanks @NobiGo, I feel like it's the right approach since all the callers of that function will benefit from this enhancement in a similar way.

core/src/main/java/org/apache/calcite/rex/RexUtil.java
```diff
@@ -2826,6 +2826,21 @@ private CorrelationFinder() {
...
+      if (!deep) {
+        return null;
+      }
+
+      for (RexNode operand : subQuery.operands) {
+        operand.accept(this);
+      }
+
+      if (!RelOptUtil.getVariablesUsed(subQuery.rel).isEmpty()) {
+        throw Util.FoundOne.NULL;
```
zabetak, line 2840: From a quick look it seems that `RelOptUtil.getVariablesUsed` on its own does exactly what you want. There is no need to call it from here. 

In fact, there is also `org.apache.calcite.plan.RelOptUtil#notContainsCorrelation` which might be exactly what you are looking for.

NobiGo: `RelOptUtil.getVariablesUsed` can collect all the variables used in the Filter operator, including those from the input plan and the filter's Condition. Here, we only need to focus on the variables within the Condition.
`RelOptUtil.notContainsCorrelation` determines whether a specific correlation is included. Here, we need to check whether a correlation exists.
Did I misunderstand something?

zabetak: My bad, I rushed a bit through the code and didn't realize that it was a full-blown RelNode visitor. The changes here do make sense now.

zabetak (approved):

asolimando (approved):
LGTM, thanks @NobiGo!

PR #4228, opened 2025-03-07
[CALCITE-2115] Remove not generated mongo doc, and popular sales_fact_1997/1998 data

tuichenchuxin:
- Since the implementation of group distinct is implemented through mongo API, group, etc. are deleted in the statement.
- sales_fact_1997， sales_fact_1998 has been removed from the database. So can't be found in schema. I think these test for sales_fact_1997, sales_fact_1998 can be removed.

https://issues.apache.org/jira/browse/CALCITE-2115

### DIFF
--- a/mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
+++ b/mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
@@ -104,6 +104,12 @@ public class MongoAdapterTest implements SchemaFactory {
     populate(database.getCollection("warehouse"),
         requireNonNull(FoodmartJson.class.getResource("/warehouse.json"),
             "url"));
+    populate(database.getCollection("sales_fact_1997"),
+            requireNonNull(FoodmartJson.class.getResource("/sales_fact_1997.json"),
+             "url"));
+    populate(database.getCollection("sales_fact_1998"),
+            requireNonNull(FoodmartJson.class.getResource("/sales_fact_1998.json"),
+             "url"));
 
     // Manually insert data for data-time test.
     MongoCollection<BsonDocument> datatypes = database.getCollection("datatypes")
@@ -284,7 +290,6 @@ public class MongoAdapterTest implements SchemaFactory {
             "CITY=LAWTON; LONGITUDE=null; LATITUDE=null; POP=45542; STATE=OK; ID=73505");
   }
 
-  @Disabled("broken; [CALCITE-2115] is logged to fix it")
   @Test void testUnionPlan() {
     assertModel(MODEL)
         .query("select * from \"sales_fact_1997\"\n"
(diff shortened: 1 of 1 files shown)


tuichenchuxin:
> question is, what version of MongoDB is Calcite based on? Do we need to explain it?

Thank you for review.
I think we need to explain what version we can support. for example 4.2 + can be supported.
I think the latest version should be compatible with the old mongodb version. So we just need to confirm the minimum supported version.

mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
```diff
@@ -284,37 +284,6 @@ private CalciteAssert.AssertThat assertModel(URL url) {
             "CITY=LAWTON; LONGITUDE=null; LATITUDE=null; POP=45542; STATE=OK; ID=73505");
   }
 
-  @Disabled("broken; [CALCITE-2115] is logged to fix it")
```
NobiGo, line 287: I don't understand why we directly deleted the test cases that couldn't run. A better approach should be to make them run as we want.

tuichenchuxin: @NobiGo I found sales_for_1997 is in foodmart-data-json-0.4.jar, so I popular it, then test can run correctly.

mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
```diff
@@ -605,12 +607,7 @@ private CalciteAssert.AssertThat assertModel(URL url) {
                     + "    ]\n"
                     + "  }\n"
                     + "}",
-                "{$project: {CITY: '$city', STATE: '$state'}}",
-                "{$group: {_id: {CITY: '$CITY', STATE: '$STATE'}}}",
-                "{$project: {_id: 0, CITY: '$_id.CITY', STATE: '$_id.STATE'}}",
-                "{$group: {_id: '$STATE', CDC: {$sum: {$cond: [ {$eq: ['CITY', null]}, 0, 1]}}}}",
-                "{$project: {STATE: '$_id', CDC: '$CDC'}}",
-                "{$sort: {STATE: 1}}"));
+                "{$project: {STATE: '$state', CITY: '$city'}}"));
```
NobiGo, line 610: I'm not very familiar with the mongo adapter. However, judging from the plan, the script here seems to have problems. The previous test cases looked correct from the perspective of SQL semantics. But now this plan has lost the `group by` and `order by` clauses. I guess we can check whether the related operations have been completed through Enumerable by adding explainContains.

tuichenchuxin: I found the implementation of group and count distinct is not implemented by mongo statements, but through code and API
https://github.com/apache/calcite/blob/618e601b136e92db933523f77dd7af3c1dfe2779/mongodb/src/main/java/org/apache/calcite/adapter/mongodb/MongoToEnumerableConverter.java#L113-L116

https://github.com/apache/calcite/blob/618e601b136e92db933523f77dd7af3c1dfe2779/mongodb/src/main/java/org/apache/calcite/adapter/mongodb/MongoTable.java#L131-L153

tuichenchuxin: @NobiGo Thank you for your suggestion, I have added the explainContains, and the plan indicates  the related operations have been completed.

PR #4229, opened 2025-03-07
[CALCITE-6874] FilterCalcMergeRule/ProjectCalcMergeRule should not me…

NobiGo:
…rge a Filter/Project to Calc when it contains Subquery


core/src/main/java/org/apache/calcite/rel/rules/FilterCalcMergeRule.java
```diff
@@ -105,7 +106,8 @@ public interface Config extends RelRule.Config {
     default Config withOperandFor(Class<? extends Filter> filterClass,
         Class<? extends Calc> calcClass) {
       return withOperandSupplier(b0 ->
-          b0.operand(filterClass).oneInput(b1 ->
+          b0.operand(filterClass).predicate(filter ->
```
mihaibudiu, line 109: I would personally indent this using a dot on each line, I find that easier to follow.
```
b0.operand()
   .predicate()
   .oneInput()
```

mihaibudiu (approved):

PR #4230, opened 2025-03-07
[CALCITE-6877] Generate LogicalProject in RelRoot.project() when mapping is not name trivial

nielspardon:
See https://issues.apache.org/jira/browse/CALCITE-6877


nielspardon:
> Interesting that this does not affect any existing tests.

Indeed, I was also a bit suprised that this code has been there for 9 years and nobody has encountered this before. 

One thing I found was that when you start with a SQL statement through the SQL parser it combines the column reduction and renaming into a single `LogicalProject` which ends up within the `RelRoot.rel` so through the normal test approach in e.g. `SqlToRelConverterTest` you will not encounter this scenario.

PR #4231, opened 2025-03-07
[CALCITE-6788] LoptOptimizeJoinRule should be able to delegate costs to the planner

rubenada:
A backwards-compatible approach for CALCITE-6788


asolimando (approved):
I like the functional interface approach, pretty neat, I see no harm as the backward compatibility is guaranteed as it clearly emerges from the tests you have added

PR #4233, opened 2025-03-09
[CALCITE-6878] Implement FilterSortTransposeRule

xiedeyantu:
link: [[CALCITE-6878]](https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6878)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/FilterSortTransposeRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/FilterSortTransposeRule.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelOptUtil;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Filter;
+import org.apache.calcite.rel.core.Sort;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.tools.RelBuilder;
+
+import org.immutables.value.Value;
+
+/**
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/rules/FilterSortTransposeRule.java
```diff
@@ -0,0 +1,79 @@
...
+@Value.Enclosing
+public class FilterSortTransposeRule
+    extends RelRule<FilterSortTransposeRule.Config>
+    implements TransformationRule {
+  protected FilterSortTransposeRule(FilterSortTransposeRule.Config config) {
+    super(config);
+  }
+
+  @Deprecated // to be removed before 2.0
+  public FilterSortTransposeRule(RelBuilderFactory relBuilderFactory) {
```
mihaibudiu, line 43: why create functions which are already deprecated?

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/rel/rules/FilterSortTransposeRule.java
xiedeyantu: > why create functions which are already deprecated?

ok, I removed it.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -574,6 +574,45 @@ private static boolean skipItem(RexNode expr) {
...
+        .project(b.field(0))
+        .sortLimit(0, 1, b.field(0))
+        .filter(b.lessThan(b.field(0), b.literal(10)))
+        .build();
+    relFn(relFn).withRule(CoreRules.FILTER_SORT_TRANSPOSE).checkUnchanged();
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6878">[CALCITE-6878]
+   * Add the planner rule that pushes the Filter past a Sort</a>. */
```
NobiGo, line 605: Please ensure the Java documentation is precisely the same as the Jira title.

NobiGo (approved):

core/src/main/java/org/apache/calcite/rel/rules/FilterSortTransposeRule.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /** Rule configuration. */
+  @Value.Immutable
+  public interface Config extends RelRule.Config {
+    Config DEFAULT = ImmutableFilterSortTransposeRule.Config.of()
+        .withOperandSupplier(f ->
+            f.operand(Filter.class)
+                .oneInput(s ->
+                    s.operand(Sort.class).anyInputs()));
```
suibianwanwank, line 66:
```suggestion
                    s.operand(Sort.class).predicate(RelOptUtil::isPureOrder).anyInputs()));
```
Based on this modification, we can remove this code:
```
if (sort.offset != null || sort.fetch != null) {
    return;
}
```

suibianwanwank (approved):
LGTM, Just a minor suggestion leave, feel free to ignore if you prefer the current approach.

core/src/main/java/org/apache/calcite/rel/rules/FilterSortTransposeRule.java
xiedeyantu: > Based on this modification, we can remove this code:
> 
> ```
> if (sort.offset != null || sort.fetch != null) {
>     return;
> }
> ```

Thanks, It's better than mine.

PR #4235, opened 2025-03-10
[CALCITE-6431] Implement the SINGLE_VALUE aggregation in HiveSqlDiale…

NobiGo:
…ct And SparkSQLDialect

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/HiveSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/HiveSqlDialect.java
@@ -19,17 +19,24 @@ package org.apache.calcite.sql.dialect;
 import org.apache.calcite.config.NullCollation;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.SqlAlienSystemTypeNameSpec;
+import org.apache.calcite.sql.SqlBasicCall;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlDataTypeSpec;
 import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlLiteral;
 import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlNodeList;
 import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.SqlSyntax;
 import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.fun.SqlCase;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.type.BasicSqlType;
 import org.apache.calcite.util.RelToSqlConverterUtil;
+
+import com.google.common.collect.ImmutableList;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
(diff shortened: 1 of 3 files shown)


NobiGo:
Before this https://github.com/apache/calcite/pull/3817  Hive and Spark don't throw exceptions as we expected.  It's okay now.

core/src/main/java/org/apache/calcite/sql/dialect/HiveSqlDialect.java
```diff
@@ -166,4 +173,36 @@ public HiveSqlDialect(Context context) {
     }
     return super.getCastSpec(type);
   }
+
+  /**
+   * Rewrite SINGLE_VALUE.
+   *
+   * <pre>In Hive, ASSERT_TRUE(false) will throw assertion failed exception.</pre>
```
mihaibudiu, line 180: I would swap this <pre> with the next one.
It makes it look like you are solving a problem related to ASSERT_TRUE.

core/src/main/java/org/apache/calcite/sql/dialect/HiveSqlDialect.java
```diff
@@ -166,4 +173,36 @@ public HiveSqlDialect(Context context) {
     }
     return super.getCastSpec(type);
   }
+
+  /**
+   * Rewrite SINGLE_VALUE.
```
mihaibudiu, line 178: SINGLE_VALUE(result)

core/src/main/java/org/apache/calcite/sql/dialect/SparkSqlDialect.java
```diff
@@ -152,4 +159,39 @@ public SparkSqlDialect(SqlDialect.Context context) {
...
+   *
+   * <blockquote><pre>
+   * CASE COUNT(*)
+   * WHEN 0 THEN NULL
+   * WHEN 1 THEN MIN(&lt;result&gt;)
+   * ELSE RAISE_ERROR("more than one value in agg SINGLE_VALUE")
+   * </pre></blockquote>
+   */
+  @Override public SqlNode rewriteSingleValueExpr(SqlNode aggCall, RelDataType relDataType) {
+    final SqlNode operand = ((SqlBasicCall) aggCall).operand(0);
```
mihaibudiu, line 177: I don't know if you care to preserve the position of the aggCall in the constructed tree.

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql/dialect/SparkSqlDialect.java
NobiGo: The current conversion here is consistent with other dialect conversions, so no adjustments have been made.

core/src/main/java/org/apache/calcite/sql/dialect/HiveSqlDialect.java
NobiGo: Fixed.

core/src/main/java/org/apache/calcite/sql/dialect/HiveSqlDialect.java
NobiGo: Fixed.

NobiGo:
> hi @NobiGo
> 
> ```sql
> create table correlated_scalar_t1(c1 bigint, c2 bigint);
> create table correlated_scalar_t2(c1 bigint, c2 bigint);
> insert into correlated_scalar_t1 values (1,null),(null,1),(1,2), (null,2),(1,3), (2,4), (2,5), (3,3), (3,4), (20,2), (22,3), (24,4),(null,null);
> insert into correlated_scalar_t2 values (1,null),(null,1),(1,4), (1,2), (null,3), (2,4), (3,7), (3,9),(null,null),(5,1);
> 
> select c1 from correlated_scalar_t1 where correlated_scalar_t1.c2 > (select c1 from correlated_scalar_t2 where correlated_scalar_t1.c1 = correlated_scalar_t2.c1 and correlated_scalar_t2.c2 < 4) order by c1;
> ```
> 
> can you add this case into calicte and check correctness?
> 
> ```sql
> with t1 (c1, c2) as (values (1,null),(null,1),(1,2), (null,2),(1,3), (2,4), (2,5), (3,3), (3,4), (20,2), (22,3), (24,4),(null,null)),  t2 (c1, c2) as (values (1,null),(null,1),(1,4), (1,2), (null,3), (2,4), (3,7), (3,9),(null,null),(5,1)) select c1 from t1 where t1.c2 > (select c1 from t2 where t1.c1 = t2.c1 and t2.c2 < 4) order by c1;
> ```

This is an open-source project. You can freely verify it. If you encounter a new problem, you can register it in Jira.

PR #4236, opened 2025-03-10
[CALCITE-6880] Implement IntersectToSemiJoinRule

xiedeyantu:
link: [CALCITE-6880](https://issues.apache.org/jira/browse/CALCITE-6880)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
@@ -1,0 +1,108 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Intersect;
+import org.apache.calcite.rel.core.JoinRelType;
+import org.apache.calcite.rel.logical.LogicalIntersect;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.tools.RelBuilder;
+
+import org.immutables.value.Value;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Planner rule that translates a {@link org.apache.calcite.rel.core.Intersect}
+ * to a series of {@link org.apache.calcite.rel.core.Join} that type is
+ * {@link org.apache.calcite.rel.core.JoinRelType#SEMI}.
+ */
+@Value.Enclosing
+public class IntersectToSemiJoinRule
+    extends RelRule<IntersectToSemiJoinRule.Config>
+    implements TransformationRule {
+
+  /** Creates an IntersectToSemiJoinRule. */
+  protected IntersectToSemiJoinRule(Config config) {
+    super(config);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    final Intersect intersect = call.rel(0);
+    if (intersect.all) {
+      return; // nothing we can do
+    }
+
+    final RelBuilder builder = call.builder();
+    final RexBuilder rexBuilder = builder.getRexBuilder();
+
+    List<RelNode> inputs = intersect.getInputs();
+    if (inputs.size() != 2) {
+      return;
+    }
+
+    RelNode left = inputs.get(0);
+    RelNode right = inputs.get(1);
+
+    List<RexNode> conditions = new ArrayList<>();
+    int fieldCount = left.getRowType().getFieldCount();
+
+    for (int i = 0; i < fieldCount; i++) {
+      RelDataType leftFieldType = left.getRowType().getFieldList().get(i).getType();
+      RelDataType rightFieldType = right.getRowType().getFieldList().get(i).getType();
+
+      conditions.add(
+          builder.isNotDistinctFrom(
+              rexBuilder.makeInputRef(leftFieldType, i),
+              rexBuilder.makeInputRef(rightFieldType, i + fieldCount)));
+    }
+    RexNode condition = RexUtil.composeConjunction(rexBuilder, conditions);
+
+    builder.push(left)
+        .push(right)
+        .join(JoinRelType.SEMI, condition)
+        .distinct();
+
+    call.transformTo(builder.build());
+  }
+
+  /** Rule configuration. */
+  @Value.Immutable
+  public interface Config extends RelRule.Config {
+    Config DEFAULT = ImmutableIntersectToSemiJoinRule.Config.of()
+        .withOperandFor(LogicalIntersect.class);
+
+    @Override default IntersectToSemiJoinRule toRule() {
+      return new IntersectToSemiJoinRule(this);
+    }
+
+    /** Defines an operand tree for the given classes. */
+    default Config withOperandFor(Class<? extends Intersect> intersectClass) {
+      return withOperandSupplier(b -> b.operand(intersectClass).anyInputs())
+          .as(Config.class);
+    }
+  }
+}
--- a/core/src/test/java/org/apache/calcite/test/JdbcTest.java
+++ b/core/src/test/java/org/apache/calcite/test/JdbcTest.java
@@ -3968,6 +3968,26 @@ public class JdbcTest {
             "commission=null");
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6880">[CALCITE-6880]
+   * Implement IntersectToSemiJoinRule</a>. */
+  @Test void testIntersectToSemiJoin() {
+    final String sql = ""
+        + "select \"commission\" from \"hr\".\"emps\"\n"
+        + "intersect\n"
+        + "select \"commission\" from \"hr\".\"emps\" where \"empid\">=150";
+    CalciteAssert.hr()
+        .query(sql)
+        .withHook(Hook.PLANNER, (Consumer<RelOptPlanner>)
+            planner -> {
+              planner.removeRule(CoreRules.INTERSECT_TO_DISTINCT);
+              planner.removeRule(EnumerableRules.ENUMERABLE_INTERSECT_RULE);
+              planner.addRule(CoreRules.INTERSECT_TO_SEMI_JOIN);
+            })
+        .explainContains("")
+        .returnsUnordered("commission=500",
+            "commission=null");
+  }
+
   @Test void testExcept() {
     final String sql = ""
         + "select \"empid\", \"name\" from \"hr\".\"emps\" where \"deptno\"=10\n"
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -3570,6 +3570,48 @@ class RelOptRulesTest extends RelOptTestBase {
         .check();
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6880">
+   * [CALCITE-6880] Implement IntersectToSemiJoinRule</a>. */
+  @Test void testIntersectToSemiJoin() {
+    final String sql = "select ename from emp where deptno = 10\n"
+        + "intersect\n"
+        + "select ename from emp where deptno = 20\n";
+    sql(sql).withRule(CoreRules.INTERSECT_TO_SEMI_JOIN)
+        .check();
+  }
+
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6880">
+   * [CALCITE-6880] Implement IntersectToSemiJoinRule</a>. */
+  @Test void testIntersectToSemiJoinMultiCol() {
+    final String sql = "select deptno, ename from emp where deptno = 10\n"
+        + "intersect\n"
+        + "select deptno, ename from emp where deptno = 20\n";
+    sql(sql).withRule(CoreRules.INTERSECT_TO_SEMI_JOIN)
+        .check();
+  }
+
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6880">
+   * [CALCITE-6880] Implement IntersectToSemiJoinRule</a>. */
+  @Test void testIntersectToSemiJoinAll() {
+    final String sql = "select ename from emp where deptno = 10\n"
+        + "intersect\n"
+        + "select ename from emp where deptno = 20\n"
+        + "intersect all\n"
+        + "select ename from emp where deptno = 30\n";
+    sql(sql).withRule(CoreRules.INTERSECT_TO_SEMI_JOIN)
+        .check();
+  }
+
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6880">
+   * [CALCITE-6880] Implement IntersectToSemiJoinRule</a>. */
+  @Test void testIntersectToSemiJoinCoercion() {
+    final String sql = "select ename from emp where deptno = 10\n"
+        + "intersect\n"
+        + "select job from emp where deptno = 20\n";
+    sql(sql).withRule(CoreRules.INTERSECT_TO_SEMI_JOIN)
+        .check();
+  }
+
   /** Tests {@link CoreRules#MINUS_MERGE}, which merges 2
    * {@link Minus} operators into a single {@code Minus} with 3
    * inputs. */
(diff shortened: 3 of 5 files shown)


core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
```diff
@@ -0,0 +1,103 @@
...
+    implements TransformationRule {
+
+  /** Creates an IntersectToSemiJoinRule. */
+  protected IntersectToSemiJoinRule(Config config) {
+    super(config);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public void onMatch(RelOptRuleCall call) {
```
mihaibudiu, line 53: I am not sure this is correct for multisets.
The result needs to return a row with the MIN of the two counts.

mihaibudiu: this is not about the number of inputs, it's about INTERSECT ALL

mihaibudiu: I don't even know what the right way to implement INTERSECT ALL is. I suspect you will need some aggregates.
But yes, with joins you can only implement INTERSECT.

xiedeyantu: > I don't even know what the right way to implement INTERSECT ALL is. I suspect you will need some aggregates. But yes, with joins you can only implement INTERSECT.

Yes, we cannot know the ndv of the column, which is an inaccurate statistical information. There may be no better way to implement intersect all. I have changed the code to only support intersect.

core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
```diff
@@ -0,0 +1,108 @@
...
+          rexBuilder.makeInputRef(
+              left.getRowType().getFieldList().get(i).getType(), i),
+          rexBuilder.makeInputRef(
+              right.getRowType().getFieldList().get(i).getType(), i + fieldCount)));
+    }
+    RexNode condition = RexUtil.composeConjunction(rexBuilder, conditions);
+
+    RelNode root = builder.push(left)
+        .push(right)
+        .join(JoinRelType.SEMI, condition)
```
mihaibudiu, line 85: I am not sure this is enough, you need a distinct somewhere too.
Otherwise this will have duplicates in the outputs if there are duplicates in the inputs.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -3527,6 +3527,38 @@ private void checkPushJoinThroughUnionOnRightDoesNotMatchSemiOrAntiJoin(JoinRelT
         .check();
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6880">
+   * [CALCITE-6880] Implement IntersectToSemiJoinRule</a>. */
```
mihaibudiu, line 3531: Please add a test case where the intersected sets do not have exactly the same column types. Coercion should take care of it, but let's make sure.

core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
xiedeyantu: > I am not sure this is enough, you need a distinct somewhere too. Otherwise this will have duplicates in the outputs if there are duplicates in the inputs.

I added distinct after join, which can achieve deduplication. I don't understand your suggestion. Should I add a distinct to each child of join? Can you give me a detailed description?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xiedeyantu: > Please add a test case where the intersected sets do not have exactly the same column types. Coercion should take care of it, but let's make sure.

OK, I will add test for it.

core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
mihaibudiu: one after or one on each input will work I think

xiedeyantu: > one after or one on each input will work I think

I understand that adding a distinct after join can also ensure deduplication.I still don't understand why I need to add distinct to each child of join.

mihaibudiu: I said "or"

xiedeyantu: > I said "or"

Oh, then my way of writing is also OK, right?

mihaibudiu: Maybe I didn't notice the distinct.

xiedeyantu: > Maybe I didn't notice the distinct.

Maybe yes, thanks. I will add another use case for type inconsistency later. Do you have any other suggestions for me?

mihaibudiu: I would have liked some end-to-end tests (inputs->outputs), but that is not as easy; Quidem tests today cannot invoke optimizations. I keep wanting to work on that subject, it would make everything much easier to test.

xiedeyantu: > I would have liked some end-to-end tests (inputs->outputs), but that is not as easy; Quidem tests today cannot invoke optimizations. I keep wanting to work on that subject, it would make everything much easier to test.

I have studied this test mode and feel it is very good. I want to know what stage Quidem tests has reached now and what I can do for this test mode.

mihaibudiu: in general it's very hard to tell what a plan does; in this case, the plans are simple enough that inspecting them by eye is reasonable.

xiedeyantu: > in general it's very hard to tell what a plan does; in this case, the plans are simple enough that inspecting them by eye is reasonable.

I have an idea. Now that Calcite can convert plans into substrait plans and DuckDB can support the execution of substrait plans, can we complete end-to-end testing? I seem to remember that there is a query engine that also uses DuckDB for result verification.

mihaibudiu: This is a very complicated topic; see this recent thread on the mailing list: https://lists.apache.org/thread/zj0f6fdxdt8v6yjqx5nhsnpjotpo2qtq

core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    RelNode left = inputs.get(0);
+    RelNode right = inputs.get(1);
+
+    List<RexNode> conditions = new ArrayList<>();
+    int fieldCount = left.getRowType().getFieldCount();
+
+    // Converting the join condition to
+    // AND(=(COALESCE($0, 0), COALESCE($1, 0)), =(IS NULL($0), IS NULL($0))
```
xiedeyantu, line 76: Use this convertion, we can use HashJoin

core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
```diff
@@ -0,0 +1,139 @@
...
+    builder.push(left)
+        .push(right)
+        .join(JoinRelType.SEMI, condition)
+        .distinct();
+
+    List<RexNode> projects = new ArrayList<>();
+    for (RexNode rexNode : builder.fields()) {
+      RelDataType originalType =
+          oriRowType.getFieldList().get(projects.size()).getType();
+      if (!originalType.equals(rexNode.getType())) {
```
xiedeyantu, line 112: In the setop operator, it is common to encounter a mixture of Java type and SQL type.

xiedeyantu:
@mihaibudiu I have made some changes, please help me review them again.

core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
mihaibudiu: isn't IS NOT DISTINCT FROM the same thing but simpler?

xiedeyantu: > isn't IS NOT DISTINCT FROM the same thing but simpler?

IS NOT DISTINCT FROM is convert to a condition with OR, it won't convert to HashJoin, and result of computing  IS NOT DISTINCT FROM is wrong currently.  We can see https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6904

xiedeyantu: > isn't IS NOT DISTINCT FROM the same thing but simpler?

AND(=(COALESCE($0, 0), COALESCE($1, 0)), =(IS NULL($0), IS NULL($0)) also can be samplier to =($0,$1) when both of them IS NOT NULL now.

mihaibudiu: I don't understand what you are saying.
IS NOT DISTINCT FROM is really equality testing, where NULL is equal only to NULL.
What does this have anything to do with OR?
This stage of the optimizer is completely agnostic to the backend that will be used to generate code, so the issue about enumerable is not relevant.

xiedeyantu: > I don't understand what you are saying. IS NOT DISTINCT FROM is really equality testing, where NULL is equal only to NULL. What does this have anything to do with OR? This stage of the optimizer is completely agnostic to the backend that will be used to generate code, so the issue about enumerable is not relevant.

I mean IS NOT DISTINCT FROM will be samplier to x=y is TRUE OR ((x is null) and (y is null)), this condition won't be used HashJoin.

mihaibudiu: I think IS NOT DISTINCT FROM is exactly what Java calls `==`.

xiedeyantu: Another problem is maybe a bug as I said in jira 6904.We can't get right result with IS NOT DISTINCT FROM currently. EnumableHashJoinRule make wrong convertion, the `filterNulls` is ignored.

mihaibudiu: you do not make this rule more complex because another rule is wrong.
you should fix the rule which is wrong.

xiedeyantu: > I think IS NOT DISTINCT FROM is exactly what Java calls `==`.

Yes, I agree with you! But the jdbcTest can't pass. We may remove the test temporarily.

xiedeyantu: > you do not make this rule more complex because another rule is wrong. you should fix the rule which is wrong.

Thanks, I will change to IS NOT DISTINCT FROM, and remove JdbcTest.

mihaibudiu: the first will work if I am right, you should double check.
removing tests is counterproductive, but you can temporarily disable a test.
there are many examples - see the Bug class.
But ideally the fix for the bug should be merged first...

core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
mihaibudiu: I am not sure what this statement means.
In our compiler we never use Java types, so they never show up in the Rel tree.
And I am not sure casts between Java types and SQL types are even legal.

core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
xiedeyantu: Got it. But for jira 6904 I can't find a best way to fix it.

core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
xiedeyantu: Here I refer to the logic of the existing code, if I remember correctly it should be AggJoinTranspose.

mihaibudiu: I still don't understand what the relationship is between types and rewrite rules.

mihaibudiu: Normally the validator inserts casts to ensure that all collections in an intersect have the same types.
(Assuming you use type coercion).

xiedeyantu: I am also confused, but it happens, usually when operating set operators

xiedeyantu:
@mihaibudiu Thanks, I have changed as your suggestion.

core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
```diff
@@ -0,0 +1,111 @@
...
+      return;
+    }
+
+    RelNode left = inputs.get(0);
+    RelNode right = inputs.get(1);
+
+    List<RexNode> conditions = new ArrayList<>();
+    int fieldCount = left.getRowType().getFieldCount();
+
+    // Converting the join condition to
```
mihaibudiu, line 73: this comment seems wrong now

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -3925,6 +3926,28 @@ public void checkOrderBy(final boolean desc,
         .returnsUnordered("empid=150; name=Sebastian");
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6880">
```
mihaibudiu, line 3929: this comment seems wrong
why does this case fail? 
does it invoke the new rule?

xiedeyantu: Yes, the new rule can cause this problem. Should I keep this case and modify the comments?

core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
xiedeyantu: Thanks, fixed

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -3925,6 +3926,28 @@ public void checkOrderBy(final boolean desc,
         .returnsUnordered("empid=150; name=Sebastian");
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6904">
+   * [CALCITE-6904] IS_NOT_DISTINCT_FROM is converted error in ElementerableJoinRule</a>. */
```
mihaibudiu, line 3930: there is no Elementerable

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -3925,6 +3926,28 @@ public void checkOrderBy(final boolean desc,
...
+  @Test void testIntersectToSemiJoin() {
+    final String sql = ""
+        + "select \"commission\" from \"hr\".\"emps\"\n"
+        + "intersect\n"
+        + "select \"commission\" from \"hr\".\"emps\" where \"empid\">=150";
+    CalciteAssert.hr()
+        .query(sql)
+        .withHook(Hook.PLANNER, (Consumer<RelOptPlanner>)
+            planner -> {
+              planner.removeRule(CoreRules.INTERSECT_TO_DISTINCT);
```
mihaibudiu, line 3941: can you just remove this new rule here as well and not disable the test?

xiedeyantu: This is due to jira 6904, which has nothing to do with the new rule. The purpose of adding this case is to test IntersectToSemiJoin. If this rule is removed, this test may not be needed.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
xiedeyantu: Fixed, thanks.

xiedeyantu:
@mihaibudiu Hi, Is there anything else I need to modify here?

mihaibudiu:
If the results produced by this rule are wrong then we cannot yet merge this.
If they are wrong because of 6904, let's try to solve that one first and then merge this.

xiedeyantu:
> If the results produced by this rule are wrong then we cannot yet merge this. If they are wrong because of 6904, let's try to solve that one first and then merge this.

Got it, thanks, it's a right way. I keep this pr here.

xiedeyantu:
@mihaibudiu Hi, if #4260 is ready, I can continue to work on this pr. WDYT

mihaibudiu:
We will merge that one soon, and then you should rebase this PR. Sorry for my delay with #4260

xiedeyantu:
> We will merge that one soon, and then you should rebase this PR. Sorry for my delay with #4260

No worries at all! You’ve already been incredibly helpful and responsive. I really appreciate all your support. I’ll go ahead and rebase this PR once the other one is merged. Thanks again!

xiedeyantu:
@mihaibudiu I have completed the rebase. Could you please help me see if there is anything else that needs to be adjusted? Thanks for your efficient review.

mihaibudiu (approved):

PR #4237, opened 2025-03-10
[CALCITE-6881] Druid Adapter can not handle for strlen call

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6881
char_length and charactor_lenght both would transform to strlen, but strlen cannot work well in DruidAdapter.

test:
DruidAdapterIT#testExpressionsSTRLENFilter
DruidAdapter2IT#testExpressionsSTRLENFilter


site/_docs/reference.md
```diff
@@ -1452,6 +1452,7 @@ comp:
 | string &#124;&#124; string | Concatenates two character strings
 | CHAR_LENGTH(string)        | Returns the number of characters in a character string
 | CHARACTER_LENGTH(string)   | As CHAR_LENGTH(*string*)
+| STRLEN(string)             | As CHAR_LENGTH(*string*)
```
mihaibudiu, line 1455: should this be in the section about dialect-specific operators
https://calcite.apache.org/docs/reference.html#dialect-specific-operators
?

xuzifu666: @mihaibudiu There is no Druid Dialect in doc now, so we should also need to add Druid Dialect? 
currently as list:
The 'C' (compatibility) column contains value:
* '*' for all libraries,
* 'b' for Google BigQuery ('fun=bigquery' in the connect string),
* 'c' for Apache Calcite ('fun=calcite' in the connect string),
* 'f' for Snowflake ('fun=snowflake' in the connect string),
* 'h' for Apache Hive ('fun=hive' in the connect string),
* 'm' for MySQL ('fun=mysql' in the connect string),
* 'q' for Microsoft SQL Server ('fun=mssql' in the connect string),
* 'o' for Oracle ('fun=oracle' in the connect string),
* 'r' for Amazon RedShift ('fun=redshift' in the connect string),
* 's' for Apache Spark ('fun=spark' in the connect string).

mihaibudiu: I guess you are not supporting a new function, just generating code for a function that has a new name in another dialect. In this case perhaps you should not add the function at all.

I am also not sure the function should be in StdOperatorTable; it is not a standard operator.

xuzifu666: OK, I know STRLEN is not a standard operator, currently if add a function in SqlStdOperatorTable would force to add function in reference, so I would remove it in doc and SqlStdOperatorTable

mihaibudiu:
Current version does recursion too.
Still to do: 
- udfs, 
- keep static declarations local

core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
```diff
@@ -85,7 +85,10 @@ public enum SqlLibrary {
   SNOWFLAKE("f", "snowflake"),
   /** A collection of operators that are in Apache Spark but not in standard
    * SQL. */
-  SPARK("s", "spark");
+  SPARK("s", "spark"),
+  /** A collection of operators that are in Apache Druid but not in standard
+   * SQL. */
+  DRUID("d", "druid");
```
NobiGo, line 91: We shouldn't add DRUID as a SqlLibrary. DRUID is not a Jdbc database. Each adapter has a specific section that indicates which functions are supported and which are not by the current adapter. This is because only the supported functions can enable some API conversions.

xuzifu666:
According comments, I had a mind that maybe the fast way to fix the bug is add a alias for CHAR_LENGTH with naming STRLEN, but need add STRLEN to SqlStdOperatorTable, do you agree with me?  @mihaibudiu @NobiGo

NobiGo:
-1 about this PR. For me, STRLEN is a Druid API rather than a JBDC database Function.

xuzifu666:
OK, would close the pr firstly and if had a better way to handle it would open again, thank for your review @NobiGo @mihaibudiu

PR #4240, opened 2025-03-12
[CALCITE-6885] SqlToRelConverter#convertUsing should not fail if commonTypeForBinaryComparison returns null

rubenada:
See CALCITE-6885

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -3458,22 +3458,21 @@ public class SqlToRelConverter {
           validator().getTypeCoercion().commonTypeForBinaryComparison(
               comparedTypes.get(0), comparedTypes.get(1));
       if (resultType == null) {
-        // This should never happen, since the program has been validated.
-        throw new IllegalArgumentException("Cannot join on field `" + name
-            + "` because the types are not comparable: " + comparedTypes);
-      }
-
-      List<RexNode> castedOperands = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -3458,22 +3458,20 @@ private RexNode convertUsing(SqlValidatorNamespace leftNamespace,
...
-      }
-
-      List<RexNode> castedOperands = new ArrayList<>();
-      for (int i = 0; i < operands.size(); i++) {
-        RexNode operand = operands.get(i);
-        RelDataType fieldType = comparedTypes.get(i);
-        RexNode expr = operand;
-        if (!fieldType.equals(resultType)) {
-          expr = rexBuilder.makeCast(resultType, operand, true, false);
+        list.add(rexBuilder.makeCall(SqlStdOperatorTable.EQUALS, operands));
```
mihaibudiu, line 3461: Frankly, I never understood what it means for the result to be `null`.
Ideally the API of the coercion would document the intended meaning.
As least I would add a comment here explaining that the call is left unchanged in this case.
I guess running multiple coercions, each of which handles some cases, is a legitimate use case.

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
rubenada: Thanks for the quick feedback @mihaibudiu . I agree that the API is not very clear in this regard.
I have added a comment as suggested.

PR #4242, opened 2025-03-13
[CALCITE-6887] In should distinct values with ReduceExpressionRule

xuzifu666:

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/core/Values.java
+++ b/core/src/main/java/org/apache/calcite/rel/core/Values.java
@@ -66,7 +66,7 @@ public abstract class Values extends AbstractRelNode implements Hintable {
 
   //~ Instance fields --------------------------------------------------------
 
-  public final ImmutableList<ImmutableList<RexLiteral>> tuples;
+  public ImmutableList<ImmutableList<RexLiteral>> tuples;
 
   //~ Constructors -----------------------------------------------------------
 
--- a/core/src/main/java/org/apache/calcite/rel/rules/ReduceExpressionsRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/ReduceExpressionsRule.java
@@ -33,6 +33,7 @@ import org.apache.calcite.rel.core.Window;
 import org.apache.calcite.rel.logical.LogicalCalc;
 import org.apache.calcite.rel.logical.LogicalFilter;
 import org.apache.calcite.rel.logical.LogicalProject;
+import org.apache.calcite.rel.logical.LogicalValues;
 import org.apache.calcite.rel.logical.LogicalWindow;
 import org.apache.calcite.rel.metadata.RelMetadataQuery;
 import org.apache.calcite.rel.type.RelDataType;
@@ -70,6 +71,7 @@ import org.apache.calcite.util.ImmutableBitSet;
 import org.apache.calcite.util.Pair;
 import org.apache.calcite.util.Util;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 
@@ -737,6 +739,17 @@ public abstract class ReduceExpressionsRule<C extends ReduceExpressionsRule.Conf
       RelOptPredicateList predicates, boolean treatDynamicCallsAsConstant) {
     // Replace predicates on CASE to CASE on predicates.
     boolean changed = new CaseShuttle().mutate(expList);
+
+    // distinct for IN rexNode
+    for (RexNode rexNode : predicates.pulledUpPredicates) {
+      if (rexNode.isA(SqlKind.IN)) {
+        RexSubQuery rexSubQuery = (RexSubQuery) rexNode;
+        List<ImmutableList<RexLiteral>> collect =
+            ((LogicalValues) rexSubQuery.rel).tuples
+                .stream().distinct().collect(Collectors.toList());
+        ((LogicalValues) rexSubQuery.rel).tuples = ImmutableList.copyOf(collect);
+      }
+    }
 
     // Find reducible expressions.
     final List<RexNode> constExps = new ArrayList<>();
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -348,6 +348,18 @@ class RelOptRulesTest extends RelOptTestBase {
   }
 
   /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6887">[CALCITE-6887]
+   * In should distinct values with ReduceExpressionRule</a>. */
+  @Test void testReduceExpressionsWithIn() {
+    final String sql = "select deptno, sal "
+        + "from emp "
+        + "where deptno in (1,1,2,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,3,1) ";
+    sql(sql).withRule(CoreRules.PROJECT_REDUCE_EXPRESSIONS)
+        .check();
+  }
+
+  /**
    * Test case for <a href="https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6746">
    * [CALCITE-6746] Optimization rule ProjectWindowTranspose is unsound</a>. */
   @Test void testConstantWindow() {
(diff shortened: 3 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/rules/ReduceExpressionsRule.java
```diff
@@ -738,6 +740,17 @@ protected static boolean reduceExpressionsInternal(RelNode rel,
     // Replace predicates on CASE to CASE on predicates.
     boolean changed = new CaseShuttle().mutate(expList);
 
+    // distinct for IN rexNode
+    for (RexNode rexNode : predicates.pulledUpPredicates) {
+      if (rexNode.isA(SqlKind.IN)) {
+        RexSubQuery rexSubQuery = (RexSubQuery) rexNode;
+        List<ImmutableList<RexLiteral>> collect =
+            ((LogicalValues) rexSubQuery.rel).tuples
```
mihaibudiu, line 748: why is rel here always LogicalValues?

xuzifu666: Hi @mihaibudiu ，in my test here I need to optimize the values of rexSubQuery by changing tuples，so need to transform to LogicalValues，if not nessary I refer add a check for rexSubQuery.rel first, WDYT？

mihaibudiu: I expected the code to have a test: `reqSubQuery.rel instanceof LogicalValues`
Why does it work without it?
If it always does, please add a comment explaining.

xuzifu666: I aggree with you and I would add the logic for it, thanks.

xuzifu666: Done, PTAL @mihaibudiu

core/src/main/java/org/apache/calcite/rel/core/Values.java
```diff
@@ -66,7 +66,7 @@ public abstract class Values extends AbstractRelNode implements Hintable {
 
   //~ Instance fields --------------------------------------------------------
 
-  public final ImmutableList<ImmutableList<RexLiteral>> tuples;
+  public ImmutableList<ImmutableList<RexLiteral>> tuples;
```
mihaibudiu, line 69: this is very dangerous.
all these classes are immutable for a reason.

core/src/main/java/org/apache/calcite/rel/rules/ReduceExpressionsRule.java
```diff
@@ -738,6 +740,20 @@ protected static boolean reduceExpressionsInternal(RelNode rel,
...
+                .stream().distinct().collect(Collectors.toList());
+        // In would transform to RexSubQuery and rel should be LogicalValues
+        if (rexSubQuery.rel instanceof LogicalValues) {
+          ((LogicalValues) rexSubQuery.rel).tuples = ImmutableList.copyOf(collect);
```
mihaibudiu, line 752: you cannot mutate the fields of a subquery; that subquery may be used by other objects as well.
you have to create a new object.

xuzifu666: OK,  would change it as your suggesstion. @mihaibudiu

xuzifu666: @mihaibudiu According my analysis，there is a problem  create a new object of SubQuery instead of  mutating the fields of a subquery ,since pulledUpPredicates is final ImmutableList<RexNode>. Maybe we cannot  change the construct of predicate by this way.

mihaibudiu: Just allocate a new instance of LogicalValues, or use the builder to build one

suibianwanwank:
@xuzifu666 Thanks for your work. We will optimize `IN` to `join+agg` in subqueryRemoveRule. So for this optimization. I prefer to add AggregateValueReduceRule than ReduceExpression.

core/src/main/java/org/apache/calcite/rel/rules/ReduceExpressionsRule.java
suibianwanwank: Frankly, I don't understand why it has to be value here. can you add test
```
    final String sql = "select deptno, sal "
        + "from emp "
        + "where deptno in (select 1 from emp) ";
```
 to be sure?

core/src/main/java/org/apache/calcite/rel/rules/ReduceExpressionsRule.java
```diff
@@ -738,6 +740,20 @@ protected static boolean reduceExpressionsInternal(RelNode rel,
     // Replace predicates on CASE to CASE on predicates.
     boolean changed = new CaseShuttle().mutate(expList);
 
+    // distinct for IN rexNode
+    for (RexNode rexNode : predicates.pulledUpPredicates) {
```
NobiGo, line 744: I think changing this VALUE here complicates the issue. Maybe it would be better to look into the "substituteSubQuery" option, before converting it to OR or RexSubquery.

xuzifu666: Yes，I agree，would do some changes in future.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -347,6 +347,18 @@ private static boolean skipItem(RexNode expr) {
...
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6887">[CALCITE-6887]
+   * In should distinct values with ReduceExpressionRule</a>. */
+  @Test void testReduceExpressionsWithIn() {
+    final String sql = "select deptno, sal "
+        + "from emp "
+        + "where deptno in (1,1,2,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,3,1) ";
+    sql(sql).withRule(CoreRules.PROJECT_REDUCE_EXPRESSIONS)
```
julianhyde, line 358: I would like to see a test case where the values are not sorted. `in (2, 1, 2, 3)` should become `in (2, 1, 3)`. This ensures stability.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -347,6 +347,18 @@ private static boolean skipItem(RexNode expr) {
         .check();
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6887">[CALCITE-6887]
+   * In should distinct values with ReduceExpressionRule</a>. */
+  @Test void testReduceExpressionsWithIn() {
+    final String sql = "select deptno, sal "
```
julianhyde, line 355: In your test query, put '\n' at end of lines and remove ' ' from end of lines.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xuzifu666: Yes, this is work well because values number in IN lower than 20, would transform to OR with multiple EQUAL and distinct the values. so I did not add case about with test it local. @julianhyde

mihaibudiu:
what is the status of this PR?

xuzifu666:
Move the pr to: https://github.com/apache/calcite/pull/4369 and would close the pr first.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
asolimando: This comment seems important, it's not been incorporated into #4371. I understand you need at least 20 elements to avoid the `OR` expansion, it's OK, just make sure you have a non-sorted example with at least 20 elements as Julian suggests.

xuzifu666: Thanks for reminder, and I would add this to PR. @asolimando

xuzifu666: @asolimando @julianhyde I had added the case with non-sorted and less then 20 values in commit: https://github.com/apache/calcite/pull/4371/commits/c4273c449855a8dc8caa48c0b3f49d27a782495e

PR #4244, opened 2025-03-14
[CALCITE-6880] Implement IntersectReorderRule

xiedeyantu:
link: [CALCITE-6891](https://issues.apache.org/jira/browse/CALCITE-6891)


core/src/main/java/org/apache/calcite/rel/rules/IntersectReorderRule.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    final Intersect intersect = call.rel(0);
+    final RelMetadataQuery mq = call.getMetadataQuery();
+    final List<RelNode> inputs = intersect.getInputs();
+
+    final List<Pair<RelNode, Double>> inputsWithRowCounts = new ArrayList<>();
+
+    for (RelNode input : inputs) {
+      Double rowCount = mq.getRowCount(input);
```
NobiGo, line 55: Do we only need to assess whether RowCount can be replaced by Cumulative Cost? Or just consider RowCount only.

xiedeyantu: > Do we only need to assess whether RowCount can be replaced by Cumulative Cost? Or just consider RowCount only.

Thanks, I thought about your question. In intersect, the number of output columns of the children is equal. In my opinion, RowCount may represent cost.

core/src/main/java/org/apache/calcite/rel/rules/IntersectReorderRule.java
```diff
@@ -0,0 +1,99 @@
...
+@Value.Enclosing
+public class IntersectReorderRule extends RelRule<IntersectReorderRule.Config>
+      implements SubstitutionRule {
+  /** Creates an IntersectReorderRule. */
+  protected IntersectReorderRule(Config config) {
+    super(config);
+  }
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    final Intersect intersect = call.rel(0);
```
suibianwanwank, line 49: I suggest it could be simpler here, For example:
```
    final Intersect intersect = call.rel(0);
    final RelMetadataQuery mq = call.getMetadataQuery();
    final List<RelNode> inputs = intersect.getInputs();

    List<RelNode> sortedInputs = inputs.stream()
        .sorted(Comparator.comparingDouble(input -> mq.getRowCount(input)))
        .collect(Collectors.toList());

    if (inputs.equals(sortedInputs)) {
        return;
    }

    final RelBuilder relBuilder = call.builder();
    relBuilder.pushAll(sortedInputs);
    relBuilder.intersect(intersect.all, sortedInputs.size());

    call.transformTo(relBuilder.build());
```

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -10048,4 +10048,35 @@ private void checkLoptOptimizeJoinRule(LoptOptimizeJoinRule rule) {
...
   }
+
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6891">
+   * [CALCITE-6891] Implement IntersectReorderRule</a>. */
+  @Test void testIntersectReorderRule() {
+    final String sql = "select deptno from emp\n"
+        + "intersect\n"
+        + "select deptno from dept\n";
+
+    sql(sql).withVolcanoPlanner(true, p -> {
```
suibianwanwank, line 10059: Maybe it would be clearer to apply the rules by adding different filters for the same scan in the test case.

core/src/main/java/org/apache/calcite/rel/rules/IntersectReorderRule.java
xiedeyantu: Thanks, It's seems good!

core/src/main/java/org/apache/calcite/rel/rules/IntersectReorderRule.java
NobiGo: The presence of indexes and other related conditions in the table may affect the performance of this query. Therefore, we should not only focus on the number of rows.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xiedeyantu: Thanks, fixed.

core/src/main/java/org/apache/calcite/rel/rules/IntersectReorderRule.java
suibianwanwank: I prefer to use rowCount. An example that may not be entirely accurate is that the left input has a low cost but a large row count, while the right input has a small amount of data but a high cost.
```sql
SELECT order_id, customer_id  
FROM orders 
INTERSECT  
SELECT order_id, customer_id  
FROM (  
    SELECT order_id, customer_id  
    FROM orders  
      WHERE total_amount > (SELECT AVG(total_amount) FROM orders WHERE status = 'COMPLETED')  
    ORDER BY created_at DESC  
    LIMIT 1  
) t;
```
Unless the output of a partial Intersect is empty, there is little point in using the cost. WDYT

suibianwanwank (approved):

core/src/main/java/org/apache/calcite/rel/rules/IntersectReorderRule.java
NobiGo: In fact, the Cumulative Cost currently being compared is also RowCount. The Cumulative Cost currently consists of three parts: RowCount、CPU and IO. I don't understand why the data volume returned is large despite being a low-cost solution. This actually depends on the implementation of CBO.

xiedeyantu: As far as I understand, the calculation of Intersect is related to the Cumulative Cost. Currently, the cost value is composed of three parts that as you said. The cost of CPU and IO is basically the same when the number of columns is the same. The overall cost can be simplified to RowCount, so using RowCount here can reflect the cost. WDYT

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -10059,4 +10059,35 @@ private void checkLoptOptimizeJoinRule(LoptOptimizeJoinRule rule) {
     sql(sql).withRule(CoreRules.INTERSECT_TO_EXISTS)
         .checkUnchanged();
   }
+
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6891">
```
NobiGo, line 10063: Please keep this Java comment consistent with the format of other comments.

xiedeyantu: Thanks, Done.

xiedeyantu: @NobiGo Hi, need I to do something else?

xiedeyantu:
@NobiGo Hi, could you help me review this PR again?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -10059,4 +10059,35 @@ private void checkLoptOptimizeJoinRule(LoptOptimizeJoinRule rule) {
...
+    sql(sql).withVolcanoPlanner(true, p -> {
+      p.addRule(CoreRules.INTERSECT_REORDER);
+      p.addRule(EnumerableRules.ENUMERABLE_TABLE_SCAN_RULE);
+      p.addRule(EnumerableRules.ENUMERABLE_INTERSECT_RULE);
+      p.addRule(EnumerableRules.ENUMERABLE_FILTER_RULE);
+      p.addRule(EnumerableRules.ENUMERABLE_PROJECT_RULE);
+    }).check();
+  }
+
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6891">
```
NobiGo, line 10079: The style of the javadoc needs to be consistent with that of others.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -5980,6 +5980,64 @@ LogicalProject(EMPNO=[$0], ENAME=[$1], T=[$10])
...
+  LogicalProject(DEPTNO=[$7])
+    LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+  LogicalProject(DEPTNO=[$7])
+    LogicalFilter(condition=[>($7, 5)])
+      LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+]]>
+    </Resource>
+    <Resource name="planAfter">
+      <![CDATA[
+EnumerableIntersect(all=[false])
```
NobiGo, line 6005: These are all test cases for intersect. we can also cover the case of `intersect all`.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xiedeyantu: Thanks, fixed.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
xiedeyantu: Thanks, I added the case of `intersect all`.

core/src/main/java/org/apache/calcite/rel/rules/IntersectReorderRule.java
```diff
@@ -0,0 +1,81 @@
...
+    call.transformTo(relBuilder.build());
+  }
+
+  /** Rule configuration. */
+  @Value.Immutable
+  public interface Config extends RelRule.Config {
+    Config DEFAULT = ImmutableIntersectReorderRule.Config.of()
+        .withOperandSupplier(b0 ->
+            b0.operand(LogicalIntersect.class)
+                .predicate(intersect -> intersect.getInputs().size() > 1)
```
NobiGo, line 73: Why must the condition here be greater than 1? Do we have any input cases?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -10059,4 +10059,52 @@ private void checkLoptOptimizeJoinRule(LoptOptimizeJoinRule rule) {
...
+  /** Test case of
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6891">[CALCITE-6891]
+   * Implement IntersectReorderRule</a>. */
+  @Test void testIntersectReorderRuleAll() {
+    final String sql = "select deptno from emp\n"
+        + "intersect all\n"
+        + "select deptno from emp where deptno > 5\n";
+    sql(sql).withVolcanoPlanner(true, p -> {
+      p.addRule(CoreRules.INTERSECT_REORDER);
+      p.addRule(EnumerableRules.ENUMERABLE_TABLE_SCAN_RULE);
```
NobiGo, line 10104: Should we need this ENUMERABLE-related plan?

xiedeyantu: This is to demonstrate the effectiveness of cost optimization in execution, as the converted plan was indeed chosen in CBO. Do I need to adjust it to a logical plan?

core/src/main/java/org/apache/calcite/rel/rules/IntersectReorderRule.java
xiedeyantu: The number of `getInputs()` in the `intersect` can be arbitrary (of course, handwritten SQL may usually make the children >=2), and a judgment is added here that only children greater than 1 will decide whether to perform the exchange.

NobiGo: I've noticed that others also write it this way. Let's leave it at this for now.

NobiGo (approved):

PR #4247, opened 2025-03-16
[CALCITE-6894] ClickHouse dialect can support another form of writing aggFilterWhere

xiedeyantu:
Jira link: [CALCITE-6894](https://issues.apache.org/jira/browse/CALCITE-6894)

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/ClickHouseSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/ClickHouseSqlDialect.java
@@ -25,6 +25,7 @@ import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlDataTypeSpec;
 import org.apache.calcite.sql.SqlDateLiteral;
 import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.SqlFilterOperator;
 import org.apache.calcite.sql.SqlLiteral;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlTimeLiteral;
@@ -37,6 +38,8 @@ import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.util.RelToSqlConverterUtil;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Locale;
 
 import static org.apache.calcite.util.RelToSqlConverterUtil.unparseBoolLiteralToCondition;
 
@@ -76,6 +79,10 @@ public class ClickHouseSqlDialect extends SqlDialect {
 
   @Override public boolean supportsAliasedValues() {
     return false;
+  }
+
+  @Override public boolean supportsAggregateFunctionFilter() {
+    return true;
   }
 
   @Override public CalendarPolicy getCalendarPolicy() {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/ClickHouseSqlDialect.java
```diff
@@ -181,6 +188,21 @@ private static SqlDataTypeSpec createSqlDataTypeSpecByName(String typeAlias,
       return;
     }
 
+    if (call.getOperator() instanceof SqlFilterOperator) {
+      SqlCall aggCall = call.operand(0);
+      if (call.getOperandList().size() > 1 && call.operand(1) != null) {
+        SqlCall filterCondition = call.operand(1);
+        String functionName = aggCall.getOperator().getName();
+        writer.print(functionName.toLowerCase(Locale.ROOT) + "If");
```
mihaibudiu, line 196: does this work for all aggregate functions?
Is there a stddevif?

xiedeyantu: Thanks for review it. Clickhouse has such a logic. It will concatenate the aggregate function and -If instead of implementing a new xxxIf function. For example, whether stddevIf exists depends on whether Clickhouse supports stddev. I understand that the conversion of pr is correct. If more precision is needed, we may need to do a good job of mapping the functions from calcite to Clickhouse, but it seems that each dialect is being implemented one by one, which may be an area that needs improvement.

mihaibudiu: Whether clickhouse supports stddev is what I am asking.
Calcite supports a set of aggregates, and clickhouse supports another set. If the Calcite set is included in the Clickhouse set, this change is fine.

xiedeyantu: I think what you're saying is another issue, even if CASE WHEN is still used, if ClickHouse doesn't support STDDEV, then it can't also work. What do you think about creating another topic if needed?

caicancai: @mihaibudiu In my impression, ClickHouse does not seem to support stddevif

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -361,7 +361,13 @@ private static String toSql(RelNode root, SqlDialect dialect,
         + "FROM `foodmart`.`product`\n"
         + "WHERE `product_id` > 0\n"
         + "GROUP BY `product_id`";
+    final String expectedClickHouse = "SELECT"
```
caicancai, line 364: Can you add more tests for different types of aggregate functions?

core/src/main/java/org/apache/calcite/sql/dialect/ClickHouseSqlDialect.java
xiedeyantu: Thanks, do I just need to add some extra comments? I want to confirm whether this conversion is needed. In fact, this conversion cannot solve the function mapping between the two dialects. This is actually similar to the syntax sugar of ClickHouse dialect.

xiedeyantu: > Whether clickhouse supports stddev is what I am asking. Calcite supports a set of aggregates, and clickhouse supports another set. If the Calcite set is included in the Clickhouse set, this change is fine.

@mihaibudiu After seeing your and cancai's replies, maybe I didn't explain clearly enough. Let me elaborate on the issue I mentioned:
ClickHouse doesn't support stddev. Converting stddev(x) filter(where) to stddevIf(x, condition) will throw an error in ClickHouse. What I meant in my previous reply was that using stddev(case when) would also throw an error. This PR is just making a formal transformation.
If we want to properly solve whether aggregate functions or other functions can be correctly translated, maybe creating a new JIRA ticket would be better. What do you think?

mihaibudiu: Clickhouse supports stddev_samp and stddev_pop.
Converting unsupported code to unsupported code is not really helpful.
I think this code should handle correctly the supported functions and should report an error for the other cases

xiedeyantu: Well, I understand. We may not need this PR. Correctly supporting function conversion may be a problem faced by all dialects. If this PR only handles some aggregation functions supported by ClickHouse, it may not be a good solution. So I will temporarily close this PR for now.

xiedeyantu: But the - If form does have performance improvements, it's not just a formal transformation. see https://github.com/ClickHouse/ClickHouse/pull/17043 and others.

PR #4249, opened 2025-03-16
[CALCITE-6820] Trino dialect implementation

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6820

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/TrinoSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/TrinoSqlDialect.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.sql.dialect;
+
+import org.apache.calcite.avatica.util.Casing;
+import org.apache.calcite.config.NullCollation;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.util.RelToSqlConverterUtil;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * A <code>SqlDialect</code> implementation for the Trino database.
+ */
+public class TrinoSqlDialect extends PrestoSqlDialect {
+  public static final Context DEFAULT_CONTEXT = SqlDialect.EMPTY_CONTEXT
+      .withDatabaseProduct(DatabaseProduct.TRINO)
+      .withIdentifierQuoteString("\"")
+      .withUnquotedCasing(Casing.UNCHANGED)
+      .withNullCollation(NullCollation.LAST);
+
+  public static final SqlDialect DEFAULT = new TrinoSqlDialect(DEFAULT_CONTEXT);
+
+  /**
+   * Creates a TrinoSqlDialect.
+   */
+  public TrinoSqlDialect(Context context) {
+    super(context);
+  }
+
+  @Override public void unparseOffsetFetch(SqlWriter writer, @Nullable SqlNode offset,
+      @Nullable SqlNode fetch) {
+    unparseFetchUsingAnsi(writer, offset, fetch);
+  }
+
+  @Override public void unparseCall(SqlWriter writer, SqlCall call,
+      int leftPrec, int rightPrec) {
+    if (call.getOperator() == SqlStdOperatorTable.SUBSTRING) {
+      RelToSqlConverterUtil.specialOperatorByName("SUBSTRING")
+          .unparse(writer, call, 0, 0);
+    } else {
+      super.unparseCall(writer, call, leftPrec, rightPrec);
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/TrinoSqlDialect.java
```diff
@@ -0,0 +1,235 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * A <code>SqlDialect</code> implementation for the Trino database.
+ */
+public class TrinoSqlDialect extends SqlDialect {
```
caicancai, line 57: Why not improve based on presto dialect

xuzifu666: Current version of the Trino Dialect seems work well in Trino, some more improvement may need furture update.

xuzifu666: Had changed to base on presto dialect，thanks for the reminder. @caicancai

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -393,6 +394,7 @@ private static String toSql(RelNode root, SqlDialect dialect,
         .withMysql().ok(expectedMysql)
         .withClickHouse().ok(expectedClickHouse)
         .withPresto().ok(expectedPresto)
+        .withTrino().ok(expectedPresto)
```
xiedeyantu, line 397: Use `expectedPresto` for verification directly. What if the result is consistent with presto?

xuzifu666: This is for more clear for test case, and maybe not nessary to change, thanks ~

caicancai (approved):

core/src/main/java/org/apache/calcite/sql/dialect/TrinoSqlDialect.java
```diff
@@ -0,0 +1,78 @@
...
+        super.unparseMapValue(writer, call, leftPrec, rightPrec);
+        break;
+      case CHAR_LENGTH:
+        SqlCall lengthCall = SqlLibraryOperators.LENGTH
+            .createCall(SqlParserPos.ZERO, call.getOperandList());
+        super.unparseCall(writer, lengthCall, leftPrec, rightPrec);
+        break;
+      default:
+        // Current impl is same with Postgresql.
+        PostgresqlSqlDialect.DEFAULT.unparseCall(writer, call, leftPrec, rightPrec);
```
NobiGo, line 73: Here, the grammar is kept consistent with PG, but is it quite different from Presto?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -1216,6 +1228,10 @@ private static String toSql(RelNode root, SqlDialect dialect,
         + "FROM \"foodmart\".\"product\"\n"
         + "GROUP BY ROLLUP(\"product_class_id\")\n"
         + "LIMIT 5";
+    final String expectedTrino = "SELECT \"product_class_id\", COUNT(*) AS \"C\"\n"
```
xiedeyantu, line 1231: Trino itself also supports limit, do we need to convert limit into fetch here, and keep same as presto?

core/src/main/java/org/apache/calcite/sql/dialect/TrinoSqlDialect.java
xuzifu666: According my test about trino，here still not found other differences，and if had found I and others would fix it directly，much of syntax keep the same behavior with presto，so IMO here can keep it firstly @NobiGo

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xuzifu666: Other dialects would support fetch next syntax first if they supported, and the change here is different from presto, so I prefer keep it the same with other dialect.

core/src/main/java/org/apache/calcite/sql/dialect/TrinoSqlDialect.java
NobiGo: All the functions here are consistent with PG. They have nothing to do with Presto. As far as I know, Trino is more similar to Presto.

xuzifu666: OK, here is fix the function SUBSTR, I would change here not use PG Dialect. @NobiGo

xuzifu666: Had changed, PTAL，thanks @NobiGo

core/src/main/java/org/apache/calcite/sql/dialect/TrinoSqlDialect.java
```diff
@@ -0,0 +1,95 @@
...
+    unparseFetchUsingAnsi(writer, offset, fetch);
+  }
+
+  @Override public void unparseCall(SqlWriter writer, SqlCall call,
+      int leftPrec, int rightPrec) {
+    if (call.getOperator() == SqlStdOperatorTable.SUBSTRING) {
+      RelToSqlConverterUtil.specialOperatorByName("SUBSTRING")
+          .unparse(writer, call, 0, 0);
+    } else if (call.getOperator() == SqlStdOperatorTable.APPROX_COUNT_DISTINCT) {
+      RelToSqlConverterUtil.specialOperatorByName("APPROX_DISTINCT")
```
NobiGo, line 65: This is the same as Presto. We don't need to rewrite it.

core/src/main/java/org/apache/calcite/sql/dialect/TrinoSqlDialect.java
```diff
@@ -0,0 +1,95 @@
...
+          .unparse(writer, call, 0, 0);
+    } else {
+      switch (call.getKind()) {
+      case MAP_VALUE_CONSTRUCTOR:
```
NobiGo, line 69: This is the same as Presto. We don't need to rewrite it.

NobiGo (commented):
In Trino's dialect, should we only retain the differences?

xuzifu666:
> In Trino's dialect, should we only retain the differences?

Very thanks for reminder，I had only retain the differences. @NobiGo

PR #4250, opened 2025-03-17
[CALCITE-6875] EnumerableFilterRule/EnumerableProjectRule should not …

NobiGo:
…convert a Logical Filter/Project to Enumerable Filter/Project when it contains Subquery

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableFilterRule.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableFilterRule.java
@@ -21,6 +21,7 @@ import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.convert.ConverterRule;
 import org.apache.calcite.rel.core.Filter;
 import org.apache.calcite.rel.logical.LogicalFilter;
+import org.apache.calcite.rex.RexUtil;
 
 /**
  * Rule to convert a {@link LogicalFilter} to an {@link EnumerableFilter}.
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/logical/LogicalFilter.java
```diff
@@ -135,9 +133,7 @@ public static LogicalFilter create(final RelNode input, RexNode condition,
     final RelMetadataQuery mq = cluster.getMetadataQuery();
     final RelTraitSet traitSet = cluster.traitSetOf(Convention.NONE)
         .replaceIfs(RelCollationTraitDef.INSTANCE,
-            () -> RelMdCollation.filter(mq, input))
-        .replaceIf(RelDistributionTraitDef.INSTANCE,
-            () -> RelMdDistribution.filter(mq, input));
+            () -> RelMdCollation.filter(mq, input));
```
mihaibudiu, line 136: I don't know how this change is related.

NobiGo: If we don't do this. 
testEnumerableFilterRule will throw an exception:
```
There are not enough rules to produce a node with desired properties: convention=ENUMERABLE, dist=any.
Missing conversion is LogicalFilter[convention: NONE -> ENUMERABLE]
There is 1 empty subset: rel#39:RelSubset#1.ENUMERABLE.broadcast, the relevant part of the original plan is as follows
14:LogicalFilter(condition=[NOT(<= SOME($5, {
LogicalProject(COMM=[$6])
}))])
  8:LogicalTableScan(subset=[rel#13:RelSubset#0.NONE.any], table=[[CATALOG, SALES, EMP]])
```
The mismatch of Trait information leads to the invalidity of the transformed plan.

NobiGo: I'm not sure if this is a correct modification because no other test cases cover this point.

NobiGo: Create CALCITE-6899 to log this issue and use a dynamic table without RelDistributionTraitDef to ensure the test passes.

xiedeyantu: I try to remove `p.addRelTraitDef(RelDistributionTraitDef.INSTANCE)`, error can not throw.

NobiGo: @xiedeyantu We can discuss this issue in CAICITE-6899

mihaibudiu (approved):

PR #4255, opened 2025-03-19
[CALCITE-6901] SINGLE_VALUE rewrite to wrong sql in MySQL dialect

NobiGo:


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -5603,7 +5603,7 @@ private void checkLiteral2(String expression, String expected) {
     final String expectedMysql = "SELECT `product`.`product_class_id` AS `C`\n"
         + "FROM `foodmart`.`product`\n"
         + "LEFT JOIN (SELECT CASE COUNT(*) "
-        + "WHEN 0 THEN NULL WHEN 1 THEN `product_class_id` ELSE (SELECT NULL\n"
+        + "WHEN 0 THEN NULL WHEN 1 THEN MIN(`product_class_id`) ELSE (SELECT NULL\n"
```
mihaibudiu, line 5606: would be nice if someone had tested this code in some way

mihaibudiu (approved):

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
NobiGo: I've tested it on the local service.

PR #4256, opened 2025-03-21
[CALCITE-2109] MongoAdapter: Support in condition with and condition

tuichenchuxin:
see https://issues.apache.org/jira/browse/CALCITE-2109

- Support in condition with and condition
  - refactor translateOr and translateAnd

### DIFF
--- a/mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
+++ b/mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
@@ -859,4 +859,64 @@ public class MongoAdapterTest implements SchemaFactory {
         .limit(2)
         .returns("STATE=VT; AVG(pop)=26408\nSTATE=AK; AVG(pop)=26856\n");
   }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-2109">[CALCITE-2109]
+   * Mongo adapter: unable to translate (A AND B) conditional case</a>. */
+  @Test void testTranslateAndInCondition() {
+    assertModel(MODEL)
+        .query("select state, city from zips "
+            + "where city='LEWISTON' and state in ('ME', 'VT') "
+            + "order by state")
+        .queryContains(
+            mongoChecker(
+            "{$match: {$and: [{$or: [{state: \"ME\"}, {state: \"VT\"}]}, {city: \"LEWISTON\"}]}}",
+            "{$project: {STATE: '$state', CITY: '$city'}}",
+            "{$sort: {STATE: 1}}"))
+        .returns("STATE=ME; CITY=LEWISTON\n");
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-2109">[CALCITE-2109]
+   * Mongo adapter: unable to translate (A AND B) conditional case</a>. */
+  @Test void testTranslateOrAndCondition() {
+    assertModel(MODEL)
+        .query("select state, city from zips "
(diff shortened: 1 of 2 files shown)


mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
```diff
@@ -859,4 +859,26 @@ private static Consumer<List> mongoChecker(final String... expected) {
         .limit(2)
         .returns("STATE=VT; AVG(pop)=26408\nSTATE=AK; AVG(pop)=26856\n");
   }
+
+  @Test void testTranslateAndInCondition() {
```
xiedeyantu, line 863: It may be necessary to include Jira's link and description on each test.

tuichenchuxin: @xiedeyantu 
Thank you for your suggestion. Done.

mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
```diff
@@ -859,4 +859,32 @@ private static Consumer<List> mongoChecker(final String... expected) {
...
+        .returns("STATE=ME; CITY=LEWISTON\n");
+  }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-2109">[CALCITE-2109]
+   * Mongo adapter: unable to translate (A AND B) conditional case</a>. */
+  @Test void testAndAlwaysFalseCondition() {
+    assertModel(MODEL)
+        .query("select state, city from zips "
+            + "where city='LEWISTON' and 1=0 "
```
NobiGo, line 885: `1=0`  makes the condition of this plan disappear in the Enumerable phase. This test for this PR should be invalid.

mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
```diff
@@ -859,4 +859,32 @@ private static Consumer<List> mongoChecker(final String... expected) {
...
         .returns("STATE=VT; AVG(pop)=26408\nSTATE=AK; AVG(pop)=26856\n");
   }
+
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-2109">[CALCITE-2109]
+   * Mongo adapter: unable to translate (A AND B) conditional case</a>. */
+  @Test void testTranslateAndInCondition() {
+    assertModel(MODEL)
+        .query("select state, city from zips "
+            + "where city='LEWISTON' and state in ('ME', 'VT') "
```
NobiGo, line 869: How about adding a test case of "A OR B AND C"?

mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
tuichenchuxin: > `1=0` makes the condition of this plan disappear in the Enumerable phase. This test for this PR should be invalid.

@NobiGo Yes, but in the issue. This case will throw exception, so, I added this.
if you think it's unnecessary, I'll delete it.

mongodb/src/test/java/org/apache/calcite/adapter/mongodb/MongoAdapterTest.java
tuichenchuxin: @NobiGo sure, done.

NobiGo (approved):

caicancai:
Whether to support CNF

PR #4258, opened 2025-03-22
[CALCITE-6906] Support ClickHouse Dialect with cube with syntax

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6906

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -636,13 +636,21 @@ class RelToSqlConverterTest {
     sql(query).ok(expected);
   }
 
-  @Test void testSelectQueryWithHiveCube() {
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6906">[CALCITE-6906]
+   * Support ClickHouse Dialect with cube with syntax</a>. */
+  @Test void testSelectQueryWithCube() {
     String query = "select \"product_class_id\", \"product_id\", count(*) "
             + "from \"product\" group by cube(\"product_class_id\", \"product_id\")";
     String expected = "SELECT `product_class_id`, `product_id`, COUNT(*)\n"
             + "FROM `foodmart`.`product`\n"
             + "GROUP BY `product_class_id`, `product_id` WITH CUBE";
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -636,13 +636,21 @@ private static String toSql(RelNode root, SqlDialect dialect,
...
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6906">[CALCITE-6906]
+   * Support ClickHouse Dialect with cube with syntax</a>. */
+  @Test void testSelectQueryWithCube() {
     String query = "select \"product_class_id\", \"product_id\", count(*) "
             + "from \"product\" group by cube(\"product_class_id\", \"product_id\")";
     String expected = "SELECT `product_class_id`, `product_id`, COUNT(*)\n"
             + "FROM `foodmart`.`product`\n"
             + "GROUP BY `product_class_id`, `product_id` WITH CUBE";
-    sql(query).withHive().ok(expected);
+    String expectedClickHouse = "SELECT `product_class_id`, `product_id`, COUNT(*)\n"
```
mihaibudiu, line 648: This does not seem to match the docs of clickhouse: https://clickhouse.com/docs/sql-reference/statements/select/group-by#cube-modifier

xuzifu666: In doc description the same query also can be written using WITH keyword.
like：SELECT year, month, day, count(*) FROM t GROUP BY year, month, day WITH CUBE;
and Ialso test about it.

mihaibudiu: Why is this PR needed if the other syntax is correct?

xuzifu666: IMO if user migrate other SQL system which support with
 cube syntax to clickhouse, in this condition they not need to change SQL, if clickhouse self supprt the syntax maybe better.

xuzifu666: OK, I know thanks for your reminder, here maybe not need to change.@mihaibudiu

PR #4259, opened 2025-03-23
[CALCITE-4708] Infer generic Collections field type with ReflectiveSchema

a-simeshin:
Infer list generic type while Table instance is created at ReflectiveSchema class

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/java/ReflectiveSchema.java
+++ b/core/src/main/java/org/apache/calcite/adapter/java/ReflectiveSchema.java
@@ -61,6 +61,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Collection;
 import java.util.Collections;
@@ -194,7 +195,7 @@ public class ReflectiveSchema
   /** Returns a table based on a particular field of this schema. If the
    * field is not of the right type to be a relation, returns null. */
   private <T> @Nullable Table fieldRelation(final Field field) {
-    final Type elementType = getElementType(field.getType());
+    final Type elementType = getElementType(field);
     if (elementType == null) {
       return null;
     }
@@ -217,12 +218,17 @@ public class ReflectiveSchema
 
   /** Deduces a collection's element type;
    * same logic as {@link #toEnumerable}. */
-  private static @Nullable Type getElementType(Class<?> clazz) {
+  private static @Nullable Type getElementType(Field field) {
+    final Class<?> clazz = field.getType();
     if (clazz.isArray()) {
       return clazz.getComponentType();
     }
     if (Iterable.class.isAssignableFrom(clazz)) {
-      return Object.class;
+      try {
+        return ((ParameterizedType) field.getGenericType()).getActualTypeArguments()[0];
+      } catch (Exception ignored) {
+        return Object.class;
+      }
     }
     return null; // not a collection/array/iterable
   }
@@ -298,7 +304,7 @@ public class ReflectiveSchema
    *
    * <p>The following example instantiates a {@code FoodMart} object as a schema
    * that contains tables called {@code EMPS} and {@code DEPTS} based on the
-   * object's fields.
+   * object's fields as arrays and {@code DISCOUNTS} with {@code OFFERS} as collections.
    *
    * <blockquote><pre>
    * schemas: [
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/apache/calcite/test/ReflectiveSchemaTest.java
```diff
@@ -972,4 +980,48 @@ public static class DateColumnSchema {
...
+        Arguments.of("cats_stack"),
+        Arguments.of("cats_vector"),
+        Arguments.of("cats_array_dequeue"),
+        Arguments.of("cats_hash_set"),
+        Arguments.of("cats_linked_hash_set"),
+        Arguments.of("cats_tree_set"));
+  }
+
+  /**
+   * Test that the Collections works as field of ReflectiveSchema.
```
mihaibudiu, line 1000: "Test that Collections can be used as field"

core/src/test/java/org/apache/calcite/test/ReflectiveSchemaTest.java
```diff
@@ -972,4 +980,48 @@ public static class DateColumnSchema {
...
+  @ParameterizedTest @MethodSource("collectionsSchemaFieldNames")
+  void testSelectCollectionsFieldsSchema(String fieldName) {
+    CalciteAssert.that()
+        .withSchema("s", COLLECTIONS_SCHEMA)
+        .query("select * from \"s\".\"" + fieldName + "\"")
+        .returns("name=Simba; age=2\nname=Luna; age=1\nname=Bella; age=3\n");
+  }
+
+  /**
+   * Test that the Collections works as field of ReflectiveSchema with corrects
```
mihaibudiu, line 1013: correct statistics

core/src/test/java/org/apache/calcite/test/ReflectiveSchemaTest.java
```diff
@@ -737,25 +742,28 @@ private void checkOp(CalciteAssert.AssertThat with, String fn) {
...
-            + "\n"
-            + "\n");
-    // List is implicitly converted to Enumerable
-    with.query("select * from \"s\".\"list\"")
-        .returns("\n"
-            + "\n"
-            + "\n"
-            + "\n");
+
+    final String employees = new StringBuilder()
```
mihaibudiu, line 753: why is this result different?
was the previous result wrong?

core/src/main/java/org/apache/calcite/adapter/java/ReflectiveSchema.java
```diff
@@ -320,6 +326,8 @@ private static class ReflectiveTable
    * &nbsp;
    *   Employee[] EMPS;
    *   Department[] DEPTS;
+   *   List&lt;Discount&gt; DISCOUNTS;
```
mihaibudiu, line 329: was this an omission?

a-simeshin: Im really sorry for the long inactive time.

There is a checkstyle rule and wo `&lt;` I have an error `Javadoc comment at column 11 has parse error. Missed HTML close tag 'Offer'. Sometimes it means that close tag missed for one of previous tags.`

core/src/test/java/org/apache/calcite/test/ReflectiveSchemaTest.java
a-simeshin: >why is this result different?

Before this change, ReflectiveSchema correctly determined that there were 3 records in the enumerable table, but there was no mechanism for deduce fields within in a collection type. 

The mechanism worked only for arrays, but now it works for collections, so it not only defines 3 objects in the enumerable table, but also correctly handles the selection by this table with the display of fields instead of empty strings.

Comment in the test before: 
`// Enumerable field returns 3 records with 0 fields`

>was the previous result wrong?

The previous result was correct for the implementation before the PR changes, but now the test needs to be adjusted.

core/src/test/java/org/apache/calcite/test/ReflectiveSchemaTest.java
a-simeshin: Fixed

core/src/test/java/org/apache/calcite/test/ReflectiveSchemaTest.java
a-simeshin: Fixed

mihaibudiu (commented):
Your PR title does not match the JIRA issue name.
Also, the last comment in the JIRA issue is the following, from @julianhyde 

> A lot of the confusion (including my own) is because no one knows what is supposed to work. Is there any documentation on the web site how to write a reflective schema? The PR should add (or update) documentation.

I completely agree, these changes seem to make sense on their own, but there is no spec really about what ReflectiveSchema is supposed to do (or not do). Could you find some place to document this? At the very least, this could be a large JavaDoc for the ReflectiveSchema class.

PR #4260, opened 2025-03-23
[CALCITE-6904] IS_NOT_DISTINCT_FROM is incorrectly handled by EnumerableJoinRule

xiedeyantu:
Jira link: [CALCITE-6904](https://issues.apache.org/jira/browse/CALCITE-6904)

### DIFF
--- a/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
+++ b/core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
@@ -425,6 +425,7 @@ import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_JSON_OBJECT;
 import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_JSON_SCALAR;
 import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_JSON_VALUE;
 import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_NOT_A_SET;
+import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_NOT_DISTINCT_FROM;
 import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_NOT_EMPTY;
 import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_NOT_FALSE;
 import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_NOT_JSON_ARRAY;
@@ -1008,6 +1009,7 @@ public class RexImpTable {
       define(IS_NOT_TRUE, new IsNotTrueImplementor());
       define(IS_FALSE, new IsFalseImplementor());
       define(IS_NOT_FALSE, new IsNotFalseImplementor());
+      define(IS_NOT_DISTINCT_FROM, new IsNotDistinctFromImplementor());
 
       // LIKE, ILIKE, RLIKE and SIMILAR
       defineReflective(LIKE, BuiltInMethod.LIKE.method,
@@ -4738,6 +4740,51 @@ public class RexImpTable {
     @Override Expression implementSafe(final RexToLixTranslator translator,
         final RexCall call, final List<Expression> argValueList) {
       return Expressions.equal(argValueList.get(0), NULL_EXPR);
+    }
+  }
+
+  /** Implementor for the {@code IS NOT DISTINCT FROM} SQL operator. */
+  private static class IsNotDistinctFromImplementor extends AbstractRexCallImplementor {
+    IsNotDistinctFromImplementor() {
+      super("is_not_distinct_from", NullPolicy.NONE, false);
+    }
+
+    @Override public RexToLixTranslator.Result implement(final RexToLixTranslator translator,
+        final RexCall call, final List<RexToLixTranslator.Result> arguments) {
+      final RexToLixTranslator.Result left = arguments.get(0);
+      final RexToLixTranslator.Result right = arguments.get(1);
+
+      // Generated expression:
+      // left IS NULL ?
+      //   (right IS NULL ? TRUE : FALSE) :  -> when left is null
+      //   (right IS NULL ? FALSE :          -> when left is not null
+      //     left.equals(right))             -> when both are not null, compare values
+      final Expression valueExpression =
+          Expressions.condition(left.isNullVariable,
+          Expressions.condition(right.isNullVariable, BOXED_TRUE_EXPR, BOXED_FALSE_EXPR),
+          Expressions.condition(right.isNullVariable, BOXED_FALSE_EXPR,
+              Expressions.equal(left.valueVariable, right.valueVariable)));
+
+      BlockBuilder builder = translator.getBlockBuilder();
+      final ParameterExpression valueVariable =
+          Expressions.parameter(valueExpression.getType(),
+              builder.newName(variableName + "_value"));
+      final ParameterExpression isNullVariable =
+          Expressions.parameter(Boolean.TYPE,
+              builder.newName(variableName + "_isNull"));
+
+      builder.add(
+          Expressions.declare(Modifier.FINAL, valueVariable, valueExpression));
+      builder.add(
+          Expressions.declare(Modifier.FINAL, isNullVariable, FALSE_EXPR));
+
+      return new RexToLixTranslator.Result(isNullVariable, valueVariable);
+    }
+
+    @Override Expression implementSafe(final RexToLixTranslator translator,
+        final RexCall call, final List<Expression> argValueList) {
+      throw new IllegalStateException("This implementSafe should not be called,"
+          + " please call implement(...)");
     }
   }
 
--- a/core/src/main/java/org/apache/calcite/rel/core/JoinInfo.java
+++ b/core/src/main/java/org/apache/calcite/rel/core/JoinInfo.java
@@ -38,7 +38,7 @@ import static java.util.Objects.requireNonNull;
  * <p>It is useful for the many algorithms that care whether a join has an
  * equi-join condition.
  *
- * <p>You can create one using {@link #of}, or call
+ * <p>You can create one using {@link #createWithStrictEquality}, or call
  * {@link Join#analyzeCondition()}; many kinds of join cache their
  * join info, especially those that are equi-joins.
  *
@@ -66,6 +66,22 @@ public class JoinInfo {
     final List<RexNode> nonEquiList = new ArrayList<>();
     RelOptUtil.splitJoinCondition(left, right, condition, leftKeys, rightKeys,
         filterNulls, nonEquiList);
+    return new JoinInfo(ImmutableIntList.copyOf(leftKeys),
+        ImmutableIntList.copyOf(rightKeys), ImmutableList.copyOf(nonEquiList));
+  }
+
+  /** Creates a {@code JoinInfo} by analyzing a condition.
+   * A JoinInfo created using this method only considers
+   * EQUALS operations equi-conditions; in contrast
+   * IS_NOT_DISTINCT_FROM operations are NOT considered equalities;
+   * they are inserted in the nonEquiList. */
+  public static JoinInfo createWithStrictEquality(RelNode left,
+      RelNode right, RexNode condition) {
+    final List<Integer> leftKeys = new ArrayList<>();
+    final List<Integer> rightKeys = new ArrayList<>();
+    final List<RexNode> nonEquiList = new ArrayList<>();
+    RelOptUtil.splitJoinCondition(left, right, condition, leftKeys, rightKeys,
+        null, nonEquiList);
     return new JoinInfo(ImmutableIntList.copyOf(leftKeys),
         ImmutableIntList.copyOf(rightKeys), ImmutableList.copyOf(nonEquiList));
   }
--- a/core/src/test/java/org/apache/calcite/test/JdbcTest.java
+++ b/core/src/test/java/org/apache/calcite/test/JdbcTest.java
@@ -3950,6 +3950,24 @@ public class JdbcTest {
         .returnsUnordered("empid=150; name=Sebastian");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6904">[CALCITE-6904]
+   * IS_NOT_DISTINCT_FROM is converted error in EnumerableJoinRule</a>. */
+  @Test void testIsNotDistinctFrom() {
+    final String sql = ""
+        + "select \"t1\".\"commission\" from \"hr\".\"emps\" as \"t1\"\n"
+        + "join\n"
+        + "\"hr\".\"emps\" as \"t2\"\n"
+        + "on \"t1\".\"commission\" is not distinct from \"t2\".\"commission\"";
+    CalciteAssert.hr()
+        .query(sql)
+        .explainContains("NestedLoopJoin(condition=[IS NOT DISTINCT FROM($0, $1)]")
+        .returnsUnordered("commission=1000",
+            "commission=250",
+            "commission=500",
+            "commission=null");
+  }
+
   @Test void testExcept() {
     final String sql = ""
         + "select \"empid\", \"name\" from \"hr\".\"emps\" where \"deptno\"=10\n"
--- a/core/src/main/java/org/apache/calcite/rel/core/Join.java
+++ b/core/src/main/java/org/apache/calcite/rel/core/Join.java
@@ -104,7 +104,7 @@ public abstract class Join extends BiRel implements Hintable {
     this.condition = requireNonNull(condition, "condition");
     this.variablesSet = ImmutableSet.copyOf(variablesSet);
     this.joinType = requireNonNull(joinType, "joinType");
-    this.joinInfo = JoinInfo.of(left, right, condition);
+    this.joinInfo = JoinInfo.createWithStrictEquality(left, right, condition);
     this.hints = ImmutableList.copyOf(hints);
   }
 
--- a/core/src/main/java/org/apache/calcite/rel/rules/LoptOptimizeJoinRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/LoptOptimizeJoinRule.java
@@ -2053,7 +2053,7 @@ public class LoptOptimizeJoinRule
    */
   private static boolean areSelfJoinKeysUnique(RelMetadataQuery mq,
       RelNode leftRel, RelNode rightRel, RexNode joinFilters) {
-    final JoinInfo joinInfo = JoinInfo.of(leftRel, rightRel, joinFilters);
+    final JoinInfo joinInfo = JoinInfo.createWithStrictEquality(leftRel, rightRel, joinFilters);
 
     // Make sure each key on the left maps to the same simple column as the
     // corresponding key on the right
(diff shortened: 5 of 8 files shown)


core/src/main/java/org/apache/calcite/rel/core/JoinInfo.java
```diff
@@ -50,30 +52,59 @@ public class JoinInfo {
 
   /** Creates a JoinInfo. */
   protected JoinInfo(ImmutableIntList leftKeys, ImmutableIntList rightKeys,
-      ImmutableList<RexNode> nonEquiConditions) {
-    this.leftKeys = requireNonNull(leftKeys, "leftKeys");
-    this.rightKeys = requireNonNull(rightKeys, "rightKeys");
-    this.nonEquiConditions =
-        requireNonNull(nonEquiConditions, "nonEquiConditions");
+      ImmutableList<RexNode> nonEquiConditions, ImmutableList<Boolean> filterNulls,
```
mihaibudiu, line 55: changing the signature of protected methods is a breaking change, which will affect projects that use calcite.
I don't understand why JoinInfo needs to be changed if the problem is in the EnumerableJoinRule.

mihaibudiu: What is the problem of joins and null handling?
Joins can use arbitrary predicates.
What is an "equality condition"?

xiedeyantu: Maybe you can the details in jira 6904. leftKeys and rightKeys in JoinInfo contain equi condition.

mihaibudiu: CALCITE-6904 does not say anything about standard joins, it says there is a problem with EnumerableJoinRule

xiedeyantu: In EnumerableJoinRule, JoinInfo is used to split the Equi and non-equi conditions, and then the two are combined as the join condition. And then will use HashJoin. If it does not do such a split, it is fine to use NestedLoopJoin directly with the original join condition.

xiedeyantu: Maybe modify RelOptUtil.splitJoinCondition is a right way?

mihaibudiu: Maybe the question boils down at how HashJoin is supposed to be implemented.
In our implementation when the condition is equality (=) we filter out nulls prior to the join.
If the condition is `IS NOT DISTINCT FROM` we don't filter the nulls.
You can use a HashJoin in both cases.

xiedeyantu: So you mean that it is reasonable for us to modify the execution logic of hashjoin? You said before that we should not change this.

mihaibudiu: I am not sure what "the execution logic of hashjoin" is. I am not sure where I made that statement, I could not find it in the JIRA case or this conversation. 

There must be some conversion rules in the Enumerable implementation which decide when and how to implement a LogicalJoin using a HashJoin - I haven't looked at that part of the code. Maybe you are suggesting that the logic in that transformation is incorrect or inefficient.

xiedeyantu: > Maybe the question boils down at how HashJoin is supposed to be implemented. In our implementation when the condition is equality (=) we filter out nulls prior to the join. If the condition is `IS NOT DISTINCT FROM` we don't filter the nulls. You can use a HashJoin in both cases.

I mean, do we need to modify the logic here?

mihaibudiu: I am tempted to say "no" - JoinInfo is in the rel/core package, and has nothing to do with the enumerable implementation. I think any changes belong to that package.

xiedeyantu: I wonder if it would be better to treat IS NOT DISTINCT FROM as a non-equi condition and use NestedLoopJoin to execute it? I think it should be possible to implement it by modifying RelOptUtil.splitJoinCondition without modifying JoinInfo.

mihaibudiu: There are lots of tests using joins and Enumerable which work.
What does not work?

xiedeyantu: Maybe here https://github.com/apache/calcite/blob/c72e267f9c1b94abb3f4a7e31d14f7b4ed676ce1/core/src/main/java/org/apache/calcite/plan/RelOptUtil.java#L1697

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -3950,6 +3950,24 @@ public void checkOrderBy(final boolean desc,
...
+   * IS_NOT_DISTINCT_FROM is converted error in EnumerableJoinRule</a>. */
+  @Test void testIsNotDistinctFrom() {
+    final String sql = ""
+        + "select \"t1\".\"commission\" from \"hr\".\"emps\" as \"t1\"\n"
+        + "join\n"
+        + "\"hr\".\"emps\" as \"t2\"\n"
+        + "on \"t1\".\"commission\" is not distinct from \"t2\".\"commission\"";
+    CalciteAssert.hr()
+        .query(sql)
+        .explainContains("")
```
mihaibudiu, line 3964: why have an empty string for the explain?

core/src/main/java/org/apache/calcite/rel/core/JoinInfo.java
```diff
@@ -66,8 +66,36 @@ public static JoinInfo of(RelNode left, RelNode right, RexNode condition) {
     final List<RexNode> nonEquiList = new ArrayList<>();
     RelOptUtil.splitJoinCondition(left, right, condition, leftKeys, rightKeys,
         filterNulls, nonEquiList);
-    return new JoinInfo(ImmutableIntList.copyOf(leftKeys),
-        ImmutableIntList.copyOf(rightKeys), ImmutableList.copyOf(nonEquiList));
+
+    /**
+     * In join implementation when the condition is equality (=),
+     * we filter out nulls prior to the join. If the condition is
```
mihaibudiu, line 72: There is no "we".
Moreover, this code has nothing to do with filtering.
This code is about building a description of the join predicate.
I think the JoinInfo class needs better documentation; in particular, it is not clear what the field nonEquiConditions contains. Perhaps some of this JavaDoc should move to document the fields themselves. Then we would not need to discuss about whether IS NOT DISTINCT FROM is an equiCondition or not.

core/src/main/java/org/apache/calcite/rel/core/JoinInfo.java
```diff
@@ -66,8 +66,36 @@ public static JoinInfo of(RelNode left, RelNode right, RexNode condition) {
...
+    final List<Integer> rightKeysRewritten = new ArrayList<>();
+    final List<RexNode> nonEquiListRewritten = new ArrayList<>();
+    nonEquiListRewritten.addAll(nonEquiList);
+    for (int i = 0; i < filterNulls.size(); i++) {
+      if (filterNulls.get(i)) {
+        leftKeysRewritten.add(leftKeys.get(i));
+        rightKeysRewritten.add(rightKeys.get(i));
+      } else {
+        nonEquiListRewritten.add(
+            // Until IsNotDistinctFromImplementor is implemented,
```
mihaibudiu, line 90: If this code is just a temporary patch, wouldn't it be easier to implement the IsNotDistinctFromImplementor and not change this method?

core/src/main/java/org/apache/calcite/rel/core/JoinInfo.java
xiedeyantu: You are right, this description should be removed.

core/src/main/java/org/apache/calcite/rel/core/JoinInfo.java
xiedeyantu: Previously mentioned in [CALCITE-6902](https://issues.apache.org/jira/browse/CALCITE-6902) that IsNotDistinctFromImplementor is temporarily not implemented, so I think it's time to implement it. I will try to implement an IsNotDistinctFromImplementor.

xiedeyantu:
@mihaibudiu May I trouble you to help me check if my modifications match your comments.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -425,6 +425,7 @@
 import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_JSON_SCALAR;
 import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_JSON_VALUE;
 import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_NOT_A_SET;
+import static org.apache.calcite.sql.fun.SqlStdOperatorTable.IS_NOT_DISTINCT_FROM;
```
mihaibudiu, line 428: what about IS_DISTINCT_FROM?
Does that need an implementor, or is it converted to IS_NOT_DISTINCT_FROM?

core/src/main/java/org/apache/calcite/rel/core/JoinInfo.java
```diff
@@ -62,10 +62,9 @@ protected JoinInfo(ImmutableIntList leftKeys, ImmutableIntList rightKeys,
   public static JoinInfo of(RelNode left, RelNode right, RexNode condition) {
     final List<Integer> leftKeys = new ArrayList<>();
     final List<Integer> rightKeys = new ArrayList<>();
-    final List<Boolean> filterNulls = new ArrayList<>();
     final List<RexNode> nonEquiList = new ArrayList<>();
     RelOptUtil.splitJoinCondition(left, right, condition, leftKeys, rightKeys,
-        filterNulls, nonEquiList);
+        null, nonEquiList);
```
mihaibudiu, line 67: is this change needed?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4741,6 +4743,47 @@ private static class IsNullImplementor extends AbstractRexCallImplementor {
...
+      builder.add(
+          Expressions.declare(Modifier.FINAL, isNullVariable, isNullExpression));
+
+      return new RexToLixTranslator.Result(isNullVariable, valueVariable);
+    }
+
+    @Override Expression implementSafe(final RexToLixTranslator translator,
+        final RexCall call, final List<Expression> argValueList) {
+      throw new IllegalStateException("This implementSafe should not be called,"
+          + " please call implement(...)");
```
mihaibudiu, line 4783: I am not sure this error message is helpful - who is the "please" for?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4741,6 +4743,47 @@ private static class IsNullImplementor extends AbstractRexCallImplementor {
...
 
+  /** Implementor for the {@code IS NOT DISTINCT FROM} SQL operator. */
+  private static class IsNotDistinctFromImplementor extends AbstractRexCallImplementor {
+    IsNotDistinctFromImplementor() {
+      super("is_not_distinct_from", NullPolicy.NONE, false);
+    }
+
+    @Override public RexToLixTranslator.Result implement(final RexToLixTranslator translator,
+        final RexCall call, final List<RexToLixTranslator.Result> arguments) {
+      final RexToLixTranslator.Result left = arguments.get(0);
```
mihaibudiu, line 4754: can you add a comment showing the generated java code?

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
```diff
@@ -4741,6 +4743,47 @@ private static class IsNullImplementor extends AbstractRexCallImplementor {
...
+          Expressions.condition(right.isNullVariable, BOXED_TRUE_EXPR, BOXED_FALSE_EXPR),
+          Expressions.condition(right.isNullVariable, BOXED_FALSE_EXPR,
+              Expressions.equal(left.valueVariable, right.valueVariable)));
+
+      BlockBuilder builder = translator.getBlockBuilder();
+      final ParameterExpression valueVariable =
+          Expressions.parameter(valueExpression.getType(),
+              builder.newName(variableName + "_value"));
+      final Expression isNullExpression = FALSE_EXPR;
+      final ParameterExpression isNullVariable =
```
mihaibudiu, line 4768: is this declaration necessary or can you just use FALSE_EXPR below?

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -3950,6 +3950,24 @@ public void checkOrderBy(final boolean desc,
         .returnsUnordered("empid=150; name=Sebastian");
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6904">[CALCITE-6904]
+   * IS_NOT_DISTINCT_FROM is converted error in EnumerableJoinRule</a>. */
+  @Test void testIsNotDistinctFrom() {
```
mihaibudiu, line 3956: Can you also add a quidem test in one .iq file?

mihaibudiu (commented):
This looks much better.

core/src/main/java/org/apache/calcite/rel/core/JoinInfo.java
xiedeyantu: Yes, if filterNulls is not null, splitJoinCondition method will take IS NOT DISTINCT FROM as "equi condition". If filterNulls is null, it can skip  IS NOT DISTINCT FROM correctly.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
xiedeyantu: I will add comment later.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
xiedeyantu: Here we need to Implement `implementSafe` method, I referred to other implementations and also threw an error here.

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
xiedeyantu: IS_DISTINCT_FROM will expand to `x=y IS NOT TRUE and ((x is not null) or (y is not null))`. There is currently no method to collapse it, so it may not be necessary to implement it temporarily?

core/src/test/java/org/apache/calcite/test/JdbcTest.java
xiedeyantu: Done

core/src/main/java/org/apache/calcite/adapter/enumerable/RexImpTable.java
xiedeyantu: Fixed

core/src/main/java/org/apache/calcite/rel/core/JoinInfo.java
mihaibudiu: I am worried about this change.
This is a breaking change in a public API.
Users of this method will see different behaviors.
Where is this method called when the EnumerableJoinRule fires?

suibianwanwank: I haven't reviewed the code in detail, but my guess is the issue arises from the equivalent conditional determination process, where non-equi conditions are fetched from the joinInfo(ggenerated during the join construction) and treated as equivCond if they aren't in the list.

If that's the case, IMO a  better approach would be to record the list of nullFilter in the JoinInfo. And use nullFilter to fix where you fundamentally want to fix the problem🤔️.

mihaibudiu: wouldn't that be also an incompatible change?
this may be backwards compatible only if new fields are added and the content of the old fields would not be changed

xiedeyantu: In my opinion, the purpose is to allow IS NOT DISTINCT FROM expressions to be handled correctly. The splitJoinCondition method is included in the JoinInfo.of method. JoinInfo.of is only used in the implementation class of EnumerableXxxJoin, that is, when handling equality and non-equality conditions, so it may not have any impact.

mihaibudiu: You cannot see the code that is not in this repository

mihaibudiu: If this function has no other uses, then it's easy: just define a new function to create a JoinInfo the way you like it, and call that function from EnumerableXxxJoin. I would also document the fields of JoinInfo to explain what they contain - what is considered equality.

xiedeyantu: I understand what you mean, but now there is a problem. If I remember correctly, there are three places that call JoinInfo.of, involving HashJoin, NestedLoopJoin, MergeJoin. These three places need to be replaced. If I create a new function (such as JoinInfo.create), then JoinInfo.of will not have other methods calling it. Do I still need to create a new method?

mihaibudiu: As I said, there may be a tool built on Calcite which calls this method.
So changing its behavior is a breaking change.

xiedeyantu: Oh, I understand now. Can I keep it unchanged and create a new JoinInfo.create to replace the current JoinInfo.of? Also, do I need to modify other places besides here?

mihaibudiu: Yes, but I would name the function `createWithStrictEquality`, or something like this, and explain how it differs from `of`.
I think that if all the tests pass you are fine - if you don't change public interfaces.
After your change `JoinInfo` will represent two slightly different data structures, depending on how it's created, so it is very important to document clearly how the result created by `of` differs from the result created by your function.

xiedeyantu: Got it, thanks. I will change it soon.

xiedeyantu:
@mihaibudiu I have finished, please have a look, thanks.

mihaibudiu (approved):

mihaibudiu:
This is almost good to merge, I only made two suggestions for improving the documentation.

caicancai (approved):

xiedeyantu:
> This is almost good to merge, I only made two suggestions for improving the documentation.

Hi, I want to confirm with you again, is there anything else I should modify in the document you mentioned?

mihaibudiu (approved):

mihaibudiu:
Look like my comments didn't make it to github; maybe I forgot to press the right button.
I apologize about that.

xiedeyantu:
> Look like my comments didn't make it to github; maybe I forgot to press the right button. I apologize about that.

Thanks, I have addressed.

PR #4264, opened 2025-03-25
[CALCITE-6911] SqlItemOperator.inferReturnType throws AssertionError for out of bounds accesses

mihaibudiu:


mihaibudiu:
I have kept the assertions unchanged, but added additional validation which should prevent these assertions from ever being fired. The validation has to be done in `checkOperandTypes` to provide nice validator error messages; by the time `inferReturnTypes` is reached we no longer have a handle to the required callBinding.

mihaibudiu:
To make it easier to review, please note that the new code in SqlItemOperator is essentially lifted from the existing code from SqlItemOperator.inferReturnType; it is just that these checks should be done earlier, and should not throw AssertionError, but report a normal error.

caicancai (approved):
LGTM

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -13289,10 +13289,9 @@ private static void checkArrayConcatAggFuncFails(SqlOperatorFixture t) {
     f.check("select \"T\".\"X\"[2] "
             + "from (VALUES (ROW(ROW(3, CAST(NULL AS INTEGER)), ROW(4, 8)))) as T(x, y)",
         SqlTests.ANY_TYPE_CHECKER, isNullValue());
-    f.checkFails("select \"T\".\"X\"[1 + CAST(NULL AS INTEGER)] "
+    f.checkFails("select ^\"T\".\"X\"[1 + CAST(NULL AS INTEGER)]^ "
```
NobiGo, line 13292: Can we be more precise here, such as displaying `1+CAST (NULL AS INTEGER)` in the exception message?

mihaibudiu: The position is included, that's what ^ stands for in these tests. The position is between the two carets.

PR #4267, opened 2025-03-26
[CALCITE-6914] Expand join-dependent predicates from disjunction

silundong:
For the following SQL:
`
select t1.name from t1, t2 
where t1.id = t2.id 
and ( 
    (t1.id > 20 and t2.height < 50)  
    or 
    (t1.weight < 200 and t2.sales > 100) 
) 
`
The disjunction `(t1.id > 20 and t2.height < 50) or (t1.weight < 200 and t2.sales > 100)` uses t1 and t2, thus cannot be pushed below the join. 
In fact, we can extract redundant `(t1.id > 20 or t1.weight < 200)` , `(t2.height < 50 or t2.sales > 100)` from this predicate by a new rule and push down these two predicates to both side of Join input.
More detail: https://issues.apache.org/jira/browse/CALCITE-6914

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
@@ -850,4 +850,16 @@ public class CoreRules {
   @Experimental
   public static final DphypJoinReorderRule HYPER_GRAPH_OPTIMIZE =
       DphypJoinReorderRule.Config.DEFAULT.toRule();
+
+  /** Rule that expands disjunction in the condition of a {@link Filter}.
+   *
+   * @see #EXPAND_JOIN_DISJUNCTION_GLOBAL */
+  public static final ExpandDisjunctionForTableRule EXPAND_FILTER_DISJUNCTION_GLOBAL =
+      ExpandDisjunctionForTableRule.Config.FILTER.toRule();
+
+  /** Rule that expands disjunction in the condition of a {@link Join}.
+   *
+   * @see #EXPAND_FILTER_DISJUNCTION_GLOBAL */
+  public static final ExpandDisjunctionForTableRule EXPAND_JOIN_DISJUNCTION_GLOBAL =
+      ExpandDisjunctionForTableRule.Config.JOIN.toRule();
 }
--- a/core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
@@ -1,0 +1,430 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelOptUtil;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Filter;
+import org.apache.calcite.rel.core.Join;
+import org.apache.calcite.rel.metadata.RelMetadataQuery;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexInputRef;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexTableInputRef;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.tools.RelBuilder;
+import org.apache.calcite.util.ControlFlowException;
+import org.apache.calcite.util.ImmutableBitSet;
+
+import com.google.common.collect.Lists;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.immutables.value.Value;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Rule to expand disjunction for single table in condition of a {@link Filter} or {@link Join},
+ * It makes sense to make this optimization part of the predicate pushdown. For example:
+ *
+ * <blockquote><pre>{@code
+ * select t1.name from t1, t2
+ * where t1.id = t2.id
+ * and (
+ *  (t1.id > 20 and t2.height < 50)
+ *  or
+ *  (t1.weight < 200 and t2.sales > 100)
+ * )
+ * }</pre></blockquote>
+ *
+ * <p>we can expand to obtain the condition
+ *
+ * <blockquote><pre>{@code
+ * select t1.name from t1, t2
+ * where t1.id = t2.id
+ * and (
+ *  (t1.id > 20 and t2.height < 50)
+ *  or
+ *  (t1.weight < 200 and t2.sales > 100)
+ * )
+ * and (t1.id > 20 or t1.weight < 200)
+ * and (t2.height < 50 or t2.sales > 100)
+ * }</pre></blockquote>
+ *
+ * <p>new generated predicates are redundant, but they could be pushed down to
+ * scan operator of t1/t2 and reduce the cardinality.
+ *
+ * <p>Note: This rule should only be applied once and it should be used before
+ * {@link CoreRules#FILTER_INTO_JOIN} and {@link CoreRules#JOIN_CONDITION_PUSH}.
+ * In order to avoid it interacting with other expression-related rules and causing
+ * an infinite loop, it is recommended to use it alone in HepPlanner first,
+ * and then do predicate pushdown.
+ *
+ * @see CoreRules#EXPAND_FILTER_DISJUNCTION_GLOBAL
+ * @see CoreRules#EXPAND_JOIN_DISJUNCTION_GLOBAL
+ */
+@Value.Enclosing
+public class ExpandDisjunctionForTableRule
+    extends RelRule<ExpandDisjunctionForTableRule.Config>
+    implements TransformationRule {
+
+  /**
+   * Creates a ExpandDisjunctionForTableRule.
+   */
+  protected ExpandDisjunctionForTableRule(Config config) {
+    super(config);
+  }
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    config.matchHandler().accept(this, call);
+  }
+
+  /**
+   * Expand predicates for condition of a {@link Filter} or {@link Join}.
+   *
+   * @param condition   Condition to be expanded in Filter or Join
+   * @param relNode     The Filter or Join node
+   * @param fieldList   The field list of the Filter or Join inputs.
+   * @param relBuilder  Builder
+   * @param mq          RelMetadataQuery, which is used to get the expression lineage
+   *
+   * @return The redundant predicate expanded from the condition parameter. It is formed by
+   * concatenating the predicates that can be pushed down to single table with AND.
+   */
+  private RexNode apply(
+      RexNode condition,
+      RelNode relNode,
+      List<RelDataTypeField> fieldList,
+      RelBuilder relBuilder,
+      RelMetadataQuery mq) {
+    final Map<Integer, RexTableInputRef.RelTableRef> inputRefToTableRef = new HashMap<>();
+
+    ImmutableBitSet columnBits = RelOptUtil.InputFinder.bits(condition);
+    if (columnBits.isEmpty()) {
+      return condition;
+    }
+
+    // For each table discover the conditions that refer to columns of that
+    // table (and refer to no other tables)
+    for (int columnBit : columnBits.asList()) {
+      Set<RexNode> exprLineage =
+          mq.getExpressionLineage(
+              relNode,
+              RexInputRef.of(columnBit, fieldList));
+      // If mq.getExpressionLineage cannot get result, skip it
+      if (exprLineage == null) {
+        continue;
+      }
+      Set<RexTableInputRef.RelTableRef> relTableRefs =
+          RexUtil.gatherTableReferences(Lists.newArrayList(exprLineage));
+      // If the column is computed by an expression that depends on multiple tables, skip it
+      if (relTableRefs.size() != 1) {
+        continue;
+      }
+      inputRefToTableRef.put(columnBit, relTableRefs.iterator().next());
+    }
+
+    ExpandDisjunctionForTableHelper expandHelper =
+        new ExpandDisjunctionForTableHelper(inputRefToTableRef, relBuilder, config.bloat());
+    Map<RexTableInputRef.RelTableRef, RexNode> expandResult = expandHelper.expand(condition);
+
+    RexNode newCondition = condition;
+    for (RexNode expandCondition : expandResult.values()) {
+      newCondition = relBuilder.and(newCondition, expandCondition);
+    }
+    return newCondition;
+  }
+
+  private static void matchFilter(ExpandDisjunctionForTableRule rule, RelOptRuleCall call) {
+    Filter filter = call.rel(0);
+    RelMetadataQuery mq = call.getMetadataQuery();
+    RelBuilder relBuilder = call.builder();
+
+    RexNode newCondition =
+        rule.apply(
+            filter.getCondition(),
+            filter,
+            filter.getRowType().getFieldList(),
+            relBuilder,
+            mq);
+    if (newCondition.equals(filter.getCondition())) {
+      return;
+    }
+    Filter newFilter = filter.copy(filter.getTraitSet(), filter.getInput(), newCondition);
+    call.transformTo(newFilter);
+  }
+
+  private static void matchJoin(ExpandDisjunctionForTableRule rule, RelOptRuleCall call) {
+    Join join = call.rel(0);
+    RelMetadataQuery mq = call.getMetadataQuery();
+    RelBuilder relBuilder = call.builder();
+
+    List<RelDataTypeField> fieldList =
+        Lists.newArrayList(join.getLeft().getRowType().getFieldList());
+    fieldList.addAll(join.getRight().getRowType().getFieldList());
+    RexNode newCondition =
+        rule.apply(
+            join.getCondition(),
+            join,
+            fieldList,
+            relBuilder,
+            mq);
+    if (newCondition.equals(join.getCondition())) {
+      return;
+    }
+    Join newJoin =
+        join.copy(
+            join.getTraitSet(),
+            newCondition,
+            join.getLeft(),
+            join.getRight(),
+            join.getJoinType(),
+            join.isSemiJoinDone());
+    call.transformTo(newJoin);
+  }
+
+  /**
+   * Helper class to expand predicates.
+   */
+  private static class ExpandDisjunctionForTableHelper {
+    private final Map<Integer, RexTableInputRef.RelTableRef> inputRefToTableRef;
+
+    private final RelBuilder relBuilder;
+
+    private final int maxNodeCount;
+
+    // Used to record the number of redundant expressions expanded.
+    private int currentCount;
+
+    private ExpandDisjunctionForTableHelper(
+        Map<Integer, RexTableInputRef.RelTableRef> inputRefToTableRef,
+        RelBuilder relBuilder,
+        int maxNodeCount) {
+      this.inputRefToTableRef = inputRefToTableRef;
+      this.relBuilder = relBuilder;
+      this.maxNodeCount = maxNodeCount;
+    }
+
+    private Map<RexTableInputRef.RelTableRef, RexNode> expand(RexNode condition) {
+      try {
+        this.currentCount = 0;
+        return expandDeep(condition);
+      } catch (OverLimitException e) {
+        return new HashMap<>();
+      }
+    }
+
+    /**
+     * Expand predicates recursively that can be pushed down to single table.
+     *
+     * @param condition   Predicate to be expanded
+     * @return  A map from a table to a (combined) predicate that can be pushed down
+     * and only depends on columns of the table
+     */
+    private Map<RexTableInputRef.RelTableRef, RexNode> expandDeep(RexNode condition) {
+      Map<RexTableInputRef.RelTableRef, RexNode> additionalConditions = new HashMap<>();
+
+      ImmutableBitSet inputRefs = RelOptUtil.InputFinder.bits(condition);
+      if (inputRefs.isEmpty()) {
+        return additionalConditions;
+      }
+
+      RexTableInputRef.RelTableRef tableRef = inputRefsBelongToOneTable(inputRefs);
+      if (tableRef != null) {
+        // The condition already belongs to one table, return it directly
+        checkExpandCount(condition.nodeCount());
+        additionalConditions.put(tableRef, condition);
+        return additionalConditions;
+      }
+
+      // Recursively expand the expression according to whether it is a conjunction
+      // or a disjunction. If it is neither a disjunction nor a conjunction, it cannot
+      // be expanded further and an empty Map is returned.
+      switch (condition.getKind()) {
+      case AND:
+        List<RexNode> andOperands = RexUtil.flattenAnd(((RexCall) condition).getOperands());
+        for (RexNode andOperand : andOperands) {
+          Map<RexTableInputRef.RelTableRef, RexNode> operandResult = expandDeep(andOperand);
+          combinePredicatesUsingAnd(additionalConditions, operandResult);
+        }
+        break;
+      case OR:
+        List<RexNode> orOperands = RexUtil.flattenOr(((RexCall) condition).getOperands());
+        additionalConditions.putAll(expandDeep(orOperands.get(0)));
+        for (int i = 1; i < orOperands.size(); i++) {
+          Map<RexTableInputRef.RelTableRef, RexNode> operandResult =
+              expandDeep(orOperands.get(i));
+          combinePredicatesUsingOr(additionalConditions, operandResult);
+
+          if (additionalConditions.isEmpty()) {
+            break;
+          }
+        }
+        break;
+      default:
+        break;
+      }
+
+      return additionalConditions;
+    }
+
+    private RexTableInputRef.@Nullable RelTableRef inputRefsBelongToOneTable(
+        ImmutableBitSet inputRefs) {
+      RexTableInputRef.RelTableRef tableRef = inputRefToTableRef.get(inputRefs.nth(0));
+      if (tableRef == null) {
+        return null;
+      }
+      for (int inputBit : inputRefs.asList()) {
+        RexTableInputRef.RelTableRef inputBitTableRef = inputRefToTableRef.get(inputBit);
+        if (!tableRef.equals(inputBitTableRef)) {
+          return null;
+        }
+      }
+      return tableRef;
+    }
+
+    /**
+     * Combine predicates that depend on the same table using conjunctions.
+     * The result is returned by modifying baseMap. For example:
+     *
+     * <p>baseMap: {t1: p1, t2: p2}
+     *
+     * <p>forMergeMap: {t1: p11, t3: p3}
+     *
+     * <p>result: {t1: p1 AND p11, t2: p2, t3: p3}
+     *
+     * @param baseMap       Additional predicates that current conjunction has already saved
+     * @param forMergeMap   Additional predicates that current operand has expanded
+     */
+    private void combinePredicatesUsingAnd(
+        Map<RexTableInputRef.RelTableRef, RexNode> baseMap,
+        Map<RexTableInputRef.RelTableRef, RexNode> forMergeMap) {
+      for (Map.Entry<RexTableInputRef.RelTableRef, RexNode> entry : forMergeMap.entrySet()) {
+        RexNode mergedRex =
+            relBuilder.and(
+                entry.getValue(),
+                baseMap.getOrDefault(entry.getKey(), relBuilder.literal(true)));
+        int oriCount = entry.getValue().nodeCount()
+            + (baseMap.containsKey(entry.getKey())
+                ? baseMap.get(entry.getKey()).nodeCount()
+                : 0);
+        checkExpandCount(mergedRex.nodeCount() - oriCount);
+        baseMap.put(entry.getKey(), mergedRex);
+      }
+    }
+
+    /**
+     * Combine predicates that depend on the same table using disjunctions.
+     * The result is returned by modifying baseMap. For example:
+     *
+     * <p>baseMap: {t1: p1, t2: p2}
+     *
+     * <p>forMergeMap: {t1: p11, t3: p3}
+     *
+     * <p>result: {t1: p1 OR p11}
+     *
+     * @param baseMap       Additional predicates that current disjunction has already saved
+     * @param forMergeMap   Additional predicates that current operand has expanded
+     */
+    private void combinePredicatesUsingOr(
+        Map<RexTableInputRef.RelTableRef, RexNode> baseMap,
+        Map<RexTableInputRef.RelTableRef, RexNode> forMergeMap) {
+      if (baseMap.isEmpty()) {
+        return;
+      }
+
+      Iterator<Map.Entry<RexTableInputRef.RelTableRef, RexNode>> iterator =
+          baseMap.entrySet().iterator();
+      while (iterator.hasNext()) {
+        int forMergeNodeCount = 0;
+
+        Map.Entry<RexTableInputRef.RelTableRef, RexNode> entry = iterator.next();
+        if (!forMergeMap.containsKey(entry.getKey())) {
+          checkExpandCount(-entry.getValue().nodeCount());
+          iterator.remove();
+          continue;
+        } else {
+          forMergeNodeCount = forMergeMap.get(entry.getKey()).nodeCount();
+        }
+        RexNode mergedRex =
+            relBuilder.or(
+                entry.getValue(),
+                forMergeMap.get(entry.getKey()));
+        int oriCount = entry.getValue().nodeCount() + forMergeNodeCount;
+        checkExpandCount(mergedRex.nodeCount() - oriCount);
+        baseMap.put(entry.getKey(), mergedRex);
+      }
+    }
+
+    /**
+     * Check whether the number of redundant expressions generated in the expansion
+     * exceeds the limit.
+     */
+    private void checkExpandCount(int changeCount) {
+      currentCount += changeCount;
+      if (maxNodeCount > 0 && currentCount > maxNodeCount) {
+        throw OverLimitException.INSTANCE;
+      }
+    }
+
+    /** Exception to catch when we pass the limit. */
+    private static class OverLimitException extends ControlFlowException {
+      protected static final OverLimitException INSTANCE = new OverLimitException();
+
+      private OverLimitException() {}
+    }
+  }
+
+  /** Rule configuration. */
+  @Value.Immutable(singleton = false)
+  public interface Config extends RelRule.Config {
+    Config FILTER = ImmutableExpandDisjunctionForTableRule.Config.builder()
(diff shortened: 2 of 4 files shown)


mihaibudiu:
You can test your changes locally using `./gradlew build`

silundong:
> You can test your changes locally using `./gradlew build`

Thank you for your suggestion, the format of javadoc was wrong before. Could you please take a look at this PR?

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,335 @@
...
+        new ExpandDisjuctionHelper(inputRefToTableRef, relBuilder);
+    Map<RexTableInputRef.RelTableRef, RexNode> expandResult = expandHelper.expandDeep(condition);
+    RexNode newCondition = condition;
+    for (RexNode expandCondition : expandResult.values()) {
+      newCondition = relBuilder.and(newCondition, expandCondition);
+    }
+    return newCondition;
+  }
+
+  private static void matchFilter(ExpandDisjuctionRule rule, RelOptRuleCall call) {
```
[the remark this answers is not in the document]
silundong: Yes, it wasn't used. The interface MatchHandler extends BiConsumer:
`MatchHandler<R extends RelOptRule> extends BiConsumer<R, RelOptRuleCall>`
It must have two parameters: RelOptRule and RelOptRuleCall. I hope to use MatchHandler to easily match Filter and Join, but I don't need the RelOptRule parameter. 
If the rule parameter must be deleted here, do you have any good suggestions?

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    config.matchHandler().accept(this, call);
+  }
+
+  /**
+   * Expand predicates for condition of a {@link Filter} or {@link Join}.
+   */
+  private static RexNode apply(
```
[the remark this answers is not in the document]
silundong: I will update it.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,335 @@
...
+        additionalConditions.put(tableRef, condition);
+        return additionalConditions;
+      }
+
+      // Recursively expand the expression according to whether it is a conjunction
+      // or a disjunction. If it is neither a disjunction nor a conjunction, it cannot
+      // be expanded further and an empty Map is returned.
+      switch (condition.getKind()) {
+      case AND:
+        List<RexNode> andOperands = RexUtil.flattenAnd(((RexCall) condition).getOperands());
```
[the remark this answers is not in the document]
silundong: The complexity here depends on the size of the tree with AND and OR as non-leaf nodes and other expressions as leaf nodes. I don't think the tree would be particularly deep for usual SQL. The code currently has an early return logic, that is, the columns used in the current expression are all from the same table.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,335 @@
...
+      RelNode relNode,
+      List<RelDataTypeField> fieldList,
+      RelBuilder relBuilder,
+      RelMetadataQuery mq) {
+    HashMap<Integer, RexTableInputRef.RelTableRef> inputRefToTableRef = new HashMap<>();
+    ImmutableBitSet columnBits = RelOptUtil.InputFinder.bits(condition);
+    if (columnBits.isEmpty()) {
+      return condition;
+    }
+    // Trace which table does the column referenced in the condition come from
```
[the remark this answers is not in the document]
silundong: Taking matching to Filter as an example, I don’t know the structure of Joins under the current Filter, which columns belong to which Join, I only know which table a column comes from, so at this time I can only expand the additional predicates that belong to a single table from the disjunction and push them down to TableScan later.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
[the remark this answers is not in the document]
asolimando: In similar cases, whenever we risk a combinatorial explosion, we have a parameter to control it, because what everyone of us consider usual SQL varies a lot in my experience, and there are machine-generated SQL statements that can be 10MB in query text, we must have a way to impose a limit.

Does something like [this](https://github.com/search?q=repo%3Aapache%2Fcalcite%20bloat&type=code) would be applicable here?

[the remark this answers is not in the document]
silundong: Thank you for your further explanation, it seems that restrictions are necessary, I will update.

asolimando:
@silundong unless strictly necessary (CI changes from `main` that can't be picked up otherwise, etc.) please avoid force-pushing while the review is in progress as it makes it difficult for reviewers to do incremental diff via GitHub UI. You can push as many commits as needed, you will be required to squash them before merging.

silundong:
> @silundong unless strictly necessary (CI changes from `main` that can't be picked up otherwise, etc.) please avoid force-pushing while the review is in progress as it makes it difficult for reviewers to do incremental diff via GitHub UI. You can push as many commits as needed, you will be required to squash them before merging.

Sorry for making the review difficult. Should I revert the commit to the state of the last review and update the changes to the new commit?

silundong:
I have made some changes, could you please review it again? @asolimando @mihaibudiu

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+    config.matchHandler().accept(this, call);
+  }
+
+  /**
+   * Expand predicates for condition of a {@link Filter} or {@link Join}.
+   *
+   * @param condition   Condition to be expanded in Filter or Join
+   * @param relNode     The Filter or Join node
+   * @param fieldList   The field list of the Filter or Join inputs. Build the referenced columns
+   *                    in the predicate as RexInputRef and  find out which table the RexInputRef
```
asolimando, line 99:
```suggestion
   *                    in the predicate as RexInputRef and find out which table the RexInputRef
```

asolimando (commented):
@silundong I don't see anything trying to prevent the rule to be applied indefinitely. Now that we have the maximum expansion it might stop at some point but this won't prevent the rule to be tried. I have seen lots of similar problems leading to OOMs in the past, I think it's dangerous to integrate such a rule as-is, and most definitely I wouldn't put it into the standard core rules which get applied by default.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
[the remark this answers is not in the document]
mihaibudiu: If the parameter has to be there, just call it unusedCall.
This is a private function, I don't understand how it helps implement BiConsumer.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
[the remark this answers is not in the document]
mihaibudiu: Indeed, for filter this is true, but for joins it may be useful.
This could be a second optimization rule that you could add, which does not use tables to separate conditions, but only the join inputs. Would you consider implementing that? I think it could be very useful.

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
```diff
@@ -841,4 +841,16 @@ private CoreRules() {}
   @Experimental
   public static final DphypJoinReorderRule HYPER_GRAPH_OPTIMIZE =
       DphypJoinReorderRule.Config.DEFAULT.toRule();
+
+  /** Rule that expand disjuction in condition of a {@link Filter}.
+   *
+   * @see #EXPAND_JOIN_DISJUCTION */
+  public static final ExpandDisjuctionRule EXPAND_FILTER_DISJUCTION =
```
mihaibudiu, line 848: the word is DISJUNCTION, with an N

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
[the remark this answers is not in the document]
mihaibudiu: Are you still planning to add a blot control?
Please use new commits, to make reviewing easier, and then request a new review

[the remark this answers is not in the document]
silundong: @mihaibudiu I have added a restriction. As @asolimando  said, I directly force pushed and requested your second review. I will avoid force pushing for future commits. Sorry for the difficulty in reviewing this time.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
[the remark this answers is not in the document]
silundong: Yes, it could be another new rule. This new rule :
1. Only matches Join and expands the disjunction in the condition for both inputs.
2. Same as the notes for ExpandDisjuctionRule, it must be applied before FilterIntoJoinRule and only once (described in [jira 6914](https://issues.apache.org/jira/browse/CALCITE-6914)).

If the above two points are ok, I will try to implement it in the current PR. @mihaibudiu

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
silundong: I will update it.

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
silundong: I will update it.

mihaibudiu:
Please request another review when this is ready

silundong:
> @silundong I don't see anything trying to prevent the rule to be applied indefinitely. Now that we have the maximum expansion it might stop at some point but this won't prevent the rule to be tried. I have seen lots of similar problems leading to OOMs in the past, I think it's dangerous to integrate such a rule as-is, and most definitely I wouldn't put it into the standard core rules which get applied by default.

@asolimando I have checked in `matchFilter` and `matchJoin` whether the expanded newCondition is equal to the original condition. If they are the same, it will be returned directly. 
For example the rule is matched for the first time and the condition is successfully expanded, a new Filter/Join is generated. When the rule is matched for the second time, the conditions that have been successfully expanded will not change, so it will be returned directly. I don't think it will be applied infinitely. Is there something else I haven't considered?

mihaibudiu:
The concern is not infinite loops, but large increases.
For example, converting a boolean formula to CNF form can produce expressions that are exponential in the size of the original expression. Even a quadratic increase can be very large: if the original formula has 1000 terms, the result can have 1 million.
I have seen discussions about very large formulas that can arise in SQL programs.

silundong:
> The concern is not infinite loops, but large increases. For example, converting a boolean formula to CNF form can produce expressions that are exponential in the size of the original expression. Even a quadratic increase can be very large: if the original formula has 1000 terms, the result can have 1 million. I have seen discussions about very large formulas that can arise in SQL programs.

Thanks for your explanation. So it should count and check the upper bound every time it gets a new redundant predicate, right?

mihaibudiu:
If you can compute the size of the final formula based on the structure of the original formulas, then you can do that upfront and avoid expansion altogether. That's why I asked about the complexity. An alternative is to do the expansion work and if the result becomes too big abort and undo the transformation. @asolimando pointed to a BLOAT parameter which can control the maximum size.

silundong:
> If you can compute the size of the final formula based on the structure of the original formulas, then you can do that upfront and avoid expansion altogether. That's why I asked about the complexity. An alternative is to do the expansion work and if the result becomes too big abort and undo the transformation. @asolimando pointed to a BLOAT parameter which can control the maximum size.

Thank you, I will make changes in the next commit and try to add the rule we mentioned.

silundong:
@asolimando Got it. Your concerns are also what I have considered. As the notes I described in jira, this rule must be guaranteed to be applied only once to avoid generating same redundant expression. A simple example is that it cannot be in the same rule set as `FilterJoinRule`. We need to apply this rule first and then `FilterJoinRule`. Of course, as you said, other rules may also interact with this rule and cause an infinite loop, especially additional expression simplification.

I initially tried to perform disjunction expansion directly in `FilterJoinRule` to solve the [TODO](https://github.com/apache/calcite/blob/b394abca407eea06ee8f8b50f30bd443f286bb42/core/src/main/java/org/apache/calcite/rel/rules/FilterJoinRule.java#L105), but I encountered a similar problem. The simplification of the expression caused the `FilterJoinRule` rule to loop infinitely. So I think it is a better and safer solution to use a single rule for disjunction expansion before predicate pushdown. My original intention is also to try not to mix this rule with other rules, especially those related to expressions.

You mean that this rule should not be in `CoreRules` now, and it can exist alone, right?

asolimando:
> @asolimando Got it. Your concerns are also what I have considered. As the notes I described in jira, this rule must be guaranteed to be applied only once to avoid generating same redundant expression. A simple example is that it cannot be in the same rule set as `FilterJoinRule`. We need to apply this rule first and then `FilterJoinRule`. Of course, as you said, other rules may also interact with this rule and cause an infinite loop, especially additional operations for expressions.
> 
> My intention is that it is applied in HepPlanner and avoids interaction with other rules related to expression processing. It is applied separately, and then performs predicate pushdown, expression simplification, etc. This is indeed more strict, but I think it is safe and effective.
> 
> So should I remove it from `CoreRules` now (is that what 'not have it added to core right away and leave it as an optional rule' means?), or is there a better solution?

I apologize because my comment was indeed ambiguous, what I meant was not to remove it from `CoreRules`, but to make sure the rule doesn't end up in one the sets of standard rules as defined in `RelOptRules`, this is not the case at the moment.

I think that a word of caution could be added as a comment in the javadoc of the rule itself, briefly describing known issues (what rules could be problematic when paired with this one, how to limit the danger via the bloat parameter, etc.), this will help users understanding the implications of adding this rule, especially in combination with some others known to cause infinite loops.

In the current javadoc, it is correctly suggested to use this rule before `CoreRules#FILTER_INTO_JOIN` and `CoreRules#JOIN_CONDITION_PUSH`, maybe we can be more explicit suggesting the use of the `HepPlanner` too?

It can help for discussions like [this one](https://lists.apache.org/thread/0yotdv62r6b3wg1qztftdpnkzpc4dvzv) in the mailing list to have better description of what rules to add and how they are expected to be used.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+          mq.getExpressionLineage(
+              relNode,
+              RexInputRef.of(columnBit, fieldList));
+      // If mq.getExpressionLineage cannot get result, skip it
+      if (exprLineage == null) {
+        continue;
+      }
+      Set<RexTableInputRef.RelTableRef> relTableRefs =
+          RexUtil.gatherTableReferences(Lists.newArrayList(exprLineage));
+      // If the column come from multiple tables, skip it
```
asolimando, line 129:
```suggestion
      // If the column comes from multiple tables, skip it
```

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+   *
+   * @return Expanded predicates
+   */
+  private RexNode apply(
+      RexNode condition,
+      RelNode relNode,
+      List<RelDataTypeField> fieldList,
+      RelBuilder relBuilder,
+      RelMetadataQuery mq) {
+    HashMap<Integer, RexTableInputRef.RelTableRef> inputRefToTableRef = new HashMap<>();
```
asolimando, line 112: Prefer interface to concrete classes unless there is a good reason not to, you might consider using the `final` modifier to stress the fact that the variable is not going to be overwritten later on (the method is long enough that such hint helps), in this and other cases where relevant.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+    ImmutableBitSet columnBits = RelOptUtil.InputFinder.bits(condition);
```
asolimando, line 113: Nit: you should group together (i.e., no newline) lines that make sense to be considered together (like a variable used in a loop coming right after it, or a variable used in the following statement), and split groups of lines that are conceptually separated, like "extract input ref and their mappings to tables" and "computing the new filter conditions".

I know it's not a deal breaker, but I find it quite hard to read as-is, it would be nice if you could do a pass over the code to improve readability.

Here, for instance, I'd put a newline between 112 and 113 (which can go with the `if` that follows as they are strictly related).

For formatting, you can generally take a look around at similar code to take inspiration.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+      case AND:
+        List<RexNode> andOperands = RexUtil.flattenAnd(((RexCall) condition).getOperands());
+        for (RexNode andOperand : andOperands) {
+          Map<RexTableInputRef.RelTableRef, RexNode> operandResult = expandDeep(andOperand);
+          mergeAnd(additionalConditions, operandResult);
+        }
+        break;
+      case OR:
+        List<RexNode> orOperands = RexUtil.flattenOr(((RexCall) condition).getOperands());
+        additionalConditions.putAll(expandDeep(orOperands.get(0)));
```
asolimando, line 258: Nit: can't you handle the case for element 0 in the for loop as well?

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+        }
+        break;
+      default:
+        break;
+      }
+
+      return additionalConditions;
+    }
+
+    private RexTableInputRef.@Nullable RelTableRef inputRefsBelongOneTable(
```
asolimando, line 272: Nit: inputRefsBelongToOneTable?

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+        ImmutableBitSet inputRefs) {
+      RexTableInputRef.RelTableRef tableRef = inputRefToTableRef.get(inputRefs.nth(0));
+      if (tableRef == null) {
+        return null;
+      }
+      for (int inputBit : inputRefs.asList()) {
+        RexTableInputRef.RelTableRef inputBitTableRef = inputRefToTableRef.get(inputBit);
+        if (inputBitTableRef == null || !inputBitTableRef.equals(tableRef)) {
```
asolimando, line 280:
```suggestion
        if (!tableRef.equals(inputBitTableRef)) {
```
you know from line 275 that `tableRef` is not null, `equals` will return `false` in case `inputBitTableRef` is `null` or different, so you can handle both cases in a single shot

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+          return null;
+        }
+      }
+      return tableRef;
+    }
+
+    /**
+     * For additional predicates of each operand in conjuction, all of them should be retained and
```
asolimando, line 288: conjuNction (I saw Mihai spotted a similar mispelling for disjunction), can you do a massive grep on this?

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+     * use 'AND' to combine expressions.
+     *
```
asolimando, line 290: Can you provide a simple example in the javadoc explaining what the method does? similarly for `mergeOr`.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+            relBuilder.or(
+                entry.getValue(),
+                forMergeMap.get(entry.getKey()));
+        baseMap.put(entry.getKey(), mergedRex);
+      }
+    }
+
+    private void incrementAndCheck() {
+      if (maxProcessLimitNodes > 0 && ++currentCount > maxProcessLimitNodes) {
+        throw OverLimitException.INSTANCE;
```
asolimando, line 338: throwing creates a stacktrace, which is a very expensive operation, we should use a boolean guard (possibly in an array if mutating it in other places is needed) to indicate that we should "exit", you can probably take inspiration from how [RexShuttle does it](https://github.com/apache/calcite/blob/b394abca407eea06ee8f8b50f30bd443f286bb42/core/src/main/java/org/apache/calcite/rex/RexShuttle.java#L42)

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+
+    private final Map<Integer, RexTableInputRef.RelTableRef> inputRefToTableRef;
+
+    private final RelBuilder relBuilder;
+
+    private final int maxProcessLimitNodes;
+
+    private int currentCount;
+
+    private ExpandDisjuctionHelper(
```
asolimando, line 207: This reminds me somewhat a `RexVisitor`, can you take a look at some implementations in the project to see if it can fit your needs?

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+  }
+
+  /** Rule configuration. */
+  @Value.Immutable(singleton = false)
+  public interface Config extends RelRule.Config {
+    Config FILTER = ImmutableExpandDisjuctionRule.Config.builder()
+        .withMatchHandler(ExpandDisjuctionRule::matchFilter)
+        .build()
+        .withOperandSupplier(b ->
+            b.operand(Filter.class).anyInputs());
```
asolimando, line 357: Nit: If you fit into 80 chars you can always place it in a single line

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
```diff
@@ -0,0 +1,379 @@
...
+        .build()
+        .withOperandSupplier(b ->
+            b.operand(Join.class).anyInputs());
+
+    @Override default ExpandDisjuctionRule toRule() {
+      return new ExpandDisjuctionRule(this);
+    }
+
+    default int processLimit() {
+      return 10000;
```
asolimando, line 370: This is configurable, right? 10k is just the default?

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
silundong: For the operands of Or, I want to take the common set of single table among them. For example : 
Or
  (t1.p1, t2.p1, t3.p1)
  (t2.p2, t3.p2, t4.p2)
  (t2.p3, t3.p3, t5.p3)

t2, t3 are tables that exist for each operand, so I can extract the predicates involved.
Handle the 0th element first, that is to get the initial state of the common set. Subsequent loop will only reduce the number of elements in this common set.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
silundong: Here I refer to the usage of [RexUtil#CnfHelper](https://github.com/apache/calcite/blob/b394abca407eea06ee8f8b50f30bd443f286bb42/core/src/main/java/org/apache/calcite/rex/RexUtil.java#L2644). OverLimitException extends ControlFlowException, which overrides the fillInStackTrace method and no longer fills the stack trace.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
silundong: Yes. This value currently means the upper limit of the recursive call to `expandDeep`. It will be changed to the upper limit of the number of new expressions as mentioned by Mihai.

silundong:
> I apologize because my comment was indeed ambiguous, what I meant was not to remove it from `CoreRules`, but to make sure the rule doesn't end up in one the sets of standard rules as defined in `RelOptRules`, this is not the case at the moment.
> 
> I think that a word of caution could be added as a comment in the javadoc of the rule itself, briefly describing known issues (what rules could be problematic when paired with this one, how to limit the danger via the bloat parameter, etc.), this will help users understanding the implications of adding this rule, especially in combination with some others known to cause infinite loops.
> 
> In the current javadoc, it is correctly suggested to use this rule before `CoreRules#FILTER_INTO_JOIN` and `CoreRules#JOIN_CONDITION_PUSH`, maybe we can be more explicit suggesting the use of the `HepPlanner` too?
> 
> It can help for discussions like [this one](https://lists.apache.org/thread/0yotdv62r6b3wg1qztftdpnkzpc4dvzv) in the mailing list to have better description of what rules to add and how they are expected to be used.

No worries, thank you for your suggestions and comments. I responded to some of your comments that I think may be fine, and I will update the others and rule that Mihaibudiu hopes for in the next commit.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
asolimando: I missed that, thank for pointing that out, it's fine then

silundong:
@mihaibudiu I have updated the current rule and am ready to push. Would you like me to create a new PR to develop the rule you want to add?

mihaibudiu:
I think it's simpler to make a new rule based on a new JIRA ticket and a new PR.
If it will reuse code from this, maybe we should get this merged first.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjuctionRule.java
silundong: I tried to extends `RexVisitorImpl`. The logic of `expandDeep` can be transplanted to `visitCall`, but other methods (visitInputRef, visitLiteral...) may also need to be overridden, such as expressions like` id > 30 or booleanInputRef`. However, `expandDeep` does not need to consider these, which feels more convenient.

silundong:
> I think it's simpler to make a new rule based on a new JIRA ticket and a new PR. If it will reuse code from this, maybe we should get this merged first.

Ok, I have created a new [jira ticket](https://issues.apache.org/jira/browse/CALCITE-6946), and the new rule will be developed based on it.
I have updated based on the previous comments. Could you please review it again?

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
```diff
@@ -0,0 +1,430 @@
...
+    private Map<RexTableInputRef.RelTableRef, RexNode> expandDeep(RexNode condition) {
+      Map<RexTableInputRef.RelTableRef, RexNode> additionalConditions = new HashMap<>();
+
+      ImmutableBitSet inputRefs = RelOptUtil.InputFinder.bits(condition);
+      if (inputRefs.isEmpty()) {
+        return additionalConditions;
+      }
+
+      RexTableInputRef.RelTableRef tableRef = inputRefsBelongToOneTable(inputRefs);
+      // The condition already belongs to one table, return it directly
```
mihaibudiu, line 254: this comment should be after the if

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
```diff
@@ -0,0 +1,430 @@
...
+        RexTableInputRef.RelTableRef inputBitTableRef = inputRefToTableRef.get(inputBit);
+        if (!tableRef.equals(inputBitTableRef)) {
+          return null;
+        }
+      }
+      return tableRef;
+    }
+
+    /**
+     * For additional predicates of each operand in conjunction, all of them should be retained and
```
mihaibudiu, line 308: combine predicates that depend on the same table using conjunctions
specify that the result is returned by modifying baseMap
(you could also rename the function to combinePredicatesUsingAnd)

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
```diff
@@ -0,0 +1,430 @@
...
+            + (baseMap.containsKey(entry.getKey())
+                ? baseMap.get(entry.getKey()).nodeCount()
+                : 0);
+        checkExpandCount(mergedRex.nodeCount() - oriCount);
+        baseMap.put(entry.getKey(), mergedRex);
+      }
+    }
+
+    /**
+     * Only if all operands in disjunction have additional predicates for table t1, we use 'OR'
```
mihaibudiu, line 338: see my comments for javadoc above

silundong:
> I think this is pretty good, I mostly made comments about the documentation.

Think you for your suggestion, I will update it.

mihaibudiu:
can you also please fix "disjunction" in the jira title, PR title and commit message (the last one when you squash the commits)?

mihaibudiu:
When this is ready for review please request a new one

silundong:
> When this is ready for review please request a new one

Thanks for you review. I have updated it.

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
```diff
@@ -850,4 +850,16 @@ private CoreRules() {}
   @Experimental
   public static final DphypJoinReorderRule HYPER_GRAPH_OPTIMIZE =
       DphypJoinReorderRule.Config.DEFAULT.toRule();
+
+  /** Rule that expand disjuction in condition of a {@link Filter}.
```
mihaibudiu, line 854: this "disjunction" is still misspelled
it should also be "expands disjunction in the condition of a filter"
same problems in the comment below

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
```diff
@@ -0,0 +1,429 @@
...
+      RelBuilder relBuilder,
+      RelMetadataQuery mq) {
+    final Map<Integer, RexTableInputRef.RelTableRef> inputRefToTableRef = new HashMap<>();
+
+    ImmutableBitSet columnBits = RelOptUtil.InputFinder.bits(condition);
+    if (columnBits.isEmpty()) {
+      return condition;
+    }
+
+    // Trace which table does the column referenced in the condition come from
```
mihaibudiu, line 128: I think this is easier to read if written as 
"for each table discover the conditions that refer to columns of that table (and refer to no other tables)"

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
```diff
@@ -0,0 +1,429 @@
...
+          mq.getExpressionLineage(
+              relNode,
+              RexInputRef.of(columnBit, fieldList));
+      // If mq.getExpressionLineage cannot get result, skip it
+      if (exprLineage == null) {
+        continue;
+      }
+      Set<RexTableInputRef.RelTableRef> relTableRefs =
+          RexUtil.gatherTableReferences(Lists.newArrayList(exprLineage));
+      // If the column comes from multiple tables, skip it
```
mihaibudiu, line 140: What does it mean for a column to "come" from multiple tables?
You mean, it is computed by an expression that depends on multiple tables?

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
```diff
@@ -0,0 +1,429 @@
...
+        RexTableInputRef.RelTableRef inputBitTableRef = inputRefToTableRef.get(inputBit);
+        if (!tableRef.equals(inputBitTableRef)) {
+          return null;
+        }
+      }
+      return tableRef;
+    }
+
+    /**
+     * Combine predicates that depend on the same table using conjunctions, and
```
mihaibudiu, line 307: Combine predicates that depend on the same table using conjunctions.
The result is returned by modifying baseMap.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
```diff
@@ -0,0 +1,429 @@
...
+            + (baseMap.containsKey(entry.getKey())
+                ? baseMap.get(entry.getKey()).nodeCount()
+                : 0);
+        checkExpandCount(mergedRex.nodeCount() - oriCount);
+        baseMap.put(entry.getKey(), mergedRex);
+      }
+    }
+
+    /**
+     * Combine predicates that depend on the same table using disjunctions, and
```
mihaibudiu, line 337: same change as above.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
```diff
@@ -0,0 +1,429 @@
...
+        .withMatchHandler(ExpandDisjunctionForTableRule::matchJoin)
+        .build()
+        .withOperandSupplier(b -> b.operand(Join.class).anyInputs());
+
+    @Override default ExpandDisjunctionForTableRule toRule() {
+      return new ExpandDisjunctionForTableRule(this);
+    }
+
+    /**
+     * Limit how much complexity can increase during expanding.
```
mihaibudiu, line 416: Limit to the size growth of the condition allowed during expansion.

mihaibudiu (approved):
My comments are only about comments, but otherwise this looks good to merge.
I apologize if my previous suggestions were not clear enough.

mihaibudiu:
@asolimando were your comments addressed?
I'd like to merge this (after my last batch of comments are addressed as well), it looks very useful.

silundong:
> My comments are only about comments, but otherwise this looks good to merge. I apologize if my previous suggestions were not clear enough.

Sorry for the spelling and grammar errors, I have updated them.

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
```diff
@@ -0,0 +1,430 @@
...
+              relNode,
+              RexInputRef.of(columnBit, fieldList));
+      // If mq.getExpressionLineage cannot get result, skip it
+      if (exprLineage == null) {
+        continue;
+      }
+      Set<RexTableInputRef.RelTableRef> relTableRefs =
+          RexUtil.gatherTableReferences(Lists.newArrayList(exprLineage));
+      // If the column is computed by an expression that depends on multiple tables, skip it
+      if (relTableRefs.isEmpty() || relTableRefs.size() > 1) {
```
xuzifu666, line 142: Here maybe just use refTableRefs != null is ok?

xuzifu666, line 142: Here maybe just use  **(refTableRefs != null)**  is ok?

mihaibudiu: if relTableRefs could be null this would have failed in CI due to error check

silundong: > Here maybe just use **(refTableRefs != null)** is ok?

`RexUtil.gatherTableReferences` will not return null.

xuzifu666: Yes, my mistake, (relTableRefs.size != 1) equivalent to (relTableRefs.isEmpty() || relTableRefs.size() > 1)? or keep it as current~

silundong: Yes, `(relTableRefs.size != 1)` is more concise, I will update it.

asolimando (approved):
LGTM, all the important changes I asked were applied already

mihaibudiu:
> > My comments are only about comments, but otherwise this looks good to merge. I apologize if my previous suggestions were not clear enough.
> 
> Sorry for the spelling and grammar errors, I have updated them.

Especially in large PRs, please add new commits until the reviews are finished, because otherwise it's very hard to see what is changed. The commits can be squashed before merging

PR #4268, opened 2025-03-26
[CALCITE-6915] Generalize terminology Linter to allow pattern based checks in commit messages

zabetak:
The main goal is to capture more invalid term usages and easier definition of new rules.

1. Use pattern based definition in term rules and allow multiple valid/golden terms
2. Use List based rule collection for faster iteration; since we iterate over all rules using Map is wasteful
3. Make existing rules case-insensitive to capture more invalid usages of DBMS terms
4. Pre-compile and instantiate patterns and rules only once


caicancai (approved):
Thanks. LGTM

mihaibudiu:
@zabetak do you want to merge this?

PR #4273, opened 2025-03-27
[CALCITE-6919] Invalid unparse for CHAR without precision in PostgreSQL

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6919

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/PostgresqlSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/PostgresqlSqlDialect.java
@@ -75,6 +75,13 @@ public class PostgresqlSqlDialect extends SqlDialect {
             return super.getMaxPrecision(typeName);
           }
         }
+
+        @Override public int getDefaultPrecision(SqlTypeName typeName) {
+          if (typeName == SqlTypeName.CHAR) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/PostgresqlSqlDialect.java
```diff
@@ -75,6 +75,13 @@ public class PostgresqlSqlDialect extends SqlDialect {
             return super.getMaxPrecision(typeName);
           }
         }
+
+        @Override public int getDefaultPrecision(SqlTypeName typeName) {
```
mihaibudiu, line 79: was the previous behavior incorrect?
CHAR(1) is a legal type.

xuzifu666: Related to the issue: https://issues.apache.org/jira/browse/CALCITE-6423
select cast('a' as char);
select cast('a' as char(1));
both of them return a;
and had a discuss in JIRA: https://issues.apache.org/jira/browse/CALCITE-6919 @mihaibudiu

xuzifu666:
Close it first before discussion finished in JIRA @mihaibudiu

PR #4274, opened 2025-03-27
[CALCITE-6920] The type derived for a cast to INT ARRAY always has non-nullable elements

mihaibudiu:


caicancai (commented):
Is there a similar situation for MAP type?

caicancai (approved):
Is there a similar situation for MAP type?

mihaibudiu:
The pr fixed both. I mentioned this in the jira

PR #4276, opened 2025-03-29
[CALCITE-6926] ClickHouse Float type should not be converted to Double type

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6926


xiedeyantu (approved):

caicancai (approved):

xuzifu666:
Thanks for your review! @caicancai , could you help set the tag to 'disscus in jira' firstly?

PR #4277, opened 2025-03-29
[CALCITE-6927] Add rule for join condition remove IS NOT DISTINCT FROM

xiedeyantu:
please see: [CALCITE-6927](https://issues.apache.org/jira/browse/CALCITE-6927)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/JoinConditionExpandIsNotDistinctFromRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/JoinConditionExpandIsNotDistinctFromRule.java
@@ -1,0 +1,141 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Join;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexShuttle;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.tools.RelBuilder;
+
+import org.immutables.value.Value;
+
+/**
+ * Planner rule that replaces {@code IS NOT DISTINCT FROM}
+ * in a {@link Join} condition with logically equivalent operations.
+ *
+ * @see SqlStdOperatorTable#IS_NOT_DISTINCT_FROM
+ * @see RelBuilder#isNotDistinctFrom
+ */
+@Value.Enclosing
+public final class JoinConditionExpandIsNotDistinctFromRule
+    extends RelRule<JoinConditionExpandIsNotDistinctFromRule.Config>
+    implements TransformationRule {
+
+  /** Creates a JoinConditionExpandIsNotDistinctFromRule. */
+  JoinConditionExpandIsNotDistinctFromRule(Config config) {
+    super(config);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/calcite/rel/rules/JoinConditionExpandIsNotDistinctFromRule.java
```diff
@@ -0,0 +1,132 @@
...
+            join.getCluster().getRexBuilder());
+
+    RelNode newJoin = join.accept(rewriteShuttle);
+    call.transformTo(newJoin);
+  }
+
+  //~ Inner Classes ----------------------------------------------------------
+
+  /** Shuttle that removes 'x IS NOT DISTINCT FROM y' and converts it
+   * to '(COALESCE(x, '') = COALESCE(y, ''))
```
mihaibudiu, line 75: '' is not an accurate description of what's happening.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionExpandIsNotDistinctFromRule.java
```diff
@@ -0,0 +1,132 @@
...
+
+      if (call.getOperator()
+          == SqlStdOperatorTable.IS_NOT_DISTINCT_FROM) {
+        RexCall tmpCall = (RexCall) newCall;
+
+        RexNode operand0 = tmpCall.operands.get(0);
+        RexNode operand1 = tmpCall.operands.get(1);
+
+        RexNode operand0Zero =
+            rexBuilder.makeZeroLiteral(operand0.getType());
```
mihaibudiu, line 96: this only works if the type is a numeric type.
joins work on arbitrary types, including arrays and maps

mihaibudiu:
This does not look like a traditional optimization rule, since it does not really change the plan in any way, it just replaces some expressions with equivalent ones. I don't know if there is a better place for it, though.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionExpandIsNotDistinctFromRule.java
xiedeyantu: I missed this question. I will think about how to make it compatible with complex types.

xiedeyantu: Done

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionExpandIsNotDistinctFromRule.java
xiedeyantu: Done

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionExpandIsNotDistinctFromRule.java
```diff
@@ -0,0 +1,159 @@
...
+        RexUtil.findOperatorCall(SqlStdOperatorTable.IS_NOT_DISTINCT_FROM,
+        oldJoinCond);
+
+    if (found == null) {
+      // no longer contains isNotDistinctFromOperator
+      return;
+    }
+
+    for (RexNode op : found.getOperands()) {
+      if (op.getType().getSqlTypeName() == SqlTypeName.MAP) {
```
mihaibudiu, line 71: I actually think maps can be compared, but it's fine to skip this case too.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionExpandIsNotDistinctFromRule.java
```diff
@@ -0,0 +1,159 @@
...
+      }
+      return newCall;
+    }
+  }
+
+  private static RexNode makeZeroLiteral(RexBuilder rexBuilder, RexNode operand) {
+    if (operand.getType().getSqlTypeName() == SqlTypeName.ARRAY) {
+      ArraySqlType arrayType = (ArraySqlType) operand.getType();
+      RelDataType elementType = arrayType.getComponentType();
+      RexNode elementZero = rexBuilder.makeZeroLiteral(elementType);
```
mihaibudiu, line 141: You need to call the function recursively, because ARRAY can have ARRAY as elements.
Also, this may work for MULTISET too.

mihaibudiu: there is also the case of ROW types.
Maybe you want to enhance the makeZeroLiteral function to handle these as well?

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionExpandIsNotDistinctFromRule.java
xiedeyantu: I referred to the implementation of spark, it does not support map comparison. But I agree with your, map should be comparable. I will skip the map comparison for now, and modify it later if I have more references. WDYT

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionExpandIsNotDistinctFromRule.java
xiedeyantu: Great Idea! I think I can create a new ticket to complete this suggestion first.

xiedeyantu: @mihaibudiu I have used the new method `makeZeroRexNode` to replace the `makeZeroLiteral`. Could please help me to have a look again?

mihaibudiu (approved):
If there are no other comments I plan to merge this soon.

PR #4278, opened 2025-03-29
[CALCITE-6928] Sonar analysis fails

xiedeyantu:


xiedeyantu:
I haven't found a good solution yet, temporarily close this PR.

PR #4279, opened 2025-03-31
[CALCITE-6929] PrestoSqlDialect unparseCall should not reuse from PostgreSqlDialect

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6929

Currently all exsisted test cases can cover the changed code.

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/PrestoSqlDialect.java
@@ -17,6 +17,7 @@
 package org.apache.calcite.sql.dialect;
 
 import org.apache.calcite.avatica.util.Casing;
+import org.apache.calcite.avatica.util.TimeUnitRange;
 import org.apache.calcite.config.NullCollation;
 import org.apache.calcite.rel.type.RelDataType;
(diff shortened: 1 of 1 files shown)


mihaibudiu:
This PR is missing a test case

xuzifu666:
> This PR is missing a test case

thanks, according the changed code, RelToSqlConverterTest##testFloor could cover the  condition of PR, WDYT? @mihaibudiu

mihaibudiu:
If that test works before and after your change, it means that it is not covering your changes.

xuzifu666:
> If that test works before and after your change, it means that it is not covering your changes.

Yes，this PR is just a refactoring. 
before the PR:
PrestoSqlDialect##unparseCall  default part extend from PostgreSqlDialect##unparseCall , I worry about that if future changed code in PostgreSqlDialect##unparseCall would impact PrestoSqlDialect##unparseCall  directly, If contains some feature change different from Presto and developer not change also in Presto would bring risk，as I know Presto syntax contains much differences to PostgreSQL.
after the PR:
PrestoSqlDialect is independent from any other Dialect. Maybe this way is suitable. 
Do you agree with me or contains other suggestion？

PR #4280, opened 2025-03-31
[CALCITE-6930] Implementing JoinConditionOrExpansionRule

xiedeyantu:
Please see: [CALCITE-6930](https://issues.apache.org/jira/browse/CALCITE-6930)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/JoinCommuteRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/JoinCommuteRule.java
@@ -137,6 +137,12 @@ public class JoinCommuteRule
         .build();
   }
 
+  public static RexNode swapJoinCond(RexNode cond, Join join, RexBuilder rexBuilder) {
+    final VariableReplacer variableReplacer =
+        new VariableReplacer(rexBuilder, join.getLeft().getRowType(), join.getRight().getRowType());
+    return variableReplacer.apply(cond);
+  }
+
   @Override public boolean matches(RelOptRuleCall call) {
     Join join = call.rel(0);
     // SEMI and ANTI join cannot be swapped.
--- a/core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
@@ -1,0 +1,385 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelOptUtil;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Join;
+import org.apache.calcite.rel.core.JoinRelType;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexInputRef;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.tools.RelBuilder;
+
+import org.immutables.value.Value;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Planner rule that matches a
+ * {@link org.apache.calcite.rel.core.Join}
+ * and expands OR clauses in join conditions.
+ *
+ * <p>This rule expands OR conditions in join clauses into
+ * multiple separate join conditions, allowing the optimizer
+ * to handle these conditions more efficiently.
+ *
+ * <p>The following is an example of inner join,
+ * and other examples for other kinds of joins
+ * can be found in the code below.
+ *
+ * <p>Project[*]
+ *    └── Join[OR(t1.id=t2.id, t1.age=t2.age), inner]
+ *        ├── TableScan[t1]
+ *        └── TableScan[t2]
+ *
+ * <p>into
+ *
+ * <p>Project[*]
+ *    └── UnionAll
+ *        ├── Join[t1.id=t2.id, inner]
+ *        │   ├── TableScan[t1]
+ *        │   └── TableScan[t2]
+ *        └── Join[t1.age=t2.age AND t1.id≠t2.id, inner]
+ *            ├─── TableScan[t1]
+ *            └─── TableScan[t2]
+ *
+ */
+@Value.Enclosing
+public class JoinConditionOrExpansionRule
+    extends RelRule<JoinConditionOrExpansionRule.Config>
+    implements TransformationRule {
+
+  /** Creates an JoinConditionExpansionOrRule. */
+  protected JoinConditionOrExpansionRule(Config config) {
+    super(config);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public boolean matches(RelOptRuleCall call) {
+    Join join = call.rel(0);
+    List<RexNode> orConds = RelOptUtil.disjunctions(join.getCondition());
+    return orConds.size() > 1;
+  }
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    Join join = call.rel(0);
+    RelBuilder relBuilder = call.builder();
+
+    RelNode expanded;
+    switch (join.getJoinType()) {
+    case INNER:
+      expanded = expandInnerJoin(join, relBuilder);
+      break;
+    case ANTI:
+      expanded = expandAntiJoin(join, relBuilder);
+      break;
+    case LEFT:
+      expanded = expandLeftOrRightJoin(join, true, relBuilder);
+      break;
+    case RIGHT:
+      expanded = expandLeftOrRightJoin(join, false, relBuilder);
+      break;
+    case FULL:
+      expanded = expandFullJoin(join, relBuilder);
+      break;
+    default:
+      return;
+    }
+    call.transformTo(expanded);
+  }
+
+  private List<RexNode> splitCond(Join join) {
+    final RexBuilder builder = join.getCluster().getRexBuilder();
+    final List<RexNode> orConds = RelOptUtil.disjunctions(join.getCondition());
+    final int leftFieldCount = join.getLeft().getRowType().getFieldCount();
+
+    final List<RexNode> result = new ArrayList<>();
+    final List<RexNode> otherBuffer = new ArrayList<>();
+
+    for (RexNode cond : orConds) {
+      if (isValidCond(cond, leftFieldCount)) {
+        if (!otherBuffer.isEmpty()) {
+          result.add(RexUtil.composeDisjunction(builder, otherBuffer));
+          otherBuffer.clear();
+        }
+        result.add(cond);
+      } else {
+        otherBuffer.add(cond);
+      }
+    }
+
+    if (!otherBuffer.isEmpty()) {
+      result.add(RexUtil.composeDisjunction(builder, otherBuffer));
+    }
+
+    return result;
+  }
+
+  private boolean isValidCond(RexNode node, int leftFieldCount) {
+    if (!(node instanceof RexCall)) {
+      return false;
+    }
+
+    RexCall call = (RexCall) node;
+    SqlKind kind = call.getKind();
+    switch (kind) {
+    case EQUALS:
+    case IS_NOT_DISTINCT_FROM:
+      RexNode left = call.getOperands().get(0);
+      RexNode right = call.getOperands().get(1);
+
+      if (left instanceof RexInputRef && right instanceof RexInputRef) {
+        RexInputRef leftRef = (RexInputRef) left;
+        RexInputRef rightRef = (RexInputRef) right;
+        return (leftRef.getIndex() < leftFieldCount && rightRef.getIndex() >= leftFieldCount)
+            || (leftRef.getIndex() >= leftFieldCount && rightRef.getIndex() < leftFieldCount);
+      }
+      return false;
+    case AND:
+      return call.getOperands().stream()
+          .allMatch(op -> isValidCond(op, leftFieldCount));
+    default:
+      return false;
+    }
+  }
+
+  /**
+   * This method will make the following conversions.
+   *
+   * <p>Project[*]
+   *    └── Join[OR(t1.id=t2.id, t1.age=t2.age), left]
+   *        ├── TableScan[t1]
+   *        └── TableScan[t2]
+   *
+   * <p>into
+   *
+   * <p>Project[*]
+   *    └── UnionAll
+   *        ├── Join[t1.id=t2.id, inner]
+   *        │   ├── TableScan[t1]
+   *        │   └── TableScan[t2]
+   *        ├── Join[t1.age=t2.age AND t1.id≠t2.id, inner]
+   *        │   ├── TableScan[t1]
+   *        │   └── TableScan[t2]
+   *        └── Project[t1-side cols + NULLs]
+   *            └── Join[t1.id=t2.id, anti]
+   *                ├── Join[t1.age=t2.age, anti]
+   *                │   ├── TableScan[t1]
+   *                │   └── TableScan[t2]
+   *                └── TableScan[t2]
+   */
+  private RelNode expandLeftOrRightJoin(Join join, boolean isLeftJoin,
+      RelBuilder relBuilder) {
+    List<RexNode> orConds = splitCond(join);
+    List<RelNode> joins = expandLeftOrRightJoinToRelNodes(join, orConds, isLeftJoin, relBuilder);
+    return relBuilder.pushAll(joins)
+        .union(true, joins.size())
+        .build();
+  }
+
+  private List<RelNode> expandLeftOrRightJoinToRelNodes(Join join, List<RexNode> orConds,
+      boolean isLeftJoin, RelBuilder relBuilder) {
+    List<RelNode> joins = new ArrayList<>();
+    joins.addAll(expandInnerJoinToRelNodes(join, orConds, relBuilder));
+    joins.add(expandAntiJoinToRelNode(join, orConds, isLeftJoin, true, relBuilder));
+    return joins;
+  }
+
+  /**
+   * This method will make the following conversions.
+   *
+   * <p>Project[*]
+   *    └── Join[OR(t1.id=t2.id, t1.age=t2.age), full]
+   *        ├── TableScan[t1]
+   *        └── TableScan[t2]
+   *
+   * <p>into
+   *
+   * <p>Project[*]
+   *    └── UnionAll
+   *        ├── Join[t1.id=t2.id, inner]
+   *        │   ├── TableScan[t1]
+   *        │   └── TableScan[t2]
+   *        ├── Join[t1.age=t2.age AND t1.id≠t2.id, inner]
+   *        │   ├── TableScan[t1]
+   *        │   └── TableScan[t2]
+   *        ├── Project[t1-side cols + NULLs]
+   *        │   └── Join[t1.id=t2.id, anti]
+   *        │       ├── Join[t1.age=t2.age, anti]
+   *        │       │   ├── TableScan[t1]
+   *        │       │   └── TableScan[t2]
+   *        │       └── TableScan[t2]
+   *        └── Project[NULLs + t2-side cols]
+   *            └── Join[t2.id=t1.id, anti]
+   *                ├── Join[t2.age=t1.age, anti]
+   *                │   ├── TableScan[t2]
+   *                │   └── TableScan[t1]
+   *                └── TableScan[t1]
+   */
+  private RelNode expandFullJoin(Join join, RelBuilder relBuilder) {
+    List<RexNode> orConds = splitCond(join);
+    List<RelNode> joins = new ArrayList<>();
+    joins.addAll(expandInnerJoinToRelNodes(join, orConds, relBuilder));
+    joins.add(expandAntiJoinToRelNode(join, orConds, false, true, relBuilder));
+    joins.add(expandAntiJoinToRelNode(join, orConds, true, true, relBuilder));
+
+    relBuilder.pushAll(joins)
+        .union(true, joins.size());
+
+    final List<RexNode> projects = join.getRowType().getFieldList().stream()
+        .map(field -> {
+          RexNode rexNode = relBuilder.field(field.getIndex());
+          return field.getType().equals(rexNode.getType())
+              ? rexNode
+              : relBuilder.getRexBuilder().makeCast(field.getType(), rexNode, true, false);
+        }).collect(Collectors.toList());
+
+    return relBuilder.project(projects)
+        .build();
+  }
+
+  /**
+   * This method will make the following conversions.
+   *
+   * <p>Project[*]
+   *    └── Join[OR(t1.id=t2.id, t1.age=t2.age), inner]
+   *        ├── TableScan[t1]
+   *        └── TableScan[t2]
+   *
+   * <p>into
+   *
+   * <p>Project[*]
+   *    └── UnionAll
+   *        ├── Join[t1.id=t2.id, inner]
+   *        │   ├── TableScan[t1]
+   *        │   └── TableScan[t2]
+   *        └── Join[t1.age=t2.age AND t1.id≠t2.id, inner]
+   *            ├─── TableScan[t1]
+   *            └─── TableScan[t2]
+   */
+  private RelNode expandInnerJoin(Join join, RelBuilder relBuilder) {
+    List<RexNode> orConds = splitCond(join);
+    List<RelNode> joins = expandInnerJoinToRelNodes(join, orConds, relBuilder);
+    return relBuilder.pushAll(joins)
+        .union(true, joins.size())
+        .build();
+  }
+
+  private List<RelNode> expandInnerJoinToRelNodes(Join join, List<RexNode> orConds,
+      RelBuilder relBuilder) {
+    List<RelNode> joins = new ArrayList<>();
+    for (int i = 0; i < orConds.size(); i++) {
+      RexNode orCond = orConds.get(i);
+      for (int j = 0; j < i; j++) {
+        orCond = relBuilder.and(orCond, relBuilder.not(orConds.get(j)));
+      }
+
+      relBuilder.push(join.getLeft())
+          .push(join.getRight())
+          .join(JoinRelType.INNER, orCond);
+
+      joins.add(relBuilder.build());
+    }
+    return joins;
+  }
+
+  /**
+   * This method will make the following conversions.
+   *
+   * <p>Project[*]
+   *    └── Join[OR(id=id0, age=age0), anti]
+   *        ├── TableScan[tbl]
+   *        └── TableScan[tbl]
+   *
+   * <p>into
+   *
+   * <p>HashJoin[id=id0, anti]
+   *    ├── HashJoin[age=age0, anti]
+   *    │   ├── TableScan[tbl]
+   *    │   └── TableScan[tbl]
+   *    └── TableScan[tbl]
+   */
+  private RelNode expandAntiJoin(Join join, RelBuilder relBuilder) {
+    List<RexNode> orConds = splitCond(join);
+    return expandAntiJoinToRelNode(join, orConds, true, false, relBuilder);
+  }
+
+  private RelNode expandAntiJoinToRelNode(Join join, List<RexNode> orConds,
+      boolean isLeftAnti, boolean isAppendNulls, RelBuilder relBuilder) {
+    RelNode left = isLeftAnti ? join.getLeft() : join.getRight();
+    RelNode right = isLeftAnti ? join.getRight() : join.getLeft();
+
+    RelNode top = left;
+    for (int i = 0; i < orConds.size(); i++) {
+      RexNode orCond = orConds.get(i);
+      relBuilder.push(top)
+          .push(right)
+          .join(JoinRelType.ANTI,
+              isLeftAnti
+                  ? orCond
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,164 @@
...
+    RelBuilder relBuilder = call.builder();
+    List<RexNode> orConds = RelOptUtil.disjunctions(join.getCondition());
+
+    if (orConds.size() <= 1) {
+      return;
+    }
+
+    int leftFieldCount = join.getLeft().getRowType().getFieldCount();
+    for (RexNode cond : orConds) {
+      if (!isSimpleCondition(cond, leftFieldCount)) {
```
mihaibudiu, line 89: why is this necessary? The transformation is legal anyway.
Ideally this notion of "simple" should be parameterized, so one could plug in a different implementation.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,164 @@
...
+      }
+      condsAppend.add(orCond);
+      relBuilder.push(join.getLeft())
+          .push(join.getRight())
+          .join(join.getJoinType(), orCond);
+    }
+
+    relBuilder.union(true, orConds.size());
+
+    final List<RexNode> projects = join.getRowType().getFieldList().stream()
```
mihaibudiu, line 108: I didn't expect this to be necessary.
Why would types be different?

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: My initial thought is that all the expanded joins can become HashJoin. If one of them becomes NestedLoopJoin, this conversion has no advantage and CBO may not choose it. I read the jira you replied to me and it seems that there is no restriction here?

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: I'll add a detailed error report here later.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
mihaibudiu: you are thinking about this optimization only, but combined with other optimizations there could be additional benefits.
For example, this could help in combination with this new rule that is being added: https://issues.apache.org/jira/browse/CALCITE-6914

xiedeyantu: Yes, my friend silun is working on it.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: Strangely, I didn't get any error messages, and it worked fine without this logic. So I just removed it.

xiedeyantu:
> When you address comments you should not force-push, but rather add new commits; this makes it much easier to review the changes.

Thank you for pointing out the problem. I will follow your suggestion later.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,115 @@
...
+    Config DEFAULT = ImmutableJoinConditionOrExpansionRule.Config.of()
+        .withOperandFor(Join.class);
+
+    @Override default JoinConditionOrExpansionRule toRule() {
+      return new JoinConditionOrExpansionRule(this);
+    }
+
+    /** Defines an operand tree for the given classes. */
+    default Config withOperandFor(Class<? extends Join> joinClass) {
+      return withOperandSupplier(b -> b.operand(joinClass).anyInputs())
```
mihaibudiu, line 111: I left two design questions in Jira. I should have probably asked them sooner, but I only got time to think about them now.

xiedeyantu: I tried to implement outer join and included semi/anti join, I will update jira later.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,297 @@
...
+   * <p>Project[*]
+   *    └── Join[OR(id=id0, age=age0), left]
+   *        ├── TableScan[tbl]
+   *        └── TableScan[tbl]
+   *
+   * <p>into
+   *
+   * <p>Project[*]
+   *    └── UnionAll
+   *        ├── HashJoin[id=id0, inner]
```
mihaibudiu, line 104: technically I don't think these are hash joins, they are all logical joins, aren't they?

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,297 @@
...
+          .push(join.getRight())
+          .join(joinType, orCond);
+      left = relBuilder.build();
+    }
+
+    relBuilder.push(left);
+    int fieldCount = join.getRowType().getFieldCount();
+    List<RexNode> projects = new ArrayList<>(relBuilder.fields());
+    while (fieldCount > relBuilder.fields().size()) {
+      projects.add(relBuilder.literal(null));
```
mihaibudiu, line 274: does the builder infer types for these?

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,297 @@
...
+   * <p>Project[*]
+   *    └── UnionAll
+   *        ├── HashJoin[id=id0, inner]
+   *        │   ├── TableScan[tbl]
+   *        │   └── TableScan[tbl]
+   *        ├── HashJoin[age=age0 AND id≠id0, inner]
+   *        │   ├── TableScan[tbl]
+   *        │   └── TableScan[tbl]
+   *        └── Project[tbl-side cols + NULLs]
+   *            └── HashJoin[id=id0, anti]
```
mihaibudiu, line 111: this expansion looks wrong to me; the antijoin is not symmetric.
I think it is a chain of antijoins with the tablescane on the left, isn't it?

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -6357,6 +6357,144 @@ LogicalProject(EMPNO=[$3], ENAME=[$4], JOB=[$5], MGR=[$6], HIREDATE=[$7], SAL=[$
 LogicalJoin(condition=[true], joinType=[inner])
   LogicalTableScan(table=[[scott, EMP]])
   LogicalTableScan(table=[[scott, DEPT]])
+]]>
```
mihaibudiu, line 6360: these tests are very hard for humans to certify.
it would be very nice if we could execute these plans somehow on some input data to compare results with the non-expanded form.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: Yes, you're right!

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
xiedeyantu: This is a very good suggestion. I also thought about it when writing the test. Is there any existing test framework that can support this kind of logic? I don’t seem to have found it. Do I need to make an extension for JdbcTest?

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: If I understand correctly, I may change the comments to tbl1 and tbl2 to make it clearer. Otherwise, I may not understand your meaning.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
mihaibudiu: Maybe CalciteASsert.consistentResult is the function needed.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: Builder can not do it. It is a SEMI/ANTI JOIN, there are fewer columns than INNER JOIN, and the columns will not match during UNION. NULL is used here to fill in the gaps, and the JOIN will infer types using derive method later.

mihaibudiu: but can you help the builder by supplying nulls with the correct type?

xiedeyantu: I think I understand what you mean. Using literal directly may cause potential problems in some special cases.

mihaibudiu: If you already know the type, it's probably best to use it rather than infer it again

xiedeyantu: Yes, maybe I can know the types.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
xiedeyantu: Thanks, I'll see how to adjust the test.

xiedeyantu: I found a way to verify it, do you think it works?

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: Done

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: Done

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: Done

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,297 @@
...
+   *        │   └── TableScan[tbl]
+   *        ├── HashJoin[age=age0 AND id≠id0, inner]
+   *        │   ├── TableScan[tbl]
+   *        │   └── TableScan[tbl]
+   *        ├── Project[tbl-side cols + NULLs]
+   *        │   └── HashJoin[id=id0, anti]
+   *        │       ├── HashJoin[age=age0, anti]
+   *        │       │   ├── TableScan[tbl]
+   *        │       │   └── TableScan[tbl]
+   *        │       └── TableScan[EMP]
```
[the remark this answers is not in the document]
xiedeyantu: Done

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,297 @@
...
+    joins.add(expandSemiOrAntiJoin(join, JoinRelType.SEMI, orConds, relBuilder));
+    return joins;
+  }
+
+  /**
+   * This method will make the following conversions.
+   *
+   * <p>Project[*]
+   *    └── Join[OR(id=id0, age=age0), full]
+   *        ├── TableScan[tbl]
```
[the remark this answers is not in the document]
xiedeyantu: Done

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,299 @@
...
+  protected JoinConditionOrExpansionRule(Config config) {
+    super(config);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    Join join = call.rel(0);
+    RelBuilder relBuilder = call.builder();
+    List<RexNode> orConds = RelOptUtil.disjunctions(join.getCondition());
```
mihaibudiu, line 58: This expansion is most useful when some of the orConds are conjunctions of table columns.
In fact, I think you don't even need to expand ALL the orConds, you could stop after one of them.
Is it worth analyzing the orConds and making a list of the ones that can lead to conjunctions that can be efficiently handled by equijoins and expanding only these?

mihaibudiu (commented):
This looks pretty good, and it is surprisingly short given how much it accomplishes.
I still have an algorithmic question, about the nature of the expressions in the OR.

xiedeyantu:
> This looks pretty good, and it is surprisingly short given how much it accomplishes. I still have an algorithmic question, about the nature of the expressions in the OR.

Sorry, I may not understand what you mean. What question about OR do I need to reply to?

mihaibudiu:
> > This looks pretty good, and it is surprisingly short given how much it accomplishes. I still have an algorithmic question, about the nature of the expressions in the OR.
> 
> Sorry, I may not understand what you mean. What question about OR do I need to reply to?

This expansion is most useful when some of the orConds are conjunctions of table columns.
In fact, I think you don't even need to expand ALL the orConds, you could stop after one of them.
Is it worth analyzing the orConds and making a list of the ones that can lead to conjunctions that can be efficiently handled by equijoins and expanding only these?

xiedeyantu:
> > > This looks pretty good, and it is surprisingly short given how much it accomplishes. I still have an algorithmic question, about the nature of the expressions in the OR.
> > 
> > 
> > Sorry, I may not understand what you mean. What question about OR do I need to reply to?
> 
> This expansion is most useful when some of the orConds are conjunctions of table columns. In fact, I think you don't even need to expand ALL the orConds, you could stop after one of them. Is it worth analyzing the orConds and making a list of the ones that can lead to conjunctions that can be efficiently handled by equijoins and expanding only these?

I agree with you. In fact, I considered this in my initial view. You can check if `isSimpleCondition` is what you mean (maybe the name needs to be changed). The link is [here ](https://github.com/apache/calcite/commit/749c0725643cf3eb7953b236c0595966198d606d#diff-5761c993bb47ab9756426564eb187d027412f8e6469a0c0df8ce0935d2222fd8R7-R121)

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: @mihaibudiu I can't think of a good way to split out part of an `OR` and leave the rest unsplit. I can only think of expanding all `OR` or none of them. for example, whether all sub-expressions of an `OR` expansion can be converted into equijoins. Could you give me some advice?

mihaibudiu: I think you can analyze each of the orConds separately, and make a list of the ones that are "useful". Then you rewrite the OR as (useful0 OR useful1 OR ... OR (everythingelse)), where "everythingelse" is an OR which contains the expressions which are not useful. Then you expand only the useful conditions, and treat "everythingelse" as a simple atomic boolean expression, without regards to its content.

Regarding when an expression is "useful": An expression is useful if it is:
- a conjunction that has at least one useful subexpression
- it is a comparison using EQUALS or IS NOT DISTINCT FROM between two columns in the two inputs of the join

There is one more subtlety to worry about: if OR has lazy evaluation, you cannot actually reorder the expressions in the OR, because you may produce runtime exceptions. For example TRUE OR a/0 = b cannot be reordered.

mihaibudiu: Notice that SQL does not specify whether OR is lazy, but it is lazy in many dialects. I don't know if other parts of Calcite pay attention to this fact.

mihaibudiu:
Indeed, isSimpleCondition is pretty close to what you need.

mihaibudiu:
Consider this condition left.x = right.y OR (left.x < 5 AND left.y > 6)
In this case "everything else" is "left.x < 5 AND left.y > 6".
Regarding IS NOT DISTINCT FROM, that's the concern of other optimizations. It clearly defines an equijoin.

xiedeyantu:
> Consider this condition left.x = right.y OR (left.x < 5 AND left.y > 6) In this case "everything else" is "left.x < 5 AND left.y > 6". Regarding IS NOT DISTINCT FROM, that's the concern of other optimizations. It clearly defines an equijoin.

In this condition left.x = right.y OR (left.x < 5 AND left.y > 6). In my opinion, it may not be expanded at all. I have not thought about how to expand part of it and not expand the other part.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,299 @@
...
+
+  private RelNode expandRightJoin(Join join, List<RexNode> orConds,
+      RelBuilder relBuilder) {
+    List<RelNode> joins = expandRightJoinToRelNodes(join, orConds, relBuilder);
+    return relBuilder.pushAll(joins)
+        .union(true, joins.size())
+        .build();
+  }
+
+  private List<RelNode> expandRightJoinToRelNodes(Join join, List<RexNode> orConds,
```
suibianwanwank, line 141: Why is this a semi join here? If I understand correctly, for a right join this should swap the join inputs and use an anti join to ensure any unmatched condition return the right table column.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,299 @@
...
+   */
+  private RelNode expandLeftJoin(Join join, List<RexNode> orConds,
+      RelBuilder relBuilder) {
+    List<RelNode> joins = expandLeftJoinToRelNodes(join, orConds, relBuilder);
+    return relBuilder.pushAll(joins)
+        .union(true, joins.size())
+        .build();
+  }
+
+  private List<RelNode> expandLeftJoinToRelNodes(Join join, List<RexNode> orConds,
```
suibianwanwank, line 125: The logic for handling Left Join and Right Join is very similar—perhaps we can merge them.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: Thanks! There is indeed a problem here. My idea is that `RIGHT JOIN` does not support direct `OR` expansion for the time being, and it can be converted to `LEFT JOIN` through `JOIN_COMMUTE_OUTER` and then `OR` expansion, which can avoid repeated `JOIN` commute in `JOIN_CONDITION_OR_EXPANSION_RULE` and some complex adaptation. WDYT?

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: As above, we can temporarily not support direct `OR` expansion of `RIGHT JOIN`.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
asolimando: I haven't looked at all details, but the principle of keeping individual rules as simple as possible and exploit the combined effect of multiple rules is in general a good design.

That's, for instance, the reason why we keep simplification to the minimum in all rules, assuming that rules from `ReduceExpressionsRule` will take care of the rest.

It might be worth writing a test in `RelOptRuleTest` to cover such an example?

xiedeyantu: Hey, great minds think alike! In the next submission, I will add a case of how to handle RIGHT JOIN through JOIN_COMMUTE_OUTER and JOIN_CONDITION_OR_EXPANSION_RULE. Thanks!

xiedeyantu:
@mihaibudiu According to your suggestion, I split orConds. According to @suibianwanwank  and @asolimando 's suggestions, I adjusted the implementation of right join (in order to implement full join, right anti join expansion needs to be implemented, and right join is supported by the way). And added a test in RelOptRuleTest. Please take a look again. Thanks!

core/src/main/java/org/apache/calcite/rel/rules/JoinCommuteRule.java
```diff
@@ -137,6 +137,12 @@ public JoinCommuteRule(Class<? extends Join> clazz,
         .build();
   }
 
+  public static RexNode swapJoinCond(RexNode cond, Join join, RexBuilder rexBuilder) {
```
mihaibudiu, line 140: Does this have to be public or can it be moved to the file where it's used?

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -6613,26 +6613,26 @@ EnumerableProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL
...
       <![CDATA[
-EnumerableProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], EMPNO0=[$9], ENAME0=[$10], JOB0=[$11], MGR0=[$12], HIREDATE0=[$13], SAL0=[$14], COMM0=[$15], DEPTNO0=[$16], SLACKER0=[$17])
-  EnumerableUnion(all=[true])
-    EnumerableHashJoin(condition=[=($0, $9)], joinType=[inner])
-      EnumerableTableScan(table=[[CATALOG, SALES, EMP]])
-      EnumerableTableScan(table=[[CATALOG, SALES, EMP]])
-    EnumerableHashJoin(condition=[AND(=($5, $14), <>($0, $9))], joinType=[inner])
-      EnumerableTableScan(table=[[CATALOG, SALES, EMP]])
-      EnumerableTableScan(table=[[CATALOG, SALES, EMP]])
+LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], EMPNO0=[$9], ENAME0=[$10], JOB0=[$11], MGR0=[$12], HIREDATE0=[$13], SAL0=[$14], COMM0=[$15], DEPTNO0=[$16], SLACKER0=[$17])
```
mihaibudiu, line 6628: this is better

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -55,40 +62,92 @@ protected JoinConditionOrExpansionRule(Config config) {
...
+
+  private List<RexNode> splitCond(Join join) {
+    RexBuilder builder = join.getCluster().getRexBuilder();
+    List<RexNode> conds = new ArrayList<>();
+    List<RexNode> otherConds = new ArrayList<>();
+    List<RexNode> orConds = RelOptUtil.disjunctions(join.getCondition());
+    int leftFieldCount = join.getLeft().getRowType().getFieldCount();
+
+    for (RexNode cond : orConds) {
+      if (!isUsefulCond(cond, leftFieldCount)) {
```
mihaibudiu, line 106: So you are not preserving the order...
This may cause problems with operations that cause runtime exceptions. 
Note that almost anything can cause runtime exceptions, e.g., arithmetic overflow.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -256,26 +313,50 @@ private List<RelNode> expandInnerJoinToRelNodes(Join join, List<RexNode> orConds
...
-      projects.add(
+    if (!isAppendNulls) {
+      return top;
+    }
+
+    relBuilder.push(top);
+    int fieldCount = 0;
+    List<RexNode> fields = new ArrayList<>(relBuilder.fields());
+    List<RexNode> nulls = new ArrayList<>();
+    while (fieldCount < right.getRowType().getFieldCount()) {
```
mihaibudiu, line 349: maybe a for loop is better

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
mihaibudiu: If you want to be conservative you have to:
- either detect whether an expression can cause a runtime exception and only move the ones that never do
- preserve the order and stop at the first expression which is not useful and put everything from that point on in a single big OR which is not expanded

Can you add a test for such a case so we can see what happens?
For example an expression such as `A <= 0 OR LN(A) <= 0`

core/src/main/java/org/apache/calcite/rel/rules/JoinCommuteRule.java
xiedeyantu: It relies on VariableReplacer, which is a private class, so I wrap a method here.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: I kept the order of the join conditions and also fixed a few details you mentioned. Thanks so much!

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,384 @@
...
+    final RexBuilder builder = join.getCluster().getRexBuilder();
+    final List<RexNode> orConds = RelOptUtil.disjunctions(join.getCondition());
+    final int leftFieldCount = join.getLeft().getRowType().getFieldCount();
+
+    final List<RexNode> result = new ArrayList<>();
+    final List<RexNode> otherBuffer = new ArrayList<>();
+
+    for (RexNode cond : orConds) {
+      if (isUsefulCond(cond, leftFieldCount)) {
+        if (!otherBuffer.isEmpty()) {
```
mihaibudiu, line 108: This small fragment could be a utility function "creatDisjunction".
I wonder whether this exists already somewhere.

xiedeyantu: Done!

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,384 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Planner rule that matches a
+ * {@link org.apache.calcite.rel.core.Join}
+ * and expands OR clauses in join conditions.
+ *
+ * <p>this rule would expand the OR condition into
```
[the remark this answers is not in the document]
xiedeyantu: Done!

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,384 @@
...
+ * two separate join conditions, allowing the optimizer
```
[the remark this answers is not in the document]
xiedeyantu: Done!

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
```diff
@@ -0,0 +1,401 @@
...
+    if (!(node instanceof RexCall)) {
+      return false;
+    }
+
+    RexCall call = (RexCall) node;
+    SqlKind kind = call.getKind();
+    switch (kind) {
+    case EQUALS:
+    case IS_NOT_DISTINCT_FROM:
+      RexNode left = call.getOperands().get(0);
```
suibianwanwank, line 156: Perhaps this method name could be clearer, like `containEquivCond`.  
BTW, since Calcite is a general-purpose database with a strict type system. In many cases, the inputRef in join conditions will contain casts, which may cause some optimizations to fail. This issue exists across several rules. I'll try to collect such examples and create a new ticket to address.

suibianwanwank (approved):
Overall LGTM, I left some comments.

core/src/main/java/org/apache/calcite/rel/rules/JoinCommuteRule.java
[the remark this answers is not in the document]
xiedeyantu: This involves the use of a private class, which I don't want to change. This rule also has some static methods for external use.

core/src/main/java/org/apache/calcite/rel/rules/JoinConditionOrExpansionRule.java
xiedeyantu: For example, it also includes is not distinct from, and the cast you mentioned, maybe it is not very appropriate to use whether it contains equal conditions, or is there any better name you can recommend?

suibianwanwank:
> I successfully executed `./gradlew build` locally. I don't know why continuous-integration/jenkins/pr-head failed, but it should have nothing to do with pr.

From the error message reported, it appears that RelOptRuleTest is in error.

xiedeyantu:
> > I successfully executed `./gradlew build` locally. I don't know why continuous-integration/jenkins/pr-head failed, but it should have nothing to do with pr.
> 
> From the error message reported, it appears that RelOptRuleTest is in error.

Yes, I know. But it work locally.

xiedeyantu:
I used `./gradlew build` and `build` in idea to compile locally for more than 10 times and they were all successful, all after clean and rebuild. It should be ruled out that it is a code problem.

mihaibudiu:
can you try to create another PR with the exact same changes, so we can see whether the CI succeeds there?

xiedeyantu:
> can you try to create another PR with the exact same changes, so we can see whether the CI succeeds there?

@mihaibudiu Very strange! I created another PR and it passed, and the sonar of this PR also passed with the new PR. New pr is #4300.

mihaibudiu:
I will close this PR in favor of #4300

PR #4288, opened 2025-04-04
[CALCITE-6938] Support zero value creation of nested data types

xiedeyantu:
Please see: [CALCITE-6938](https://issues.apache.org/jira/browse/CALCITE-6938)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexBuilder.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexBuilder.java
@@ -1924,6 +1924,27 @@ public class RexBuilder {
     return makeLiteral(zeroValue(type), type);
   }
 
+  public RexNode makeZeroRexNode(RelDataType type) {
+    switch (type.getSqlTypeName()) {
+    case ARRAY:
+      return makeCast(type,
+          makeCall(type, SqlStdOperatorTable.ARRAY_VALUE_CONSTRUCTOR, ImmutableList.of()));
+    case MULTISET:
+      return makeCast(type,
+          makeCall(type, SqlStdOperatorTable.MULTISET_VALUE, ImmutableList.of()));
+    case MAP:
+      return makeCast(type,
+          makeCall(type, SqlStdOperatorTable.MAP_VALUE_CONSTRUCTOR, ImmutableList.of()));
+    case ROW:
+      List<RexNode> zeroFields = type.getFieldList().stream()
+              .map(field -> makeZeroRexNode(field.getType()))
+              .collect(Collectors.toList());
+      return makeCall(type, SqlStdOperatorTable.ROW, zeroFields);
+    default:
+      return makeZeroLiteral(type);
+    }
+  }
+
   private static Comparable zeroValue(RelDataType type) {
     switch (type.getSqlTypeName()) {
     case CHAR:
--- a/core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
+++ b/core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
@@ -31,7 +31,10 @@ import org.apache.calcite.sql.SqlCollation;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.fun.SqlLibraryOperators;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.type.ArraySqlType;
 import org.apache.calcite.sql.type.BasicSqlType;
+import org.apache.calcite.sql.type.MapSqlType;
+import org.apache.calcite.sql.type.MultisetSqlType;
 import org.apache.calcite.sql.type.SqlTypeFactoryImpl;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.test.CustomTypeSystems;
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1917,6 +1917,34 @@ public RexLiteral makeZeroLiteral(RelDataType type) {
     return makeLiteral(zeroValue(type), type);
   }
 
+  public RexNode makeZeroForNestedType(RelDataType type) {
+    switch (type.getSqlTypeName()) {
+    case ARRAY:
```
mihaibudiu, line 1922: so you think a zero array is an array containing a zero?
or should it be an empty array?
the question is "zero for what algebraic operation"?
for concatenation the empty array is the zero.

mihaibudiu: same question for multisets and maps

mihaibudiu: I think an empty object makes more sense

xiedeyantu: Yes, I also think what you said is very reasonable. Maybe I should implement an XxxZero for each nested type and implement the Comparable interface. I am trying to modify it. Could you give some suggestions?

mihaibudiu: Why do you need the Comparable interface?
Just modify this function to return empty collections for ARRAY, MAP, MULTISET.

mihaibudiu: Yes, but I think it will just work, no need to worry about it.
Just invoke the constructor for array with no arguments and cast it to the correct type.

xiedeyantu: Got it, thanks a lot!

xiedeyantu: Please take a look, I have addressed, thanks!

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1918,31 +1918,24 @@ public RexLiteral makeZeroLiteral(RelDataType type) {
...
 
   public RexNode makeZeroForNestedType(RelDataType type) {
+    SqlOperator op;
     switch (type.getSqlTypeName()) {
     case ARRAY:
-      ArraySqlType arrayType = (ArraySqlType) type;
-      RexNode zeroElement = makeZeroForNestedType(arrayType.getComponentType());
-      return makeCall(type, SqlStdOperatorTable.ARRAY_VALUE_CONSTRUCTOR,
-          ImmutableList.of(zeroElement));
+      op = SqlStdOperatorTable.ARRAY_VALUE_CONSTRUCTOR;
```
mihaibudiu, line 1924: What is the type of the constructed array?
I wonder whether this will create problems because the type needs to be inferred.
But you know the type already, so you can synthesize a `CAST(ARRAY[] AS type)`.
Same below.

xiedeyantu: When using CAST, you may need to write something like ARRAY[INTEGER]

mihaibudiu: I am not sure what this statements means. This is not legal SQL.
And here I am talking about the internal representation.

xiedeyantu: You mean to keep the actual type and the writing consistent to avoid problems with `CAST(ARRAY[] AS type)` in SQL, right? I'll think about it and may need to adjust it again.

mihaibudiu: This should be a simple change. Just wrap the constructor call into a cast that produces `type`.

xiedeyantu: Thanks! I will fix it soon.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1917,6 +1917,27 @@ public RexLiteral makeZeroLiteral(RelDataType type) {
...
+      break;
+    case MULTISET:
+      op = SqlStdOperatorTable.MULTISET_VALUE;
+      break;
+    case MAP:
+      op = SqlStdOperatorTable.MAP_VALUE_CONSTRUCTOR;
+      break;
+    case ROW:
+      op = SqlStdOperatorTable.ROW;
```
mihaibudiu, line 1933: I don't think this works for ROW. For ROW your original solution was correct: you need to create a zero for each field using a recursive call.

core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
```diff
@@ -1104,6 +1107,140 @@ private static Stream<Arguments> testData4testMakeZeroLiteral() {
...
+
+    // MAP<VARCHAR, INTEGER>
+    RelDataType mapType = new MapSqlType(varcharType, integerType, false);
+    RexNode expectedMap =
+        b.makeCast(
+            mapType, b.makeCall(mapType, SqlStdOperatorTable.MAP_VALUE_CONSTRUCTOR,
+            ImmutableList.of()));
+
+    // ROW<INTEGER, VARCHAR>
+    RelDataType rowType =
```
mihaibudiu, line 1183: I don't think this would pass validation.
Which means that these tests are not really conclusive, since they can generate incorrect programs. Is there a way you could make sure that these programs are valid?

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
xiedeyantu: I tried to sort out the logic according to your suggestion. ROW needs to be processed recursively and filled with real zero values. Should MAP have the same logic as ROW or ARRAY? If both use recursive logic, then ARRAY and MULITSET only need to do a CAST at the innermost level?

mihaibudiu: Yes, ROW needs real zeros, MAP, ARRAY, and MULTISET should be empty.
The empty value is the zero value for monoid provided by the concatenation operation for these data structures.

core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
xiedeyantu: Is it that the test for ROW needs to be modified, but not the others? I saw that makeZoreLiteral also verifies how the zero value is constructed, and I referred to its logic.

mihaibudiu: it would be nice if there was a way to validate all of them.
I didn't think exactly how.

xiedeyantu: Should I keep this validation logic? Should I restore the zero value implementation logic of ROW to the previous version?

mihaibudiu: I think zero was correct for ROW previously.
For validation you can keep this, but can you make sure that there is additional validation? I haven't looked to see how builder results are validated.

xiedeyantu: Sorry, I didn't express myself clearly. I changed the zero value logic of ROW, and its corresponding validation will also be changed. I will keep the rest as it is.

mihaibudiu: the fact that there was no error shows that the validation is not strong enough. I don't think you should be able to create a ROW value with no fields when the type has fields.

mihaibudiu: if you find a way to validate the row, it will work for all other values.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
xiedeyantu: Done!

mihaibudiu:
Please squash the commits so we can merge this.

xiedeyantu:
> Please squash the commits so we can merge this.

@mihaibudiu Thanks, I have rebased.

PR #4289, opened 2025-04-04
[CALCITE-6939] Add support for Lateral Column Alias

mihaibudiu:
This is based on an implementation from https://github.com/apache/calcite/pull/2931 by @jaystarshot from 2022. 
However, this implementation does only one of the many things from the original. Moreover, it uses a conformance feature flag to gate the effects of this change.


mihaibudiu:
To save you the trouble from reading the JIRA issue, this adds a conformance flag to allow such programs become legal:
`SELECT 1 as x, x+1 as y`, where `x` can be used immediately once it has been defined.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -6677,6 +6677,32 @@ public boolean isBangEqualAllowed() {
         .withConformance(lenient).fails("Column 'E' not found in any table");
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6939">
+   * [CALCITE-6939] Add support for Lateral Column Alias</a>. */
+  @Test void testAliasInSelect() {
```
caicancai, line 6682: Does it support multi-layer nesting?

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -7393,20 +7393,65 @@ private SqlNode nthSelectItem(int ordinal, final SqlParserPos pos) {
...
+          if (!(ex instanceof CalciteContextException)) {
+            throw ex;
+          }
+          if (validator.getConformance().isSelectAlias() && id.isSimple()) {
+            // Try to lookup the item in the select list itself
+            String name = id.getSimple();
+            @Nullable SqlNode expr = null;
+            final SqlNameMatcher nameMatcher =
+                validator.catalogReader.nameMatcher();
+            int n = 0;
```
caicancai, line 7427: I feel that the variable name `n` is a bit vague. Is there a more appropriate name?

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -6677,6 +6677,32 @@ public boolean isBangEqualAllowed() {
...
+    final SqlConformance withSelectAlias = new SqlAbstractConformance() {
+      @Override public boolean isSelectAlias() {
+        return true;
+      }
+    };
+
+    // Standard conformance
+    sql("select 1 AS x, ^x^ as Y")
```
NobiGo, line 6690: 1: Does this currently support SQL execution? Or is it only supporting validate?
2: If the table includes x and y columns, what would the output be if it is renamed as x?
Some cases I test in Spark:
```
create table table(x int,y int);
insert into table values(1, 2);
select 2 as x, x + 2  from table; // output=> 2       3
 select 2 as x1, x1 + 2; // Error in query: cannot resolve '`x1`' given input columns: []; line 1 pos **17;**
```

core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
```diff
@@ -102,6 +102,16 @@ public interface SqlConformance {
    */
   boolean isGroupByAlias();
 
+  /**
+   * Whether to allow aliases from the {@code SELECT} clause to be used as
+   * column names in the same {@code SELECT} clause.
+   * E.g., SELECT 1 as X, X+1 as Y;
+   * Name lookup considers an identifier in the same SELECT only
+   * if other lookups failed.
+   * Supported by Spark, Snowflake, BigQuery.
```
NobiGo, line 111: Spark currently does not support.

mihaibudiu: https://www.databricks.com/blog/introducing-support-lateral-column-alias

mihaibudiu: May be wrong about bigquery, though. See the issue on JIRA.

caicancai: I remember that the documentation of Dataricks and Spark is sometimes different. I am not sure if this is also the case.

mihaibudiu: What other SQL do they support besides Spark SQL?

caicancai: The SQL engine inside Databricks has been rewritten in C++, which is different from the open source Spark. Maybe there are some changes in the rewriting process, which are difficult for us to verify. Maybe we should focus on the open source Spark version.

caicancai: I see. I have Dataricks' Spark here. I can help you verify it tomorrow.

caicancai: I see. @NobiGo Which version of Spark did you verify this on? Can you tell me? I can verify this on DataRisks later.

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
mihaibudiu: I have added some tests to show that it does now

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
mihaibudiu: I have added some SqlOperatorTests to show that this works end-to-end.

NobiGo: I tested it again using 3.5.1 and it is supported. Previously, I used 3.1.2.

core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
NobiGo: I retested with 3.5.1 and it is supported.

core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
```diff
@@ -102,6 +102,28 @@ public interface SqlConformance {
...
   boolean isGroupByAlias();
 
+  /**
+   * Value describing how to perform lookup for values defined in a SELECT statement.
+   */
+  enum SelectAliasLookup {
+    /** Values defined in a SELECT statement are not visible within the statement. */
+    UNSUPPORTED,
+    /** Values defined in a SELECT statement are visible to the right of their definition. */
+    LEFT_TO_RIGHT,
```
NobiGo, line 112:
```
What if the table t already has a column called “a” ?
Should “b” be 42 + 1, or should it be the column “a” + 1 ?
```
In Spark:
```
select 2 as x, x + 2 from table;
```
When the table has column x. and value 1. The SQL will return `2   3`.
I'm not sure if any other databases return `2 4` means the priority of column aliases is higher than the definition of the table.

mihaibudiu: As I have commented in the JIRA, if you would do that, you would break all existing queries that redefine columns that exist already. So I don't think any database can afford to do that.

mihaibudiu: If 'x' exists in the FROM, it is used from the FROM. 
Only when lookup in the FROM fails it is made in the local SELECT.

NobiGo (approved):
LGTM Based on the test cases.

mihaibudiu:
If you have comments on this topic please weigh in either here or in the jira.
Otherwise I plan to merge this.

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -7389,15 +7396,50 @@ private SqlNode nthSelectItem(int ordinal, final SqlParserPos pos) {
...
+    private SelectExpander(SqlValidatorImpl validator, SelectScope scope,
+        SqlSelect select, Map<String, SqlNode> expansions, Set<String> lookingUp) {
       super(validator, scope);
       this.select = select;
+      this.expansions = expansions;
+      this.lookingUp = lookingUp;
+    }
+
+    /**
+     * Create an expander for the items in a SELECT list.
```
julianhyde, line 7421: s/Create/Creates/

core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
```diff
@@ -102,6 +102,28 @@ public interface SqlConformance {
    */
   boolean isGroupByAlias();
 
+  /**
+   * Value describing how to perform lookup for values defined in a SELECT statement.
```
julianhyde, line 106: 'lookup for values' - these are not values, these are aliases. It's not a SELECT statement, it's the SELECT list.

core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
```diff
@@ -102,6 +102,28 @@ public interface SqlConformance {
...
+   */
+  enum SelectAliasLookup {
+    /** Values defined in a SELECT statement are not visible within the statement. */
+    UNSUPPORTED,
+    /** Values defined in a SELECT statement are visible to the right of their definition. */
```
julianhyde, line 111: Mention 'lateral'

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -7389,15 +7396,50 @@ private SqlNode nthSelectItem(int ordinal, final SqlParserPos pos) {
...
+   * local select expressions into the expressions themselves.
    */
   static class SelectExpander extends Expander {
     final SqlSelect select;
-
-    SelectExpander(SqlValidatorImpl validator, SelectScope scope,
-        SqlSelect select) {
+    // Maps simple identifiers to their expansions.
+    final Map<String, SqlNode> expansions;
+    // List of identifiers that are currently beeing looked-up.  Used to detect
```
julianhyde, line 7407: s/beeing/being/

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -7406,7 +7448,78 @@ static class SelectExpander extends Expander {
       if (node != id) {
         return node;
       } else {
-        return super.visit(id);
+        try {
+          return super.visit(id);
+        } catch (Exception ex) {
+          if (!(ex instanceof CalciteContextException)) {
```
julianhyde, line 7454: This is a lot of code for a `catch (Exception)` block.

julianhyde (commented):
We need a lot more tests. We need tests at the SQL (Quidem) level to make sure that expansion doesn't produce invalid plans. 

Something like the SQL queries in https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-name-resolution 

I am especially worried about matching of GROUP BY columns. I can "SELECT x + y" if I have "GROUP BY x + y" but what if "y" has a different meaning in the two expressions?

mihaibudiu:
I will check to see how quidem tests can change conformance

mihaibudiu:
> I am especially worried about matching of GROUP BY columns. I can "SELECT x + y" if I have "GROUP BY x + y" but what if "y" has a different meaning in the two expressions?

Some of this is already a consideration when using isGroupbyAlias.

mihaibudiu:
I hope I have made it work with GROUP BY and HAVING. Indeed, there were some issues.

mihaibudiu:
There is a second commit which attempted to address all the issues raised by @julianhyde

core/src/main/java/org/apache/calcite/sql/validate/SqlConformance.java
caicancai: dataricks test success

caicancai (approved):

PR #4291, opened 2025-04-07
[CALCITE-6921] REGEXP_REPLACE with empty string causes Exception

kramerul:


NobiGo (approved):

NobiGo:
@kramerul Hi, could you please improve the test cases like Calcite-6923? To have a clearer view of the impact of our current PR on SQL.

kramerul:
Hi @NobiGo,

for [Calcite-6923](https://issues.apache.org/jira/browse/CALCITE-6923), it was necessary to test that backslashes are handled correctly. Therefore, Julian suggested to add additional tests (including *.iq files). In this case it's not necessary to have additional test for escape characters.

PR #4295, opened 2025-04-09
[CALCITE-6944] Cannot parse parenthesized partition by in Table Function

juntaozhang:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6944


xiedeyantu:
Could you change the title of the PR to make it consistent with your commit content?

juntaozhang:
as https://issues.apache.org/jira/browse/CALCITE-6944 discussion, `<table argument>` with parentheses is not the standard in [SO/IEC 19075-7:2021](https://www.iso.org/standard/78938.html), so close this PR.

PR #4296, opened 2025-04-10
[CALCITE-6945] Use LITERAL_AGG to simplify SubQueryRemoveRule by avoi…

NobiGo:
…ding the extra Project


core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
```diff
@@ -729,21 +728,10 @@ public static Exists createExistsPlan(
...
-    final int projectedKeyCount = exprs.size();
-    exprs.add(rexBuilder.makeLiteral(true));
-
     ret = relBuilder.push(ret)
-        .project(exprs)
         .aggregate(
-            relBuilder.groupKey(ImmutableBitSet.range(projectedKeyCount)),
-            relBuilder.min(relBuilder.field(projectedKeyCount)))
+            relBuilder.groupKey(relBuilder.fields()),
+            relBuilder.literalAgg(true))
```
mihaibudiu, line 734: what does LITERAL_AGG return for an empty set?

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -15595,15 +15585,15 @@ where exists(select * from dept where emp.deptno = dept.deptno)]]>
 LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8])
   LogicalJoin(condition=[=($7, $9)], joinType=[inner])
     LogicalTableScan(table=[[CATALOG, SALES, EMP]])
-    LogicalProject(DEPTNO=[$0])
+    LogicalProject(DEPTNO3=[$0])
```
mihaibudiu, line 15588: this suggests that an ALIAS is missing somewhere

core/src/main/java/org/apache/calcite/plan/RelOptUtil.java
NobiGo: `select literal_agg(true) from emp where 1=0;`
I make this work in Calcite, and it returns 0 rows.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
NobiGo: I have rolled back the relevant changes, it seems to be caused by the logic above.

mihaibudiu (approved):
Do we know if any end-to-end test exercises this rule?
Maybe some JDBCTest?

NobiGo:
> Do we know if any end-to-end test exercises this rule? Maybe some JDBCTest?

We have sub-query.iq to cover this.

mihaibudiu:
Do IQ tests use this rule?

NobiGo:
> Do IQ tests use this rule?

Yes, this SubQueryRemoveRule is effective in [Programs#standard](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/tools/Programs.java#L312).

PR #4297, opened 2025-04-10
[CALCITE-6947] Support LogicalSnapshot in RelShuttle

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6947

Currently RelShuttle do not support LogicalTableModify, need to support LogicalTableModify in RelShuttle.


xiedeyantu:
A small suggestion, if this makes sense, are there some missing logical operators that need to be added?

PR #4298, opened 2025-04-10
[CALCITE-6948] Implement MinusToAntiJoinRule

xiedeyantu:
Please see: [CALCITE-6948](https://issues.apache.org/jira/browse/CALCITE-6948)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/MinusToAntiJoinRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/MinusToAntiJoinRule.java
@@ -1,0 +1,108 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.JoinRelType;
+import org.apache.calcite.rel.core.Minus;
+import org.apache.calcite.rel.logical.LogicalMinus;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.tools.RelBuilder;
+
+import org.immutables.value.Value;
+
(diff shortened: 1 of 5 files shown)


mihaibudiu (approved):

xiedeyantu:
@mihaibudiu Hi, this pr has a minor change, please have a look thanks!

core/src/main/java/org/apache/calcite/rel/rules/MinusToAntiJoinRule.java
```diff
@@ -74,10 +74,17 @@ protected MinusToAntiJoinRule(Config config) {
       RelDataType leftFieldType = left.getRowType().getFieldList().get(i).getType();
       RelDataType rightFieldType = right.getRowType().getFieldList().get(i).getType();
 
-      conditions.add(
-          relBuilder.isNotDistinctFrom(
-              rexBuilder.makeInputRef(leftFieldType, i),
-              rexBuilder.makeInputRef(rightFieldType, i + fieldCount)));
+      if (!leftFieldType.isNullable() && !rightFieldType.isNullable()) {
```
mihaibudiu, line 77: I wonder whether this is enough.
You can join on ROW types, and in this case I wonder whether the check should be done on each field.

xiedeyantu: I suddenly thought of a question. We need to judge whether it is `nullable` more precisely, so as to convert `is not distinct` from to `=`?

mihaibudiu: it will work for scalar types as it is; you could check that

xiedeyantu: Will it have any impact if we use is not distinct from uniformly and do not make more complex judgments (return to the logic of the previous version)?

mihaibudiu: the previous is correct, but this one can be slightly more efficient.
However, perhaps this kind of rewrite should be in the scope of a different rule, which could perform it everywhere?

xiedeyantu: So I change it back?

mihaibudiu: Personally I think it's cleaner if each optimization is done by a separate transformation rather than each transformation trying to perform many small optimizations by itself.

xiedeyantu: I agree with you, I will convert the new commit soon.

mihaibudiu:
If not one has comments I plan to merge this soon.
@suibianwanwank you have looked at similar stuff recently, if you want to take a peek.

core/src/test/java/org/apache/calcite/test/JdbcTest.java
```diff
@@ -4015,6 +4015,34 @@ public void checkOrderBy(final boolean desc,
             "commission=null");
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6948">[CALCITE-6948]
+   * Implement IntersectToSemiJoinRule</a>. */
+  @Test void testMinusToAntiJoinRule() {
```
mihaibudiu, line 4020: maybe you will be able to write a few quidem tests too after we merge   https://github.com/apache/calcite/pull/4301

xiedeyantu: Yes, I am submitting #4301 to the draft branch now and would like everyone to see if this approach is acceptable. Then I will adjust the implementation method according to everyone's suggestions. The cases written in JdbcTest can all be completed using quidem tests.

PR #4301, opened 2025-04-12
[CALCITE-6335] Quidem tests should allow specifying optimization passes to apply to programs

xiedeyantu:


core/src/test/resources/sql/test.iq
```diff
@@ -0,0 +1,63 @@
+# test.iq - Join query tests
```
mihaibudiu, line 1: you should probably call this file something else, perhaps planner.iq?
the comment is no longer accurate

core/src/test/resources/sql/test.iq
```diff
@@ -0,0 +1,63 @@
...
++---+
+| I |
++---+
+| 1 |
++---+
+(1 row)
+
+!ok
+
+EnumerableNestedLoopJoin(condition=[IS NOT DISTINCT FROM($0, $1)], joinType=[semi])
```
mihaibudiu, line 36: you should have two tests after setting the rules, so we can check whether the rules set apply to the end of the file

testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
```diff
@@ -71,6 +78,10 @@ public abstract class QuidemTest {
 
   private static final Pattern PATTERN = Pattern.compile("\\.iq$");
 
+  private List<RelOptRule> addRulesCache = new ArrayList<>();
```
mihaibudiu, line 81: even though these are private, I would still document them.

testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
```diff
@@ -171,6 +182,33 @@ protected void checkRun(String path) throws Exception {
...
+                      planner -> {
+                          for (RelOptRule rule : addRulesCache) {
+                            planner.addRule(rule);
+                          }
+                          for (RelOptRule rule : removeRulesCache) {
+                            planner.removeRule(rule);
+                          }
+                      }));
+            }
+            if (propertyName.equals("reset-rules")) {
```
mihaibudiu, line 198: I am not sure this undoes the effect of the previous.
In particular, I don't know what the order of the rules is in the planner, and where in the planner the inserted rules will be
Ideally you can save the entire list of planner rules and restore it. I don't know if the hooks allow that

testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
```diff
@@ -187,6 +225,41 @@ protected void checkRun(String path) throws Exception {
...
+    Pattern pattern = Pattern.compile("([+-])(CoreRules|EnumerableRules)\\.(\\w+)");
+    Matcher matcher = pattern.matcher(value);
+
+    while (matcher.find()) {
+      char operation = matcher.group(1).charAt(0);
+      String ruleSource = matcher.group(2);
+      String ruleName = matcher.group(3);
+
+      try {
+        if (ruleSource.equals("CoreRules")) {
```
mihaibudiu, line 238: Maybe this can be the default - if no source is specified, to make the specification of rules shorter

core/src/test/resources/sql/test.iq
```diff
@@ -0,0 +1,63 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+!use post
+!set outputformat mysql
+!set rules "-CoreRules.INTERSECT_TO_DISTINCT,-EnumerableRules.ENUMERABLE_INTERSECT_RULE,+CoreRules.INTERSECT_TO_SEMI_JOIN"
```
mihaibudiu, line 20: this doesn't really "set" the rules, it updates them

core/src/test/resources/sql/test.iq
```diff
@@ -0,0 +1,63 @@
...
+(1 row)
+
+!ok
+
+EnumerableNestedLoopJoin(condition=[IS NOT DISTINCT FROM($0, $1)], joinType=[semi])
+  EnumerableValues(tuples=[[{ 0 }, { 1 }]])
+  EnumerableCalc(expr#0=[{inputs}], expr#1=[0], expr#2=[>($t0, $t1)], EXPR$0=[$t0], $condition=[$t2])
+    EnumerableValues(tuples=[[{ 0 }, { 1 }]])
+!plan
+!set reset-rules ""
```
mihaibudiu, line 41: this should probably be called "!reset rules"

mihaibudiu (commented):
It's nice that this is so simple.
I hope you can make "reset" work too.

core/src/test/resources/sql/test.iq
xiedeyantu: Good point! I didn’t think that far ahead when writing the demo.

testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
xiedeyantu: Good catch! I’ll have to give this some more thought.

core/src/test/resources/sql/test.iq
xiedeyantu: I took a shortcut here by piggybacking on the existing logic for 'set'. We could probably extend it with custom commands like !update-rules and !reset-rules, though that might be a bit more work. If needed, I can try implementing it!

mihaibudiu:
the most important must-have feature is "reset". otherwise you essentially need to create one file per set of rules you want to test
when you add rules, are they inserted in front of the existing ones? how does the hook work?

xiedeyantu:
> the most important must-have feature is "reset". otherwise you essentially need to create one file per set of rules you want to test when you add rules, are they inserted in front of the existing ones? how does the hook work?

The logic here is consistent with that of Hook.PLANNER in JdbcTest. Hook added rules will be added at the end. Hook.PLANNER is entered after all rules are loaded, and it will contain all default rules.  In this PR, I record the original default rules when I modify the rule group for the first time, so that I can reset it.

core/src/test/resources/sql/planner.iq
```diff
@@ -0,0 +1,84 @@
...
+
+!ok
+
+EnumerableNestedLoopJoin(condition=[IS NOT DISTINCT FROM($0, $1)], joinType=[semi])
+  EnumerableValues(tuples=[[{ 0 }, { 1 }]])
+  EnumerableCalc(expr#0=[{inputs}], expr#1=[0], expr#2=[>($t0, $t1)], EXPR$0=[$t0], $condition=[$t2])
+    EnumerableValues(tuples=[[{ 0 }, { 1 }]])
+!plan
+
+# Test no reset-rules can keep rules
```
mihaibudiu, line 42: test that rules set by "!set update-rules" are sticky

core/src/test/resources/sql/planner.iq
```diff
@@ -0,0 +1,84 @@
...
+
+!ok
+
+EnumerableNestedLoopJoin(condition=[IS NOT DISTINCT FROM($0, $1)], joinType=[semi])
+  EnumerableValues(tuples=[[{ 0 }, { 1 }]])
+  EnumerableCalc(expr#0=[{inputs}], expr#1=[0], expr#2=[>($t0, $t1)], EXPR$0=[$t0], $condition=[$t2])
+    EnumerableValues(tuples=[[{ 0 }, { 1 }]])
+!plan
+
+!set reset-rules ""
```
mihaibudiu, line 62: this is a bit ugly
if you want to reuse "!set", maybe you can say "!set optimization-rules original". 
Here "original" is not the name of any rule

core/src/test/resources/sql/planner.iq
```diff
@@ -0,0 +1,84 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+!use post
+!set outputformat mysql
+!set update-rules "-INTERSECT_TO_DISTINCT,-EnumerableRules.ENUMERABLE_INTERSECT_RULE,+CoreRules.INTERSECT_TO_SEMI_JOIN"
```
mihaibudiu, line 20: let's call this "optimization-rules" instead of "update-rules"

testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
mihaibudiu: can you write a test to verify that the planner rules are the same after a reset?
you don't have to do it from the quidem tests, it can be a separate test that just uses the hook

testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
```diff
@@ -71,6 +77,9 @@ public abstract class QuidemTest {
 
   private static final Pattern PATTERN = Pattern.compile("\\.iq$");
 
+  // cache original rules
```
mihaibudiu, line 80: "Saved original planner rules".

testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
```diff
@@ -187,6 +209,57 @@ protected void checkRun(String path) throws Exception {
     }
   }
 
+  private void updatePlanner(RelOptPlanner planner, String value) {
+    List<RelOptRule> rulesAdd = new ArrayList<>();
+    List<RelOptRule> rulesRemove = new ArrayList<>();
+    parseRules(value, rulesAdd, rulesRemove);
+    rulesAdd.forEach(planner::addRule);
+    rulesRemove.forEach(planner::removeRule);
```
mihaibudiu, line 217: I don't know if the order of rules in the planner matters.
but just in case, perhaps it's safer to first remove, then add

testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
xiedeyantu: Sorry, I have been thinking for a long time and I don't have a good way. In my opinion, the internal logic is to set the original rule set back when it is reset. If it is not set back, then test3 should not have intersect. I can easily verify this problem by logging locally. But I can't think of a good way to write the test.

core/src/test/resources/sql/planner.iq
```diff
@@ -59,9 +59,9 @@ EnumerableNestedLoopJoin(condition=[IS NOT DISTINCT FROM($0, $1)], joinType=[sem
     EnumerableValues(tuples=[[{ 0 }, { 1 }]])
 !plan
 
-!set reset-rules ""
+!set planner-rules original
 
-# Test reset-rules can reset init rules
+# Test that rules set by "!set planner-rules original" are sticky
```
mihaibudiu, line 64: "sticky" means that they can be reused across multiple tests. 
this is not what you are testing here.
you are really testing that this brings back the original rules (to some degree)

mihaibudiu (approved):
This is very good, I think we can merge it once you fix these minor problems.

core/src/test/resources/sql/planner.iq
xiedeyantu: Sorry, I fixed this. When I was developing, I found that multiple SQL statements would be affected by the changed rules if I didn't reset them. So I specifically verified this problem when debugging.

NobiGo (commented):
Updating the process of Rules in Planner makes it seem very complicated. We can't just construct a new Planner directly?

mihaibudiu:
> Updating the process of Rules in Planner makes it seem very complicated. We can't just construct a new Planner directly?

The planner has a lot of rules, I don't expect you want to insert them all by hand. All the enumerable rules are needed to generate running code.

core/src/test/resources/sql/planner.iq
```diff
@@ -0,0 +1,84 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+!use post
+!set outputformat mysql
+!set planner-rules "-INTERSECT_TO_DISTINCT,-EnumerableRules.ENUMERABLE_INTERSECT_RULE,+CoreRules.INTERSECT_TO_SEMI_JOIN"
```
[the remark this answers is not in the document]
mihaibudiu: I suggested to have that prefix implicit, because it makes the text shorter.

core/src/test/resources/sql/planner.iq
```diff
@@ -0,0 +1,84 @@
...
+
+!ok
+
+EnumerableNestedLoopJoin(condition=[IS NOT DISTINCT FROM($0, $1)], joinType=[semi])
+  EnumerableValues(tuples=[[{ 0 }, { 1 }]])
+  EnumerableCalc(expr#0=[{inputs}], expr#1=[0], expr#2=[>($t0, $t1)], EXPR$0=[$t0], $condition=[$t2])
+    EnumerableValues(tuples=[[{ 0 }, { 1 }]])
+!plan
+
+# Test that rules set by "!set planner-rules" are sticky
```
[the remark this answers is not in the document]
mihaibudiu: no different, it just tests that the planner rules are not reset from test to test

core/src/test/resources/sql/planner.iq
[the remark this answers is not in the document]
NobiGo: If both CoreRules and EnumerableRules can be omitted, I think there is no problem. But only omitting CoreRules seems a bit strange. However, the rules we usually test are CoreRules. 
If we can only omit one of them, we need to add some explanations.

mihaibudiu:
I expect most tests will just insert corerules

NobiGo:
> > Updating the process of Rules in Planner makes it seem very complicated. We can't just construct a new Planner directly?
> 
> The planner has a lot of rules, I don't expect you want to insert them all by hand. All the enumerable rules are needed to generate running code.

If you directly rebuild a default Planner, it will come with default Rules. Is it because I misunderstood somewhere?

xiedeyantu:
> > > Updating the process of Rules in Planner makes it seem very complicated. We can't just construct a new Planner directly?
> > 
> > 
> > The planner has a lot of rules, I don't expect you want to insert them all by hand. All the enumerable rules are needed to generate running code.
> 
> If you directly rebuild a default Planner, it will come with default Rules. Is it because I misunderstood somewhere?

When executing the `.iq` file, the planner is no longer rebuild for each SQL statement. The planner is now reused within each connection. We may need to refactor this framework to achieve rebuilding the planner every time.

core/src/test/resources/sql/planner.iq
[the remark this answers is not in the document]
xiedeyantu: @NobiGo This is a good suggestion, I will add a usage note to the `planner.iq` file. WDYT?

testkit/src/main/java/org/apache/calcite/test/QuidemTest.java
```diff
@@ -171,6 +180,26 @@ protected void checkRun(String path) throws Exception {
               int thresholdValue = ((BigDecimal) value).intValue();
               closer.add(Prepare.THREAD_INSUBQUERY_THRESHOLD.push(thresholdValue));
             }
+            // Configures query planner rules via "!set planner-rules" command.
```
xiedeyantu, line 183: Added some instructions

core/src/test/resources/sql/planner.iq
```diff
@@ -0,0 +1,89 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+!use post
+!set outputformat mysql
+
+# The following line is used to update the default planner rules.
```
xiedeyantu, line 21: Added some instructions

mihaibudiu:
Since I have approved, I will wait for @NobiGo to approve before merging

core/src/test/resources/sql/planner.iq
```diff
@@ -0,0 +1,89 @@
...
+# "+" means add a rule, "-" means remove a rule. Only the "CoreRules." can be omitted.
+!set planner-rules "-INTERSECT_TO_DISTINCT,-EnumerableRules.ENUMERABLE_INTERSECT_RULE,+CoreRules.INTERSECT_TO_SEMI_JOIN"
```
mihaibudiu, line 23: I suspect the quotes are actually not necessary, is that right?
but maybe it's useful to keep them, since we may change Quidem to allow multi-line strings

core/src/test/resources/sql/planner.iq
xiedeyantu: I'm very sorry, so I left a mark to make up for my mistake.

core/src/test/resources/sql/planner.iq
xiedeyantu: It is still needed at this stage. If there are no double quotes, the reading will be considered to end when a comma is encountered.

xiedeyantu:
@NobiGo Hi, please take a look, if there some suggestions for me?

xiedeyantu:
@NobiGo Sorry to bother you again, is there anything else I can do?

caicancai (approved):

mihaibudiu:
Another question is whether we wait for a new release of Quidem to pick up the multi-line strings.
Then you can include the upgrade in this PR.
But we should merge this soon, it will be very useful.

xiedeyantu:
> Another question is whether we wait for a new release of Quidem to pick up the multi-line strings. Then you can include the upgrade in this PR. But we should merge this soon, it will be very useful.

I suggest that if this PR only lacks multi-line strings, can it be merged first. CALCITE-6975 may need to provide a 0.12-SNAPSHOT version for validation, otherwise Julian may not release version 0.12. WDYT？

mihaibudiu:
I think he suggested for you to test locally using a snapshot. Did you do that?

xiedeyantu:
> I think he suggested for you to test locally using a snapshot. Did you do that?

Oh, sorry, maybe I misunderstood. I previously modified the `Quidem.java` file directly in calcite for validation, so I will use 0.12-SNAPSHOT for local validation.

xiedeyantu:
> I think he suggested for you to test locally using a snapshot. Did you do that?

I have tested it locally and it works.

NobiGo (approved):

mihaibudiu:
Ok, let's merge this as is, and we can edit once we have multi-line strings.

PR #4302, opened 2025-04-12
[CALCITE-6949] ClickHouse not support floor date to SECOND/MILLISECOND/MICROSECOND/NANOSECOND

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6949

### DIFF
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -5355,6 +5355,61 @@ class RelToSqlConverterTest {
         .withMysql().ok(expectedMysql);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6949">[CALCITE-6949]
+   * ClickHouse not support floor date to SECOND/MILLISECOND/MICROSECOND/NANOSECOND</a>. */
+  @Test void testFloorClickHouse() {
+    String query = "SELECT floor(\"hire_date\" TO YEAR) FROM \"employee\"";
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -5355,6 +5355,61 @@ private void checkLiteral2(String expression, String expected) {
...
+        + "FROM `foodmart`.`employee`";
+    sql(query).withClickHouse().ok(expectedClickHouse);
+
+    String query1 = "SELECT floor(\"hire_date\" TO MONTH) FROM \"employee\"";
+    String expectedClickHouse1 = "SELECT toStartOfMonth(`hire_date`)\n"
+        + "FROM `foodmart`.`employee`";
+    sql(query1).withClickHouse().ok(expectedClickHouse1);
+
+    String query2 = "SELECT floor(\"hire_date\" TO WEEK) FROM \"employee\"";
+    String expectedClickHouse2 = "SELECT toMonday(`hire_date`)\n"
```
mihaibudiu, line 5373: In some places weeks are considered to start on Sunday, so the meaning of the `FLOOR(... WEEK)` function may vary depending on the source dialect or even the locale. But I don't know how this can be fixed.

mihaibudiu (approved):

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xuzifu666: Yes, this toMonday code just a test for previous code and seems different system keep a separate standard. @mihaibudiu

PR #4303, opened 2025-04-13
[CALCITE-5985] FilterTableFunctionTransposeRule should not use "Logical" RelNodes

xiedeyantu:
Please link: [CALCITE-5985](https://issues.apache.org/jira/browse/CALCITE-5985)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
@@ -30,6 +30,7 @@ import org.apache.calcite.rel.core.Project;
 import org.apache.calcite.rel.core.Sample;
 import org.apache.calcite.rel.core.SetOp;
 import org.apache.calcite.rel.core.Sort;
+import org.apache.calcite.rel.core.TableFunctionScan;
 import org.apache.calcite.rel.core.TableScan;
 import org.apache.calcite.rel.core.Union;
 import org.apache.calcite.rel.core.Values;
@@ -42,8 +43,8 @@ import org.apache.calcite.rel.logical.LogicalJoin;
 import org.apache.calcite.rel.logical.LogicalMatch;
 import org.apache.calcite.rel.logical.LogicalProject;
 import org.apache.calcite.rel.logical.LogicalSortExchange;
-import org.apache.calcite.rel.logical.LogicalTableFunctionScan;
 import org.apache.calcite.rel.logical.LogicalWindow;
+import org.apache.calcite.rel.rules.FilterTableFunctionTransposeRule.Config;
 import org.apache.calcite.rel.rules.materialize.MaterializedViewRules;
 import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexUtil;
@@ -285,11 +286,12 @@ public class CoreRules {
   public static final FilterSampleTransposeRule FILTER_SAMPLE_TRANSPOSE =
       FilterSampleTransposeRule.Config.DEFAULT.toRule();
 
-  /** Rule that pushes a {@link LogicalFilter}
-   * past a {@link LogicalTableFunctionScan}. */
+  /** Rule that pushes a {@link Filter}
+   * past a {@link TableFunctionScan}. */
   public static final FilterTableFunctionTransposeRule
-      FILTER_TABLE_FUNCTION_TRANSPOSE =
-      FilterTableFunctionTransposeRule.Config.DEFAULT.toRule();
+      FILTER_TABLE_FUNCTION_TRANSPOSE = Config.DEFAULT
+          .withOperandFor(Filter.class, TableFunctionScan.class)
+          .toRule();
 
   /** Rule that matches a {@link Filter} on a {@link TableScan}. */
   public static final FilterTableScanRule FILTER_SCAN =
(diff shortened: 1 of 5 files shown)


core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -2371,6 +2371,15 @@ private void checkSemiOrAntiJoinProjectTranspose(JoinRelType type) {
         .check();
   }
 
+  @Test void testFilterTableFunctionScanTranspose() {
```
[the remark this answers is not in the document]
xiedeyantu: The existing rules were adjusted to match Filter and TableFunctionScan instead of the "logical" Filter and "logical" TableFunctionScan, with no other changes. A test was added to cover this rule, as it was previously untested.

testkit/src/main/java/org/apache/calcite/test/MockSqlOperatorTable.java
```diff
@@ -208,6 +213,39 @@ public DedupFunction() {
     }
   }
 
+  /** "TEST" user-defined table function. */
+  public static class TestTableFunction extends SqlFunction
```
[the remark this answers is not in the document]
xiedeyantu: The purpose of this TestTableFunction is to illustrate how this rule works, since none of the existing TableFunctions can properly demonstrate its effect.

NobiGo (commented):
Test cases that do not belong to the Logical plan need to be included.

xiedeyantu:
> Test cases that do not belong to the Logical plan need to be included.

This is a great suggestion. However, it's a bit tricky because all SQL are converted into "logical" plans. Could you give me some advice?

mihaibudiu:
> > Test cases that do not belong to the Logical plan need to be included.
> 
> This is a great suggestion. However, it's a bit tricky because all SQL are converted into "logical" plans. Could you give me some advice?

Why did you make this change? You must have a scenario in mind where Logical is wrong. Can you create a short reproduction for that scenario?

xiedeyantu:
> > > Test cases that do not belong to the Logical plan need to be included.
> > 
> > 
> > This is a great suggestion. However, it's a bit tricky because all SQL are converted into "logical" plans. Could you give me some advice?
> 
> Why did you make this change? You must have a scenario in mind where Logical is wrong. Can you create a short reproduction for that scenario?

I saw this jira question and I think it is appropriate. It may not be wrong to use "logical", but it can be extended to non-"logical" operators. NobiGo's suggestion is good, but I haven't found a good way to write a case yet. For example, plans are all Enumerable operators.

mihaibudiu (approved):

mihaibudiu:
Let's wait to see if @NobiGo has comments before merging

xiedeyantu:
@NobiGo Could you help me to have a look again? Thanks!

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
```diff
@@ -288,8 +290,16 @@ private CoreRules() {}
   /** Rule that pushes a {@link LogicalFilter}
    * past a {@link LogicalTableFunctionScan}. */
   public static final FilterTableFunctionTransposeRule
-      FILTER_TABLE_FUNCTION_TRANSPOSE =
-      FilterTableFunctionTransposeRule.Config.DEFAULT.toRule();
+      LOGICAL_FILTER_LOGICAL_TABLE_FUNCTION_TRANSPOSE = Config.DEFAULT
+          .withOperandFor(LogicalFilter.class, LogicalTableFunctionScan.class)
```
asolimando, line 294: Maybe I am missing something here, but both `LogicalFilter` and `LogicalTableFunctionScan` are subclasses of `Filter` and `TableFunctionScan` (respectively), why do we need to duplicate the rule matching condition here? I don't recall us doing it for other Calcite rules which most probably have the same problem (I recall a few changes lifting the Logical operator to the parent class some time ago which didn't require any introduction of new methods nor deprecating anything, but I can't find it at the moment).

If we provide the rule operand matching for `X` won't it work and get applied to `LogicalX`? In that case you don't need to introduce the `LOGICAL_` elements.

xiedeyantu: Because I saw Julian mention this statement in JIRA: "Perhaps retain an instance of the rule that continues to match only LogicalFilter and LogicalTableScan, and also add a new instance that is generic." Therefore, I kept a Logical_X rule.

asolimando: Thanks for the explanation @xiedeyantu, maybe @julianhyde can elaborate more on this? I still don't get the rationale behind the request, I don't feel strongly about it, it's more to understand better.

mihaibudiu:
If we get a second approval I will merge this.

xiedeyantu:
> If we get a second approval I will merge this.

Thanks!

asolimando (approved):
LGTM, let's wait a bit more as there might be pending comments, if not we can merge

xiedeyantu:
@NobiGo I might have to trouble you to take another look, if you have time.

NobiGo (approved):

asolimando:
@xiedeyantu can you squash the commits and make sure the commit title matches the Jira title so that we can merge?

xiedeyantu:
> @xiedeyantu can you squash the commits and make sure the commit title matches the Jira title so that we can merge?

Done!

PR #4306, opened 2025-04-15
[CALCITE-6958] JDBC adapter for MySQL not support floor date to MILLISECOND/MICROSECOND

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6958

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
@@ -54,6 +54,7 @@ import com.google.common.collect.ImmutableList;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.util.List;
+import java.util.Locale;
 
 /**
  * A <code>SqlDialect</code> implementation for the MySQL database.
@@ -347,6 +348,8 @@ public class MysqlSqlDialect extends SqlDialect {
     }
 
     String format;
+    boolean needSubStr = false;
+    int substringLength = 0;
     switch (unit) {
     case YEAR:
       format = "%Y-01-01";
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
```diff
@@ -366,17 +369,39 @@ private static void unparseFloor(SqlWriter writer, SqlCall call) {
     case SECOND:
       format = "%Y-%m-%d %H:%i:%s";
       break;
+    case MILLISECOND:
+      needSubStr = true;
+      format = "%Y-%m-%d %H:%i:%s.%f";
+      substringLength = 23;
```
mihaibudiu, line 375: I am not crazy about this approach, although it probably works.
I think it would be more elegant to only apply SUBSTR to the fractional seconds part (or some arithmetic).
But this is probably more efficient.

Does this work correctly in MySQL when the fractional part is 0 or has trailing zeros?

xuzifu666: Hi @mihaibudiu , thanks for the suggestion, do you mean only use date_format to archive it? 
According to my previous research, currently MySQL only can extract microsecond from the way:
、、、
SELECT microsecond(NOW(6));
>953213
、、、

but seem there is no other way to extract millisecond：
、、、
SELECT MILLISECOND(NOW(6));
>[42000]: FUNCTION test_db.MILLISECOND does not exist
、、、

and seems no other way to get MILLISECOND directly.

I also test trailing zeros in fractional part:
、、、
select DATE_FORMAT(NOW(6), '%Y-%m-%d %H:%i:%s.000'); 
>2025-04-15 02:15:50.000

select DATE_FORMAT(NOW(6), '%Y-%m-%d %H:%i:%s.000'); 
>2025-04-15 02:15:50.000

select DATE_FORMAT(NOW(6), '%Y-%m-%d %H:%i:%s.%f000'); 
>2025-04-15 02:17:10.718958000
、、、

seems not as expected, but users just use this way(substr + dateformat) to get millisecond/microsecond as a universal solution in many real production practice.

WDYT?

mihaibudiu: I didn't mean to call MILLISECONDS, I was thinking to just use `SUBSTRING(format(date, "%f"), ...)`
But I guess these are equivalent.

xuzifu666: Yes,I agree，they are equivalent. I had referred to the relevant MySQL documents about date and format usage, seems current version maybe a suitable way to archive it. Some senior MySQL users also use this way to get millisecond/microsecond from timestamp. If can offer some more suggestions, I would do more research about it. @mihaibudiu

mihaibudiu (approved):

PR #4308, opened 2025-04-15
[CALCITE-6955] PruneEmptyRules does not handle the all attribute of SetOp correctly

xiedeyantu:
Please see: [CALCITE-6955](https://issues.apache.org/jira/browse/CALCITE-6955)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
@@ -406,6 +406,9 @@ public abstract class PruneEmptyRules {
               + RelOptUtil.toString(union);
           if (nonEmptyInputs == 0) {
             relBuilder.push(union).empty();
+          } else if (nonEmptyInputs == 1 && !union.all) {
+            relBuilder.distinct();
+            relBuilder.convert(union.getRowType(), true);
           } else {
             relBuilder.union(union.all, nonEmptyInputs);
             relBuilder.convert(union.getRowType(), true);
@@ -447,6 +450,9 @@ public abstract class PruneEmptyRules {
               + RelOptUtil.toString(minus);
           if (nonEmptyInputs == 0) {
             relBuilder.push(minus).empty();
+          } else if (nonEmptyInputs == 1 && !minus.all) {
+            relBuilder.distinct();
+            relBuilder.convert(minus.getRowType(), true);
           } else {
             relBuilder.minus(minus.all, nonEmptyInputs);
             relBuilder.convert(minus.getRowType(), true);
--- a/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
@@ -3744,6 +3744,34 @@ LogicalProject(X=[$0], Y=[$1])
 ]]>
     </Resource>
   </TestCase>
+  <TestCase name="testEmptyFilterProjectUnion2">
+    <Resource name="sql">
+      <![CDATA[select * from (
+select * from (values (10, 1), (30, 3), (30, 3)) as t (x, y)
+union
+select * from (values (20, 2))
+)
+where x + y > 30]]>
+    </Resource>
+    <Resource name="planBefore">
+      <![CDATA[
+LogicalProject(X=[$0], Y=[$1])
+  LogicalFilter(condition=[>(+($0, $1), 30)])
+    LogicalUnion(all=[false])
+      LogicalProject(X=[$0], Y=[$1])
+        LogicalValues(tuples=[[{ 10, 1 }, { 30, 3 }, { 30, 3 }]])
+      LogicalProject(EXPR$0=[$0], EXPR$1=[$1])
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
```diff
@@ -406,6 +406,9 @@ public interface UnionEmptyPruneRuleConfig extends PruneEmptyRule.Config {
               + RelOptUtil.toString(union);
           if (nonEmptyInputs == 0) {
             relBuilder.push(union).empty();
+          } else if (nonEmptyInputs == 1 && !union.all) {
```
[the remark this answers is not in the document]
xiedeyantu: If intersect contains two children, one is values(1) and the other is values(), then the result will be an empty set. This is another logic, and the original implementation is as expected.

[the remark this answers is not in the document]
mihaibudiu: that is only true if the number of empty inputs is > 0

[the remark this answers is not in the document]
xiedeyantu: Yes, this rule requires matching an empty value.
```
UnionEmptyPruneRuleConfig DEFAULT = ImmutableUnionEmptyPruneRuleConfig.of()
        .withOperandSupplier(b0 ->
            b0.operand(Union.class).unorderedInputs(b1 ->
                b1.operand(Values.class)
                    .predicate(Values::isEmpty).noInputs()))
        .withDescription("Union");
```

mihaibudiu (approved):

xiedeyantu:
> Frankly, this would be very amenable to having an end-to-end test showing incorrect results before. This could be done easily after we merge #4301

I totally agree with you. If some real data may be exposed.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -3768,6 +3796,22 @@ LogicalIntersect(all=[false])
     <Resource name="planAfter">
       <![CDATA[
 LogicalValues(tuples=[[]])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testEmptyIntersect2">
```
NobiGo, line 3802: This test case does not have a corresponding one in Java.

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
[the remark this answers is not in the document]
NobiGo: > If intersect contains two children, one is values(1) and the other is values(), then the result will be an empty set. This is another logic, and the original implementation is as expected.

Do we have test cases covering this?

NobiGo (commented):
Needs end-to-end tests.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
xiedeyantu: Done!

core/src/main/java/org/apache/calcite/rel/rules/PruneEmptyRules.java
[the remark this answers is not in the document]
xiedeyantu: testEmptyIntersect can cover this case

xiedeyantu:
@NobiGo I've added the `.iq` test cases.

mihaibudiu:
Waiting for @NobiGo to confirm that he's satisfied

core/src/test/resources/sql/set-op.iq
```diff
@@ -258,4 +258,32 @@ UNION
...
+union
+select * from (values (20, 2)) as t (x, y) where x > 30;
++----+---+
+| X  | Y |
++----+---+
+| 10 | 1 |
+| 30 | 3 |
++----+---+
+(2 rows)
+
```
NobiGo, line 273: We need to verify whether the optimizations related to this PR have taken effect through `!plan`.

xiedeyantu: The corresponding plan is already present in RelOptRulesTest.xml. Would you like me to add an additional one here as well.

NobiGo: Yes, this is different. In RelOptRulesTest, we have actively added the plan. For the iq file, we are using the default optimization plan.

NobiGo: If this plan doesn't take effect, this SQL statement should still be able to execute successfully.

xiedeyantu: Oops! You're completely right. I'll add it soon.

xiedeyantu: @NobiGo Done!

NobiGo (approved):

NobiGo:
@xiedeyantu Please squash the commit info.

xiedeyantu:
> @xiedeyantu Please squash the commit info.

Done!

PR #4310, opened 2025-04-15
[CALCITE-6960] The tests for SemiJoinRemoveRule should explicitly include a semi join

xiedeyantu:


mihaibudiu (approved):

mihaibudiu:
I guess the tests are equivalent, but now you make the before plan explicit in the xml

xiedeyantu:
> I guess the tests are equivalent, but now you make the before plan explicit in the xml

Yes, I think it would be easier for other people to understand.

PR #4312, opened 2025-04-16
[CALCITE-6957] The RelOptRulesTest tests should fail if the xml file contains tests that do not exist in Java

xiedeyantu:


core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -586,62 +586,6 @@ LogicalAggregate(group=[{7, 10}])
   LogicalJoin(condition=[=($7, $9)], joinType=[left])
     LogicalTableScan(table=[[CATALOG, SALES, EMP]])
     LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
-]]>
```
asolimando, line 589: We can clearly see that such a check was long due

core/src/test/java/org/apache/calcite/test/JavaXmlConsistencyChecker.java
```diff
@@ -0,0 +1,143 @@
...
+    Set<String> javaMethodNames = parseMethodNamesFromJava(javaFilePath);
+
+    List<String> missingMethods = new ArrayList<>();
+    for (String testCase : testCaseNames) {
+      if (!javaMethodNames.contains(testCase)) {
+        missingMethods.add(testCase);
+      }
+    }
+
+    if (!missingMethods.isEmpty()) {
```
asolimando, line 90: At the moment, if you have a method in Java and no corresponding XML element, what happens?

If it fails, then it's OK to check the way you propose. If not, maybe we should check that the two set of names is identical.

Can you please try and update here?

asolimando (commented):
LGTM, I have left a question about the check strategy, the rest is OK for me

core/src/test/java/org/apache/calcite/test/JavaXmlConsistencyChecker.java
asolimando: OK great, so adding a Java method which is missing the XML counterpart doesn't fail silently, then I am fine with the current proposal, and it's great it applies to all XML-based test suites.

asolimando (approved):
LGTM!

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
xiedeyantu: Yes, I actually overlooked this problem. @NobiGo  spotted it when reviewing my code, so I tried to fix it.

mihaibudiu:
This looks fine, but I don't understand the last comment in the Jira, so I will wait to merge this.

xiedeyantu:
> This looks fine, but I don't understand the last comment in the Jira, so I will wait to merge this.

I just want to give an example to explain the problem more specifically. If there is any ambiguity, I can delete it.

NobiGo (commented):
Have you reviewed the DiffRepository test method? Can we improve this method? Currently, any new files that need testing must be added manually.

xiedeyantu:
> Have you reviewed the DiffRepository test method? Can we improve this method? Currently, any new files that need testing must be added manually.

I reviewed this and noticed that it also requires explicit class name, such as:
﻿
```
  @Override RelOptFixture fixture() {
    RelOptFixture fixture = super.fixture()
        .withDiffRepos(DiffRepository.lookup(RelOptRulesTest.class));
    diffRepos = fixture.diffRepos();
    return fixture;
  }
```

Additionally, I have a concern. Currently, all XML files are located under `core/src/test/resources/org/apache/calcite/test/`, alongside `.csv` files. Would it be appropriate for me to scan all XML files under this path programmatically, or should we keep explicitly specifying the files?

NobiGo:
> > Have you reviewed the DiffRepository test method? Can we improve this method? Currently, any new files that need testing must be added manually.
> 
> I reviewed this and noticed that it also requires explicit class name, such as:
> 
> ```
>   @Override RelOptFixture fixture() {
>     RelOptFixture fixture = super.fixture()
>         .withDiffRepos(DiffRepository.lookup(RelOptRulesTest.class));
>     diffRepos = fixture.diffRepos();
>     return fixture;
>   }
> ```
> 
> Additionally, I have a concern. Currently, all XML files are located under `core/src/test/resources/org/apache/calcite/test/`, alongside `.csv` files. Would it be appropriate for me to scan all XML files under this path programmatically, or should we keep explicitly specifying the files?

Yes. This is the way the test cases are used. When developers apply this for verification, I hope that the corresponding verification procedures have already been included in it.

xiedeyantu:
> > > Have you reviewed the DiffRepository test method? Can we improve this method? Currently, any new files that need testing must be added manually.
> > 
> > 
> > I reviewed this and noticed that it also requires explicit class name, such as:
> > ```
> >   @Override RelOptFixture fixture() {
> >     RelOptFixture fixture = super.fixture()
> >         .withDiffRepos(DiffRepository.lookup(RelOptRulesTest.class));
> >     diffRepos = fixture.diffRepos();
> >     return fixture;
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Additionally, I have a concern. Currently, all XML files are located under `core/src/test/resources/org/apache/calcite/test/`, alongside `.csv` files. Would it be appropriate for me to scan all XML files under this path programmatically, or should we keep explicitly specifying the files?
> 
> Yes. This is the way the test cases are used. When developers apply this for verification, I hope that the corresponding verification procedures have already been included in it.

I didn't fully understand your description. But I seem to understand what you mean, you want to be smarter. Maybe I still haven't been able to solve my confusion. DiffRepository also requires specified input to tell it what the file is. I don't know which Java files need to be compared with which XML files. All I can think of is to explicitly specify them.

xiedeyantu:
@NobiGo I refactored the code, please see if this is acceptable？

asolimando:
> @NobiGo I refactored the code, please see if this is acceptable？

@xiedeyantu checkstyle is failing, can you check locally and fix the problems?

xiedeyantu:
> > @NobiGo I refactored the code, please see if this is acceptable？
> 
> @xiedeyantu checkstyle is failing, can you check locally and fix the problems?

Fixed, Thanks!

NobiGo:
I will check this PR carefully. It will take some time.

xiedeyantu:
> I will check this PR carefully. It will take some time.

Thanks!

xiedeyantu:
@NobiGo has a better implement, so I close this PR to support that.

PR #4318, opened 2025-04-18
[CALCITE-6944] Align toSqlString with SQL std for Table Args in PTF

juntaozhang:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6944

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlCall.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlCall.java
@@ -116,16 +116,23 @@ public abstract class SqlCall extends SqlNode {
         getOperandList());
   }
 
+  private boolean needsParentheses(SqlWriter writer, int leftPrec, int rightPrec) {
+    if (getKind() == SqlKind.SET_SEMANTICS_TABLE) {
+      return false;
+    }
+    final SqlOperator operator = getOperator();
+    return leftPrec > operator.getLeftPrec()
+        || (operator.getRightPrec() <= rightPrec && (rightPrec != 0))
+        || writer.isAlwaysUseParentheses() && isA(SqlKind.EXPRESSION)
+        || (operator.getRightPrec() <= rightPrec + 1 && isA(SqlKind.COMPARISON));
+  }
+
   @Override public void unparse(
       SqlWriter writer,
       int leftPrec,
       int rightPrec) {
-    final SqlOperator operator = getOperator();
     final SqlDialect dialect = writer.getDialect();
-    if (leftPrec > operator.getLeftPrec()
-        || (operator.getRightPrec() <= rightPrec && (rightPrec != 0))
-        || writer.isAlwaysUseParentheses() && isA(SqlKind.EXPRESSION)
-        || (operator.getRightPrec() <= rightPrec + 1 && isA(SqlKind.COMPARISON))) {
+    if (needsParentheses(writer, leftPrec, rightPrec)) {
       final SqlWriter.Frame frame = writer.startList("(", ")");
       dialect.unparseCall(writer, this, 0, 0);
       writer.endList(frame);
--- a/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
+++ b/testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
@@ -4705,7 +4705,15 @@ public class SqlParserTest {
     // test one partition key for input table
     final String sql = "select * from table(topn(table orders partition by productid, 3))";
     final String expected = "SELECT *\n"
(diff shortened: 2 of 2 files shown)


mihaibudiu:
The PR and commit must use the same title as the JIRA, so you will have to edit them to make them identical.

core/src/main/java/org/apache/calcite/sql/SqlCall.java
```diff
@@ -116,16 +116,21 @@ public int operandCount() {
         getOperandList());
   }
 
+  private boolean isParenthesesRequired(SqlWriter writer, int leftPrec, int rightPrec) {
```
mihaibudiu, line 119: I would rename this function to "needsParentheses".

core/src/main/java/org/apache/calcite/sql/SqlCall.java
```diff
@@ -116,16 +116,21 @@ public int operandCount() {
...
+    final SqlOperator operator = getOperator();
+    boolean condition = leftPrec > operator.getLeftPrec()
+        || (operator.getRightPrec() <= rightPrec && (rightPrec != 0))
+        || writer.isAlwaysUseParentheses() && isA(SqlKind.EXPRESSION)
+        || (operator.getRightPrec() <= rightPrec + 1 && isA(SqlKind.COMPARISON));
+    return condition && !(getKind() == SqlKind.SET_SEMANTICS_TABLE);
```
mihaibudiu, line 125: It looks that this condition is easier to write starting with this test:
`if (getKind() == SET_SEMANTICS_TABLE) return false;`
Then followed by the original expression

juntaozhang:
> The PR and commit must use the same title as the JIRA, so you will have to edit them to make them identical.

Thank you for the correction.  I change JIRA name, becuase it can match the final conclusion.

core/src/main/java/org/apache/calcite/sql/SqlCall.java
juntaozhang: 👌

core/src/main/java/org/apache/calcite/sql/SqlCall.java
juntaozhang: ok

mihaibudiu (approved):

mihaibudiu:
In the issue you filed you gave a specific test case which failed.
I don't see that covered in the existing tests.

mihaibudiu:
In other words, did you solve the problem which motivated this PR?

juntaozhang:
> In the issue you filed you gave a specific test case which failed. I don't see that covered in the existing tests.
hi @mihaibudiu can you help clerify which specific test case not pass?  Let me double check.

mihaibudiu:
```
"f(a => TABLE t PARTITION BY f1 ORDER BY f2, b => 1)";
```

juntaozhang:
> In other words, did you solve the problem which motivated this PR?
Just as the failed cases described in JIRA, these two cases of after PR can pass.
```
    String sqlExpected = "f(a => TABLE t PARTITION BY f1 ORDER BY f2, b => 1)";
    String sqlActual = parseExpression(sqlExpected)
        .toSqlString(new AnsiSqlDialect(SqlDialect.EMPTY_CONTEXT)).getSql();
    parseExpression(sqlActual);
```

```
    String sqlExpected = "f(TABLE t PARTITION BY f1 ORDER BY f2,1)";
    String sqlActual = parseExpression(sqlExpected)
        .toSqlString(new AnsiSqlDialect(SqlDialect.EMPTY_CONTEXT)).getSql();
    parseExpression(sqlActual);
```

juntaozhang:
> ```
> "f(a => TABLE t PARTITION BY f1 ORDER BY f2, b => 1)";
> ```

That's weird,  it's ok in my side, are there any special environments or configurations?
image

mihaibudiu:
I don't see this in the new tests

juntaozhang:
> I don't see this in the new tests

I know your concerns, le't me clarify it, just as the conclusion of our discussion at the end in JIRA：
```	
2. The toSqlString method should not return an unparsable and nonstandard result to users, the parentheses need be removed from the <table argument>.
```

So the exsiting cases only need to modify to cover the PR, because we don't change the parse logic, so new case is not needed.

If you think the named parameter syntax need to enrich the test cases, in this point I am happy add it in the PR.

juntaozhang:
> I don't see this in the new tests

done

mihaibudiu (approved):

mihaibudiu:
This seems ready to merge, please squash the commits to a single commit with the same message as the JIRA title

juntaozhang:
> This seems ready to merge, please squash the commits to a single commit with the same message as the JIRA title

done, thank you for your guidance.

PR #4319, opened 2025-04-18
[CALCITE-6964] SqlDelete#getOperandList return operands' both order and size not match with SqlDelete#setOperand

joewee:
Please view the detail : [CALCITE-6964]

### DIFF
--- a/core/src/test/java/org/apache/calcite/sql/SqlCallOperandsTest.java
+++ b/core/src/test/java/org/apache/calcite/sql/SqlCallOperandsTest.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.sql;
+
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/sql/SqlCallOperandsTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasSize;
+
+/**
+ * Tests for SqlCall operands' order and size match with set.
+ */
+public class SqlCallOperandsTest {
```
xuzifu666, line 33: You can add jira link to here.

xuzifu666 (commented):
LGTM, only left a minor comment

core/src/test/java/org/apache/calcite/sql/SqlCallOperandsTest.java
joewee: Ok, i have added the jira link .

core/src/test/java/org/apache/calcite/sql/SqlCallOperandsTest.java
```diff
@@ -0,0 +1,72 @@
...
+            null,
+            null,
+            null,
+            null);
+    SqlIdentifier alias = new SqlIdentifier("alias", SqlParserPos.ZERO);
+    sqlDelete.setOperand(0, targetTable);
+    sqlDelete.setOperand(1, condition);
+    sqlDelete.setOperand(2, sourceSelect);
+    sqlDelete.setOperand(3, alias);
+    final List<SqlNode> operandList = sqlDelete.getOperandList();
```
mihaibudiu, line 65: Here you are checking that the list you write is the list you get.
So if the list is permuted, this test will still succeed.
But what you should check is that the operands correspond to their intended use.
So you should read them back using getTargetTable(), getAlias(), etc.

joewee: Maybe the test case method name is ambiguity,  this case purpose is ensure  getOperandList return operands' order and size match with setOperand to solve getOperandList return a mismatch size and order operands like before:  
`  @Override public List<SqlNode> getOperandList() {
    return ImmutableNullableList.of(targetTable, condition, alias);
  }`
So should i modify the test method name to 'testSqlDeleteGetOperandsSizeAndOrderMatchWithSetOperand' ? @mihaibudiu

mihaibudiu: I don't see how this test does what you claim.
If you swap alias and condition the test will still pass.
What you should check is that the condition is indeed obtained by sqlDelete.getCondition()

mihaibudiu (approved):

mihaibudiu:

joewee:
> Can you please squash the commits so we can merge this?

OK

PR #4314, opened 2025-04-17
[CALCITE-6967] Unparsing STARTS_WITH/ENDS_WITH/BIT functions is incorrect for the Clickhouse dialect

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6967

PR #4320, opened 2025-04-18
[CALCITE-6969] Support ClickHouse in SqlLibrary

xuzifu666:
Prior to the https://github.com/apache/calcite/pull/4314
description in JIRA：https://issues.apache.org/jira/browse/CALCITE-6969

### DIFF
--- a/site/_docs/reference.md
+++ b/site/_docs/reference.md
@@ -2758,6 +2758,7 @@ The 'C' (compatibility) column contains value:
 * 'p' for PostgreSQL ('fun=postgresql' in the connect string),
 * 'r' for Amazon RedShift ('fun=redshift' in the connect string),
 * 's' for Apache Spark ('fun=spark' in the connect string).
+* 'i' for ClickHouse('fun=clickhouse' in the connect string).
 
 One operator name may correspond to multiple SQL dialects, but with different
 semantics.
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlLibrary.java
@@ -85,7 +85,10 @@ public enum SqlLibrary {
   SNOWFLAKE("f", "snowflake"),
   /** A collection of operators that are in Apache Spark but not in standard
    * SQL. */
-  SPARK("s", "spark");
+  SPARK("s", "spark"),
+  /** A collection of operators that are in ClickHouse but not in standard
+   * SQL. */
+  CLICKHOUSE("i", "clickhouse");
 
   /** Map from {@link Enum#name() name} and {@link #fun} to library. */
   public static final Map<String, SqlLibrary> MAP;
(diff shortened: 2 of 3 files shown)


xuzifu666:
@mihaibudiu PTAL when you have a free time~

site/_docs/reference.md
```diff
@@ -2758,6 +2758,7 @@ The 'C' (compatibility) column contains value:
 * 'p' for PostgreSQL ('fun=postgresql' in the connect string),
 * 'r' for Amazon RedShift ('fun=redshift' in the connect string),
 * 's' for Apache Spark ('fun=spark' in the connect string).
+* 'i' for ClickHouse('fun=clickhouse' in the connect string).
```
xiedeyantu, line 2761: Ask a question not related to PR, why is there only one letter to represent the dialect here? It feels very abstract.

xiedeyantu:
LGTM

site/_docs/reference.md
xuzifu666: As I know it is a doc standard and there a document checker for it @xiedeyantu

mihaibudiu: If you see the reference page, each function has a list of dialects:
https://calcite.apache.org/docs/reference.html#dialect-specific-operators
I suppose the intent is to keep that list shorter so it can be displayed nicely.

mihaibudiu:
> @mihaibudiu PTAL when you have a free time~

Waiting for free time is not a good idea, it may never happen. I will do it now.

mihaibudiu (approved):
This change is good, but I suspect we cannot close the issue until someone went through the list of functions and labeled the ones that already work in clickhouse.

xuzifu666:
> This change is good, but I suspect we cannot close the issue until someone went through the list of functions and labeled the ones that already work in clickhouse.

I would do this work,  list and labeled related functions in clickhouse.@mihaibudiu

xuzifu666:
Hi @mihaibudiu I had list related funcions and tagged them work well in ClickHouse, Mainly concentrated on string and bit functions（can refer to https://issues.apache.org/jira/browse/CALCITE-6967）：
<html>
<body>

Calcite | ClickHouse
-- | --
String
STARTS_WITH | startsWith
ENDS_WITH | endsWith
Bit
BITAND | bitAnd
BITOR | bitOr
BITXOR | bitXor
BITNOT | bitNot


</body>
</html>
and I update the test in Calcite,  once the PR merged I can work continue in https://github.com/apache/calcite/pull/4314

NobiGo (approved):

mihaibudiu:
I was assuming you would also edit reference.md with this information

xuzifu666:
> I was assuming you would also edit reference.md with this information

Yes, I would update these infomations in follow pr which adapt funcsions. @mihaibudiu

caicancai:
> > I was assuming you would also edit reference.md with this information
> 
> Yes, I would update these infomations in follow pr which adapt funcsions. @mihaibudiu

I hope to complete it in PR, because it is not a difficult thing

caicancai:
I raised some questions in jira

caicancai (approved):

PR #4321, opened 2025-04-18
[CALCITE-6957] The RelOptRulesTest tests should fail if the xml file …

NobiGo:
…contains tests that do not exist in Java

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/DiffRepository.java
+++ b/testkit/src/main/java/org/apache/calcite/test/DiffRepository.java
@@ -46,13 +46,18 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Writer;
+import java.lang.reflect.Method;
 import java.net.URL;
 import java.util.AbstractList;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.stream.Collectors;
 import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
@@ -194,6 +199,8 @@ public class DiffRepository {
   private final @Nullable DiffRepository baseRepository;
   private final int indent;
   private final ImmutableSortedSet<String> outOfOrderTests;
+  private final Boolean existsMethodOnlyInXml;
+  private final SortedMap<String, Node> xmlTestCases;
   private Document doc;
   private final Element root;
   private final URL refFile;
@@ -213,7 +220,7 @@ public class DiffRepository {
    */
   private DiffRepository(URL refFile, File logFile,
       @Nullable DiffRepository baseRepository, @Nullable Filter filter,
-      int indent) {
+      int indent, Set<String> javaTestMethods) {
     this.baseRepository = baseRepository;
     this.filter = filter;
     this.indent = indent;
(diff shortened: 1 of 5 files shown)


mihaibudiu:
What's the relation between this and #4312?

NobiGo:
> What's the relation between this and #4312?

This PR can be directly replaced with #4321.

After this PR, we can solve the only existing XML methods by copying and pasting the file_actual.xml and don't need to delete them one by one. 
In addition, the exception information will be clearer. For example, in PR's unit test DiffRepositoryTest#testMethodOnlyExistsInXml:
The exception will be:
```
java.lang.IllegalArgumentException: Actual and reference files differ. If you are adding new tests, replace the reference file with the current actual file, after checking its content.
diff /calcite/testkit/build/diffrepo/test/org/apache/calcite/test/DiffRepositoryTest_actual.xml /calcite/testkit/src/test/resources/org/apache/calcite/test/DiffRepositoryTest.xml
26d25
<   <TestCase name="testMethodOnlyExistsInXml1"/>

```

testkit/src/main/java/org/apache/calcite/test/DiffRepository.java
```diff
@@ -641,13 +652,47 @@ private static ImmutableSortedSet<String> validate(Element root) {
...
-            && previousName.compareTo(name) > 0) {
-          outOfOrderNames.add(name);
-        }
-        previousName = name;
       }
     }
+    return testCases;
+  }
+
+  /** Validates if there is methods that only exists in XML.
```
mihaibudiu, line 660: This should also be "Checks"
If there "are methods that only exist in XML"

testkit/src/test/java/org/apache/calcite/test/DiffRepositoryTest.java
```diff
@@ -44,4 +44,17 @@ public class DiffRepositoryTest {
...
+    boolean assertPassed = false;
+    DiffRepository r = DiffRepository.lookup(DiffRepositoryTest.class);
+    final String actual = "testMethodOnlyExistsInXml1";
+    try {
+      r.checkActualAndReferenceFiles();
+      assertPassed = true;
+    } catch (IllegalArgumentException e) {
+      assertThat(e.getMessage(), containsString(actual));
+    }
+    assertThat("First assertion must always fail", assertPassed, is(false));
```
mihaibudiu, line 58: I think "unreachable" is a better description

mihaibudiu (approved):

xiedeyantu:
This implementation is fantastic! If developers add tests in Java and XML formats, there will be no need to focus on case consistency issues. I will close my PR #4312 and switch to supporting this.

xiedeyantu (approved):

testkit/src/test/java/org/apache/calcite/test/DiffRepositoryTest.java
```diff
@@ -44,4 +44,17 @@ public class DiffRepositoryTest {
     }
     assertThat("First assertion must always fail", assertPassed, is(false));
   }
+
+  @Test void testMethodOnlyExistsInXml() {
+    boolean assertPassed = false;
+    DiffRepository r = DiffRepository.lookup(DiffRepositoryTest.class);
+    final String actual = "testMethodOnlyExistsInXml1";
```
xiedeyantu, line 51: There's a small question, should this be "testMethodOnlyExistsInXml1_ actual"?

testkit/src/test/java/org/apache/calcite/test/DiffRepositoryTest.java
NobiGo: I don't make changes here because it is consistent with another unit test.

testkit/src/test/java/org/apache/calcite/test/DiffRepositoryTest.java
NobiGo: You can refer to the comments above about the detailed exception info, both are acceptable.

NobiGo: The file here should be `DiffRepositoryTest_actual.xml`.

xiedeyantu: I understand, this may just be a placeholder.

PR #4322, opened 2025-04-20
[CALCITE-6973] Add rule to convert Minus to Filter

xiedeyantu:
Please see: [CALCITE-6973](https://issues.apache.org/jira/browse/CALCITE-6973)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
@@ -383,6 +383,10 @@ public class CoreRules {
   public static final IntersectToSemiJoinRule INTERSECT_TO_SEMI_JOIN =
           IntersectToSemiJoinRule.Config.DEFAULT.toRule();
 
+  /** Rule that translates a {@link Minus} to {@link Filter}. */
+  public static final MinusToFilterRule MINUS_TO_FILTER =
+      MinusToFilterRule.Config.DEFAULT.toRule();
+
   /** Rule that translates a distinct
    * {@link Minus} into a group of operators
    * composed of {@link Union}, {@link Aggregate}, etc. */
--- a/core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
@@ -1,0 +1,121 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Filter;
+import org.apache.calcite.rel.core.Minus;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.tools.RelBuilder;
+
+import org.immutables.value.Value;
+
+/**
+ * Rule that replaces {@link Minus} operator with {@link Filter}
+ * when both inputs are from the same source with only filter conditions differing.
+ * Only inspect a single {@link Filter} layer in the inputs of {@link Minus}.
+ * For inputs with nested {@link Filter}s, apply {@link CoreRules#FILTER_MERGE}
+ * as a preprocessing step.
+ *
+ * <p>Example transformation:
+ * <blockquote><pre>
+ * SELECT mgr, comm FROM emp WHERE mgr = 12
+ * EXCEPT
+ * SELECT mgr, comm FROM emp WHERE comm = 5
+ *
+ * to
+ *
+ * SELECT DISTINCT mgr, comm FROM emp
+ * WHERE mgr = 12 AND NOT(comm = 5)
+ * </pre></blockquote>
+ */
+@Value.Enclosing
+public class MinusToFilterRule
+    extends RelRule<MinusToFilterRule.Config>
+    implements TransformationRule {
+
+  /** Creates an MinusToFilterRule. */
+  protected MinusToFilterRule(Config config) {
+    super(config);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    final Minus minus = call.rel(0);
+    if (minus.all || minus.getInputs().size() != 2) {
+      return;
+    }
+    final RelBuilder builder = call.builder();
+    final RelNode leftInput = call.rel(1);
+    final Filter rightInput = call.rel(2);
+
+    if (!RexUtil.isDeterministic(rightInput.getCondition())) {
+      return;
+    }
+
+    RelNode leftBase;
+    RexNode leftCond = null;
+    if (leftInput instanceof Filter) {
+      Filter leftFilter = (Filter) leftInput;
+      leftBase = leftFilter.getInput().stripped();
+      leftCond = leftFilter.getCondition();
+    } else {
+      leftBase = leftInput.stripped();
+    }
+
+    final RelNode rightBase = rightInput.getInput().stripped();
+    if (!leftBase.equals(rightBase)) {
+      return;
+    }
+
+    // Right input is Filter, right cond should be not null
+    final RexNode finalCond = leftCond != null
+        ? builder.and(leftCond, builder.not(rightInput.getCondition()))
+        : builder.not(rightInput.getCondition());
+
+    builder.push(leftBase)
+        .filter(finalCond)
+        .distinct();
+
+    call.transformTo(builder.build());
+  }
+
+  /** Rule configuration. */
+  @Value.Immutable
+  public interface Config extends RelRule.Config {
+    Config DEFAULT = ImmutableMinusToFilterRule.Config.of()
+        .withOperandFor(Minus.class, RelNode.class, Filter.class);
+
+    @Override default MinusToFilterRule toRule() {
+      return new MinusToFilterRule(this);
+    }
+
+    /** Defines an operand tree for the given classes. */
+    default Config withOperandFor(Class<? extends Minus> minusClass,
+        Class<? extends RelNode> relNodeClass, Class<? extends Filter> filterClass) {
+      return withOperandSupplier(
+          b0 -> b0.operand(minusClass).inputs(
+              b1 -> b1.operand(relNodeClass).anyInputs(),
+              b2 -> b2.operand(filterClass).anyInputs()))
+          .as(Config.class);
+    }
+  }
+}
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -10421,4 +10421,30 @@ class RelOptRulesTest extends RelOptTestBase {
         .withRule(CoreRules.JOIN_CONDITION_PUSH)
         .check();
   }
+
+  /** Test case of
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6973">[CALCITE-6973]
+   * Add rule for convert Minus to Filter</a>. */
+  @Test void testMinusToFilterRule() {
+    final String sql = "SELECT mgr, comm FROM emp WHERE mgr = 12\n"
+        + "EXCEPT\n"
+        + "SELECT mgr, comm FROM emp WHERE comm = 5\n";
+    sql(sql)
+        .withPreRule(CoreRules.PROJECT_FILTER_TRANSPOSE)
+        .withRule(CoreRules.MINUS_TO_FILTER)
+        .check();
+  }
+
+  /** Test case of
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6973">[CALCITE-6973]
+   * Add rule for convert Minus to Filter</a>. */
+  @Test void testMinusToFilterRule2() {
+    final String sql = "SELECT mgr, comm FROM emp\n"
+        + "EXCEPT\n"
+        + "SELECT mgr, comm FROM emp WHERE comm = 5\n";
+    sql(sql)
+        .withPreRule(CoreRules.PROJECT_FILTER_TRANSPOSE)
+        .withRule(CoreRules.MINUS_TO_FILTER)
+        .check();
+  }
 }
--- a/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
@@ -9015,6 +9015,59 @@ LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2])
 ]]>
     </Resource>
   </TestCase>
+  <TestCase name="testMinusToFilterRule">
+    <Resource name="sql">
+      <![CDATA[SELECT mgr, comm FROM emp WHERE mgr = 12
+EXCEPT
+SELECT mgr, comm FROM emp WHERE comm = 5
+]]>
+    </Resource>
+    <Resource name="planBefore">
+      <![CDATA[
+LogicalMinus(all=[false])
+  LogicalFilter(condition=[=($0, 12)])
+    LogicalProject(MGR=[$3], COMM=[$6])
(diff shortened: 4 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
```diff
@@ -0,0 +1,137 @@
...
+  @Value.Immutable
+  public interface Config extends RelRule.Config {
+    Config DEFAULT = ImmutableMinusToFilterRule.Config.of()
+        .withOperandFor(Minus.class);
+
+    @Override default MinusToFilterRule toRule() {
+      return new MinusToFilterRule(this);
+    }
+
+    /** Limit how much complexity can increase during merging.
```
mihaibudiu, line 125: I am not sure this is necessary here, because the result is always linear in the size of the input.

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
```diff
@@ -0,0 +1,137 @@
...
+      return;
+    }
+
+    // Verify they come from the same source
+    if (!leftBase.equals(rightBase)) {
+      return;
+    }
+
+    final RexNode finalCond =
+        builder.and(extractFilterCond(leftInput), builder.not(rightCond));
```
mihaibudiu, line 85: where is the check that both filters produce the same fields?

xiedeyantu: Yes, by identifying the beginning and ending double quotes.

xiedeyantu: I understand that there is no need to judge the same column, as long as the condition of `right` is reversed. WDYT

NobiGo: How about `SELECT mgr, comm FROM emp WHERE mgr = 12 EXCEPT SELECT mgr1, comm1 FROM emp WHERE comm = 5`? We need to ensure that the two inputs return the same columns. Only then will this Rule be valid.

xiedeyantu: Maybe [here](https://github.com/apache/calcite/pull/4322/files/1a743cc2f3947a79a084bd1fe433a2740961a2cf#diff-7dde415fabdd836a8181e96f6405a572bd5a9c07d67394d596dbbd8039e1454aR80) can solve this question？I will check if it comes from the same source.

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
xiedeyantu: Done!

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
NobiGo: Get it.

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
```diff
@@ -0,0 +1,130 @@
...
+        .distinct();
+
+    call.transformTo(builder.build());
+  }
+
+  /**
+   * Returns the base source of the given node,
+   * skipping first filter.
+   */
+  private static RelNode getBaseSource(RelNode node) {
```
mihaibudiu, line 97: I would call this "getUnfilteredInput"
Note that "source" rather implies a table, as opposed to an input of the current dataflow node.

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
```diff
@@ -0,0 +1,130 @@
...
+    RelNode stripped = node.stripped();
+    if (stripped instanceof Filter) {
```
mihaibudiu, line 99: can this ever not be a Filter?

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
```diff
@@ -0,0 +1,130 @@
...
+      return node;
+    }
+  }
+
+  private static RexNode extractFilterCond(RelNode node) {
+    RelNode stripped = node.stripped();
+    if (stripped instanceof Filter) {
+      return ((Filter) stripped).getCondition();
+    }
+    return node.getCluster().getRexBuilder().makeLiteral(true);
```
mihaibudiu, line 111: why accept non-filters and treat them like "true"?
this seems to do unnecessary rewrites

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
xiedeyantu: It can indeed be other operators. I want to convert the plan through some pre-rules first, and then make this rule effective. The problem to be dealt with here is similar to the problem dealt with by some existing rules.

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
mihaibudiu: you can match on Minus(Filter, Filter)

xiedeyantu: Yes, it should be an empty values.

xiedeyantu: > you can match on Minus(Filter, Filter)

Great idea, this is better.

xiedeyantu:
@mihaibudiu I have refactored this pr, please take another look. Thanks!

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
```diff
@@ -0,0 +1,133 @@
...
+  /** Creates an MinusToFilterRule. */
+  protected MinusToFilterRule(Config config) {
+    super(config);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    final Minus minus = call.rel(0);
+    if (minus.all || minus.getInputs().size() != 2) {
```
mihaibudiu, line 65: I expect size cannot be != 2, since the rule would not be applied
So this is fine, but it could also be `assert getInputs().size() == 2;`

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
```diff
@@ -0,0 +1,133 @@
...
+
+    call.transformTo(builder.build());
+  }
+
+  private static Pair<RelNode, @Nullable RexNode> getUnfilteredInputAndMergeCond(
+      final RelNode node) {
+    RexBuilder builder = node.getCluster().getRexBuilder();
+    RelNode stripped = node.stripped();
+    RexNode cond = null;
+    while (stripped instanceof Filter) {
```
mihaibudiu, line 100: here you handle nested filters; perhaps you should document this in the JavaDoc.

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
xiedeyantu: I only considered the 2-way scenario. Changing to assert may allow non-2-way scenarios to enter the rule, which may cause other input filters to be ignored. I think that handwritten SQL generally does not have n-way scenarios, so 2-way is supported for the time being. Do I need to make any further changes?

xiedeyantu: This test can build 3-way Minus
```
  @Test void testMinusToFilterRule2() {
        .scan("EMP")
        .project(b.field("MGR"), b.field("COMM"))
        .filter(b.lessThan(b.field("MGR"), b.literal(20)))
        .scan("EMP")
        .project(b.field("MGR"), b.field("COMM"))
        .filter(b.lessThan(b.field("MGR"), b.literal(10)))
        .scan("EMP")
        .project(b.field("MGR"), b.field("COMM"))
        .filter(b.lessThan(b.field("MGR"), b.literal(9)))
        .minus(false, 3)
        .build();

    relFn(relFn)
        .withRule(CoreRules.MINUS_TO_FILTER)
        .check();
  }
```

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
xiedeyantu: Done!

mihaibudiu: I mean the JavaDoc for the rule.

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
mihaibudiu: I said that withOperandSupplier will only accept minus with two inputs. isn't that right?

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
xiedeyantu: Done!

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -10407,4 +10407,38 @@ private void checkLoptOptimizeJoinRule(LoptOptimizeJoinRule rule) {
...
+  }
+
+  /** Test case of
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6973">[CALCITE-6973]
+   * Add rule for convert Minus to Filter</a>. */
+  @Test void testMinusToFilterRule2() {
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .scan("EMP")
+        .project(b.field("MGR"), b.field("COMM"))
+        .filter(b.lessThan(b.field("MGR"), b.literal(20)))
```
NobiGo, line 10431: We have `FilterMergeRule` to merge Filter, so I think this Rule should focus on Minus to Filter. There is no need to do everything under this Rule.

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
```diff
@@ -0,0 +1,141 @@
...
+    @Override default MinusToFilterRule toRule() {
+      return new MinusToFilterRule(this);
+    }
+
+    /** Defines an operand tree for the given classes. */
+    default Config withOperandFor(Class<? extends Minus> minusClass,
+        Class<? extends RelNode> relNodeClass, Class<? extends Filter> filterClass) {
+      return withOperandSupplier(
+          b0 -> b0.operand(minusClass).inputs(
+              b1 -> b1.operand(relNodeClass).anyInputs(),
```
NobiGo, line 136: We should have a test case covering RelNode and Filter. Now only include Filter and Filter.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
asolimando: I strongly agree, individual rules should be kept as simple as possible and should rely on other rules whenever it makes sense like in this case

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
```diff
@@ -381,6 +381,10 @@ private CoreRules() {}
   public static final IntersectToSemiJoinRule INTERSECT_TO_SEMI_JOIN =
           IntersectToSemiJoinRule.Config.DEFAULT.toRule();
 
+  /** Rule to translates a {@link Minus} to {@link Filter}. */
```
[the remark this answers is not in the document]
xiedeyantu: Done!

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xiedeyantu: @NobiGo @asolimando Done!

mihaibudiu:
I think it is a good design principle to have each rule be as simple as possible, and not do work that another rule can do.
However, there is a balancing act in practice, because the complexity of the search process is exponential in the number of rules that can be applied at each step, so there may be performance benefits in having a rule perform optimizations that it *knows* for sure are beneficial - rather than hoping the optimizer will guess what to do next.

xiedeyantu:
> I think it is a good design principle to have each rule be as simple as possible, and not do work that another rule can do. However, there is a balancing act in practice, because the complexity of the search process is exponential in the number of rules that can be applied at each step, so there may be performance benefits in having a rule perform optimizations that it _knows_ for sure are beneficial - rather than hoping the optimizer will guess what to do next.

I agree with @mihaibudiu 's opinion, what do you think? @asolimando @NobiGo

xiedeyantu:
> Notice that I said there is an argument for both cases, but I didn't say which one is right.

I understand what you mean. I actually rewrote it according to their ideas to see what most people think.

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
```diff
@@ -0,0 +1,118 @@
...
+
+    final RelNode leftBase = leftInput instanceof Filter
+        ? ((Filter) leftInput).getInput().stripped()
+        : leftInput.stripped();
+    final RelNode rightBase = rightInput.getInput().stripped();
+    if (!leftBase.equals(rightBase)) {
+      return;
+    }
+
+    RexNode leftCond = leftInput instanceof Filter
```
asolimando, line 82: Nit: You are checking twice this condition in a complex expression, I think it would improve readability to extract the boolean variable and use that instead. WDYT?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -10407,4 +10407,30 @@ private void checkLoptOptimizeJoinRule(LoptOptimizeJoinRule rule) {
...
+  @Test void testMinusToFilterRule() {
+    final String sql = "SELECT mgr, comm FROM emp WHERE mgr = 12\n"
+        + "EXCEPT\n"
+        + "SELECT mgr, comm FROM emp WHERE comm = 5\n";
+    sql(sql)
+        .withPreRule(CoreRules.PROJECT_FILTER_TRANSPOSE)
+        .withRule(CoreRules.MINUS_TO_FILTER)
+        .check();
+  }
+
```
asolimando, line 10423: Can we add a test case involving `FILTER_MERGE` rule too just to check that we can cover equivalent transformations as the previous version?

asolimando:
> I think it is a good design principle to have each rule be as simple as possible, and not do work that another rule can do. However, there is a balancing act in practice, because the complexity of the search process is exponential in the number of rules that can be applied at each step, so there may be performance benefits in having a rule perform optimizations that it _knows_ for sure are beneficial - rather than hoping the optimizer will guess what to do next.

I agree on your statement, I have baked into rules some "transformations" (e.g., predicate simplifications despite there is a rule to do that) when we knew we had to do it. In this case, `FILTER_MERGE` rule is part of the standard rule set in Calcite and it's a "cheap" rule, that's why I am more inclined to avoid merging filters in this rule, but I am happy to discuss trade-offs if you think otherwise.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xiedeyantu: In my opinion, it may not be necessary to add this case. This case may only show whether the filter is merged, which has little to do with this PR.

xiedeyantu:
> > I think it is a good design principle to have each rule be as simple as possible, and not do work that another rule can do. However, there is a balancing act in practice, because the complexity of the search process is exponential in the number of rules that can be applied at each step, so there may be performance benefits in having a rule perform optimizations that it _knows_ for sure are beneficial - rather than hoping the optimizer will guess what to do next.
> 
> I agree on your statement, I have baked into rules some "transformations" (e.g., predicate simplifications despite there is a rule to do that) when we knew we had to do it. In this case, `FILTER_MERGE` rule is part of the standard rule set in Calcite and it's a "cheap" rule, that's why I am more inclined to avoid merging filters in this rule, but I am happy to discuss trade-offs if you think otherwise.

I think you've convinced me. I just need to focus on the issues I should be dealing with.

asolimando:
> Bailu Ding, Vivek Narasayya and Surajit Chaudhuri (2024), “Extensible Query Optimizers in Practice”, Foundations and Trends® in Databases: Vol. 14, No. 3-4, pp 186–402. DOI: 10.1561/1900000077

Great reference, thanks for sharing, Mihai!

asolimando (approved):
LGTM

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
xiedeyantu: Done!

asolimando:
@xiedeyantu, sorry for noticing late but I think we should rephrase as "Add rule to convert Minus to Filter" or "Add rule for converting Minus to Filter", and update the Jira title/commit (and references in code) accordingly. For the commit title it can be done upon the final squashing.

mihaibudiu:
I think this has enough approvals, can you please squash the commits?

xiedeyantu:
> squash

Thanks! I have squashed and rebased main branch.

xiedeyantu:
> I think this has enough approvals, can you please squash the commits?

Thanks! I have squashed and rebased main branch.

xiedeyantu:
This ci pr-head checker failure has nothing to do with this PR.

PR #4323, opened 2025-04-21
[CALCITE-6974] Default typesystem has incorrect limits for DECIMAL for Presto/MySQL/Phoenix

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6974

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
@@ -71,10 +71,25 @@ public class MysqlSqlDialect extends SqlDialect {
             return 65535;
           case TIMESTAMP:
             return 6;
+          case DECIMAL:
+            return 65;
           default:
             return super.getMaxPrecision(typeName);
           }
         }
+
+        // We can refer to document of MySQL 5.x and 8.x:
+        // https://dev.mysql.com/doc/refman/8.4/en/precision-math-decimal-characteristics.html
+        // https://dev.mysql.com/doc/refman/5.7/en/precision-math-decimal-characteristics.html
+        @Override public int getMaxScale(SqlTypeName typeName) {
+          switch (typeName) {
+          case DECIMAL:
+            return 30;
+          default:
+            return super.getMaxScale(typeName);
+          }
+        }
+
         @Override public int getDefaultPrecision(SqlTypeName typeName) {
           if (typeName == SqlTypeName.CHAR) {
             return RelDataType.PRECISION_NOT_SPECIFIED;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
```diff
@@ -71,10 +71,22 @@ public class MysqlSqlDialect extends SqlDialect {
...
           case TIMESTAMP:
             return 6;
+          case DECIMAL:
+            return 65;
           default:
             return super.getMaxPrecision(typeName);
           }
         }
+
+        @Override public int getMaxScale(SqlTypeName typeName) {
```
mihaibudiu, line 81: I think this is wrong: https://dev.mysql.com/doc/refman/8.4/en/precision-math-decimal-characteristics.html
The docs say that 30 is the max value.
Please double check these values before you make the fixes.

xuzifu666: It‘s my mistake!  thanks for the correction，I would doube check them again. @mihaibudiu

xuzifu666: Had changed it after check from documents，and test precision and scale in Presto/MySQL/Phoenix real environment which as expected. PTAL @mihaibudiu

mihaibudiu: ideally you can add here links to the documentation and specify also the version of the database, so people can check whether the values are still accurate as things change

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java
xuzifu666: Done，had add document links of MySQL 5.x and 8.x(which are the most used version), thanks for the suggestion. @mihaibudiu

core/src/main/java/org/apache/calcite/sql/dialect/PhoenixSqlDialect.java
```diff
@@ -17,22 +17,51 @@
...
 import org.apache.calcite.sql.type.AbstractSqlType;
+import org.apache.calcite.sql.type.SqlTypeName;
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * A <code>SqlDialect</code> implementation for the Apache Phoenix database.
  */
 public class PhoenixSqlDialect extends SqlDialect {
+  public static final RelDataTypeSystem TYPE_SYSTEM =
```
mihaibudiu, line 36: are there links for these other type systems too?

xuzifu666: Done, had added links for other systems, thanks @mihaibudiu

PR #4324, opened 2025-04-22
[Calcite-6946] Expand predicates from disjunction for inputs of Join

silundong:
Similar to Calcite-6914, we can expand redundant predicates from the disjunction and push them down. However, the predicates expanded by Calcite-6914 must belong to a single table, that is, if join_type does not restrict pushdown, the redundant predicates can be pushed directly on TableScan.

However, pushing predicates down to the inputs of Join is also very useful ([link](https://github.com/apache/calcite/pull/4267#discussion_r2021576009)), for example:

```
select t1.id from t1, t2, t3
where t1.id = t2.id
and t1.id = t3.id
and (
    (t1.age < 50 and t3.age > 20)
    or
    (t2.weight > 70 and t3.height < 180)
) 
```
Because `(t1.age < 50 and t3.age > 20)` or `(t2.weight > 70 and t3.height < 180)` a is a disjunction and involves multiple tables, it cannot be pushed down.

However, we can expand it to `(t1.age < 50 or t2.weight > 70)` , `(t3.age > 20 or t3.height < 180)`, and push them down to both sides of Join.

More details: https://issues.apache.org/jira/browse/CALCITE-6946

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForJoinInputsRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForJoinInputsRule.java
@@ -1,0 +1,242 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelOptUtil;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.core.Filter;
+import org.apache.calcite.rel.core.Join;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.tools.RelBuilder;
+import org.apache.calcite.util.ImmutableBitSet;
+
+import org.immutables.value.Value;
+
+import java.util.Map;
+
+/**
+ * Rule to expand disjunction for join inputs in condition of a {@link Filter} or {@link Join}.
+ * For example:
+ *
+ * <blockquote><pre>{@code
+ * select t1.id from t1, t2, t3
+ * where t1.id = t2.id
+ * and t1.id = t3.id
+ * and (
+ *     (t1.age < 50 and t3.age > 20)
+ *     or
+ *     (t2.weight > 70 and t3.height < 180)
+ * )
+ * }</pre></blockquote>
+ *
+ * <p>we can expand to obtain the condition
+ *
+ * <blockquote><pre>{@code
+ * select t1.id from t1, t2, t3
+ * where t1.id = t2.id
+ * and t1.id = t3.id
+ * and (
+ *     (t1.age < 50 and t3.age > 20)
+ *     or
+ *     (t2.weight > 70 and t3.height < 180)
+ * )
+ * and (t1.age < 50 or t2.weight > 70)
+ * and (t3.age > 20 or t3.height < 180)
+ * }</pre></blockquote>
+ *
+ * <p>new generated predicates are redundant, but they could be pushed down to
+ * join inputs([t1, t2], [t3]) and reduce the cardinality.
+ *
+ * <p>Note: This rule is similar to the {@link ExpandDisjunctionForTableRule},
+ * but this rule divides the predicates by Join inputs.
+ *
+ * @see CoreRules#EXPAND_FILTER_DISJUNCTION_LOCAL
+ * @see CoreRules#EXPAND_JOIN_DISJUNCTION_LOCAL
+ */
+@Value.Enclosing
+public class ExpandDisjunctionForJoinInputsRule
+    extends RelRule<ExpandDisjunctionForJoinInputsRule.Config>
+    implements TransformationRule {
+
+  /**
+   * Creates a ExpandDisjunctionForJoinInputsRule.
+   */
+  protected ExpandDisjunctionForJoinInputsRule(Config config) {
+    super(config);
+  }
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    config.matchHandler().accept(this, call);
+  }
+
+  private static void matchFilter(ExpandDisjunctionForJoinInputsRule rule, RelOptRuleCall call) {
+    Filter filter = call.rel(0);
+    Join join = call.rel(1);
+    RelBuilder builder = call.builder();
+
+    ExpandDisjunctionForJoinInputsHelper helper =
+        new ExpandDisjunctionForJoinInputsHelper(
+            join.getLeft().getRowType().getFieldCount(),
+            join.getRight().getRowType().getFieldCount(),
+            join.getJoinType().canPushLeftFromAbove(),
+            join.getJoinType().canPushRightFromAbove(),
+            builder,
+            rule.config.bloat());
+    Map<String, RexNode> expandResult = helper.expand(filter.getCondition());
+
+    RexNode newCondition =
+        builder.and(
+            filter.getCondition(),
+            expandResult.getOrDefault(helper.leftKey, builder.literal(true)),
+            expandResult.getOrDefault(helper.rightKey, builder.literal(true)));
+    if (newCondition.equals(filter.getCondition())) {
+      return;
+    }
+    Filter newFilter = filter.copy(filter.getTraitSet(), join, newCondition);
+    call.transformTo(newFilter);
+  }
+
+  private static void matchJoin(ExpandDisjunctionForJoinInputsRule rule, RelOptRuleCall call) {
+    Join join = call.rel(0);
+    RelBuilder builder = call.builder();
+
+    ExpandDisjunctionForJoinInputsHelper helper =
+        new ExpandDisjunctionForJoinInputsHelper(
+            join.getLeft().getRowType().getFieldCount(),
+            join.getRight().getRowType().getFieldCount(),
+            join.getJoinType().canPushLeftFromWithin(),
+            join.getJoinType().canPushRightFromWithin(),
+            builder,
+            rule.config.bloat());
+    Map<String, RexNode> expandResult = helper.expand(join.getCondition());
+
+    RexNode newCondition =
+        call.builder().and(
+            join.getCondition(),
+            expandResult.getOrDefault(helper.leftKey, builder.literal(true)),
+            expandResult.getOrDefault(helper.rightKey, builder.literal(true)));
+    if (newCondition.equals(join.getCondition())) {
+      return;
+    }
+    Join newJoin =
+        join.copy(
+            join.getTraitSet(),
+            newCondition,
+            join.getLeft(),
+            join.getRight(),
+            join.getJoinType(),
+            join.isSemiJoinDone());
+    call.transformTo(newJoin);
+  }
+
+  /**
+   * Helper class to expand predicates.
+   */
+  private static class ExpandDisjunctionForJoinInputsHelper
+      extends RexUtil.ExpandDisjunctionHelper<String> {
+
+    private final ImmutableBitSet leftBitmap;
+
+    private final ImmutableBitSet rightBitmap;
+
+    private final boolean canPushLeft;
+
+    private final boolean canPushRight;
+
+    private final String leftKey = "left";
+
+    private final String rightKey = "right";
+
+    private ExpandDisjunctionForJoinInputsHelper(
+        int leftFieldCount,
+        int rightFieldCount,
+        boolean canPushLeft,
+        boolean canPushRight,
+        RelBuilder relBuilder,
+        int maxNodeCount) {
+      super(relBuilder, maxNodeCount);
+      this.leftBitmap = ImmutableBitSet.range(0, leftFieldCount);
+      this.rightBitmap = ImmutableBitSet.range(leftFieldCount, leftFieldCount + rightFieldCount);
+      this.canPushLeft = canPushLeft;
+      this.canPushRight = canPushRight;
+    }
+
+    @Override protected boolean canReturnEarly(
+        RexNode condition,
+        Map<String, RexNode> additionalConditions) {
+      boolean earlyReturn = false;
+
+      ImmutableBitSet inputRefs = RelOptUtil.InputFinder.bits(condition);
+      if (inputRefs.isEmpty()) {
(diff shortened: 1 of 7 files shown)


silundong:
@mihaibudiu Could you please review the code?

asolimando:
@silundong could you check the first four issues highlighted by Sonar? https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=4324&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForJoinInputsRule.java
```diff
@@ -0,0 +1,388 @@
...
+
+import org.immutables.value.Value;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Rule to expand disjunction for join inputs in condition of a {@link Filter} or {@link Join}.
```
asolimando, line 39: Can you try to rephrase this sentence?

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForJoinInputsRule.java
```diff
@@ -0,0 +1,388 @@
...
+        return new HashMap<>();
+      }
+    }
+
+    /**
+     * Expand predicates recursively that can be pushed down to Join inputs.
+     *
+     * @param condition   Predicate to be expanded
+     * @return  A map from Join left/right input to a (combined) predicate that can be pushed down
+     * and only depends on columns of the join inputs.
```
asolimando, line 206: Maybe the extra "one" makes it more precise?

```suggestion
     * and only depends on columns of one the join inputs.
```

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForJoinInputsRule.java
```diff
@@ -0,0 +1,388 @@
...
+          iterator.remove();
+          continue;
+        } else {
+          forMergeNodeCount = forMergeMap.get(entry.getKey()).nodeCount();
+        }
+        RexNode mergedRex =
+            relBuilder.or(
+                entry.getValue(),
+                forMergeMap.get(entry.getKey()));
+        int oriCount = entry.getValue().nodeCount() + forMergeNodeCount;
```
asolimando, line 330: Nit: originalCount isn't much longer and makes it more readable IMO

asolimando (commented):
LGTM but I feel there is some redundancy with the previous rules that could be easily reduced, could you give it a shot @silundong?

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForJoinInputsRule.java
silundong: I will update it and the documentation you mentioned above.

silundong:
> LGTM but I feel there is some redundancy with the previous rules that could be easily reduced, could you give it a shot @silundong?

Sorry, I didn’t understand that. Could you elaborate a little bit?

asolimando:
> > LGTM but I feel there is some redundancy with the previous rules that could be easily reduced, could you give it a shot @silundong?
> 
> Sorry, I didn’t understand that. Could you elaborate a little bit?

I have reviewed CALCITE-6914 as well and lots of the code implemented here is nearly identical to code provided there, I am wondering if we could abstract some parts a bit and re-use them here. I haven't tried to do that, it would be great if you could take a look and give it a shot if it sounds doable.

silundong:
> > > LGTM but I feel there is some redundancy with the previous rules that could be easily reduced, could you give it a shot @silundong?
> > 
> > 
> > Sorry, I didn’t understand that. Could you elaborate a little bit?
> 
> I have reviewed [CALCITE-6914](https://issues.apache.org/jira/browse/CALCITE-6914) as well and lots of the code implemented here is nearly identical to code provided there, I am wondering if we could abstract some parts a bit and re-use them here. I haven't tried to do that, it would be great if you could take a look and give it a shot if it sounds doable.

Yes, they are similar. However, their logic in splitting predicates is not consistent (single table/same side of Join inputs), so it may not be easy to reuse, but I will try it. :)

silundong:
@mihaibudiu Could you please take a look?

core/src/main/java/org/apache/calcite/rex/RexUtil.java
```diff
@@ -2799,6 +2800,172 @@ private RexNode or(Iterable<? extends RexNode> nodes) {
     }
   }
 
+  /**
+   * Helper class that expands predicates from disjunctions(splited by K).
```
asolimando, line 2804:
```suggestion
   * Helper class that expands predicates from disjunctions (splitted by K).
```

core/src/main/java/org/apache/calcite/rex/RexUtil.java
```diff
@@ -2799,6 +2800,172 @@ private RexNode or(Iterable<? extends RexNode> nodes) {
...
+   *
+   * @param <K> The type used to split the predicates.
```
asolimando, line 2806: I'd reference to `K` as "an input source K" or similar, which can be a table or a join depending on how the rule is being instantiated. I don't feel strongly about the name, but I think we need to characterize more precisely what this `K` means, I understand what this means only because I have seen the code already, I feel it's too abstract as it's phrased now

asolimando (commented):
Thanks @silundong for improving the re-use and modularity of the PR.

I have left a couple of comments on the documentation, making it more abstract/modular requires some extra effort to makes it precise and concrete enough but that's the last concern I have on the contribution.

core/src/main/java/org/apache/calcite/rex/RexUtil.java
silundong: I'll try to describe it in more detail

core/src/main/java/org/apache/calcite/rex/RexUtil.java
```diff
@@ -2799,6 +2800,175 @@ private RexNode or(Iterable<? extends RexNode> nodes) {
     }
   }
 
+  /**
+   * Helper class that expands predicates from disjunctions (splited by K).
```
asolimando, line 2804:
```suggestion
   * Helper class that expands predicates from disjunctions (split by K).
```

core/src/main/java/org/apache/calcite/rel/rules/ExpandDisjunctionForTableRule.java
```diff
@@ -207,86 +204,35 @@ private static void matchJoin(ExpandDisjunctionForTableRule rule, RelOptRuleCall
...
-     * Expand predicates recursively that can be pushed down to single table.
-     *
-     * @param condition   Predicate to be expanded
-     * @return  A map from a table to a (combined) predicate that can be pushed down
-     * and only depends on columns of the table
-     */
-    private Map<RexTableInputRef.RelTableRef, RexNode> expandDeep(RexNode condition) {
-      Map<RexTableInputRef.RelTableRef, RexNode> additionalConditions = new HashMap<>();
-
+    @Override protected boolean canEarlyReturn(
```
mihaibudiu, line 219: canReturnEarly sounds a little better.

core/src/main/java/org/apache/calcite/rex/RexUtil.java
```diff
@@ -2799,6 +2800,175 @@ private RexNode or(Iterable<? extends RexNode> nodes) {
...
+   *           predicates that can be pushed down to a single table from the disjunction, K is
+   *           {@link RelTableRef}; if you want to expand predicates that can be pushed down to
+   *           Join inputs from the disjunction, K is string whose value is 'left' or 'right'.
+   */
+  public abstract static class ExpandDisjunctionHelper<K> {
+    private final RelBuilder relBuilder;
+
+    private final int maxNodeCount;
+
+    // Used to record the number of redundant expressions expanded.
```
mihaibudiu, line 2816: Is this the increase of the expression size?
Maybe that's what it should be called "expressionIncreaseAmount"

core/src/main/java/org/apache/calcite/rex/RexUtil.java
```diff
@@ -2799,6 +2800,175 @@ private RexNode or(Iterable<? extends RexNode> nodes) {
...
+        return expandDeep(condition);
+      } catch (OverLimitException e) {
+        return new HashMap<>();
+      }
+    }
+
+    /**
+     * Expand predicates recursively that can be pushed down to a single K.
+     *
+     * @param condition   Predicate to be expanded
```
mihaibudiu, line 2836: can you add a java doc for K?
I know it's present above, but it's worth repeating.

core/src/main/java/org/apache/calcite/rex/RexUtil.java
```diff
@@ -2799,6 +2800,175 @@ private RexNode or(Iterable<? extends RexNode> nodes) {
...
+     */
+    private void combinePredicatesUsingAnd(
+        Map<K, RexNode> baseMap,
+        Map<K, RexNode> forMergeMap) {
+      for (Map.Entry<K, RexNode> entry : forMergeMap.entrySet()) {
+        RexNode mergedRex =
+            relBuilder.and(
+                entry.getValue(),
+                baseMap.getOrDefault(entry.getKey(), relBuilder.literal(true)));
+        int oriCount = entry.getValue().nodeCount()
```
mihaibudiu, line 2902: originalCount?

core/src/main/java/org/apache/calcite/rex/RexUtil.java
```diff
@@ -2799,6 +2800,175 @@ private RexNode or(Iterable<? extends RexNode> nodes) {
...
+     */
+    protected void checkExpandCount(int changeCount) {
+      currentCount += changeCount;
+      if (maxNodeCount > 0 && currentCount > maxNodeCount) {
+        throw OverLimitException.INSTANCE;
+      }
+    }
+
+    /** Exception to catch when we pass the limit. */
+    private static class OverLimitException extends ControlFlowException {
```
mihaibudiu, line 2966: I think I saw similar exception used elsewhere. Is there a public version you could use?
private is fine, but if we can reuse something, maybe we should.
Why do you need to subclass ControlFlowException and you can't just reuse it? Could be thrown elsewhere?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -10447,4 +10447,41 @@ private void checkLoptOptimizeJoinRule(LoptOptimizeJoinRule rule) {
         .withRule(CoreRules.MINUS_TO_FILTER)
         .check();
   }
+
+  @Test void testExpandFilterDisjunctionForJoinInput() {
```
mihaibudiu, line 10451: this could really benefit from some quidem tests, using the new capability of specifying optimizations

core/src/main/java/org/apache/calcite/rex/RexUtil.java
silundong: I've seen other places use it through subclass extends, maybe that makes it easier to understand why the exception is thrown. There's an OverflowError in CnfHelper, which also prevents combinatorial explosion. Maybe a common exception can be abstracted under RexUtil, which can be used by both CnfHelper and ExpandDisjunctionHelper?

core/src/main/java/org/apache/calcite/rex/RexUtil.java
silundong: I've seen other places use it through subclass extends, maybe that makes it easier to understand why the exception is thrown. There's an OverflowError in CnfHelper, which also prevents combinatorial explosion. Maybe a common exception can be abstracted under RexUtil, which can be used by both CnfHelper and ExpandDisjunctionHelper?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
silundong: Sorry, I don't understand what I should do here. Could you elaborate a little bit?

mihaibudiu: Can you write a couple of tests in a file named .iq using this optimization?
See this for an example: https://github.com/apache/calcite/pull/4301

silundong: Thank you, I will update it.

silundong:
> @silundong please avoid force-pushing while the review is still in progress unless strictly necessary to rebase on `main`, by reading the latest comments from Mihai I realized he has asked for changes I already asked and that got lost as the GH UI can't keep up when you force push

Sorry. I may need to use the new features provided by #4301 to complete the test mentioned by Mihai.

asolimando:
> > @silundong please avoid force-pushing while the review is still in progress unless strictly necessary to rebase on `main`, by reading the latest comments from Mihai I realized he has asked for changes I already asked and that got lost as the GH UI can't keep up when you force push
> 
> Sorry. I may need to use the new features provided by #4301 to complete the test mentioned by Mihai.

As I said, whenever needed it's not a problem, but let's avoid it unless necessary as it makes the review process more complicated. We are anyway always squashing before merging, don't worry about piling up intermediate commits until the PR is finalized.

core/src/test/resources/sql/planner.iq
```diff
@@ -88,4 +88,61 @@ EnumerableIntersect(all=[false])
...
+!ok
+
+EnumerableHashJoin(condition=[AND(=($0, $6), OR(AND(>($1, 11), <=($7, 32)), AND(<($5, 255), >=($8, 344))))], joinType=[inner])
+  EnumerableMergeJoin(condition=[AND(=($0, $3), OR(>($1, 11), <($5, 255)))], joinType=[inner])
+    EnumerableValues(tuples=[[{ 1, 11, 111 }, { 2, 12, 122 }, { 3, 13, 133 }, { 4, 14, 144 }, { 5, 15, 155 }]])
+    EnumerableValues(tuples=[[{ 1, 21, 211 }, { 2, 22, 222 }, { 3, 23, 233 }, { 4, 24, 244 }, { 5, 25, 255 }]])
+  EnumerableCalc(expr#0..2=[{inputs}], expr#3=[32], expr#4=[<=($t1, $t3)], expr#5=[344], expr#6=[>=($t2, $t5)], expr#7=[OR($t4, $t6)], proj#0..2=[{exprs}], $condition=[$t7])
+    EnumerableValues(tuples=[[{ 1, 31, 311 }, { 2, 32, 322 }, { 3, 33, 333 }, { 4, 34, 344 }, { 5, 35, 355 }]])
+!plan
+
```
xiedeyantu, line 147: Please add `!set planner-rules original` here to avoid affecting the following cases.

core/src/main/java/org/apache/calcite/rex/RexUtil.java
```diff
@@ -2984,6 +2970,9 @@ private static class RexShiftShuttle extends RexShuttle {
     }
   }
 
+  /** Exception to catch when an expression explodes exponentially. */
+  private static class ExponentialExplosionError extends ControlFlowException {}
```
mihaibudiu: It doesn't have to be exponential, sometime even a small polynomial can be too much.
I would call this "PlanTooComplex," since it indicates that optimizing the plan produces a result that is too complex, either at the Rel or at the Rex level (add this to the JavaDoc).

mihaibudiu (approved):

mihaibudiu:
I think you can also squash the commits, but please consider also renaming the exception.

silundong:
Squashed. Thanks for reviewing!

PR #4327, opened 2025-04-23
[CALCITE-6979] Invalid unparse for IS TRUE,IS FALSE,IS NOT TRUE and IS NOT FALSE in ClickHouseDialect

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6979

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/ClickHouseSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/ClickHouseSqlDialect.java
@@ -21,6 +21,7 @@ import org.apache.calcite.config.NullCollation;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
 import org.apache.calcite.rel.type.RelDataTypeSystemImpl;
+import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.sql.SqlAbstractDateTimeLiteral;
 import org.apache.calcite.sql.SqlBasicTypeNameSpec;
 import org.apache.calcite.sql.SqlCall;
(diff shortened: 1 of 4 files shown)


NobiGo (changes requested):

core/src/main/java/org/apache/calcite/sql/dialect/ClickHouseSqlDialect.java
```diff
@@ -112,6 +116,65 @@ public ClickHouseSqlDialect(Context context) {
     return CalendarPolicy.SHIFT;
   }
 
+  @Override public RexNode prepareUnparse(RexNode arg) {
```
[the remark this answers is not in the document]
xuzifu666: Thanks for the suggestion, would changed as it. @NobiGo

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
```diff
@@ -64,6 +68,79 @@ public static void unparseHiveTrim(
     }
   }
 
+  /**
+   * Unparses IS_TURE/IS_FALSE/IS_NOT_TRUE/IS_NOT_FALSE to right format for dialects.
+   *
+   * <p>For example :
+   *
+   * <blockquote><pre>
```
NobiGo, line 76: 1)Unparses IS TRUE,IS FALSE,IS NOT TRUE and IS NOT FALSE.
2) Generalize this expression using A( A IS NULL ), and explain that an exception will be thrown when A is a non-deterministic function.

xuzifu666: Had changed, please help to review with the comments whether is OK, thanks~ @NobiGo

NobiGo (approved):

NobiGo:
We'll wait for 48 hours to see if there are any other comments.

PR #4331, opened 2025-04-24
[CALCITE-6982] Removes cast from string support with IS_NULL/IS_NOT_NULL

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6982

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -354,7 +354,8 @@ public abstract class SqlImplementor {
     case IS_NOT_NULL:
       final RexCall call0 = (RexCall) node;
       final RexNode o = call0.operands.get(0);
-      if (o.getKind() == SqlKind.CAST) {
+      if (o.getKind() == SqlKind.CAST
+          && dialect.supportsImplicitTypeCoercion((RexCall) o)) {
         return call0.clone(call0.getType(),
             ImmutableList.of(((RexCall) o).operands.get(0)));
       }
--- a/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
+++ b/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
@@ -350,6 +350,15 @@ public abstract class SqlImplementor {
    */
   private static RexNode stripCastFromString(RexNode node, SqlDialect dialect) {
     switch (node.getKind()) {
+    case IS_NULL:
+    case IS_NOT_NULL:
+      final RexCall call0 = (RexCall) node;
+      final RexNode o = call0.operands.get(0);
+      if (o.getKind() == SqlKind.CAST) {
+        return call0.clone(call0.getType(),
+            ImmutableList.of(((RexCall) o).operands.get(0)));
+      }
+      break;
     case EQUALS:
     case IS_NOT_DISTINCT_FROM:
     case NOT_EQUALS:
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -8270,7 +8270,7 @@ class RelToSqlConverterTest {
 
   /** Test case for
    * <a href="https://issues.apache.org/jira/browse/CALCITE-6982">[CALCITE-6982]
-   * Removes cast from string also support with IS_NULL/IS_NOT_NULL</a>. */
+   * Removes cast from string support with IS_NULL/IS_NOT_NULL</a>. */
   @Test void testRemoveCastStringInForceConversion() {
     final String query = "select \"employee_id\" "
         + "from \"foodmart\".\"employee\" "
--- a/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
+++ b/core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
@@ -8268,22 +8268,27 @@ class RelToSqlConverterTest {
     sql(query2).ok(expected2);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6982">[CALCITE-6982]
+   * Removes cast from string also support with IS_NULL/IS_NOT_NULL</a>. */
   @Test void testCastInStringIntegerComparison() {
     final String query = "select \"employee_id\" "
         + "from \"foodmart\".\"employee\" "
-        + "where 10 = cast('10' as int) and \"birth_date\" = cast('1914-02-02' as date) or "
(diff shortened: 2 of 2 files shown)


core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -8268,22 +8268,27 @@ private void checkLiteral2(String expression, String expected) {
...
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6982">[CALCITE-6982]
+   * Removes cast from string also support with IS_NULL/IS_NOT_NULL</a>. */
   @Test void testCastInStringIntegerComparison() {
     final String query = "select \"employee_id\" "
         + "from \"foodmart\".\"employee\" "
-        + "where 10 = cast('10' as int) and \"birth_date\" = cast('1914-02-02' as date) or "
+        + "where 10 is distinct from cast('10' as int) and 10 = cast('10' as int) and \"birth_date\" = cast('1914-02-02' as date) or "
```
NobiGo, line 8277: Please add a new test case to verify this.

NobiGo (commented):
The Cast is necessary in basic behavior. For the Database that supports 10=‘10’ and enables automatic forced type conversion, we can remove it.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -350,6 +350,15 @@ public static SqlNode convertConditionToSqlNode(RexNode node,
    */
   private static RexNode stripCastFromString(RexNode node, SqlDialect dialect) {
     switch (node.getKind()) {
+    case IS_NULL:
+    case IS_NOT_NULL:
+      final RexCall call0 = (RexCall) node;
+      final RexNode o = call0.operands.get(0);
+      if (o.getKind() == SqlKind.CAST) {
```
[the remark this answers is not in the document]
xuzifu666: Basic type conversion cast(var as type) seems would not be null when var is not null (such as numberic, date, string, nested type), and I would do more research about it, thanks~ @mihaibudiu

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xuzifu666: OK, I would add a new test about it. thanks @NobiGo

xuzifu666: Hi, @NobiGo @mihaibudiu I had do more research about it in these days.

Currently whether there is this pr or not, this sql(which would remove cast from string):
```
select "employee_id" from "foodmart"."employee" where 10 = cast('10' as int)
```
would convert to 
```
SELECT "employee_id" FROM "foodmart"."employee" WHERE 10='10'
```
because default calcite dialect would remove cast from string which we can refer to SqlDialect##supportsImplicitTypeCoercion,
if operand is character type in cast would be string directly.
this pr is aim to remove unnecessary cast, such as：
```
select "employee_id" from "foodmart"."employee" where 10 is distinct from cast('10' as int)
```
before the pr would convert to:
```
SELECT \"employee_id\"\nFROM \"foodmart\".\"employee\"\nWHERE (10 IS NOT NULL OR CAST('10' AS INTEGER) IS NOT NULL) AND 10 = '10' IS NOT TRUE
```
we can find that all filter had remove cast from string except is null/ is not null.
this should be converted to:
```
SELECT "employee_id" FROM "foodmart"."employee" WHERE (10 IS NOT NULL OR '10' IS NOT NULL) AND 10 = '10' IS NOT TRUE
```
at the same time, if some special dialect such as postgresql, would not more cast if supportsImplicitTypeCoercion not permit type is Numberic, 
this case also added to the new test case, it would be:
```
SELECT "employee_id" WHERE (10 IS NOT NULL OR CAST('10' AS INTEGER) IS NOT NULL) AND 10 = CAST('10' AS INTEGER) IS NOT TRUE.
```

Summarize that the pr is aim to resolve remove cast from string not thorough in IS NULL/IS NOT NULL condition.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -8268,6 +8268,31 @@ private void checkLiteral2(String expression, String expected) {
     sql(query2).ok(expected2);
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6982">[CALCITE-6982]
```
mihaibudiu, line 8272: does not match new jira title

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
```diff
@@ -350,6 +350,16 @@ public static SqlNode convertConditionToSqlNode(RexNode node,
    */
   private static RexNode stripCastFromString(RexNode node, SqlDialect dialect) {
     switch (node.getKind()) {
+    case IS_NULL:
```
mihaibudiu, line 353: the problem is that an expression such as 10 = '10' can be interpreted in two ways:
`CAST(10 AS VARCHAR) = '10'` or `10 = CAST('10' AS INT)`
To strip the cast we have to know what the target dialect would do.
These will NOT produce the same result for all expressions.
In particular, CAST to INT will produce a runtime exception if the value is not a number, while CAST to VARCHAR will never produce a runtime exception.

xuzifu666: Yes, I agree with it, but some dialect had restricted conversion conditions, such as Presto which not allowed varchar type convert to numberic type in supportsImplicitTypeCoercion method(maybe is aim to resolve the problem you mentioned), 
so do you suggest to do more cast handle for the problem？ @mihaibudiu

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xuzifu666: Had fixed it，thanks for the reminder.

core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java
mihaibudiu: Are you saying that in Presto there is no explicit cast from VARCHAR to a numeric type? 

I think that only the dialect can decide which cast to strip.
I don't think this logic can be in SqlImplementor.

xuzifu666: Yes, Presto had restricted cast the type from string to numberic, and this logic should not do this, maybe I need a more deepthinking about it, thanks for guidance! @mihaibudiu

xuzifu666:
This PR maybe unnessary, close it first.

PR #4333, opened 2025-04-24
[CALCITE-6976] Assertion failure while typechecking SUBSTR(TIMESTAMP,…)

mihaibudiu:
It looks like the family type checker for optional arguments is broken, so I have filed https://issues.apache.org/jira/browse/CALCITE-6984
This is a workaround for this particular narrower issue

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
+++ b/core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
@@ -1077,9 +1077,8 @@ public abstract class OperandTypes {
           SqlTypeFamily.INTEGER);
 
   public static final SqlSingleOperandTypeChecker STRING_INTEGER_OPTIONAL_INTEGER =
-      family(
-          ImmutableList.of(SqlTypeFamily.STRING, SqlTypeFamily.INTEGER,
-              SqlTypeFamily.INTEGER), i -> i == 2);
+      family(SqlTypeFamily.STRING, SqlTypeFamily.INTEGER)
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/sql/type/OperandTypes.java
```diff
@@ -1077,9 +1077,8 @@ public static SqlSingleOperandTypeChecker same(int operandCount,
           SqlTypeFamily.INTEGER);
 
   public static final SqlSingleOperandTypeChecker STRING_INTEGER_OPTIONAL_INTEGER =
-      family(
-          ImmutableList.of(SqlTypeFamily.STRING, SqlTypeFamily.INTEGER,
```
NobiGo, line 1081: These two should have the same meaning or aim to achieve the same effect. Should we fix the first API then?

mihaibudiu: they should, but I don't know how to fix the API, because I don't really understand the rules.
I commented about this in the JIRA ticket
I tried to mark this API Deprecated, but there is still one use I could not remove.

NobiGo: I noticed that another PR has made the same changes. If this API is incorrect, I suggest deprecating this API.

mihaibudiu: I need to file another issue; this API is still used by `JSON_QUERY`, which I think is actually not type checked at all because of this.

PR #4334, opened 2025-04-24
[CALCITE-6984] FamilyOperandTypeChecker with a Predicate describing

mihaibudiu:
optional arguments does not reject mistyped expressions

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlJsonQueryFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlJsonQueryFunction.java
@@ -30,8 +30,8 @@ import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlOperatorBinding;
 import org.apache.calcite.sql.SqlWriter;
 import org.apache.calcite.sql.parser.SqlParserPos;
-import org.apache.calcite.sql.type.OperandTypes;
 import org.apache.calcite.sql.type.ReturnTypes;
+import org.apache.calcite.sql.type.SqlSingleOperandTypeChecker;
 import org.apache.calcite.sql.type.SqlTypeFamily;
 import org.apache.calcite.sql.type.SqlTypeName;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/calcite/sql/fun/SqlJsonQueryFunction.java
```diff
@@ -45,12 +45,24 @@
...
 
+import static org.apache.calcite.sql.type.OperandTypes.family;
+
 import static java.util.Objects.requireNonNull;
 
 /**
  * The <code>JSON_QUERY</code> function.
  */
 public class SqlJsonQueryFunction extends SqlFunction {
+  static final SqlSingleOperandTypeChecker TYPE_CHECKER =
```
xiedeyantu, line 56: I have a small question, can it be written like this here?
```
  public static final SqlSingleOperandTypeChecker TYPE_CHECKER =
      family(SqlTypeFamily.ANY, SqlTypeFamily.CHARACTER,
          SqlTypeFamily.ANY, SqlTypeFamily.ANY, SqlTypeFamily.ANY)
          .or(family(SqlTypeFamily.ANY));
```

mihaibudiu: I think that the form you propose means "either 5 arguments, or just 1".
ANY means "the argument may have any type".
But what we need is "5 or 6".

xiedeyantu: Thanks! Got it.

asolimando (approved):
LGTM

xiedeyantu:
LGTM

mihaibudiu:
@xiedeyantu you can approve even if you are not a committer.

xiedeyantu (approved):

xiedeyantu:
> @xiedeyantu you can approve even if you are not a committer.

Thanks for your guidance. I always thought that it was necessary for the commtter to execute approve.

PR #4335, opened 2025-04-25
[CALCITE-6985] Add rule to transform MIN/MAX with ORDER BY and LIMIT 1

xiedeyantu:
Please see: [CALCITE-6985](https://issues.apache.org/jira/browse/CALCITE-6985)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/AggregateMinMaxToLimitRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/AggregateMinMaxToLimitRule.java
@@ -1,0 +1,128 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Aggregate;
+import org.apache.calcite.rel.core.AggregateCall;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.tools.RelBuilder;
+
+import org.immutables.value.Value;
+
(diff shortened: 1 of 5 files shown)


mihaibudiu:
can you please fix the conflict?

xiedeyantu:
> can you please fix the conflict?

Thanks, done!

core/src/main/java/org/apache/calcite/rel/rules/AggregateMinMaxToLimitRule.java
```diff
@@ -0,0 +1,124 @@
...
+  @Override public void onMatch(RelOptRuleCall call) {
+    final Aggregate agg = call.rel(0);
+    RelNode aggInput = agg.getInput().stripped();
+    RelBuilder builder = call.builder();
+
+    builder.push(aggInput);
+    List<RexNode> newProjects = new ArrayList<>();
+    for (AggregateCall aggCall : agg.getAggCallList()) {
+      int idx = aggCall.getArgList().get(0);
+      final RexNode field = builder.field(idx);
```
mihaibudiu, line 87: you may want to check that this is a deterministic expression

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -10532,4 +10532,10 @@ private void checkLoptOptimizeJoinRule(LoptOptimizeJoinRule rule) {
         .withRule(CoreRules.JOIN_CONDITION_PUSH, CoreRules.FILTER_INTO_JOIN)
         .check();
   }
+
+  @Test void testAggregateMinMaxToLimitRule() {
```
mihaibudiu, line 10536: we should take advantage of the new Quidem tests with optimization rules to check this

xiedeyantu: There are some problems when executing Quidem test, and the error is as follows:
```
> Caused by: java.util.NoSuchElementException
> 	at java.util.AbstractList$Itr.next(AbstractList.java:364)
> 	at org.apache.calcite.rel.core.Sort.explainTerms(Sort.java:267)
> 	at org.apache.calcite.rel.AbstractRelNode.getDigestItems(AbstractRelNode.java:414)
> 	at org.apache.calcite.rel.AbstractRelNode.deepHashCode(AbstractRelNode.java:397)
```
Maybe it has something to do with this
```
  public RexNode desc(RexNode node) {
    return call(SqlStdOperatorTable.DESC, node);
  }
```
I'll try to solve this problem first.

xiedeyantu: Sorry, ignore the error above. It's due to some other reasons, not related to this PR. I'll check what the problem is later. The test has been added, please review it again, thanks!

core/src/main/java/org/apache/calcite/rel/rules/AggregateMinMaxToLimitRule.java
mihaibudiu: I still think this is required to approve this PR.
If the expression is not deterministic, cloning it will produce a different result.

xiedeyantu: Sorry I missed it.

mihaibudiu (approved):
this is good, minus the style violations

xiedeyantu:
> this is good, minus the style violations

Sorry, fixed.

mihaibudiu:
Can you please squash the commits?

xiedeyantu:
> Can you please squash the commits?

Done!

PR #4336, opened 2025-04-26
[CALCITE-6986] Parser rejects SQL sources that produce an empty statement list

mihaibudiu:


testkit/src/main/java/org/apache/calcite/sql/test/SqlTestFactory.java
```diff
@@ -130,6 +130,12 @@ protected SqlTestFactory(CatalogReaderFactory catalogReaderFactory,
 
   /** Creates a parser. */
   public SqlParser createParser(String sql) {
+    if (sql.isEmpty()) {
+      // I could not figure out how to convince the grammar to accept an empty
```
xiedeyantu, line 134: I'm not very familiar with this place, maybe it's a JavaCC issue as you said, but it does feel strange.

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
```diff
@@ -2111,6 +2111,14 @@ void checkPeriodPredicate(Checker checker) {
             + "FROM `DEPT`))) AND (3 = 4))");
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-6986">[CALCITE-6986]
+   * Parser rejects SQL sources that produce an empty statement list</a>. */
+  @Test public void testEmpty() {
+    sql("").list().ok();
```
NobiGo, line 2117: I suggest adding new test cases to form a contrast. For example: 
```
sql("").ok()
sql(" ").ok()
sql("-- comment").ok()
```
Throwing an exception is also a valid test case.

NobiGo (approved):

testkit/src/main/java/org/apache/calcite/sql/parser/SqlParserTest.java
mihaibudiu: These are actually not very interesting test cases, because `sql(...).ok()` expects exactly ONE SQL statement, and there there isn't any in these strings. So these test cases are not testing this issue. @julianhyde made a comment about this in the Jira.

PR #4338, opened 2025-04-27
[CALCITE-6983] SortJoinTransposeRule should not push SORT past a UNION when SORT's fetch is DynamicParam

juntaozhang:
Detail bug report in [CALCITE-6983](https://issues.apache.org/jira/browse/CALCITE-6983)
Fix if already sort already pushed down, then skip.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/SortJoinTransposeRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SortJoinTransposeRule.java
@@ -18,6 +18,7 @@ package org.apache.calcite.rel.rules;
 
 import org.apache.calcite.plan.RelOptRuleCall;
 import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.plan.hep.HepRelVertex;
 import org.apache.calcite.rel.RelCollation;
 import org.apache.calcite.rel.RelCollations;
 import org.apache.calcite.rel.RelFieldCollation;
@@ -30,6 +31,7 @@ import org.apache.calcite.rel.logical.LogicalJoin;
 import org.apache.calcite.rel.logical.LogicalSort;
 import org.apache.calcite.rel.metadata.RelMdUtil;
 import org.apache.calcite.rel.metadata.RelMetadataQuery;
+import org.apache.calcite.rex.RexDynamicParam;
 import org.apache.calcite.tools.RelBuilderFactory;
 
 import org.immutables.value.Value;
@@ -101,6 +103,9 @@ public class SortJoinTransposeRule
               mq, join.getRight(), joinInfo.rightSet())) {
         return false;
       }
+      if (alreadyPushedThrough(join.getLeft())) {
+        return false;
+      }
     } else if (join.getJoinType() == JoinRelType.RIGHT) {
       if (sort.getCollation() != RelCollations.EMPTY) {
         for (RelFieldCollation relFieldCollation
@@ -116,11 +121,24 @@ public class SortJoinTransposeRule
               mq, join.getLeft(), joinInfo.leftSet())) {
         return false;
       }
+      if (alreadyPushedThrough(join.getRight())) {
+        return false;
+      }
     } else {
       return false;
     }
 
     return true;
+  }
+
+  // Returns true if the sort has already been pushed through
+  private boolean alreadyPushedThrough(RelNode rel) {
+    if (!(rel instanceof HepRelVertex)) {
+      return false;
+    }
+    RelNode currentRel = ((HepRelVertex) rel).getCurrentRel();
+    return currentRel instanceof Sort
+        && ((Sort) currentRel).fetch instanceof RexDynamicParam;
   }
 
   @Override public void onMatch(RelOptRuleCall call) {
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -7962,6 +7963,50 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
         .checkUnchanged();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6983">[CALCITE-6983]
+   * SortJoinTransposeRule should not push SORT past a UNION when SORT's fetch is DynamicParam
+   </a>. */
+  @Test void testSortJoinTranspose8() {
```
xiedeyantu, line 7970: This case seems to have nothing to do with PR.

core/src/main/java/org/apache/calcite/rel/rules/SortJoinTransposeRule.java
```diff
@@ -116,13 +121,26 @@ public SortJoinTransposeRule(Class<? extends Sort> sortClass,
...
+      }
     } else {
       return false;
     }
 
     return true;
   }
 
+  // Returns true if the sort has already been pushed down
+  private boolean alreadyPushedDown(RelNode rel) {
```
xiedeyantu, line 135: Thank you for submitting the PR. I have two small questions. The first is whether the `sort` containing the `RexDynamicParam` parameter should be pushed down? The second is if it can be pushed down, then in this method you only check whether the `sort` contains `RexDynamicParam`, but can you guarantee that this `sort` is the same as the top-level `sort`?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -7962,6 +7963,50 @@ private void checkSemiJoinRuleOnAntiJoin(RelOptRule rule) {
...
+        + "  SELECT empno as empno1 from emp where job = ?\n"
+        + " ) AS x\n"
+        + " left join (\n"
+        + "  SELECT empno as empno2 from emp where mgr = ?\n"
+        + " ) AS y on x.empno1 = y.empno2\n"
+        + " FETCH NEXT ? ROWS ONLY";
+    sql(sql).withPlanner(hepPlanner).check();
+  }
+
+  @Test void testSortJoinTranspose9() {
```
xiedeyantu, line 7992: There is a missing annotation like "/** Test case for ...".

core/src/main/java/org/apache/calcite/rel/rules/SortJoinTransposeRule.java
juntaozhang: I'm very glad that you can review this PR. Your two questions seem to be answered in `testSortJoinTranspose6`[CALCITE-1507](https://issues.apache.org/jira/browse/CALCITE-1507), it explains when pushing down has no impact. This PR fix if the pushdown succeeds, under certain rules, it could enter a loop of transpositions, leading to a stack overflow.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
juntaozhang: I simplified the user's SQL to better illustrate the issue, which can indeed be confusing. Taking your suggestion, I have reverted to using the original SQL.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
juntaozhang: Done, thanks for your reminding.

core/src/main/java/org/apache/calcite/rel/rules/SortJoinTransposeRule.java
juntaozhang: Not sure if I understood it correctly, feedback is welcome.

NobiGo (changes requested):
Check out the comments on Jira.

core/src/main/java/org/apache/calcite/rel/rules/SortJoinTransposeRule.java
```diff
@@ -116,13 +121,26 @@ public SortJoinTransposeRule(Class<? extends Sort> sortClass,
...
+        return false;
+      }
     } else {
       return false;
     }
 
     return true;
   }
 
+  // Returns true if the sort has already been pushed through
```
suibianwanwank, line 134: I haven't studied the original logic of this rule in detail, but frankly speaking, such approaches usually can't fundamentally prevent StackOverflow. Assuming there's a rule called SortFilterTranspose, the Sort node would be pushed under Filter again after application, creating an infinite loop of applying this same rule.

xiedeyantu: +1

juntaozhang:
hi @suibianwanwank @xiedeyantu, thanks for your reply.
This PR may not align well with the intent of the Jira issue. Additionally, as @NobiGo pointed out in their Jira comment, this does not conform to the usage specification of preparedStatement. Therefore, I am submitting another [PR](https://github.com/apache/calcite/pull/4344/files) that adheres to the statement provided in the Jira.

Thank you all.

xiedeyantu:
You can continue to modify this PR without closing it.

PR #4339, opened 2025-04-27
Krooswu/new operators

Dwrite:


NobiGo:
@Dwrite please ensure that the Commit info in the PR is consistent with the Jira summary. If the PR has implemented "<<", please add relevant descriptions in Jira.

Dwrite:
@NobiGo hi, thanks for your time. already add relevant descriptions to Jira. I have currently implemented support for the ^ (bitwise XOR) and << (bitwise left shift) operators. However, due to the testing framework using {^} as a parser position. I have only added a parser test for it, and I need find new test for it. do you have some suggestions ?

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3226,6 +3226,21 @@ public static ByteString bitNot(ByteString b) {
...
+
+  public static long leftShift(long a, int b) {
+    return a << b;
+  }
+
+  public static long leftShift(int a, long b) {
+    return a << b;
+  }
+
+  public static long leftShift(Long a, Long b) {
```
mihaibudiu, line 3241: will this cover all supported types, like TINYINT?
What is the semantics of shift when the shift amount is negative or larger than the value size?

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -1299,6 +1299,33 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
...
+   * <code>{@code <<}</code> operator.
+   */
+  public static final SqlBinaryOperator LEFTSHIFT_OPERATOR =
+      new SqlBinaryOperator("<<",                           // Operator name
+          SqlKind.OTHER_FUNCTION,             // SqlKind
+          32,
+          true,
+          ReturnTypes.LARGEST_INT_OR_FIRST_NON_NULL,
+          null,
+          OperandTypes.family(SqlTypeFamily.INTEGER, SqlTypeFamily.INTEGER));
```
mihaibudiu, line 1326: are you sure that any integer type is fine? This creates a very large number of configurations. I would expect that at least the shift amount does not support all integer types.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -16294,6 +16299,51 @@ private static void checkLogicalOrFunc(SqlOperatorFixture f) {
...
+    assertEquals(2, call.getOperandList().size());
+  }
+
+  @Test void testLeftShiftScalarFunc() {
+    final SqlOperatorFixture f = fixture();
+    f.setFor(SqlStdOperatorTable.LEFTSHIFT_OPERATOR, VmName.EXPAND);
+
+
+    // Basic test cases
+    f.checkScalar("2 << 2", "8", "INTEGER NOT NULL");
```
mihaibudiu, line 16322: can you test with values out of range 
- overflows caused by <<
- shift amount too large
- shift amount negative

core/src/main/codegen/templates/Parser.jj
```diff
@@ -7920,6 +7920,7 @@ SqlBinaryOperator BinaryRowOperator() :
 {
     // <IN> is handled as a special case
     <EQ> { return SqlStdOperatorTable.EQUALS; }
+|   <LEFTSHIFT> { return SqlStdOperatorTable.LEFTSHIFT_OPERATOR; }
```
xuzifu666, line 7923: Just curious，should we also need to implement RIGHTSHIFT？

Dwrite: yeah. I am trying to add it, But I need to resolve this parsing conflict with `cast(a as map<varchar multiset, map<int, int>>)`

Dwrite: still can not fix this conflict. wii add RIGHTSHIFT function to support.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
Dwrite: Just add more test cases. Since Java provides a standard way to handle this overloading, should we consider reusing that approach directly in this implementation?

NobiGo:
> @NobiGo hi, thanks for your time. already add relevant descriptions to Jira. I have currently implemented support for the ^ (bitwise XOR) and << (bitwise left shift) operators. However, due to the testing framework using {^} as a parser position. I have only added a parser test for it, and I need find new test for it. do you have some suggestions ?

I haven't carefully read the implementation logic of `^` in the test cases. I wonder if there are escape characters to distinguish it?

mihaibudiu:
The code to parse the string and carets is hand-written, the right solution is indeed to adapt it to treat a double caret as an escaped caret. (or some other similar escaping mechanism.)

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
Dwrite: Just add more test cases. Since Java provides a standard way to handle this overloading, should we consider reusing that approach directly in this implementation?

Dwrite:
@NobiGo Is it OK to add some ^ test cases using a whitelist?

mihaibudiu:

Dwrite:
@mihaibudiu 
Thanks for checking in! The current status is that the PR adds support for the ^, <<, and & operators, as well as the RightShift function. I've added complete test cases for all of them except for ^, which is currently tested using a whitelist-based approach, since writing exhaustive tests for it is relatively complex.

Could you help review whether this approach for ^ is acceptable?

mihaibudiu:
The PR title does not match the issue; please edit the issue.
In general, I think it's preferable to solve one problem in one PR, so probably this one should be 3 separate PRs (and 3 issues).
But I will review it as it is.

Dwrite:
@mihaibudiu  
Thanks for the suggestion. I've split this PR as requested.
- This PR now only supports `^` as per CALCITE-6731

core/src/main/java/org/apache/calcite/sql/parser/StringAndPos.java
```diff
@@ -28,6 +31,39 @@ public class StringAndPos {
   public final String sql;
   public final int cursor;
   public final @Nullable SqlParserPos pos;
+  private static final Set<String> BITXOR_OPERATOR_WHITELISTED_SQLS;
+
+  static {
```
mihaibudiu: are these test cases?
they do not belong to this class; they should in a test package

core/src/main/java/org/apache/calcite/sql/parser/StringAndPos.java
```diff
@@ -65,6 +101,10 @@ private StringAndPos(String sql, int cursor, @Nullable SqlParserPos pos) {
    * </ul>
    */
   public static StringAndPos of(String sql) {
+
+    if (BITXOR_OPERATOR_WHITELISTED_SQLS.contains(sql)) {
```
mihaibudiu: the right way to do this is to introduce an escape mechanism, similar to escaping quotes in strings
For example, parse ^^ into a ^ in StringAndPos.
If you do that, please add test cases for the StringAndPos parser.

mihaibudiu (changes requested):

core/src/main/java/org/apache/calcite/sql/parser/StringAndPos.java
Dwrite: Thanks for the suggestion!

I've updated the implementation to support an escape mechanism in StringAndPos, where ^^ is parsed into a single ^, similar to quote escaping.
I also added corresponding test cases in StringAndPosTest to verify the behavior.

Let me know if you have any further comments — would appreciate another round of review when you have a chance.

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3226,7 +3226,6 @@ public static ByteString bitNot(ByteString b) {
 
     return new ByteString(result);
   }
-
```
mihaibudiu, line 3419: this edit is not particularly useful

core/src/main/java/org/apache/calcite/sql/parser/StringAndPos.java
```diff
@@ -52,8 +53,54 @@ private StringAndPos(String sql, int cursor, @Nullable SqlParserPos pos) {
   }
 
   /**
-   * Looks for one or two carets in a SQL string, and if present, converts
-   * them into a parser position.
+   * Checks if the SQL expression is a simple XOR pattern that can be safely processed.
+   *
+   * <p>This method provides special handling for the BITXOR_OPERATOR (^) in Apache Calcite
```
mihaibudiu: I think these patterns are overly specific and not really that useful just for allowing people to skip one caret in a test case.
If you really want special processing, you could treat a string with a single caret specially, by interpreting the caret as a normal character.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -16294,6 +16299,48 @@ private static void checkLogicalOrFunc(SqlOperatorFixture f) {
...
+    // Basic XOR between two integer literals
+    f.checkScalar("2 ^ 3", "1", "INTEGER NOT NULL");
+    f.checkScalar("CAST(2 AS INTEGER) ^ CAST(3 AS BIGINT)", "1", "BIGINT NOT NULL");
+    f.checkScalar("-5 ^ 7", "-4", "INTEGER NOT NULL");
+    f.checkScalar("-5 ^ -7", "2", "INTEGER NOT NULL");
+    f.checkScalar("CAST(-5 AS TINYINT) ^ CAST(7 AS TINYINT)", "-4", "TINYINT NOT NULL");
+    f.checkScalar("CAST(-5 AS TINYINT) ^ CAST(-31 AS TINYINT)", "26", "TINYINT NOT NULL");
+    f.checkType("CAST(2 AS TINYINT) ^ CAST(6 AS TINYINT)", "TINYINT NOT NULL");
+    f.checkType("CAST(2 AS SMALLINT) ^ CAST(6 AS SMALLINT)", "SMALLINT NOT NULL");
+    f.checkType("CAST(2 AS BIGINT) ^ CAST(6 AS BIGINT)", "BIGINT NOT NULL");
```
mihaibudiu: we now support unsigned numbers, can you add one test with one of these?

Dwrite: Added

core/src/main/java/org/apache/calcite/sql/parser/StringAndPos.java
Dwrite: Thanks for the suggestion!

However, simply treating a string with a single caret (^) as a literal character won't work in practice — it causes some of our test cases to fail, particularly those involving parsing or validation errors where caret-based range highlighting is required.

That’s why we use patterns like ^...^ or more structured escape sequences — they provide clarity and work consistently across different test scenarios.

Happy to adjust the pattern matching logic if you have a concrete suggestion that avoids this ambiguity, but we do need some kind of disambiguation.

mihaibudiu: how bad do the tests look if you require escaping in all these cases?
that may be preferable

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
```diff
@@ -3417,6 +3417,66 @@ public static ByteString bitNot(ByteString b) {
...
+  }
+
+  /**
+   * Bitwise function <code>BITXOR</code> applied to {@link org.joou.ULong} values.
+   * Returns {@code null} if any operand is null.
+   */
+  public static @PolyNull ULong bitXor(@PolyNull ULong b0, @PolyNull ULong b1) {
+    return (b0 == null || b1 == null)
+        ? castNonNull(null)
+        : ULong.valueOf(b0.longValue() ^ b1.longValue());
```
mihaibudiu: I hope this handles values that overflow correctly.
Is there a test for this case (Unsigned value > INT_MAX)

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
```diff
@@ -1298,6 +1298,19 @@ public class SqlStdOperatorTable extends ReflectiveSqlOperatorTable {
...
+  /**
+   * <code>{@code ^}</code> operator.
+   */
+  public static final SqlBinaryOperator BITXOR_OPERATOR =
+      new SqlBinaryOperator(
+          "^",
+          SqlKind.BITXOR,
+          40,        // Precedence between addition (40) and multiplication (60)
+          true,
+          ReturnTypes.LARGEST_INT_OR_FIRST_NON_NULL,  // Returns same type as inputs when nullable
```
mihaibudiu: I am not sure what FIRST_NON_NULL means, and the comment makes it even more ambiguous.
- can the inputs have different types? Type coercion should make sure that does not happen, but it's not always on 
- for different types I expect that the rules for arithmetic could apply here too

core/src/main/java/org/apache/calcite/sql/parser/StringAndPos.java
mihaibudiu: Let me explain why I think these patterns are not a good idea: they essentially define a weird language for writing StringAndPos strings, whose rules cannot be summarized concisely for users. I'd rather have a simple precise language less convenient for users than a weird one.

testkit/src/test/java/org/apache/calcite/test/StringAndPosTest.java
```diff
@@ -0,0 +1,66 @@
...
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.CoreMatchers.nullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasToString;
+
+/**
+ * Unit tests for {@link StringAndPos}, particularly around caret handling.
+ */
+public class StringAndPosTest {
```
mihaibudiu: if you don't add these special parsing rules, these tests are no longer necessary either.

mihaibudiu (commented):
Could you also add a few tests in a quidem file (suffix .iq)?
The reason I am asking this is because all the tests you wrote involve only constants, so they will be optimized by the compiler.
For quidem tests that read from a table the evaluation path is slightly different.

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
Dwrite: Thanks for raising this — great point.

You're right that FIRST_NON_NULL can be a bit ambiguous. My understanding is that LARGEST_INT_OR_FIRST_NON_NULL behaves like:

If both inputs are integer types (including unsigned ones), it returns the "largest" type that can hold both.

If either is NULL, and inference is based on that input alone, it falls back to the type of the first non-null argument.

It's a safeguard when type coercion is disabled or incomplete.

In theory, with coercion enabled, inputs should be normalized to the same type (e.g., INTEGER ^ BIGINT becomes BIGINT ^ BIGINT). But when coercion is off, this return type inference helps prevent type mismatch errors and makes the behavior consistent for combinations like UByte ^ Integer, which we've added support for.

Also, I’ve updated the comment to make this behavior clearer in the code.

That said, I agree this return type logic may benefit from being aligned more closely with arithmetic rules. If there’s a more precise or canonical way to represent this (e.g., a custom inference rule or a cleaned-up version of LARGEST_INT_OR_FIRST_NON_NULL), I’d be happy to experiment with it.

Let me know if you’d prefer I try that here or in a follow-up.

testkit/src/test/java/org/apache/calcite/test/StringAndPosTest.java
Dwrite: deleted

core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java
Dwrite: Added.

core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java
Dwrite: I've updated the comment to clarify the meaning of FIRST_NON_NULL in this context — it means that the result type will be derived from the first non-null operand, which aligns with the behavior in SqlFunction BITXOR.

Regarding mixed types: yes, type coercion is typically expected to ensure operands are of compatible types. However, since coercion can be disabled in some configurations, this rule falls back to arithmetic-compatible behavior if the operand types differ.

Let me know if you'd prefer a different strategy for type derivation here — happy to adjust.

core/src/main/java/org/apache/calcite/sql/parser/StringAndPos.java
Dwrite: I totally understand the concern about adding too many special patterns potentially making the test input language confusing.

However, after careful consideration and multiple attempts, this approach with these specific patterns is currently the simplest way to guarantee idempotency of the of(String sql) method and keep all existing test cases passing.

Simply treating a single caret ^ as a normal character unfortunately breaks certain tests that rely on caret-based highlighting for error ranges.

If you have any alternative approach that can keep idempotency and pass all tests without introducing these patterns, I’d be very glad to explore it together!

Dwrite:
> Could you also add a few tests in a quidem file (suffix .iq)? The reason I am asking this is because all the tests you wrote involve only constants, so they will be optimized by the compiler. For quidem tests that read from a table the evaluation path is slightly different.

which file are you refering to ? I could not find the suffix.iq file.

mihaibudiu:
```
$ find . -name *.iq
./core/src/test/resources/sql/measure-paper.iq
./core/src/test/resources/sql/unnest.iq
./core/src/test/resources/sql/sequence.iq
./core/src/test/resources/sql/planner.iq
./core/src/test/resources/sql/unsigned.iq
./core/src/test/resources/sql/winagg.iq
./core/src/test/resources/sql/agg.iq
...
```

core/src/main/java/org/apache/calcite/sql/parser/StringAndPos.java
mihaibudiu: I am not sure why idempotency is important.
If I understand right, the function already supports escaping carets
What's so bad in escaping all carets in the tests for XOR?

mihaibudiu (commented):
I will wait for an iq test to finish reviewing.

Dwrite:
Thank you for the feedback and for reviewing the PR!

I've added the corresponding .iq test case to demonstrate the issue and verify the fix.

To clarify: the reason idempotency is broken is not due to ^^ sequences — those are already correctly handled by the parser. The issue lies in this specific logic:
```
else if (secondCaret < 0) {
  String sqlSansCaret =
      sql.substring(0, firstCaret)
          + sql.substring(firstCaret + 1);
  ...
}
```
This block assumes that a single caret ^ always indicates an error marker, and thus removes it unconditionally. However, in valid SQL expressions such as SELECT 2 ^ 3, the caret is a legitimate binary operator (bitwise XOR). As a result, when running QuidemTest, the caret is stripped out, breaking the SQL statement and causing the test to fail on subsequent runs — violating idempotency.

The proposed fix adds a minimal check to distinguish between ^ used as part of an error marker and ^ used as a valid SQL character. This ensures tests like SELECT 2 ^ 3 remain valid and idempotent across runs.

While escaping all carets manually (using ^^) would work, it introduces unnecessary complexity and reduces readability — especially for contributors unfamiliar with this nuance. This fix aims for a cleaner and more robust solution that maintains natural SQL syntax in tests.

Thanks again for your time and review — much appreciated!

core/src/test/resources/sql/operator.iq
```diff
@@ -702,4 +702,54 @@ FOO BOOLEAN(1) NOT NULL
 
 !ok
 
+SELECT 2 ^ 3 AS foo;
```
mihaibudiu, line 735: these tests are all tests with constants, please add some non-constant tests, e.g., using empno/empid.
Also, please mix operand types.
That's the whole point of iq test - they can run on table data. Better coverage than operator tests.

mihaibudiu:
I think that doubling a caret is something much simpler for contributors to do to write tests than to understand the rules about when a single caret works and when a double one is required.

core/src/test/resources/sql/operator.iq
Dwrite: Thanks for the suggestion! I’ve added non-constant IQ tests with mixed operand types (e.g., empno ^ mgr). Hope this looks good now — happy to adjust further if needed.

Dwrite:
> I think that doubling a caret is something much simpler for contributors to do to write tests than to understand the rules about when a single caret works and when a double one is required.

I agree that doubling carets is simple in principle — and I’ve tried that approach. Unfortunately, it still breaks idempotency in some cases, especially when Quidem interprets and rewrites the input/output. That’s why I went for a solution that preserves the original SQL and ensures the tests remain stable across runs.

mihaibudiu:
I don't understand what Quidem has to do with carets.
Maybe I am missing something; can you explain what the problem is with an example?
I still don't know what is the idempotency you are mentioning. What is the operation that has to be applied twice and produce the same results?

Dwrite:
> Quidem has to do with carets.

You're right — Quidem has nothing to do with carets.
The issue I mentioned is unrelated to Quidem itself.
When using ^^ (double caret), I've seen that repeated evaluation or type checking (e.g. inside f.checkScalar) can produce inconsistent results.
This breaks idempotency in tests, since the same expression might yield different results across passes.
That's why I avoided using ^^ even though it works syntactically.

Dwrite:
Hi mihaibudiu, all checks have passed. May I ask if there's anything missing or any concerns that are blocking approval at this point? I'd be happy to address them. Thanks!

mihaibudiu:
Can you show me a test that has problems with double carets? Would be good to understand what is going on

Dwrite:
java.lang.RuntimeException: Error while parsing query: values (2 ^^ 3)

	at org.apache.calcite.sql.test.AbstractSqlTester.parseAndValidate(AbstractSqlTester.java:158)
	at org.apache.calcite.sql.test.AbstractSqlTester.check(AbstractSqlTester.java:240)
	at org.apache.calcite.sql.test.SqlTester.check(SqlTester.java:158)
	at org.apache.calcite.test.SqlOperatorFixtureImpl.lambda$checkScalar$2(SqlOperatorFixtureImpl.java:240)
	at org.apache.calcite.sql.test.AbstractSqlTester.forEachQuery(AbstractSqlTester.java:454)
	at org.apache.calcite.test.SqlOperatorFixtureImpl.checkScalar(SqlOperatorFixtureImpl.java:239)
	at org.apache.calcite.sql.test.SqlOperatorFixture.checkScalar(SqlOperatorFixture.java:232)
	at org.apache.calcite.test.SqlOperatorTest.testBitXorOperatorScalarFunc(SqlOperatorTest.java:16468)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:727)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at 
....
Caused by: org.apache.calcite.sql.parser.SqlParseException: Encountered "^ ^" at line 1, column 11.

Hi,

Just to clarify, Quidem replaces ^^ with a single caret ^ before executing the tests. So in test inputs, we need to write ^^ to represent an actual caret.

The issue is that this replacement happens multiple times during test processing, which means the caret count reduces each time the test runs. This breaks idempotency, because the SQL changes after the first run, causing failures or unexpected behavior on subsequent runs.

That’s why relying solely on doubling carets (^^) isn’t an ideal solution. We need a way to handle single carets properly to ensure consistent behavior across multiple executions.

mihaibudiu:
I think this signals a bug in the testing framework, I will take a look at this.

mihaibudiu:
The problem is the following: the testing framework uses a "consumer" to check the results, with the same API used for negative and positive test cases. There is an implicit assumption that only negative tests may have carets in the SQL strings. So when a positive test is executed, the code paths for handling escaped carets are never executed. I will see whether I can fix this.

Dwrite:
> The problem is the following: the testing framework uses a "consumer" to check the results, with the same API used for negative and positive test cases. There is an implicit assumption that only negative tests may have carets in the SQL strings. So when a positive test is executed, the code paths for handling escaped carets are never executed. I will see whether I can fix this.

Got it, thanks again for the explanation! Just to confirm — would you suggest I hold off on this PR until the testing framework is updated, or is there anything you'd prefer me to adjust in the meantime?

mihaibudiu:
Let’s wait a bit. Thank you

Dwrite:
> Let’s wait a bit. Thank you

Sure, thanks!
Just to mention — I’ve added the previously added << operator in a separate PR: https://github.com/apache/calcite/pull/4478.

mihaibudiu (commented):
I have pushed an extra commit which should fix the problems of the test framework with carets. It would be nice if someone else than me could review that. It was a bit messy to cleanup.

(I have also removed your custom caret parsing function)

Dwrite:
> I have pushed an extra commit which should fix the problems of the test framework with carets. It would be nice if someone else than me could review that. It was a bit messy to cleanup.
> 
> (I have also removed your custom caret parsing function)

sure. I just asked the `[xuzifu666](https://github.com/xuzifu666)` to review

xuzifu666:
> > I have pushed an extra commit which should fix the problems of the test framework with carets. It would be nice if someone else than me could review that. It was a bit messy to cleanup.
> > (I have also removed your custom caret parsing function)
> 
> sure. I just asked the `[xuzifu666](https://github.com/xuzifu666)` to review

I think these changes are fine.

core/src/main/java/org/apache/calcite/materialize/TileSuggester.java
```diff
@@ -60,11 +60,11 @@ public Iterable<? extends Lattice.Tile> tiles() {
     final double f = statisticsProvider.getFactRowCount();
     final ImmutableMap.Builder<Parameter, Object> map = ImmutableMap.builder();
     if (lattice.algorithmMaxMillis >= 0) {
-      map.put(Algorithm.ParameterEnum.timeLimitSeconds,
+      map.put(Algorithm.ParameterEnum.TIME_LIMIT_SECONDS,
```
mihaibudiu: why are these changes in the diff?
Can you please rebase on the latest main and force-push again?

Dwrite: sure.updated

mihaibudiu (approved):
Please squash the commits so we can merge.
If it's not too much work, you can leave my changes in a separate commit - but that may be difficult.

Dwrite:
> Please squash the commits so we can merge. If it's not too much work, you can leave my changes in a separate commit - but that may be difficult.

Thanks for the review! I've squashed my commits and kept your fixes in a separate commit as suggested.

mihaibudiu:
I think there is a linter rule which will prohibit you from having a commit message starting with "Fix".
You can try to run this locally with ./gradlew build.

mihaibudiu:
If you will change the commits, maybe you can combine the two changes from the first two commits into a single one and use that message.

Dwrite:
> If you will change the commits, maybe you can combine the two changes from the first two commits into a single one and use that message.
sure. Just squash the commits to one commit.

mihaibudiu (approved):

mihaibudiu:
Appreciate your patience with the long review process

Dwrite:
> Appreciate your patience with the long review process

Thank you! This is actually my first PR to Calcite, and I really appreciate your patience and detailed guidance throughout the long review process. I’ve learned a lot from it. Looking forward to contributing more and doing better in the future!

PR #4342, opened 2025-04-28
[CALCITE-6988] DuckDB dialect implementation

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6988

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/dialect/DuckDBSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/DuckDBSqlDialect.java
@@ -1,0 +1,162 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.sql.dialect;
+
+import org.apache.calcite.avatica.util.TimeUnitRange;
+import org.apache.calcite.config.NullCollation;
+import org.apache.calcite.rel.type.RelDataTypeSystem;
+import org.apache.calcite.rel.type.RelDataTypeSystemImpl;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.SqlLiteral;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.fun.SqlLibraryOperators;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.type.SqlTypeName;
+
+/**
+ * A <code>SqlDialect</code> implementation for the DuckDB database.
+ */
+public class DuckDBSqlDialect extends SqlDialect {
+  public static final RelDataTypeSystem TYPE_SYSTEM =
+      new RelDataTypeSystemImpl() {
+
+        // We can refer to document of DuckDB 1.2.x:
+        // https://duckdb.org/docs/stable/sql/data_types/numeric#fixed-point-decimals
+        @Override public int getMaxPrecision(SqlTypeName typeName) {
+          switch (typeName) {
+          case DECIMAL:
+            return 38;
+          default:
+            return super.getMaxPrecision(typeName);
+          }
+        }
+
+        @Override public int getMaxScale(SqlTypeName typeName) {
+          switch (typeName) {
+          case DECIMAL:
+            return 38;
+          default:
+            return super.getMaxScale(typeName);
+          }
+        }
+
+        @Override public int getMaxNumericScale() {
+          return getMaxScale(SqlTypeName.DECIMAL);
+        }
+      };
+
+  public static final SqlDialect.Context DEFAULT_CONTEXT = SqlDialect.EMPTY_CONTEXT
+      .withDatabaseProduct(SqlDialect.DatabaseProduct.DUCKDB)
+      .withIdentifierQuoteString("\"")
+      // Refer to document: https://duckdb.org/docs/stable/sql/query_syntax/orderby.html
+      .withNullCollation(NullCollation.LAST)
+      .withDataTypeSystem(TYPE_SYSTEM);
+
+  public static final SqlDialect DEFAULT = new DuckDBSqlDialect(DEFAULT_CONTEXT);
+
+  /** Creates a DuckDBSqlDialect. */
+  public DuckDBSqlDialect(SqlDialect.Context context) {
+    super(context);
+  }
+
+  @Override public boolean supportsApproxCountDistinct() {
+    return true;
+  }
+
+  @Override public boolean supportsAliasedValues() {
+    return false;
+  }
+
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/sql/dialect/DuckDBSqlDialect.java
```diff
@@ -0,0 +1,157 @@
...
+
+  public static final SqlDialect.Context DEFAULT_CONTEXT = SqlDialect.EMPTY_CONTEXT
+      .withDatabaseProduct(SqlDialect.DatabaseProduct.DUCKDB)
+      .withIdentifierQuoteString("\"")
+      .withNullCollation(NullCollation.LAST)
+      .withDataTypeSystem(TYPE_SYSTEM);
+
+  public static final SqlDialect DEFAULT = new DuckDBSqlDialect(DEFAULT_CONTEXT);
+
+  /** Creates a PhoenixSqlDialect. */
```
xiedeyantu, line 71: DuckDBSqlDialect

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7842,9 +7890,12 @@ private void checkLiteral2(String expression, String expected) {
     // since it seems to be used across more dialects.
     final String expectedSnowflake = "SELECT LENGTH(\"brand_name\")\n"
         + "FROM \"foodmart\".\"product\"";
+    final String expectedDuckDB = "SELECT LENGTH(\"brand_name\")\n"
```
xiedeyantu, line 7893: same as expectedSnowflake?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -5361,52 +5362,92 @@ private void checkLiteral2(String expression, String expected) {
 
   /** Test case for
    * <a href="https://issues.apache.org/jira/browse/CALCITE-6949">[CALCITE-6949]
-   * ClickHouse not support floor date to SECOND/MILLISECOND/MICROSECOND/NANOSECOND</a>. */
+   * ClickHouse not support floor date to SECOND/MILLISECOND/MICROSECOND/NANOSECOND</a>,
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6988">[CALCITE-6988]
+   * DuckDB dialect implementation</a>.
+   * */
   @Test void testFloorClickHouse() {
```
xiedeyantu, line 5369: Change the method name to testFloorXxx?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xuzifu666: Thanks~ had changed the code.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xuzifu666: +1, had changed the name to testFloorFunction. @xiedeyantu

xiedeyantu:
LGTM

core/src/main/java/org/apache/calcite/sql/dialect/DuckDBSqlDialect.java
```diff
@@ -0,0 +1,157 @@
...
+      format = "second";
+      break;
+    case MILLISECOND:
+      format = "milliseconds";
+      break;
+    case MICROSECOND:
+      format = "microseconds";
+      break;
+    default:
+      throw new AssertionError("MYSQL does not support FLOOR for time unit: "
```
mihaibudiu, line 144: Why MYSQL?

core/src/main/java/org/apache/calcite/sql/dialect/DuckDBSqlDialect.java
```diff
@@ -0,0 +1,157 @@
...
+
+        @Override public int getMaxNumericScale() {
+          return getMaxScale(SqlTypeName.DECIMAL);
+        }
+      };
+
+  public static final SqlDialect.Context DEFAULT_CONTEXT = SqlDialect.EMPTY_CONTEXT
+      .withDatabaseProduct(SqlDialect.DatabaseProduct.DUCKDB)
+      .withIdentifierQuoteString("\"")
+      .withNullCollation(NullCollation.LAST)
```
mihaibudiu, line 66: Is this LAST or HIGH?
These are not the same.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -5958,6 +5999,13 @@ private void checkLiteral2(String expression, String expected) {
         .ok(expected);
   }
 
+  @Test void testDuckDBDecimalPrecision() {
+    final String sql = "SELECT CAST(1.23 AS DECIMAL(38, 37))";
+    final String expected = "SELECT 1.2300000000000000000000000000000000000";
```
mihaibudiu, line 6004: I think this test would be more robust if the last digit was not 0
you are relying on the formatting to print zeros which are not significant, and this may not always be true

xuzifu666: From definition：
LAST：Nulls last for DESC and ASC.
HIGH：Nulls first for DESC, nulls last for ASC.

system env test：
before query
insert into t3 select 1,'aa';
insert into t3 select 1,'bb';
insert into t3 select 1,NULL;
insert into t3 select 1,NULL;

then query order by desc/asc
D select * from t3 order by data asc;
┌───────┬─────────┐
│  id   │  data   │
│ int32 │ varchar │
├───────┼─────────┤
│     1 │ aa      │
│     1 │ bb      │
│     1 │         │
└───────┴─────────┘
D select * from t3 order by data desc;
┌───────┬─────────┐
│  id   │  data   │
│ int32 │ varchar │
├───────┼─────────┤
│     1 │ bb      │
│     1 │ aa      │
│     1 │         │
└───────┴─────────┘

we can find that NULL is always at last, so here should be LAST.

core/src/main/java/org/apache/calcite/sql/dialect/DuckDBSqlDialect.java
xuzifu666: From definition：
LAST：Nulls last for DESC and ASC.
HIGH：Nulls first for DESC, nulls last for ASC.

system env test：
before query
insert into t3 select 1,'aa';
insert into t3 select 1,'bb';
insert into t3 select 1,NULL;
insert into t3 select 1,NULL;

then query order by desc/asc
D select * from t3 order by data asc;
┌───────┬─────────┐
│  id   │  data   │
│ int32 │ varchar │
├───────┼─────────┤
│     1 │ aa      │
│     1 │ bb      │
│     1 │         │
└───────┴─────────┘
D select * from t3 order by data desc;
┌───────┬─────────┐
│  id   │  data   │
│ int32 │ varchar │
├───────┼─────────┤
│     1 │ bb      │
│     1 │ aa      │
│     1 │         │
└───────┴─────────┘

we can find that NULL is always at last, so here should be LAST.

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
xuzifu666: Thanks for the reminder, had add a case with non-zero at last. @mihaibudiu

core/src/main/java/org/apache/calcite/sql/dialect/DuckDBSqlDialect.java
mihaibudiu: Can you add a link to https://duckdb.org/docs/stable/sql/query_syntax/orderby.html so that people can check easily?

xuzifu666: OK, had added the link to the code~ @mihaibudiu

xuzifu666:
> I think you can also squash the commits so we can merge this

Had squashed the commits, thanks @mihaibudiu

PR #4344, opened 2025-04-28
[CALCITE-6983] SortJoinTransposeRule should not push SORT past a UNION when SORT's fetch is DynamicParam

juntaozhang:
Details in [CALCITE-6983](https://issues.apache.org/jira/browse/CALCITE-6983)


core/src/main/java/org/apache/calcite/rel/rules/SortJoinTransposeRule.java
```diff
@@ -79,13 +80,18 @@ public SortJoinTransposeRule(Class<? extends Sort> sortClass,
...
+    // 1) If the fetch of sort is dynamic parameter, we bail out
+    // 2) If join is not a left or right outer, we bail out
+    // 3) If sort is not a trivial order-by, and if there is
     // any sort column that is not part of the input where the
     // sort is pushed, we bail out
-    // 3) If sort has an offset, and if the non-preserved side
+    // 4) If sort has an offset, and if the non-preserved side
     // of the join is not count-preserving against the join
     // condition, we bail out
+    if (sort.fetch instanceof RexDynamicParam) {
```
mihaibudiu, line 91: if the limit is a dynamic parameter is this ok?

mihaibudiu: in general, I wonder how many rules are ok when queries contain dynamic parameters.
but I don't know if this is the place to answer this question.

juntaozhang: > in general, I wonder how many rules are ok when queries contain dynamic parameters. but I don't know if this is the place to answer this question.

Thanks for your review. I have no idea.
FYI: [CALCITE-6647 SortUnionTransposeRule should not push SORT past a UNION when SORT's fetch is DynamicParam](https://issues.apache.org/jira/browse/CALCITE-6647) .
This [PR](https://github.com/apache/calcite/pull/4338) is guaranteed that the push througth operation not loop.

mihaibudiu: I will file a new issue.

mihaibudiu: https://issues.apache.org/jira/browse/CALCITE-6990

xiedeyantu: > > in general, I wonder how many rules are ok when queries contain dynamic parameters. but I don't know if this is the place to answer this question.
> 
> Thanks for your review. I have no idea. FYI: [CALCITE-6647 SortUnionTransposeRule should not push SORT past a UNION when SORT's fetch is DynamicParam](https://issues.apache.org/jira/browse/CALCITE-6647) . This [PR](https://github.com/apache/calcite/pull/4338) is guaranteed that the push througth operation not loop.

The `SortUnionTransposeRule` will push `sort` down to all `inputs`, and the number of sort with `DynamicParam` will increase from 1 to n. But the `SortJoinTransposeRule` will only push `sort` down to one `input`. I don't know if this scenario supports `DynamicParam`. I also wanted to ask this question in the previous PR. It may need to be discussed in Mihai's Jira.

juntaozhang: Hi @xiedeyantu 
Currently we only handle left and right join, if only have limit, sort push through preserved side didn't change sql result.
If with offset, it will more complicated, sort only push through to preserved side, and the other side need count-preserving, this will not change the sql result.

"pushdown" is incorrect, community use "push through" before, it means that the original sort still there, and the sort will be add to the corresponding place. Thus, multiple dynamic params may diff from the semantic expression required by the original SQL when in Preparestatement as @NobiGo  mentioned.
This is my understanding. Welcome any corrections。

juntaozhang: > can you add a test using `?` for OFFSET too?

Thanks for your reminding, done

mihaibudiu:
Until your PR is approved, you should try not to force-push, since it makes reviewing changes harder.

mihaibudiu (approved):

juntaozhang:
> Until your PR is approved, you should try not to force-push, since it makes reviewing changes harder.

ok

PR #3819, opened 2024-06-09
[CALCITE-6432] Infinite loop for JoinPushTransitivePredicatesRule

asdfgh19:
jira: [CALCITE-6432](https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6432)

Infinite loop for JoinPushTransitivePredicatesRule when there are multiple project expressions reference the same input field.

PR #4346, opened 2025-04-28
[CALCITE-6432] Infinite loop for JoinPushTransitivePredicatesRule

xiedeyantu:


xiedeyantu:
I found [#3819 ](https://github.com/apache/calcite/pull/3819) is a very usefull PR, but it seems to be inactive. It has a small problem and I made some changes. If the original contributor can continue to complete his PR. I can close this PR at any time.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -634,6 +641,126 @@ public RelOptPredicateList getPredicates(RelSubset r,
     return Util.first(list, RelOptPredicateList.EMPTY);
   }
 
+  /**
```
mihaibudiu, line 644: is this changed?
github makes it difficult to tell whether the code was just moved.
why move it anyway?

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -194,15 +194,11 @@ public RelOptPredicateList getPredicates(Project project,
     final List<RexNode> projectPullUpPredicates = new ArrayList<>();
 
     ImmutableBitSet.Builder columnsMappedBuilder = ImmutableBitSet.builder();
-    Mapping m =
-        Mappings.create(MappingType.PARTIAL_FUNCTION,
-            input.getRowType().getFieldCount(),
-            project.getRowType().getFieldCount());
-
+    Map<Integer, BitSet> equivalence = new HashMap<>();
```
mihaibudiu, line 197: can you add a comment describing what this map does, including the keys and values?

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -7448,27 +7448,6 @@ LogicalProject(SAL=[$5])
               LogicalProject(SAL=[$5], $f9=[=($5, 4)])
                 LogicalFilter(condition=[AND(=($7, 20), >($5, 1000))])
                   LogicalTableScan(table=[[CATALOG, SALES, EMPNULLABLES]])
-]]>
```
mihaibudiu, line 7451: where is the planAfter?

xiedeyantu: After modification, this case becomes checkUnchanged.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
xiedeyantu: Done!

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
xiedeyantu: I refactored the code to make the changes clearer.

xiedeyantu:
@mihaibudiu Please help me to take a look again if you have time. Thanks!

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -966,15 +986,18 @@ protected EquivalenceFinder() {
...
       final BitSet[] columnSets;
       final int[] iterationIdx;
       @Nullable Mapping nextMapping;
       boolean firstCall;
+      int sourceCount;
+      int targetCount;
 
       @SuppressWarnings("JdkObsolete")
-      ExprsItr(ImmutableBitSet fields) {
+      ExprsItr(ImmutableBitSet fields, Map<Integer, BitSet> equivalence,
```
mihaibudiu, line 999: is the supression of JdkObsolete still necessary?

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
```diff
@@ -218,8 +214,30 @@ public RelOptPredicateList getPredicates(Project project,
...
-        r2 = r2.accept(new RexPermuteInputsShuttle(m, input));
-        projectPullUpPredicates.add(r2);
+        ImmutableBitSet fields = RelOptUtil.InputFinder.bits(r2);
+        // If r2 cannot find input (such as SubQuery),
+        // it will directly return without adjusting mapping.
+        if (fields.isEmpty()) {
+          projectPullUpPredicates.add(r2);
+          continue;
+        }
+        // Using Map<Integer, BitSet>, the result will be
```
mihaibudiu, line 224: There is no need to describe alternative designs.
I would change this comment and add it to the declaration of the map itself, above.
I would document the map: keys are field indexes (RexInputRef) that appear in the project, values are sets of field indexes (RexInputRef) that appear in r2.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
xiedeyantu: To be honest, I don't really understand the purpose of this annotation, so I just kept it as it is.

mihaibudiu:
```
@SuppressWarnings("JdkObsolete") is an annotation in Java that instructs the compiler to ignore warnings related to the usage of obsolete APIs from the JDK, such as old collection classes like Vector, Hashtable, or Enumeration. These APIs are considered outdated and are generally discouraged in favor of newer, more efficient alternatives like ArrayList, HashMap, or the Iterator interface.
```

mihaibudiu: I don't see any such usage.
If you are using an IDE, you can remove the warning and see whether it complains.
Cleaning up code is always useful, even if it is not the goal of your PR.

xiedeyantu: Thanks! I will try it.

core/src/main/java/org/apache/calcite/rel/metadata/RelMdPredicates.java
xiedeyantu: I updated the doc and removed the annotations. I tested locally that both java8 and java17 were successfully and to see if ci can succeed.

mihaibudiu (approved):

mihaibudiu:
I think you can also squash the commits so we can merge

xiedeyantu:
> I think you can also squash the commits so we can merge

Okay, I'll go right now.

xiedeyantu:
> I think you can also squash the commits so we can merge

Done! Please wait the CI.

PR #4350, opened 2025-04-30
[CALCITE-5387] Type-mismatch on nullability in JoinPushTransitivePredicatesRule RelRule

xiedeyantu:


xiedeyantu:
This PR continues the unfinished work of #3452 .

core/src/main/java/org/apache/calcite/rex/RexPermuteInputsShuttle.java
```diff
@@ -54,14 +55,29 @@ public class RexPermuteInputsShuttle extends RexShuttle {
   public RexPermuteInputsShuttle(
       Mappings.TargetMapping mapping,
       RelNode... inputs) {
-    this(mapping, fields(inputs));
+    this(mapping, fields(inputs), false);
+  }
+
+  public RexPermuteInputsShuttle(
```
mihaibudiu, line 61: can you please document the new flag using JavaDoc?

core/src/main/java/org/apache/calcite/rex/RexPermuteInputsShuttle.java
```diff
@@ -85,9 +101,15 @@ private static ImmutableList<RelDataTypeField> fields(RelNode[] inputs) {
   @Override public RexNode visitInputRef(RexInputRef local) {
     final int index = local.getIndex();
     int target = mapping.getTarget(index);
-    return new RexInputRef(
-        target,
-        local.getType());
+    if (!isStrictTypeMatch) {
```
mihaibudiu, line 104: this doesn't look to me like it's a "strict" comparison
basically you are copying types from different fields: source, or destination
Am I right?

xiedeyantu: Yes, maybe I need to change the name `isMatchTargetType`? This affects the creation of subsequent Filters in the rule.

core/src/main/java/org/apache/calcite/rex/RexPermuteInputsShuttle.java
```diff
@@ -54,14 +55,41 @@ public class RexPermuteInputsShuttle extends RexShuttle {
   public RexPermuteInputsShuttle(
       Mappings.TargetMapping mapping,
       RelNode... inputs) {
-    this(mapping, fields(inputs));
+    this(mapping, fields(inputs), false);
+  }
+
+  /**
+   * Creates a RexPermuteInputsShuttle with target type matching control.
```
mihaibudiu, line 62: what is 'control'? It's not mentioned anywhere else.

core/src/main/java/org/apache/calcite/rex/RexPermuteInputsShuttle.java
```diff
@@ -54,14 +55,41 @@ public class RexPermuteInputsShuttle extends RexShuttle {
...
+   *
+   * <p>The mapping provides at most one target for every source. If a source
```
mihaibudiu, line 64: I think this paragraph belongs in the documentation for the mapping parameter.
I don't think you should discuss here what TargetMapping does, that should belong to the documentation of TargetMapping itself. What you should say is what this function assumes about the mapping. Perhaps you mean to say that every RexInputRef in the expression that will be analyzed by this shuttle needs to be a key in this mapping?

core/src/main/java/org/apache/calcite/rex/RexPermuteInputsShuttle.java
```diff
@@ -54,14 +55,41 @@ public class RexPermuteInputsShuttle extends RexShuttle {
...
+   * has no targets and is referenced in the expression,
+   * {@link org.apache.calcite.util.mapping.Mappings.TargetMapping#getTarget(int)}
+   * will give an error. Otherwise the mapping gives a unique target.
+   *
+   * @param mapping          Mapping that defines source-to-target relation
+   * @param shouldMatchTargetType If true, enforces use target type when remapping inputs
```
mihaibudiu, line 70: Based on this description, this should be called 'matchTargetType'. There is no need for 'should'.
I would describe what this does for both 'true' and 'false' cases.
In the 'false' case each input ref is substituted with a reference with the same type as the original input.
In the 'true' case each input ref is substituted with a reference with the same type as the target type.

mihaibudiu (approved):
Only some documentation improvements are required.
I have thus approved, but will wait for these to happen to merge.
Since this is a small PR you can probably also squash the commits after you make the documentation changes.

core/src/main/java/org/apache/calcite/rex/RexPermuteInputsShuttle.java
xiedeyantu: I modified the description of `matchTargetType` according to your suggestion, and kept the other descriptions unchanged. Do you think it's ok?

core/src/main/java/org/apache/calcite/rex/RexPermuteInputsShuttle.java
mihaibudiu: what I suggested here is that the second phrase and third should be removed. (If a source... unique target). They do not document what this method does, they document what getTarget does.

xiedeyantu: Done!

mihaibudiu:
I plan to merge this soon if there are no other comments.

PR #4351, opened 2025-04-30
[CALCITE-6995] Support FULL JOIN in StarRocks/Doris Dialect

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6995

### DIFF
--- a/core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
+++ b/core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
@@ -176,7 +176,9 @@ public final class CalciteSystemProperty<T> {
               "H2",
               "MYSQL",
               "ORACLE",
-              "POSTGRESQL"));
+              "POSTGRESQL",
+              "STARROCKS",
+              "DORIS"));
 
   /**
    * Path to the dataset file that should used for integration tests.
--- a/core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
+++ b/core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
@@ -1357,6 +1357,27 @@ class JdbcAdapterTest {
 
   /**
    * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6995">[CALCITE-6995]
+   * Support FULL JOIN in StarRocks/Doris Dialect</a>. */
+  @Test void testFullJoinSupportedDialect() {
+    CalciteAssert.model(JdbcTest.SCOTT_MODEL)
+        .enable(CalciteAssert.DB != CalciteAssert.DatabaseInstance.H2
+            && CalciteAssert.DB != CalciteAssert.DatabaseInstance.MYSQL)
+        .query("select empno, ename, e.deptno, dname\n"
+            + "from scott.emp e full join scott.dept d\n"
+            + "on e.deptno = d.deptno")
+        .explainContains("PLAN=JdbcToEnumerableConverter\n"
+            + "  JdbcProject(EMPNO=[$0], ENAME=[$1], DEPTNO=[$2], DNAME=[$4])\n"
+            + "    JdbcJoin(condition=[=($2, $3)], joinType=[full])\n"
+            + "      JdbcProject(EMPNO=[$0], ENAME=[$1], DEPTNO=[$7])\n"
+            + "        JdbcTableScan(table=[[SCOTT, EMP]])\n"
+            + "      JdbcProject(DEPTNO=[$0], DNAME=[$1])\n"
+            + "        JdbcTableScan(table=[[SCOTT, DEPT]])\n\n")
+        .runs();
+  }
+
+  /**
+   * Test case for
    * <a href="https://issues.apache.org/jira/browse/CALCITE-5243">[CALCITE-5243]
    * "SELECT NULL AS C causes NoSuchMethodException: java.sql.ResultSet.getVoid(int)</a>. */
   @Test void testNullSelect() {
--- a/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
+++ b/core/src/main/java/org/apache/calcite/sql/dialect/StarRocksSqlDialect.java
@@ -18,6 +18,7 @@ package org.apache.calcite.sql.dialect;
 
 import org.apache.calcite.avatica.util.TimeUnitRange;
 import org.apache.calcite.config.NullCollation;
+import org.apache.calcite.rel.core.JoinRelType;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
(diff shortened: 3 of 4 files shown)


core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -1335,7 +1335,9 @@ private LockWrapper exclusiveCleanDb(Connection c) throws SQLException {
   /**
    * Test case for
    * <a href="https://issues.apache.org/jira/browse/CALCITE-4619">[CALCITE-4619]
-   * "Full join" generates an incorrect execution plan under mysql</a>. */
```
asolimando, line 1338: I feel a separate test case would be cleaner, `testFullJoinSupportedDialect` or similar for those who support it, the existing (untouched?) for those who don't

asolimando (changes requested):

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
xuzifu666: Thanks for the suggestion and agree with it，had add a new test case for the issue.

asolimando:
@xuzifu666 can you check sonar report and fix what makes sense? Thanks!

asolimando:
> > @xuzifu666 can you check sonar report and fix what makes sense? Thanks!
> 
> In my view，sonar test is success，and other test is ok，which sonar error do you mean？（can show the link to me）@asolimando

It's [this link](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=4351&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true) that gets published as a recap in messages like [this one](https://github.com/apache/calcite/pull/4351#issuecomment-2841049798).

In this case there are 6 new issues, some are clearly addressable and correct (dangling imports), some others could be addressed (repetition of string). There might also be false positives sometimes, or debatable issues like "brain methods" etc. which are very much present in the codebase and mostly accepted.

In general it's worth checking with grain of salt, not every issue reported has to be addressed, but for trivial things like imports it spares time for everyone and makes the review total time shorter.

When you see "success" it means the analysis succeeded, not that there aren't problems.

core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
```diff
@@ -214,6 +214,10 @@ public final class CalciteSystemProperty<T> {
         return ".";
       });
 
+
```
asolimando, line 217: Let's have them as `private static final` in `JdbcAdapterTest` for now, we can move them in a "common" location and adjust visibility if they are going to be used elsewhere

NobiGo:
> @xuzifu666 can you check sonar report and fix what makes sense? Thanks!

I haven't paid attention to this test before. It often runs rather slowly. I will pay more attention to it later.

core/src/main/java/org/apache/calcite/config/CalciteSystemProperty.java
xuzifu666: Maybe we cannot put them in JdbcAdapterTest because need these static String in CalciteAssert (in Calcite testkit and JdbcAdapterTest is in core moudle ), how about put them in CalciteAssert with private static final ? @asolimando

xuzifu666:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=apache_calcite&pullRequest=4351) **Quality Gate passed**
> Issues  [0 New issues](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=4351&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=apache_calcite&pullRequest=4351&issueStatuses=ACCEPTED)
> 
> Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=apache_calcite&pullRequest=4351&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [90.9% Coverage on New Code](https://sonarcloud.io/component_measures?id=apache_calcite&pullRequest=4351&metric=new_coverage&view=list)  [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=apache_calcite&pullRequest=4351&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=apache_calcite&pullRequest=4351)

According sonar report warning had been fixed. @asolimando

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -1355,6 +1355,27 @@ private LockWrapper exclusiveCleanDb(Connection c) throws SQLException {
         .runs();
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6995">[CALCITE-6995]
+   * StarRocks/Doris should support full join</a>.*/
+  @Test void testFullJoinSupportedDialect() {
+    CalciteAssert.model(JdbcTest.SCOTT_MODEL)
+        .enable(CalciteAssert.DB != CalciteAssert.DatabaseInstance.H2
```
asolimando, line 1364: Q: I see no other tests where we enable when the DB is not set to a given set of values (inequality), we always enable on specific values (equality). In this case, do we want to enable it specifically for Doris/StarRocks or for all DBs except the two you have excluded?

xuzifu666: I think here should test for all DBs except the two excluded, because other db can support full join. @asolimando

asolimando (approved):
LGTM!

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
```diff
@@ -1355,6 +1355,27 @@ private LockWrapper exclusiveCleanDb(Connection c) throws SQLException {
         .runs();
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6995">[CALCITE-6995]
+   * StarRocks/Doris should support full join</a>.*/
```
NobiGo, line 1361: We should be consistent with the others. `</a>. */`

NobiGo:
I hope we can change this pr summary to `Support FULL JOIN in StarRocks/Doris Dialect`

core/src/test/java/org/apache/calcite/test/JdbcAdapterTest.java
xuzifu666: Thanks for remidner, had fixed it.

xuzifu666:
> I hope we can change this pr summary to `Support FULL JOIN in StarRocks/Doris Dialect`

Make sense, I had changed the title to Support FULL JOIN in StarRocks/Doris Dialect, thanks~ @NobiGo

NobiGo (approved):

PR #4352, opened 2025-04-30
[CALCITE-6981] Runtime conversion of DECIMAL ARRAY to INT ARRAY fails…

NobiGo:
… with a ClassCastException

### DIFF
--- a/core/src/test/resources/sql/cast.iq
+++ b/core/src/test/resources/sql/cast.iq
@@ -1707,4 +1707,91 @@ WHERE x BETWEEN 2 AND 4
 
 !ok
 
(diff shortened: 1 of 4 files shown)


core/src/test/resources/sql/cast.iq
```diff
@@ -1707,4 +1707,16 @@ WHERE x BETWEEN 2 AND 4
 
 !ok
 
+# [CALCITE-6981] Runtime conversion of DECIMAL ARRAY to INT ARRAY fails with a ClassCastException
```
mihaibudiu, line 1710: will this work with nested arrays (INT ARRAY ARRAY)?
maybe you can add a test to check

mihaibudiu (approved):

PR #4353, opened 2025-04-30
[CALCITE-6991] Validator cannot infer type for COALESCE when call is not expanded

juntaozhang:
details info see [CALCITE-6991](https://issues.apache.org/jira/browse/CALCITE-6991)

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlCoalesceFunction.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlCoalesceFunction.java
@@ -16,20 +16,33 @@
  */
 package org.apache.calcite.sql.fun;
 
+import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlCallBinding;
 import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlFunctionCategory;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.SqlOperatorBinding;
 import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.type.OperandTypes;
 import org.apache.calcite.sql.type.ReturnTypes;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
 import org.apache.calcite.sql.type.SqlTypeTransforms;
+import org.apache.calcite.sql.type.SqlTypeUtil;
 import org.apache.calcite.sql.validate.SqlValidator;
+import org.apache.calcite.sql.validate.implicit.TypeCoercion;
 import org.apache.calcite.util.Util;
 
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+import java.util.ArrayList;
 import java.util.List;
+
+import static org.apache.calcite.util.Static.RESOURCE;
+
+import static java.util.Objects.requireNonNull;
 
 /**
  * The <code>COALESCE</code> function.
@@ -79,5 +92,43 @@ public class SqlCoalesceFunction extends SqlFunction {
     SqlNode elseExpr = Util.last(operands);
     assert call.getFunctionQuantifier() == null;
     return SqlCase.createSwitched(pos, null, whenList, thenList, elseExpr);
+  }
+
+  @Override @NonNull public RelDataType inferReturnType(
+      SqlOperatorBinding opBinding) {
+    RelDataType returnType = getReturnTypeInference().inferReturnType(opBinding);
+    if (returnType == null
+        && opBinding instanceof SqlCallBinding
+        && ((SqlCallBinding) opBinding).isTypeCoercionEnabled()) {
+      SqlCallBinding callBinding = (SqlCallBinding) opBinding;
+      List<RelDataType> argTypes = new ArrayList<>();
+      for (SqlNode operand : callBinding.operands()) {
+        RelDataType type = SqlTypeUtil.deriveType(callBinding, operand);
+        argTypes.add(type);
+      }
+      TypeCoercion typeCoercion = callBinding.getValidator().getTypeCoercion();
+      RelDataType commonType = typeCoercion.getWiderTypeFor(argTypes, true);
+      if (null != commonType) {
+        // COALESCE type coercion, find a common type across all branches and casts
+        // operands to this common type if necessary.
+        boolean coerced = typeCoercion.caseOrEquivalentCoercion(callBinding);
+        if (coerced) {
+          return SqlTypeUtil.deriveType(callBinding);
+        }
+      }
+      throw callBinding.newValidationError(RESOURCE.illegalMixingOfTypes());
+    }
+
+    if (returnType == null) {
+      throw new IllegalArgumentException(
+          "Cannot infer return type for " + opBinding.getOperator() + "; operand types: "
+              + opBinding.collectOperandTypes());
+    }
+
+    return returnType;
+  }
+
+  @Override public SqlReturnTypeInference getReturnTypeInference() {
+    return requireNonNull(super.getReturnTypeInference(), "returnTypeInference");
   }
 }
--- a/core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercion.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercion.java
@@ -129,10 +129,29 @@ public interface TypeCoercion {
   /**
    * Coerces CASE WHEN statement branches to one common type.
    *
-   * <p>Rules: Find common type for all the then operands and else operands,
-   * then try to coerce the then/else operands to the type if needed.
+   * @deprecated Use {@link #caseOrEquivalentCoercion} instead.
    */
-  boolean caseWhenCoercion(SqlCallBinding binding);
+  @Deprecated boolean caseWhenCoercion(SqlCallBinding binding);
+
+  /**
+   * Type coercion in CASE WHEN, COALESCE, and NULLIF.
+   *
+   * <p>Rules:
+   * <ol>
+   *   <li>
+   *     CASE WHEN collect all the branches types including then
+   *     operands and else operands to find a common type, then cast the operands to the common type
+   *     when needed.</li>
+   *   <li>
+   *     COALESCE collect all the branches types to find a common type,
+   *     then cast the operands to the common type when needed.</li>
+   *   <li>
+   *     NULLIF returns the first operand if the two operands are not equal,
+   *     otherwise it returns a null value of the type of the first operand,
+   *     without return type coercion.</li>
+   * </ol>
+   */
+  boolean caseOrEquivalentCoercion(SqlCallBinding binding);
 
   /**
    * Type coercion with inferred type from passed in arguments and the
--- a/core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
@@ -383,11 +383,12 @@ public class TypeCoercionImpl extends AbstractTypeCoercion {
   }
 
   /**
-   * CASE and COALESCE type coercion, collect all the branches types including then
+   * CASE WHEN type coercion, collect all the branches types including then
    * operands and else operands to find a common type, then cast the operands to the common type
    * when needed.
    */
-  @Override public boolean caseWhenCoercion(SqlCallBinding callBinding) {
+  @SuppressWarnings("deprecation")
+  public boolean caseWhenCoercion(SqlCallBinding callBinding) {
     // For sql statement like:
     // `case when ... then (a, b, c) when ... then (d, e, f) else (g, h, i)`
     // an exception throws when entering this method.
@@ -418,6 +419,40 @@ public class TypeCoercionImpl extends AbstractTypeCoercion {
             || coerced;
       }
       return coerced;
+    }
+    return false;
+  }
+
+  /**
+   * Coerces CASE WHEN and COALESCE statement branches to a unified type.
+   * NULLIF returns the same type as the first operand without return type coercion.
+   */
+  @Override public boolean caseOrEquivalentCoercion(SqlCallBinding callBinding) {
+    if (callBinding.getCall().getKind() == SqlKind.COALESCE) {
+      // For sql statement like: `coalesce(a, b, c)`
+      return coalesceCoercion(callBinding);
+    } else if (callBinding.getCall().getKind() == SqlKind.NULLIF) {
+      // For sql statement like: `nullif(a, b)`
+      return false;
+    } else {
+      assert callBinding.getCall() instanceof SqlCase;
+      return caseWhenCoercion(callBinding);
+    }
+  }
+
+  /**
+   * COALESCE type coercion, collect all the branches types to find a common type,
+   * then cast the operands to the common type when needed.
+   */
+  private boolean coalesceCoercion(SqlCallBinding callBinding) {
+    List<RelDataType> argTypes = new ArrayList<>();
+    SqlValidatorScope scope = getScope(callBinding);
+    for (SqlNode node : callBinding.operands()) {
+      argTypes.add(validator.deriveType(scope, node));
+    }
+    RelDataType widerType = getWiderTypeFor(argTypes, true);
+    if (null != widerType) {
+      return coerceOperandsType(scope, callBinding.getCall(), widerType);
     }
     return false;
   }
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlCaseOperator.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlCaseOperator.java
@@ -272,7 +272,7 @@ public class SqlCaseOperator extends SqlOperator {
         // (with the correct nullability) in SqlValidator
         // instead of the commonType as the return type.
         if (null != commonType) {
-          coerced = typeCoercion.caseWhenCoercion(callBinding);
+          coerced = typeCoercion.caseOrEquivalentCoercion(callBinding);
           if (coerced) {
             ret = SqlTypeUtil.deriveType(callBinding);
           }
--- a/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
+++ b/core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
@@ -704,6 +704,16 @@ public class SqlValidatorTest extends SqlValidatorTestCase {
         .columnType("DECIMAL(5, 2)");
     expr("nullif(345.21, 2e0)")
         .columnType("DECIMAL(5, 2)");
+    expr("nullif(DATE '2020-01-01', '2020-01-01')")
+        .columnType("DATE");
+    expr("nullif('2020-01-01', DATE '2020-01-01')")
+        .columnType("CHAR(10)");
+    expr("nullif(DATE '2020-01-01', '2020-01-01')")
+        .withValidatorConfig(c -> c.withCallRewrite(false))
+        .columnType("DATE");
+    expr("nullif('2020-01-01', DATE '2020-01-01')")
+        .withValidatorConfig(c -> c.withCallRewrite(false))
+        .columnType("CHAR(10)");
     wholeExpr("nullif(1,2,3)")
         .fails("Invalid number of arguments to function 'NULLIF'. Was "
             + "expecting 2 arguments");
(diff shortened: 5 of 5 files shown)


core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -383,11 +384,16 @@ protected boolean booleanEquality(SqlCallBinding binding,
...
 
   /**
-   * CASE and COALESCE type coercion, collect all the branches types including then
+   * CASE WHEN and COALESCE type coercion, collect all the branches types including then
    * operands and else operands to find a common type, then cast the operands to the common type
    * when needed.
    */
   @Override public boolean caseWhenCoercion(SqlCallBinding callBinding) {
+    if (!(callBinding.getCall() instanceof SqlCase)) {
+      // For sql statement like: `coalesce(a, b, c)`
```
mihaibudiu, line 393: This is better. The JavaDoc suggests that this was the original intent of the code too, but maybe it was never realized.

core/src/main/java/org/apache/calcite/sql/fun/SqlCoalesceFunction.java
```diff
@@ -80,4 +88,30 @@ public SqlCoalesceFunction() {
     assert call.getFunctionQuantifier() == null;
     return SqlCase.createSwitched(pos, null, whenList, thenList, elseExpr);
   }
+
+  @Override public RelDataType inferReturnType(
```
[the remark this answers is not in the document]
juntaozhang: I'm a bit confused because neither the parent class nor the other subclasses have added it. If we force it, there will be some warnings:
```
Method annotated with @Nullable must not override @DefaultQualifier method 
```
Do we need it here?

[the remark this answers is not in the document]
mihaibudiu: You need to make the checker framework happy

core/src/main/java/org/apache/calcite/sql/SqlOperator.java
```diff
@@ -556,10 +556,28 @@ protected void preValidateCall(SqlValidator validator,
    * {@link SqlCall})
    * @return inferred return type
    */
-  public RelDataType inferReturnType(
-      SqlOperatorBinding opBinding) {
+  public RelDataType inferReturnType(SqlOperatorBinding opBinding) {
```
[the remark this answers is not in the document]
juntaozhang: I'm also a little unsure and need your further guidance:
```
  protected final RelDataType inferReturnType(
      SqlOperatorBinding opBinding) {
    if (returnTypeInference != null) {
      RelDataType returnType = returnTypeInference.inferReturnType(opBinding);
      //=====================
      // here is our main logic, in this case there are some duplicated code, returnTypeInference and operandTypeInference need change private to protected
     if (opBinding instanceof SqlCallBinding) {
        SqlCallBinding callBinding = (SqlCallBinding) opBinding;
        if (callBinding.isTypeCoercionEnabled()) {
          List<RelDataType> argTypes = new ArrayList<>();
          for (SqlNode operand : callBinding.operands()) {
            RelDataType type = SqlTypeUtil.deriveType(callBinding, operand);
            argTypes.add(type);
          }
          TypeCoercion typeCoercion = callBinding.getValidator().getTypeCoercion();
          RelDataType commonType = typeCoercion.getWiderTypeFor(argTypes, true);
          if (null != commonType) {
            boolean coerced = typeCoercion.caseWhenCoercion(callBinding);
            if (coerced) {
              return SqlTypeUtil.deriveType(callBinding);
            }
          }
          throw callBinding.newValidationError(RESOURCE.illegalMixingOfTypes());
        }
      }
      //============================
      if (operandTypeInference != null
          && opBinding instanceof SqlCallBinding
          && this instanceof SqlFunction) {
        final SqlCallBinding callBinding = (SqlCallBinding) opBinding;
        final List<RelDataType> operandTypes = opBinding.collectOperandTypes();
        if (operandTypes.stream().anyMatch(t -> t.getSqlTypeName() == SqlTypeName.ANY)) {
          final RelDataType[] operandTypes2 = operandTypes.toArray(new RelDataType[0]);
          operandTypeInference.inferOperandTypes(callBinding, returnType, operandTypes2);
          ((SqlValidatorImpl) callBinding.getValidator())
              .callToOperandTypesMap
              .put(callBinding.getCall(), ImmutableList.copyOf(operandTypes2));
        }
      }

      return returnType;
    }
```

core/src/main/java/org/apache/calcite/sql/fun/SqlCoalesceFunction.java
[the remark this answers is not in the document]
juntaozhang: got it

core/src/main/java/org/apache/calcite/sql/SqlOperator.java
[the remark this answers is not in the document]
juntaozhang: Done, rollback inferReturnType to original behaviour.

core/src/main/java/org/apache/calcite/sql/SqlOperator.java
```diff
@@ -114,10 +114,10 @@ public abstract class SqlOperator {
   private final int rightPrec;
 
   /** Used to infer the return type of a call to this operator. */
-  private final @Nullable SqlReturnTypeInference returnTypeInference;
+  protected final @Nullable SqlReturnTypeInference returnTypeInference;
```
mihaibudiu, line 117: why is this change needed?
There are get functions to access these fields.

core/src/main/java/org/apache/calcite/sql/fun/SqlCoalesceFunction.java
```diff
@@ -80,4 +94,52 @@ public SqlCoalesceFunction() {
...
+
+    if (operandTypeInference != null && opBinding instanceof SqlCallBinding) {
+      final SqlCallBinding callBinding = (SqlCallBinding) opBinding;
+      final List<RelDataType> operandTypes = opBinding.collectOperandTypes();
+      if (operandTypes.stream().anyMatch(t -> t.getSqlTypeName() == SqlTypeName.ANY)) {
+        final RelDataType[] operandTypes2 = operandTypes.toArray(new RelDataType[0]);
+        operandTypeInference.inferOperandTypes(callBinding, returnType, operandTypes2);
+        ((SqlValidatorImpl) callBinding.getValidator())
+            .callToOperandTypesMap
+            .put(callBinding.getCall(), ImmutableList.copyOf(operandTypes2));
```
mihaibudiu, line 139: I am not sure that inferReturnType is allowed to change the callBinding. This leads to bugs such https://issues.apache.org/jira/browse/CALCITE-6743

The change to the callBinding has to be done in inferOperandTypes.

juntaozhang: Thank you for your kindly reminding.  This is indeed redundant processing, as this value is null during construction. I truly appreciate your review.

core/src/main/java/org/apache/calcite/sql/SqlOperator.java
juntaozhang: Got it, thanks for your reminding.

core/src/main/java/org/apache/calcite/sql/fun/SqlCoalesceFunction.java
```diff
@@ -80,4 +93,40 @@ public SqlCoalesceFunction() {
     assert call.getFunctionQuantifier() == null;
     return SqlCase.createSwitched(pos, null, whenList, thenList, elseExpr);
   }
+
+  @Override @NonNull public RelDataType inferReturnType(
```
mihaibudiu, line 97: I am afraid that this solution has the same problems as your previous one: it modifies (using type coercion) operand types during the *return* type inference. I think you are only allowed to change the operand types during *operand* type inference.

mihaibudiu: strange, I see that Case is in fact doing the same thing

mihaibudiu: I take back my comment above and I will approve this solution.
I guess I don't yet understand completely how the type checker works.
I see that `SqlOperator.validateOperands` calls `checkOperandTypes`, then `inferReturnType`, which in turn may call `operandTypeInference`. This is strange - inferring operand types after they have already been checked.
But we won't fix this problem in this PR.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -383,11 +384,16 @@ protected boolean booleanEquality(SqlCallBinding binding,
   }
 
   /**
-   * CASE and COALESCE type coercion, collect all the branches types including then
+   * CASE WHEN and COALESCE type coercion, collect all the branches types including then
    * operands and else operands to find a common type, then cast the operands to the common type
    * when needed.
    */
   @Override public boolean caseWhenCoercion(SqlCallBinding callBinding) {
+    if (!(callBinding.getCall() instanceof SqlCase)) {
```
mihaibudiu, line 392: How about you invert the flow of control here and call `coalesceCoercion` only if `call.getKind() == SqlKind.COALESCE`?
That is more robust.

core/src/main/java/org/apache/calcite/sql/fun/SqlCoalesceFunction.java
```diff
@@ -80,4 +93,40 @@ public SqlCoalesceFunction() {
...
+      SqlCallBinding callBinding = (SqlCallBinding) opBinding;
+      List<RelDataType> argTypes = new ArrayList<>();
+      for (SqlNode operand : callBinding.operands()) {
+        RelDataType type = SqlTypeUtil.deriveType(callBinding, operand);
+        argTypes.add(type);
+      }
+      TypeCoercion typeCoercion = callBinding.getValidator().getTypeCoercion();
+      RelDataType commonType = typeCoercion.getWiderTypeFor(argTypes, true);
+      if (null != commonType) {
+        boolean coerced = typeCoercion.caseWhenCoercion(callBinding);
```
mihaibudiu, line 112: please add a comment here that `caseWhenCoercion` is supposed to handle `COALESCE` too - otherwise readers may thing it's a bug.
I think the method name is wrong, but we cannot change it, since it's public.
(one thing we can do is to deprecate this method and create one with the proper name)

mihaibudiu (approved):
This is pretty close, but please make the changes suggested too.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
juntaozhang: Yeal, very good proposal.

core/src/main/java/org/apache/calcite/sql/fun/SqlCoalesceFunction.java
juntaozhang: I totally agree, and in ansi doc, `COALESCE` and `NULLIF` is a define in `case abbreviation`[SQL 2023 standard, 6.12\<case expression\>](https://www.iso.org/standard/76584.html).

How about add new API name it by `caseExpressionCoercion `? 
```
<case expression> ::= <case abbreviation> | <case specification>
```

mihaibudiu: what we can do is to add a new API caseOrEquivalentCoercion and make the existing one call this one, and deprecate the existing one

juntaozhang: Done, thank you for your suggestion.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -383,12 +383,16 @@ protected boolean booleanEquality(SqlCallBinding binding,
...
+    return caseOrEquivalentCoercion(callBinding);
+  }
+
   /**
    * CASE WHEN and COALESCE type coercion, collect all the branches types including then
    * operands and else operands to find a common type, then cast the operands to the common type
    * when needed.
    */
-  @Override public boolean caseWhenCoercion(SqlCallBinding callBinding) {
+  @Override public boolean caseOrEquivalentCoercion(SqlCallBinding callBinding) {
```
mihaibudiu, line 395: do we need to worry about NULLIF?
Can you take a look to see whether that's also expanded?

mihaibudiu: BTW: I would add in the Java doc what you wrote about the ANSI standard: these 3 functions are considered "Related"

mihaibudiu (approved):
I like this.
Would be nice to get more comments on it.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -382,12 +383,22 @@ protected boolean booleanEquality(SqlCallBinding binding,
...
+
   /**
-   * CASE and COALESCE type coercion, collect all the branches types including then
+   * CASE WHEN and COALESCE type coercion, collect all the branches types including then
    * operands and else operands to find a common type, then cast the operands to the common type
    * when needed.
    */
-  @Override public boolean caseWhenCoercion(SqlCallBinding callBinding) {
+  @Override public boolean caseOrEquivalentCoercion(SqlCallBinding callBinding) {
+    if (!(callBinding.getCall() instanceof SqlCase)
```
xiedeyantu, line 396: Do we need to judge the SqlCase type here?

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -382,12 +383,22 @@ protected boolean booleanEquality(SqlCallBinding binding,
     return false;
   }
 
+  @Deprecated @Override public boolean caseWhenCoercion(SqlCallBinding callBinding) {
+    return caseOrEquivalentCoercion(callBinding);
```
xiedeyantu, line 387: If you want to keep the original logic, why return the new method?

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
juntaozhang: > do we need to worry about NULLIF? Can you take a look to see whether that's also expanded?

After my study, `NULLIF ` `returnType` use `ARG0_FORCE_NULLABLE`, it means using first arg type , but `COALESCE` is `LEAST_RESTRICTIVE` need `inferReturnType`. 
So, typeCoercion only happen in `case when` and `COALESCE`.

`NULLIF (V1, V2) `  definiton:
```
CASE WHEN V1=V2 THEN NULL ELSE V1 END
```

mihaibudiu: from your description it seems that NULLIF is wrong too: the arguments should be coerced and then the type of ARG0 is correct.

can you also please add a test case for NULLIF?

If we are changing the API to handle all CASE-equivalents, we need to handle all equivalent operations.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercion.java
```diff
@@ -127,12 +127,19 @@ boolean rowTypeCoercion(
   boolean binaryComparisonCoercion(SqlCallBinding binding);
 
   /**
-   * Coerces CASE WHEN statement branches to one common type.
+   * Coerces CASE WHEN and COALESCE statement branches to one common type.
```
mihaibudiu, line 130: if we add NULLIF, it will need to be here as well

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
xiedeyantu: Maybe I didn't describe it clearly. Is `callBinding.getCall().getKind() == SqlKind.COALESCE` not enough to determine whether it is COALESC?

mihaibudiu: it is.
I think looking at the kind is more efficient too than calling instanceof
I favor positive checks == instead of !=, the meaning is unambiguous when you refactor the code.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
xiedeyantu: Why is it marked solved without any explanation?

mihaibudiu: I don't think we promise to preserve the original functionality for deprecated methods, just their signature.
But there is an argument to be made for preserving the original functionality too.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -383,11 +383,28 @@ protected boolean booleanEquality(SqlCallBinding binding,
...
+    if (callBinding.getCall().getKind() == SqlKind.COALESCE) {
+      // For sql statement like: `coalesce(a, b, c)`
+      return coalesceCoercion(callBinding);
+    }
+
+    if (callBinding.getCall().getKind() == SqlKind.NULLIF) {
+      // For sql statement like: `nullif(a, b)`
+      return nullifCoercion(callBinding);
+    }
+
```
mihaibudiu, line 407: can you also add an assertion here that the remaining call is CASE?

mihaibudiu: for symmetry you may also want to move this code into a separate private function

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -422,6 +439,34 @@ protected boolean booleanEquality(SqlCallBinding binding,
...
+    }
+    RelDataType widerType = getWiderTypeFor(argTypes, true);
+    if (null != widerType) {
+      return coerceOperandsType(scope, callBinding.getCall(), widerType);
+    }
+    return false;
+  }
+
+  /**
+   * NULLIF type coercion, cast the second operand type to the first operand type when needed.
```
mihaibudiu, line 460: Are you sure this is right?
When you translate this to a CASE statement the wider type will be used by coercion.
You have to preserve this behavior.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
xiedeyantu: I understand that if we want to keep the original function, we should keep it intact. Or we can just modify the original method directly.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
mihaibudiu: Maybe there is a misunderstanding, I didn't say that the NULLIF coercion has to be removed, I said it has to behave like the other two coercions, using the wider type.

There is a commutative diagram which must hold at all times:

```
NULLIF ----------------> CASE
     | coercion                      | coercion
     v                                    v
NULLIF----------------->CASE
```

In other words, if you choose to expand NULLIF into CASE and then coerce, or if you coerce first and then expand, you should get the exact same result. So once you fix the coercion for CASE (which is fixed in the original code), you get a fixed coercion rule for NULLIF. All this is true for COALESCE as well.

BTW: I would recommend putting this diagram in the comments of the coercion for NULLIF; if someone implements a new validator, this rule has to be obeyed.

mihaibudiu: it would also be nice to have a test for this diagram, but it won't be trivial to write.

mihaibudiu:
this was pretty close, I hope we can finalize it.
I think you need the nullif coercion

juntaozhang:
> this was pretty close, I hope we can finalize it. I think you need the nullif coercion

I’m happy to hear from you. You said NULLIF should return the wider type, like the other coercions. But I’ve found it actually returns the type of the first argument, without coercion. It would be a big help if you could explain this, this will help me finish the PR.

My related statement is outlined in my previous reply.

mihaibudiu:
It looks to me that NULLIF is the same as 
```
CASE 
    WHEN expression1 = expression2 THEN NULL
    ELSE expression1
END
```
I guess that this indeed does return always the type of expression1, made nullable.
So I guess we are good.
There is a coercion, but that is not visible in the type inference.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercion.java
```diff
@@ -129,10 +129,27 @@ boolean rowTypeCoercion(
...
+   * <ol>
+   *   <li>
+   *     CASE WHEN collect all the branches types including then
+   *     operands and else operands to find a common type, then cast the operands to the common type
+   *     when needed.</li>
+   *   <li>
+   *     COALESCE collect all the branches types to find a common type,
+   *     then cast the operands to the common type when needed.</li>
+   *   <li>
+   *     NULLIF return type is the first operand type, with no coercion applied.</li>
```
mihaibudiu, line 149: made nullable if necessary

mihaibudiu (approved):

mihaibudiu (commented):
Can you please squash the commits?

PR #2364, opened 2021-03-05
[CALCITE-4340] Fixing Subqueries in On Clauses

jamesstarr:
[CALCITE-4340] Throwing an exception when correlated sub-queries on clauses are found.
Fixing the offset of field references for sub-queries in on clauses.

PR #4354, opened 2025-04-30
[CALCITE-6963] SqlToRelConverter fails when subquery is in join on clause

xiedeyantu:


core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -3686,6 +3686,26 @@ LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$
...
+and t1.ename in (select t3.ename from emp t3)]]>
+    </Resource>
+    <Resource name="plan">
+      <![CDATA[
+LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8])
+  LogicalJoin(condition=[=($7, $9)], joinType=[left])
+    LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+    LogicalJoin(condition=[=($1, $2)], joinType=[inner])
+      LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
+      LogicalAggregate(group=[{}], agg#0=[SINGLE_VALUE($0)])
```
mihaibudiu, line 3705: why is this plan correct?
this seems to assume that ename is a key

mihaibudiu: this PR could benefit from some quidem tests too

xiedeyantu: I tried to write a quidem test and use expand=true/false to test the execution results, and they are consistent. Is this OK?

mihaibudiu: I see why the aggregate is fine, because IN is satisfied with a set, even if the table is a multiset.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
```diff
@@ -1227,6 +1227,16 @@ private void substituteSubQuery(Blackboard bb, SubQuery subQuery) {
...
         // values list into an inline table for the
         // reference to Q below.
+      } else if (query instanceof SqlSelect) {
+        // During JOIN ON condition processing, bb.root is
+        // initially null but subquery expansion requires
+        // a non-null bb.root. For IN/NOT IN subqueries
+        // processed via convertExists, we use the join's
+        // left input as bb.root. Exception: in UPDATE queries,
+        // bb.inputs remains null and subquery rewriting is skipped.
+        if (bb.inputs != null && bb.inputs.size() == 2) {
```
mihaibudiu, line 1237: is there a way to assert here that this is indeed inside a JOIN?
I am worried about other patterns matching.

xiedeyantu: It's not easy to obviously demonstrate the problem in a quidem test, but I did enter my changes when debugging.

mihaibudiu: This request is not about quidem tests.
The comments says "we use the join's left input".
How do you know that's from a join?
Can you check that?

xiedeyantu: Currently, only when there is a subquery in the join on condition, `bb.root` will be null and `query` will be of `SqlSelect` type. I wrote `bb.inputs.size() == 2` in the code to confirm that `bb.inputs` does have the two i`nputs` to join.

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
xiedeyantu: Here I refer to the processing of the `in` subquery in `where`, and use the same processing method.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
mihaibudiu: there are other operators with 2 inputs

xiedeyantu: I understand that `join` is the only method that can have subqueries and multiple `inputs`. I didn't expect other cases. The `substituteSubQuery` method also involves subqueries. I don't know if I understand it correctly.

mihaibudiu: why is the left input to the join the correct root?
what if that is itself a join?

xiedeyantu: If the subquery is in `where`, the `inputs` may be a `join`. If the subquery is in the `on` condition of a `join`, the `inputs` will be `left` and `right`. The `root` used in the [register](https://github.com/apache/calcite/pull/4354/files#diff-5d7c774cd39d94b119eedcf3d361e72406c7864e916736c533bbe8f893ca72bcR1336 ) method is `left`.

mihaibudiu: Yes, but the docs say that the root cannot be a join.
I am wondering about the case when the left input of the join is a join.
Maybe you can at least `assert !(root instanceof Join)` after the assignment?
Moreover, directly assigning to root is strange, I have never seen this pattern in Calcite.
I would have expected a use of setRoot.
It is true that there is another example in this file, and that code was written by @julianhyde.

xiedeyantu: Oh, I can use setRoot instead.

mihaibudiu: Frankly, I am not sure this fix is correct for all possible queries (because I don't really know how the blackboard works), but it is an improvement, since some correct queries which used to be rejected are now correctly compiled, so I will approve.

mihaibudiu (approved):

mihaibudiu:
Since this is a small fix, you can probably squash the commits.
I hope that someone more knowledgeable than me can also take a look, so I will wait a bit to merge it.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
dssysolyatin: > Currently, only when there is a subquery in the join on condition, bb.root will be null and query will be of SqlSelect type. I wrote bb.inputs.size() == 2 in the code to confirm that bb.inputs does have the two inputs to join

Is it possible to use `bb.scope` to check if it's a join ? The current statement is probably correct for now, but it might be difficult to debug if any issues come up later

core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java
```diff
@@ -5679,4 +5679,11 @@ void checkUserDefinedOrderByOver(NullCollation nullCollation) {
     String sql = "SELECT CAST(CAST(? AS INTEGER) AS CHAR)";
     sql(sql).ok();
   }
+
+  @Test void testJoinOnConditionWithSubquery() {
```
suibianwanwank, line 5683: @xiedeyantu, can you add this test , I think it may not pass the test .
```
  @Test void testJoinOnConditionWithSubquery2() {
    final String sql = "select t1.* from emp t1\n"
        + "left join dept t2 on t1.deptno = t2.deptno\n"
        + "and t2.deptno in (select t3.ename from emp t3)";
    sql(sql).ok();
  }
```

xiedeyantu: This is a good case. I may need to determine whether deptno comes from the left or right.

core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
xiedeyantu: @dssysolyatin Thanks for your suggestion, that might be a good idea.

suibianwanwank:
Before this test gets fixed together, I'm removing the LGTM tags to avoid the PR being merged by mistake.

xiedeyantu:
@mihaibudiu @dssysolyatin @suibianwanwank It may not be possible to simply select one from `bb.inputs` as `bb.root`. If the first operand of the in only involves the `left` or `right` of the `join`, perhaps we can choose one as `bb.root`. If the first operand involves both `left` and `right` (for example, `t1.deptno * t2.deptno in (...)`), but the current `join` has not been created yet, it may require a lot of changes to adapt. I haven't thought of a good way to be compatible with all cases. But this situation can be handled in `SubQueryRemoveRule`. So when encountering this complex case, forcing `expand` to `false` to let the case pass, I don't know if it is a good way to deal with it. If you think it is not good, then I will close this PR first.

mihaibudiu:
To find a solution we would need to know that "root" represents in the blackboard, and why it is needed during the conversion process. Then we can figure out whether there is a suitable value for it while converting this join.

Even if you cannot handle the most general case, if you can handle this case, it's already an improvement, as long as other cases are not handled incorrectly.

xiedeyantu:
> To find a solution we would need to know that "root" represents in the blackboard, and why it is needed during the conversion process. Then we can figure out whether there is a suitable value for it while converting this join.
> 
> Even if you cannot handle the most general case, if you can handle this case, it's already an improvement, as long as other cases are not handled incorrectly.

So do you think my current changes are acceptable? Or should I continue with my original idea? I haven't come up with a complete solution for my original idea.

mihaibudiu:
If would be useful to throw if you cannot handle the query, rather than produce a potentially incorrect result

xiedeyantu:
> If would be useful to throw if you cannot handle the query, rather than produce a potentially incorrect result

The current idea is to keep the subquery in the on condition unexpanded, so that no error will occur in the result.

mihaibudiu:
As long as correct inputs produce correct outputs, and more correct inputs are handled, this is an improvement.

xiedeyantu:
> As long as correct inputs produce correct outputs, and more correct inputs are handled, this is an improvement.

If you basically approve of my current changes, shall we continue to wait for suggestions from other reviewers?

mihaibudiu:
As I said, I don't know how the blackboard works, so I appreciate other comments.

xiedeyantu:
Please see [CALCITE-4340](https://issues.apache.org/jira/browse/CALCITE-4340) I found that someone had discovered this issue a long time ago, and it was a related situation. Julian also explained why there are two logic for eliminating subqueries. So is it better to directly hand over this situation to SubQueryRemoveRule?

mihaibudiu:
There is an open PR, we could take a look to see what it's doing: https://github.com/apache/calcite/pull/2364

xiedeyantu:
> There is an open PR, we could take a look to see what it's doing: #2364

I have roughly read this PR. It is about the subquery expansion in sql to rel. I wonder if we need to implement a similar logic as SubQueryRemoveRule here? I have an idea. Should the subquery logic in sql to rel be refactored to keep it consistent with SubQueryRemoveRule?

mihaibudiu:
That discussion happened before I started working on Calcite, and this is a part of the code I am not familiar with.
You should probably ask this question in the JIRA of the old issue.

xiedeyantu:
> That discussion happened before I started working on Calcite, and this is a part of the code I am not familiar with. You should probably ask this question in the JIRA of the old issue.

I don't really understand the differences between the two parts yet, maybe I'll ask a question when I figure it out. So should this PR be closed now, or should I wait and see if there are any comments?

suibianwanwank:
I seem to have some misunderstanding here. According to the documentation below, the default value of `isExpand` should be `false`, and when `expand` is `true`, certain issues may not necessarily be maintained. However, in `SqlToRelConverter`, the default value of `expand` appears to be `true`, which is causing me some confusion.
```
    /** Returns the {@code expand} option. Controls whether to expand
     * sub-queries. If false (the default), each sub-query becomes a
     * {@link org.apache.calcite.rex.RexSubQuery}.
     *
     * <p>Setting {@code expand} to true is deprecated. Expansion still works,
     * but there will be less development effort in that area. */
    @Value.Default default boolean isExpand() {
      return false;
    }
```

mihaibudiu:
Frankly I don't know why the comment recomments "isExpand" is false. Many subqueries expanded in that case cannot yet be handled by the existing decorrelator. At least in our compiler we have kept "isExpand" set to "true".

PR #4355, opened 2025-05-01
[CALCITE-6978] Translation to relational algebra of correlated query with COALESCE causes an assertion failure

mihaibudiu:
I don't know if this is the right fix, but I can't imagine another one that does not require heroic efforts.
See my comments in JIRA.


caicancai (approved):

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -1528,7 +1528,9 @@ private void handleOffsetFetch(@Nullable SqlNode offset, @Nullable SqlNode fetch
           ((SqlBasicCall) call).setOperator(overloads.get(0));
         }
       }
-      if (config.callRewrite()) {
+      if (config.callRewrite()
+          // Do not rewrite calls that contain subqueries
+          && !SqlUtil.containsCall(call, c -> c.getKind() == SqlKind.SELECT)) {
```
xiedeyantu, line 1533: I assume this is reasonable, would it be better to make a judgment[here](https://github.com/apache/calcite/blob/eeb9c3d076889a5bf25d28a24318e5c1c6c89549/core/src/main/java/org/apache/calcite/sql/fun/SqlCoalesceFunction.java#L57).

mihaibudiu: There are two kinds of code transformations: mechanisms and policies.
In this instance, mechanisms are code rewrites.
Policies decide when code rewrites have to be done.
`rewriteCall` is a mechanism function.
This one is a policy function.
As such, I think it's a mistake to push the decision to `rewriteCall`.

xiedeyantu: Got it. I have another small question. The title says it is related to subqueries in COALESCE, but the PR seems to handle all subqueries. Do we need to cover all subqueries? If so, do we need to add test cases?

xiedeyantu: callRewrite also looks like a parameter similar to expand. If turn it off, can we get the correct result?

mihaibudiu: it does for this test case

mihaibudiu: As @suibianwanwank has pointed out in the Jira as well, I think this should be OFF by default. I have changed in our compiler to be this way.

xiedeyantu: Sorry, I missed some information in JIRA. I think this fix should be accepted as long as it is correct.

mihaibudiu: The only problem with this PR is that the user asks for a function expansion, and this expansion is not done for some functions which have subqueries as arguments. This means that users may see these functions in the produced code although they expect them to be gone. I think there should be an alternative way to perform these expansions after decorrelation.

suibianwanwank (approved):

core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -1528,7 +1528,9 @@ private void handleOffsetFetch(@Nullable SqlNode offset, @Nullable SqlNode fetch
           ((SqlBasicCall) call).setOperator(overloads.get(0));
         }
       }
-      if (config.callRewrite()) {
+      if (config.callRewrite()
```
NobiGo, line 1531: This modification makes this logic less pure. I assume that when callRewrite is set to true, the original author intends to rewrite all expressions, not just under specific conditions. We should ensure that the SQL can still be executed after the rewrite, except when including nondeterministic functions, where there might be incorrect results.

mihaibudiu: I have filed https://issues.apache.org/jira/browse/CALCITE-7012

PR #4358, opened 2025-05-03
[CALCITE-6998] The command !set planner-rules does not support rules with multiple Config

xiedeyantu:
Please see: [CALCITE-6998](https://issues.apache.org/jira/browse/CALCITE-6998)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/RuleConfig.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/RuleConfig.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * The {@code @RuleConfig} annotation is exclusively used for testing purposes
+ * to identify configuration names when a rule contains multiple Configs.
+ *
+ * <p>Usage Example:
+ *
+ * <pre>
+ * &#64;RuleConfig(value = "FILTER")
+ * public static final ExpandDisjunctionForJoinInputsRule
+ *   EXPAND_FILTER_DISJUNCTION_LOCAL =
+ *       ExpandDisjunctionForJoinInputsRule.Config.FILTER.toRule();
+ * &#64;RuleConfig(value = "JOIN")
+ * public static final ExpandDisjunctionForJoinInputsRule
+ *   EXPAND_JOIN_DISJUNCTION_LOCAL =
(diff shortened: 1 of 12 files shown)


xiedeyantu:
@mihaibudiu Do you have time to help me review this PR?

xiedeyantu:
@mihaibudiu You may be the one who understands this PR best.

julianhyde:
This is a feature request, not a bug. So please add to the jira case a specification for the feature you are adding.

core/src/test/resources/sql/planner.iq
```diff
@@ -20,6 +20,8 @@
 
 # The following line is used to update the default planner rules.
 # "+" means add a rule, "-" means remove a rule. Only the "CoreRules." can be omitted.
+# If multiple rules are involved, they need to be divided into multiple lines, one rule per line,
```
mihaibudiu, line 23: I don't see a test using the new feature.

core/src/main/java/org/apache/calcite/rel/rules/RuleConfig.java
```diff
@@ -0,0 +1,39 @@
...
+ * and is used to identify the Config name when there
+ * are multiple Configs in each rule.
+ *
+ * <p>For example:
+ * ExpandDisjunctionForJoinInputsRule contains two
+ * Config(FILTER and JOIN). We should add annotations
+ * in {@link CoreRules}. See
+ * {@link CoreRules#EXPAND_JOIN_DISJUNCTION_LOCAL}
+ */
+@Target(ElementType.FIELD)
```
mihaibudiu, line 35: As @julianhyde said, if you add a short description to the JIRA about the design it may make it much easier to review this. 

There should be no comments with "we should...". Perhaps you mean "Each CoreRule must have an annotation..."

core/src/test/resources/sql/planner.iq
xiedeyantu: I updated the documentation in `planner.iq`. This is a format specification to prevent subsequent developers from writing multiple formats.

xiedeyantu:
> This is a feature request, not a bug. So please add to the jira case a specification for the feature you are adding.

I have updated the instructions in jira. Please review it when you have time.

core/src/main/java/org/apache/calcite/rel/rules/RuleConfig.java
xiedeyantu: I have also updated the documentation for this annotation. Please help me take a look.

xiedeyantu:
@mihaibudiu I have found that the current error in CI for MacOS is that `planner.iq` runs out of time. It does include the last test case and not included in my local system is 3s and 1m3s. Do you have any good suggestions for dealing with this timeout issue? I haven't come up with a good solution yet.

mihaibudiu:
I am not an expert in planner issues.
But other people have pointed out possible explanations; the planner gets into an infinite loop, or the planner search space becomes too large.
I wonder whether this rule should not be in CoreRules, maybe then the planner won't try to use it always.

silundong:
Perhaps the current Quidem is not suitable for testing this rule? The planner created in the current Quidem test is VolcanoPlanner, and the default rules is registered. However, this rule is recommended to be used alone in HepPlanner to avoid infinite loops (or making the search space of VolcanoPlanner too large).

mihaibudiu:
I have filed https://issues.apache.org/jira/browse/CALCITE-7004, but that is somewhat independent of this PR.

xiedeyantu:
> I have filed https://issues.apache.org/jira/browse/CALCITE-7004, but that is somewhat independent of this PR.

Got it, thanks! Is there anything else that needs to be modified in this PR?

core/src/main/java/org/apache/calcite/rel/rules/RuleConfig.java
```diff
@@ -0,0 +1,55 @@
...
+ * public static final ExpandDisjunctionForJoinInputsRule
+ *   EXPAND_JOIN_DISJUNCTION_LOCAL =
+ *       ExpandDisjunctionForJoinInputsRule.Config.JOIN.toRule();
+ * </pre>
+ *
+ * <p>Key Characteristics:
+ * <ul>
+ *   <li>Test-scoped annotation (not used in production code)</li>
+ *   <li>Required when a rule class has multiple Configs</li>
+ *   <li>Annotation value must match the Config name</li>
```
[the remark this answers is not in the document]
xiedeyantu: Done!

PR #4359, opened 2025-05-04
[CALCITE-6145] Function 'TRIM' without parameters throw NullPointerException

xiedeyantu:
Only squash the #3869 and rebase main.

### DIFF
--- a/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
+++ b/testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
@@ -11943,6 +11943,7 @@ public class SqlOperatorTest {
     f.checkNull("trim(null)");
     f.checkNull("trim(cast(null as varchar(1)) from 'a')");
     f.checkNull("trim('a' from cast(null as varchar(1)))");
+    f.checkNull("trim(null)");
 
     // SQL:2003 6.29.9 and SQL:2016 6.30.11: trim string must have length=1.
     // Failure occurs at runtime.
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -11954,6 +11955,7 @@ void assertSubFunReturns(boolean binary, String s, int start,
         "Invalid argument '': the length of the string describing "
             + "the trimmed character must be 1",
         true);
+    f.checkFails("trim()", "Invalid number of arguments to function 'TRIM'. Was expecting at least 2 arguments", false);
```
xiedeyantu, line 11958: This test cannot pass...

mihaibudiu: then the PR is incomplete

xiedeyantu: Yes, give me some time and I'll see if I can handle it.

xiedeyantu: Please remove the LGTM tag before I fix it.

xiedeyantu:
At present, errors will occur during the parsing stage. If TRIM cannot accept empty parameters, then this should be reasonable. I also reported parsing errors when verifying in MySQL and PGSQL. So does this PR still need for the current code?
```
Caused by: org.apache.calcite.sql.parser.SqlParseException: Encountered ")" at line 1, column 14.
Was expecting one of:
    "ABS" ...
    "ARRAY" ...
    "AVG" ...
    "BOTH" ...
```

xuzifu666:
> So does this PR still need for the current code?

There are compatibility issues in syntax here which would block horoughly fix.

mihaibudiu:
See https://issues.apache.org/jira/browse/CALCITE-6708
and https://issues.apache.org/jira/browse/CALCITE-6709
So perhaps this has been fixed indirectly.

xiedeyantu:
> > So does this PR still need for the current code?
> 
> There are compatibility issues in syntax here which would block horoughly fix.

Do we still need to fix this error? I think it's reasonable. What's your opinion?

mihaibudiu:
I will mark the issue resolved, you can close this PR

PR #4360, opened 2025-05-04
[CALCITE-6999] Invalid unparse for TRIM in PrestoDialect

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-6999


xiedeyantu (approved):
LGTM

xuzifu666:
> Why the BigQuery dialect files were modified ?

The logic can be reused by other dialects，so move it to RelToSqlConverterUtil avoid duplicate code. @caicancai

caicancai:
There are two reasons. The first is that you cannot be sure that every dialect has this function and the format is the same. The second is that you did not modify the above comment.

caicancai:
> > There are two reasons. The first is that you cannot be sure that every dialect has this function and the format is the same. The second is that you did not modify the above comment.
> 
> Yes, here need to adapt system which support trim/ltrim/rtrim functions and presto is supported well(which test result in JIRA:https://issues.apache.org/jira/browse/CALCITE-6999), and I had updated the comments with presto link. thanks for reminder @caicancai

So now all dialects support trim function parsing?

xuzifu666:
> > > There are two reasons. The first is that you cannot be sure that every dialect has this function and the format is the same. The second is that you did not modify the above comment.
> > 
> > 
> > Yes, here need to adapt system which support trim/ltrim/rtrim functions and presto is supported well(which test result in JIRA:https://issues.apache.org/jira/browse/CALCITE-6999), and I had updated the comments with presto link. thanks for reminder @caicancai
> 
> So now all dialects support trim function parsing?

This can not confirm，but which need to use(like BigQuery/Presto) to convert right way  can use the method.

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
```diff
@@ -68,6 +68,43 @@ public static void unparseHiveTrim(
...
   }
 
+  /**
+   * For usage of TRIM, LTRIM and RTRIM in BQ see
+   * <a href="https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#trim">
+   * BQ Trim Function</a>,
+   * <a href="https://prestodb.io/docs/current/functions/string.html#trim-string-varchar">
+   * Presto Trim Function</a>.
+   */
+  public static void unparseTrim(SqlWriter writer, SqlCall call, int leftPrec,
```
mihaibudiu, line 78: if this function is specific to BigQuery, the function name should reflect that.
I would not abbreviate BQ in the JavaDoc, why force people to try to guess what it is when you can just write it down?

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
```diff
@@ -68,6 +68,43 @@ public static void unparseHiveTrim(
...
+      break;
+    default:
+      operatorName = call.getOperator().getName();
+      break;
+    }
+    final SqlWriter.Frame trimFrame = writer.startFunCall(operatorName);
+    call.operand(2).unparse(writer, leftPrec, rightPrec);
+
+    // If the trimmed character is a non-space character, add it to the target SQL.
+    // eg: TRIM(BOTH 'A' from 'ABCD'
```
mihaibudiu, line 98: missing parens in comment

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
```diff
@@ -68,6 +68,43 @@ public static void unparseHiveTrim(
...
+    // Output Query: TRIM('ABC', 'A')
+    String value = requireNonNull(valueToTrim.toValue(), "valueToTrim.toValue()");
+    if (!value.matches("\\s+")) {
```
mihaibudiu, line 101: is this correct?
is trim(s, "\n") the same as trimming all whitespaces?

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
xuzifu666: Yes，make sense， this method not only used for BigQuery，can be used by any need convert trim(both from) to trim/ltrim/rtrim functions, so method name may not need to be changed, I had updated the bbreviate BQ in JavaDoc.Thanks for suggestion @mihaibudiu

mihaibudiu: the current name is too generic, this function is not suitable for unparsingTrim for all dialects, as the name would suggest
I think unparseTrimBigQuery is better, or unparseTrimLR if you want.
Many dialects can share it.

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
xuzifu666: Here should be correct, I had test for it:
```
select trim('   aa c', ' ');
>aa c
```
trim with '\n'

```
select trim('   aa c', '\n');
>   aa c
```
trim(s, "\n")  is not the same as trimming all whitespaces.

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
xuzifu666: Yes, I thought unparseTrimLR is better, so changed the function to unparseTrimLR~ @mihaibudiu

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
mihaibudiu: What about a literal tab character (not the \t escape sequence)?

xuzifu666: Yes，you are right，maybe we can use StringUtils.isWhitespace(value) to instead of it? which can only match write spaces. @mihaibudiu

xuzifu666:
Hi @mihaibudiu all comments had been addressed, whether is OK currently?

core/src/main/java/org/apache/calcite/util/RelToSqlConverterUtil.java
```diff
@@ -68,6 +70,44 @@ public static void unparseHiveTrim(
...
+      break;
+    }
+    final SqlWriter.Frame trimFrame = writer.startFunCall(operatorName);
+    call.operand(2).unparse(writer, leftPrec, rightPrec);
+
+    // If the trimmed character is a non-space character, add it to the target SQL.
+    // eg: TRIM(BOTH 'A' from 'ABCD')
+    // Output Query: TRIM('ABC', 'A')
+    String value = requireNonNull(valueToTrim.toValue(), "valueToTrim.toValue()");
+    if (!StringUtils.isWhitespace(value)) {
```
mihaibudiu, line 104: I still don't think this is correct.
The spec says that the default value is just a single *space*.
So the test should be for value.equals(" "), or it can be removed altogether.

xuzifu666: Make sense, I agree with it and had changed it, thanks @mihaibudiu

mihaibudiu (approved):
you can squash the commits

xuzifu666:
Thanks for review，I had squashed commits. @mihaibudiu

PR #4361, opened 2025-05-05
[CALCITE-7000] Extend IntersectToSemiJoinRule to support n-way inputs

xiedeyantu:
Jira link: [CALCITE-7000](https://issues.apache.org/jira/browse/CALCITE-7000)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
@@ -34,9 +34,51 @@ import java.util.ArrayList;
 import java.util.List;
 
 /**
- * Planner rule that translates a {@link org.apache.calcite.rel.core.Intersect}
+ * Planner rule that translates a {@link Intersect}
  * to a series of {@link org.apache.calcite.rel.core.Join} that type is
- * {@link org.apache.calcite.rel.core.JoinRelType#SEMI}.
+ * {@link JoinRelType#SEMI}. This rule supports n-way Intersect conversion,
+ * as this rule can be repeatedly applied during query optimization to
+ * refine the plan.
+ *
+ * <h2>Example</h2>
+ *
+ <p>Original sql:
+ * <pre>{@code
+ * select ename from emp where deptno = 10
+ * intersect
+ * select deptno from emp where ename in ('a', 'b')
+ * intersect
+ * select ename from empnullables
+ * }</pre>
+ *
+ * <p>Original plan:
+ * <pre>{@code
+ * LogicalIntersect(all=[false])
+ *   LogicalProject(ENAME=[$1])
+ *     LogicalFilter(condition=[=($7, 10)])
+ *       LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+ *   LogicalProject(DEPTNO=[CAST($7):VARCHAR NOT NULL])
+ *     LogicalFilter(condition=[OR(=($1, 'a'), =($1, 'b'))])
+ *       LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+ *   LogicalProject(ENAME=[$1])
+ *     LogicalTableScan(table=[[CATALOG, SALES, EMPNULLABLES]])
+ * }</pre>
+ *
+ * <p>Plan after conversion:
+ * <pre>{@code
+ * LogicalProject(ENAME=[CAST($0):VARCHAR])
+ *   LogicalAggregate(group=[{0}])
+ *     LogicalJoin(condition=[<=>(CAST($0):VARCHAR, CAST($1):VARCHAR)], joinType=[semi])
+ *       LogicalJoin(condition=[=(CAST($0):VARCHAR, $1)], joinType=[semi])
+ *         LogicalProject(ENAME=[$1])
+ *           LogicalFilter(condition=[=($7, 10)])
+ *             LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+ *         LogicalProject(DEPTNO=[CAST($7):VARCHAR NOT NULL])
+ *           LogicalFilter(condition=[OR(=($1, 'a'), =($1, 'b'))])
+ *             LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+ *       LogicalProject(ENAME=[$1])
+ *         LogicalTableScan(table=[[CATALOG, SALES, EMPNULLABLES]])
+ * }</pre>
  */
 @Value.Enclosing
 public class IntersectToSemiJoinRule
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/rules/IntersectToSemiJoinRule.java
```diff
@@ -34,9 +34,52 @@
...
+ * Planner rule that translates a {@link Intersect}
  * to a series of {@link org.apache.calcite.rel.core.Join} that type is
- * {@link org.apache.calcite.rel.core.JoinRelType#SEMI}.
+ * {@link JoinRelType#SEMI}. This rule supports n-way Intersect conversion,
+ * as this rule can be repeatedly applied during query optimization to
+ * refine the plan.
+ *
+ * <h2>Example</h2>
+ *
+ <p>Original sql:
```
mihaibudiu, line 45: Is this the most general example for this rule?
If so, please explain in the comments the constraints on the supported predicates - e.g., they all have to involve the same field.
If not, please provide a more general example where the predicates look very different

xiedeyantu: Done!

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -3633,6 +3633,19 @@ private void checkPushJoinThroughUnionOnRightDoesNotMatchSemiOrAntiJoin(JoinRelT
         .check();
   }
 
+  /** Test case for <a href="https://issues.apache.org/jira/browse/CALCITE-7000">[CALCITE-7000]
+   * Extend IntersectToSemiJoinRule to support n-way inputs</a>. */
+  @Test void testIntersectToSemiJoin2() {
```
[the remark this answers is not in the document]
xiedeyantu: Done!

core/src/test/resources/sql/planner.iq
```diff
@@ -128,7 +128,60 @@ EnumerableCalc(expr#0..2=[{inputs}], $f0=[$t1], $f1=[$t2])
         EnumerableCalc(expr#0=[{inputs}], expr#1=[IS NOT NULL($t0)], DEPTNO=[$t0], $condition=[$t1])
           EnumerableValues(tuples=[[{ 10 }, { 10 }, { 20 }, { 30 }, { 30 }, { 50 }, { 50 }, { 60 }, { null }]])
 !plan
+!set planner-rules original
+
```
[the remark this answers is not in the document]
xiedeyantu: Done!

core/src/test/resources/sql/planner.iq
```diff
@@ -132,12 +132,13 @@ EnumerableCalc(expr#0..2=[{inputs}], $f0=[$t1], $f1=[$t2])
...
 # [CALCITE-7000] Extend IntersectToSemiJoinRule to support n-way inputs
 !set planner-rules "-EnumerableRules.ENUMERABLE_INTERSECT_RULE,
+  -CoreRules.INTERSECT_TO_DISTINCT,
   +CoreRules.INTERSECT_TO_SEMI_JOIN"
 select a from (values (1), (2), (3), (4), (5)) as t1 (a)
 intersect
 select a from (values (1), (2)) as t2 (a)
 intersect
-select a from (values (1), (4)) as t3 (a);
+select a from (values (1), (4), (null)) as t3 (a);
```
mihaibudiu: I think the idea is to run the same test with both old and new rules and see that the results are the same
(there is no way to say "the same", but the table output should be the same).
So can you also run it once with the existing rules?

mihaibudiu (approved):

mihaibudiu:
You can probably squash the commits at the same time you add the last test. Thank you.

core/src/test/resources/sql/planner.iq
xiedeyantu: Thank you for your detailed review. The reason for this result is that there is no good way to control the order of rules, as you mentioned before. This is the only way to do it now, and it may be the same for JdbcTest. I will think about what good way to control the order later.

mihaibudiu: I don't think the order is important. Can you run the same query before you modify the rules?

xiedeyantu: Sorry, I didn't explain it clearly. It is because the modified rule depends on INTERSECT_MERGE, and if we had done INTERSECT_TO_SSI_JOIN first, it would have produced the current result.

mihaibudiu: is there a way to run this example in quidem with and without the new rule and compare results?

xiedeyantu: I have currently modified the code and it seems that I can no longer execute the rules before the modification. Do you want to run the 2-way rule? If you want to run without INTERSECT_TO_SEMI_JOIN, there is currently a case in the submission that uses  INTERSECT for calculations.

mihaibudiu: I don't really understand what you are saying.
What does it mean "I can no longer execute the rules before the modification"?
You are not executing any rules, the compiler is, when preparing a test.

core/src/test/resources/sql/planner.iq
```diff
@@ -128,7 +128,64 @@ EnumerableCalc(expr#0..2=[{inputs}], $f0=[$t1], $f1=[$t2])
...
++-----+
+| A   |
++-----+
+| 1.0 |
++-----+
+(1 row)
+
+!ok
+
+EnumerableIntersect(all=[false])
```
xiedeyantu, line 179: @mihaibudiu Do you want this case?

mihaibudiu: yes, this looks good

xiedeyantu:
> You can probably squash the commits at the same time you add the last test. Thank you.

Should I try hard coding and verify it locally?

mihaibudiu:
not sure what you mean by "hard coding"
I have approved

xiedeyantu:
> not sure what you mean by "hard coding" I have approved

I tried to execute this case in JdbcTest and found that 3-way INTERSECT triggered the rule, but it was abandoned by Volcano because of its high cost. Then I'll squash commit first, and then deal with other problems if they arise?
```
  @Test void testIntersectToSemiJoin2() {
    final String sql = ""
        + "select a from (values (1.0), (2.0), (3.0), (4.0), (5.0)) as t1 (a)\n" +
        "intersect\n" +
        "select a from (values (1), (2)) as t2 (a)\n" +
        "intersect\n" +
        "select a from (values (1.0), (4.0), (null)) as t3 (a)";
    CalciteAssert.hr()
        .query(sql)
        .withHook(Hook.PLANNER, (Consumer<RelOptPlanner>)
            planner -> {
              List<RelOptRule> rules = planner.getRules();
              for (RelOptRule rule : rules) {
                planner.removeRule(rule);
              }
              planner.addRule(CoreRules.INTERSECT_TO_SEMI_JOIN);
              planner.addRule(EnumerableRules.ENUMERABLE_JOIN_RULE);
              planner.addRule(EnumerableRules.ENUMERABLE_VALUES_RULE);
              planner.addRule(EnumerableRules.ENUMERABLE_AGGREGATE_RULE);
              planner.addRule(EnumerableRules.ENUMERABLE_CALC_RULE);
              planner.addRule(EnumerableRules.ENUMERABLE_PROJECT_RULE);
              planner.addRule(EnumerableRules.ENUMERABLE_FILTER_RULE);
              planner.addRule(CoreRules.INTERSECT_MERGE);
              planner.addRule(CoreRules.INTERSECT_TO_SEMI_JOIN);
            })
        .explainContains("")
        .returnsUnordered("A=1.0");
  }
```
graphviz

mihaibudiu:
You are pointing to a shortcoming of the implementation of the quidem tests that specify optimization rules.
If the hook uses the Volcano planner, there is no guarantee that the rules are applied.
Ideally the rules specified using the !set commands would be inserted in a Hep program executed *before* the Volcano planner. I will file an issue for this.

xiedeyantu:
> You are pointing to a shortcoming of the implementation of the quidem tests that specify optimization rules. If the hook uses the Volcano planner, there is no guarantee that the rules are applied. Ideally the rules specified using the !set commands would be inserted in a Hep program executed _before_ the Volcano planner. I will file an issue for this.

It may not be very accurate, the rule was indeed used, but it was not selected in the end.

caicancai (approved):
LGTM. Good job

core/src/test/resources/sql/planner.iq
```diff
@@ -128,9 +128,64 @@ EnumerableCalc(expr#0..2=[{inputs}], $f0=[$t1], $f1=[$t2])
...
++-----+
+| A   |
++-----+
+| 1.0 |
++-----+
+(1 row)
 
+!ok
+
+EnumerableCalc(expr#0=[{inputs}], expr#1=[CAST($t0):DECIMAL(11, 1)], A=[$t1])
```
mihaibudiu, line 152: so this plan wasn't generated by the new rule?

mihaibudiu:
I think we can merge this as it is.
If we fix the issue with how `set planner-rule` works this test may need to change, and it may become more relevant.

xiedeyantu:
> I think we can merge this as it is. If we fix the issue with how `set planner-rule` works this test may need to change, and it may become more relevant.

Okay, I will think about how to improve the `set planner-rules` testing.

PR #4362, opened 2025-05-06
[CALCITE-6950] Use ANY operator to check if an element exists in an a…

NobiGo:
…rray throws exception

### DIFF
--- a/core/src/main/java/org/apache/calcite/sql/SqlKind.java
+++ b/core/src/main/java/org/apache/calcite/sql/SqlKind.java
@@ -1418,6 +1418,8 @@ public enum SqlKind {
           CREATE_TYPE, DROP_TYPE,
           SET_OPTION, OTHER_DDL);
 
+  public static final EnumSet<SqlKind> COLLECTION_EXPRESSION =
+      EnumSet.of(ARRAY_VALUE_CONSTRUCTOR, MULTISET_VALUE_CONSTRUCTOR);
   /**
    * Category consisting of query node types.
    *
--- a/core/src/main/java/org/apache/calcite/sql/fun/SqlQuantifyOperator.java
+++ b/core/src/main/java/org/apache/calcite/sql/fun/SqlQuantifyOperator.java
@@ -18,10 +18,12 @@ package org.apache.calcite.sql.fun;
 
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlCallBinding;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlNodeList;
 import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.SqlWriter;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.type.SqlTypeUtil;
 import org.apache.calcite.sql.validate.SqlValidator;
@@ -99,22 +101,32 @@ public class SqlQuantifyOperator extends SqlInOperator {
       SqlValidatorScope scope, SqlCall call) {
     final SqlNode left = call.operand(0);
     final SqlNode right = call.operand(1);
-    if (right instanceof SqlNodeList && ((SqlNodeList) right).size() == 1) {
-      final RelDataType rightType =
-          validator.deriveType(scope, ((SqlNodeList) right).get(0));
-      if (SqlTypeUtil.isCollection(rightType)) {
-        final RelDataType componentRightType =
-            requireNonNull(rightType.getComponentType());
-        final RelDataType leftType = validator.deriveType(scope, left);
-        if (SqlTypeUtil.sameNamedType(componentRightType, leftType)
-            || SqlTypeUtil.isNull(leftType)
-            || SqlTypeUtil.isNull(componentRightType)) {
-          return validator.getTypeFactory().createTypeWithNullability(
-              validator.getTypeFactory().createSqlType(SqlTypeName.BOOLEAN),
-              rightType.isNullable()
-                  || componentRightType.isNullable()
-                  || leftType.isNullable());
+    if (right instanceof SqlNodeList || right.isA(SqlKind.QUERY)) {
+      return null;
+    }
+    RelDataType rightType = validator.deriveType(scope, right);
+    if (SqlTypeUtil.isCollection(rightType)) {
+      RelDataType componentRightType = requireNonNull(rightType.getComponentType());
+      RelDataType leftType = validator.deriveType(scope, left);
+      SqlCallBinding callBinding = new SqlCallBinding(validator, scope, call);
+      if (callBinding.isTypeCoercionEnabled()) {
+        boolean coerced = callBinding.getValidator().getTypeCoercion()
+            .quantifyOperationCoercion(callBinding);
+        if (coerced) {
+          // Update the node data type if we coerced any type.
+          leftType = validator.deriveType(scope, call.operand(0));
+          rightType = validator.deriveType(scope, call.operand(1));
+          componentRightType = rightType.getComponentType();
         }
+      }
+      assert componentRightType != null;
+      if (SqlTypeUtil.sameNamedType(componentRightType, leftType)
+          || SqlTypeUtil.isNull(leftType) || SqlTypeUtil.isNull(componentRightType)) {
+        return validator.getTypeFactory().createTypeWithNullability(
+            validator.getTypeFactory().createSqlType(SqlTypeName.BOOLEAN),
+            rightType.isNullable()
+                || componentRightType.isNullable()
+                || leftType.isNullable());
       }
     }
     return null;
@@ -158,6 +170,25 @@ public class SqlQuantifyOperator extends SqlInOperator {
       }
     default:
       throw new AssertionError("unexpected " + kind);
+    }
+  }
+
+  @Override public void unparse(
+      SqlWriter writer,
+      SqlCall call,
+      int leftPrec,
+      int rightPrec) {
+    assert call.operandCount() == 2;
+    call.operand(0).unparse(writer, leftPrec, rightPrec);
+    writer.setNeedWhitespace(true);
+    writer.sep(getName());
+    writer.setNeedWhitespace(true);
+    if (call.operand(1).isA(SqlKind.QUERY)) {
+      call.operand(1).unparse(writer, 0, 0);
+    } else {
+      final SqlWriter.Frame frame = writer.startList("(", ")");
+      call.operand(1).unparse(writer, 0, 0);
+      writer.endList(frame);
     }
   }
 }
--- a/core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercion.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercion.java
@@ -120,6 +120,9 @@ public interface TypeCoercion {
    */
   boolean inOperationCoercion(SqlCallBinding binding);
 
+  /** Handles type coercion for Quantify operation. */
+  boolean quantifyOperationCoercion(SqlCallBinding binding);
+
   /** Coerces operand of binary arithmetic expressions to Numeric type.*/
   boolean binaryArithmeticCoercion(SqlCallBinding binding);
 
--- a/core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
+++ b/core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
@@ -38,6 +38,7 @@ import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.type.SqlOperandMetadata;
 import org.apache.calcite.sql.type.SqlTypeFamily;
 import org.apache.calcite.sql.type.SqlTypeMappingRule;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.type.SqlTypeUtil;
 import org.apache.calcite.sql.validate.SqlValidator;
 import org.apache.calcite.sql.validate.SqlValidatorScope;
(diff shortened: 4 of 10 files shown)


core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -563,6 +564,44 @@ protected boolean booleanEquality(SqlCallBinding binding,
     return false;
   }
 
+  @Override public boolean quantifyOperationCoercion(SqlCallBinding binding) {
```
mihaibudiu, line 567: is this only when the argument is an array?
If true, this should be documented.
(The function probably needs JavaDoc anyway)

core/src/main/java/org/apache/calcite/sql/SqlKind.java
```diff
@@ -1418,6 +1418,8 @@ public enum SqlKind {
           CREATE_TYPE, DROP_TYPE,
           SET_OPTION, OTHER_DDL);
 
+  public static final EnumSet<SqlKind> COLLECTION_EXPRESSION =
```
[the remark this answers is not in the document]
NobiGo: According to https://www.postgresql.org/docs/current/functions-comparisons.html#FUNCTIONS-COMPARISONS-ANY-SOME, it should only involve ARRAY.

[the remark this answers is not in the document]
mihaibudiu: This is true if you are thinking about quantifier queries, but this set of SqlKind does not seem to be about that.
Maybe you can change the set's name.
Or maybe you don't really need this constant at all.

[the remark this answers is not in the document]
NobiGo: Delete.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -563,6 +564,76 @@ protected boolean booleanEquality(SqlCallBinding binding,
...
+   * to find the common type of LHS's type and RHS's component type.
+   * <li>If the common type differs from the LHS type, then coerced LHS type,
+   * and nullable remains unchanged.
+   * <li>Create a new Collection type that matches the common type,
+   * and nullable keep same as RHS's component type and RHS's collection type.
+   * <li>If this new Collection type differs from the RHS type, adjust the RHS type as needed.
+   *
+   *<pre>
+   * field1       ARRAY(field2, field3, field4)
+   *    |                |       |        |
```
mihaibudiu, line 586: lines are not aligned with pluses

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -563,6 +564,76 @@ protected boolean booleanEquality(SqlCallBinding binding,
...
+    final RelDataType rightWidenType =
+        binding.getTypeFactory().enforceTypeWithNullability(widenType, type2.isNullable());
+    RelDataType collectionWidenType =
+        binding.getTypeFactory().createArrayType(rightWidenType, -1);
+    collectionWidenType =
+        binding
+            .getTypeFactory()
+            .enforceTypeWithNullability(collectionWidenType, collectionType.isNullable());
+    boolean coercedRight =
+        coerceOperandType(getScope(binding), binding.getCall(), 1, collectionWidenType);
```
mihaibudiu, line 630: I hope these calls adjust the types in the typeMap and the callbinding too.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -16099,6 +16099,28 @@ void testTimestampDiff(boolean coercionEnabled) {
...
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6950">[CALCITE-6950]
+   * Use ANY operator to check if an element exists in an array throws exception</a>. */
+  @Test void testQuantifyCollectionOperators2() {
+    final SqlOperatorFixture f = fixture();
+    QUANTIFY_OPERATORS.forEach(operator -> f.setFor(operator, SqlOperatorFixture.VmName.EXPAND));
+
+    f.checkNull("1.0 = some (ARRAY[2,3,null])");
```
mihaibudiu, line 16109: why is this null?

NobiGo: Because the value of `1.0 == null` is NULL, So the value of the expression is null.

mihaibudiu: So this stops at the first match that returns true.
But if there is no match, what does it return, the last comparison, or the AND of all comparisons?
I.e., what is the output for 1 = some(array[2, NULL, 3])?

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
NobiGo: Are you saying here that we don't need to update the type of "call" again, instead update it directly within "coerceOperandType"?

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
NobiGo: ` 1 = some(array[2, NULL, 3])` will return NULL, I will update the unit test.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
NobiGo: I understand what you mean. We can discuss this in creating a new PR It's not a problem that can be quickly modified.

mihaibudiu: I want to avoid the bug as in https://issues.apache.org/jira/browse/CALCITE-6743

NobiGo: The RHS type has been updated by:
```
if (coercedRight) {
      updateInferredType(node2, collectionWidenType);
    } 
```
I'm not sure if this is the same issue, but it would be helpful to add a description of the usage issues that may arise when having inaccurate types in CALCITE-6723.
 I will also review another PR mentioned in the comments on CALCITE-6723, as there may be some duplicate code that can achieve the same capability. Currently, I am unsure.

mihaibudiu: After 6743 the program includes casts, but the validator types in the typeMap are still the old ones, before the casts were inserted.

NobiGo: I will check this.

NobiGo: This SQL:
`select 1.0 = some (ARRAY[2,3,null]);`
The nodeToTypeMap is:
Screen Shot 2025-05-11 at 08 45 12

Our question is Which of the following three are legal？

1: ARRAY [CAST (2 AS INTEGER), CAST (3 AS INTEGER), NULL] ->DECIMAL (11,1) ARRAY
2: CAST(ARRAY[CAST(2 AS INTEGER), CAST(3 AS INTEGER), NULL] AS DECIMAL(11, 1) ARRAY) -> DECIMAL (11,1) ARRAY
3: ARRAY [CAST(CAST (2 AS INTEGER) as DECIMAL(11,1)), CAST(CAST (3 AS INTEGER) as  DECIMAL (11,1)), CAST(NULL as DECIMAL(11,1))]->DECIMAL (11,1) ARRAY

mihaibudiu: which one is "legal" depends on the dialect that the function belongs to; the existing behavior attempted to model Spark.
But the main requirement is for the nodes that end up in the Rel tree to all be present in the nodeToTypeMap.

NobiGo: According to the screenshot above, does the current PR also have the same issue? The node `ARRAY [CAST (2 AS INTEGER), CAST (3 AS INTEGER), NULL] ->DECIMAL (11,1) ARRAY` has been updated.

NobiGo: @mihaibudiu Does this discussion affect the merging of the current issues?

mihaibudiu: I was assuming this pr is fine and that you have checked

NobiGo: According to my understanding, I have checked it.

core/src/main/java/org/apache/calcite/sql/fun/SqlQuantifyOperator.java
```diff
@@ -160,4 +172,23 @@ public class SqlQuantifyOperator extends SqlInOperator {
       throw new AssertionError("unexpected " + kind);
     }
   }
+
+  @Override public void unparse(
```
mihaibudiu, line 176: is there a test for this new function?

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercion.java
```diff
@@ -120,6 +120,12 @@ boolean rowTypeCoercion(
    */
   boolean inOperationCoercion(SqlCallBinding binding);
 
+  /** Handles type coercion for Quantify operation {@link org.apache.calcite.sql.fun.SqlQuantifyOperator}.
```
mihaibudiu, line 123: Since there are multiple types of quantify operations, I would say "operations" here, and perhaps give an example `SOME(ARRAY [...])` to make it easier to understand.

Does this handle both cases for subqueries and collection constructors? If so, it should be mentioned here as well.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -563,6 +564,78 @@ protected boolean booleanEquality(SqlCallBinding binding,
...
+   * <p>STRATEGIES
+   *
+   * <p>To determine the common type:
+   *
+   * <ul>
+   *
+   * <li>When the LHS has a Simple type and RHS has a Collection type determined by {@link SqlTypeUtil#isCollection},
+   * to find the common type of LHS's type and RHS's component type.
+   * <li>If the common type differs from the LHS type, then coerced LHS type,
+   * and nullable remains unchanged.
```
mihaibudiu, line 579: what is "nullable"? It appears here for the first time.

mihaibudiu: it's probably the nullability of the result type.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -563,6 +564,78 @@ protected boolean booleanEquality(SqlCallBinding binding,
...
+   * <li>If this new Collection type differs from the RHS type, adjust the RHS type as needed.
+   *
+   *<pre>
+   * field1       ARRAY(field2, field3, field4)
+   *    |                |       |       |
+   *    |                +-------+-------+
+   *    |                        |
+   *    |                  component type
+   *    |                        |
+   *    +------common type-------+
```
mihaibudiu, line 591: So in general you need to insert a cast on both sides; on the right hand side the cast can be to a different array type.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -563,6 +564,78 @@ protected boolean booleanEquality(SqlCallBinding binding,
...
+   *</pre>
+   *
+   * <li>Notice: If either LHS or RHS has a {@link SqlTypeName#NULL} type, it will directly return without any adjusting.
```
mihaibudiu, line 594: why not insert casts in this case too?

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
```diff
@@ -563,6 +564,78 @@ protected boolean booleanEquality(SqlCallBinding binding,
...
+    final SqlCall sqlCall = binding.getCall();
+    final SqlValidatorScope scope = binding.getScope();
+    final SqlNode node1 = binding.operand(0);
+    final SqlNode node2 = binding.operand(1);
+    RelDataType widenType = commonTypeForBinaryComparison(type1, type2);
+    if (widenType == null) {
+      widenType = getTightestCommonType(type1, type2);
+    }
+    if (widenType == null) {
+      return false;
```
mihaibudiu, line 614: what does "false" mean here?
it looks to me like an error.
but below "false" seems to mean "no change".
Where is an error reported if coercion cannot bring the types to a common type?

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
mihaibudiu: can you document the return result?

core/src/main/java/org/apache/calcite/sql/fun/SqlQuantifyOperator.java
NobiGo: 1. `testAllListWithOneElement` 
2. `SqlOperatorUnparseTest#testQuantifyCollectionOperators2`
These two can cover this test case.

NobiGo: Changing the Parser.jj parsing will complicate this issue, so I have reverted this part of the modification. The corrections will be made in a new PR later.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercion.java
NobiGo: Added.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
NobiGo: I made some revisions.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
NobiGo: Added.

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
NobiGo: Remove it. The original code logic here has already covered it. If the LHS or RHS 's type is null, just return .

core/src/main/java/org/apache/calcite/sql/validate/implicit/TypeCoercionImpl.java
NobiGo: `False` means will throw compatible types exception.

testkit/src/main/java/org/apache/calcite/test/SqlOperatorTest.java
```diff
@@ -16145,6 +16145,41 @@ void testTimestampDiff(boolean coercionEnabled) {
...
+    f.checkNull("1.0 = some (ARRAY[2,null,3])");
+    f.checkBoolean("1.0 = some (ARRAY[1,2,null])", true);
+    f.checkBoolean("3.0 = some (ARRAY[1,2])", false);
```
mihaibudiu, line 16163: how do these tests verify that the coercion happened?
we don't know what the inserted casts are.
maybe a test that unparses the query would be more useful

NobiGo: Added.

mihaibudiu: I meant that if you add such tests to SqlToRelConverterTest, the xml file will show the casts explicitly inserted.
Alternatively, you can add such tests to a test fixture which unparses and then (e.g., RelToSqlConverterTest) we can again see also the explicit casts inserted. 
Sorry for being unclear, but I think it is good to be explicit about the checking; here you want to really check that the proper casts are inserted, not only that the query is accepted.

NobiGo: It seems that we have discovered a new issue CALCITE-7056.

mihaibudiu: sorry about the extra complications...

mihaibudiu: maybe showing the plan is simpler

NobiGo: I have added it in `SqlToRelConverterTest`.  
In RelToSqlConverterTest, an exception will be thrown, regardless of whether it is a simple one or not.

NobiGo: The test cases always ensure the accuracy of the PR, so I understand your concern.

NobiGo: If you think there is a risk, I can first resolve CALCITE-7056 and then proceed to review this PR.

mihaibudiu (approved):

PR #4363, opened 2025-05-07
[CALCITE-7001] Cast of malformed literal to TIMESTAMP WITH LOCAL TIME  ZONE need to throw informative error

zstan:
honestly, this PR looks like a bit hack as for me, but we can start with it and simplify all related stuff in future

### DIFF
--- a/core/src/main/java/org/apache/calcite/util/TimestampWithTimeZoneString.java
+++ b/core/src/main/java/org/apache/calcite/util/TimestampWithTimeZoneString.java
@@ -22,12 +22,17 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
-import java.util.Locale;
 import java.util.TimeZone;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import static org.apache.calcite.avatica.util.DateTimeUtils.DATE_FORMAT_STRING;
+import static org.apache.calcite.avatica.util.DateTimeUtils.TIMESTAMP_FORMAT_STRING;
+import static org.apache.calcite.util.DateTimeStringUtils.getDateFormatter;
+import static org.apache.calcite.util.Static.RESOURCE;
+
 import static java.lang.Math.floorMod;
+import static java.util.Objects.requireNonNull;
 
 /**
  * Timestamp with time-zone literal.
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/calcite/util/TimestampWithTimeZoneString.java
```diff
@@ -202,4 +212,17 @@ public TimestampString getLocalTimestampString() {
...
     return timeZone;
   }
+
+  private static DateTimeUtils.PrecisionTime parseDateTime(String tsStr, TimeZone timeZone) {
+    DateTimeUtils.PrecisionTime pt =
+        DateTimeUtils.parsePrecisionDateTimeLiteral(tsStr, requireNonNull(TIMESTAMP_FORMAT.get()),
+            timeZone, -1);
+
+    if (pt == null) {
+      throw RESOURCE.illegalLiteral("TIMESTAMP WITH LOCAL TIME ZONE", tsStr,
```
mihaibudiu, line 222: it is very useful for users to be able to see the illegal timestamp, can it be included in the message?

mihaibudiu (approved):

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
```diff
@@ -1836,6 +1838,44 @@ private void thereAndBack(byte[] bytes) {
...
+
+    ret = toTimestampWithLocalTimeZone("1970-01-01 00:00:01 "
+        + TimeZone.getTimeZone("UTC").getID());
+    assertThat(ret, is(1000L));
+
+    // exceptional scenarios
+    try {
+      ret = toTimestampWithLocalTimeZone("malformed", TimeZone.getDefault());
+      fail("expected error, got " + ret);
+    } catch (RuntimeException e) {
```
xiedeyantu, line 1860: It may be better to change it to `CalciteException` here.

core/src/main/java/org/apache/calcite/util/TimestampWithTimeZoneString.java
```diff
@@ -41,21 +46,39 @@ public class TimestampWithTimeZoneString
   final TimestampString localDateTime;
   final TimeZone timeZone;
   final String v;
+  private final DateTimeUtils.PrecisionTime pt;
+
+  private static final ThreadLocal<@Nullable SimpleDateFormat> TIMESTAMP_FORMAT =
```
xuzifu666, line 51: We can confirm TimestampWithTimeZoneString##constructor would be call by multiple thread?  If not we may not need to use ThreadLocal.

core/src/test/java/org/apache/calcite/test/SqlFunctionsTest.java
xuzifu666: +1

zstan: done

core/src/main/java/org/apache/calcite/util/TimestampWithTimeZoneString.java
zstan: It already done, i fix a tests

core/src/main/java/org/apache/calcite/util/TimestampWithTimeZoneString.java
zstan: It not easy, also if it thread safety right now we have no guarantee that it remain same in future.
DateFormat is not thread safe (check java doc) and checker framework reports about erroneous usage without thread local mechanics. Thus i forced to use such a code, if you have different approach here - plz let me know. Thanks !

xiedeyantu (approved):
LGTM

xuzifu666 (approved):

mihaibudiu (approved):
I think you can squash the commits

zstan:
@mihaibudiu squashed

PR #4364, opened 2025-05-08
[CALCITE-6994] Enhance RexBuilder#makeIn to create SEARCH for MULTISET literals

xiedeyantu:
Jira: [CALCITE-6994](https://issues.apache.org/jira/browse/CALCITE-6994)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rex/RexBuilder.java
+++ b/core/src/main/java/org/apache/calcite/rex/RexBuilder.java
@@ -28,6 +28,7 @@ import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rel.type.RelDataTypeSystemImpl;
 import org.apache.calcite.runtime.FlatLists;
+import org.apache.calcite.runtime.FlatLists.AbstractFlatList;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlCollation;
 import org.apache.calcite.sql.SqlIntervalQualifier;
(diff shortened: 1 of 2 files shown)


mihaibudiu:
I requested a review from @NobiGo since he filed the issue.

xiedeyantu:
> I requested a review from @NobiGo since he filed the issue.

Thanks!

core/src/test/java/org/apache/calcite/rex/RexBuilderTest.java
```diff
@@ -981,6 +981,34 @@ inCall, hasToString("SEARCH($0, Sarg["
...
+            + "[100:INTEGER, 200:INTEGER]:INTEGER NOT NULL MULTISET, "
+            + "[300:INTEGER, 400:INTEGER]:INTEGER NOT NULL MULTISET"
+            + "]:INTEGER NOT NULL MULTISET)";
+
+    RexNode m1 = rexBuilder.makeLiteral(ImmutableList.of(100, 200), multisetType, true);
+    RexNode m2 = rexBuilder.makeLiteral(ImmutableList.of(300, 400), multisetType, true);
+    RexNode inCall1 = rexBuilder.makeIn(column, ImmutableList.of(m1, m2));
+    assertThat(inCall1, hasToString(expected));
+
+    RexNode m3 = rexBuilder.makeLiteral(ImmutableList.of(100, 200), multisetType, false);
```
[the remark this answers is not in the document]
xiedeyantu: > The third parameter 'allowCast' is not reflected in the test case.

The last parameter 'false' is 'allowCast'.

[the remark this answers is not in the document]
xiedeyantu: Thank you for your review. I have added test cases for `allowCast`. Could you please help me review them again.

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1871,10 +1872,28 @@ && areAssignable(arg, Arrays.asList(lower, upper))) {
     switch (point.getKind()) {
     case LITERAL:
       final RexLiteral literal = (RexLiteral) point;
+      if (literal.getValue() instanceof AbstractFlatList) {
```
xiedeyantu, line 1875: Hi @NobiGo I'm not sure if this modification is appropriate. Can you help me to have a look?

NobiGo: Can't we use a list here?
 If it is an AbstractFlatList, does it restrict users from using an ImmutableIntList to build it?
 Can only include the integer value?

core/src/main/java/org/apache/calcite/rex/RexBuilder.java
```diff
@@ -1871,10 +1872,28 @@ && areAssignable(arg, Arrays.asList(lower, upper))) {
...
+        AbstractFlatList list = (AbstractFlatList) literal.getValue();
```
NobiGo, line 1876: `literal.getValueAs` will remove the cast.

xiedeyantu: When constructing MultiSet, there is this logic that uses `FlatLists.of`, which stores `RexLiteral`. So there needs to be an additional layer of processing here, but it is not limited to `Int`.
```
      if (allowCast) {
        return makeCall(SqlStdOperatorTable.MULTISET_VALUE, operands);
      } else {
        return new RexLiteral((Comparable) FlatLists.of(operands), type,
            sqlTypeName);
      }
```

xiedeyantu: I found using `literal.getValueAs` will return @Nullable. Additional processing may be required. So I changed it back again. Is that OK?

mihaibudiu:
Is this ready?

xiedeyantu:
> Is this ready?

I think it is almost ready. This is the implementation I can think of so far.

xiedeyantu:
This modification approach doesn't seem very elegant and has been temporarily closed.

PR #4366, opened 2025-05-09
[CALCITE-7002] Create an optimization rule to eliminate UNION from the same source with different filters

xiedeyantu:
Jira link: [CALCITE-7002](https://issues.apache.org/jira/browse/CALCITE-7002)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
@@ -388,7 +388,21 @@ public class CoreRules {
   public static final IntersectToSemiJoinRule INTERSECT_TO_SEMI_JOIN =
           IntersectToSemiJoinRule.Config.DEFAULT.toRule();
 
+  /** Rule that translates a {@link Union} to {@link Filter}. */
+  public static final SetOpToFilterRule UNION_FILTER_TO_FILTER =
+      SetOpToFilterRule.Config.UNION.toRule();
+
+  /** Rule that translates a {@link Intersect} to {@link Filter}. */
+  public static final SetOpToFilterRule INTERSECT_FILTER_TO_FILTER =
+      SetOpToFilterRule.Config.INTERSECT.toRule();
+
   /** Rule that translates a {@link Minus} to {@link Filter}. */
+  public static final SetOpToFilterRule MINUS_FILTER_TO_FILTER =
+      SetOpToFilterRule.Config.MINUS.toRule();
+
+  /** Rule that sames as {@link CoreRules#MINUS_FILTER_TO_FILTER},
+   * But it is deprecated. */
+  @Deprecated
   public static final MinusToFilterRule MINUS_TO_FILTER =
       MinusToFilterRule.Config.DEFAULT.toRule();
 
--- a/core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
@@ -51,7 +51,10 @@ public class MinusToFilterRule
     extends RelRule<MinusToFilterRule.Config>
     implements TransformationRule {
 
-  /** Creates an MinusToFilterRule. */
+  /**
+   * This rule is replaced by {@link SetOpToFilterRule}.
+   * Please see {@link CoreRules#MINUS_FILTER_TO_FILTER} */
+  @Deprecated
   protected MinusToFilterRule(Config config) {
     super(config);
   }
--- a/core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
@@ -1,0 +1,287 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Filter;
+import org.apache.calcite.rel.core.Intersect;
+import org.apache.calcite.rel.core.Minus;
+import org.apache.calcite.rel.core.SetOp;
+import org.apache.calcite.rel.core.Union;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.tools.RelBuilder;
+import org.apache.calcite.util.Pair;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.immutables.value.Value;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Rule that replaces {@link SetOp} operator with {@link Filter}
+ * when both inputs are from the same source with only filter conditions differing.
+ * For nested filters, the rule {@link CoreRules#FILTER_MERGE}
+ * should be used prior to invoking this one.
+ *
+ * <p>Example:
+ *
+ * <p>UNION
+ * <blockquote><pre>
+ * SELECT mgr, comm FROM emp WHERE mgr = 12
+ * UNION
+ * SELECT mgr, comm FROM emp WHERE comm = 5
+ *
+ * is rewritten to
+ *
+ * SELECT DISTINCT mgr, comm FROM emp
+ * WHERE mgr = 12 OR comm = 5
+ * </pre></blockquote>
+ *
+ * <p>UNION with multiple inputs
+ * <blockquote><pre>
+ * SELECT deptno FROM emp WHERE deptno = 12
+ * UNION
+ * SELECT deptno FROM dept WHERE deptno = 5
+ * UNION
+ * SELECT deptno FROM emp WHERE deptno = 6
+ * UNION
+ * SELECT deptno FROM dept WHERE deptno = 10
+ *
+ * is rewritten to
+ *
+ * SELECT deptno FROM emp WHERE deptno = 12 OR deptno = 6
+ * UNION
+ * SELECT deptno FROM dept WHERE deptno = 5 OR deptno = 10
+ * </pre></blockquote>
+ *
+ * <p>INTERSECT
+ * <blockquote><pre>
+ * SELECT mgr, comm FROM emp WHERE mgr = 12
+ * INTERSECT
+ * SELECT mgr, comm FROM emp WHERE comm = 5
+ *
+ * is rewritten to
+ *
+ * SELECT DISTINCT mgr, comm FROM emp
+ * WHERE mgr = 12 AND comm = 5
+ * </pre></blockquote>
+ *
+ * <p>EXCEPT
+ * <blockquote><pre>
+ * SELECT mgr, comm FROM emp WHERE mgr = 12
+ * EXCEPT
+ * SELECT mgr, comm FROM emp WHERE comm = 5
+ *
+ * is rewritten to
+ *
+ * SELECT DISTINCT mgr, comm FROM emp
+ * WHERE mgr = 12 AND NOT(comm = 5)
+ * </pre></blockquote>
+ */
+@Value.Enclosing
+public class SetOpToFilterRule
+    extends RelRule<SetOpToFilterRule.Config>
+    implements TransformationRule {
+
+  /** Creates an SetOpToFilterRule. */
+  protected SetOpToFilterRule(Config config) {
+    super(config);
+  }
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    config.matchHandler().accept(this, call);
+  }
+
+  private void match(RelOptRuleCall call) {
+    final SetOp setOp = call.rel(0);
+    final List<RelNode> inputs = setOp.getInputs();
+    if (setOp.all || inputs.size() < 2) {
+      return;
+    }
+
+    final RelBuilder builder = call.builder();
+    Pair<RelNode, @Nullable RexNode> first = extractSourceAndCond(inputs.get(0).stripped());
+
+    // Groups conditions by their source relational node and input position.
+    // - Key: Pair of (sourceRelNode, inputPosition)
+    //   - inputPosition is null for mergeable conditions
+    //   - inputPosition contains original index for non-mergeable inputs
+    // - Value: List of conditions
+    //
+    // For invalid conditions (non-deterministic expressions or containing subqueries),
+    // positions are tagged with their input indices to skip unmergeable inputs
+    // during map-based grouping. Other positions are set to null.
+    Map<Pair<RelNode, @Nullable Integer>, List<@Nullable RexNode>> sourceToConds =
+        new LinkedHashMap<>();
+
+    RelNode firstSource = first.left;
+    sourceToConds.computeIfAbsent(Pair.of(firstSource, null),
+        k -> new ArrayList<>()).add(first.right);
+
+    for (int i = 1; i < inputs.size(); i++) {
+      final RelNode input = inputs.get(i).stripped();
+      final Pair<RelNode, @Nullable RexNode> pair = extractSourceAndCond(input);
+      sourceToConds.computeIfAbsent(Pair.of(pair.left, pair.right != null ? null : i),
+          k -> new ArrayList<>()).add(pair.right);
+    }
+
+    if (sourceToConds.size() == inputs.size()) {
+      return;
+    }
+
+    int branchCount = 0;
+    for (Map.Entry<Pair<RelNode, @Nullable Integer>, List<@Nullable RexNode>> entry
+        : sourceToConds.entrySet()) {
+      Pair<RelNode, @Nullable Integer> left = entry.getKey();
+      List<@Nullable RexNode> conds = entry.getValue();
+      // Single null condition indicates pass-through branch,
+      // directly add its corresponding input to the new inputs list.
+      if (conds.size() == 1 && conds.get(0) == null) {
+        builder.push(left.left);
+        branchCount++;
+        continue;
+      }
+
+      List<RexNode> condsNonNull = conds.stream().map(e -> {
+        assert e != null;
+        return e;
+      }).collect(Collectors.toList());
+
+      RexNode combinedCond =
+          combineConditions(builder, condsNonNull, setOp, left.left == firstSource);
+
+      builder.push(left.left)
+          .filter(combinedCond);
+      branchCount++;
+    }
+
+    // RelBuilder will not create 1-input SetOp
+    // and remove the distinct after a SetOp
+    buildSetOp(builder, branchCount, setOp)
+        .distinct();
+    call.transformTo(builder.build());
+  }
+
+  private static RelBuilder buildSetOp(RelBuilder builder, int count, RelNode setOp) {
+    if (setOp instanceof Union) {
+      return builder.union(false, count);
+    } else if (setOp instanceof Intersect) {
+      return builder.intersect(false, count);
+    } else if (setOp instanceof Minus) {
+      return builder.minus(false, count);
+    }
+    // unreachable
+    throw new IllegalStateException("unreachable code");
+  }
+
+  private static Pair<RelNode, @Nullable RexNode> extractSourceAndCond(RelNode input) {
+    if (input instanceof Filter) {
+      Filter filter = (Filter) input;
+      if (!RexUtil.isDeterministic(filter.getCondition())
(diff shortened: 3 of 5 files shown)


core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
```diff
@@ -0,0 +1,188 @@
...
+  private void match(RelOptRuleCall call) {
+    final SetOp setOp = call.rel(0);
+    if (setOp.all || setOp.getInputs().size() != 2) {
+      return;
+    }
+    final RelBuilder builder = call.builder();
+    final RelNode leftInput = call.rel(1);
+    final Filter rightInput = call.rel(2);
+
+    if (!RexUtil.isDeterministic(rightInput.getCondition())) {
```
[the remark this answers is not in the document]
xiedeyantu: Thank you for your review, I think this is a great question. My idea is that if both `inputs` have `comm=rand()`, merging filters may result in a composite `comm=rand()`, which may have an impact on the outcome. Another reason is that if there is only one `rand()` in the inputs, it may be pushed forward by other rules in the future and affect the results. So I want to be more rigorous here. If there are any issues, please help to correct them.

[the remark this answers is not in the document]
mihaibudiu: Yes, I think many other rules may be broken because they copy liberally expressions without checking whether the expressions are deterministic.

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
```diff
@@ -388,9 +388,17 @@ private CoreRules() {}
   public static final IntersectToSemiJoinRule INTERSECT_TO_SEMI_JOIN =
           IntersectToSemiJoinRule.Config.DEFAULT.toRule();
 
+  /** Rule that translates a {@link Union} to {@link Filter}. */
+  public static final SetOpToFilterRule UNION_TO_FILTER =
```
mihaibudiu, line 392: A better name would have been `UNION_FILTER_TO_FILTER`, but because we already have a rule for `MINUS`, we should keep this name, which has the same structure.

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
```diff
@@ -1,121 +0,0 @@
...
- * SELECT mgr, comm FROM emp WHERE comm = 5
- *
- * to
- *
- * SELECT DISTINCT mgr, comm FROM emp
- * WHERE mgr = 12 AND NOT(comm = 5)
- * </pre></blockquote>
- */
-@Value.Enclosing
-public class MinusToFilterRule
```
mihaibudiu, line 50: I don't think you can delete a public class.
Just mark it `@deprecated` and point to the replacement.
I know, this is annoying.

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
```diff
@@ -0,0 +1,188 @@
...
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.tools.RelBuilder;
+
+import org.immutables.value.Value;
+
+/**
+ * Rule that replaces {@link SetOp} operator with {@link Filter}
+ * when both inputs are from the same source with only filter conditions differing.
+ * Only inspect a single {@link Filter} layer in the inputs of {@link SetOp}.
```
mihaibudiu, line 36: I don't know what "inspect" means here, but I think this phrase is not necessary.

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
```diff
@@ -0,0 +1,188 @@
...
+ * For inputs with nested {@link Filter}s, apply {@link CoreRules#FILTER_MERGE}
```
mihaibudiu, line 37: who should apply? 
If this rule applies, say "this rule applies"
if the user is supposed to apply, say "for nested filters, the rule ... should be used prior to invoking this one".

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * <p>Example:
+ *
+ * <p>UNION
+ * <blockquote><pre>
+ * SELECT mgr, comm FROM emp WHERE mgr = 12
+ * UNION
+ * SELECT mgr, comm FROM emp WHERE comm = 5
+ *
+ * to
```
mihaibudiu, line 48: is rewritten to

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
```diff
@@ -0,0 +1,188 @@
...
+
+  //~ Methods ----------------------------------------------------------------
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    config.matchHandler().accept(this, call);
+  }
+
+  private void match(RelOptRuleCall call) {
+    final SetOp setOp = call.rel(0);
+    if (setOp.all || setOp.getInputs().size() != 2) {
```
mihaibudiu, line 96: I think this would work with more than 2 as well.

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
```diff
@@ -0,0 +1,188 @@
...
+    } else if (setOp instanceof Intersect) {
+      finalCond = leftCond != null
+          ? builder.and(leftCond, rightInput.getCondition())
+          : rightInput.getCondition();
+    } else if (setOp instanceof Minus) {
+      finalCond = leftCond != null
+          ? builder.and(leftCond, builder.not(rightInput.getCondition()))
+          : builder.not(rightInput.getCondition());
+    } else {
+      // Should not enter here
```
mihaibudiu, line 137: throw some exception that says "unreachable".
maybe there are other examples you can copy.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -10515,6 +10515,115 @@ private void checkLoptOptimizeJoinRule(LoptOptimizeJoinRule rule) {
...
+  /** Test case of
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-7002">[CALCITE-7002]
+   * Create an optimization rule to eliminate UNION
+   * from the same source with different filters</a>. */
+  @Test void testIntersectToFilterRule3() {
+    final String sql = "SELECT mgr, comm FROM emp WHERE comm = 5\n"
+        + "INTERSECT\n"
+        + "SELECT e1.mgr, e1.comm\n"
+        + "FROM emp e1\n"
+        + "WHERE EXISTS (\n"
```
mihaibudiu, line 10618: these correlated queries are more complicated than really necessary to test this rule.
why are you using them?

mihaibudiu (approved):

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xiedeyantu: I remember you mentioned in jira that it should be applicable to correlated subqueries, so I added a case. If it is not needed, I can delete it (maybe it is really not necessary to add it here).

mihaibudiu: What i meant is maybe the rule does not apply when a predicate contains a query

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
xiedeyantu: Although the name of this implementation class has been changed (to SetOpToFilterRule), the logic of MINUS_TO_FILTER has not been changed and the name is still MINUS_TO_FILTER. Can I continue to keep UNION and INTERSECT consistent with it (UNOIN_TO_FILTER and INTERSECT_TO_FILTER)?

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
xiedeyantu: Oh, I suddenly understand what you mean. I don't want to delete the MinusToFilterRule class, but mark it as obsolete. Then I don't want to delete MINUS_TO_FILTER in CoreRules. The new SetOpToFilterRule is completely new. So the name here should be written as MINUS_FILTERTO_FILTER, and the others are similar. Do I understand it correctly now?

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xiedeyantu: Got it, then I will prohibit subqueries.

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
xiedeyantu: Yes, it can support multiple inputs. But I am worried about a case, if input1 and input2 come from the same source, but input3 does not, do I need to handle this complex scenario? Maybe if I don't do SETOP_MERGE, there will be no such problem, and I don't need to support multiple inputs. I don't have a good idea yet.

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
xiedeyantu: Sorry, please ignore this reply

NobiGo (approved):
LGTM

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
mihaibudiu: That sounds like a good approach, but I would have a _ before the TO.

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
mihaibudiu: Indeed, if you have an N-ary operation in the most general case you can "group" the inputs (in a `List<List<RelOp>>`) and for each such list make a single WHERE, and the result is a setop between all the lists

core/src/main/java/org/apache/calcite/rel/rules/MinusToFilterRule.java
xiedeyantu: Sorry, it was my writing mistake. It should be MINUS_FILTER_TO_FILTER.

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
xiedeyantu: This is a very good idea! Do I need to implement this in this PR as well?

mihaibudiu: Frankly, I don't think it will be much more work.
But you can do it in two PRs if you want.

mihaibudiu:
If you don't plan to generalize this, let me know and we'll merge it.
Otherwise the new commit may also make these failing tests pass, they seem to be network-related.

xiedeyantu:
> If you don't plan to generalize this, let me know and we'll merge it. Otherwise the new commit may also make these failing tests pass, they seem to be network-related.

Please wait me to implement your idea.

xiedeyantu:
@mihaibudiu I have implemented the n-ary logic, please help me check if it meets expectations.

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
```diff
@@ -0,0 +1,241 @@
...
+ * <blockquote><pre>
+ * SELECT mgr, comm FROM emp WHERE mgr = 12
+ * EXCEPT
+ * SELECT mgr, comm FROM emp WHERE comm = 5
+ *
+ * is rewritten to
+ *
+ * SELECT DISTINCT mgr, comm FROM emp
+ * WHERE mgr = 12 AND NOT(comm = 5)
+ * </pre></blockquote>
```
mihaibudiu, line 82: you should also describe what happens if there are multiple inputs here.
only one setop will be sufficient for that case.

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
```diff
@@ -0,0 +1,241 @@
...
+    final SetOp setOp = call.rel(0);
+    final List<RelNode> inputs = setOp.getInputs();
+    if (setOp.all || inputs.size() < 2) {
+      return;
+    }
+
+    final RelBuilder builder = call.builder();
+    Pair<RelNode, RexNode> first = extractSourceAndCond(inputs.get(0).stripped());
+    // Map to store conditions grouped by <sourceRelNode, inputPosition>.
+    // Valid cond positions are set to null and others are tagged with
```
mihaibudiu, line 110: what makes a position "invalid?"
you haven't explained yet

mihaibudiu: the checker is unhappy about your missing annotations

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
```diff
@@ -0,0 +1,241 @@
...
+
+      RexNode combinedCond =
+          combineConditions(builder, conds, setOp, left.left == firstSource);
+
+      builder.push(left.left)
+          .filter(combinedCond);
+      branchCount++;
+    }
+
+    buildSetOp(builder, branchCount, setOp)
```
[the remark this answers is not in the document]
mihaibudiu: you should not do the setop if there is only one input left

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
```diff
@@ -0,0 +1,241 @@
...
+        // just return current input and null cond.
+        return Pair.of(input, null);
+      }
+      return Pair.of(filter.getInput().stripped(), filter.getCondition());
+    }
+    return Pair.of(input.stripped(),
+        input.getCluster().getRexBuilder().makeLiteral(true));
+  }
+
+  private static RexNode andFirstNotRest(RelBuilder builder, List<RexNode> conds) {
```
mihaibudiu, line 183: a short comment would help, although the function is pretty clear
builds `conds.get(0) && !conds.get(1) && ... `

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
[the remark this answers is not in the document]
mihaibudiu: but maybe the builder takes care of both of these?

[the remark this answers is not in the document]
xiedeyantu: Yes, the builder can help me to do this.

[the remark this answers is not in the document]
xiedeyantu: I have Completed the modification of these comments.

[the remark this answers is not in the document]
xiedeyantu: Please wait for me to fix the ci error.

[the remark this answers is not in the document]
xiedeyantu: @mihaibudiu I have fixed CI errors. Can you please help me to have a look?

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
```diff
@@ -388,7 +388,20 @@ private CoreRules() {}
...
+  /** Rule that translates a {@link Intersect} to {@link Filter}. */
+  public static final SetOpToFilterRule INTERSECT_FILTER_TO_FILTER =
+      SetOpToFilterRule.Config.INTERSECT.toRule();
+
   /** Rule that translates a {@link Minus} to {@link Filter}. */
+  public static final SetOpToFilterRule MINUS_FILTER_TO_FILTER =
+      SetOpToFilterRule.Config.MINUS.toRule();
+
+  /** Rule that sames as {@link CoreRules#MINUS_FILTER_TO_FILTER},
+   * But it is deprecated. */
```
NobiGo, line 404: Use ` @Deprecated`

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
```diff
@@ -0,0 +1,285 @@
...
+ * UNION
+ * SELECT deptno FROM emp WHERE deptno = 6
+ * UNION
+ * SELECT deptno FROM dept WHERE deptno = 10
+ *
+ * is rewritten to
+ *
+ * SELECT deptno FROM emp WHERE deptno = 12 OR deptno = 5
+ * UNION
+ * SELECT deptno FROM emp WHERE deptno = 5 OR deptno = 10
```
NobiGo, line 75: `deptno = 6`  OR `dept`?

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -9173,6 +9266,131 @@ LogicalAggregate(group=[{0, 1}])
       <![CDATA[SELECT mgr, comm FROM emp
 EXCEPT
 SELECT mgr, comm FROM emp WHERE comm = 5
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testMinusToFilterRule3">
```
NobiGo, line 9272: Why `Minus`?

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -9173,6 +9266,131 @@ LogicalAggregate(group=[{0, 1}])
...
+      LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
+]]>
+    </Resource>
+    <Resource name="planAfter">
+      <![CDATA[
+LogicalMinus(all=[false])
+  LogicalFilter(condition=[SEARCH($0, Sarg[(6..12]])])
+    LogicalProject(DEPTNO=[$7])
+      LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+  LogicalFilter(condition=[>($0, 8)])
```
NobiGo, line 9306: `condition=[>($0, 8)]` this condition is right?

xiedeyantu: Original SQL:
```
SELECT deptno FROM emp WHERE deptno > 6
EXCEPT
SELECT deptno FROM dept WHERE deptno > 8
EXCEPT
SELECT deptno FROM emp WHERE deptno > 12
EXCEPT
SELECT deptno FROM dept WHERE deptno > 10
```
Firstly can be convert to:
```
SELECT deptno FROM emp WHERE deptno > 6 and deptno <= 12
EXCEPT
(
SELECT deptno FROM dept WHERE deptno > 8
UNION
SELECT deptno FROM dept WHERE deptno > 10
)
```
Secondly can be convert to:
```
SELECT deptno FROM emp WHERE deptno > 6 and deptno <= 12
EXCEPT
SELECT deptno FROM dept WHERE deptno > 8
```
So it should be consistent with the current plan.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -10511,7 +10511,157 @@ private void checkLoptOptimizeJoinRule(LoptOptimizeJoinRule rule) {
...
-        .withRule(CoreRules.MINUS_TO_FILTER)
+        .withRule(CoreRules.MINUS_FILTER_TO_FILTER)
+        .check();
+  }
+
+  /** Test case of
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-7002">[CALCITE-7002]
+   * Create an optimization rule to eliminate UNION
+   * from the same source with different filters</a>. */
+  @Test void testMinusToFilterRule3() {
```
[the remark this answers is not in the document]
xiedeyantu: Done!

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -10492,7 +10492,7 @@ private void checkLoptOptimizeJoinRule(LoptOptimizeJoinRule rule) {
   /** Test case of
    * <a href="https://issues.apache.org/jira/browse/CALCITE-6973">[CALCITE-6973]
    * Add rule for convert Minus to Filter</a>. */
-  @Test void testMinusToFilterRule() {
+  @Test void testMinusToFilterRuleWithTwoFilter() {
```
asolimando, line 10495:
```suggestion
  @Test void testMinusToFilterRuleWithTwoFilters() {
```

asolimando (commented):
Can you check sonar issues? Some seem legitimate

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xiedeyantu: Done!

xiedeyantu: The local test passes, but pr-head always fails. I will push the code later. Sorry.

xiedeyantu: Thanks, I saw this error. But it can indeed pass through locally. I agree with your idea that we cannot merge the PR of CI. Give me some time, and I will find a way to get CI approved.

asolimando: It's very weird that it passes locally and fails in CI, it would help tracking when this started to help figure out the source of the issue.

mihaibudiu (approved):

mihaibudiu:
I think you can squash the commits

core/src/main/java/org/apache/calcite/rel/rules/SetOpToFilterRule.java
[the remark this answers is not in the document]
mihaibudiu: can you also add a comment about the fact that the builder will not create 1-input setop and remove the distinct after a setop?

PR #4368, opened 2025-05-10
[CALCITE-7006] Incorrect left join results with IS NOT DISTINCT FROM under specific plan

xiedeyantu:
JIRA link: [CALCITE-7006](https://issues.apache.org/jira/browse/CALCITE-7006)

### DIFF
--- a/core/src/test/resources/sql/join.iq
+++ b/core/src/test/resources/sql/join.iq
@@ -854,4 +854,154 @@ EnumerableCalc(expr#0..1=[{inputs}], A=[$t0])
     EnumerableValues(tuples=[[{ 1 }, { null }]])
 !plan
 
+# Test for Double
+SELECT t1.a
+FROM (VALUES (1.0e0), (2.0e0), (null)) AS t1(a)
+JOIN (VALUES (1.0e0), (null)) AS t2(a)
+ON t1.a IS NOT DISTINCT FROM t2.a;
++-----+
+| A   |
++-----+
+| 1.0 |
+|     |
++-----+
+(2 rows)
+
+!ok
+
+# Test for Int
+SELECT t1.a
+FROM (VALUES (1), (2), (null)) AS t1(a)
+JOIN (VALUES (1), (null)) AS t2(a)
+ON t1.a IS NOT DISTINCT FROM t2.a;
++---+
+| A |
++---+
+| 1 |
+|   |
++---+
+(2 rows)
+
+!ok
+
+# Test for Varchar/Char
+SELECT t1.a
+FROM (VALUES ('x'), ('y'), (null)) AS t1(a)
+JOIN (VALUES ('x'), (null)) AS t2(a)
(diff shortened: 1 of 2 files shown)


core/src/test/resources/sql/join.iq
```diff
@@ -854,4 +854,154 @@ EnumerableCalc(expr#0..1=[{inputs}], A=[$t0])
     EnumerableValues(tuples=[[{ 1 }, { null }]])
 !plan
 
+# Test for Double
+SELECT t1.a
```
mihaibudiu, line 858: I think these are DECIMAL.
For DOUBLE you have to write 1.0e0

mihaibudiu: the test is good, but add a DOUBLE test too
In general, joining on double is tricky, but for this case it should work fine

core/src/test/resources/sql/join.iq
```diff
@@ -854,4 +854,154 @@ EnumerableCalc(expr#0..1=[{inputs}], A=[$t0])
...
+| A          |
++------------+
+| 2025-05-10 |
+|            |
++------------+
+(2 rows)
+
+!ok
+
+# Test for TimesStamp
```
mihaibudiu, line 932: typo

core/src/test/resources/sql/join.iq
```diff
@@ -854,4 +854,154 @@ EnumerableCalc(expr#0..1=[{inputs}], A=[$t0])
...
+| A                   |
++---------------------+
+| 2025-05-10 10:22:34 |
+|                     |
++---------------------+
+(2 rows)
+
+!ok
+
+# Test for Decimal
```
mihaibudiu, line 947: were these tests failing previously?

mihaibudiu (approved):

core/src/test/resources/sql/join.iq
xiedeyantu: Yes

xiedeyantu:
@mihaibudiu I have fixed the comment above.

core/src/test/resources/sql/join.iq
mihaibudiu: the test for DECIMAL was good too, but it's also covered by the ARRAY tests below

mihaibudiu:
I will merge this once CI passes; it's a serious bug. Surprising it wasn't found earlier.

xiedeyantu:
> I will merge this once CI passes; it's a serious bug. Surprising it wasn't found earlier.

Please wait for me to squish these commits. The reason is that previously it was `is not distinct from` and could not match null values. After implementing `IsNotDistinctitFromImplementer,` the `equal` method was not implemented correctly, which led to this issue.

xiedeyantu:
@mihaibudiu The continuous-integration CI failure seems to have nothing to do with this PR. If necessary, I can trigger CI again.

PR #4369, opened 2025-05-12
[CALCITE-6887] ReduceExpressionsRule applied to 'IN subquery' should make the values distinct if the subquery is a constant Values

xuzifu666:

According the discussion in: https://github.com/apache/calcite/pull/4242
add a AggregateValueReduceRule to distinct IN filter values.

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
@@ -61,8 +61,7 @@ public class AggregateValueReduceRule
   protected AggregateValueReduceRule(Config config) {
     super(config);
   }
-
-  @Deprecated // to be removed before 2.0
+  
   public AggregateValueReduceRule(RelBuilderFactory relBuilderFactory) {
     this(Config.DEFAULT
         .withRelBuilderFactory(relBuilderFactory)
--- a/core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
@@ -41,13 +41,15 @@ import java.util.stream.Collectors;
  * <p>Sample query where this matters:
  *
  * <blockquote><code>
- * SELECT deptno, sal FROM EMP WHERE deptno IN (1,1,3,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1)
+ * LogicalAggregate(group=[{0}])
+ * LogicalValues(tuples=[[{ 1 }, { 1 }, { 3 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 },
+ * { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }]])
  * </code></blockquote>
  *
  * <p>should be converted to:
  *
  * <blockquote><code>
- * SELECT deptno, sal FROM EMP WHERE deptno IN (1,3)
+ * LogicalValues(tuples=[[{ 1 }, { 3 }]])
  * </code></blockquote>
  *
  * @see CoreRules#AGGREGATE_VALUES_REDUCE
@@ -70,7 +72,6 @@ public class AggregateValueReduceRule
 
   @Override public void onMatch(RelOptRuleCall call) {
     final Values values = call.rel(1);
-    Util.discard(values);
     final RelBuilder relBuilder = call.builder();
 
     List<ImmutableList<RexLiteral>> distinctValues =
--- a/core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
@@ -61,7 +61,7 @@ public class AggregateValueReduceRule
   protected AggregateValueReduceRule(Config config) {
     super(config);
   }
-  
+
   public AggregateValueReduceRule(RelBuilderFactory relBuilderFactory) {
     this(Config.DEFAULT
         .withRelBuilderFactory(relBuilderFactory)
--- a/core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.rel.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelRule;
+import org.apache.calcite.rel.core.Aggregate;
+import org.apache.calcite.rel.core.Values;
+import org.apache.calcite.rex.RexLiteral;
+import org.apache.calcite.tools.RelBuilder;
+import org.apache.calcite.tools.RelBuilderFactory;
+import org.apache.calcite.util.Util;
+
+import com.google.common.collect.ImmutableList;
+
+import org.immutables.value.Value;
+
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
```diff
@@ -0,0 +1,104 @@
...
+ * <p>Sample query where this matters:
+ *
+ * <blockquote><code>
+ * SELECT deptno, sal FROM EMP WHERE deptno IN (1,1,3,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1)
+ * </code></blockquote>
+ *
+ * <p>should be converted to:
+ *
+ * <blockquote><code>
+ * SELECT deptno, sal FROM EMP WHERE deptno IN (1,3)
```
xiedeyantu, line 50: Rule examples can be more universal. If it is difficult to describe with SQL, a simple plan can be used. For example:
```
<p>Sample plan where this matters:

<blockquote><code>
LogicalAggregate(group=[{0}])
  LogicalValues(tuples=[[{ 1 }, { 1 }, { 3 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }]])
</code></blockquote>

<p>should be converted to:

<blockquote><code>
LogicalValues(tuples=[[{ 1 }, { 3 }]])
</code></blockquote>
```
You can use this case to receive this result.
```
  @Test void testReduceInValues() {
        .values(new String[]{"v"}, 1,1,3,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1)
        .distinct()
        .build();
    relFn(relFn)
        .withRule(CoreRules.AGGREGATE_VALUES_REDUCE)
        .check();
  }
```

core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
```diff
@@ -0,0 +1,104 @@
...
+
+  public AggregateValueReduceRule(RelBuilderFactory relBuilderFactory) {
+    this(Config.DEFAULT
+        .withRelBuilderFactory(relBuilderFactory)
+        .as(Config.class));
+  }
+
+  @Override public void onMatch(RelOptRuleCall call) {
+    final Values values = call.rel(1);
+    Util.discard(values);
```
xiedeyantu, line 73: What is the purpose of this discard?

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
```diff
@@ -847,6 +847,10 @@ private CoreRules() {}
   public static final AggregateValuesRule AGGREGATE_VALUES =
       AggregateValuesRule.Config.DEFAULT.toRule();
 
+  /** Rule that applies an {@link Aggregate} to a distinct {@link Values}. */
```
xiedeyantu, line 850: /** Rule that applies {@link Aggregate} and {@link Values} to a distinct {@link Values}. */ Is that right?

xuzifu666: Yes，your suggestion is more suitable.  had changed as it. @xiedeyantu

core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
xuzifu666: Here not need the discard call, this is just to avoid warning, removed it.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
xuzifu666: Had changed it~

xuzifu666:
@xiedeyantu Thanks for review, had modified the code according to suggestion.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
```diff
@@ -0,0 +1,104 @@
...
+public class AggregateValueReduceRule
+    extends RelRule<AggregateValueReduceRule.Config>
+    implements SubstitutionRule {
+
+  /** Creates an AggregateValuesRule. */
+  protected AggregateValueReduceRule(Config config) {
+    super(config);
+  }
+
+  public AggregateValueReduceRule(RelBuilderFactory relBuilderFactory) {
```
xiedeyantu, line 66: This is an old interface method, there is no need to support it anymore.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
```diff
@@ -0,0 +1,104 @@
...
+
+import com.google.common.collect.ImmutableList;
+
+import org.immutables.value.Value;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Rule that applies an {@link Aggregate} to a distinct {@link Values}.
```
xiedeyantu, line 35: Maybe need change here too.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
xuzifu666: Yes, not need it~

core/src/main/java/org/apache/calcite/rel/rules/AggregateValueReduceRule.java
xuzifu666: Done

xiedeyantu:
Sorry, I noticed that there is also an AGGREGATE_VALUES rule that simplifies AGG + VALUES to VALUES. Can we extend this rule? we also paying attention to the `aggCalls` and `groupSets` in the AGG. It might only support scenarios where `groupSets.size() == 1`.

xuzifu666:
> Sorry, I noticed that there is also an AGGREGATE_VALUES rule that simplifies AGG + VALUES to VALUES. Can we extend this rule? we also paying attention to the `aggCalls` and `groupSets` in the AGG. It might only support scenarios where `groupSets.size() == 1`.

Add a new Rule or extend AGGREGATE_VALUES maybe both can resolve the issue, and maybe we can  wait for others' suggession~ @xiedeyantu

asolimando:
> > Sorry, I noticed that there is also an AGGREGATE_VALUES rule that simplifies AGG + VALUES to VALUES. Can we extend this rule? we also paying attention to the `aggCalls` and `groupSets` in the AGG. It might only support scenarios where `groupSets.size() == 1`.
> 
> Add a new Rule or extend AGGREGATE_VALUES maybe both can resolve the issue, and we can wait for others' suggession~ @xiedeyantu

IMO, if the existing rule can be expanded with a functionality that has no foreseeable drawbacks, and it doesn't cause any regressions, I think it's preferrable than adding a new rule

xuzifu666:
> > > Sorry, I noticed that there is also an AGGREGATE_VALUES rule that simplifies AGG + VALUES to VALUES. Can we extend this rule? we also paying attention to the `aggCalls` and `groupSets` in the AGG. It might only support scenarios where `groupSets.size() == 1`.
> > 
> > 
> > Add a new Rule or extend AGGREGATE_VALUES maybe both can resolve the issue, and we can wait for others' suggession~ @xiedeyantu
> 
> IMO, if the existing rule can be expanded with a functionality that has no foreseeable drawbacks, and it doesn't cause any regressions, I think it's preferrable than adding a new rule

OK, thanks for suggestion and I would extend it as suggestion. @asolimando @xiedeyantu

xuzifu666:
@asolimando @xiedeyantu Direct modification is a bit troublesome，so I make a new PR, we can continue in https://github.com/apache/calcite/pull/4371

asolimando:
> @asolimando @xiedeyantu Direct modification is a bit troublesome，so I make a new PR, we can continue in #4371 and this pr would be closed.

@xuzifu666 we are at 3 PRs for the same ticket, I have the feeling that we (might) have lost of lot of valuable comments along the way.

Even if we were somehow tracking all questions carefully, when you re-open the PR including the asked modifications, you can't expect people down the line checking multiple PRs to understand the rationale of what was done.

PRs are very important, not only for reviewing purposes, they are also a written trace of what lead to the code as-is, and it's therefore crucial to keep it clean and accessible, same for the Jira discussions.

If you are facing issues with git or anything else please ask here (or in the ML) and we can surely help out, tooling can be problematic, there is absolutely no problem with that, feel free to ask.

xuzifu666:
> > @asolimando @xiedeyantu Direct modification is a bit troublesome，so I make a new PR, we can continue in #4371 and this pr would be closed.
> 
> @xuzifu666 we are at 3 PRs for the same ticket, I have the feeling that we (might) have lost of lot of valuable comments along the way.
> 
> Even if we were somehow tracking all questions carefully, when you re-open the PR including the asked modifications, you can't expect people down the line checking multiple PRs to understand the rationale of what was done.
> 
> PRs are very important, not only for reviewing purposes, they are also a written trace of what lead to the code as-is, and it's therefore crucial to keep it clean and accessible, same for the Jira discussions.
> 
> If you are facing issues with git or anything else please ask here (or in the ML) and we can surely help out, tooling can be problematic, there is absolutely no problem with that, feel free to ask.

Thanks for your suggestion, I would notice the pr manager in further!@asolimando

PR #4371, opened 2025-05-13
[CALCITE-6887] ReduceExpressionsRule applied to 'IN subquery' should make the values distinct if the subquery is a Values composed of literals

xuzifu666:

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
@@ -34,10 +34,13 @@ import org.immutables.value.Value;
 import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
- * Rule that applies {@link Aggregate} to a {@link Values} (currently just an
+ * Rule that
+ * 1. applies {@link Aggregate} to a {@link Values} (currently just an
  * empty {@code Value}s).
+ * 2. {@link Aggregate} and {@link Values} to a distinct {@link Values}.
  *
  * <p>This is still useful because {@link PruneEmptyRules#AGGREGATE_INSTANCE}
  * doesn't handle {@code Aggregate}, which is in turn because {@code Aggregate}
@@ -52,6 +55,20 @@ import java.util.List;
  * <p>This rule only applies to "grand totals", that is, {@code GROUP BY ()}.
  * Any non-empty {@code GROUP BY} clause will return one row per group key
  * value, and each group will consist of at least one row.
+ *
+ * <p>This is useful because {@link SubQueryRemoveRule}
+ * doesn't remove duplicates for IN filter values.
+ *
+ * <blockquote><code>
+ * LogicalAggregate(group=[{0}])
+ *   LogicalValues(tuples=[[{ 1 }, { 1 }, { 3 }]])
+ * </code></blockquote>
+ *
+ * <p>With the Rule would deduplicate the values and convert to:
+ *
+ * <blockquote><code>
+ * LogicalValues(tuples=[[{ 1 }, { 3 }]])
+ * </code></blockquote>
  *
  * @see CoreRules#AGGREGATE_VALUES
  */
@@ -72,6 +89,8 @@ public class AggregateValuesRule
         .as(Config.class));
   }
 
+  //~ Methods ----------------------------------------------------------------
+
   @Override public void onMatch(RelOptRuleCall call) {
     final Aggregate aggregate = call.rel(0);
     final Values values = call.rel(1);
@@ -79,33 +98,46 @@ public class AggregateValuesRule
     final RelBuilder relBuilder = call.builder();
     final RexBuilder rexBuilder = relBuilder.getRexBuilder();
 
-    final List<RexLiteral> literals = new ArrayList<>();
-    for (final AggregateCall aggregateCall : aggregate.getAggCallList()) {
-      switch (aggregateCall.getAggregation().getKind()) {
-      case COUNT:
-      case SUM0:
-        literals.add(
-            rexBuilder.makeLiteral(BigDecimal.ZERO, aggregateCall.getType()));
-        break;
+    if (aggregate.getGroupCount() == 0 && values.getTuples().isEmpty()) {
+      final List<RexLiteral> literals = new ArrayList<>();
+      for (final AggregateCall aggregateCall : aggregate.getAggCallList()) {
+        switch (aggregateCall.getAggregation().getKind()) {
+        case COUNT:
+        case SUM0:
+          literals.add(
+              rexBuilder.makeLiteral(BigDecimal.ZERO, aggregateCall.getType()));
+          break;
 
-      case MIN:
-      case MAX:
-      case SUM:
-        literals.add(rexBuilder.makeNullLiteral(aggregateCall.getType()));
-        break;
+        case MIN:
+        case MAX:
+        case SUM:
+          literals.add(rexBuilder.makeNullLiteral(aggregateCall.getType()));
+          break;
 
-      default:
-        // Unknown what this aggregate call should do on empty Values. Bail out to be safe.
-        return;
+        default:
+          // Unknown what this aggregate call should do on empty Values. Bail out to be safe.
+          return;
+        }
       }
+
+      call.transformTo(
+          relBuilder.values(ImmutableList.of(literals), aggregate.getRowType())
+              .build());
+
+      // New plan is absolutely better than old plan.
+      call.getPlanner().prune(aggregate);
+      return;
     }
 
-    call.transformTo(
-        relBuilder.values(ImmutableList.of(literals), aggregate.getRowType())
-            .build());
-
-    // New plan is absolutely better than old plan.
-    call.getPlanner().prune(aggregate);
+    if (Aggregate.isSimple(aggregate)
+        && aggregate.getAggCallList().isEmpty()
+        && aggregate.getRowType().equals(values.getRowType())) {
+      List<ImmutableList<RexLiteral>> distinctValues =
+          values.getTuples().stream().distinct().collect(Collectors.toList());
+      relBuilder.values(distinctValues, values.getRowType());
+      call.transformTo(relBuilder.build());
+      call.getPlanner().prune(aggregate);
+    }
   }
 
   /** Rule configuration. */
@@ -123,10 +155,8 @@ public class AggregateValuesRule
         Class<? extends Values> valuesClass) {
       return withOperandSupplier(b0 ->
           b0.operand(aggregateClass)
-              .predicate(aggregate -> aggregate.getGroupCount() == 0)
               .oneInput(b1 ->
                   b1.operand(valuesClass)
-                      .predicate(values -> values.getTuples().isEmpty())
                       .noInputs()))
           .as(Config.class);
     }
--- a/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
+++ b/core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
@@ -354,6 +354,73 @@ class RelOptRulesTest extends RelOptTestBase {
         .check();
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6887">[CALCITE-6887]
+   * ReduceExpressionsRule applied to 'IN subquery' should make the values distinct
+   * if the subquery is a Values composed of literals</a>. */
+  @Test void testAggregateValuesRuleWithInRepeatingValues() {
+    final String sql = "SELECT deptno, sal "
+        + "FROM EMP "
+        + "WHERE deptno IN (1,1,3,1,2,1,1,1,1,1,1,1,null,null,1,1,1,1,1,1,1)";
+
+    sql(sql)
+        .withPreRule(CoreRules.FILTER_SUB_QUERY_TO_CORRELATE)
+        .withRule(CoreRules.AGGREGATE_VALUES)
+        .check();
+  }
+
+  @Test void testReduceInValuesNonOrdered() {
+    final String sql = "SELECT deptno, sal "
+        + "FROM EMP "
+        + "WHERE deptno IN (1,1,3,1,2)";
+
+    sql(sql)
+        .withPreRule(CoreRules.FILTER_SUB_QUERY_TO_CORRELATE)
+        .withRule(CoreRules.AGGREGATE_VALUES)
+        .checkUnchanged();
+  }
+
+  @Test void testAggregateValuesRuleWithRepeatingValues() {
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .values(new String[]{"v"}, 1, 1, 3, null, null)
+        .distinct()
+        .build();
+    relFn(relFn)
+        .withRule(CoreRules.AGGREGATE_VALUES)
+        .check();
+  }
+
+  @Test void testAggregateValuesRuleWithAggregation() {
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .values(new String[]{"v1", "v2"}, 1, "a", 2, "b", 1, "a", 1, "b")
+        .aggregate(b.groupKey(ImmutableBitSet.of(0, 1)),
+            b.count(b.field(0)).as("V1_COUNT"))
+        .build();
+    relFn(relFn)
+        .withRule(CoreRules.AGGREGATE_VALUES)
+        .checkUnchanged();
+  }
+
+  @Test void testAggregateValuesRuleWithSameRowType() {
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .values(new String[]{"v1", "v2"}, 1, "a", 2, "b", 1, "a", 1, "b")
+        .aggregate(b.groupKey(ImmutableBitSet.of(0, 1)))
+        .build();
+    relFn(relFn)
+        .withRule(CoreRules.AGGREGATE_VALUES)
+        .check();
+  }
+
+  @Test void testAggregateValuesRuleWithDifferentRowTypes() {
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .values(new String[]{"v1", "v2"}, 1, "a", 2, "b", 1, "a", 1, "b")
+        .aggregate(b.groupKey(ImmutableBitSet.of(0)))
+        .build();
+    relFn(relFn)
+        .withRule(CoreRules.AGGREGATE_VALUES)
+        .checkUnchanged();
+  }
+
   /**
    * Test case for <a href="https://issues.apache.org/jira/projects/CALCITE/issues/CALCITE-6746">
    * [CALCITE-6746] Optimization rule ProjectWindowTranspose is unsound</a>. */
--- a/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
@@ -19613,4 +19613,82 @@ LogicalProject(DEPTNO=[$0], EXPR$1=[$1], EXPR$2=[CAST(/($2, $3)):INTEGER NOT NUL
 ]]>
     </Resource>
   </TestCase>
+  <TestCase name="testAggregateValuesRuleWithAggregation">
+    <Resource name="planBefore">
+      <![CDATA[
+LogicalAggregate(group=[{0, 1}], V1_COUNT=[COUNT()])
+  LogicalValues(tuples=[[{ 1, 'a' }, { 2, 'b' }, { 1, 'a' }, { 1, 'b' }]])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testAggregateValuesRuleWithDifferentRowTypes">
+    <Resource name="planBefore">
+      <![CDATA[
+LogicalAggregate(group=[{0}])
+  LogicalValues(tuples=[[{ 1, 'a' }, { 2, 'b' }, { 1, 'a' }, { 1, 'b' }]])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testReduceInValuesNonOrdered">
+    <Resource name="sql">
+      <![CDATA[SELECT deptno, sal FROM EMP WHERE deptno IN (1,1,3,1,2)]]>
+    </Resource>
+    <Resource name="planBefore">
+      <![CDATA[
+LogicalProject(DEPTNO=[$7], SAL=[$5])
+  LogicalFilter(condition=[OR(=($7, 1), =($7, 3), =($7, 2))])
+    LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testAggregateValuesRuleWithInRepeatingValues">
+    <Resource name="sql">
+      <![CDATA[SELECT deptno, sal FROM EMP WHERE deptno IN (1,1,3,1,2,1,1,1,1,1,1,1,null,null,1,1,1,1,1,1,1)]]>
+    </Resource>
+    <Resource name="planBefore">
+      <![CDATA[
+LogicalProject(DEPTNO=[$7], SAL=[$5])
+  LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8])
+    LogicalJoin(condition=[=($7, $9)], joinType=[inner])
+      LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+      LogicalAggregate(group=[{0}])
+        LogicalValues(tuples=[[{ 1 }, { 1 }, { 3 }, { 1 }, { 2 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { null }, { null }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }]])
+]]>
+    </Resource>
+    <Resource name="planAfter">
+      <![CDATA[
+LogicalProject(DEPTNO=[$7], SAL=[$5])
+  LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8])
+    LogicalJoin(condition=[=($7, $9)], joinType=[inner])
+      LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+      LogicalValues(tuples=[[{ 1 }, { 3 }, { 2 }, { null }]])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testAggregateValuesRuleWithRepeatingValues">
+    <Resource name="planBefore">
+      <![CDATA[
+LogicalAggregate(group=[{0}])
+  LogicalValues(tuples=[[{ 1 }, { 1 }, { 3 }, { null }, { null }]])
+]]>
+    </Resource>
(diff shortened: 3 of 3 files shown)


xuzifu666:
@asolimando @xiedeyantu We can continue here, and I had extend AggregateValuesRule from https://github.com/apache/calcite/pull/4369

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
```diff
@@ -53,6 +56,21 @@
  * Any non-empty {@code GROUP BY} clause will return one row per group key
  * value, and each group will consist of at least one row.
  *
+ * <p>This is useful because such as {@link SubQueryRemoveRule}
+ * doesn't distinct values for IN filter values.
+ *
+ * <blockquote><code>
+ * LogicalAggregate(group=[{0}])
+ * LogicalValues(tuples=[[{ 1 }, { 1 }, { 3 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 }, { 1 },
```
asolimando, line 64: Nit: the example could be made simpler/shorter without loss

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -354,6 +354,32 @@ private static boolean skipItem(RexNode expr) {
...
   }
 
+  /** Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-6887">[CALCITE-6887]
+   * ReduceExpressionsRule applied to 'IN subquery' should make the values distinct
+   * if the subquery is a constant Values</a>. */
+  @Test void testReduceInValues() {
+    final String sql = "SELECT deptno, sal "
+        + "FROM EMP "
+        + "WHERE deptno IN (1,1,3,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1)";
```
asolimando, line 364: I'd make the test simpler (no need for that many ending `1`s, a couple suffice) and add more tests around edge cases (rule is no-op if no duplicates, multiple duplicate values, nulls, etc)

xuzifu666: Thanks for reminder, here with many ending 1s is due to greater than 20 would construcor with IN operator (otherwise would not), and I would add more tests according your suggestion. @asolimando

xuzifu666: @asolimando According test and analyze，currently maybe cannot support complex data types，such as：
```
@Test void testReduceInValuesWithAggregateValueReduceRuleWithNestedType() {
      .values(new String[]{"v"},  new int[]{1, 2, 3})
      .distinct()
      .build();
  relFn(relFn)
      .withRule(CoreRules.AGGREGATE_VALUES)
      .check();
}
```
would error out：
```
cannot convert [I@6233d094 (class [I) to a constant
java.lang.IllegalArgumentException: cannot convert [I@6233d094 (class [I) to a constant
 at org.apache.calcite.tools.RelBuilder.literal(RelBuilder.java:497)
 at org.apache.calcite.tools.RelBuilder.tupleList(RelBuilder.java:3463)
 at org.apache.calcite.tools.RelBuilder.values(RelBuilder.java:3423)
 at org.apache.calcite.test.RelOptRulesTest.lambda$testReduceInValuesWithAggregateValueReduceRuleWithNestedType$8(RelOptRulesTest.java:394)
 at org.apache.calcite.test.RelSupplier$FnRelSupplier.apply(RelSupplier.java:131)
 at org.apache.calcite.test.RelOptFixture.toRel(RelOptFixture.java:309)
 at org.apache.calcite.test.RelOptFixture.checkPlanning(RelOptFixture.java:345)
 at org.apache.calcite.test.RelOptFixture.check(RelOptFixture.java:334)
 at org.apache.calcite.test.RelOptFixture.check(RelOptFixture.java:318)
```

reason is  RelBuilder##literal method can only handle with basic types. 
```
public RexLiteral literal(@Nullable Object value) {
    final RexBuilder rexBuilder = cluster.getRexBuilder();
    if (value == null) {
      final RelDataType type = getTypeFactory().createSqlType(SqlTypeName.NULL);
      return rexBuilder.makeNullLiteral(type);
    } else if (value instanceof Boolean) {
      return rexBuilder.makeLiteral((Boolean) value);
    } else if (value instanceof BigDecimal) {
      return rexBuilder.makeExactLiteral((BigDecimal) value);
    } else if (value instanceof Float || value instanceof Double) {
      return rexBuilder.makeApproxLiteral(
          ((Number) value).doubleValue(), getTypeFactory().createSqlType(SqlTypeName.DOUBLE));
    } else if (value instanceof Number) {
      return rexBuilder.makeExactLiteral(
          BigDecimal.valueOf(((Number) value).longValue()));
    } else if (value instanceof String) {
      return rexBuilder.makeLiteral((String) value);
    } else if (value instanceof Enum) {
      return rexBuilder.makeLiteral(value,
          getTypeFactory().createSqlType(SqlTypeName.SYMBOL));
    } else if (value instanceof DateString) {
      return rexBuilder.makeDateLiteral((DateString) value);
    } else if (value instanceof Date) {
      return rexBuilder.makeLiteral(value,
          getTypeFactory().createSqlType(SqlTypeName.DATE));
    } else {
      throw new IllegalArgumentException("cannot convert " + value
          + " (" + value.getClass() + ") to a constant");
    }
  }
```
so maybe we can just resolve non-complex data types currently and I had added more case to cover more conditions which expected as the rule.

mihaibudiu: Because these are not literals, they are constant expressions. That's not the same thing.

xuzifu666: @mihaibudiu RelBuilder##values would call tupleList method which can only support with Literal
```
  public RelBuilder values(@Nullable String[] fieldNames, @Nullable Object... values) {
    requireNonNull(fieldNames, "fieldNames");
    if (fieldNames.length == 0
        || values.length % fieldNames.length != 0
        || values.length < fieldNames.length) {
      throw new IllegalArgumentException(
          "Value count must be a positive multiple of field count");
    }
    final int rowCount = values.length / fieldNames.length;
    for (Ord<@Nullable String> fieldName : Ord.zip(fieldNames)) {
      if (allNull(values, fieldName.i, fieldNames.length)) {
        throw new IllegalArgumentException("All values of field '" + fieldName.e
            + "' (field index " + fieldName.i + ")"
            + " are null; cannot deduce type");
      }
    }
    final ImmutableList<ImmutableList<RexLiteral>> tupleList =
        **tupleList(fieldNames.length, values);**
    assert tupleList.size() == rowCount;
    final List<String> fieldNameList =
        Util.transformIndexed(Arrays.asList(fieldNames), (name, i) ->
            name != null ? name : SqlUtil.deriveAliasFromOrdinal(i));
    return values(tupleList, fieldNameList);
  }
```

 RelBuilder##tupleList 
```
  private ImmutableList<ImmutableList<RexLiteral>> tupleList(int columnCount,
      @Nullable Object[] values) {
    final ImmutableList.Builder<ImmutableList<RexLiteral>> listBuilder =
        ImmutableList.builder();
    final List<RexLiteral> valueList = new ArrayList<>();
    for (int i = 0; i < values.length; i++) {
      Object value = values[i];
      valueList.add(literal(value));
      if ((i + 1) % columnCount == 0) {
        listBuilder.add(ImmutableList.copyOf(valueList));
        valueList.clear();
      }
    }
    return listBuilder.build();
  }
```
So IMO here can only handle with Literal.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
```diff
@@ -75,10 +91,18 @@ public AggregateValuesRule(RelBuilderFactory relBuilderFactory) {
...
     final Aggregate aggregate = call.rel(0);
     final Values values = call.rel(1);
-    Util.discard(values);
     final RelBuilder relBuilder = call.builder();
     final RexBuilder rexBuilder = relBuilder.getRexBuilder();
 
+    if ((aggregate.getGroupCount() != 0 || !values.getTuples().isEmpty())
+        && aggregate.groupSets.size() == 1) {
+      List<ImmutableList<RexLiteral>> distinctValues =
+          values.getTuples().stream().distinct().collect(Collectors.toList());
```
mihaibudiu, line 100: this distinct uses Java equals(). I assume it is always correct, but I am not sure it will always find all duplicates.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
```diff
@@ -75,10 +91,18 @@ public AggregateValuesRule(RelBuilderFactory relBuilderFactory) {
...
+      relBuilder.values(distinctValues, values.getRowType());
```
mihaibudiu, line 101: doesn't the field you group on need to be the same that is aggregated as well?
where is that check?
VALUES can contain ROW objects.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
xuzifu666: Your are right, but IMO maybe can handle with it by this way, due to RelBuilder##values can only handle RexLiteral type, and all type of RexLiteral had implement Compare interface which may remove duplicates correctly.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
xuzifu666: Here would only handle with groupSets.size = 1 condition，I had a tried seems ROW type can not constructor in test case. Can we just handle with Literal type first？

xiedeyantu: Mihai may mean this: 
```
        .values(new String[]{"v1", "v2"},  1, "a", 2, "b", 1, "a", 1, "b")
        .aggregate(b.groupKey(ImmutableBitSet.of(0)))
        .build();
```
This will generate the following plan, which only does a GROUP BY on the v1 field.
```
LogicalAggregate(group=[{0}])
  LogicalValues(tuples=[[{ 1, 'a' }, { 2, 'b' }, { 1, 'a' }, { 1, 'b' }]])
```
Maybe you need to check whether this scenario is supported.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
```diff
@@ -75,10 +91,18 @@ public AggregateValuesRule(RelBuilderFactory relBuilderFactory) {
   @Override public void onMatch(RelOptRuleCall call) {
     final Aggregate aggregate = call.rel(0);
     final Values values = call.rel(1);
-    Util.discard(values);
     final RelBuilder relBuilder = call.builder();
     final RexBuilder rexBuilder = relBuilder.getRexBuilder();
 
+    if ((aggregate.getGroupCount() != 0 || !values.getTuples().isEmpty())
```
xiedeyantu, line 97: This judgment condition may have some problems. Maybe you can change like this.
```
    if (aggregate.getGroupCount() == 0 && values.getTuples().isEmpty()) {
      // old
    } else if (/* your condition */) {
      
    }
```
This may be a bad case:
```
    final String sql = "select count(*), sum(empno) from emp where false group by empno";
```

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
xuzifu666: Had test for Row type about it and it would distinct result as expected.
```
  <TestCase name="testReduceInValuesWithAggregateValueReduceRuleWithRow">
    <Resource name="planBefore">
      <![CDATA[
LogicalAggregate(group=[{0, 1}])
  LogicalValues(tuples=[[{ 1, 'a' }, { 2, 'b' }, { 1, 'a' }, { 1, 'b' }, { 1, 'a' }]])
]]>
    </Resource>
    <Resource name="planAfter">
      <![CDATA[
LogicalValues(tuples=[[{ 1, 'a' }, { 2, 'b' }, { 1, 'b' }]])
]]>
    </Resource>
```

xiedeyantu: You should use `aggregate(b.groupKey(ImmutableBitSet.of(0)))` to test. `distinct()` can not expose any problems.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xuzifu666: @asolimando I also  added the case with non-sorted and less then 20 values in commit: https://github.com/apache/calcite/commit/c4273c449855a8dc8caa48c0b3f49d27a782495e

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
```diff
@@ -53,6 +55,20 @@
  * Any non-empty {@code GROUP BY} clause will return one row per group key
  * value, and each group will consist of at least one row.
  *
+ * <p>This is useful because such as {@link SubQueryRemoveRule}
+ * doesn't distinct values for IN filter values.
+ *
+ * <blockquote><code>
+ * LogicalAggregate(group=[{0}])
+ * LogicalValues(tuples=[[{ 1 }, { 1 }, { 3 }]])
```
xiedeyantu, line 63:
```
 * LogicalAggregate(group=[{0}])
 *   LogicalValues(tuples=[[{ 1 }, { 1 }, { 3 }]])
```

xuzifu666:
Thanks for all comments, had modified as follow:
1. AggregateValuesRule is expanded to 2 modes, handling the default case and deduplication respectively;
2. Deduplication cannot be fully guaranteed for complex types such as Row, so deduplication is performed on non-complex type data of Literal first;
3. Related tests are added.
please take a look for the changes. @mihaibudiu @asolimando @xiedeyantu

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
xuzifu666:
```
doesn't the field you group on need to be the same that is aggregated as well?
where is that check?
VALUES can contain ROW objects.
```
@mihaibudiu had added check for rowType. Need to be same with that is aggregated.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -354,6 +354,72 @@ private static boolean skipItem(RexNode expr) {
...
+    sql(sql)
+        .withPreRule(CoreRules.FILTER_SUB_QUERY_TO_CORRELATE)
+        .withRule(CoreRules.AGGREGATE_VALUES_REDUCE)
+        .check();
+  }
+
+  @Test void testReduceInValuesWithNonOrder() {
```
xiedeyantu, line 371: You can just use `testAggregateValuesRuleWithInRepeatingValues` to show this logic.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -354,6 +354,72 @@ private static boolean skipItem(RexNode expr) {
...
+        + "FROM EMP "
+        + "WHERE deptno IN (1,1,3,1,2)";
+
+    sql(sql)
+        .withPreRule(CoreRules.FILTER_SUB_QUERY_TO_CORRELATE)
+        .withRule(CoreRules.AGGREGATE_VALUES_REDUCE)
+        .checkUnchanged();
+  }
+
+  @Test void testAggregateValuesRuleWithRepeatingValues() {
```
xiedeyantu, line 382: Same as above.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
```diff
@@ -53,6 +56,20 @@
...
+ * doesn't distinct values for IN filter values.
+ *
+ * <blockquote><code>
+ * LogicalAggregate(group=[{0}])
+ *   LogicalValues(tuples=[[{ 1 }, { 1 }, { 3 }]])
+ * </code></blockquote>
+ *
+ * <p>should be converted to:
+ *
+ * <blockquote><code>
```
asolimando, line 69: What happened to the aggregation here? Is it now redundant and removed due to the deduplication? If so, please state it explicitly in the comment.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
asolimando, line 371:
```suggestion
  @Test void testReduceInValuesNonOrdered() {
``` ? or `testReduceUnsortedInValues()`?

File: core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java [hunk 10]
```diff
@@ -354,6 +354,72 @@ private static boolean skipItem(RexNode expr) {
...
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .values(new String[]{"v"},  1, 1, 3, null, null)
```
asolimando, line 384:
```suggestion
        .values(new String[]{"v"}, 1, 1, 3, null, null)
```

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -354,6 +354,72 @@ private static boolean skipItem(RexNode expr) {
...
+        .distinct()
+        .build();
+    relFn(relFn)
+        .withRule(CoreRules.AGGREGATE_VALUES_REDUCE)
+        .check();
+  }
+
+  @Test void testAggregateValuesRuleWithAggCall() {
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .values(new String[]{"v1", "v2"},  1, "a", 2, "b", 1, "a", 1, "b")
```
asolimando, line 394:
```suggestion
        .values(new String[]{"v1", "v2"}, 1, "a", 2, "b", 1, "a", 1, "b")
```

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -354,6 +354,72 @@ private static boolean skipItem(RexNode expr) {
...
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .values(new String[]{"v"},  1, 1, 3, null, null)
+        .distinct()
+        .build();
+    relFn(relFn)
+        .withRule(CoreRules.AGGREGATE_VALUES_REDUCE)
+        .check();
+  }
+
+  @Test void testAggregateValuesRuleWithAggCall() {
```
asolimando, line 392:
```suggestion
  @Test void testAggregateValuesRuleWithAggregation() {
```

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -354,6 +354,72 @@ private static boolean skipItem(RexNode expr) {
...
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .values(new String[]{"v1", "v2"},  1, "a", 2, "b", 1, "a", 1, "b")
+        .aggregate(b.groupKey(ImmutableBitSet.of(0, 1)))
+        .build();
+    relFn(relFn)
+        .withRule(CoreRules.AGGREGATE_VALUES_REDUCE)
+        .check();
+  }
+
+  @Test void testAggregateValuesRuleWithDifferentRowType() {
```
asolimando, line 413:
```suggestion
  @Test void testAggregateValuesRuleWithDifferentRowTypes() {
```

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -354,6 +354,72 @@ private static boolean skipItem(RexNode expr) {
...
+    final Function<RelBuilder, RelNode> relFn = b -> b
+        .values(new String[]{"v1", "v2"},  1, "a", 2, "b", 1, "a", 1, "b")
```
asolimando, line 415:
```suggestion
        .values(new String[]{"v1", "v2"}, 1, "a", 2, "b", 1, "a", 1, "b")
```

asolimando:
> Thanks for all comments, had modified as follow:
> 
>     1. AggregateValuesRule is expanded to 2 modes, handling the default case and deduplication respectively;
> 
>     2. Deduplication cannot be fully ensured for complex types such as Row, so deduplication is performed on non-complex data types such as Literal first to ensure that deduplication is effective;
> 
>     3. Related tests are added
>        please take a look for the changes. @mihaibudiu @asolimando @xiedeyantu

I won't push back on this, but I was more convinced on having the two modes activated conditionally in the `onMatch` method, like in the previous version.

It's not always easy when to decide what makes two distinct rules or two optional behavior into the same one, as both approaches have pros and cons.

Separate rules put more burden to the end-user and overhead in the planner, to me it's only worth it when there is a clear logical separation on the modes, that users might want to pick selectively (e.g., I want mode `A` but not mode `B`).

Here I feel users would like any of the two optimizations, when their query matches the conditions, or none. I can't think of cases where one want one mode and not the other, so making two rules feels artificial.

What do you guys think?

asolimando (changes requested):
A few minor improvements, merging the two modes is a separate discussion.

xuzifu666:
Thanks for @asolimando detailed review，I agree with you，so I would addressed your comments then change it as previous version.

xuzifu666:
> > Thanks for all comments, had modified as follow:
> > ```
> > 1. AggregateValuesRule is expanded to 2 modes, handling the default case and deduplication respectively;
> > 
> > 2. Deduplication cannot be fully ensured for complex types such as Row, so deduplication is performed on non-complex data types such as Literal first to ensure that deduplication is effective;
> > 
> > 3. Related tests are added
> >    please take a look for the changes. @mihaibudiu @asolimando @xiedeyantu
> > ```
> 
> I won't push back on this, but I was more convinced on having the two modes activated conditionally in the `onMatch` method, like in the previous version.
> 
> It's not always easy when to decide what makes two distinct rules or two optional behavior into the same one, as both approaches have pros and cons.
> 
> Separate rules put more burden to the end-user and overhead in the planner, to me it's only worth it when there is a clear logical separation on the modes, that users might want to pick selectively (e.g., I want mode `A` but not mode `B`).
> 
> Here I feel users would like any of the two optimizations, when their query matches the conditions, or none. I can't think of cases where one want one mode and not the other, so making two rules feels artificial.
> 
> What do you guys think?

@asolimando Thanks for your detailed review and I agree with you, I would changed it to previous version.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
xuzifu666: @asolimando Thanks, I had changed it to: 
```
With the Rule would deduplicate the values and convert to:
```

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xuzifu666: Done

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xuzifu666: Done

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xuzifu666: Done

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xuzifu666: Done

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
```diff
@@ -72,40 +89,54 @@ public AggregateValuesRule(RelBuilderFactory relBuilderFactory) {
...
-    // New plan is absolutely better than old plan.
-    call.getPlanner().prune(aggregate);
+      call.transformTo(
+          relBuilder.values(ImmutableList.of(literals), aggregate.getRowType())
+              .build());
+
+      // New plan is absolutely better than old plan.
+      call.getPlanner().prune(aggregate);
+    } else {
+      if (Aggregate.isSimple(aggregate)
```
xiedeyantu, line 130: Here you can use a separate if, no need for an if else.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
```diff
@@ -72,40 +89,54 @@ public AggregateValuesRule(RelBuilderFactory relBuilderFactory) {
...
+          && aggregate.getAggCallList().isEmpty()
+          && aggregate.getRowType().equals(values.getRowType())) {
+        List<ImmutableList<RexLiteral>> distinctValues =
+            values.getTuples().stream().distinct().collect(Collectors.toList());
+        relBuilder.values(distinctValues, values.getRowType());
+        call.transformTo(relBuilder.build());
+        call.getPlanner().prune(aggregate);
```
xiedeyantu, line 137: Repeated code can be mentioned outside the if else.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
xuzifu666: Done

xiedeyantu (approved):
LGTM, please wait and see if @asolimando and @mihaibudiu will make any comments.

core/src/main/java/org/apache/calcite/rel/rules/AggregateValuesRule.java
xuzifu666: done

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -19610,6 +19610,46 @@ LogicalProject(DEPTNO=[$0], EXPR$1=[$1], EXPR$2=[CAST(/($2, $3)):INTEGER NOT NUL
...
+LogicalProject(DEPTNO=[$7], SAL=[$5])
+  LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8])
+    LogicalJoin(condition=[=($7, $9)], joinType=[inner])
+      LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+      LogicalValues(tuples=[[{ 1 }, { 3 }]])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testReduceInValuesWithAggregateValueReduceRule">
+    <Resource name="sql">
```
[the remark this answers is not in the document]
asolimando: CALCITE-7016 has been filed to track this, let's see where that discussion leads, we can gloss on this for now here

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
asolimando: > Because these are not literals, they are constant expressions. That's not the same thing.

Then I guess we should update the title of the Jira ticket to something like "ReduceExpressionsRule applied to 'IN subquery' should make the values distinct if the subquery is a Values composed of literals"?

asolimando (commented):
I need to take another pass on the PR but most (all?) my comments seem to have been addressed, I want to check again the other PRs to see if all comments there were addressed as well.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
xuzifu666: Make sense，I would change as your suggestion，very thanks for your review. @asolimando

asolimando (approved):
LGTM, can you please squash the commits into one @xuzifu666 making sure the commit title matches the usual format and the associated jira ticket title?

PR #4373, opened 2025-05-13
[CALCITE-7008] Extend MinusToAntiJoinRule to support n-way inputs

xiedeyantu:
Jira link: [CALCITE-7008](https://issues.apache.org/jira/browse/CALCITE-7008)

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
@@ -399,7 +399,7 @@ public class CoreRules {
       MinusToDistinctRule.Config.DEFAULT.toRule();
 
   /** Rule to translates a {@link Minus} to {@link Join} anti-join}. */
-  public static final MinusToAntiJoinRule MINUS_TO_ANTI_JOIN_RULE =
+  public static final MinusToAntiJoinRule MINUS_TO_ANTI_JOIN =
       MinusToAntiJoinRule.Config.DEFAULT.toRule();
 
   /** Rule that converts a {@link LogicalMatch} to the result of calling
--- a/core/src/main/java/org/apache/calcite/rel/rules/MinusToAntiJoinRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/MinusToAntiJoinRule.java
@@ -36,11 +36,18 @@ import java.util.List;
 /**
  * Planner rule that translates a {@link Minus}
  * to a series of {@link org.apache.calcite.rel.core.Join} that type is
- * {@link JoinRelType#ANTI}. This rule supports 2-way Minus conversion,
+ * {@link JoinRelType#ANTI}. This rule supports n-way Minus conversion,
  * as this rule can be repeatedly applied during query optimization to
  * refine the plan.
  *
- * <h2>Example</h2>
+ * <p>Example for 2-way
+ *
+ * <p>Original sql:
+ * <pre>{@code
+ * select ename from emp where deptno = 10
+ * except
+ * select ename from emp where deptno = 20
+ * }</pre>
  *
  * <p>Original plan:
  * <pre>{@code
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/apache/calcite/rel/rules/MinusToAntiJoinRule.java
```diff
@@ -36,33 +36,48 @@
...
  * Planner rule that translates a {@link Minus}
  * to a series of {@link org.apache.calcite.rel.core.Join} that type is
- * {@link JoinRelType#ANTI}. This rule supports 2-way Minus conversion,
+ * {@link JoinRelType#ANTI}. This rule supports n-way Minus conversion,
  * as this rule can be repeatedly applied during query optimization to
  * refine the plan.
  *
  * <h2>Example</h2>
  *
+ * <p>Original sql:
```
mihaibudiu, line 45: frankly, I think both examples are good, I would keep both.
the simple one is easier to read.

mihaibudiu (approved):

core/src/main/java/org/apache/calcite/rel/rules/MinusToAntiJoinRule.java
xiedeyantu: Done!

core/src/main/java/org/apache/calcite/rel/rules/CoreRules.java
```diff
@@ -399,7 +399,7 @@ private CoreRules() {}
       MinusToDistinctRule.Config.DEFAULT.toRule();
 
   /** Rule to translates a {@link Minus} to {@link Join} anti-join}. */
-  public static final MinusToAntiJoinRule MINUS_TO_ANTI_JOIN_RULE =
+  public static final MinusToAntiJoinRule MINUS_TO_ANTI_JOIN =
```
mihaibudiu, line 402: I don't think you can rename public constants, this will break user code that depends on these names.
Even if the name is not great, it's not worth the trouble changing it.

xiedeyantu: Sorry, I've restored it.

asolimando: This was introduced after `1.39.0`, so it's an unreleased feature at the moment, even assuming someone is using it from a snapshot, I personally consider breaking changes only w.r.t. released versions, so for me we are still in time to change this before `1.40.0`, but that's my personal opinion, maybe someone could confirm or deny my assumption?

xiedeyantu: @asolimando I started out with the same idea as you, but Mihai might be the one using snapshots, so I'm not sure what to do.

mihaibudiu: Don’t worry about snapshots. Let’s keep the same name as in 1.39

asolimando: @mihaibudiu, according to git blame, this whole file was not there at all in `1.39.0`, it's been added later during the development of `1.40.0`, that's why I am suggesting it's fine (under this exceptional condition) to modify a public constant without considering it a breaking change IMO. Do you share the same feeling?

mihaibudiu: Sure

asolimando: Thanks for the confirmation @mihaibudiu, @xiedeyantu if you feel like it you can change it back again

xiedeyantu: I want to change the name back to the same style as the other rules. Can you help me see if there are any other changes needed? Can I just revert and squash it?

asolimando: You mean specifically for this change or in general? For this change, just do it through the IDE's refactoring features so there should be no discrepancies due to the renaming.

At that point you can safely squash as Mihai has already approved the PR (and this change), if tests are green the contribution will be good to be merged.

xiedeyantu: Thanks for @mihaibudiu @asolimando, I have change it back again. And keep this pr in one commit.

PR #4374, opened 2025-05-13
[CALCITE-7013] Support building RexLiterals from Character values

xuzifu666:
As description in JIRA: https://issues.apache.org/jira/browse/CALCITE-7013

### DIFF
--- a/core/src/test/java/org/apache/calcite/plan/RelOptUtilTest.java
+++ b/core/src/test/java/org/apache/calcite/plan/RelOptUtilTest.java
@@ -405,6 +405,15 @@ class RelOptUtilTest {
         relBuilder.literal(true));
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-7013">[CALCITE-7013]
+   * Support building RexLiterals from Character values</a>. */
+  @Test void testCharacterLiteral() {
+    char c = 'c';
+    relBuilder.literal(c);
+  }
+
   private void splitJoinConditionHelper(RexNode joinCond, List<Integer> expLeftKeys,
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/apache/calcite/plan/RelOptUtilTest.java
```diff
@@ -405,6 +405,15 @@ private static Frameworks.ConfigBuilder config() {
         relBuilder.literal(true));
   }
 
+  /**
+   * Test case for
+   * <a href="https://issues.apache.org/jira/browse/CALCITE-7013">[CALCITE-7013]
+   * Literal should support Character basic type</a>. */
+  @Test void testCharacterLiteral() {
```
mihaibudiu, line 412: I am not sure the issue title is suitable; Literals certainly support values of Char type.
There were no suitable constructors taking just a Character value.

xuzifu666: How about change the title to "Support Character basic type in Literal", maybe is more suiatble. @mihaibudiu

mihaibudiu: It doesn't seem a much better fit.
This is really about creating some utility functions to build RexLiterals directly from Character values.

xuzifu666: Yes, you are right. Maybe "Support build RexLiterals from Character values" is more in line with this issue？ @mihaibudiu

mihaibudiu: building
yes

xuzifu666: Done, thank~ @mihaibudiu

PR #4375, opened 2025-05-13
[CALCITE-7010] The well-known count bug

suibianwanwank:
This PR combines an existing implementation of Calcite with the paper [Improving Unnesting of Complex Queries](https://15799.courses.cs.cmu.edu/spring2025/papers/11-unnesting/neumann-btw2025.pdf) to fix the classic count-bug when decorrelated subqueries. For more information, see [CALCITE-7010].
Pseudo-code given in the paper:
```
1 fun unnest(groupby, info, accessing):
2 static = groupby.groups is empty or groups.groupingsets contains ∅
3 unnest(groupby.input, info, accessing)
4 rewriteColumns(groupby, info)
5 for c in info.outerRefs:
6 add info.repr[c] to groupby.groups
7 if static:
8 replace groupby with info.D groupy, joining on mapped info.outerRefs
```
This fix affects a lot of existing test plans, and I haven't yet fully fixed it and evaluated its impact. However, based on the sub-query.iq tests, the results are very promising.
 I'll set this PR to draft until all tests are fixed. Any insight on this would be appreciated!


core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -759,6 +758,68 @@ protected RexNode removeCorrelationExpr(
     }
 
     RelNode newRel = relBuilder.build();
+    // Special case where the group by is static (i.e., a select clause with aggregation functions
```
mihaibudiu, line 761: why "static"?
there really is only a single group containing all elements.
alternatively, this is the same as grouping on 'true' (or any constant).

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -759,6 +758,68 @@ protected RexNode removeCorrelationExpr(
...
+    // but without group by).
+    if (rel.getGroupType() == Aggregate.Group.SIMPLE
+        && rel.getGroupSet().isEmpty()
+        && !frame.corDefOutputs.isEmpty()) {
+      final Pair<CorrelationId, Frame> outerFramePair = requireNonNull(this.frameStack.peek());
+      final Frame outFrame = outerFramePair.right;
+      RexBuilder rexBuilder = relBuilder.getRexBuilder();
+
+      int groupSize = (int) corDefOutputs.keySet().stream()
```
mihaibudiu, line 770: is "keySize" a better name?
or groupKeySize?

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -759,6 +758,68 @@ protected RexNode removeCorrelationExpr(
...
+          .filter(a -> a.corr.equals(outerFramePair.left))
+          .count();
+      List<RelDataTypeField> newRelFields = newRel.getRowType().getFieldList();
+      ImmutableBitSet.Builder corFieldBuilder = ImmutableBitSet.builder();
+      final Map<Integer, RexNode> specificRexNodeMap = new HashMap<>();
```
mihaibudiu, line 775: can you describe the keys and values of this map?
seems to map an index to a field access expression.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -759,6 +758,68 @@ protected RexNode removeCorrelationExpr(
...
+      final List<RexNode> conditions = new ArrayList<>();
+      for (Map.Entry<CorDef, Integer> corDefOutput : corDefOutputs.entrySet()) {
+        if (corDefOutput.getKey().corr.equals(outerFramePair.left)) {
```
mihaibudiu, line 778: I think the code is easier to read if you store coreDefOutput.getKey() and getValue() in separate variables with descriptive names.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -759,6 +758,68 @@ protected RexNode removeCorrelationExpr(
...
+
+      ImmutableBitSet groupSet = corFieldBuilder.build();
+      final RelNode join = relBuilder.push(outFrame.r)
+          .aggregate(relBuilder.groupKey(groupSet))
+          .push(newRel)
+          .join(JoinRelType.LEFT, conditions).build();
+
+      for (int i1 = 0; i1 < rel.getAggCallList().size(); i1++) {
+        AggregateCall aggCall = rel.getAggCallList().get(i1);
+        if (aggCall.getAggregation() instanceof SqlCountAggFunction) {
```
mihaibudiu, line 800: does the argument to count matter? i.e., `*`?
the paper seemed to imply it does

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
```diff
@@ -1781,7 +1781,7 @@ private void checkSemiOrAntiJoinProjectTranspose(JoinRelType type) {
             CoreRules.FILTER_INTO_JOIN,
             CoreRules.PROJECT_MERGE)
         .withRule(CoreRules.PROJECT_TO_SEMI_JOIN)
-        .check();
+        .checkUnchanged();
```
mihaibudiu, line 1784: why are these unchanged now?
This suggests that decorrelation fails in this case.

core/src/test/java/org/apache/calcite/test/SqlHintsConverterTest.java
```diff
@@ -519,48 +519,6 @@ public final Fixture sql(String sql) {
         .check();
   }
 
-  @Test void testHintsPropagationInVolcanoPlannerRules2() {
```
mihaibudiu, line 522: I am not sure it's ok to delete such tests without a very good reason

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -1003,16 +1003,22 @@ and e1.sal > (select avg(e2.sal) from emp e2
...
+    LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], DEPTNO0=[$9], NAME=[$10], DEPTNO00=[CAST($11):INTEGER], $f1=[$12])
       LogicalJoin(condition=[AND(=($9, $11), >($5, $12))], joinType=[inner])
         LogicalJoin(condition=[=($7, $9)], joinType=[inner])
           LogicalTableScan(table=[[CATALOG, SALES, EMP]])
           LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
-        LogicalAggregate(group=[{0}], agg#0=[SINGLE_VALUE($1)])
-          LogicalProject(DEPTNO=[$2], EXPR$0=[$3])
-            LogicalAggregate(group=[{0, 1, 2}], groups=[[{0, 1, 2}, {0, 2}, {1, 2}, {2}]], EXPR$0=[AVG($3)])
-              LogicalProject(COMM=[$6], MGR=[$3], DEPTNO=[$7], SAL=[$5])
+        LogicalProject(DEPTNO0=[$0], $f1=[$2])
```
mihaibudiu, line 1011: were the previous plans incorrect?

mihaibudiu (commented):
I will have to read the paper and the changes more carefully to make truly informed comments.
But if this is a bug, it has to be fixed.
The compiler should either refuse to convert a query or produce a correct result, but never produce an incorrect result from a rewrite.

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
suibianwanwank: From the plan, the original plan was also correct. The root cause of the `count bug` is that during Aggregate processing, the `corVar` field gets added to GROUP BY for deduplication and is used in subsequent JOINs. For `sum(a)`, when the input has 0 rows, it returns null, but for `sum(a) GROUP BY b`, it returns 0 rows. This is what the PR fixes. However, if a parent RelNode returns identical results for null columns and empty rows, we actually don't need extra processing—though currently there's no good way to determine this. Perhaps we could handle this case in the rule.

core/src/test/java/org/apache/calcite/test/SqlHintsConverterTest.java
suibianwanwank: I'm not sure why we're using a correlated query here to test hint passing. maybe I should change the original SQL to a more explicit join.

core/src/test/java/org/apache/calcite/test/RelOptRulesTest.java
suibianwanwank: It appears the SemiJoinRule is being tested here. After the new decorrelation plan, this rule was not successfully applied. I'll try to research this more thoroughly.

caicancai:
So cool, I will read the paper and review it.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
suibianwanwank: Actually, this isn't exactly the same:
```
select sum(i0) from  (select i0 from test where i1 = 1) a

sum |
----+
null|
(1 rows)
```

```
select sum(i0) from  (select i0 from test where i1 = 1) a group by true

sum|
---+
(0 rows)
```

suibianwanwank:
> Can we revive this PR?

I think it's ready~

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -18919,13 +18922,16 @@ LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$
     <Resource name="planAfter">
       <![CDATA[
 LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8])
-  LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], m=[CAST($9):INTEGER], c=[CAST($10):BIGINT], d=[CAST($11):BIGINT], trueLiteral=[CAST($12):BOOLEAN], NAME=[CAST($13):VARCHAR(10)])
+  LogicalProject(EMPNO=[$0], ENAME=[$1], JOB=[$2], MGR=[$3], HIREDATE=[$4], SAL=[$5], COMM=[$6], DEPTNO=[$7], SLACKER=[$8], m=[$9], c=[CAST($10):BIGINT], d=[CAST($11):BIGINT], trueLiteral=[$12], JOB0=[CAST($13):VARCHAR(10)])
```
mihaibudiu: this plan and the previous one are changed in a strange way: I would not expect the column names to change

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -160,6 +162,8 @@ public class RelDecorrelator implements ReflectiveVisitor {
   // map built during translation
   protected CorelMap cm;
 
+  protected final Deque<Pair<CorrelationId, Frame>> frameStack = new ArrayDeque<>();
```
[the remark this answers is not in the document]
mihaibudiu: well, thumbs up, but the description is still missing

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -476,14 +480,16 @@ protected RexNode removeCorrelationExpr(
   }
 
   /** Fallback if none of the other {@code decorrelateRel} methods match. */
-  public @Nullable Frame decorrelateRel(RelNode rel, boolean isCorVarDefined) {
+  public @Nullable Frame decorrelateRel(RelNode rel, boolean isCorVarDefined,
```
mihaibudiu, line 507: can you describe this new argument in one of the javadocs?
what does it mean that the parent propagates null values? 
All or some?
And where does it propagate them?

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -942,19 +1022,26 @@ private static void shiftMapping(Map<Integer, Integer> mapping, int startIndex,
...
-  public @Nullable Frame decorrelateRel(Project rel, boolean isCorVarDefined) {
+  public @Nullable Frame decorrelateRel(Project rel, boolean isCorVarDefined,
+      boolean parentPropagatesNullValues) {
     //
     // Rewrite logic:
     //
     // 1. Pass along any correlated variables coming from the input.
     //
+    for (RexNode project : rel.getProjects()) {
+      if (!Strong.isStrong(project)) {
```
mihaibudiu, line 1178: it this function always correct, or is Strong a conservative approximation?
is this correct if strong is only conservative?
what if none of the inputs is nullable, does it matter?

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
```diff
@@ -7378,20 +7378,6 @@ private void checkLiteral2(String expression, String expected) {
         callsUnparseCallOnSqlSelect[0], is(true));
   }
 
-  @Test void testCorrelate() {
```
mihaibudiu, line 7508: why delete this test?

core/src/test/java/org/apache/calcite/sql2rel/RelDecorrelatorTest.java
```diff
@@ -164,17 +164,23 @@ public static Frameworks.ConfigBuilder config() {
...
         + "      LogicalProject(DEPTNO=[$7], SAL=[$5])\n"
         + "        LogicalTableScan(table=[[scott, EMP]])\n"
-        + "    LogicalAggregate(group=[{0}], EXPR$0=[AVG($1)])\n"
-        + "      LogicalProject(DEPTNO=[$0], TOTAL=[$1])\n"
-        + "        LogicalAggregate(group=[{0}], TOTAL=[SUM($1)])\n"
-        + "          LogicalProject(DEPTNO=[$0], SAL=[$1])\n"
-        + "            LogicalFilter(condition=[IS NOT NULL($0)])\n"
-        + "              LogicalProject(DEPTNO=[$7], SAL=[$5])\n"
-        + "                LogicalTableScan(table=[[scott, EMP]])\n";
+        + "    LogicalProject(DEPTNO=[$0], EXPR$0=[$2])\n"
```
mihaibudiu, line 276: reviewing such plans is not easy...

core/src/test/java/org/apache/calcite/rel/rel2sql/RelToSqlConverterTest.java
suibianwanwank: Keeping this test seems pointless, as it attempts to make decorrelate fail using value, but this PR already supports decorrelation for value.

core/src/test/java/org/apache/calcite/sql2rel/RelDecorrelatorTest.java
suibianwanwank: I totally agree —

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
suibianwanwank: For aggregates marked as "STATIC,"  we can always ensure correctness by adding the `join` introduced in this PR, but this would complicate all plans.
The `propagatesNullValues` parameter presents an optimization opportunity—it allows us to avoid unnecessary joins in such scenarios. Therefore, even if `Strong` is conservative, we would only miss out on the corresponding optimization opportunities.
Additionally, this parameter is inspired by DuckDB's implementation. I will add more comments to clarify its usage.

core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
suibianwanwank: Fixed, I kept the aliases.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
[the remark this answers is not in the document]
suibianwanwank: Done. Thanks!

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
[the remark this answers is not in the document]
suibianwanwank: Sorry for missing that, added now.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -760,6 +766,136 @@ protected RexNode removeCorrelationExpr(
...
+    // To handle this situation, we ensure aggregated result output through pre-join
+    // Given the following plan:
+    //    LogicalCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{0}])
+    //      LogicalProject(DEPTNO=[$0])
+    //        LogicalTableScan(table=[[scott, DEPT]])
+    //      LogicalProject(EXPR$1=[IS NULL($1)])
+    //        LogicalFilter(condition=[=(0, $0)])
+    //          LogicalAggregate(group=[{}], EXPR$0=[COUNT()], EXPR$1=[SUM($0)])
+    //            LogicalFilter(condition=[=($cor0.DEPTNO, $7)])
+    //              LogicalTableScan(table=[[scott, EMP]])
```
iwanttobepowerful: plz add sql stmt for ref

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -760,6 +766,136 @@ protected RexNode removeCorrelationExpr(
...
+    //          LogicalAggregate(group=[{0}], EXPR$0=[COUNT()], EXPR$1=[SUM($0)])
+    //            LogicalProject(DEPTNO=[$7])
+    //              LogicalFilter(condition=[IS NOT NULL($7)])
+    //                LogicalTableScan(table=[[scott, EMP]])
+    // It will causes rows with `count=0` to be filtered out, and IS NULL($2) will return null
+    // instead of true. Therefore, we use LEFT JOIN to ensure that
+    // correlation fields (extra group by key) always returns the aggregation result.
+    //
+    // Rewrite Aggregate as:
+    //    LogicalProject(DEPTNO=[$0], EXPR$0=[CASE(IS NOT NULL($2), $2, 0), EXPR$1=[$(1)])
```
iwanttobepowerful: maybe parenthesis '[' not matching

suibianwanwank:
Thanks for the review! I've updated some comments and added statements for clarity. PTAL @mihaibudiu @iwanttobepowerful .

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -760,6 +766,177 @@ protected RexNode removeCorrelationExpr(
...
+    // Background:
+    //   For the query:
+    //     SELECT SUM(salary), COUNT(name) FROM A;
+    //   When table A is empty, it returns [null, 0].
+    //   But for
+    //     SELECT SUM(salary), COUNT(name) FROM A group by id
+    //   When table A is empty, it returns empty. This causes result mismatch.
+    // In the general decorrelation framework, we add corVar as an additional groupKey to
+    // rewrite Correlate as JOIN. (See the code above for details) This means that when the input
+    // is empty, the default return value of aggregate functions without a GROUP BY clause
```
mihaibudiu: I would not say that the result is ignored; there simply is no result.
I would change the comment to say: thus the result produced using a JOIN is incorrect.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -760,6 +766,177 @@ protected RexNode removeCorrelationExpr(
...
+    //      LogicalCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{0}])
+    //        LogicalProject(DEPTNO=[$0])
+    //          LogicalTableScan(table=[[scott, DEPT]])
+    //        LogicalProject(cs=[true])
+    //          LogicalFilter(condition=[=(0, $0)])
+    //            LogicalAggregate(group=[{}], EXPR$0=[COUNT()])
+    //              LogicalFilter(condition=[=($cor0.DEPTNO, $7)])
+    //                LogicalTableScan(table=[[scott, EMP]])
+    //
+    // The regular rewrite as:
```
mihaibudiu: Rewriting this as:
...
produces an incorrect result.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -760,6 +766,177 @@ protected RexNode removeCorrelationExpr(
...
+    //          LogicalTableScan(table=[[scott, DEPT]])
+    //        LogicalProject(cs=[true], DEPTNO=[$0])
+    //          LogicalFilter(condition=[=(0, $1)])
+    //            LogicalAggregate(group=[{0}], EXPR$0=[COUNT()])
+    //              LogicalProject(DEPTNO=[$7])
+    //                LogicalFilter(condition=[IS NOT NULL($7)])
+    //                  LogicalTableScan(table=[[scott, EMP]])
+    //  We can clearly observe that due to the presence of the GROUP BY clause,
+    //  COUNT(*) = 0 will never evaluate to true, since rows with zero records won't appear
+    //  in the GROUP BY results. This contradicts our intended semantics.
```
mihaibudiu: you can't really "contradict" the semantics; you just produce an incorrect result.
(you are not respecting the semantics; and it is not "our" semantics, it is the SQL semantics).

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -760,6 +766,177 @@ protected RexNode removeCorrelationExpr(
...
+    // In the general decorrelation framework, we add corVar as an additional groupKey to
+    // rewrite Correlate as JOIN. (See the code above for details) This means that when the input
+    // is empty, the default return value of aggregate functions without a GROUP BY clause
+    // is ignored. This leads to inconsistent results.
+    //
+    // We refer to this situation as: `The well-known count bug`,
+    // More details about this issue: Optimization of Nested SQL Queries Revisited
+    // (https://dl.acm.org/doi/pdf/10.1145/38714.38723)
+    //
+    // To handle this situation, we ensure aggregated result output through pre-join
```
mihaibudiu: the terminology of pre-join is confusing; perhaps you mean you are using a LEFT JOIN to ensure that an output is always produced.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -760,6 +766,177 @@ protected RexNode removeCorrelationExpr(
...
+    //  COUNT(*) = 0 will never evaluate to true, since rows with zero records won't appear
+    //  in the GROUP BY results. This contradicts our intended semantics.
+    //
+    // Rewrite Aggregate as:
+    //   SELECT d.deptno FROM dept d
+    //     JOIN (
+    //          SELECT true AS cs, d2.deptno
+    //          FROM (
+    //              SELECT d2.deptno,
+    //                     CASE WHEN cnt0 IS NOT NULL THEN cnt0 ELSE 0 END AS cnt
```
mihaibudiu: so this rewrites NULL (empty set) to 0?

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -760,6 +766,177 @@ protected RexNode removeCorrelationExpr(
...
+    //                LogicalFilter(condition=[IS NOT NULL($7)])
+    //                  LogicalTableScan(table=[[scott, EMP]])
+    //  We can clearly observe that due to the presence of the GROUP BY clause,
+    //  COUNT(*) = 0 will never evaluate to true, since rows with zero records won't appear
+    //  in the GROUP BY results. This contradicts our intended semantics.
+    //
+    // Rewrite Aggregate as:
+    //   SELECT d.deptno FROM dept d
+    //     JOIN (
+    //          SELECT true AS cs, d2.deptno
```
mihaibudiu: had to remove this "d2" to make this program compile

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -760,6 +766,177 @@ protected RexNode removeCorrelationExpr(
...
+          newProjectMap.put(corIndex + groupKeySize, left);
+          conditions.add(
+              relBuilder.isNotDistinctFrom(left,
+                  new RexInputRef(corIndex + groupKeySize,
+                      newRelFields.get(corIndex).getType())));
+        }
+      }
+
+      ImmutableBitSet groupSet = corFieldBuilder.build();
+      final RelNode join = relBuilder.push(outFrame.r)
```
mihaibudiu: you can annotate the code here with the line from the above plan that is being constructed

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -760,6 +766,177 @@ protected RexNode removeCorrelationExpr(
...
+    //                  LEFT JOIN (
+    //                      SELECT * FROM emp WHERE deptno IS NOT NULL ) e
+    //                  ON d2.deptno IS NOT DISTINCT FROM e.deptno
+    //                  GROUP BY d2.deptno
+    //              ) AS left_join
+    //          ) AS case_count
+    //          WHERE cnt = 0
+    //     ) AS d0 ON d.deptno = d0.deptno
+    // Corresponding plan:
+    //    LogicalProject(DEPTNO=[$0])
```
mihaibudiu: I suggest numbering these lines - see below.

mihaibudiu (commented):
I think I am starting to understand what's going on; made some suggestions on making the code easier to read. I will review again.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
suibianwanwank: Yes, to be more precise, for any join key in subsequent join, `count` needs to return 0, and other aggregate functions must return NULL. (NULL is also not the same as an empty set—if there's a filter like `is not null`, so the count-bug isn't limited to just `count`.)

suibianwanwank:
@mihaibudiu I've addressed all the comments. I modified it into an additional method because the excessive number of lines in the method caused the CI to fail (though most of them were comments).

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -760,11 +766,198 @@ protected RexNode removeCorrelationExpr(
...
+   *   SELECT deptno FROM dept d
+   *     WHERE 0 = (SELECT COUNT(*) FROM emp e WHERE d.deptno = e.deptno)
+   * Corresponding plan:
+   *    LogicalProject(DEPTNO=[$0])
+   *      LogicalCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{0}])
+   *        LogicalProject(DEPTNO=[$0])
+   *          LogicalTableScan(table=[[scott, DEPT]])
+   *        LogicalProject(cs=[true])
+   *          LogicalFilter(condition=[=(0, $0)])
+   *            LogicalAggregate(group=[{}], EXPR$0=[COUNT()])
```
mihaibudiu, line 840: This rel node is the input to rewriteScalarAggregate.
It would be handy to point this out.

suibianwanwank: Done.

core/src/test/resources/sql/scalar.iq
```diff
@@ -301,6 +301,7 @@ from (values (1), (3)) t1(id);
 
 !ok
 
+!if (false) {
```
suibianwanwank: @mihaibudiu I merged main and squashed the commits. Unfortunately, this test failed.

Before this PR, Value couldn't be decorrelated,so this test wouldn't cause an error. but I believe this issue is unrelated to the current PR because replacing `Value` with a `tableScan` would still cause the same error. Additionally, I'm confident this issue can be fixed by [CALCITE-7045], and I'll start working on it after this PR is merged.

mihaibudiu: yes, it seems that main fails too

suibianwanwank:
If there are no other comments, I will merge this PR today.

core/src/main/java/org/apache/calcite/sql2rel/RelDecorrelator.java
```diff
@@ -782,11 +788,198 @@ protected RexNode removeCorrelationExpr(
...
+   * [03]     LogicalProject(DEPTNO=[$0])
+   * [04]       LogicalTableScan(table=[[scott, DEPT]])
+   * [05]     LogicalProject(cs=[true], DEPTNO=[$0])
+   * [06]       LogicalFilter(condition=[=(0, $1)])
+   * [07]         LogicalProject(DEPTNO=[$0], EXPR$0=[CASE(IS NOT NULL($2), $2, 0)])
+   * [08]           LogicalJoin(condition=[IS NOT DISTINCT FROM($0, $1)], joinType=[left])
+   * [09]             LogicalAggregate(group=[{0}])
+   * [10]               LogicalProject(DEPTNO=[$0])
+   * [11]                 LogicalTableScan(table=[[scott, DEPT]])
+   * [12]             LogicalAggregate(group=[{0}], EXPR$0=[COUNT()])
```
iwanttobepowerful, line 895: Does this include the optimization of aggregation pushdown? From the perspective of intuitiveness and ensuring a one-to-one correspondence between the code and the execution plan, I think the plan posted here should be the one without the aggregation pushdown optimization.

suibianwanwank: No, this does not include any optimizations.  I guess the following comment explains the issue you're wondering about:
```
    // Build [09] LogicalAggregate(group=[{0}]) to obtain the distinct set of
    // corVar from outFrame.
```
The Aggregate here is for distinct purposes.

suibianwanwank: My oversight, thanks. I've adjusted the SQL stmt.

PR #4378, opened 2025-05-14
[CALCITE-7009] AssertionError when converting query containing multiple correlated subqueries referencing different tables in FROM

korlov42:
https://issues.apache.org/jira/browse/CALCITE-7009

This patch addresses an issue when AssertionError is thrown if multiple correlated subqueries reference different tables in FROM list. Proposed solution is to simply remove assertion check. See attached JIRA for details.

### DIFF
--- a/core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
@@ -5010,6 +5010,30 @@ LogicalProject(ENAME=[$cor0.ENAME])
 ]]>
     </Resource>
   </TestCase>
+  <TestCase name="testMultipleCorrelatedSubQueriesInSelectReferencingDifferentTablesInFrom">
+    <Resource name="sql">
+      <![CDATA[select
+(select ename from emp where empno = empnos.empno) as emp_name,
+(select name from dept where deptno = deptnos.deptno) as dept_name
+ from (values (1), (2)) as empnos(empno), (values (1), (2)) as deptnos(deptno)]]>
+    </Resource>
+    <Resource name="plan">
+      <![CDATA[
+LogicalProject(EMP_NAME=[$SCALAR_QUERY({
+LogicalProject(ENAME=[$1])
+  LogicalFilter(condition=[=($0, $cor0.EMPNO)])
+    LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+})], DEPT_NAME=[$SCALAR_QUERY({
+LogicalProject(NAME=[$1])
+  LogicalFilter(condition=[=($0, $cor0.DEPTNO)])
+    LogicalTableScan(table=[[CATALOG, SALES, DEPT]])
+})])
+  LogicalJoin(condition=[true], joinType=[inner])
+    LogicalValues(tuples=[[{ 1 }, { 2 }]])
+    LogicalValues(tuples=[[{ 1 }, { 2 }]])
+]]>
+    </Resource>
+  </TestCase>
   <TestCase name="testMultiset">
     <Resource name="plan">
       <![CDATA[
--- a/core/src/test/resources/sql/scalar.iq
+++ b/core/src/test/resources/sql/scalar.iq
@@ -301,4 +301,21 @@ from (values (1), (3)) t1(id);
 
 !ok
 
+# Several scalar sub-queries reference different tables in FROM list
+select
+    (select ename from emp where empno = empnos.empno) as emp_name,
+    (select dname from dept where deptno = deptnos.deptno) as dept_name
+  from (values (7369), (7499)) as empnos(empno), (values (10), (20)) as deptnos(deptno) order by 1, 2;
++----------+------------+
+| EMP_NAME | DEPT_NAME  |
++----------+------------+
+| ALLEN    | ACCOUNTING |
+| ALLEN    | RESEARCH   |
+| SMITH    | ACCOUNTING |
+| SMITH    | RESEARCH   |
++----------+------------+
+(4 rows)
+
+!ok
+
 # End scalar.iq
--- a/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
+++ b/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java
@@ -3110,12 +3110,6 @@ public class SqlToRelConverter {
     final ImmutableBitSet.Builder requiredColumns = ImmutableBitSet.builder();
     final List<CorrelationId> correlNames = new ArrayList<>();
 
-    // All correlations must refer the same namespace since correlation
-    // produces exactly one correlation source.
-    // The same source might be referenced by different variables since
-    // DeferredLookups are not de-duplicated at create time.
-    SqlValidatorNamespace prevNs = null;
-
     for (CorrelationId correlName : correlatedVariables) {
       DeferredLookup lookup =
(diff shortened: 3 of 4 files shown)


mihaibudiu:
These changes were introduced by @vlsi . It would be great if he could review this.

zstan (approved):

core/src/test/resources/sql/scalar.iq
```diff
@@ -301,4 +301,21 @@ from (values (1), (3)) t1(id);
...
+  from (values (7369), (7499)) as empnos(empno), (values (10), (20)) as deptnos(deptno) order by 1, 2;
++----------+------------+
+| EMP_NAME | DEPT_NAME  |
++----------+------------+
+| ALLEN    | ACCOUNTING |
+| ALLEN    | RESEARCH   |
+| SMITH    | ACCOUNTING |
+| SMITH    | RESEARCH   |
++----------+------------+
+(4 rows)
```
asolimando, line 317:
```
CREATE TABLE emp (
  empno INT,
  ename VARCHAR(100)
);

CREATE TABLE dept (
  deptno INT,
  dname VARCHAR(100)
);

INSERT INTO dept VALUES (1, 'ACCOUNTING');
INSERT INTO dept VALUES (2, 'RESEARCH');
INSERT INTO emp VALUES (7369, 'ALLEN');
INSERT INTO emp VALUES (7499, 'SMITH');

select
  (select ename from emp where empno = empnos.empno) as emp_name,
  (select dname from dept where deptno = deptnos.deptno) as dept_name
from (values (7369), (7499)) as empnos(empno), 
     (values (1), (2)) as deptnos(deptno)
order by 1, 2;
```

validated with [sqlfiddle](https://sqlfiddle.com/postgresql/online-compiler?id=406f65d5-6b88-4df5-a531-9d7dedc676f5) with PostgreSQL and it's giving the same result

asolimando (approved):
LGTM

core/src/test/resources/sql/scalar.iq
mihaibudiu: I would add this as a comment to the iq file

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
```diff
@@ -5007,6 +5007,30 @@ LogicalProject(ENAME=[$cor0.ENAME])
 }), 0)])
   LogicalProject(DEPTNO=[$7], ENAME=[$1])
     LogicalTableScan(table=[[CATALOG, SALES, EMP]])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testMultipleCorrelatedSubQueriesInSelectReferencingDifferentTablesInFrom">
```
mihaibudiu, line 5013: I thought `$cor` variables are introduced by Correlate nodes...
I guess they can also be introduced by `$SCALAR_QUERY` nodes?
Since the same name may be reused, I wonder whether there can arise ambiguity when the correlated queries are actually nested.

korlov42: I tried to add such case locally in my branch it it works just fine on `SqlToRelConverterTest`, but it fails with `Caused by: java.lang.NullPointerException: cm.mapCorToCorRel.get($cor0)` in `QuidemTest`.  I tried to add following test case:
```
select
    (select ename || ' from dept '
                  || (select dname from dept where deptno = emp.deptno and emp.empno = empnos.empno)
       from emp
    ) as ename_from_dept
  from (values (7369), (7499)) as empnos(empno) order by 1;
```

asolimando: I feel this requires some more thinking and it should be postponed to 1.41.0, what do you think @korlov42?

korlov42: One argument to include this patch in the upcoming release is that it does not affect any production, because normally productions are run without assertions enabled. It would be nice to be able to run the same queries on dev environments which works in production.
If this reasoning still doesn't convince you this patch is safe for the release, I'm ok to postpone it till the next one.

asolimando: The only instance of that failure I could fine is in CALCITE-4913, which touches the same area but doesn't seem too related, I suggest to file a dedicated issue for the bug.

I'd like to take a further look at all this, try the same query on postgres etc. etc. 

Let's have this ready for 1.41.0, thanks for your understanding!

korlov42:
folks, I'm wondering how I can facilitate the review of this patch?

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
mihaibudiu: have you filed an issue about the query that fails to compile?
I assume it's independent on this fix.

korlov42: @mihaibudiu here the JIRA for problem with nested correlated subqueries: https://issues.apache.org/jira/browse/CALCITE-7057. It can be reproduced on latest main (I've checked on commit 0dd30b5c), so it's not related to this patch.

mihaibudiu:
This PR already has 2 reviews, but @asolimando said that he wants to take another look. He did already approve it.

core/src/test/resources/org/apache/calcite/test/SqlToRelConverterTest.xml
asolimando: with the following DDL:
```CREATE TABLE emp (
  empno INT,
  ename VARCHAR(100),
  deptno INT
);

CREATE TABLE dept (
  deptno INT,
  dname VARCHAR(100)
);

INSERT INTO dept VALUES (1, 'ACCOUNTING');
INSERT INTO dept VALUES (2, 'RESEARCH');
INSERT INTO emp VALUES (7369, 'ALLEN', 1);
INSERT INTO emp VALUES (7499, 'SMITH', 2);
```

your query fails with `Query Error: more than one row returned by a subquery used as an expression` with Postgres 17 (tested on https://www.db-fiddle.com/).

I think the following query expresses what you wanted to do and it's valid according to Postgres:
```
select
    (select ename || ' from dept '
                  || (
                    select dname 
                    from dept 
                    where deptno = emp.deptno)
     from emp
     where emp.empno = empnos.empno) as ename_from_dept
  from (values (7369), (7499)) as empnos(empno)
  order by 1;
```

@korlov42 could you please check if that one is working at runtime too?

I understand that we are aiming at accepting more valid queries, but we might be uncovering some bugs that were hidden before, that's not a problem as long as we don't cause regressions and we file tickets for the problems we find along the way.

asolimando:
> This PR already has 2 reviews, but @asolimando said that he wants to take another look. He did already approve it.

Apologies for the delay, I am fine with the PR, I just asked @korlov42 to do a final check around the case we found it's failing, the outcome doesn't change my approval of the PR, but I generally try to condition merging upon the completion of related due diligence tasks, so we don't forget about them.

mihaibudiu:
I think that @korlov42 filed https://issues.apache.org/jira/browse/CALCITE-7057; that was my question too.
If there are no other comments I plan to merge this today.

mihaibudiu:
@korlov42 I am not sure what happened, but after merging into main the test from misc.iq fails.
I will comment it out and file an issue for you to look into this.

PR #4379, opened 2025-05-14
[CALCITE-7016] Sql resources should not be included in XML when using relFn in RelOptRulesTest testing

xiedeyantu:
JIRA link: [CALCITE-7016](https://issues.apache.org/jira/browse/CALCITE-7016)

### DIFF
--- a/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
+++ b/core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
@@ -1228,11 +1228,6 @@ LogicalTableScan(table=[[scott, EMP]])
     </Resource>
   </TestCase>
   <TestCase name="testAntiJoinProjectTranspose">
-    <Resource name="sql">
-      <![CDATA[select a.name from dept a
-where a.deptno not in (select b.deptno * 2 from dept);
-]]>
-    </Resource>
     <Resource name="planBefore">
       <![CDATA[
 LogicalProject(DNAME=[$1])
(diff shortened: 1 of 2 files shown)


core/src/test/resources/org/apache/calcite/test/RelOptRulesTest.xml
```diff
@@ -1228,11 +1228,6 @@ LogicalTableScan(table=[[scott, EMP]])
     </Resource>
   </TestCase>
   <TestCase name="testAntiJoinProjectTranspose">
-    <Resource name="sql">
```
NobiGo, line 1231: Is this SQL incorrect? Would it be better to add "rel" instead, for example:
```
<Resource name="rel">
 <![CDATA[
RelString.....
]]>
    </Resource>
```

xiedeyantu: Not sure if it is correct, it may be correct. But there is no direct connection between sql and plan, it is easy to make mistakes. I agree with your suggestion very much, using `rel` resource may be a great idea.

xiedeyantu: I checked and `planBefore` should be the `rel` we want. Does that mean we don't need to add a `rel` resource?

NobiGo: Yes. This SQL improves readability, but it does not perform validation, which is a risk.

xiedeyantu: @NobiGo Agree, we can wait for others' opinions. But there's an issue here: if the expected plan can be generated using SQL, should RelFn not be used? If RelFn is used, can it still be mapped to valid SQL? However, I'm unsure whether such a restriction should be imposed on developers.

PR #4328, opened 2025-04-23
CALCITE-5638 Assertion Failure during planning correlated query with orderby

zstan:
LogicalJoin can contain sub-query which correlated inputs are not considered in further LogicalJoin fields trimming

subquery: $SCALAR_QUERY with correlate
```
  LogicalProject(DEPTNO=[$1])
    LogicalFilter(condition=[=(CAST($0):CHAR(11) NOT NULL, $cor0.DNAME)])
```

```
                       ---------------------------LogicalJoin ----------------------
                       V                                                            V  
                     left                                                         right
                       |                                                            |
LogicalProject.NONE.[0, 1]                                             LogicalValues.NONE.[0]
RecordType(INTEGER DEPTNO, CHAR(11) DNAME)                             RecordType(INTEGER DEPTNO)
```

In such a case  $cor0.DNAME - has not been considered which leads to mention assertion

PR #4382, opened 2025-05-15
[CALCITE-7017] Incorrect subquery rewriting in join conditions when j…

suibianwanwank:
…oin type is outer join

### DIFF
--- a/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
+++ b/core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
@@ -30,7 +30,6 @@ import org.apache.calcite.rel.core.JoinRelType;
 import org.apache.calcite.rel.core.Project;
 import org.apache.calcite.rel.metadata.RelMdUtil;
 import org.apache.calcite.rel.metadata.RelMetadataQuery;
-import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.LogicVisitor;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexCorrelVariable;
@@ -739,8 +738,7 @@ public class SubQueryRemoveRule
           CorrelationId newCorId = builder.getCluster().createCorrel();
           CorrelationId originCorId = Iterables.getOnlyElement(variablesSet);
           RelNode newRel =
-              replaceCorrelationId(builder.getRexBuilder(), builder.build(),
-                  originCorId, newCorId, builder.peek().getRowType());
+              replaceCorrelationId(builder.getRexBuilder(), builder.build(), originCorId, newCorId);
           builder.push(newRel);
           builder.as(ctAlias);
           builder.join(JoinRelType.LEFT, trueLiteral, ImmutableSet.of(newCorId));
@@ -859,17 +857,17 @@ public class SubQueryRemoveRule
 
   /**
    * Replace every {@link RexFieldAccess} in the given {@link RelNode}
-   * that reference the specified {@link CorrelationId} with another {@link CorrelationId}.
+   * that references the specified {@link CorrelationId} with another {@link CorrelationId}.
    */
   public static RelNode replaceCorrelationId(RexBuilder rexBuilder, RelNode node,
-      final CorrelationId from, final CorrelationId to, RelDataType type) {
+      final CorrelationId from, final CorrelationId to) {
     return node.accept(new RelHomogeneousShuttle() {
       @Override public RelNode visit(RelNode other) {
         other = other.accept(new RexShuttle() {
           @Override public RexNode visitFieldAccess(RexFieldAccess fieldAccess) {
             if (fieldAccess.getReferenceExpr() instanceof RexCorrelVariable
                 && ((RexCorrelVariable) fieldAccess.getReferenceExpr()).id.equals(from)) {
-              RexNode correl = rexBuilder.makeCorrel(type, to);
+              RexNode correl = rexBuilder.makeCorrel(fieldAccess.getReferenceExpr().getType(), to);
               return rexBuilder.makeFieldAccess(correl, fieldAccess.getField().getIndex());
             }
             return fieldAccess;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -926,6 +923,11 @@ private static void matchJoin(SubQueryRemoveRule rule, RelOptRuleCall call) {
     if (inputIntersectsLeftSide) {
       builder.push(join.getLeft());
 
+      final RelOptUtil.Logic logic =
```
mihaibudiu, line 926: this code could use a little comment.
(I am not familiar with this part of the code...)

core/src/test/resources/sql/sub-query.iq
```diff
@@ -848,6 +848,38 @@ where sal + 100 not in (
...
 
+SELECT e1.*
+FROM emp e1
+LEFT JOIN dept d
+  ON e1.deptno = d.deptno
+  AND e1.empno IN (
+    SELECT e3.empno
+    FROM emp e3
+    WHERE e1.comm > e3.comm
+  )
```
iwanttobepowerful, line 885:
```sql
SELECT e1.*
FROM emp e1
LEFT JOIN dept d
  ON e1.deptno = d.deptno
  AND e1.empno not IN (    -- not
    SELECT e3.empno
    FROM emp e3
    WHERE e1.comm > e3.comm
  )
ORDER BY e1.empno, e1.deptno;
```
maybe can add this case, and show plans diff if have

xiedeyantu: I may not understand the purpose of the test `not in`. Do you want to add a case where `inputIntersectsLeftSide=false`?

asolimando: @suibianwanwank, #4328 has been merged, can you try again?

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -946,6 +948,11 @@ private static void matchJoin(SubQueryRemoveRule rule, RelOptRuleCall call) {
       builder.push(join.getLeft());
       builder.push(join.getRight());
 
+      final RelOptUtil.Logic logic =
```
asolimando, line 951: The two logic variables in the `if/else` seem identical, in case let's move it to its original location in its updated form

asolimando:
> While working on [[CALCITE-7034](https://issues.apache.org/jira/browse/CALCITE-7034)], I encountered some obstacles and might need more time to complete it. I'm inclined to defer it to 1.41.0.

That's totally fine, thanks for letting me know, I will remove the `fixVersion` from the ticket

suibianwanwank:
Thank you all for your comments and reviews. I've refactored this PR to address more related issues here. Although they are not directly connected, they were all exposed and affecting each other within the same test. Given that the code changes required are minimal, I'm inclined to resolve them together in this PR.  

For this SQL.

```
SELECT e1.*, d.deptno
FROM emp e1
LEFT JOIN dept d
  ON e1.deptno = d.deptno
  AND d.deptno IN (
    SELECT e3.deptno
    FROM emp e3
    WHERE d.deptno > e3.comm
  );
```
The current code has the following issues:  

   1. This PR addresses the problem described in the Jira ticket, where an outer join produces output even when there are no matching rows.  
   2. The previous handling of the right side of the join was incorrect—it requires not only `shiftFieldAccess` but also shifting `subQuery.getOperands()`.  
   3. In `rewriteIn`, if `variablesSet` is not empty, `builder.join` generates multiple `Correlate` nodes with the same `corId`, which leads to decorrelation failure (I believe this results in an invalid `RelNode`).

mihaibudiu:
there is one failure on MacOS
```
 FAILURE  89.6sec, org.apache.calcite.test.CoreQuidemTest > test(String)[25], [25] sql/spatial.iq
    org.opentest4j.AssertionFailedError: Files differ: /Users/runner/work/calcite/calcite/core/build/quidem/test/sql/spatial.iq /Users/runner/work/calcite/calcite/core/build/resources/test/sql/spatial.iq
    1847c1847
    < LINESTRING (0 0, 1 1, 0 0, 1 1, 2 2)
    ---
    > LINESTRING (NaN 0, 1 1, 0 0, 1 1, 2 2)
```

mihaibudiu:
The CI failure does not make any sense, maybe the CI machine is broken?

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -840,6 +857,29 @@ private static List<RexNode> fields(RelBuilder builder, int fieldCount) {
     return projects;
   }
 
+  /**
+   * Replace every {@link RexFieldAccess} in the given {@link RelNode}
+   * that reference the specified {@link CorrelationId} with another {@link CorrelationId}.
```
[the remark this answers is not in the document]
suibianwanwank: I noticed that the type parameter can probably be left out, I'll try to modify.

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -960,13 +1001,45 @@ private static void matchJoin(SubQueryRemoveRule rule, RelOptRuleCall call) {
...
 
+      final RelOptUtil.Logic logic =
+          LogicVisitor.find(join.getJoinType().generatesNullsOnLeft()
+                  ? RelOptUtil.Logic.TRUE_FALSE_UNKNOWN : RelOptUtil.Logic.TRUE,
+              ImmutableList.of(join.getCondition()), e);
+
+      RexSubQuery subQuery = e;
+
+      if (!variablesSet.isEmpty()) {
+        // Original correlates reference joint row type, but we are about to create
```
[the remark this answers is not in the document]
suibianwanwank: I added a plan, but the fieldAccess index wasn't actually printed in the plan. At least showing the plan before and after rewriting helps understand the offset reason.

suibianwanwank:
Thank you for reviewing. All issues mentioned in the comments have been resolved.

core/src/main/java/org/apache/calcite/rel/rules/SubQueryRemoveRule.java
```diff
@@ -840,6 +857,29 @@ private static List<RexNode> fields(RelBuilder builder, int fieldCount) {
     return projects;
   }
 
+  /**
+   * Replace every {@link RexFieldAccess} in the given {@link RelNode}
+   * that reference the specified {@link CorrelationId} with another {@link CorrelationId}.
+   */
+  public static RelNode replaceCorrelationId(RexBuilder rexBuilder, RelNode node,
+      final CorrelationId from, final CorrelationId to, RelDataType type) {
+    return node.accept(new RelHomogeneousShuttle() {
```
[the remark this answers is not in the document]
mihaibudiu: I am skeptical, because there is no way the RexVisitor can call your homogeneous shuttle back, since it does not have a handle to it.
Can there be subqueries in expressions at this point, or are you assuming they have been expanded?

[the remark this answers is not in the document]
suibianwanwank: You're right, now I understand what you meant. I've made the corresponding changes, but currently this rule's implementation has other issues in multi-subQuery scenarios. I might need to spend some more time on it.

mihaibudiu:
So this is now ready for review?
Should we first merge the well-known count bug fix?
Or are these completely independent?

suibianwanwank:
> Should we first merge the well-known count bug fix? Or are these completely independent?

Yes, we should merge the well-known count bug fix first. This fix has no dependencies. 

I want to wait for [CALCITE-7045] to be completed, because the current PR addresses some issues from [CALCITE-7045] to fix the tests (even though this is unrelated to the current issue), but I don't think this is the fundamental solution.

PR #4388, opened 2025-05-18
[CALCITE-7026] Validation of snapshot in case of CTEs should fail with better message than NPE

snuyanzin:


core/src/main/java/org/apache/calcite/sql/validate/SqlValidatorImpl.java
```diff
@@ -5874,6 +5874,11 @@ private void validateSnapshot(
         throw newValidationError(period,
             Static.RESOURCE.illegalExpressionForTemporal(dataType.getSqlTypeName().getName()));
       }
+      if (ns instanceof IdentifierNamespace && ns.resolve() instanceof WithItemNamespace) {
+        // If the snapshot is used over a CTE, then we don't have a concrete underlying
+        // table to operate on. This will be rechecked later in the planner rules.
```
mihaibudiu, line 5879: there is no guarantee that any planner rules will be executed after validation.

snuyanzin: i think in general you are right
I backported it from Flink where it is guaranteed...

Any proposal how to proceed with Calcite

mihaibudiu: is SYSTEM_TIME even defined for a query?
Or just for a table?
If not, maybe the query should be rejected.

snuyanzin: in case of Flink it depends... It might be defined per query however then there is extra requirement of having primary key which is checked during planning.
It seems the behavior might be depending on downstream projects.

Then I'm thinking of having a `protected` method here which is failing by default in case of queries and downstream projects could override it if needed

core/src/test/java/org/apache/calcite/test/SqlValidatorTest.java
```diff
@@ -8782,6 +8782,12 @@ void testGroupExpressionEquivalenceParams() {
     sql("select stream * from orders left join products_temporal\n"
         + "for system_time as of CURRENT_TIMESTAMP\n"
         + "on orders.productid = products_temporal.productid").ok();
+
+    // verify left join of CTEs
```
mihaibudiu, line 8786: I think that this query is illegal according to the "standard" definition of SYSTEM_TIME.
You can make the method overridable, but is there even a definition of what it means to apply SYSTEM_TIME to the result of a query?
If you want it in here it should be applied to products_temporal.

snuyanzin: made it failing and overridable
>but is there even a definition of what it means to apply SYSTEM_TIME to the result of a query?

in Flink result of query in streaming mode is changelog. Changelog has timestamp, in this case we could jump at some position in the middle with SYSTEM_TIME

mihaibudiu: One way this can be resolved is to have a conformance flag (systemTimeForQueries?) which makes this query legal, and then disable this check only if that conformance flag is enabled. Supposedly compiling for Flink would enable it.

PR #4390, opened 2025-05-19
[CALCITE-7022] Decouple ModelHandler from CalciteConnection

zabetak:
ModelHandler is mainly a parser that can transform YAML/JSON files into SchemaPlus objects.

Currently, the class requires a CalciteConnection in order to be instantiated and the latter is a pretty heavyweight object. The usage of the CalciteConnection makes it rather cumbersome to instantiate and use the parsing capabilities of the ModelHander independently.

In reality though, the handler does not need much from the CalciteConnection and we could easily refactor the handler to not depend on the connection at all.

The motivation for this change is to facilitate the creation of schema objects from YAML/JSON files.


asolimando (approved):
LGTM

NobiGo (approved):

